PR #15, opened 2021-12-01
fix missing abort_signal in call to do_validation

yanchengnv:
Added the missing abort_signal to the call of self.do_validation().


yanchengnv:
Looks like Adil has a pending PR that fixed this.

PR #31, opened 2021-12-03
Add tensorflow requirement

madil90:


test/app_testing/run_app_tests.sh
```diff
@@ -1,3 +1,4 @@
 
+pip install tensorflow
 
 python test_runner.py --poc ../../nvflare/poc --n_clients 2 --yaml test_apps/test_simple.yml --app_path test_apps --cleanup
```
YuanTingHsieh, line 4: Not sure if we need to fix this warning "No newline at end of file."

madil90: I don't think formatting rules apply to sh files

IsaacYangSLA: I also think it doesn't matter.

PR #37, opened 2021-12-03
Update streaming widget to use just function but not component

YuanTingHsieh:
- Update streaming widget to use function 
- Update tb_streaming app in test_apps


IsaacYangSLA (dismissed):
The only thing that might need another refactor is the EVENT="log_analytics" because the json is using the string.

IsaacYangSLA:
@YuanTingHsieh , can you check https://docs.github.com/en/authentication/managing-commit-signature-verification/about-commit-signature-verification and have the commit signed?

YuanTingHsieh:
> The only thing that might need another refactor is the EVENT="log_analytics" because the json is using the string.

What do we want about this?

IsaacYangSLA (approved):
Thanks for this PR.  LGTM.

PR #44, opened 2021-12-07
Add AnalyticsSender and enhance streaming widgets

YuanTingHsieh:
- Rename DataType, Data to AnalyticsDataType and AnalyticsData in nvflare/apis/analytix.py
- Add AnalyticsSender which implements the following functions:

    - info, debug, error, exception, warning, critical following the method signature of Python logger
    - add_scalar, add_scalars, add_text, add_image following the method signature of PyTorch SummaryWriter
- Add a lock for save action in AnalyticsReceiver to ensure thread safety
- Update docstrings

### DIFF
--- a/nvflare/app_common/widgets/streaming.py
+++ b/nvflare/app_common/widgets/streaming.py
@@ -13,29 +13,36 @@
 # limitations under the License.
 
 from abc import ABC, abstractmethod
+from threading import Lock
 from typing import List, Optional
 
-from nvflare.apis.analytix import Data as AnalyticsData
-from nvflare.apis.analytix import DataType as AnalyticsDataType
+from nvflare.apis.analytix import AnalyticsData, AnalyticsDataType
 from nvflare.apis.dxo import DXO
 from nvflare.apis.event_type import EventType
 from nvflare.apis.fl_component import FLComponent
-from nvflare.apis.fl_constant import FedEventHeader, FLContextKey, ReservedKey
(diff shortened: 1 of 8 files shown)


nvflare/app_common/widgets/streaming.py
```diff
@@ -170,13 +268,14 @@ def handle_event(self, event_type: str, fl_ctx: FLContext):
             record_origin = fl_ctx.get_identity_name()
 
             # if fed event use peer name to save
-            if data.get_header(FedEventHeader.ORIGIN) is not None:
+            if fl_ctx.get_prop(FLContextKey.EVENT_SCOPE) == EventScope.FEDERATION:
                 peer_name = data.get_peer_prop(ReservedKey.IDENTITY_NAME, None)
                 record_origin = peer_name
```
IsaacYangSLA, line 273: These lines are OK.  It might be more clear if we say
if fl_ctx.get_prop(....) == ... :
  record_origin = data.get_peer_prop(....)
else:
  record_origin = fl_ctx.get_identity_name()

IsaacYangSLA (approved):
I am OK with the codes.  Just one comment about how to clearly implement the logic in a few lines.

nvflare/app_common/widgets/streaming.py
IsaacYangSLA: No need to submit another commit.  I will mark this as resolved.

PR #47, opened 2021-12-08
Remove learner_service_provider_spec.py.

yhwen:
Remove the no use learner_service_provider_spec.py


IsaacYangSLA (approved):
This is an unused file.  Safe to remove it.

PR #66, opened 2021-12-11
Update built docs

nvkevlu:


nvkevlu:
I had asked Isaac about that before and he said it was okay to include. It does not seem necessary though and I can remove in future docs builds.

PR #77, opened 2021-12-17
Fix TensorBoard streaming app test's result validator

### LINKED ISSUE #76: Update TensorBoard streaming test_apps
Since fed_event does not guaranteed to be sent, we should not enforce that the server side has client TensorBoard files.

YuanTingHsieh:
Fix #76


yanchengnv (approved):
Looks good.

nvkevlu (approved):

Issue #78, opened 2021-12-17
NVFlare python version not compatible with Google colab or Google Vertex AI Notebooks


filipmu:
NVFlare requires python 3.8.10 or higher per the pypi page, and both Google Colab and Google Vertex AI Notebooks currently run python 3.7.12 and 3.7.10 respectively.  Upgrading these environments is relatively undocumented and complex.  

For reference PyTorch 1.10 works with python 3.5 or greater.

Can the dependencies on python 3.8.10 be reduced so that python 3.7 will suffice?

nvkevlu:
Unfortunately right now NVFlare does not work on python 3.7 and there are currently no plans to reduce the dependencies on python 3.8 although there are plans to make NVFlare compatible with python 3.9 in the future. Hopefully Google Colab and Google Vertex AI can be upgraded to run on python 3.8 soon.

filipmu:
OK, that's too bad.  Its going to be hard to influence Google's plans.  Also, this makes it difficult to use nvflare on Jetson, since the NVIDIA L4T ML container on NGC is on python 3.6.

I tried building with 3.7 and found 12 files that had issues - walrus operators and fstring formatting were flagged as syntax errors.  Are there other behind the scenes capabilities used from python 3.8?

nvkevlu:
We will look into this and see what we can do about it.

YuanTingHsieh:
I forgot to mention, thanks for raising this good issue!

YuanTingHsieh:
Closing this issue as the original question "Can the dependencies on python 3.8.10 be reduced so that python 3.7 will suffice?" is resolved.

The answer is yes now.

PR #79, opened 2021-12-21
Remove = in format string and remove walrus operator for supporting Python 3.7

YuanTingHsieh:
Related to #78.

- Remove "=" in f-string
- Remove walrus operator

### DIFF
--- a/nvflare/apis/impl/controller.py
+++ b/nvflare/apis/impl/controller.py
@@ -42,9 +42,9 @@ _TASK_KEY_DONE = "___done"
 
 def _check_positive_int(name, value):
     if not isinstance(value, int):
-        raise TypeError(f"{name} must be an instance of int.")
+        raise TypeError("{} must be an instance of int.".format(name))
     if value < 0:
-        raise ValueError(f"{name} must >= 0.")
+        raise ValueError("{} must >= 0.".format(name))
 
 
 def _check_inputs(task: Task, fl_ctx: FLContext, targets: Union[List[Client], List[str], None]):
@@ -79,7 +79,7 @@ class Controller(Responder, ControllerSpec, ABC):
         self._task_check_period = task_check_period
 
     def initialize_run(self, fl_ctx: FLContext):
-        """Called by runners to initailze controller with information in fl_ctx
+        """Called by runners to initialize controller with information in fl_ctx
 
         Note: Controller subclasses must not overwrite this method.
 
@@ -145,30 +145,30 @@ class Controller(Responder, ControllerSpec, ABC):
 
         client_task_to_send = None
         with self._task_lock:
-            self.logger.debug(f"{self._tasks=}")
+            self.logger.debug("self._tasks: {}".format(self._tasks))
             for task in self._tasks:
-                assert isinstance(task, Task)
-
                 if task.completion_status is not None:
                     # this task is finished (and waiting for the monitor to exit it)
                     continue
 
                 # do we need to send this task to this client?
                 # note: the task could be sent to a client multiple times (e.g. in relay)
-                # we only check the last ClientTask sent to the client!
+                # we only check the last ClientTask sent to the client
                 client_task_to_check = task.last_client_task_map.get(client.name, None)
-                self.logger.debug(f"task to check {client_task_to_check=}")
+                self.logger.debug("client_task_to_check: {}".format(client_task_to_check))
                 resend_task = False
 
                 if client_task_to_check is not None:
-                    # this client has been sent the task already.
-                    # do we need to send again?
+                    # this client has been sent the task already
                     if not isinstance(client_task_to_check, ClientTask):
-                        raise TypeError(f"{client_task_to_check=} must be an instance of ClientTask.")
+                        raise TypeError(
+                            "client_task_to_check must be an instance of ClientTask, but got {}".format(
+                                type(client_task_to_check)
+                            )
+                        )
                     if client_task_to_check.result_received_time is None:
-                        # we have not received result
-                        # the client must be working on this task, but apparently something is wrong
-                        # we resend the task!
+                        # controller has not received result from client
+                        # something wrong happens when client working on this task, so resend the task
                         resend_task = True
                         client_task_to_send = client_task_to_check
                         fl_ctx.set_prop(FLContextKey.IS_CLIENT_TASK_RESEND, True, sticky=False)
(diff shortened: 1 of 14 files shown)


nvflare/apis/impl/seq_relay_manager.py
```diff
@@ -77,7 +77,7 @@ def check_task_send(self, client_task: ClientTask, fl_ctx: FLContext) -> TaskChe
 
         # adjust client window
         win_start_idx, win_end_idx = self._determine_window(task)
-        self.logger.debug(f"{win_start_idx=} {win_end_idx=}")
+        self.logger.debug("win_start_idx {}, win_end_idx {}".format(win_start_idx, win_end_idx))
```
madil90, line 80: This will print it out as win_start_idx 2, win_end_idx 3 instead of win_start_idx=2. We should add the = or : in string

nvflare/apis/impl/seq_relay_manager.py
```diff
@@ -86,7 +86,7 @@ def check_task_send(self, client_task: ClientTask, fl_ctx: FLContext) -> TaskChe
         for i in range(win_start_idx, win_end_idx):
             if client_name == task.targets[i]:
                 # this client is in the window!
-                self.logger.debug(f"last_send_idx={i}")
+                self.logger.debug("last_send_idx={}".format(i))
```
madil90, line 89: This is not an issue (just a mention) but I don't think it's using walrus operator (just a normal f string)

nvflare/apis/impl/seq_relay_manager.py
```diff
@@ -105,7 +105,7 @@ def _determine_window(self, task: Task) -> Tuple[int, int]:
         if last_send_idx >= 0:
             # see whether the result has been received
             last_task = task.last_client_task_map[task.targets[last_send_idx]]
-            self.logger.debug(f"{last_task=}")
+            self.logger.debug("last_task {}".format(last_task))
```
madil90, line 108: Missing = or : here

nvflare/apis/impl/seq_relay_manager.py
```diff
@@ -115,21 +115,29 @@ def _determine_window(self, task: Task) -> Tuple[int, int]:
                     # we give up on this client and move to the next target
                     win_start_idx = last_send_idx + 1
                     win_start_time = last_task.task_sent_time + task_result_timeout
-                    self.logger.debug(f"client task result timed out {win_start_idx=} {win_start_time=}")
+                    self.logger.debug(
+                        "client task result timed out. win_start_idx {}, win_start_time {}".format(
```
madil90, line 119: Same issue here

setup.py
```diff
@@ -66,7 +66,7 @@
     ],
     long_description=long_description,
     long_description_content_type="text/markdown",
-    python_requires=">=3.8",
+    python_requires=">=3.7",
```
madil90, line 69: if nvflare doesn't work in python 3.9+, we should add a python_requires<3.9 also

nvflare/app_common/workflows/cyclic_ctl.py
```diff
@@ -118,7 +118,7 @@ def control_flow(self, abort_signal: Signal, fl_ctx: FLContext):
                     abort_signal=abort_signal,
                 )
                 self.persistor.save(self.last_learnable, fl_ctx)
-                self.log_debug(fl_ctx, f"Ending {current_round=}.")
+                self.log_debug(fl_ctx, "Ending current round {}.".format(current_round))
```
madil90, line 121: missing = or :

nvflare/app_common/workflows/cyclic_ctl.py
```diff
@@ -91,7 +91,7 @@ def control_flow(self, abort_signal: Signal, fl_ctx: FLContext):
                 if abort_signal.triggered:
                     return
 
-                self.log_debug(fl_ctx, f"Starting {current_round=}.")
+                self.log_debug(fl_ctx, "Starting current round {}.".format(current_round))
```
madil90, line 94: missing : or =

madil90 (changes requested):
We should add : or = to maintain the same level of print clarity as before.

setup.py
YuanTingHsieh: that I don't know though, should we add that to be safe?

nvflare/apis/impl/seq_relay_manager.py
YuanTingHsieh: yeah it is an f-string, just change it to use format if we want to support 3.5 later.

setup.py
YuanTingHsieh: @IsaacYangSLA @yhwen might know this

nvflare/apis/impl/controller.py
```diff
@@ -179,41 +179,37 @@ def process_task_request(self, client: Client, fl_ctx: FLContext) -> Tuple[str,
                     if client_task_to_check is None:
                         client_task_to_check = ClientTask(task=task, client=client)
                     check_status = manager.check_task_send(client_task_to_check, fl_ctx)
-                    self.logger.debug(f"checking client task {client_task_to_check} {client_task_to_check.client.name}")
```
nvkevlu, line 182: client_task_to_check.client.name is now no longer displayed, is that okay?

nvflare/apis/impl/controller.py
```diff
@@ -628,7 +616,8 @@ def abort_task(self, task, fl_ctx: FLContext):
 
     def _end_task(self, task_names, fl_ctx: FLContext):
         engine = fl_ctx.get_engine()
-        assert isinstance(engine, ServerEngineSpec)
+        if not isinstance(engine, ServerEngineSpec):
+            raise TypeError("engine should be an instance of ServerEngineSpec, but get {}".format(type(engine)))
```
nvkevlu, line 620: "get" should be "got" here since the error is being raised after the wrong engine type was gotten

nvkevlu (dismissed):

nvflare/apis/impl/controller.py
YuanTingHsieh: i think that is OK, but I added it back.

nvflare/apis/impl/controller.py
YuanTingHsieh: fixed

setup.py
YuanTingHsieh: I think we could verify (and possibly fix) this in another PR what do you think?

nvkevlu: Makes sense, this PR is just about 3.7

nvkevlu (approved):

IsaacYangSLA:
In addition to f-string and walrus operators, we are not sure if other places have 3.7 incompatible behaviors.  Since this is the first step to make NVFlare compatible with 3.7, and all f-string {variable=} and walrus ops are removed, this PR should be OK.

IsaacYangSLA (approved):
In the future, maybe we can use only one of the two syntax f"variable1={variable1}" and "variable1={}".format(variable1).

setup.py
IsaacYangSLA: I think since we have really tried nvflare in 3.9 or 3.10, we should use
python_requires=">=3.7,<3.9"

YuanTingHsieh:
Thanks for the review!

@madil90 can you review and approve the PR, thanks!

PR #85, opened 2021-12-28
update the statement about Python version compatibility

### LINKED ISSUE #84: Document page states NVFlare only compatible with one single Python version
Previously, NVFlare 1.X was compatible with (and ran on) Python 3.8.10 due to the pip package was released with pyc files only.  Those pyc files were compiled by Python 3.8.10 interpreter and thus must run in Python 3.8.10 environment.
In NVFlare 2.x, the pip packages are source codes, in stead of pyc files.  Therefore, the original statement may cause confusion.

image

IsaacYangSLA:
Closes #84 .

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -6,7 +6,7 @@ Installation
 Python Version
 --------------
 
-NVIDIA FLARE requires `Python 3.8.10 <https://www.python.org/downloads/release/python-3810/>`_.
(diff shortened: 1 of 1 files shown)


nvkevlu (approved):

YuanTingHsieh (approved):
LGTM

YuanTingHsieh:
Note that it may work with Python 3.7 after PR #79 is merged.

PR #86, opened 2021-12-31
Fixed the args.log_config setting.

yhwen:


nvkevlu (approved):

nvkevlu (approved):

IsaacYangSLA (approved):
The updated portion is good.  However, it seemed we piggybacked some information into args attributes while we can use another variable.

YuanTingHsieh:
Yeah, if we need this log_config in args then the current code is good. (``args.log_config``)
If we don't need it in args, then we just use a new variable.  (for example: ``app_log_config``)

yhwen:
> Yeah, if we need this log_config in args then the current code is good. (`args.log_config`) If we don't need it in args, then we just use a new variable. (for example: `app_log_config`)

args.log_config has been used outside as a variable to hold the log config.

nvkevlu (approved):

PR #99, opened 2022-01-06
Yield CUDA_VISIBLE_DEVICES settings to users

### LINKED ISSUE #98: CUDA_VISIBLE_DEVICES setting
Users should be able to control that setting outside start.sh/sub_start.sh

IsaacYangSLA:
This will close #98 .

### DIFF
--- a/nvflare/lighter/impl/master_template.yml
+++ b/nvflare/lighter/impl/master_template.yml
@@ -334,7 +334,6 @@ sub_start_cln_sh: |
 sub_start_svr_sh: |
   #!/usr/bin/env bash
   DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
-  export CUDA_VISIBLE_DEVICES=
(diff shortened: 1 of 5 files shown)


holgerroth (approved):

yanchengnv:
What is the motivation of this PR? How would the user specify CUDA device settings?

IsaacYangSLA:
> What is the motivation of this PR? How would the user specify CUDA device settings?

User will run start.sh like the following:

EXPORT CUDA_VISIBLE_DEVICES=0,1 start.sh

The original code of start.sh/sub_start.sh actually resets that env var so users can not specify that outside start.sh.  Holger has a good use case that requires specifying it like the above EXPORT statement (https://github.com/NVIDIA/NVFlare/blob/5316ed4a69328535cc58387ede1e69837f6047d8/examples/cifar10/run_secure.sh#L25).

yanchengnv:
We need to update all examples with this extra piece of information. Otherwise the examples may fail.

yhwen (approved):

PR #104, opened 2022-01-07
Added the missing config_validator parsing.

yhwen:
To fix the missing config_validator parsing.

### DIFF
--- a/nvflare/private/fed/app/fl_conf.py
+++ b/nvflare/private/fed/app/fl_conf.py
@@ -64,7 +64,7 @@ class FLServerStarterConfiger(JsonConfigurator):
         self.app_root = app_root
         self.server_config_file_name = server_config_file_name
 
-        self.config_validator = None
+        self.app_validator = None
         self.enable_byoc = False
 
     def start_config(self, config_ctx: ConfigContext):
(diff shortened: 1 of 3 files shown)


nvflare/private/fed/app/fl_conf.py
```diff
@@ -92,6 +92,10 @@ def process_config_element(self, config_ctx: ConfigContext, node: Node):
             self.enable_byoc = element
             return
 
+        if path == "config_validator" and isinstance(element, dict):
```
yanchengnv, line 95: Let's call it app_validator since it could go beyond just config files. Also the concept should be "app", not "config" which is too generic.

The provision also needs to change to call it "app_validator".

yanchengnv (changes requested):

yanchengnv (approved):

IsaacYangSLA (approved):
LGTM.

nvflare/private/fed/app/server/server_train.py
```diff
@@ -103,7 +103,7 @@ def main():
                 server_conf=first_server,
                 args=args,
                 secure_train=secure_train,
-                app_validator=deployer.config_validator,
+                app_validator=deployer.app_validator,
```
IsaacYangSLA, line 106: This line seems causing test failure.  deployer might not always have app_validator attribute.

IsaacYangSLA (changes requested):

yanchengnv (approved):

IsaacYangSLA (approved):

PR #106, opened 2022-01-07
Add option of app_validator

### LINKED ISSUE #105: Missing config_validator option in provisioning tool
Provisioning tool needs to have config_validator option so the generated fed_server.json can have that information.

IsaacYangSLA:
This will close #105.

### DIFF
--- a/nvflare/lighter/impl/static_file.py
+++ b/nvflare/lighter/impl/static_file.py
@@ -20,10 +20,11 @@ from nvflare.lighter.utils import sh_replace
 
 
 class StaticFileBuilder(Builder):
-    def __init__(self, enable_byoc=False, config_folder="", docker_image=""):
+    def __init__(self, enable_byoc=False, config_folder="", app_validator="", docker_image=""):
         self.enable_byoc = enable_byoc
         self.config_folder = config_folder
         self.docker_image = docker_image
+        self.app_validator = app_validator
 
     def _write(self, file_full_path, content, mode, exe=False):
         mode = mode + "w"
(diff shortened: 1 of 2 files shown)


nvflare/lighter/impl/static_file.py
```diff
@@ -20,10 +20,11 @@
 
 
 class StaticFileBuilder(Builder):
-    def __init__(self, enable_byoc=False, config_folder="", docker_image=""):
+    def __init__(self, enable_byoc=False, config_folder="", config_validator="", docker_image=""):
```
yanchengnv, line 23: Let's call it app_validator, instead of config_validator. App is the term we promote.

nvflare/lighter/project.yml
```diff
@@ -35,6 +35,9 @@ builders:
     args:
       # config_folder can be set to inform NVFlare where to get configuration
       config_folder: config
+      # config_validator is used to verify if uploaded app has proper structures
```
yanchengnv, line 38: Call it app_validator, not config_validator.

yanchengnv (changes requested):

nvflare/lighter/impl/static_file.py
IsaacYangSLA: Yes.  Just renamed this and other places one min ago.

nvflare/lighter/project.yml
IsaacYangSLA: Same as the above.  This was renamed.

yanchengnv (approved):

yhwen (approved):

PR #108, opened 2022-01-08
Added check to ensure the Learner returnCode is OK.

yhwen:


nvflare/app_common/executors/learner_executor.py
```diff
@@ -85,7 +85,9 @@ def train(self, shareable: Shareable, fl_ctx: FLContext, abort_signal: Signal) -
         validate_result: Shareable = self.learner.validate(shareable, fl_ctx, abort_signal)
 
         train_result = self.learner.train(shareable, fl_ctx, abort_signal)
-        if not (train_result and isinstance(train_result, Shareable)):
+        if not (
+            train_result and isinstance(train_result, Shareable) and train_result.get_return_code() == ReturnCode.OK
+        ):
```
yanchengnv, line 90: In the case that the train_result is already a valid Shareable but not OK, we should just return it. Why do we want to change the return code to EMPTY_RESULT?

nvflare/app_common/executors/learner_executor.py
```diff
@@ -105,7 +107,11 @@ def train(self, shareable: Shareable, fl_ctx: FLContext, abort_signal: Signal) -
     def submit_model(self, shareable: Shareable, fl_ctx: FLContext) -> Shareable:
         model_name = shareable.get_header(AppConstants.SUBMIT_MODEL_NAME)
         submit_model_result = self.learner.get_model_for_validation(model_name, fl_ctx)
-        if submit_model_result and isinstance(submit_model_result, Shareable):
+        if (
+            submit_model_result
+            and isinstance(submit_model_result, Shareable)
+            and submit_model_result.get_return_code() == ReturnCode.OK
```
yanchengnv, line 113: Again, if the Learner returns a valid Shareable, we should honor it and not changing its return code. We could return (and skip further processing) if Learner's return code is not OK.

yanchengnv (changes requested):

nvflare/app_common/executors/learner_executor.py
yhwen: even though the returned is a valid Shareable, but "from_shareable(train_result)" may still fail, because the content may not be valid DXO. Maybe need further enhanced check here.

yanchengnv: The content doesn't have to contain a DXO, though we encourage people to do it. If the LearnerExecutor requires that (does it?), then it should check it explicitly, even if the return code is OK.

yhwen: Changed the check the "from_shareable(train_result)" explicitly.

holgerroth (approved):

yanchengnv (approved):

IsaacYangSLA (approved):

PR #118, opened 2022-01-18
avoid running unzip subprocess in poc.py

taleinat:
An unzip executable is not always available, and calling it introduces an unnecessary implicit dependency.

This PR uses `shutil.unpack_archive()` instead.


IsaacYangSLA (approved):
Thanks for the PR.

YuanTingHsieh (approved):
Thanks for this improvement.

LGTM.

PR #124, opened 2022-01-19
Increase fed event firing frequency

IsaacYangSLA:
Add ABOUT_TO_END_RUN to flush queue
Allow thread to exit only when queue is empty during END_RUN

### DIFF
--- a/nvflare/private/fed/client/client_aux_runner.py
+++ b/nvflare/private/fed/client/client_aux_runner.py
@@ -49,6 +49,7 @@ class ClientAuxRunner(AuxRunner):
             self.sender.start()
         elif event_type == EventType.END_RUN:
             self.asked_to_stop = True
+            self.log_debug(fl_ctx, "Received END_RUN")
             if self.sender and self.sender.is_alive():
                 self.sender.join()
 
@@ -79,7 +80,7 @@ class ClientAuxRunner(AuxRunner):
             # this is fire-and-forget request
             with self.fnf_lock:
                 self.fnf_requests.append(req_to_send)
-                return make_reply(ReturnCode.OK)
+            return make_reply(ReturnCode.OK)
 
         # send regular request
         engine = fl_ctx.get_engine()
@@ -104,23 +105,27 @@ class ClientAuxRunner(AuxRunner):
         sleep_time = 0.5
         while True:
             time.sleep(sleep_time)
-            if self.asked_to_stop or self.abort_signal.triggered:
+            if self.abort_signal.triggered:
                 break
 
-            with self.fnf_lock:
-                if len(self.fnf_requests) <= 0:
+            if len(self.fnf_requests) <= 0:
+                if self.asked_to_stop:
+                    break
+                else:
                     sleep_time = 1.0
                     continue
 
-                with self.engine.new_context() as fl_ctx:
-                    bulk = Shareable()
-                    bulk.set_header(ReservedHeaderKey.TOPIC, topic)
-                    bulk.set_peer_props(fl_ctx.get_all_public_props())
+            with self.engine.new_context() as fl_ctx:
+                self.log_debug(fl_ctx, "preparing bulk send")
+                bulk = Shareable()
+                bulk.set_header(ReservedHeaderKey.TOPIC, topic)
+                bulk.set_peer_props(fl_ctx.get_all_public_props())
+                with self.fnf_lock:
                     bulk[self.DATA_KEY_BULK] = self.fnf_requests
                     reply = self.engine.aux_send(topic=topic, request=bulk, timeout=1.0, fl_ctx=fl_ctx)
-
                     rc = reply.get_return_code()
                     if rc != ReturnCode.COMMUNICATION_ERROR:
+                        self.log_debug(fl_ctx, "aux bulk send returns no ERROR")
                         # if communication error we'll retry
                         self.fnf_requests = []
             sleep_time = 0.5
--- a/nvflare/private/fed/client/client_aux_runner.py
+++ b/nvflare/private/fed/client/client_aux_runner.py
@@ -49,7 +49,6 @@ class ClientAuxRunner(AuxRunner):
             self.sender.start()
         elif event_type == EventType.END_RUN:
             self.asked_to_stop = True
-            self.log_debug(fl_ctx, "Received END_RUN")
             if self.sender and self.sender.is_alive():
                 self.sender.join()
 
@@ -116,7 +115,6 @@ class ClientAuxRunner(AuxRunner):
                     continue
 
             with self.engine.new_context() as fl_ctx:
-                self.log_debug(fl_ctx, "preparing bulk send")
                 bulk = Shareable()
                 bulk.set_header(ReservedHeaderKey.TOPIC, topic)
                 bulk.set_peer_props(fl_ctx.get_all_public_props())
@@ -125,7 +123,6 @@ class ClientAuxRunner(AuxRunner):
                     reply = self.engine.aux_send(topic=topic, request=bulk, timeout=1.0, fl_ctx=fl_ctx)
                     rc = reply.get_return_code()
                     if rc != ReturnCode.COMMUNICATION_ERROR:
-                        self.log_debug(fl_ctx, "aux bulk send returns no ERROR")
                         # if communication error we'll retry
                         self.fnf_requests = []
             sleep_time = 0.5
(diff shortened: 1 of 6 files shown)


nvflare/widgets/fed_event.py
```diff
@@ -109,29 +116,60 @@ def _receive(self, topic: str, request: Shareable, fl_ctx: FLContext) -> Shareab
...
+                    event_to_post = self.in_events.pop(0)
+            elif self.asked_to_stop:
+                # the queue is empty and we are asdked to stop.  Give
+                # it self.grace_period seconds to wait, then
+                # exit.
+                if countdown < 0:
+                    break
+                else:
+                    countdown = countdown - 1
+                    time.sleep(1)
```
SYangster, line 150: add continue statement

IsaacYangSLA: Thanks for the review.  Just added the continue after time.sleep(1).

nvflare/widgets/fed_event.py
```diff
@@ -109,29 +116,60 @@ def _receive(self, topic: str, request: Shareable, fl_ctx: FLContext) -> Shareab
...
+            n = len(self.in_events)
+            if n > 0:
+                if self.asked_to_flush:
+                    sleep_time = 0
+                else:
+                    sleep_time = self.regular_interval
+                with self.in_lock:
+                    event_to_post = self.in_events.pop(0)
+            elif self.asked_to_stop:
+                # the queue is empty and we are asdked to stop.  Give
```
nvkevlu, line 143: asdked looks like a typo here

IsaacYangSLA: My bad.  Fixed in a new commit.

nvflare/private/fed/client/client_aux_runner.py
```diff
@@ -104,23 +105,27 @@ def _send_fnf_requests(self):
...
+                else:
                     sleep_time = 1.0
                     continue
 
-                with self.engine.new_context() as fl_ctx:
-                    bulk = Shareable()
-                    bulk.set_header(ReservedHeaderKey.TOPIC, topic)
-                    bulk.set_peer_props(fl_ctx.get_all_public_props())
+            with self.engine.new_context() as fl_ctx:
+                self.log_debug(fl_ctx, "preparing builk send")
```
nvkevlu, line 119: message should probably be "preparing bulk send"

IsaacYangSLA: Good catch.  Thanks.

nvflare/private/fed/client/client_aux_runner.py
```diff
@@ -104,23 +105,27 @@ def _send_fnf_requests(self):
         sleep_time = 0.5
         while True:
             time.sleep(sleep_time)
-            if self.asked_to_stop or self.abort_signal.triggered:
+            if self.abort_signal.triggered:
                 break
 
-            with self.fnf_lock:
-                if len(self.fnf_requests) <= 0:
+            if len(self.fnf_requests) <= 0:
```
YuanTingHsieh, line 111: Just to confirm this len function does not need to be inside the lock?

nvflare/widgets/fed_event.py
```diff
@@ -109,29 +116,61 @@ def _receive(self, topic: str, request: Shareable, fl_ctx: FLContext) -> Shareab
...
-                event_to_post = self.in_events.pop(0)
-                assert isinstance(event_to_post, Shareable)
-
-            if self.asked_to_stop or self.abort_signal.triggered:
-                break
+            else:
+                sleep_time = min(sleep_time * 2, 1)
+                continue
 
+            assert isinstance(event_to_post, Shareable)
```
YuanTingHsieh, line 156: I think we don't need this statement, as self.in_events is added by this class.

nvflare/widgets/fed_event.py
```diff
@@ -109,29 +116,61 @@ def _receive(self, topic: str, request: Shareable, fl_ctx: FLContext) -> Shareab
...
             with self.engine.new_context() as fl_ctx:
                 assert isinstance(fl_ctx, FLContext)
```
YuanTingHsieh, line 158: I think we don't need this line

nvflare/widgets/fed_event.py
```diff
@@ -109,29 +116,61 @@ def _receive(self, topic: str, request: Shareable, fl_ctx: FLContext) -> Shareab
...
-                if len(self.in_events) <= 0:
+            n = len(self.in_events)
+            if n > 0:
+                if self.asked_to_flush:
+                    sleep_time = 0
+                else:
+                    sleep_time = self.regular_interval
+                with self.in_lock:
+                    event_to_post = self.in_events.pop(0)
+            elif self.asked_to_stop:
```
YuanTingHsieh, line 142: if self.asked_to_stop is True,

should we stop adding events into self.in_events?

YuanTingHsieh (changes requested):
If ask to stop, I think server side should stop receiving events.

nvflare/private/fed/client/client_aux_runner.py
IsaacYangSLA: The with self.fnf_lock statement goes down a few lines to reduce locking time.  len is read-only so we can move it out of the self.fnf_lock.

nvflare/widgets/fed_event.py
IsaacYangSLA: This is from original codes.  But I agree this assert is not needed.

nvflare/widgets/fed_event.py
IsaacYangSLA: Same as above.

nvflare/widgets/fed_event.py
IsaacYangSLA: After discussing with @yhwen , we decided not to block event receiving because components which hasn't received END_RUN might still firing fed events.  We even wait two seconds to collect those events and fire them during END_RUN to flush the queue.

YuanTingHsieh: Got it, just a note that, for the "grace_period" we set.
We might wait a bit longer than that, because we only decrease the grace_period / countdown when length of in events is 0.

YuanTingHsieh (approved):
Thanks for the explanation, LGTM.

nvflare/private/fed/client/client_runner.py
```diff
@@ -337,7 +346,13 @@ def abort(self):
...
         self._abort_current_task()
         with self.engine.new_context() as fl_ctx:
-            self.fire_event(EventType.END_RUN, fl_ctx)
+            with self.end_run_lock:
+                if not self.end_run_fired:
+                    self.fire_event(EventType.ABOUT_TO_END_RUN, fl_ctx)
+                    self.log_info(fl_ctx, "ABOUT_TO_END_RUN fired")
+                    self.fire_event(EventType.END_RUN, fl_ctx)
+                    self.log_info(fl_ctx, "END_RUN fired")
+                    self.end_run_fired = True
```
yanchengnv, line 355: We still need to trigger self.run_abort_signal here; otherwise the try_run() won't break from the loop.

yanchengnv (changes requested):

nvflare/private/fed/client/client_aux_runner.py
```diff
@@ -104,23 +105,27 @@ def _send_fnf_requests(self):
         sleep_time = 0.5
         while True:
             time.sleep(sleep_time)
-            if self.asked_to_stop or self.abort_signal.triggered:
+            if self.abort_signal.triggered:
                 break
 
-            with self.fnf_lock:
```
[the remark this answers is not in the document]
IsaacYangSLA: We discussed it yesterday.  It should be OK to move the lock down on the bulk setting section.

nvflare/private/fed/client/client_runner.py
```diff
@@ -337,7 +339,10 @@ def abort(self):
         """
         self._abort_current_task()
```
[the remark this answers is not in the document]
YuanTingHsieh: I agree we should trigger `self.run_abort_signal` in this `abort` function.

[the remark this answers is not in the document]
IsaacYangSLA: Yes.  It's needed.

yanchengnv (approved):

YuanTingHsieh (approved):
Thanks. LGTM.

PR #129, opened 2022-01-20
Fix workflow for automatically building docs, improve docs

nvkevlu:
Fix workflow by removing apidocs from gitignore so it can be checked in with the built docs into docs branch automatically, and add docstring documentation for provisioning builders and link to them in docs.

### DIFF
--- a/nvflare/lighter/impl/auth_policy.py
+++ b/nvflare/lighter/impl/auth_policy.py
@@ -20,6 +20,18 @@ from nvflare.lighter.spec import Builder, Study
 
 class AuthPolicyBuilder(Builder):
     def __init__(self, orgs, roles, groups, disabled):
+        """Creates and writes authorization.json to the server's startup directory with the authorization policy
+        defining the groups each org is in and the admin client roles which controls the allowed rights. The
+        participant information from project.yml is included in authorization.json with what orgs, groups, and roles
+        are associated with each participant. This builder also checks for errors if the arguments are specified
+        incorrectly.
+
+        Args:
+            orgs: authorization configuration for orgs (it may be helpful to build this section with the UI)
+            roles: authorization configuration for roles (it may be helpful to build this section with the UI)
+            groups: authorization configuration for groups (it may be helpful to build this section with the UI)
+            disabled: if true, all users are super with all privileges
+        """
         self.orgs = orgs
         self.roles = roles
         self.groups = groups
--- a/nvflare/lighter/impl/he.py
+++ b/nvflare/lighter/impl/he.py
@@ -27,6 +27,15 @@ class HEBuilder(Builder):
         scale_bits=40,
         scheme="CKKS",
     ):
+        """Generates Tenseal homomorphic encryption context for server and client and writes them to server and client
+        participant folders.
+
+        Args:
+            poly_modulus_degree: defaults to 8192.
+            coeff_mod_bit_sizes: defaults to [60, 40, 40].
+            scale_bits: defaults to 40.
+            scheme: defaults to "CKKS".
+        """
 
         self._context = None
         self.scheme_type_mapping = {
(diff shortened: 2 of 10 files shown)


nvflare/lighter/impl/auth_policy.py
```diff
@@ -20,6 +20,18 @@
...
+        defining the groups each org is in and the admin client roles which controls the allowed rights. The
+        participant information from project.yml is included in authorization.json with what orgs, groups, and roles
+        are associated with each participant. This builder also checks for errors if the arguments are specified
+        incorrectly.
+
+        Args:
+            orgs: authorization configuration for orgs (it may be helpful to build this section with the UI)
+            roles: authorization configuration for roles (it may be helpful to build this section with the UI)
+            groups: authorization configuration for groups (it may be helpful to build this section with the UI)
+            disabled: if true, all users are super with all privileges
```
YuanTingHsieh, line 33: All looks good.

Just one question, what thing does this argument "disabled" disable?

nvflare/lighter/impl/he.py
```diff
@@ -27,6 +27,15 @@ def __init__(
...
         scheme="CKKS",
     ):
+        """Generates Tenseal homomorphic encryption context for server and client and writes them to server and client
+        participant folders.
+
+        Args:
+            poly_modulus_degree: defaults to 8192.
+            coeff_mod_bit_sizes: defaults to [60, 40, 40].
+            scale_bits: defaults to 40.
+            scheme: defaults to "CKKS".
```
YuanTingHsieh, line 37: What does each of these arguments mean?

@holgerroth can you help?

Thanks.

nvflare/lighter/impl/static_file.py
```diff
@@ -21,6 +21,20 @@
 
 class StaticFileBuilder(Builder):
     def __init__(self, enable_byoc=False, config_folder="", app_validator="", docker_image=""):
+        """Uses the information from project.yml through study to go through the participants and write the contents of
+        each file with the template, and replacing with the appropriate values from project.yml.
+
+        Usually, two main categories of files are created in all FL participants, static and dynamic. Static files
+        have similar contents among different participants, with small differences.  For example, the differences in
+        sub_start.sh are client name and python module.  Those are basically static files.  This builder uses template
+        file and string replacement to generate those static files for each participant.
```
YuanTingHsieh, line 30: Do we have an explanation of a dynamic file?
Like a builder or something, or other than static file are all dynamic files?

nvflare/lighter/impl/template.py
```diff
@@ -20,6 +20,10 @@
 
 
 class TemplateBuilder(Builder):
+    """Loads the content of the template_file and the authz_def (section of template file with fixed authorization
+    definitions) into two key-value pairs in the build context.
+    """
+
```
YuanTingHsieh, line 26: Why do we need this class though?

For me this is more like a loader rather than "builder".

nvflare/lighter/impl/auth_policy.py
IsaacYangSLA: If disabled, this builder just enters the 'researcher' mode and makes everyone with role = super.  That is, it disables the authorization portion so everyone has the highest rights.

nvflare/lighter/impl/template.py
IsaacYangSLA: This is load the template file.  It's possible other developers have different template files and need to load them differently.  Also, because it only implements the initialize method, it's not a full 'builder' per se.

nvflare/lighter/impl/static_file.py
IsaacYangSLA: Static files are derivatives from templates.  Dynamic files are usually generated by other builders, with their 'build' methods.  For example, CertBuilder writes all certs during build.  Those certs are dynamic files.

nvflare/lighter/impl/he.py
IsaacYangSLA: I think those parameters are from tenseal package.  The parameter names should be identical to tenseal.  The actual meanings might be somewhere in the paper.  It should be safe to assume users who would like to change those parameters have already read the paper.

IsaacYangSLA (approved):
Thanks for adding docstring on lighter's builders.

YuanTingHsieh (approved):
Thanks for the PR.

And thanks @IsaacYangSLA  for the explanation.
We could improve the documentation later to include Isaac's explanation

PR #133, opened 2022-01-21
Removed the use of tmp folders for the server and client.

yhwen:
To remove the "/tmp/fl_server" and "/tmp/fl/uid" workspace dependency for FL server and client.

### DIFF
--- a/nvflare/private/fed/app/client/worker_process.py
+++ b/nvflare/private/fed/app/client/worker_process.py
@@ -35,6 +35,7 @@ from nvflare.private.fed.client.command_agent import CommandAgent
 def main():
     parser = argparse.ArgumentParser()
     parser.add_argument("--workspace", "-m", type=str, help="WORKSPACE folder", required=True)
+    parser.add_argument("--startup", "-w", type=str, help="startup folder", required=True)
 
     parser.add_argument(
         "--fed_client", "-s", type=str, help="an aggregation server specification json file", required=True
@@ -88,17 +89,17 @@ def main():
                 )
             )
 
-        workspace = os.path.join("/tmp/fl", client_name)
+        startup = args.startup
         app_root = os.path.join(args.workspace, "run_" + str(run_number), "app_" + client_name)
 
         app_log_config = os.path.join(app_root, config_folder, "log.config")
         if os.path.exists(app_log_config):
             args.log_config = app_log_config
         else:
-            args.log_config = os.path.join(workspace, "log.config")
+            args.log_config = os.path.join(startup, "log.config")
 
         conf = FLClientStarterConfiger(
-            app_root=workspace,
+            app_root=startup,
             client_config_file_name=args.fed_client,
             log_config_file_name=args.log_config,
             kv_list=args.set,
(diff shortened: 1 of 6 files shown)


IsaacYangSLA (approved):
LGTM.

nvflare/private/fed/app/client/worker_process.py
```diff
@@ -88,17 +89,17 @@ def main():
                 )
             )
 
-        workspace = os.path.join("/tmp/fl", client_name)
+        startup = args.startup
```
yanchengnv, line 92: What's the difference btw this startup and the startup on line 75? If they are the same, then we shouldn't need this new arg for startup.

nvflare/private/fed/app/client/worker_process.py
```diff
@@ -88,17 +89,17 @@ def main():
...
         app_log_config = os.path.join(app_root, config_folder, "log.config")
         if os.path.exists(app_log_config):
             args.log_config = app_log_config
         else:
-            args.log_config = os.path.join(workspace, "log.config")
+            args.log_config = os.path.join(startup, "log.config")
 
         conf = FLClientStarterConfiger(
-            app_root=workspace,
+            app_root=startup,
```
yanchengnv, line 102: Maybe the only arg needed is the "workspace". Since its structure is standard, all other files can be found there following the folder structure.

nvflare/private/fed/client/client_engine.py
```diff
@@ -45,7 +45,7 @@ def __init__(self, client, client_name, sender, args, rank, workers=5):
         self.args = args
         self.rank = rank
         self.client.process = None
-        self.client_executor = ProcessExecutor(client.client_name)
+        self.client_executor = ProcessExecutor(client.client_name, os.path.join(args.workspace, "startup"))
```
yanchengnv, line 48: Should we pass args.workspace to ProcessExecutor? It should have access to the whole workspace.

nvflare/private/fed/client/client_executor.py
```diff
@@ -29,8 +29,8 @@
 
 
 class ClientExecutor(object):
-    def __init__(self, uid) -> None:
-        pipe_path = "/tmp/fl/" + uid + "/comm"
+    def __init__(self, uid, startup) -> None:
+        pipe_path = startup + "/comm"
```
yanchengnv, line 33: Should use os.path.join() here.

nvflare/private/fed/client/client_executor.py
```diff
@@ -123,9 +123,10 @@ class ProcessExecutor(ClientExecutor):
     Run the Client executor in a child process.
     """
 
-    def __init__(self, uid):
-        ClientExecutor.__init__(self, uid)
+    def __init__(self, uid, startup):
```
yanchengnv, line 126: Use workspace instead of just startup?

nvflare/private/fed/client/client_executor.py
```diff
@@ -171,6 +172,7 @@ def start_train(self, client, args, app_root, app_custom_folder, listen_port):
         command = (
             f"{sys.executable} -m nvflare.private.fed.app.client.worker_process -m "
             + args.workspace
+            + " -w " + self.startup
```
yanchengnv, line 175: Isn't self.startup already a subfolder of args.workspace?

yanchengnv (changes requested):

IsaacYangSLA:
@SYangster could duplicate the issue on 2.0.5 and didn't see it on this PR.

yanchengnv (approved):
Looks like Yuhong won't be available to address my questions. Let's approve this and get it tested thoroughly.

IsaacYangSLA (approved):
Approved to be ready for QA.

PR #135, opened 2022-01-21
Update docstrings and clean up streaming

### LINKED ISSUE #134: Update streaming widgets docstrings and clean up
The docstrings of the streaming widgets needs to be updated, this is a follow up of https://github.com/NVIDIA/NVFlare/pull/117 .

YuanTingHsieh:
Fix #134 

- Rename log_receiver.py to log_streaming.py
- Move LogSender into log_streaming.py and rename to LogAnalyticsSender
- Update streaming.py docstrings and removed unused codes
- Update unit tests

### DIFF
--- a/nvflare/app_common/widgets/log_streaming.py
+++ b/nvflare/app_common/widgets/log_streaming.py
@@ -15,15 +15,53 @@
 import logging
 import os
 from logging import LogRecord
+from threading import Lock
 from typing import List, Optional
 
 from nvflare.apis.analytix import AnalyticsData, AnalyticsDataType
 from nvflare.apis.dxo import from_shareable
+from nvflare.apis.event_type import EventType
 from nvflare.apis.fl_constant import LogMessageTag
 from nvflare.apis.fl_context import FLContext
 from nvflare.apis.shareable import Shareable
 from nvflare.app_common.app_event_type import AppEventType
-from nvflare.app_common.widgets.streaming import AnalyticsReceiver
+from nvflare.app_common.widgets.streaming import AnalyticsReceiver, create_analytic_dxo, send_analytic_dxo
+from nvflare.widgets.widget import Widget
+
+
+class LogAnalyticsSender(Widget, logging.StreamHandler):
+    def __init__(self, log_level=""):
+        """Sends the log record.
+
+        Args:
+            log_level: log_level threshold
+        """
+        Widget.__init__(self)
+        logging.StreamHandler.__init__(self)
+        self.log_level = getattr(logging, log_level, logging.INFO)
+        self.engine = None
+
+    def handle_event(self, event_type: str, fl_ctx: FLContext):
+        if event_type == EventType.ABOUT_TO_START_RUN:
+            self.engine = fl_ctx.get_engine()
+            logging.root.addHandler(self)
+        elif event_type == EventType.END_RUN:
+            logging.root.removeHandler(self)
+
+    def emit(self, record: LogRecord):
+        """Sends the log record.
+
+        When the log_level higher than the configured level, sends the log record.
+        Args:
+            record: logging record
+        """
+        if record.levelno >= self.log_level and self.engine:
+            dxo = create_analytic_dxo(
+                tag=LogMessageTag.LOG_RECORD, value=record, data_type=AnalyticsDataType.LOG_RECORD
+            )
+            with self.engine.new_context() as fl_ctx:
+                send_analytic_dxo(self, dxo=dxo, fl_ctx=fl_ctx, event_type=AppEventType.LOGGING_EVENT_TYPE)
+            self.flush()
 
 
 class LogAnalyticsReceiver(AnalyticsReceiver):
@@ -40,6 +78,7 @@ class LogAnalyticsReceiver(AnalyticsReceiver):
         self.formatter = formatter
 
         self.handlers = {}
+        self.handlers_lock = Lock()
 
     def initialize(self, fl_ctx: FLContext):
         workspace = fl_ctx.get_engine().get_workspace()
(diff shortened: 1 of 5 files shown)


YuanTingHsieh:
One thing to discuss here, if we don't want private to import from app_common, we should promote the streaming into apis.

yhwen:
> One thing to discuss here, if we don't want private to import from app_common, we should promote the streaming into apis.

We should promote the log_streaming.py to the nvflare/widgets, and leave the LocalLogger in the apis/utils.

nvflare/app_common/widgets/log_streaming.py
```diff
@@ -14,16 +14,54 @@
...
+        Widget.__init__(self)
+        logging.StreamHandler.__init__(self)
+        self.log_level = getattr(logging, log_level, logging.INFO)
+        self.engine = None
+
+    def handle_event(self, event_type: str, fl_ctx: FLContext):
+        if event_type == EventType.ABOUT_TO_START_RUN:
+            self.engine = fl_ctx.get_engine()
+            logging.root.addHandler(self)
+        if event_type == EventType.END_RUN:
```
yanchengnv, line 48: Should be "elif".

nvflare/app_common/widgets/log_streaming.py
```diff
@@ -14,16 +14,54 @@
...
+        if record.levelno >= self.log_level and self.engine:
+            dxo = create_analytic_dxo(
+                tag=LogMessageTag.LOG_RECORD, value=record, data_type=AnalyticsDataType.LOG_RECORD
+            )
+            with self.engine.new_context() as fl_ctx:
+                send_analytic_dxo(self, dxo=dxo, fl_ctx=fl_ctx, event_type=AppEventType.LOGGING_EVENT_TYPE)
+            self.flush()
 
 
 class LogAnalyticsReceiver(AnalyticsReceiver):
```
yanchengnv, line 67: The handlers creation is not thread-safe. Even though usually the Receiver won't be running in multiple threads, it's not guaranteed (e.g. the Receiver could be used for both local and fed events).
Need to add a lock to ensure thread safety.

nvflare/private/fed/client/communicator.py
```diff
@@ -27,8 +27,8 @@
 from nvflare.apis.fl_constant import FLContextKey
 from nvflare.apis.fl_context import FLContext
 from nvflare.apis.fl_exception import FLCommunicationError
-from nvflare.apis.utils.local_logger import LocalLogger
-from nvflare.private.fed.utils.fed_utils import shareable_to_modeldata, make_context_data, make_shareeable_data
+from nvflare.app_common.widgets.log_streaming import LocalLogger
```
yanchengnv, line 30: We have a dependency violation here - private (which is part of the core) must not depend on app_common!
Looks like LocalLogger needs to be moved into the "core". We probably need to create a "local root logger" before firing ABOUT_TO_START_RUN (which will cause the LogSender to be added to the root logger).

nvflare/private/fed/client/fed_client_base.py
```diff
@@ -27,8 +27,9 @@
 from nvflare.apis.fl_exception import FLCommunicationError
 from nvflare.apis.shareable import Shareable
 from nvflare.apis.signal import Signal
-from nvflare.apis.utils.local_logger import LocalLogger
+from nvflare.app_common.widgets.log_streaming import LocalLogger
```
yanchengnv, line 30: Dependency violation.

yanchengnv (changes requested):

nvflare/private/fed/client/communicator.py
YuanTingHsieh: Thanks, this will be solved by PR #139.

YuanTingHsieh: I will rebase after that is merged.

nvflare/private/fed/client/fed_client_base.py
YuanTingHsieh: Thanks, this will be solved by PR #139, I will update this PR after that is merged.

YuanTingHsieh:
Thanks @yanchengnv for the review, I've addressed the issues.

nvflare/app_common/widgets/log_streaming.py
```diff
@@ -57,6 +96,8 @@ def save(self, fl_ctx: FLContext, shareable: Shareable, record_origin: str):
                 handler = self.handlers.get(record_origin)
```
yanchengnv, line 96: I think the lock should be placed before self.handlers.get().

yanchengnv (changes requested):

nvflare/app_common/widgets/log_streaming.py
YuanTingHsieh: Yes, it should.
I've fixed this issue.

yanchengnv (approved):

IsaacYangSLA (approved):
Approved for release to QA testing.

nvflare/app_common/widgets/log_streaming.py
```diff
@@ -19,11 +19,48 @@
...
 from nvflare.apis.fl_constant import LogMessageTag
 from nvflare.apis.fl_context import FLContext
 from nvflare.apis.shareable import Shareable
 from nvflare.app_common.app_event_type import AppEventType
-from nvflare.app_common.widgets.streaming import AnalyticsReceiver
+from nvflare.app_common.widgets.streaming import AnalyticsReceiver, create_analytic_dxo, send_analytic_dxo
+from nvflare.widgets.widget import Widget
+
+
+class LogAnalyticsSender(Widget, logging.StreamHandler):
```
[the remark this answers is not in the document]
yhwen: This does not apply any more with LogAnalyticsSender change.

yhwen (approved):

PR #137, opened 2022-01-21
Find the open port when running MultiProcessExecutor.

yhwen:


nvflare/app_common/pt/pt_multi_process_executor.py
```diff
@@ -26,5 +27,14 @@ def get_multi_process_command(self) -> str:
...
             + " --nnodes=1 --node_rank=0"
-            + ' --master_addr="localhost" --master_port=1234'
+            + ' --master_addr="localhost" --master_port='
+            + str(self._get_open_port())
         )
+
+    def _get_open_port(self):
+        s = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+        s.bind(("", 0))
+        s.listen(1)
```
IsaacYangSLA, line 37: Not sure if listen is necessary.  The kernel seems able to find a free port after trying to bind to port 0.  Other than that, it's perfect.

yhwen: bind() will find the open port. listen() is to set the connection queue. May not be absolutely necessary, but a good practice.

yanchengnv (approved):

nvflare/apis/utils/common_utils.py
```diff
@@ -0,0 +1,36 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import socket
+
+
+def get_open_ports(number):
```
IsaacYangSLA, line 18: Roughly when number == 100, there could be one duplicate port (>1024) in the returned list.  Other than that, this function seems safe.

IsaacYangSLA (approved):
Just one comment about the get_open_ports function.

PR #139, opened 2022-01-21
Refactor LocalLogger to remove the dependency on LogSender.

yhwen:
codes in apis folder should not depend on app_common folder.

That is why we need this change.

### DIFF
--- a/nvflare/apis/utils/local_logger.py
+++ b/nvflare/apis/utils/local_logger.py
@@ -15,25 +15,44 @@
 import logging
 import threading
 
-from nvflare.app_common.widgets.streaming import LogSender
-
 
 class LocalLogger:
+    handlers = None
     loggers = {}
     lock = threading.Lock()
 
     @staticmethod
+    def initialize():
+        """        Initialize the LocalLogger to keep all the handlers before the adding of LogSender handler.
+        Returns:
+
+        """
+        if not LocalLogger.handlers:
+            LocalLogger.handlers = []
+            for handler in logging.root.handlers:
+                LocalLogger.handlers.append(handler)
+
+    @staticmethod
     def get_logger(name=None) -> logging.Logger:
+        """        Get a logger only do the local logging.
+        Args:
+            name: logger name
+
+        Returns: local_logger
+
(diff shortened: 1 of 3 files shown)


nvflare/app_common/widgets/streaming.py
```diff
@@ -297,6 +298,8 @@ def __init__(self, log_level=""):
 
     def handle_event(self, event_type: str, fl_ctx: FLContext):
         if event_type == EventType.ABOUT_TO_START_RUN:
+            # initialize the LocalLogger to keep all the local handlers.
+            LocalLogger.initialize()
```
yanchengnv, line 302: LocalLogger.initialize() should be called from the framework before firing the ABOUT_TO_START_RUN event, to be safe. Otherwise someone could write another widget that adds a handler before this streaming widget's ABORT_TO_START_RUN is called.

yanchengnv (changes requested):

nvflare/apis/utils/local_logger.py
```diff
@@ -15,25 +15,44 @@
...
-
 
 class LocalLogger:
+    handlers = None
     loggers = {}
     lock = threading.Lock()
 
+    @staticmethod
+    def initialize():
+        """        Initialize the LocalLogger to keep all the handlers before the adding of LogSender handler.
```
YuanTingHsieh, line 26: I think it would be better to not include this "LogSender" in the docstring

nvflare/app_common/widgets/streaming.py
yhwen: Done

nvflare/apis/utils/local_logger.py
yhwen: I expect you to modify the docstring when you move this into the log_streaming.py.

YuanTingHsieh: no problem, let me do that in #135

YuanTingHsieh (approved):
Thanks, LGTM.

yanchengnv (approved):

PR #140, opened 2022-01-21
Make tenseal optional

### LINKED ISSUE #130: Tenseal dependency for HE is not available on ARM aarch64
The tenseal dependency is not available for the ARM aarch64 platform, causing installation to fail.  This has been reported for local development on Mac M1 and will affect other non-x86 architectures, Jetson, Clara AGX, IBM POWER, etc..

The tenseal dependency is only required when using the HEBuilder module, and it looks like all other functionality could be used without this dependency.  Can tenseal be made optional, with the caveat that HE is not available without tenseal?

One option would be providing an alternate install, a `requirements-no-tenseal.txt` that includes everything but tenseal.  For example, I generated this file in a clean venv on my linux machine using:
```
pip download nvflare -d /tmp -v \
    | grep Collecting \
    | awk '{print $2}' \
    | tr '[:upper:]' '[:lower:]' \
    | grep -v tenseal \
    | tee requirements-no-tenseal.txt
```
and verified that I can install nvflare and all deps except tenseal by copying to an aarch64 system (in this case a Jetson TX2) with:
```
python3 -m pip install --no-deps -r requirements-no-tenseal.txt
```

This is a pretty awkward solution.  It would be much cleaner to remove the tenseal dependency in the default packaging, since HE is optional, and note in the docs that tenseal must be installed when using HE.

IsaacYangSLA:
Fixes #130 

To pip install nvflare with tenseal: pip install nvflare[HE]

To install dependencies with tenseal: pip install -r requirements-extra.txt

### DIFF
--- a/nvflare/lighter/impl/he.py
+++ b/nvflare/lighter/impl/he.py
@@ -14,7 +14,10 @@
 
 import os
 
-import tenseal as ts
+try:
+    import tenseal as ts
+except ImportError:
+    ts = None
 
 from nvflare.lighter.spec import Builder
 
@@ -38,6 +41,11 @@ class HEBuilder(Builder):
             scale_bits: defaults to 40.
             scheme: defaults to "CKKS".
         """
+        if ts is None:
+            print(
+                "\n *** tenseal is not installed.  HEBuilder is ignored and no tenseal files will be generated. ***\n"
+            )
+            return
         self._context = None
         self.scheme_type_mapping = {
             "CKKS": ts.SCHEME_TYPE.CKKS,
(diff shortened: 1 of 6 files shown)


yhwen (approved):

holgerroth (dismissed):
LGTM

YuanTingHsieh (approved):
LGTM.

YuanTingHsieh:
@IsaacYangSLA @nvkevlu  we might need to update the README.md to include this [he] option in installation part.

IsaacYangSLA:
Will hold this until package name is finalized.

nvflare/lighter/impl/he.py
```diff
@@ -28,6 +31,11 @@ def __init__(
         scheme="CKKS",
     ):
 
+        if ts is None:
+            print(
```
chesterxgchen, line 35: should we using logging instead of print ? 

should the message indicate how to include teaseal? something like 
"To install dependencies with tenseal: pip install -r requirements-extra.txt"

chesterxgchen (approved):

holgerroth (approved):

PR #144, opened 2022-01-24
Add Ditto algorithm to prostate example

ZiyueXu77:
Initial implementation


ZiyueXu77:
Ditto mainly introduced: one additional model training to local learner, and proxloss for one of the two models

yanchengnv (dismissed):

ZiyueXu77:
@holgerroth @yanchengnv @Can-Zhao shall we move the "supervised trainer" to app_common/pt?

holgerroth:
> @holgerroth @yanchengnv @Can-Zhao shall we move the "supervised trainer" to app_common/pt?

It should be "SupervisedLearner", right? Let's move it under `app_common/pt/learners`?

ZiyueXu77:
ah yes right, supervised_learner, I derived two classes from it: prostate_learner with prostate specs, and ditto_learner with additional network training for ditto. We can incorporate the ditto part into base class (like fedprox), but my concern is that this base class will grow larger and larger when we include more algorithms.
prostate_learner is used for central, fedavg, and fed prox.
prostate_ditto_learner, inherited from prostate_learner and ditto_learner, is used for ditto

examples/prostate/configs/prostate_ditto/config/environment.json
```diff
@@ -0,0 +1,3 @@
+{
+    "APP_CKPT_DIR": "models"
+}
```
YuanTingHsieh, line 3: What's the difference between config_train.json and environment.json?

Can we merge this two files into one?

examples/prostate/pt/learners/supervised_prostate_learner.py
```diff
@@ -165,6 +166,12 @@ def _extra_train_config(self, fl_ctx: FLContext, config_info: str):
         self.transform_post = Compose([EnsureType(), Activations(sigmoid=True), AsDiscrete(threshold=0.5)])
 
         # Set dataset
+        # self.train_dataset = CacheDataset(
+        #     data=train_list,
+        #     transform=self.transform_train,
+        #     cache_rate=1.0,
+        #     num_workers=0,
+        # )
```
YuanTingHsieh, line 174: Suggest removing unused codes.

examples/prostate/pt/learners/supervised_learner.py
```diff
@@ -50,7 +52,7 @@ def __init__(
             a Shareable with the updated local model after running `execute()`
             or the best local model depending on the specified task.
         """
-        super().__init__()
+        super().__init__(*args, **kwargs)
```
YuanTingHsieh, line 55: Why do we need these *args, **kwargs?

examples/prostate/pt/learners/supervised_prostate_learner.py
```diff
@@ -15,7 +15,7 @@
 
 import torch
 import torch.optim as optim
-from monai.data import DataLoader, Dataset, load_decathlon_datalist
+from monai.data import DataLoader, CacheDataset, Dataset, load_decathlon_datalist
```
YuanTingHsieh, line 18: suggest removing unused import

examples/prostate/pt/learners/supervised_learner.py
```diff
@@ -23,20 +23,22 @@
...
 from nvflare.app_common.abstract.learner_spec import Learner
+from nvflare.app_common.abstract.model import ModelLearnableKey
 from nvflare.app_common.app_constant import AppConstants, ModelName, ValidateType
 from nvflare.app_common.pt.pt_fedproxloss import PTFedProxLoss
 
 
 class SupervisedLearner(Learner):
     def __init__(
-        self,
+        self, *args,
```
YuanTingHsieh, line 36: If this `SupervisedLearner` is general purpose we can move it into app_common/pt/learners as Holger suggested.

examples/prostate/configs/prostate_ditto/config/environment.json
ZiyueXu77: environment.json is needed by nvflare some versions ago, seems no longer the case, will remove it

examples/prostate/pt/learners/supervised_learner.py
ZiyueXu77: will remove

examples/prostate/pt/learners/supervised_learner.py
ZiyueXu77: Agreed, I will also move the ditto learner to it since it is task agnostic.

examples/prostate/pt/learners/supervised_prostate_learner.py
```diff
@@ -37,22 +37,24 @@
...
-class ProstateLearner(SupervisedLearner):
+class SupervisedProstateLearner(SupervisedLearner):
     def __init__(
         self,
+        *args,
         train_config_filename,
         aggregation_epochs: int = 1,
         train_task_name: str = AppConstants.TASK_TRAIN,
         submit_model_task_name: str = AppConstants.TASK_SUBMIT_MODEL,
+        **kwargs,
```
YuanTingHsieh, line 57: Do we need *args, **kwargs here?

nvflare/app_common/pt/learners/supervised_ditto_learner.py
```diff
@@ -0,0 +1,242 @@
...
+class SupervisedDittoLearner(SupervisedLearner):
+    def __init__(
+        self,
+        *args,
+        train_config_filename,
+        aggregation_epochs: int = 1,
+        local_model_epochs: int = 1,
+        train_task_name: str = AppConstants.TASK_TRAIN,
+        submit_model_task_name: str = AppConstants.TASK_SUBMIT_MODEL,
+        **kwargs,
```
YuanTingHsieh, line 39: Do we need *args, **kwargs here?

examples/prostate/pt/learners/supervised_prostate_learner.py
ZiyueXu77: yes, these are needed for multi-inheritance (same for the other class)

holgerroth (changes requested):

holgerroth (dismissed):
Nice job!

nvflare/app_common/pt/learners/supervised_learner.py
```diff
@@ -124,11 +123,6 @@ def train_config(self, fl_ctx: FLContext, train_config_file_path: str):
         with open(train_config_file_path) as file:
             config_info = json.load(file)
 
-        self.fedproxloss_mu = config_info.get("fedproxloss_mu", 0.0)
-        if self.fedproxloss_mu > 0:
-            self.log_info(fl_ctx, f"Using FedProx loss with mu {self.fedproxloss_mu}")
-            self.criterion_prox = PTFedProxLoss(mu=self.fedproxloss_mu)
-
         self._extra_train_config(fl_ctx, config_info)
```
YuanTingHsieh, line 126: The file name and the class name (supervised_learner and supervised_ditto_learner) suggests this should be general purpose.

However from the implementation, I see the users need to define the self.model (and some other stuff) somewhere for this to work. (like getting stuff from config), and the MONAI text in docstring suggest these are not general purpose.

We either need to fix the issue in this PR, or we don't move into app_common for now, we do a separate PR to move them.

YuanTingHsieh (changes requested):

nvflare/app_common/pt/learners/supervised_learner.py
ZiyueXu77: sure let's reverse back and have some further discussions for a separate PR later

YuanTingHsieh (approved):
Thanks, LGTM.

holgerroth (approved):

PR #146, opened 2022-01-25
Add protection to prevent loop stream logging.

yhwen:


nvflare/app_common/widgets/log_streaming.py
```diff
@@ -39,6 +39,11 @@ def __init__(self, log_level=""):
         Widget.__init__(self)
         logging.StreamHandler.__init__(self)
         self.log_level = getattr(logging, log_level, logging.INFO)
+        if self.log_level < logging.INFO:
+            raise ValueError(
+                f"LogAnalyticsSender log level must be higher than logging.INFO: {logging.INFO}. "
```
yanchengnv, line 44: The error msg should say that "... higher than DEBUG ...", or " ... higher than or equal to INFO".

Why do we need "{logging.INFO}" in the msg?

nvflare/private/event.py
```diff
@@ -58,7 +58,8 @@ def fire_event(event: str, handlers: list, ctx: FLContext):
                 ctx.set_prop(key=FLContextKey.EVENT_ORIGIN, value=event_origin, private=True, sticky=False)
                 ctx.set_prop(key=FLContextKey.EVENT_SCOPE, value=event_scope, private=True, sticky=False)
                 h.handle_event(event, ctx)
-            except:
-                h.log_exception(ctx, 'exception when handling event "{}"'.format(event), fire_event=False)
+            except Exception as e:
```
yanchengnv, line 61: The e is not used, so you don't need it here.

yanchengnv (changes requested):

IsaacYangSLA (dismissed):
Maybe in the future, we can check if filters are suitable for this logging routing.

nvflare/app_common/widgets/log_streaming.py
yhwen: Added self.log_level initialization enhancement.

nvflare/private/event.py
yhwen: done.

yanchengnv (changes requested):

yanchengnv (approved):
Let's improve the log_level config handling later

nvflare/app_common/widgets/log_streaming.py
```diff
@@ -30,15 +30,27 @@
...
         Args:
             log_level: log_level threshold
         """
         Widget.__init__(self)
         logging.StreamHandler.__init__(self)
-        self.log_level = getattr(logging, log_level, logging.INFO)
+        if log_level is None:
+            log_level = "INFO"
+
+        self.log_level = getattr(logging, log_level)
```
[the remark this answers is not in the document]
yhwen: I was trying to let the AttributeError throws out itself, with the proper error message already. But it's okay to catch the AttributeError and convert it to a ValueError.

IsaacYangSLA (approved):

PR #152, opened 2022-01-26
Changed the EventRecorder to only use local logger.

yhwen:


nvflare/app_common/handlers/event_recorder.py
```diff
@@ -206,6 +206,7 @@ def handle_event(self, event_type: str, fl_ctx: FLContext):
             formatter = logging.Formatter("%(asctime)s - %(name)s - %(levelname)s - %(message)s")
             output_file_handler.setFormatter(formatter)
             self.logger.addHandler(output_file_handler)
+            self.logger.propagate = False
```
IsaacYangSLA, line 209: Setting propagate to False means the log record will not be passed to its parent, which I believe is the ROOT logger.  Does ROOT logger cause any issue?

yhwen: yes

PR #159, opened 2022-01-27
Remove the log streaming codes.

yhwen:


yanchengnv (approved):

IsaacYangSLA (approved):
Not sure if we still need to do _fire_log_event.  The only destination is log_recorder, I think.

PR #169, opened 2022-02-02
Enhanced the DXO data and meta default to empty dict.

yhwen:


nvflare/apis/dxo.py
```diff
@@ -47,10 +47,15 @@ def __init__(self, data_kind: str, data: dict, meta: dict = None):
...
-            data_kind:
+            data_kind: data_kind name
             data: clear-text data
             meta: None or dict
         """
+        if data is None:
+            data = {}
+        if meta is None:
+            meta = {}
+
```
yanchengnv, line 58: Since we are doing this, we should make sure data and meta are of dict type.

yanchengnv (changes requested):

nvflare/apis/dxo.py
yhwen: We have this:

def __init__(self, data_kind: str, data: dict, meta: dict = None):

yanchengnv (changes requested):

nvflare/apis/dxo.py
yhwen: Added data and meta dict type check.

yhwen: Actually we already have the dict type check in:

err = self.validate()

yanchengnv: You are right!

yanchengnv (approved):

YuanTingHsieh (approved):

PR #178, opened 2022-02-04
Standardize error messages

SYangster:
- improving error messages to use the form "_ must be x but got y"
- adjusted some unit tests to match new error messages
- standardizes errors in all files except in fuel/ (covered by Kevin)

### DIFF
--- a/nvflare/apis/impl/controller.py
+++ b/nvflare/apis/impl/controller.py
@@ -106,9 +106,10 @@ class Controller(Responder, ControllerSpec, ABC):
         """
         collector = fl_ctx.get_prop(InfoCollector.CTX_KEY_STATS_COLLECTOR, None)
         if collector:
-            assert isinstance(
-                collector, GroupInfoCollector
-            ), "collector must be an instance of GroupInfoCollector, but got {}".format(type(collector))
+            if not isinstance(collector, GroupInfoCollector):
+                raise TypeError(
+                    "collector must be an instance of GroupInfoCollector, but got {}".format(type(collector))
+                )
             collector.set_info(
                 group_name=self._name,
                 info={
--- a/nvflare/apis/impl/controller.py
+++ b/nvflare/apis/impl/controller.py
@@ -42,24 +42,27 @@ _TASK_KEY_DONE = "___done"
 
 def _check_positive_int(name, value):
     if not isinstance(value, int):
-        raise TypeError("{} must be an instance of int.".format(name))
+        raise TypeError("{} must be an instance of int, but got {}.".format(name, type(name)))
     if value < 0:
         raise ValueError("{} must >= 0.".format(name))
 
 
 def _check_inputs(task: Task, fl_ctx: FLContext, targets: Union[List[Client], List[str], None]):
     if not isinstance(task, Task):
-        raise TypeError("task must be an instance of Task.")
+        raise TypeError("task must be an instance of Task, but got {}".format(type(task)))
 
     if not isinstance(fl_ctx, FLContext):
-        raise TypeError("fl_ctx must be an instance of FLContext.")
+        raise TypeError("fl_ctx must be an instance of FLContext, but got {}".format(type(fl_ctx)))
 
     if targets is not None:
         if not isinstance(targets, list):
-            raise TypeError("targets must be a list of Client or string.")
+            raise TypeError("targets must be a list of Client or string, but got {}".format(type(targets)))
 
-        if not all([isinstance(t, (Client, str)) for t in targets]):
-            raise TypeError("targets must be a list of Client or string.")
+        for t in targets:
+            if not isinstance(t, (Client, str)):
+                raise TypeError(
+                    "targets must be a list of Client or string, but got element of type {}".format(type(t))
+                )
 
 
 class Controller(Responder, ControllerSpec, ABC):
@@ -103,7 +106,9 @@ class Controller(Responder, ControllerSpec, ABC):
         """
         collector = fl_ctx.get_prop(InfoCollector.CTX_KEY_STATS_COLLECTOR, None)
         if collector:
-            assert isinstance(collector, GroupInfoCollector)
+            assert isinstance(
+                collector, GroupInfoCollector
+            ), "collector must be an instance of GroupInfoCollector, but got {}".format(type(collector))
             collector.set_info(
                 group_name=self._name,
                 info={
@@ -139,9 +144,9 @@ class Controller(Responder, ControllerSpec, ABC):
             Tuple[str, str, Shareable]: task_name, an id for the client_task, and the data for this requst
         """
         if not isinstance(client, Client):
-            raise TypeError("client must be an instance of Client.")
+            raise TypeError("client must be an instance of Client, but got {}".format(type(client)))
         if not isinstance(fl_ctx, FLContext):
-            raise TypeError("fl_ctx must be an instance of FLContext.")
+            raise TypeError("fl_ctx must be an instance of FLContext, but got {}".format(type(fl_ctx)))
 
         client_task_to_send = None
         with self._task_lock:
@@ -312,11 +317,11 @@ class Controller(Responder, ControllerSpec, ABC):
             ValueError: task_name is not found in the client_task
         """
         if not isinstance(client, Client):
-            raise TypeError("client must be an instance of Client.")
+            raise TypeError("client must be an instance of Client, but got {}".format(type(client)))
         if not isinstance(fl_ctx, FLContext):
-            raise TypeError("fl_ctx must be an instance of FLContext.")
+            raise TypeError("fl_ctx must be an instance of FLContext, but got {}".format(type(fl_ctx)))
         if not isinstance(result, Shareable):
-            raise TypeError("result must be an instance of Shareable.")
+            raise TypeError("result must be an instance of Shareable, but got {}".format(type(result)))
 
         with self._task_lock:
             # task_id is the uuid associated with the client_task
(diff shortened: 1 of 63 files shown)


nvflare/apis/impl/controller.py
```diff
@@ -103,7 +106,9 @@ def _set_stats(self, fl_ctx: FLContext):
         """
         collector = fl_ctx.get_prop(InfoCollector.CTX_KEY_STATS_COLLECTOR, None)
         if collector:
-            assert isinstance(collector, GroupInfoCollector)
+            assert isinstance(
+                collector, GroupInfoCollector
+            ), "collector must be an instance of GroupInfoCollector, but got {}".format(type(collector))
```
YuanTingHsieh, line 111: if we want to ensure this check,
we can use if isinstance: ... raise TypeError

nvflare/apis/utils/fl_context_utils.py
```diff
@@ -63,7 +63,9 @@ def generate_log_message(fl_ctx: FLContext, msg: str):
 
     peer_ctx = fl_ctx.get_peer_context()
     if peer_ctx:
-        assert isinstance(peer_ctx, FLContext)
+        assert isinstance(peer_ctx, FLContext), "peer_ctx must be an instance of FLContext, but got {}".format(
+            type(peer_ctx)
+        )
```
YuanTingHsieh, line 68: Same here, we need to think if we need this check or not.

If we need: suggest change to if instance raise
If we don't: remove this whole assert statement

nvflare/app_common/aggregators/accumulate_model_aggregator.py
```diff
@@ -67,7 +67,7 @@ def accept(self, shareable: Shareable, fl_ctx: FLContext) -> bool:
         except:
             self.log_exception(fl_ctx, "shareable data is not a valid DXO")
             return False
-        assert isinstance(dxo, DXO)
+        assert isinstance(dxo, DXO), "dxo must be DXO but got {}".format(type(dxo))
```
YuanTingHsieh, line 70: Same here, we need to think if we need this check or not.

If we need: suggest change to if instance raise
If we don't: remove this whole assert statement

nvflare/app_common/aggregators/accumulate_model_aggregator.py
```diff
@@ -150,7 +150,7 @@ def aggregate(self, fl_ctx: FLContext) -> Shareable:
         for v_name in vars_to_aggregate:
             n_local_iters, np_vars = [], []
             for item in self.accumulator:
-                assert isinstance(item, _AccuItem)
+                assert isinstance(item, _AccuItem), "item must be _AccuItem but got {}".format(type(item))
```
YuanTingHsieh, line 153: and here.

nvflare/app_common/aggregators/intime_accumulate_model_aggregator.py
```diff
@@ -117,7 +117,7 @@ def accept(self, shareable: Shareable, fl_ctx: FLContext) -> bool:
         except BaseException:
             self.log_exception(fl_ctx, "shareable data is not a valid DXO")
             return False
-        assert isinstance(dxo, DXO)
+        assert isinstance(dxo, DXO), "dxo must be DXO but got {}".format(type(dxo))
```
YuanTingHsieh, line 120: and here.

nvflare/app_common/executors/multi_process_executor.py
```diff
@@ -320,7 +320,7 @@ def finalize(self, fl_ctx: FLContext):
 
         # wait for all relay threads to join!
         for t in self.relay_threads:
-            assert isinstance(t, threading.Thread)
+            assert isinstance(t, threading.Thread), "t must be threading.Thread but got {}".format(type(t))
```
YuanTingHsieh, line 323: and here.

nvflare/app_common/filters/exclude_vars.py
```diff
@@ -87,7 +95,7 @@ def process(self, shareable: Shareable, fl_ctx: FLContext) -> Shareable:
             self.log_exception(fl_ctx, "shareable data is not a valid DXO")
             return shareable
 
-        assert isinstance(dxo, DXO)
+        assert isinstance(dxo, DXO), "dxo must be DXO but got {}".format(type(dxo))
```
YuanTingHsieh, line 98: and here.

nvflare/app_common/handlers/event_recorder.py
```diff
@@ -123,9 +123,9 @@ def event_error_tag(self, fl_ctx: FLContext):
 
     def validate_prop(self, prop_name: str, req: _CtxPropReq, fl_ctx: FLContext):
         stats = fl_ctx.get_prop(self._KEY_EVENT_STATS, None)
-        assert isinstance(stats, _EventStats)
+        assert isinstance(stats, _EventStats), "stats must be _EventStats but got {}".format(type(stats))
 
-        assert isinstance(req, _CtxPropReq)
+        assert isinstance(req, _CtxPropReq), "req must be _CtxPropReq but got {}".format(type(req))
```
YuanTingHsieh, line 128: these are private classes, who set the stats? req?
do we really need to check the class here?

YuanTingHsieh (changes requested):
Good and detailed changes.

If the original statement is `assert(data, class)`, we need to determine if this check is necessary:

- If it is necessary, suggest changing to `if not isinstance, raise TypeError`
- If it is not, suggest removing them

nvkevlu (approved):
All looks good to me, thanks @SYangster. @YuanTingHsieh makes a good point about the asserts. I don't know if we need to do anything in this commit as it does not look like it will harm anything, but many assert statements were only used for coding help for auto-complete in the IDE so maybe they can be removed.

nvkevlu:
Thanks for the update. What was the decision making process behind which asserts should be removed vs the ones changed to if and raise error? Some of these are different from what I was expecting, like I imagined checking for DXO is something we do want to do, and ClientEngineInternalSpec, ServerEngineSpec, and ServerEngineInternalSpec are more internal and ones we could remove.

Otherwise, I think the rest look good to me.

SYangster:
For the DXOs, I saw that from_shareable() in the try-except already does some error checking and will return a DXO. I also removed some where the variables were defined in the same file rather than passed in. For a lot of the other ones I decided to err on the side of caution/ leave them to keep the code more readable with the type checking. I can remove ClientEngineInternalSpec, ServerEngineSpec, and ServerEngineInternalSpec since those are more internal though.

yanchengnv:
Good question, Kevin.

In general, if you know that an assertion can never be wrong (e.g. the object was already created to be type X), then the assert's only use was for code autocompletion. I'm not sure we have to remove it. 

Otherwise, if the object is passed to you through a method arg, then the assert could go wrong and should checked.

nvkevlu (approved):
Thank you, this looks good and your comment helps explain and it makes sense. If you are removing some asserts for the specs already, we could, or we could leave them.

SYangster:
Ok great, I think I will leave them in for now

nvflare/app_common/workflows/cross_site_model_eval.py
```diff
@@ -125,8 +125,12 @@ def start_controller(self, fl_ctx: FLContext):
...
             self._participating_clients = [c.name for c in clients]
         else:
-            assert isinstance(self._participating_clients, list), "participating_clients must be a list"
-            assert all(isinstance(x, str) for x in self._participating_clients), "participating_clients must be strings"
+            if not isinstance(self._participating_clients, list):
+                raise TypeError(
+                    "participating_clients must be a list but got {}".format(type(self._participating_clients))
+                )
+            if not all(isinstance(x, str) for x in self._participating_clients):
+                raise TypeError("participating_clients must be strings")
```
YuanTingHsieh, line 133: Is it more proper to move this part in __init__?

nvflare/private/event.py
```diff
@@ -48,7 +48,8 @@ def fire_event(event: str, handlers: list, ctx: FLContext):
 
     if handlers:
         for h in handlers:
-            assert isinstance(h, FLComponent)
+            if not isinstance(h, FLComponent):
+                raise TypeError("h must be FLComponent but got {}".format(type(h)))
```
YuanTingHsieh, line 52: It might be hard for users to understand what is "h"

nvflare/private/fed/server/shell_cmd.py
```diff
@@ -152,7 +155,8 @@ def validate_shell_command(self, args: List[str], parse_result):
                 return "only one file is allowed"
 
             for f in parse_result.files:
-                assert isinstance(f, str)
+                if not isinstance(f, str):
+                    raise TypeError("f must be str but got {}".format(type(f)))
```
YuanTingHsieh, line 159: also this one, but we can improve these later.

nvflare/private/fed_json_config.py
```diff
@@ -163,7 +163,8 @@ def finalize_config(self, config_ctx: ConfigContext):
 
         data_filter_table = {}
         for c in self.task_data_filter_chains:
-            assert isinstance(c, FilterChain)
+            if not isinstance(c, FilterChain):
+                raise TypeError("c must be FilterChain but got {}".format(type(c)))
```
YuanTingHsieh, line 167: this one as well.

YuanTingHsieh (approved):

nvflare/private/event.py
SYangster: I agree, fixed

nvflare/app_common/workflows/cross_site_model_eval.py
SYangster: done

YuanTingHsieh (approved):

nvkevlu (approved):

PR #179, opened 2022-02-04
Add default FL app validator

YuanTingHsieh:
Add a default FL AppValidator to ensure the application's structure is valid.

- DefaultAppValidator: Validates an application has config folder and config_fed_server.json and config_fed_client.json inside it.
- FLAppValidator: The built-in FL app validator, by default it contains the DefaultAppValidator

### DIFF
--- a/nvflare/private/fed/app/fl_app_validator.py
+++ b/nvflare/private/fed/app/fl_app_validator.py
@@ -1,0 +1,42 @@
+# Copyright (c) 2021-2022, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Dict, List, Optional, Tuple
+
+from nvflare.apis.app_validation import AppValidator
+
+from .default_app_validator import DefaultAppValidator
+
+
+class FLAppValidator(AppValidator):
+    def __init__(self, custom_validators: Optional[List[AppValidator]] = None):
+        super().__init__()
+        self.validators = [DefaultAppValidator()]
+        if custom_validators:
+            if not isinstance(custom_validators, list):
+                raise TypeError("custom_validators must be list, but got {}".format(type(custom_validators)))
+            for validator in custom_validators:
+                if not isinstance(validator, AppValidator):
+                    raise TypeError("validator must be AppValidator, but got {}".format(type(validator)))
+                self.validators.append(validator)
+
+    def validate(self, app_folder: str) -> Tuple[str, Dict]:
+        final_result = {}
+        for v in self.validators:
+            err, result = v.validate(app_folder)
+            if err:
+                return err, result
+            if result:
+                final_result.update(result)
+        return "", final_result
--- a/nvflare/private/fed/app/fl_conf.py
+++ b/nvflare/private/fed/app/fl_conf.py
@@ -24,6 +24,7 @@ from nvflare.fuel.utils.wfconf import ConfigContext
 from nvflare.private.fed.client.base_client_deployer import BaseClientDeployer
 from nvflare.private.json_configer import JsonConfigurator
 
+from .fl_app_validator import FLAppValidator
 from .trainers.server_deployer import ServerDeployer
 
 FL_PACKAGES = ["nvflare"]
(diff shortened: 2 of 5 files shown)


nvflare/private/fed/app/fl_conf.py
```diff
@@ -103,6 +105,9 @@ def finalize_config(self, config_ctx: ConfigContext):
         if not secure_train:
             self.enable_byoc = True
 
+        if self.app_validator is None:
+            self.app_validator = FLAppValidator()
```
yanchengnv, line 109: Even if app_validator is not None, we still want to apply FLAppValidator to ensure the basic structure of the app.

I suggest:

1. Change the name of FLAppValidator to AppStructureValidator.
2. Write FLAppValidator that can take additional custom validators:

`
class FLAppValidator(AppValidator):

    def __init__(self, custom_validators: list):
         self.validators = [AppStructureValidator()]
         if custom_validators:
              self.validators.extend(custom_validators)

    def validate(self, app_folder: str) -> Tuple[str, Dict]:
          final_result = {}
          for v in self.validators:
              err, result = v.validate(app_folder)
              if err:
                  return err, result
              if result:
                  final_result.update(result)
          return "", final_result
`
3. custom_vals = [cust_val] if cust_val else []
4. self.validator = FLAppValidator(custom_vals)

nvflare/private/fed/app/fl_app_validator.py
```diff
@@ -0,0 +1,33 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+from typing import Dict, Tuple
+
+from nvflare.apis.app_validation import AppValidator, AppValidationKey
+
+
+class FLAppValidator(AppValidator):
```
yanchengnv, line 21: Change the class name to AppStructureValidator. See my comment on fl_conf.py

yanchengnv (changes requested):

nvflare/private/fed/app/fl_app_validator.py
YuanTingHsieh: Changed the class name to DefaultAppValidator.

Because I think the AppValidator has the same meaning as AppStructureValidator.

nvflare/private/fed/app/fl_app_validator.py
```diff
@@ -0,0 +1,53 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+from typing import Dict, Tuple, Optional, List
+
+from nvflare.apis.app_validation import AppValidator, AppValidationKey
+
+
+class DefaultAppValidator(AppValidator):
```
yanchengnv, line 21: Maybe should change the file name to default_app_validator.py?

nvflare/private/fed/app/fl_app_validator.py
```diff
@@ -0,0 +1,53 @@
...
+    def validate(self, app_folder: str) -> Tuple[str, Dict]:
+        result = dict()
+        app_root = os.path.abspath(app_folder)
+        if not os.path.exists(os.path.join(app_root, "config")):
+            return "Missing config folder inside app folder.", {}
+        if not os.path.exists(os.path.join(app_root, "config", "config_fed_server.json")):
```
yanchengnv, line 27: In theory, if the app is deployed to Server, then only config_fed_server.json is needed; if the app is for Client, then only config_fed_client.json is needed. However this is hard to do at this time since we don't know what the app is to be used for. In NVF2.1, we'll know this and the validation will be done for the whole "experiment" that contains all apps and sufficient info.

So let's just keep this in mind.

yanchengnv (changes requested):
Please just change the file name of fl_app_validator.py as commented.

nvflare/private/fed/app/fl_app_validator.py
YuanTingHsieh: add a new file called "default_app_validator.py" to put in this class.

yanchengnv (changes requested):

nvflare/private/fed/app/fl_app_validator.py
```diff
@@ -0,0 +1,39 @@
...
+from nvflare.apis.app_validation import AppValidator
+
+from .default_app_validator import DefaultAppValidator
+
+
+class FLAppValidator(AppValidator):
+    def __init__(self, custom_validators: Optional[List[AppValidator]] = None):
+        super().__init__()
+        self.validators = [DefaultAppValidator()]
+        for validator in custom_validators:
```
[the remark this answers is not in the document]
YuanTingHsieh: thanks, fixed.

yanchengnv (approved):

IsaacYangSLA (approved):
LGTM.  One question, but not critical, the FLAppValidator allows a list of AppValidator and the users actually configure either None or one custom_validator?  If None, the default one is still being used?

YuanTingHsieh:
> LGTM. One question, but not critical, the FLAppValidator allows a list of AppValidator and the users actually configure either None or one custom_validator? If None, the default one is still being used?

Yes, this argument is correct.

yhwen (approved):

PR #186, opened 2022-02-05
Fixe cross_validation locate_model should ony return the model weight…

yhwen:
…s DXO.

### DIFF
--- a/nvflare/app_common/pt/pt_file_model_locator.py
+++ b/nvflare/app_common/pt/pt_file_model_locator.py
@@ -17,6 +17,7 @@ from typing import List
 from nvflare.apis.dxo import DXO, DataKind
 from nvflare.apis.event_type import EventType
 from nvflare.apis.fl_context import FLContext
+from nvflare.app_common.abstract.model import ModelLearnableKey
 from nvflare.app_common.abstract.model_locator import ModelLocator
 from nvflare.app_common.pt.pt_file_model_persistor import PTFileModelPersistor
 
@@ -67,7 +68,8 @@ class PTFileModelLocator(ModelLocator):
         if model_name not in list(self.model_inventory.keys()):
             raise ValueError(f"model inventory does not contain: {model_name}")
 
-        weights = self.model_persistor.get_model(model_name, fl_ctx)
+        model_learnable = self.model_persistor.get_model(model_name, fl_ctx)
+        weights = model_learnable.get(ModelLearnableKey.WEIGHTS)
         dxo = DXO(data_kind=DataKind.WEIGHTS, data=weights, meta={})
 
         return dxo


holgerroth (approved):
LGTM

yanchengnv (approved):

nvflare/app_common/pt/pt_file_model_locator.py
```diff
@@ -67,7 +68,10 @@ def locate_model(self, model_name, fl_ctx: FLContext) -> DXO:
         if model_name not in list(self.model_inventory.keys()):
             raise ValueError(f"model inventory does not contain: {model_name}")
 
-        weights = self.model_persistor.get_model(model_name, fl_ctx)
+        model_learnable = self.model_persistor.get_model(model_name, fl_ctx)
+        weights = model_learnable.get(ModelLearnableKey.WEIGHTS)
+        if weights is None:
+            raise ValueError(f"model {model_name} does not have required model weights.")
         dxo = DXO(data_kind=DataKind.WEIGHTS, data=weights, meta={})
```
yanchengnv, line 75: Should change to use model_learnable_to_dxo(model_learnable) here. It also includes all meta, which will keep the DXO self-sufficient.

yhwen: Updaed to use model_learnable_to_dxo().

holgerroth (approved):

nvflare/app_common/pt/pt_file_model_locator.py
```diff
@@ -17,6 +17,7 @@
 from nvflare.apis.dxo import DXO, DataKind
 from nvflare.apis.event_type import EventType
 from nvflare.apis.fl_context import FLContext
+from nvflare.app_common.abstract.model import ModelLearnableKey, model_learnable_to_dxo
```
yanchengnv, line 20: Looks like the ModelLearnableKey is no longer needed. Please remove from the import.

yanchengnv (changes requested):
I looked at this file end-to-end. It is fragile. In the initialize() it tries to get the persistor component from the ID, but does not check the returned component. It could be None. We need to raise exception when this happens; otherwise the problem will only be detected when locate_model() is called.

nvflare/app_common/pt/pt_file_model_locator.py
yhwen: cleaned up.

holgerroth (approved):

yanchengnv (approved):

IsaacYangSLA (approved):
Most of our codes have initialize while some have _initialize.  Not a critical issue but people may wonder why.
Other than that, it looks good to me.

PR #187, opened 2022-02-07
improve docstrings for apis and widgets

SYangster:
- fix pydocstyle issues in apis and widgets

### DIFF
--- a/nvflare/apis/aux_spec.py
+++ b/nvflare/apis/aux_spec.py
@@ -27,15 +27,14 @@ class MessageSendStatus(enum.Enum):
 
 
 def aux_request_handle_func_signature(topic: str, request: Shareable, fl_ctx: FLContext) -> Shareable:
-    """
-    This is the signature of the message_handle_func.
+    """This is the signature of the message_handle_func.
 
     Args:
         topic: topic of the message to be handled
         request: the message data to be handled
         fl_ctx: FL context
 
-    Returns: a response to the requester
+    Returns: a Shareable response to the requester
 
     """
     pass
--- a/nvflare/apis/aux_spec.py
+++ b/nvflare/apis/aux_spec.py
@@ -29,6 +29,12 @@ class MessageSendStatus(enum.Enum):
 def aux_request_handle_func_signature(topic: str, request: Shareable, fl_ctx: FLContext) -> Shareable:
     """This is the signature of the message_handle_func.
 
+    The message_handle_func is a callback function that is registered to handle an aux request of a specific topic.
+    Any implementation of a message_handle_func must follow this signature.
+
+    Example from the client runner:
+        engine.register_aux_message_handler(topic=ReservedTopic.END_RUN, message_handle_func=self._handle_end_run)
+
     Args:
         topic: topic of the message to be handled
         request: the message data to be handled
--- a/nvflare/apis/client.py
+++ b/nvflare/apis/client.py
@@ -19,11 +19,13 @@ class Client:
     def __init__(self, name, token) -> None:
         """Init Client.
 
+        Represents a client, and is managed by the client manager.
+        The token is a uuid used for authorization.
+
         Args:
             name: client name
             token: client token
         """
-        # super().__init__()
         self.name = name
         self.token = token
         self.last_connect_time = time.time()
--- a/nvflare/apis/client.py
+++ b/nvflare/apis/client.py
@@ -17,6 +17,12 @@ import time
 
 class Client:
     def __init__(self, name, token) -> None:
+        """Init Client.
+
+        Args:
+            name: client name
+            token: client token
+        """
         # super().__init__()
         self.name = name
         self.token = token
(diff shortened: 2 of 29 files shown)


nvflare/apis/aux_spec.py
```diff
@@ -27,15 +27,14 @@ class MessageSendStatus(enum.Enum):
 
 
 def aux_request_handle_func_signature(topic: str, request: Shareable, fl_ctx: FLContext) -> Shareable:
-    """
-    This is the signature of the message_handle_func.
+    """This is the signature of the message_handle_func.
```
YuanTingHsieh, line 30: We will need to improves this.
What is a message_handle_func?

nvflare/apis/client.py
```diff
@@ -17,6 +17,12 @@
 
 class Client:
     def __init__(self, name, token) -> None:
+        """Init Client.
```
YuanTingHsieh, line 20: We might want to add some explanation to this class.

We can remove # super().__init__() since it is commented out.

nvflare/apis/client_engine_spec.py
```diff
@@ -23,6 +23,13 @@
 
 class TaskAssignment(object):
     def __init__(self, name: str, task_id: str, data: Shareable):
+        """Init TaskAssignment.
```
YuanTingHsieh, line 26: What does ``TaskAssignment`` mean?

Maybe @yhwen can help on some of these classes.

nvflare/apis/dxo.py
```diff
@@ -143,7 +158,7 @@ def from_shareable(s: Shareable) -> DXO:
 
 
 def from_bytes(data: bytes) -> DXO:
-    """method to convert the object bytes into Model object.
+    """Method to convert the object bytes into Model object.
```
YuanTingHsieh, line 161: Suggest -> `Convert the object bytes into a DXO object.`

nvflare/apis/dxo.py
```diff
@@ -122,6 +128,15 @@ def validate(self) -> str:
 
 
 def from_shareable(s: Shareable) -> DXO:
+    """Method to convert DXO object into Shareable.
```
YuanTingHsieh, line 131: Suggest -> `Convert Shareable into a DXO object.`

nvflare/apis/dxo.py
```diff
@@ -96,11 +96,17 @@ def update_shareable(self, s: Shareable) -> Shareable:
         return s
 
     def to_shareable(self) -> Shareable:
+        """Method to convert DXO object into Shareable.
```
YuanTingHsieh, line 99: Suggest -> `Convert DXO object into Shareable.`

nvflare/apis/shareable.py
```diff
@@ -108,7 +107,7 @@ def get_peer_prop(self, key: str, default):
         return props.get(key, default)
 
     def to_bytes(self) -> bytes:
-        """method to serialize the Model object into bytes.
+        """Method to serialize the Model object into bytes.
```
YuanTingHsieh, line 110: Suggest -> `Serialize the Shareable object into bytes.`

nvflare/apis/shareable.py
```diff
@@ -118,7 +117,7 @@ def to_bytes(self) -> bytes:
 
     @classmethod
     def from_bytes(cls, data: bytes):
-        """method to convert the object bytes into Model object.
+        """Method to convert the object bytes into Model object.
```
YuanTingHsieh, line 120: Suggest -> `Covert the data bytes into a Shareable object.`

YuanTingHsieh (changes requested):
- For method / function docstring, suggest starts with a verb. We don't need to say "method to XXX", can directly say "XXX"

- For class `__init__` function docstring, 
  - If we have provided the class docstring that explains what this class does / is, then we can just say `Init XXX`
   - If we don't have the class level docstring that explains what this class does / is, I suggest add it in `__init__` function docstring.

nvflare/apis/aux_spec.py
SYangster: I believe that is a function that gets registered for the aux message handler. I agree this isn't very clear, I can add more detail.

IsaacYangSLA: @SYangster , one way to add more details is to say "any implementation of message_handle_func has to follow this signature.  message_handle_func is a callback function to handle aux request."  You can also use this as an example so users can know more about this signature (engine.register_aux_message_handler(topic=ReservedTopic.END_RUN, message_handle_func=self._handle_end_run))

IsaacYangSLA (approved):
This docstrings improvement is OK for current stage as we mainly to ensure no pydocstyle failure.  Some of them need more clarification if time allows because, for beginners, better docstrings really help.

IsaacYangSLA (approved):
LGTM.

nvkevlu (approved):
I'd say it looks okay for this task though docstrings can continue to be improved. One thing to note is that even though it is convention that we put the docstrings in `__init__`, we probably should not put "Init classname" as the first line of the docstring if we can come up with something more descriptive to put.

YuanTingHsieh (approved):
LGTM. Agree with Kevin's suggestion.

PR #188, opened 2022-02-07
Fix CredentialType

YuanTingHsieh:
- CredentialType class needs to inherit from str, so that string comparison ``"password" == CredentialType.PASSWORD`` will return True


nvkevlu (approved):
LGTM

PR #194, opened 2022-02-09
Fix admin shutdown client issue

### LINKED ISSUE #193: Clients don't shut down
Using admin API with

```
    runner.run(run_number, args.app, restart_all_first=False, shutdown_on_error=True, shutdown_at_end=True,
               timeout=7200, min_clients=args.min_clients)
```
clients don't shut down after end of run.

Similarly when using the admin client console.

Issue was probably introduced with this change https://github.com/NVIDIA/NVFlare/pull/174/files#diff-08f576a2965a40b523edfef473ea838f2a28c0bbce28a1c0bf7015e42dade1c8L345

YuanTingHsieh:
Fix #193

### DIFF
--- a/nvflare/private/fed/server/server_engine.py
+++ b/nvflare/private/fed/server/server_engine.py
@@ -350,6 +350,8 @@ class ServerEngine(ServerEngineInternalSpec):
     def _remove_dead_client(self, token):
         client = self.server.client_manager.remove_client(token)
         self.server.remove_client_data(token)
+        if self.server.admin_server:
+            self.server.admin_server.client_dead(token)
 
     def register_aux_message_handler(self, topic: str, message_handle_func):
         self.run_manager.aux_runner.register_aux_message_handler(topic, message_handle_func)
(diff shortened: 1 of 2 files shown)


nvkevlu (approved):

nvflare/private/fed/server/server_engine.py
```diff
@@ -350,8 +350,6 @@ def remove_clients(self, clients: List[str]) -> str:
     def _remove_dead_client(self, token):
         client = self.server.client_manager.remove_client(token)
         self.server.remove_client_data(token)
-        if self.server.admin_server:
```
yhwen, line 353: admin_server need to remove the dead client as well. We need this step.

yhwen (changes requested):
We need to remove the dead client from admin_server.

nvflare/private/fed/server/server_engine.py
yhwen: The root cause of this shutdown issue is the wrong order in the _shutdown_app_on_clients() in the server/training_cmds.py.  The "shutdown" message should be sent to the clients first, then call the engine.remove_clients(). Like below:

        replies = self.send_request_to_clients(conn, message)
        self.process_replies_to_table(conn, replies)

        err = engine.remove_clients(clients)
        if err:
            conn.append_error(err)
            return False

yhwen (approved):

nvkevlu (approved):

PR #195, opened 2022-02-10
Add information in documentation to help explain fed events

nvkevlu:
Add information in documentation to help explain fed events and link to tb streaming example.

### DIFF
--- a/docs/programming_guide/event_system.rst
+++ b/docs/programming_guide/event_system.rst
@@ -148,3 +148,21 @@ command from the researcher. If needed, you can use this event to gracefully fin
 If your component provides some services that other components can use, you can put your component as a private and
 sticky prop into the fl_ctx at the time of START_RUN under a uniquely defined prop name. Other components can later get
 your component with this name and invoke the services of your component.
+
+Local Events and Fed Events
+---------------------------
+Local events are local to the client, and federated events or fed events are broadcast to other sites as well.
+
+The :class:`ConvertToFedEvent<nvflare.app_common.widgets.convert_to_fed_event.ConvertToFedEvent>` widget will convert
+local events to federated events.
+
+One example of how this is applied in use is in log streaming as seen in the `hello-pt-tb example <https://github.com/NVIDIA/NVFlare/tree/main/examples/hello-pt-tb>`_.
+
+The :class:`AnalyticsSender<nvflare.app_common.widgets.streaming.AnalyticsSender>` triggers an event called "analytix_log_stats",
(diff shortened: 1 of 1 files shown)


docs/programming_guide/event_system.rst
```diff
@@ -148,3 +148,19 @@ command from the researcher. If needed, you can use this event to gracefully fin
 If your component provides some services that other components can use, you can put your component as a private and
 sticky prop into the fl_ctx at the time of START_RUN under a uniquely defined prop name. Other components can later get
 your component with this name and invoke the services of your component.
+
+Local Events and Fed Events
+---------------------------
+The :class:`ConvertToFedEvent<nvflare.app_common.widgets.convert_to_fed_event.ConvertToFedEvent>` widget will convert
+local events to federated events (meaning the event goes across to server).
```
YuanTingHsieh, line 155: Does Fed event only go from client to server?
Can it be another direction (server to client)?

YuanTingHsieh: I suggest we explain what is Fed events, (for example they are events that go across site).

Then we can mention to convert local events into fed events people can use the `ConvertToFedEvent` widget.

In the end, we say the example on how to configure this widget is in `hello-pt-tb example <https://github.com/NVIDIA/NVFlare/tree/main/examples/hello-pt-tb>`_

The following section if it is already included in the tutorial text. maybe we don't need to duplicate those.

yhwen: This maybe a slightly different issue. But maybe we should add an event_type parameter in AnalyticsSender. Right now it's sending default event type "analytix_log_stats".  However, in the ConvertToFedEvent and TBAnalyticsReceiver JSON configuration, the user needs to config the "analytix_log_stats" in the events to work. It's not natural for user to know where this event is coming from. If we add an "event_type" in the AnalyticsSender, we can make the event sending, converting and receiving in pairs. That would be easier for people to connect the dots and understand.

yanchengnv: Even though fed event could go from server to client, we don't have a use case for that currently, and it is more complicated (e.g. which client(s) to send). We probably should only focus on fed events from clients to server, for now.

nvkevlu: Modified the wording based on these comments. I kept the section even if it adds a bit of duplicated information because it addresses what was raised in the QA issue and this is a location where it can be useful to have this.

YuanTingHsieh (approved):

PR #199, opened 2022-02-10
admin client automatically logouts after timeout

SYangster:
- admin client quits if session timed out
- updated error message to be more clear to users

### DIFF
--- a/nvflare/fuel/hci/client/cli.py
+++ b/nvflare/fuel/hci/client/cli.py
@@ -266,6 +266,11 @@ class AdminClient(cmd.Cmd):
             ent = entries[0]
             resp = ent.handler(args, self.api)
             self.print_resp(resp)
+            if "data" in resp:
+                for item in resp["data"]:
+                    item_type = item["type"]
+                    if item_type == "error" and "session timed out" in item["data"]:
+                        return True
             return
 
         entries = self.api.server_cmd_reg.get_command_entries(cmd_name)
--- a/nvflare/fuel/hci/client/fl_admin_api.py
+++ b/nvflare/fuel/hci/client/fl_admin_api.py
@@ -215,7 +215,7 @@ class FLAdminAPI(AdminAPI, FLAdminAPISpec):
                 if data["type"] == "string" or data["type"] == "error":
                     reply_data_list.append(data["data"])
             reply_data_full_response = "\n".join(reply_data_list)
-            if "not authenticated - no user" in reply_data_full_response:
+            if "session timed out" in reply_data_full_response:
                 raise ConnectionRefusedError(reply_data_full_response)
             if "Failed to communicate" in reply_data_full_response:
                 raise ConnectionError(reply_data_full_response)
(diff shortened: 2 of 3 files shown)


nvflare/fuel/hci/server/login.py
```diff
@@ -202,5 +202,9 @@ def pre_command(self, conn: Connection, args: List[str]):
             conn.set_prop(ConnProps.TOKEN, token)
             return True
         else:
-            conn.append_error("not authenticated - no user")
+            conn.append_error(
+                "admin client session timed out after {} seconds of inactivity - logging out".format(
+                    self.session_mgr.idle_timeout
+                )
+            )
```
YuanTingHsieh, line 209: If we gets no session, the reason could be

1. The session was created, but it timed out -> then this error msg here is correct.
2. The session was never created for this client token (due to authentication) -> then this error message is misleading?

SYangster: For 2, if it fails to authenticate and thus never creates the session, wouldn't it never reach that function since it would 
'conn.append_string("REJECT")' which would throw some sort of error?

YuanTingHsieh: Or we could combine the two error messages,
something like "admin client session not found, it could due to session timed out or the user is not authenticated."

nvflare/fuel/hci/client/cli.py
```diff
@@ -266,6 +266,11 @@ def _do_default(self, line):
             ent = entries[0]
             resp = ent.handler(args, self.api)
             self.print_resp(resp)
+            if "data" in resp:
+                for item in resp["data"]:
+                    item_type = item["type"]
+                    if item_type == "error" and "session timed out" in item["data"]:
```
yanchengnv, line 272: We should use the "return code" in the resp (not "data" ) to communicate this kind of error.
The return code should be defined as a string without spaces (e.g. session_timeout).
Please check with Kevin about usage of the "return code".

nvflare/fuel/hci/client/fl_admin_api.py
```diff
@@ -215,7 +215,7 @@ def _get_processed_cmd_reply_data(self, command) -> Tuple[bool, str, Dict[str, A
                 if data["type"] == "string" or data["type"] == "error":
                     reply_data_list.append(data["data"])
             reply_data_full_response = "\n".join(reply_data_list)
-            if "not authenticated - no user" in reply_data_full_response:
+            if "session timed out" in reply_data_full_response:
```
yanchengnv, line 218: Use "return code" - see my previous comment.

nvflare/fuel/hci/server/login.py
```diff
@@ -202,5 +202,9 @@ def pre_command(self, conn: Connection, args: List[str]):
             conn.set_prop(ConnProps.TOKEN, token)
             return True
         else:
-            conn.append_error("not authenticated - no user")
+            conn.append_error(
```
yanchengnv, line 205: Use "return code" - check with Kevin how to do it.

yanchengnv (changes requested):

SYangster:
Update: We've decided to have the admin client logout without needing a user command first, which requires additional changes. Closing PR for now

PR #209, opened 2022-02-15
admin client automatically logouts after timeout

SYangster:
- Admin client now automatically logouts after timeout with needing a user input


nvflare/fuel/hci/client/cli.py
```diff
@@ -348,8 +374,7 @@ def cmdloop(self, intro=None):
                         except EOFError:
                             line = "bye"
                         except KeyboardInterrupt:
-                            self.stdout.write("\n")
-                            line = "\n"
+                            return
```
YuanTingHsieh, line 377: why are we removing these statements?

nvflare/fuel/hci/server/sess.py
```diff
@@ -147,3 +154,21 @@ def handle_list_sessions(self, conn: Connection, args: List[str]):
                     "{}".format(time.time() - s.last_active_time),
                 ]
             )
+
```
YuanTingHsieh, line 157: might want to change the "handle_list_sessions"'s docstring.

And not sure if list(self.session.values) should be inside the lock.

nvflare/fuel/hci/server/login.py
```diff
@@ -176,8 +176,8 @@ def handle_logout(self, conn: Connection, args: List[str]):
         conn.append_string("OK")
 
     def pre_command(self, conn: Connection, args: List[str]):
-        if args[0] in [LOGIN_CMD_NAME, CERT_LOGIN_CMD_NAME]:
-            # skip login command
+        if args[0] in [LOGIN_CMD_NAME, CERT_LOGIN_CMD_NAME, "_check_session"]:
+            # skip login and check session command
```
YuanTingHsieh, line 180: what about "list_sessions" command?

We can consider make "list_sessions" and "_check_session" as constants in sess.py

For example: `CHECK_SESSION_CMD_NAME = "_check_session"`
and then we just import that constants to be used here.

nvflare/fuel/hci/client/cli.py
```diff
@@ -118,6 +121,25 @@ def __init__(
...
+            time.sleep(60)
+            res = self.api.check_session()
+
+            for item in res["data"]:
+                item_type = item["type"]
+                if item_type == "string":
+                    if item["data"] == "REJECT":
+                        session_active = False
+                elif item_type == "error":
+                    self.write_error(item["data"])
```
YuanTingHsieh, line 139: if this suddenly shows up in admin console would users be confused about what the message is for?

SYangster: This error would only be present for the response of check_session() if the session timed out.

nvflare/fuel/hci/server/sess.py
SYangster: I don't think python has any read locks unfortunately, but I think it should be fine just to use the same lock.

nvflare/fuel/hci/client/cli.py
```diff
@@ -118,6 +121,25 @@ def __init__(
             poc=poc,
         )
 
+        self._session_monitor = threading.Thread(target=self._check_session, args=())
+        self._session_monitor.daemon = True
```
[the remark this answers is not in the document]
SYangster: I was thinking that since this thread is just polling the server, it would fit as a daemon thread to run in the background that would be easily killed if the main thread exits.

nvflare/fuel/hci/client/cli.py
SYangster: What is the behavior we want on KeyboardInterrupt? I changed it to exit when catching a KeyboardInterrupt while waiting at the input() call since I had the thread raise signal.SIGINT when the session was expired.

nvflare/fuel/hci/client/api.py
```diff
@@ -164,6 +165,28 @@ def _load_client_cmds(self, cmd_modules):
     def register_command(self, cmd_entry):
         self.all_cmds.append(cmd_entry.name)
 
+    def session_monitor(self, session_callback):
+        _session_monitor_thread = threading.Thread(target=self._check_session, args=(session_callback,), daemon=True)
+        _session_monitor_thread.start()
```
YuanTingHsieh, line 170: If the api is a long lived object in the system,
after someone called api.session_monitor() it will start this session monitor thread,
I think we would expect there is a method that can end / join this thread back, rather than close this thread when the whole process end?

SYangster: Yes I definitely agree. I just spoke with Yan about this as well.

nvflare/fuel/hci/client/cli.py
```diff
@@ -132,6 +136,17 @@ def __init__(
...
+        signal.signal(signal.SIGUSR1, partial(self.session_signal_handler))
+
+    def session_ended(self, message):
+        self.write_error(message)
+        os.kill(os.getpid(), signal.SIGUSR1)
+
+    def session_signal_handler(self, signum, frame):
+        if self.session_monitor_thread.is_alive():
+            self.session_monitor_thread.join()
+        sys.exit()
```
YuanTingHsieh, line 148: Since the start of session monitor thread is defined in api's "session_monitor" method.

Is it possible to put the join thread logic in api's method?

So that who uses api does not need to worry about this?

YuanTingHsieh:
Since this is a new feature, we might want to merge this back into the 2.1 dev branch,
what do you think @IsaacYangSLA

nvflare/fuel/hci/client/cli.py
SYangster: Good idea. I can move the join logic to the api or I can still return the thread and instead let the user both start and join. Either way works depending on what we want.

nvkevlu:
Works well for the admin client. For the FLAdminAPI, the message that returns to the user ends up being "Exception: None" for a RuntimeError because _get_processed_cmd_reply_data in fl_admin_api.py does not look at the data because it checks for if reply.get("status") != APIStatus.SUCCESS when it gets APIStatus.ERROR_INACTIVE_SESSION and raises RuntimeError . Maybe we can just move the block to the bottom so it is after where you checked for "session timed out" in the full response.

nvflare/fuel/hci/client/api.py
```diff
@@ -167,19 +170,29 @@ def _load_client_cmds(self, cmd_modules):
     def register_command(self, cmd_entry):
         self.all_cmds.append(cmd_entry.name)
 
-    def session_monitor(self, session_callback, interval=5):
-        _session_monitor_thread = threading.Thread(
+    def start_session_monitor(self, session_callback, interval=5):
+        if isinstance(self.sess_monitor_thread, threading.Thread) and self.sess_monitor_thread.is_alive():
```
YuanTingHsieh, line 174: Can this check be simplified as 
`if self.sess_monitor_thread and self.sess_monitor_thread.is_alive()`

SYangster: Sure

nvflare/fuel/hci/client/api.py
```diff
@@ -164,9 +170,56 @@ def _load_client_cmds(self, cmd_modules):
...
+            self.sess_monitor_thread.join()
+
+    def _check_session(self, session_callback, interval):
+        error_msg = ""
+        connection_error_counter = 0
+        while True:
+            time.sleep(interval)
+
+            if not self.sess_monitor_active:
+                return
```
YuanTingHsieh, line 194: What if user call `start_session_monitor` and then call `close_session_monitor` right away?

In that case the `self.api_active` remains True and `session_callback` is not called?

nvflare/fuel/hci/client/api.py
```diff
@@ -367,6 +422,9 @@ def do_command(self, command):
         return self.server_execute(command)
 
     def server_execute(self, command, reply_processor=None):
+        if not self.api_active:
+            return {"status": APIStatus.ERROR_INACTIVE_SESSION, "details": "API session is inactive"}
```
YuanTingHsieh, line 426: I think the name `api_active` should be changed for clarity.
Since this actually means if the server session is active or not.

For example, the client side commands can actually run without issues even if server session is inactive.

nvflare/fuel/hci/client/api.py
SYangster: If they call these sequentially in a single thread it will be ok, but I agree, if they call them in separate threads there could be issues. I can add a lock.

nvflare/fuel/hci/client/api.py
SYangster: Sure, how about `server_session_active`

nvflare/fuel/hci/client/api.py
YuanTingHsieh: got it, I think we can assume single thread for this api.
In this case, do we need this if statement?

SYangster: (resolved offline) Assuming api is single threaded in this pr, but may change to support multi-threads in the future.

YuanTingHsieh (approved):
LGTM, it would be good later to add some tests into CI/CD

nvflare/fuel/hci/client/api.py
```diff
@@ -164,9 +170,56 @@ def _load_client_cmds(self, cmd_modules):
...
+            self.close_session_monitor()
+        self.sess_monitor_thread = threading.Thread(
+            target=self._check_session, args=(session_ended_callback, interval), daemon=True
+        )
+        self.sess_monitor_active = True
+        self.sess_monitor_thread.start()
+
+    def close_session_monitor(self):
+        self.sess_monitor_active = False
+        if self.sess_monitor_thread:
```
yanchengnv, line 184: Change to:
`if self.sess_monitor_thread and self.sess_monitor_thread.is_alive():`

nvflare/fuel/hci/client/api.py
```diff
@@ -164,9 +170,56 @@ def _load_client_cmds(self, cmd_modules):
...
+            self.sess_monitor_thread.join()
+
```
yanchengnv, line 186: Add:
`self.sess_monitor_thread = None`
after join.

nvflare/fuel/hci/client/api.py
```diff
@@ -164,9 +170,56 @@ def _load_client_cmds(self, cmd_modules):
...
+                break
+
+        self.server_sess_active = False
+        session_ended_callback(error_msg)
+
     def logout(self):
         """Send logout command to server."""
-        return self.server_execute("_logout")
+        resp = self.server_execute("_logout")
+        self.server_sess_active = False
```
yanchengnv, line 221: Since self.server_sess_active is set to False here, should the _check_session() method also check this flag? If it's False, then perhaps the monitor should end?

nvflare/fuel/hci/server/login.py
```diff
@@ -202,5 +202,10 @@ def pre_command(self, conn: Connection, args: List[str]):
             conn.set_prop(ConnProps.TOKEN, token)
             return True
         else:
-            conn.append_error("not authenticated - no user")
+            conn.append_string("session_inactive")
```
yanchengnv, line 205: These two appends seem to be reversed.  Should be:
`conn.append_error("session_inactive")
conn.append_string("user not authenticated or ...")
`

Basically, the "error" should be terse and easy for machine to check; whereas "string" is for end users.

If change this, you also need to change the logic in API to check the error properly.

yanchengnv (changes requested):

nvflare/fuel/hci/client/api.py
SYangster: If this flag gets set to False, then the next time _check_session() calls server_execute(), then it will receive APIStatus.ERROR_INACTIVE_SESSION and then the monitor will end.

nvflare/fuel/hci/server/login.py
```diff
@@ -176,8 +176,8 @@ def handle_logout(self, conn: Connection, args: List[str]):
         conn.append_string("OK")
 
     def pre_command(self, conn: Connection, args: List[str]):
-        if args[0] in [LOGIN_CMD_NAME, CERT_LOGIN_CMD_NAME]:
-            # skip login command
+        if args[0] in [LOGIN_CMD_NAME, CERT_LOGIN_CMD_NAME, CHECK_SESSION_CMD_NAME, LIST_SESSIONS_CMD_NAME]:
```
yanchengnv, line 179: I don't think the LIST_SESSIONS_CMD_NAME should be included here, because this command is issued by the user himself. This means that the user is still active, and the "last_active_time" should be updated.

yanchengnv (changes requested):

YuanTingHsieh (changes requested):
As @IsaacYangSLA suggested,
Let's create a branch from 2.1 and apply your commits on top of that.
So we get a clean commit history in 2.1.

YuanTingHsieh:
Close because @SYangster move the changes into #244

PR #210, opened 2022-02-16
Add persistence to admin cli

SYangster:
- added persistent command history to admin client cli
- uses readline module (only Unix compatible)

### DIFF
--- a/nvflare/fuel/hci/client/cli.py
+++ b/nvflare/fuel/hci/client/cli.py
@@ -22,6 +22,11 @@ from datetime import datetime
 from enum import Enum
 from typing import List, Optional
 
+try:
+    import readline
+except ImportError:
+    readline = None
+
 from nvflare.fuel.hci.cmd_arg_utils import join_args, split_to_args
 from nvflare.fuel.hci.reg import CommandModule, CommandModuleSpec, CommandRegister, CommandSpec
 from nvflare.fuel.hci.security import hash_password, verify_password
@@ -50,22 +55,6 @@ class CredentialType(str, Enum):
 
 
 class AdminClient(cmd.Cmd):
-    """Admin command prompt for submitting admin commands to the server through the CLI.
-
-    Args:
-        host: cn provisioned for the project, with this fully qualified domain name resolving to the IP of the FL server
-        port: port provisioned as admin_port for FL admin communication, by default provisioned as 8003, must be int
-        prompt: prompt to use for the command prompt
-        ca_cert: path to CA Cert file, by default provisioned rootCA.pem
-        client_cert: path to admin client Cert file, by default provisioned as client.crt
-        client_key: path to admin client Key file, by default provisioned as client.key
-        server_cn: server cn
-        require_login: whether to require login
-        credential_type: what type of credential to use
-        cmd_modules: command modules to load and register
-        debug: whether to print debug messages. False by default.
-    """
-
     def __init__(
         self,
         host,
@@ -79,7 +68,26 @@ class AdminClient(cmd.Cmd):
         credential_type: str = CredentialType.PASSWORD,
         cmd_modules: Optional[List] = None,
         debug: bool = False,
+        cli_history_dir: str = "",
+        cli_history_size: int = 1000,
     ):
+        """Admin command prompt for submitting admin commands to the server through the CLI.
+
+        Args:
+            host: cn provisioned for the project, with this fully qualified domain name resolving to the IP of the FL server
+            port: port provisioned as admin_port for FL admin communication, by default provisioned as 8003, must be int
+            prompt: prompt to use for the command prompt
+            ca_cert: path to CA Cert file, by default provisioned rootCA.pem
+            client_cert: path to admin client Cert file, by default provisioned as client.crt
+            client_key: path to admin client Key file, by default provisioned as client.key
+            server_cn: server cn
+            require_login: whether to require login
+            credential_type: what type of credential to use
+            cmd_modules: command modules to load and register
+            debug: whether to print debug messages. False by default.
+            cli_history_dir: directory path to save the cli history file
+            cli_history_size: the maximum number of commands to save in the cli history file. Defaults to 1000.
+        """
         cmd.Cmd.__init__(self)
         self.intro = "Type help or ? to list commands.\n"
         self.prompt = prompt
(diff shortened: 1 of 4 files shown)


nvflare/fuel/hci/client/cli.py
```diff
@@ -79,6 +84,8 @@ def __init__(
         credential_type: str = CredentialType.PASSWORD,
         cmd_modules: Optional[List] = None,
         debug: bool = False,
+        cli_history_dir: str = "",
+        cli_history_size: int = 1000,
     ):
```
YuanTingHsieh, line 89: Can you move the docstring from class level under `__init__` and add the new arguments?
Thanks!

SYangster: Done

nvflare/poc/admin/startup/fl_admin.sh
```diff
@@ -10,4 +10,4 @@ fi
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 echo $DIR
 mkdir -p $DIR/../transfer
-python3 -m nvflare.fuel.hci.tools.admin --host ${server} --port 8003 --prompt "> " --with_file_transfer --upload_dir=$DIR/../transfer --download_dir=$DIR/../transfer --with_shell --with_login
+python3 -m nvflare.fuel.hci.tools.admin --host ${server} --port 8003 --prompt "> " --with_file_transfer --upload_dir=$DIR/../transfer --download_dir=$DIR/../transfer --with_shell --with_login --cli_history_dir=$DIR
```
IsaacYangSLA, line 13: Can you also update the master_template.yml in lighter module so both POC and provisioned files contain this new option?

IsaacYangSLA (changes requested):

nvflare/poc/admin/startup/fl_admin.sh
SYangster: Done

IsaacYangSLA (dismissed):
LGTM.

nvkevlu (approved):

nvflare/fuel/hci/client/cli.py
```diff
@@ -118,6 +126,12 @@ def __init__(
             poc=poc,
         )
 
+        if readline and cli_history_dir and os.path.exists(cli_history_dir):
+            self.cli_history_file = os.path.join(cli_history_dir, ".cli_history")
```
yanchengnv, line 130: A common practice is to define a ".nvflare" folder in the user's home dir. Then you can keep anything in there.

nvflare/fuel/hci/client/cli.py
```diff
@@ -206,6 +220,14 @@ def complete(self, text, state):
         results = [x + " " for x in self.api.all_cmds if x.startswith(text)] + [None]
         return results[state]
 
+    def preloop(self):
+        if readline and self.cli_history_file and os.path.exists(self.cli_history_file):
+            readline.read_history_file(self.cli_history_file)
+
+    def postloop(self):
```
yanchengnv, line 227: So if the client crashed before exiting the loop, history won't be written?

nvflare/lighter/impl/master_template.yml
```diff
@@ -177,7 +177,7 @@ fl_admin_sh: |
   #!/usr/bin/env bash
   DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
   mkdir -p $DIR/../transfer
-  python3 -m nvflare.fuel.hci.tools.admin --host {~~cn~~} --port {~~admin_port~~} --prompt "> " --with_file_transfer --upload_dir=$DIR/../transfer --download_dir=$DIR/../transfer --with_shell --with_login --with_ssl --cred_type cert --ca_cert=$DIR/rootCA.pem --client_cert=$DIR/client.crt --client_key=$DIR/client.key
+  python3 -m nvflare.fuel.hci.tools.admin --host {~~cn~~} --port {~~admin_port~~} --prompt "> " --with_file_transfer --upload_dir=$DIR/../transfer --download_dir=$DIR/../transfer --with_shell --with_login --with_ssl --cred_type cert --ca_cert=$DIR/rootCA.pem --client_cert=$DIR/client.crt --client_key=$DIR/client.key --cli_history_dir=$DIR
```
yanchengnv, line 180: if we always save history to ~/.nvflare, then no need to change this file.

nvflare/poc/admin/startup/fl_admin.sh
yanchengnv, line 13: No need here if we save history to ~/.nvflare

yanchengnv (changes requested):

nvflare/fuel/hci/client/cli.py
SYangster: Ah yes, precmd() and postcmd() would probably be better then.

nvflare/fuel/hci/client/cli.py
SYangster: Got it, will change to this.

yanchengnv (approved):
Looks good

nvflare/fuel/hci/client/cli.py
```diff
@@ -22,6 +22,11 @@
 from enum import Enum
 from typing import List, Optional
 
+try:
+    import readline
+except ImportError:
+    readline = None
+
```
YuanTingHsieh, line 29: Why do we have try except block here?

If readline is not built-in, we should make it a requirements and remove all the `if readline` statement so that we can support this cli history feature for all the users.

YuanTingHsieh (changes requested):

nvflare/fuel/hci/client/cli.py
SYangster: readline is only for Unix, so this is to avoid causing errors for Windows or OS X

YuanTingHsieh: I see that the readline module is not supported in Windows system.
In that case, we say we only support cli history feature for Unix and mac OS then?

YuanTingHsieh: We can add a note later in the docstring in a separate PR, but we can merge this now.

YuanTingHsieh (approved):

PR #215, opened 2022-02-16
Add versioneer so nvflare.__version__ has meaningful value

IsaacYangSLA:
After this PR, the build process can be simplified to 

1. git tag 2.0.100 (locally or in github and no PR required)
2. python3 setup.py bdist_wheel
3. publish wheel to pypi


nvflare/__init__.py
```diff
@@ -12,4 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-__version__ = "2.0.7"
+from . import _version
+
+__version__ = _version.get_versions()["version"]
```
YuanTingHsieh, line 17: Does get_versions() always return a dict that has the key "version"?

YuanTingHsieh: Seeing from the code, I think the answer is Yes.

YuanTingHsieh (approved):
LGTM

nvkevlu (approved):

PR #244, opened 2022-02-28
Quit admin after timeout dev

SYangster:
- Admin client now automatically logouts after timeout with needing a user input


YuanTingHsieh (approved):
LGTM.

PR #274, opened 2022-03-09
Update with fed_admin config

nvkevlu:
Still need to move overseer agent into admin API layer but this should be working for now with fed_admin.json at least

### DIFF
--- a/nvflare/fuel/hci/client/cli.py
+++ b/nvflare/fuel/hci/client/cli.py
@@ -70,13 +70,16 @@ class AdminClient(cmd.Cmd):
         require_login: whether to require login
         credential_type: what type of credential to use
         cmd_modules: command modules to load and register
+        overseer_end_point: end point for the overseer in order to find the active server
+        project: project name to provide overseer
+        name: name of the provisioned admin to provide to the overseer
         debug: whether to print debug messages. False by default.
     """
 
     def __init__(
         self,
-        host,
-        port: int,
+        host=None,
+        port: int = None,
         prompt: str = "> ",
         ca_cert=None,
         client_cert=None,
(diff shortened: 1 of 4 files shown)


nvflare/fuel/hci/tools/admin.py
```diff
@@ -12,90 +12,99 @@
...
         cmd_modules=modules,
         ca_cert=ca_cert,
         client_cert=client_cert,
         client_key=client_key,
-        require_login=args.with_login,
-        credential_type=CredentialType.PASSWORD if args.cred_type == "password" else CredentialType.CERT,
+        require_login=admin_config["with_login"],
+        credential_type=CredentialType.PASSWORD if admin_config["cred_type"] == "password" else CredentialType.CERT,
         debug=args.with_debug,
+        overseer_end_point=admin_config["overseer_agent"]["args"]["overseer_end_point"],
```
yhwen, line 105: The Admin config already creates the overseer-agent instance. We should pass the overseer-agent to the AdminClient, instead of the overseer_end_point only.

nvflare/fuel/hci/client/cli.py
```diff
@@ -122,14 +128,17 @@ def __init__(
...
+            name=name,
             debug=self.debug,
             poc=poc,
         )
 
         signal.signal(signal.SIGUSR1, partial(self.session_signal_handler))
 
         self.ssid = None
-        self.overseer_agent = self._create_overseer_agent()
+        self.overseer_agent = self._create_overseer_agent(overseer_end_point, project, name)
```
yhwen, line 141: When creating the AdminClient, we can already get the overseer-agent instance. No need to create here again.

yhwen (approved):

PR #280, opened 2022-03-10
Clean up codes inside nvflare/private/fed/app

YuanTingHsieh:
- Define new constants in `nvflare/private/defs.py` to replace hard coded strings
- Move `server_deployer.py` into nvflare/private/fed/server (because client deployer is in nvflare/private/fed/client)


YuanTingHsieh:
Move the changes into PR #286

PR #286, opened 2022-03-10
Clean up codes in nvflare/private/fed/app

YuanTingHsieh:
- Define new constants in `nvflare/private/defs.py` to replace hard coded strings
- Move `client_deployer.py` into nvflare/private/fed/app/deployer


YuanTingHsieh:
@yhwen can you review this PR? thanks

yhwen (changes requested):
Actually instead of moving:
private/fed/app/trainers/server_deployer.py → private/fed/server/server_deployer.py

We should move:
private/fed/client/base_client_deploy.py -> private/fed/app/trainers/base_client_deployer.py

And rename "trainers" folder to "deployer". These are app related functions.

yhwen (approved):

PR #294, opened 2022-03-11
update admin logout for ha

SYangster:
- add user configured admin timeout
- fix bye/shutdown

### DIFF
--- a/nvflare/fuel/hci/client/api.py
+++ b/nvflare/fuel/hci/client/api.py
@@ -272,7 +272,7 @@ class AdminAPI(AdminAPISpec):
 
     def _check_session(self, session_ended_callback, interval):
         error_msg = ""
-        connection_error_counter = 0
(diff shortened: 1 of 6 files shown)


YuanTingHsieh:
If the server dies and never comeback, the admin will not timeout?

YuanTingHsieh:
If the server dies and never comeback, the admin will not timeout?

nvkevlu:
Sorry for the conflicts, you can delete the comment that is in cli.py.

SYangster:
> If the server dies and never comeback, the CLI will not timeout and close itself?

> If the server dies and never comeback, the admin will not timeout?

I originally had this, but we decided that there are no guarantees on how long the switch will take (wherein it cannot reach a server), so instead we keep waiting.

YuanTingHsieh:
Got it, so we decide not to support #194 ?

PR #302, opened 2022-03-14
Refactor overseer_agent to AdminAPI from cli, add ha commands

nvkevlu:
Refactor overseer_agent to AdminAPI from cli, so it can also be used by the FLAdminAPI. For HA it is now required to enable auto_login, since the admin client needs to login after switching to a new SP.

### DIFF
--- a/nvflare/fuel/hci/client/api.py
+++ b/nvflare/fuel/hci/client/api.py
@@ -247,8 +247,6 @@ class AdminAPI(AdminAPISpec):
     def _load_client_cmds(self, cmd_modules):
         if cmd_modules:
             for m in cmd_modules:
-                if not isinstance(m, CommandModule):
-                    raise TypeError("cmd_modules must be a list of CommandModule, but got element of type {}".format(type(m)))
                 self.client_cmd_reg.register_module(m, include_invisible=False)
         self.client_cmd_reg.finalize(self.register_command)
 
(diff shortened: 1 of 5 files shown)


nvflare/fuel/hci/client/api.py
```diff
@@ -158,13 +198,58 @@ def __init__(
...
+        return False
+
     def _load_client_cmds(self, cmd_modules):
         if cmd_modules:
-            if not isinstance(cmd_modules, list):
-                raise TypeError("cmd_modules must be a list")
             for m in cmd_modules:
                 if not isinstance(m, CommandModule):
-                    raise TypeError("cmd_modules must be a list of CommandModule")
+                    raise TypeError("cmd_modules must be a list of CommandModule, but got element of type {}".format(type(m)))
```
YuanTingHsieh, line 252: we can remove the checks here since you move these checks upfront.

nvflare/fuel/hci/client/api.py
```diff
@@ -135,12 +156,31 @@ def __init__(
...
         self.server_cn = server_cn
         self.debug = debug
 
+        # for overseer agent
+        self.ssid = None
+
         # for login
         self.token = None
         self.login_result = None
+        if auto_login:
```
YuanTingHsieh, line 175: why do we need auto_login, can you explain the purpose of this in the PR?

YuanTingHsieh (approved):
LGTM.

- We need to think about if we still need "host", "port", "server_cn" anymore if we have overseer_agent passed in.
- We need to think about if it is reasonable to pass "password" as an argument to api.

PR #305, opened 2022-03-14
Persist wf_index in server_runner

YuanTingHsieh:
If there are multiple workflows, we need to persist the `wf_index` so that we can resume from there.


YuanTingHsieh:
@yhwen what would happen if the run_number is missing in the snapshot?

yhwen (approved):

PR #306, opened 2022-03-15
Fix filesystem storage object creation bugs

SYangster:
- disallow creation of object at nonempty directory
- disallow creation of object inside preexisting object

### DIFF
--- a/nvflare/app_common/storages/filesystem_storage.py
+++ b/nvflare/app_common/storages/filesystem_storage.py
@@ -110,6 +110,15 @@ class FilesystemStorage(StorageSpec):
         if self._object_exists(full_uri) and not overwrite_existing:
             raise Exception("object {} already exists and overwrite_existing is False".format(uri))
 
+        path_parts = Path(uri).parts
+        for i in range(0, len(path_parts) - 1):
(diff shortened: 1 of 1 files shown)


YuanTingHsieh:
This newly added logic LGTM.


I have questions regarding the existing codes, in some places I see

`full_uri = os.path.join(self.root_dir, uri.lstrip("/"))`


By doing this, if the URI is "/home/XXX" and the root is "hello" we will change this into "hello/home/XXX"?

So we are assuming the leftmost "/" is translated to self.root_dir?

YuanTingHsieh:
This newly added logic LGTM.


I have questions regarding the existing codes, in some places I see

`full_uri = os.path.join(self.root_dir, uri.lstrip("/"))`


By doing this, if the URI is "/home/XXX" and the root is "hello" we will change this into "hello/home/XXX"?

So we are assuming the leftmost "/" is translated to self.root_dir?

SYangster:
Yes self.root_dir is by default "/", I can enforce that self.root_dir is an absolute path.

YuanTingHsieh (approved):

SYangster:
Replaced with #325

PR #309, opened 2022-03-15
Re-organise widgets and handlers

### LINKED ISSUE #202: Improve this debug messages in InTimeModelSelector
we might want to improve this debug message, but this can be done in separate PR.

_Originally posted by @YuanTingHsieh in https://github.com/NVIDIA/NVFlare/pull/127#discussion_r804237821_

YuanTingHsieh:
Fix #202 

- Move things from handlers folder into widgets folder
- Rename IntimeModelSelectionHandler into IntimeModelSelector


nvflare/app_common/widgets/intime_model_selector.py
```diff
@@ -16,20 +16,20 @@
...
 from nvflare.apis.fl_constant import FLContextKey, ReservedKey
 from nvflare.apis.fl_context import FLContext
 from nvflare.apis.shareable import Shareable
 from nvflare.app_common.app_constant import AppConstants
 from nvflare.app_common.app_event_type import AppEventType
+from nvflare.widgets.widget import Widget
 
 
-class IntimeModelSelectionHandler(FLComponent):
+class IntimeModelSelector(Widget):
```
holgerroth, line 27: This name change will brake many examples. Can you add an alias to the old name to keep backward capability?

holgerroth (changes requested):
Looks good in general but I'm worried about backward compatibility. Seems like some breaking changes for many examples.

nvflare/app_common/widgets/intime_model_selector.py
YuanTingHsieh: thanks for the feedback, I've added the alias down there.

holgerroth (approved):
LGTM

PR #330, opened 2022-03-18
Add event_type to AnalyticsSender

### LINKED ISSUE #200: Add an event_type argument in AnalyticsSender
Right now AnalyticsSender is sending default event type "analytix_log_stats".

However, in the ConvertToFedEvent and TBAnalyticsReceiver JSON configuration, the user needs to config the "analytix_log_stats" in the events to work.
It's not natural for user to know where this event is coming from.
If we add an "event_type" argument  in the AnalyticsSender, we can make the event sending, converting and receiving in pairs.
That would be easier for people to understand.

YuanTingHsieh:
Fix #200


holgerroth (approved):
Good idea

PR #332, opened 2022-03-22
Initial checkin of code with job management and admin commands

nvkevlu:


nvflare/private/fed/server/job_cmds.py
```diff
@@ -0,0 +1,120 @@
...
+from nvflare.fuel.hci.conn import Connection
+from nvflare.fuel.hci.reg import CommandModule, CommandModuleSpec, CommandSpec
+from nvflare.fuel.hci.zip_utils import unzip_all_from_bytes
+from nvflare.mt.job_def import JobMetaKey
+
+
+class JobCommandModule(CommandModule):
+    """Command module with commands for job management."""
+
+    def __init__(self, upload_dir: str, download_dir: str):  # , job_def_manager: JobDefManagerSpec):
```
YuanTingHsieh, line 30: why do we need upload_dir and download_dir here?

nvflare/mt/job_def_manager.py
```diff
@@ -0,0 +1,186 @@
...
+
+from nvflare.apis.job_def_manager_spec import JobDefManagerSpec, RunStatus
+from nvflare.apis.storage import StorageSpec
+from nvflare.apis.study_manager_spec import StudyManagerSpec
+from nvflare.mt.job_def import JobMetaKey, Job, job_from_meta
+
+
+class _JobChecker:
+    def check_job(self, meta: dict) -> bool:
+        pass
```
YuanTingHsieh, line 29: suggest make this abstract class

nvflare/mt/job_def_manager.py
```diff
@@ -0,0 +1,186 @@
...
+
+class _StatusChecker(_JobChecker):
+    def __init__(self, status_to_check):
+        self.result = []
+        self.status_to_check = status_to_check
+
+    def check_job(self, meta: dict):
+        if meta[JobMetaKey.STATUS] == self.status_to_check:
+            self.result.append(meta)
+        return True
```
YuanTingHsieh, line 40: These checker looks more like a filter?
The "check_job" always return True.
In this case should we rename them?

nvflare/apis/study_manager_spec.py
```diff
@@ -0,0 +1,87 @@
...
+        self.description = description
+        self.sites = sites
+        self.users = users
+        self.start_time = start_time
+        self.end_time = end_time
+        self.reviewers = reviewers
+        self.create_time = None
+
+
+class StudyManagerSpec(object):
```
YuanTingHsieh, line 40: This should be an abstract class.
And we have a valid implementation somewhere.

But I understand we are doing this now for convenience.

nvflare/mt/job_def_manager.py
```diff
@@ -0,0 +1,186 @@
...
+        self.store = store
+        self.uri_root = uri_root
+        self.result_uri_root = "results"
+        self.study_manager = study_manager
+
+    def job_uri(self, jid: str):
+        return "/" + self.uri_root + "/" + jid
+
+    def job_result_uri(self, jid: str, result_id: str):
+        return "/" + self.uri_root + "/" + self.result_uri_root + "/" + jid + "/" + result_id
```
YuanTingHsieh, line 84: what is this function's usage?

nvflare/mt/job_def.py
```diff
@@ -0,0 +1,112 @@
...
+        Job object.
+
+    """
+    return Job(
+        meta.get(JobMetaKey.JOB_ID),
+        meta.get(JobMetaKey.STUDY_NAME),
+        meta.get(JobMetaKey.RESOURCE_SPEC),
+        meta.get(JobMetaKey.DEPLOY_MAP),
+        meta,
+    )
```
YuanTingHsieh, line 97: What if some of these keys are missing in meta?

nvflare/mt/job_def_manager.py
```diff
@@ -0,0 +1,186 @@
...
+    def update_meta(self, jid: str, meta):
+        self.store.update_meta(uri=self.job_uri(jid), meta=meta, replace=False)
+
+    def list_all(self) -> List[Job]:
+        checker = _AllJobsChecker()
+        self._scan(checker)
+        return checker.result
+
+    def _scan(self, checker: _JobChecker):
+        jid_paths = self.store.list_objects("/" + self.uri_root)
```
YuanTingHsieh, line 153: I suggest we remove the "/" in the front, it should still work.

nvflare/mt/job_def_manager.py
```diff
@@ -0,0 +1,186 @@
...
+class SimpleJobDefManager(JobDefManagerSpec):
+    def __init__(self, study_manager: StudyManagerSpec, store: StorageSpec, uri_root: str = "jobs"):
+        super().__init__()
+        self.store = store
+        self.uri_root = uri_root
+        self.result_uri_root = "results"
+        self.study_manager = study_manager
+
+    def job_uri(self, jid: str):
+        return "/" + self.uri_root + "/" + jid
```
YuanTingHsieh, line 81: I suggest we remove "/" in the front.
For the second "/", suggest use os.path.join or os.sep

nvflare/private/fed/server/server_cmd_modules.py
```diff
@@ -14,13 +14,23 @@
...
 
 class ServerCommandModules:
-    cmd_modules = [ShellCommandModule(), SystemCommandModule(), TrainingCommandModule(), InfoCollectorCommandModule()]
+    cmd_modules = [
+        ShellCommandModule(),
+        SystemCommandModule(),
+        TrainingCommandModule(),
+        JobCommandModule(
+            upload_dir="/workspace/nvflare_provis/workspace/example_project/prod_01/admin@nvidia.com/transfer",
+            download_dir="/workspace/nvflare_provis/workspace/example_project/prod_01/admin@nvidia.com/transferdl",
```
YuanTingHsieh, line 30: Why do we need upload_dir and download_dir here?

nvflare/private/fed/server/job_cmds.py
```diff
@@ -0,0 +1,120 @@
...
+from typing import List
+
+from nvflare.fuel.hci.client.api_status import APIStatus
+from nvflare.fuel.hci.conn import Connection
+from nvflare.fuel.hci.reg import CommandModule, CommandModuleSpec, CommandSpec
+from nvflare.fuel.hci.zip_utils import unzip_all_from_bytes
+from nvflare.mt.job_def import JobMetaKey
+
+
+class JobCommandModule(CommandModule):
```
YuanTingHsieh, line 27: worth discussing: Should this command be put under "mt" folder?

nvflare/apis/job_def_manager_spec.py
```diff
@@ -0,0 +1,81 @@
...
+from abc import ABC, abstractmethod
+from typing import List
+
+from nvflare.mt.job_def import Job, RunStatus
+
+
+class JobDefManagerSpec(ABC):
+    def __init__(self):
+        """Job Definition Management API."""
+        self.job_index = []
```
YuanTingHsieh, line 24: Do we need this init anymore?

nvflare/apis/job_def_manager_spec.py
```diff
@@ -0,0 +1,81 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from abc import ABC, abstractmethod
+from typing import List
+
+from nvflare.mt.job_def import Job, RunStatus
```
YuanTingHsieh, line 18: I think we either
1. move this job_def_manager_spec.py into mt
2. move the Job, RunStatus def inside this file

nvflare/mt/job_def.py
nvkevlu: For now it will just create, we can add more in validations later. I addressed everything else in the commit update.

nvkevlu:
Thanks for the comments, I cleaned things up and addressed the issues as mentioned in the comments with the updated commit.

nvflare/fuel/hci/tools/admin.py
```diff
@@ -100,6 +100,8 @@ def main():
         ca_cert=ca_cert,
         client_cert=client_cert,
         client_key=client_key,
+        upload_dir=admin_config.get("upload_dir"),
+        download_dir=admin_config.get("download_dir"),
```
YuanTingHsieh, line 104: One note here, it need not to be resolved in this PR.

I think we don't need these two folders, and we should remove https://github.com/NVIDIA/NVFlare/blob/dev-2.1/nvflare/fuel/hci/client/api.py#L130-L133

I think the API itself does not need to have "FileTransferModule" built in, what do you think?

YuanTingHsieh, line 104: If the `FileTransferModule` is required for api.py,
let's make it a default just like `HACommandModule`

YuanTingHsieh (approved):
Thanks for the quick update, overall LGTM.

Minor notes worth discussing can be addressed in later PRs.

PR #341, opened 2022-03-23
Storage improvements

SYangster:
- add automated S3 storage test using local minio server
- update storage spec and docstrings
- move validate args to utils

### DIFF
--- a/nvflare/apis/storage.py
+++ b/nvflare/apis/storage.py
@@ -97,7 +97,8 @@ class StorageSpec(ABC):
         Args:
             path: the path to the objects
 
-        Returns: list of URIs of objects
+        Returns:
+            list of URIs of objects
 
         """
         pass
@@ -109,7 +110,8 @@ class StorageSpec(ABC):
         Args:
             uri: URI of the object
 
-        Returns: meta info of the object.
+        Returns:
+            meta info of the object.
 
         Raises exception when:
         - invalid args
@@ -125,7 +127,8 @@ class StorageSpec(ABC):
         Args:
             uri: URI of the object
 
-        Returns: meta info of the object.
+        Returns:
+            meta info of the object.
 
         Raises exception when:
         - invalid args
@@ -141,7 +144,8 @@ class StorageSpec(ABC):
         Args:
             uri: URI of the object
 
-        Returns: data of the object.
+        Returns:
+            data of the object.
 
         Raises exception when:
         - invalid args
@@ -157,7 +161,8 @@ class StorageSpec(ABC):
         Args:
             uri: URI of the object
 
-        Returns: meta info and data of the object.
+        Returns:
+            meta info and data of the object.
 
         Raises exception when:
         - invalid args
--- a/nvflare/apis/storage.py
+++ b/nvflare/apis/storage.py
@@ -12,18 +12,21 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from abc import ABC, abstractmethod
 from typing import ByteString, List, Tuple
 
 
-class StorageSpec(object):
+class StorageSpec(ABC):
+    """Functional spec of object storage.
+
+    An object is identified by a URI (unique resource identifier).
+    Each object contains:
+        - content (data)
+        - meta info that describes the control info of the object.
 
     """
-    This defines the functional spec of object storage.
-    Object - an object is identified by a URI (unique resource identifier).
-    An object has content (data)
-    An object has meta info that describes the control info of the object.
-    """
 
+    @abstractmethod
     def create_object(self, uri: str, data: ByteString, meta: dict, overwrite_existing: bool):
         """Create a new object or update an existing object
 
@@ -36,7 +39,6 @@ class StorageSpec(object):
         Returns:
 
         Raises exception when:
-
         - invalid URI specification
         - invalid args
         - object already exists and overwrite_existing is False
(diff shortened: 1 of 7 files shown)


nvflare/apis/storage.py
```diff
@@ -62,14 +65,14 @@ def update_meta(self, uri: str, meta: dict, replace: bool):
         Returns:
 
         Raises exception when:
-
-        - no such object
         - invalid args
+        - no such object
         - error updating the object
```
YuanTingHsieh, line 70: I think in the spec's docstring we don't need to specify what exception needs to be raised.

As we don't expect that from spec, that should come from the implementation side.

nvflare/app_common/storages/filesystem_storage.py
```diff
@@ -12,44 +12,30 @@
...
 import pickle
 import shutil
-from functools import wraps
 from pathlib import Path
 from typing import ByteString, List, Tuple
 
 from nvflare.apis.storage import StorageSpec
+from nvflare.apis.utils.format_check import validate_class_methods_args
 
+URI_ROOT = os.path.abspath(os.sep)
```
YuanTingHsieh, line 24: suggest `URI_ROOT= "/"`

I think os.sep does not always equal to the root directory of the system.

SYangster: `os.path.abspath` called on `os.sep` does give the root directory of the system regardless of the OS

test/unit_test/app_common/test_storage.py
```diff
@@ -45,50 +46,58 @@ def random_meta():
...
-def teardown_class(cls):
-    tmp_dir.cleanup()
+@pytest.mark.xdist_group(name="storage_tests_group")
+@pytest.mark.parametrize("storage", [FilesystemStorage, S3Storage])
+class TestStorage:
 
+    @classmethod
+    def setup_class(cls):
+        global tmp_dir
+        global minio_ps
```
YuanTingHsieh, line 61: Can we make `tmp_dir`, `minio_ps`, `storages` class variables?
I see they are only being used in the class.

test/unit_test/app_common/test_storage.py
```diff
@@ -45,50 +46,58 @@ def random_meta():
...
+        env = os.environ.update({"MINIO_ROOT_USER": "admin", "MINIO_ROOT_PASSWORD": "password"})
+        for command in commands:
+            minio_ps = subprocess.Popen(command, stderr=subprocess.PIPE, stdout=subprocess.PIPE, env=env)
+            if command[0] != "./minio":
+                minio_ps.communicate()
+
+        time.sleep(15)
+
+        storages[FilesystemStorage] = FilesystemStorage(root_dir=os.path.join(tmp_dir_name, "filesystem-storage"))
+        storages[S3Storage] = S3Storage(endpoint="localhost:9000", access_key="admin", secret_key="password", secure=False, bucket_name="nvflare-storage")
```
YuanTingHsieh, line 83: suggest using "string" as key instead of class
so `storages["FilesystemStorage"]`

test/unit_test/app_common/test_storage.py
```diff
@@ -45,50 +46,58 @@ def random_meta():
...
+tmp_dir = None
+storages = defaultdict()
+minio_ps = None
 
 
-@classmethod
-def teardown_class(cls):
-    tmp_dir.cleanup()
+@pytest.mark.xdist_group(name="storage_tests_group")
+@pytest.mark.parametrize("storage", [FilesystemStorage, S3Storage])
```
YuanTingHsieh, line 55: suggest using string as keys.

nvflare/apis/storage.py
YuanTingHsieh: On the other hand, if we do require users to raise these exception in these situations, then they can stay in the docstring.

nvflare/app_common/storages/filesystem_storage.py
YuanTingHsieh: Got it, sounds good.

nvflare/apis/storage.py
SYangster: Agreed, I will add try-excepts around the minio API calls to throw specified exceptions as well

nvflare/app_common/storages/s3_storage.py
```diff
@@ -137,106 +160,138 @@ def update_data(self, uri: str, data: ByteString):
...
+                - if error accessing object
 
         """
         if not self._object_exists(uri):
             raise RuntimeError("object {} does not exist".format(uri))
 
-        return ast.literal_eval(self.s3_client.stat_object(self.bucket_name, uri)._metadata["x-amz-meta-user_metadata"])
+        try:
+            return ast.literal_eval(
+                self.s3_client.stat_object(self.bucket_name, uri)._metadata["x-amz-meta-user_metadata"]
```
YuanTingHsieh, line 230: How do we get this "x-amz-meta-user_metadata"?

nvflare/app_common/storages/s3_storage.py
```diff
@@ -137,106 +160,138 @@ def update_data(self, uri: str, data: ByteString):
...
-            metadata={"user_metadata": self.get_meta(uri)},
-            part_size=5242880,
-        )
+        try:
+            self.s3_client.put_object(
+                self.bucket_name,
+                uri,
+                data=io.BytesIO(data),
+                length=-1,
+                metadata={"user_metadata": self.get_meta(uri)},
```
YuanTingHsieh, line 179: Suggest using a constant something like
`_META_DATA_KEY = "user_metadata"`

YuanTingHsieh (approved):
Thanks for the improvements, LGTM!

PR #342, opened 2022-03-24
Exit worker process automatically when parent process is killed

### LINKED ISSUE #279: stop_fl.sh does not terminate client side worker process
The stop_fl.sh just does `touch shutdown.fl` , wait 5 seconds, then kill the process of pid.fl.

The worker process of client is not killed / terminated.

Steps to reproduce:
1. Launch server (start.sh)
2. Launch client (start.sh)
3. Using Admin: Upload app
4. Using Admin: Deploy app
5. Using Admin: Start app
6. Use stop_fl.sh

YuanTingHsieh:
Fix #279

- Pass parent_pid into worker process so it can check if parent_pid is different from os.getppid()
- Remove unused codes and cleanup

### DIFF
--- a/nvflare/fuel/hci/tools/admin.py
+++ b/nvflare/fuel/hci/tools/admin.py
@@ -48,14 +48,15 @@ def main():
 
     except ConfigError as ex:
         print("ConfigError:", str(ex))
+        return
 
     try:
         admin_config = conf.config_data["admin"]
     except KeyError:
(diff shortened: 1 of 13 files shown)


yanchengnv:
Please close this PR - we won't do it this way for 2.1.

Please work with Sean to create a PR for 2.0 only. Only minimize changes.

PR #345, opened 2022-03-24
Add study config user app

### LINKED ISSUE #344: Need an app to create study configuration

IsaacYangSLA:
Fixes #344

### DIFF
--- a/nvflare/lighter/study.py
+++ b/nvflare/lighter/study.py
@@ -1,0 +1,124 @@
+# Copyright (c) 2021-2022, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import absolute_import
+
+import argparse
+import json
+import os
+from datetime import datetime
+
+import yaml
+
+
+def get_input(prompt, item_list, multiple=False):
+    answer = input(prompt)
+    result = None
+    if multiple:
+        try:
+            if answer == "":
+                result = []
+            else:
+                result = [item_list[int(i)] for i in answer.split(",")]
+        except BaseException:
+            print("Input contains errors (non-integer or out of index range)")
+    else:
+        try:
+            result = item_list[int(answer)]
+        except ValueError:
+            print(f"Expect integer but got {answer.__class__.__name__}")
+        except IndexError:
+            print("Number out of index range")
+    if result is None:
+        exit(1)
+    return result
+
+
+def main():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-p", "--project_file", type=str, default="project.yml", help="file to describe FL project")
+
+    args = parser.parse_args()
+
+    current_path = os.getcwd()
+
+    # main project file
+    project_file = args.project_file
+    project_full_path = os.path.join(current_path, project_file)
+    if not os.path.exists(project_full_path):
+        print(f"{project_full_path} not found.  Running study requires that file.")
+        exit(0)
+
+    project = yaml.load(open(project_full_path, "r"), Loader=yaml.Loader)
+    api_version = project.get("api_version")
+    if api_version not in [3]:
+        raise ValueError(f"API version expected 3 but found {api_version}")
+
+    admin_list = list()
+    client_list = list()
+    for p in project.get("participants"):
+        if p.get("type") == "admin":
+            admin_list.append(p.get("name"))
+        elif p.get("type") == "client":
+            client_list.append(p.get("name"))
+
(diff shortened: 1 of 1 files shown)


nvflare/lighter/study.py
```diff
@@ -0,0 +1,124 @@
...
+        except BaseException:
+            print("Input contains errors (non-integer or out of index range)")
+    else:
+        try:
+            result = item_list[int(answer)]
+        except ValueError:
+            print(f"Expect integer but got {answer.__class__.__name__}")
+        except IndexError:
+            print("Number out of index range")
+    if result is None:
```
yanchengnv, line 43: If result is None, then it means that user didn't select anything, right? We should print an error message before exit.

nvflare/lighter/study.py
```diff
@@ -0,0 +1,124 @@
...
+    client_list_string = ", ".join([f"{i}:{v}" for i, v in enumerate(client_list)])
+    study_name = input("Please enter the name of this study: ")
+    study_description = input("and brief description: ")
+    contact = get_input(f"select one admin for contact {admin_list_string}: ", admin_list)
+    participanting_admins = get_input(
+        f"select participanting admins (separated by ',') {admin_list_string} ", admin_list, multiple=True
+    )
+    participanting_clients = get_input(
+        f"select participanting clients (separated by ',') {client_list_string} ", client_list, multiple=True
+    )
```
yanchengnv, line 86: We should further validate user's input of participating_admins and participanting_clients. For example, the user could selected the same entry multiple times. We should either disallow this or de-dup the result.

nvflare/lighter/study.py
```diff
@@ -0,0 +1,124 @@
...
+            admin_list.append(p.get("name"))
+        elif p.get("type") == "client":
+            client_list.append(p.get("name"))
+
+    admin_list_string = ", ".join([f"{i}:{v}" for i, v in enumerate(admin_list)])
+    client_list_string = ", ".join([f"{i}:{v}" for i, v in enumerate(client_list)])
+    study_name = input("Please enter the name of this study: ")
+    study_description = input("and brief description: ")
+    contact = get_input(f"select one admin for contact {admin_list_string}: ", admin_list)
+    participanting_admins = get_input(
```
yanchengnv, line 81: Typo: participanting_admins should be participating_admins.

nvflare/lighter/study.py
```diff
@@ -0,0 +1,124 @@
...
+        f"select participanting admins (separated by ',') {admin_list_string} ", admin_list, multiple=True
+    )
+    participanting_clients = get_input(
```
yanchengnv, line 84: Typo: participanting_clients should be participating_clients

nvflare/lighter/study.py
```diff
@@ -0,0 +1,124 @@
...
+        f"select participanting clients (separated by ',') {client_list_string} ", client_list, multiple=True
+    )
+    participanting_clients_string = ", ".join([f"{i}:{v}" for i, v in enumerate(participanting_clients)])
```
yanchengnv, line 87: Typo: participanting_clients_string should be participating_clients_string

nvflare/lighter/study.py
```diff
@@ -0,0 +1,124 @@
...
+        contact=contact,
+        participanting_admins=participanting_admins,
+        participanting_clients=participanting_clients,
+        reviewers=reviewer_dict,
+        start_date=start_date,
+        end_date=end_date,
+    )
+    with open(f"{study_name}.json", "wt") as f:
+        f.write(json.dumps(study_config, indent=2))
+
```
yanchengnv, line 121: Maybe also print a message to tell the user where the generated study file is?

yanchengnv (changes requested):

nvflare/lighter/study.py
```diff
@@ -0,0 +1,124 @@
...
+            participanting_clients,
+            multiple=True,
+        )
+        reviewer_dict[admin] = reviewed_clients
+    start_date = input("input start date of this study (MM/DD/YYYY): ")
+    try:
+        start_date = datetime.strptime(start_date, "%m/%d/%Y").date().isoformat()
+    except:
+        print(f"Expect MM/DD/YYYY but got {start_date}")
+        exit(1)
```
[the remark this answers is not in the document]
IsaacYangSLA: Now every input, except the first two, have a loop so users can enter again if the input is not correct.

nvflare/lighter/study.py
IsaacYangSLA: A message is shown when result is an empty list in multiple=True.  When result is None (exception caught), users will be prompted to enter the answer again.

nvflare/lighter/study.py
IsaacYangSLA: The duplicates are removed.  Since we only allow index as input, if the int(item) has error, it will ask for user input again.

nvflare/lighter/study.py
IsaacYangSLA: Fixed.

nvflare/lighter/study.py
IsaacYangSLA: A message is added.

nvflare/lighter/study.py
IsaacYangSLA: Fixed.

nvflare/lighter/study.py
IsaacYangSLA: Also fixed.

yanchengnv (approved):

PR #350, opened 2022-03-25
move job def related into apis, and add get_apps

nvkevlu:


nvflare/apis/impl/job_def_manager.py
```diff
@@ -15,15 +15,17 @@
...
 from abc import ABC, abstractmethod
 from typing import Any, Dict, List
 
+from nvflare.apis.job_def import Job, JobMetaKey, job_from_meta
+from nvflare.apis.job_def_manager_spec import JobDefManagerSpec, RunStatus
 from nvflare.apis.storage import StorageSpec
 from nvflare.apis.study_manager_spec import StudyManagerSpec
-from nvflare.mt.job_def import Job, JobMetaKey, job_from_meta
-from nvflare.mt.job_def_manager_spec import JobDefManagerSpec, RunStatus
+from nvflare.fuel.hci.zip_utils import unzip_all_from_bytes, zip_directory_to_bytes
```
YuanTingHsieh, line 28: Worth discussing:

Should we move these functionalities in apis/utils?

YuanTingHsieh (approved):

PR #353, opened 2022-03-25
worker process check parent process and exit when parent died

### LINKED ISSUE #279: stop_fl.sh does not terminate client side worker process

YuanTingHsieh:
Fix #279 in 2.0

### DIFF
--- a/nvflare/private/fed/app/client/worker_process.py
+++ b/nvflare/private/fed/app/client/worker_process.py
@@ -16,6 +16,7 @@
 
 import argparse
 import os
+import sys
 import threading
 import time
 import traceback
@@ -64,6 +65,7 @@ def main():
     thread = threading.Thread(target=check_parent_alive, args=(args.parent_pid, stop_event))
     thread.start()
 
+    args.train_config = os.path.join("config", "config_train.json")
     config_folder = kv_list.get("config_folder", "")
     secure_train = kv_list.get("secure_train", True)
     if config_folder == "":
@@ -71,9 +73,17 @@ def main():
     else:
         args.client_config = os.path.join(config_folder, "config_fed_client.json")
     args.config_folder = config_folder
-    args.train_config = os.path.join("config", "config_train.json")
     args.env = os.path.join("config", "environment.json")
 
+    try:
+        remove_restart_file(args)
+    except BaseException:
+        print("Could not remove the restart.fl / shutdown.fl file.  Please check your system before starting FL.")
+        sys.exit(-1)
+
+    restart_file = os.path.join(args.workspace, "restart.fl")
+    if os.path.exists(restart_file):
+        os.remove(restart_file)
     print("starting the client .....")
 
     deployer = None
@@ -139,6 +149,7 @@ def main():
             handlers=conf.runner_config.handlers,
             conf=conf,
         )
+        federated_client.run_manager = run_manager
 
         with run_manager.new_context() as fl_ctx:
             fl_ctx.set_prop(FLContextKey.CLIENT_NAME, client_name, private=False)
@@ -153,6 +164,10 @@ def main():
             run_manager.add_handler(client_runner)
             fl_ctx.set_prop(FLContextKey.RUNNER, client_runner, private=True)
 
+            # # Start the thread for responding the inquire
+            # federated_client.stop_listen = False
+            # thread = threading.Thread(target=listen_command, args=[federated_client, int(listen_port), client_runner])
+            # thread.start()
             # Start the command agent
             command_agent = CommandAgent(federated_client, int(listen_port), client_runner)
             command_agent.start(fl_ctx)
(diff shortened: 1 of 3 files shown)


nvflare/private/fed/app/client/worker_process.py
```diff
@@ -32,23 +34,37 @@
 from nvflare.private.fed.client.command_agent import CommandAgent
 
 
+def check_parent_alive(parent_pid, stop_event: threading.Event):
+    while True:
+        if stop_event.is_set():
+            break
+        # TODO: this check only valid in UNIX
```
nvidianz, line 41: According to the documentation, this works on Windows since 3.2.

nvflare/private/fed/app/client/worker_process.py
```diff
@@ -32,23 +34,37 @@
...
+        if os.getppid() != parent_pid:
+            # if parent is not alive, kill its worker process
+            os.killpg(os.getpgid(os.getpid()), 9)
```
nvidianz, line 44: Maybe sys.exit() is good enough and works on Windows too.

nvflare/private/fed/app/client/worker_process.py
```diff
@@ -32,23 +34,37 @@
...
+    """Worker process start program."""
     parser = argparse.ArgumentParser()
     parser.add_argument("--workspace", "-m", type=str, help="WORKSPACE folder", required=True)
     parser.add_argument("--startup", "-w", type=str, help="startup folder", required=True)
-
     parser.add_argument(
         "--fed_client", "-s", type=str, help="an aggregation server specification json file", required=True
     )
-
+    parser.add_argument("--parent_pid", type=int, help="parent process id", required=True)
```
nvidianz, line 57: Can we just get ppid on startup and save that?

nvflare/private/fed/app/client/worker_process.py
YuanTingHsieh: Hi @nvidianz,
`os.getppid()` works on Windows that is correct.
But there is a note (https://docs.python.org/3/library/os.html#os.getppid): 

"When the parent process has exited, on Unix the id returned is the one of the init process (1), on Windows it is still the same id, which may be already reused by another process."

I did find a solution to this though, since we already included psutil.
I can use  `psutil.pid_exists`, let me change this.

nvflare/private/fed/app/client/worker_process.py
YuanTingHsieh: When using PyTorch Dataset, we need to do this `os.killpg(os.getpgid(os.getpid()), 9)`
Otherwise only the child process will be killed (the process spawn by torch is not killed.

nvflare/private/fed/app/client/worker_process.py
YuanTingHsieh: sure, thanks for the suggestion

nvflare/private/fed/client/client_executor.py
```diff
@@ -184,25 +182,23 @@ def start_train(self, client, args, app_root, app_custom_folder, listen_port):
...
-            f"{sys.executable} -m nvflare.private.fed.app.client.worker_process -m "
-            + args.workspace
-            + " -w "
-            + self.startup
-            + " -s fed_client.json "
-            " --set" + command_options + " print_conf=True"
+            f"{sys.executable} -m nvflare.private.fed.app.client.worker_process"
+            f" --workspace {args.workspace}"
+            f" --startup {self.startup}"
+            f" --fed_client fed_client.json"
```
IsaacYangSLA, line 188: This one may have coding style error.

IsaacYangSLA (approved):

yhwen (changes requested):
Can we limit the changes only to the relevant worker process terminate after parent process died? Child process creation and management will be re-engineer to support the multi-Run. The change in this PR will be conflict with that.

yhwen (approved):

nvflare/private/fed/app/client/worker_process.py
```diff
@@ -158,6 +175,7 @@ def main():
 
         federated_client.status = ClientStatus.STARTED
         client_runner.run(app_root, args)
+        stop_event.set()
```
yanchengnv, line 178: I believe you need to join the thread here; otherwise the process may hang.

yanchengnv (changes requested):

yanchengnv (changes requested):

nvflare/private/fed/app/client/worker_process.py
```diff
@@ -170,17 +190,20 @@ def main():
             command_agent.shutdown()
         if deployer:
             deployer.close()
+        if federated_client:
+            federated_client.close()
+        if thread and thread.is_alive():
```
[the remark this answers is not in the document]
YuanTingHsieh: thanks for this feedback. I've made the change as suggested.

yhwen (approved):

YuanTingHsieh:
Hi @yanchengnv , can you review again, thanks!

yanchengnv (approved):

PR #355, opened 2022-03-28
Add FedSM example using prostate experiment setting

ZiyueXu77:
Functional first version of FedSM with prostate experiment, but need significant polish.
Two major issues at this time:

1. Workflow using send_and_wait, making it inefficient, needs a new "broadcast_and_wait" with client customization capability to replace the current implementation.

2. Issue with persistor: fedsm needs three persistors, but if referencing the same persistor python code (global model and selector model can use the same script), only the first one will be initialized (that's why there are three custom persistors under pt/persistors and two of them are identical) - needs further investigation.

### DIFF
--- a/examples/FedSM/configs/prostate_fedsm/config/config_fed_server.json
+++ b/examples/FedSM/configs/prostate_fedsm/config/config_fed_server.json
@@ -1,0 +1,136 @@
+{
+  "format_version": 2,
+  "min_clients": 2,
+  "num_rounds": 100,
+  "server": {
+    "heart_beat_timeout": 600
+  },
+  "task_data_filters": [],
+  "task_result_filters": [],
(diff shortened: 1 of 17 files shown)


examples/FedSM/configs/prostate_fedsm/config/config_fed_server.json
```diff
@@ -0,0 +1,136 @@
...
+              2
+            ],
+            "num_res_units": 2
+          }
+        }
+      }
+    },
+    {
+      "id": "persistor_select",
+      "path": "pt.persistors.pt_file_model_persistor_ext.PTFileModelPersistorExt",
```
ZiyueXu77, line 44: remove "_ext" to reproduce the error behavior

PR #359, opened 2022-03-28
Make study spec consistent with study tool

### LINKED ISSUE #358: Minor inconsistency between study config generation tool and study spec

IsaacYangSLA:
Fixes #358


nvflare/apis/study_manager_spec.py
```diff
@@ -12,40 +12,44 @@
...
+        self.participating_clients = participating_clients
+        self.participating_admins = participating_admins
+        self.start_date = start_date
+        self.end_date = end_date
         self.reviewers = reviewers
-        self.create_time = None
+        self.created_at = datetime.utcnow().isoformat()
 
 
 class StudyManagerSpec(object):
```
YuanTingHsieh, line 43: we should make this class an abstract class and its method abstractmethod so who inherits it need to implement those methods.

IsaacYangSLA: This study is concrete, not a spec, I believe.  The study itself is like a dataclass.  It does not have methods to implement.  Also, the __init__ also means the class here is not a spec.

YuanTingHsieh: Thanks, I agree study is a dataclass.
What about "StudyManagerSpec"?

IsaacYangSLA: Now it is an ABC with all methods as abstractmethod.

PR #365, opened 2022-03-29
Rename study to project in all files in lighter

### LINKED ISSUE #364: Conflicting names between jobs and lighter
Study was used in lighter and jobs also starts using the same name.

IsaacYangSLA:
Fixes #364


YuanTingHsieh:
@IsaacYangSLA 

Why are we changing the names from study to project in lighter folder?

What is the definition of "study" and "project"?

IsaacYangSLA:
The concept of study in current lighter is a long-term concept.  However, in the job/study case, the study is a subset of that lighter study.  Therefore, it's better to avoid name reuse.

YuanTingHsieh (approved):
LGTM

PR #382, opened 2022-04-01
Remove pickle from job commands

nvkevlu:


nvflare/fuel/hci/client/file_transfer.py
```diff
@@ -336,7 +335,8 @@ def upload_job(self, args, api: AdminAPISpec):
         data = zip_directory_to_bytes(self.upload_dir, folder_name)
 
         b64str = bytes_to_b64str(data)
-        meta_b64str = bytes_to_b64str(pickle.dumps(meta))
+        serialized_meta = json.dumps(meta).encode("utf-8")
```
YuanTingHsieh, line 338: @nvkevlu  If meta has key that is not of "string" type, would this still work as expected?

YuanTingHsieh (approved):
LGTM

PR #396, opened 2022-04-06
Fix job def and manager specs

YuanTingHsieh:
- Spec: list_all -> get_all_jobs
- Updated docstrings
- Fixed usages
- Fixed format issues in the files changed

Note that we might need to add try block around the store API calls because they might throw exceptions.


nvkevlu (approved):
Thank you YuanTing, this looks great

PR #397, opened 2022-04-07
fix upload and download job commands, add clone_job

nvkevlu:


nvflare/fuel/hci/server/file_transfer.py
```diff
@@ -31,6 +32,7 @@
 from nvflare.fuel.hci.conn import Connection
 from nvflare.fuel.hci.reg import CommandModule, CommandModuleSpec, CommandSpec
 from nvflare.fuel.hci.zip_utils import unzip_all_from_bytes, zip_directory_to_bytes
+from nvflare.private.fed.server.server_engine import ServerEngine
```
YuanTingHsieh, line 35: I think a check to make sure engine is ServerEngine is important as that is the way we ensure engine.job_def_manager exists.

But I think the "fuel" package should not import class from "private"?

nvkevlu: removed the imports for now

YuanTingHsieh: If the file_transfer module depends on server_engine's job_def_manager instance,
should we move the file_transfer module into "private"?

YuanTingHsieh (approved):

YuanTingHsieh:
It is worth discussing that for upload_job and downlaod_job should we put them here or in the job_cmd.py inside the private package.

PR #401, opened 2022-04-08
update the commands in the FLAdminAPI

nvkevlu:


nvflare/fuel/hci/client/fl_admin_api_spec.py
```diff
@@ -110,21 +112,24 @@ def upload_app(self, app: str) -> FLAdminAPIResponse:
...
         Returns: FLAdminAPIResponse
 
         """
         pass
 
     @abstractmethod
-    def deploy_app(self, app: str, target_type: TargetType, targets: Optional[List[str]] = None) -> FLAdminAPIResponse:
+    def deploy_app(
+        self, run_destination: str, app: str, target_type: TargetType, targets: Optional[List[str]] = None
+    ) -> FLAdminAPIResponse:
```
YuanTingHsieh, line 125: Do we still have deploy_app, start_app?
I thought it should only be "upload_job"

nvkevlu: I asked Yuhong and he said that we are still going to have those

nvflare/fuel/hci/client/fl_admin_api_spec.py
```diff
@@ -77,26 +77,28 @@ def check_status(self, target_type: TargetType, targets: Optional[List[str]] = N
...
-        """Sets a current run number.
+    def delete_run(self, run_number: str) -> FLAdminAPIResponse:
+        """Deletes a specified run.
+
+        This deletes the run folder corresponding to the run number on the server and
+        all connected clients. This is not reversible.
 
         Args:
-            run_number: run number in order to set for the current experiment, must be integer greater than 0
+            run_number (str): run number to delete
```
YuanTingHsieh, line 87: If run_number is not a number anymore,
I think we should change it.

Maybe something like "run_name"/"run_id"?

YuanTingHsieh (approved):

PR #403, opened 2022-04-11
The privilege yaml file must be signed and loaded by secure content s…

IsaacYangSLA:
…ervice

### DIFF
--- a/nvflare/lighter/impl/signature.py
+++ b/nvflare/lighter/impl/signature.py
@@ -26,21 +26,22 @@ class SignatureBuilder(Builder):
     can be cryptographically verified to ensure any tampering is detected. This builder writes the signature.pkl file.
     """
 
+    def _do_sign(self, root_pri_key, dest_dir):
+        signatures = sign_all(dest_dir, root_pri_key)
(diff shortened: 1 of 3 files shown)


nvflare/lighter/impl/signature.py
```diff
@@ -27,14 +27,20 @@ class SignatureBuilder(Builder):
     """
 
     def build(self, project: Project, ctx: dict):
+        root_pri_key = ctx.get("root_pri_key")
+
+        overseer = project.get_participants_by_type("overseer", first_only=True)
+        dest_dir = self.get_kit_dir(overseer, ctx)
+        signatures = sign_all(dest_dir, root_pri_key)
+        json.dump(signatures, open(os.path.join(dest_dir, "signature.json"), "wt"))
```
YuanTingHsieh, line 35: just a minor improvement, these three lines are repeating for overseer, server and client
maybe we can consider make a method?

IsaacYangSLA: Good catch.  Thanks.  Just pushed another commit with refactored function.

YuanTingHsieh (approved):
Thanks, LGTM.

PR #405, opened 2022-04-12
Re-factor unit test's folder

YuanTingHsieh:
- Re-factor unit tests following this guideline: package a/b/c/mycode.py _the corresponding test code should be test/a/b/c/mycode_test.py_
- Rename test_xxx.py -> xxx_test.py
- Put all controller related tests into controller_test.py


IsaacYangSLA (approved):
Let's merge this and add other test cases accordingly.

PR #409, opened 2022-04-13
Fix error when running provision with different overseer

IsaacYangSLA:
Cert builder was not initialized with proper issuer


nvflare/lighter/impl/cert.py
```diff
@@ -60,6 +60,7 @@ def initialize(self, ctx):
             )
             self.pub_key = self.pri_key.public_key()
             self.subject = self.root_cert.subject
+            self.issuer = self.subject.get_attributes_for_oid(NameOID.COMMON_NAME)[0].value
```
YuanTingHsieh, line 63: Does this always work for every certs?
Or does this assume the certs were created in a specific way?

IsaacYangSLA: For our persistent state, it is always true.  If it fails, then it's an exception and the provisioning stops because we can not proceed with incorrect root cert.

YuanTingHsieh (approved):

PR #410, opened 2022-04-14
Add system state into overseer reply.

IsaacYangSLA:
Include admin API (for privileged users only) to set system state

### DIFF
--- a/nvflare/ha/overseer/utils.py
+++ b/nvflare/ha/overseer/utils.py
@@ -33,6 +33,20 @@ else:
     from .mem_store import get_all_sp, get_primary_sp, get_sp_by, update_sp  # noqa
 
 
+system_state = "ready"
+
+
+def get_system_state():
+    global system_state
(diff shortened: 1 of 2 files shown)


YuanTingHsieh (approved):

nvflare/ha/overseer/utils.py
```diff
@@ -33,6 +33,20 @@
     from .mem_store import get_all_sp, get_primary_sp, get_sp_by, update_sp  # noqa
 
 
+system_state = "ready"
```
YuanTingHsieh, line 36: If we re-write this file as a class, the system state can be treated as class variable?

IsaacYangSLA: Yes.  But right now, I think it's better to keep it as simple as possible.  Later, if we need to add more functionalities into overseer, we can start to refactor its base utils into different classes.

PR #411, opened 2022-04-14
Scheduler integration

yhwen:


nvflare/private/fed/server/job_runner.py
```diff
@@ -0,0 +1,216 @@
...
+        run_number = job.job_id
+        workspace = os.path.join(self.workspace_root, WorkspaceConstants.WORKSPACE_PREFIX + run_number)
+        count = 1
+        while os.path.exists(workspace):
+            run_number = run_number + "_" + str(count)
+            workspace = os.path.join(self.workspace_root, WorkspaceConstants.WORKSPACE_PREFIX + run_number)
+            count += 1
+
+        for app_name, participants in job.get_deployment().items():
+            app_data = job.get_application(app_name, fl_ctx)
```
YuanTingHsieh, line 66: if here is the only usage of get_application, I would suggest we just use

`engine.get_component(SystemComponents.JOB_MANAGER).get_app`

directly.

nvflare/apis/job_def.py
```diff
@@ -108,14 +111,14 @@ def get_deployment(self) -> Dict[str, List[str]]:
         """
         return self.deploy_map
 
-    def get_application(self, participant, fl_ctx: FLContext) -> bytes:
+    def get_application(self, app_name, fl_ctx: FLContext) -> bytes:
```
YuanTingHsieh, line 114: Since the Job class means only the job config, I think this method should not be in this class.

nvflare/apis/server_engine_spec.py
```diff
@@ -115,3 +115,15 @@ def persist_components(self, fl_ctx: FLContext, completed: bool):
...
+        pass
+
+    def check_client_resources(self, resource_reqs):
+        pass
+
+    def cancel_client_resources(self, resource_check_results, resource_reqs):
+        pass
+
+    def start_client_job(self, client_sites, run_number):
+        pass
```
YuanTingHsieh, line 129: We might need to add docstrings to these methods to explain why they need to be here

nvflare/apis/fl_constant.py
```diff
@@ -222,3 +222,30 @@ class RunProcessKey(object):
...
+    """hard coded file names inside the workspace folder."""
+
+    LOGGING_CONFIG = "log.config"
+    AUDIT_LOG = "audit.log"
+
+    # these two files is used by shell scripts to determine restart / shutdown
+    RESTART_FILE = "restart.fl"
+    SHUTDOWN_FILE = "shutdown.fl"
+
+    WORKSPACE_PREFIX = ""
```
YuanTingHsieh, line 250: is this intended? the workspace_prefix is empty string.

nvflare/apis/job_def.py
```diff
@@ -31,7 +32,7 @@ class RunStatus(str, Enum):
 class JobMetaKey(str, Enum):
     STUDY_NAME = "study_name"
     JOB_ID = "job_id"
-    JOB_NAME = "job_name"
+    JOB_FOLDER_NAME = "job_folder_name"
```
YuanTingHsieh, line 35: I think we also need job name?

nvflare/private/fed/utils/fed_utils.py
```diff
@@ -40,3 +44,22 @@ def make_context_data(fl_ctx):
...
+def deploy_app(app_name, site_name, workspace, app_data):
+    try:
+        dest = os.path.join(workspace, WorkspaceConstants.APP_PREFIX + site_name)
+        # Remove the previous deployed app.
+        if os.path.exists(dest):
+            shutil.rmtree(dest)
+        if not os.path.exists(dest):
+            os.makedirs(dest)
+        unzip_all_from_bytes(app_data, dest)
+        app_file = os.path.join(workspace, "fl_app.txt")
```
YuanTingHsieh, line 58: is there a constant defined for "fl_app.txt"?

nvflare/private/fed/server/training_cmds.py
```diff
@@ -52,14 +52,14 @@ def get_spec(self):
...
-                    visible=True,
-                ),
+                # CommandSpec(
+                #     name=AdminCommandNames.DEPLOY_APP,
+                #     description="deploy FL app to client/server",
+                #     usage="deploy_app run_number app server|client <client-name>|all",
+                #     handler_func=self.deploy_app,
+                #     authz_func=self.authorize_deploy_app,
+                #     visible=True,
+                # ),
```
YuanTingHsieh, line 62: We can just delete this along with the function associated with it.
Since the old codes are in main branch.

yhwen: changed to make the deploy_app command invisible.

nvflare/apis/impl/list_resource_manager.py
```diff
@@ -55,7 +56,7 @@ def check_resources(self, resource_requirement: dict, fl_ctx: FLContext) -> (boo
                 if len(self.resources[k]) < resource_requirement[k]:
                     check_result = False
                     break
-        return check_result, None
+        return check_result,  str(uuid.uuid4())
```
YuanTingHsieh, line 59: we don't need this token in this `ListResourceManager`.

This token would be needed on the ResourceManager that reserves certain resources upon checking, so they need the token to cancel those reserved resources if the job is not deployed.

nvflare/apis/fl_constant.py
yhwen: This is intended. We could decide to add any prefix to it to only change this one place.

nvflare/private/fed/client/client_engine_executor_spec.py
```diff
@@ -20,6 +20,53 @@
...
+            request: request to be sent
+            timeout: number of secs to wait for replies. 0 means fire-and-forget.
+            fl_ctx: FL context
+
+        Returns: a reply Shareable
+
+        """
+        pass
+
+    def fire_and_forget_aux_request(self, topic: str, request: Shareable, fl_ctx: FLContext) -> Shareable:
```
YuanTingHsieh, line 57: This was a helper function.
We need to check its usage and definition to see if we require this method to be implemented by the child classes

nvflare/apis/impl/list_resource_manager.py
yhwen: resource manager need to return a token back to the allocated resources. The allocated resources are used by the resource consumer. Resource consumer needs to check this token matching. This logic is missing in the resource consumer right now.

nvflare/private/fed/client/client_run_manager.py
```diff
@@ -157,6 +157,9 @@ def send_aux_request(self, topic: str, request: Shareable, timeout: float, fl_ct
     def register_aux_message_handler(self, topic: str, message_handle_func):
         self.aux_runner.register_aux_message_handler(topic, message_handle_func)
 
+    def fire_and_forget_aux_request(self, topic: str, request: Shareable, fl_ctx: FLContext) -> Shareable:
+        return self.send_aux_request(topic, request, 0.0, fl_ctx)
+
```
YuanTingHsieh, line 162: this should either be here, or in the 'ClientEngineInternalSpec`

nvflare/private/fed/server/server_engine.py
```diff
@@ -439,7 +444,9 @@ def new_context(self) -> FLContext:
         if self.run_manager:
             return self.run_manager.new_context()
         else:
-            return FLContext()
+            # return FLContext()
+            return FLContextManager(engine=self, identity_name=self.server.project_name, run_num="",
+                                    public_stickers={}, private_stickers={}).new_context()
```
YuanTingHsieh, line 449: what's the difference between this and the previous method of creating a FLContext?

nvflare/apis/server_engine_spec.py
```diff
@@ -115,3 +115,15 @@ def persist_components(self, fl_ctx: FLContext, completed: bool):
 
     def restore_components(self, snapshot: RunSnapshot, fl_ctx: FLContext):
         pass
+
+    def send_admin_requests(self, requests):
+        pass
```
YuanTingHsieh, line 120: note that if this is specified here, we are requiring the ServerEngine implementation to always have an admin.

nvflare/private/fed/client/scheduler_cmds.py
```diff
@@ -0,0 +1,123 @@
...
+        resource_manager = engine.get_component(SystemComponents.RESOURCE_MANAGER)
+        if not isinstance(resource_manager, ResourceManagerSpec):
+            raise RuntimeError(
+                f"resource_manager should be of type ResourceManagerSpec, but got {type(ResourceManagerSpec)}."
+            )
+        with engine.new_context() as fl_ctx:
+            result = Shareable()
+            try:
+                # resource_spec = req.get_header(ShareableHeader.RESOURCE_SPEC)
+                resource_spec = pickle.loads(req.body)
```
YuanTingHsieh, line 115: note that these pickle usages will need to be changed later.

nvflare/private/fed/server/job_cmds.py
```diff
@@ -19,11 +19,14 @@
...
-from nvflare.fuel.hci.reg import CommandModule, CommandModuleSpec, CommandSpec
+from nvflare.fuel.hci.reg import CommandModuleSpec, CommandSpec
 from nvflare.private.fed.server.server_engine import ServerEngine
+from nvflare.private.fed.server.server_engine_internal_spec import ServerEngineInternalSpec
+from nvflare.security.security import Action
+from .training_cmds import TrainingCommandModule
 
 
-class JobCommandModule(CommandModule):
+class JobCommandModule(TrainingCommandModule):
```
YuanTingHsieh, line 29: why do we want to inherit from TrainingCommandModule?

nvflare/private/fed/app/deployer/server_deployer.py
```diff
@@ -98,10 +105,37 @@ def deploy(self, args):
         """
         first_server, services = self.create_fl_server(args, secure_train=self.secure_train)
         services.deploy(args, grpc_args=first_server, secure_train=self.secure_train)
+
+        job_runner = JobRunner(workspace_root=args.workspace)
+        workspace = Workspace(args.workspace, "server", args.config_folder)
+        run_manager = RunManager(
+            server_name=services.project_name,
```
YuanTingHsieh, line 112: Why server_name is project_name?

nvflare/private/fed/server/server_engine_internal_spec.py
```diff
@@ -195,6 +197,16 @@ def set_run_manager(self, run_manager: RunManager):
         """
         pass
 
+    @abstractmethod
+    def set_job_runner(self, job_runner: JobRunner, job_manager: JobDefManagerSpec):
```
YuanTingHsieh, line 201: since it set both job_runner and job_manager,
should we change the name to reflects this?

nvflare/apis/impl/list_resource_manager.py
YuanTingHsieh: The original "token" here is designed for reserve / unreserve resources (like if requesting some cloud resources)
@yanchengnv can correct me if I am wrong.

In this implementation, we are not reserving any resources when we do check_resource.
Thus, the token is unused in this implementation.

Why do we need another token in resource consumer?

YuanTingHsieh:
@yhwen can you also run `python -m isort ./nvflare` and `python -m black ./nvflare` to make sure the code format is good?
thanks

yhwen:
> 

Done

nvflare/private/fed/server/job_cmds.py
yhwen: To inherit the auth methods.

nvflare/apis/server_engine_spec.py
```diff
@@ -115,3 +115,15 @@ def persist_components(self, fl_ctx: FLContext, completed: bool):
...
+
+    def check_client_resources(self, resource_reqs):
```
YuanTingHsieh, line 122: we should add @abstractmethod decorator to these methods so that the child class is required to implement these methods.

YuanTingHsieh, line 122: we should add @abstractmethod decorator to these methods so that the child class is required to implement these methods.

nvflare/apis/client_engine_spec.py
```diff
@@ -64,40 +64,3 @@ def get_all_components(self) -> dict:
 
     def get_widget(self, widget_id: str) -> Widget:
```
YuanTingHsieh, line 65: also suggest add `@abstractmethod` to all the methods that we require the child class to implement

nvflare/private/fed/server/server_engine.py
yhwen: FLContextManager.new_context() will create default properties instead of empty one.

nvidianz (approved):
We can work on the proposed changes in the main repo

PR #414, opened 2022-04-15
Added components for poc mode.

yhwen:


IsaacYangSLA (approved):
Later I will create another PR to update the fed_server.json so study manager will also use component to get store.

PR #418, opened 2022-04-18
Fix store in study manager as it's now a component

IsaacYangSLA:


nvflare/apis/impl/study_manager.py
```diff
@@ -95,6 +91,13 @@ def add_study(self, study: Study, fl_ctx: FLContext) -> Tuple[Optional[Study], s
         serialized_study = self._study_to_bytes(study)
         try:
             store = self._get_store(fl_ctx)
+        except BaseException as e:
+            return None, e.message
+        try:
+            self._existing_studies = json.loads(store.get_data(StudyManager.STORAGE_KEY).decode("utf-8"))
+        except BaseException:
+            self._existing_studies = list()
```
YuanTingHsieh, line 99: Should we define `_existing_studies=None` in the `__init__` method?

IsaacYangSLA: You are right.  Or maybe we should set it to an empty list.

YuanTingHsieh (approved):
LGTM

PR #422, opened 2022-04-20
Fixed the cross_site_validation for multi_run.

yhwen:


nvflare/app_common/workflows/cross_site_model_eval.py
```diff
@@ -128,7 +128,9 @@ def start_controller(self, fl_ctx: FLContext):
 
         # If the list of participating clients is not provided, include all clients currently available.
         if not self._participating_clients:
-            self._participating_clients = [c.name for c in engine.get_clients()]
+            engine.sync_clients_from_main_process()
```
YuanTingHsieh, line 131: Instead of asking user to call this,
Can we put `sync_clients_from_main_process()` inside the `get_clients()` method of the engine?

nvflare/apis/server_engine_spec.py
```diff
@@ -33,6 +33,15 @@ def fire_event(self, event_type: str, fl_ctx: FLContext):
     def get_clients(self) -> List[Client]:
         pass
 
+    @abstractmethod
+    def sync_clients_from_main_process(self):
```
YuanTingHsieh, line 37: Is this really required?

We need this because our implementation has child and parent processes, the user might implement the server engine in another way?

nvflare/app_common/workflows/cross_site_model_eval.py
yhwen: No, we need to keep both methods. The behavior is different when calling from parent or child process.

nvflare/apis/server_engine_spec.py
yhwen: These are 2 separate method calls. Need to expose both of them to the user.

YuanTingHsieh: Is this method only required in the child process's engine? 

If it is we should not put this in spec.

YuanTingHsieh: I think we should not put this in spec if this method only needed in child process, what do you think @yanchengnv ?

yhwen: The spec serves the purpose of defining the method needs to be implemented in the ServerEngine. it also serves the purpose for user to know what's function available to use. This is the case for user needs to know this method needs to be called first, then called the get_clients().

Another option is to wrap these to another method only will be used in the child process.

nvidianz (approved):

nvflare/apis/server_engine_spec.py
YuanTingHsieh: In this version, we did not distinguish parent and child process in server.
Both of them are creating FederatedServer and ServerEngine.
We should just create FederatedServer at the parent process.
And ServerEngine should be separate into 2 classes as some methods are only for parent process and some for child process.

PR #423, opened 2022-04-20
Update setup.py for release

IsaacYangSLA:


IsaacYangSLA:
Hold this as dependencies are not finalized.

IsaacYangSLA:
This PR updates metadata for pypi site (classifiers), replaces test with tests on the exclude section and other release info.

PR #426, opened 2022-04-21
Add default env var PYTHONPATH if it is not set

YuanTingHsieh:
- Not all environment will set the PYTHONPATH environment variable


IsaacYangSLA (approved):
Thanks.  LGTM.

yhwen (approved):

PR #427, opened 2022-04-21
Tests run with PASSED

IsaacYangSLA:
All credits go to @SYangster .


tests/integration_test/validators/sag_result_validator.py
```diff
@@ -51,6 +51,42 @@ def check_sag_results(server_data, run_data):
     return True
 
 
+def ha_check_sag_results(server_data, run_data):
```
YuanTingHsieh, line 54: if this logic is the same as "SagResultValidator" maybe we just use that for now?

tests/integration_test/system_test.py
```diff
@@ -46,7 +47,8 @@ def read_yaml(yaml_file_path):
 
 params = [
     # "./test_examples.yml",
-    "./test_internal.yml"
+    # "./test_internal.yml"
+    "./test_ha.yml"
```
YuanTingHsieh, line 51: can we uncomment ./test_internal.yml?

tests/integration_test/validators/sag_result_validator.py
IsaacYangSLA: The results are different because we couldn't stop the server at exact same location.  It runs one more round.

tests/integration_test/system_test.py
IsaacYangSLA: I think it's OK.

YuanTingHsieh (approved):
LGTM

PR #428, opened 2022-04-21
Remove unused app constant

YuanTingHsieh:
- Code cleanup: Remove unused constants

### DIFF
--- a/nvflare/app_common/app_constant.py
+++ b/nvflare/app_common/app_constant.py
@@ -24,36 +24,12 @@ class ExecutorTasks:
 
 class AppConstants(object):
 
-    CONFIG_PATH = "config_path"
-    MODEL_NETWORK = "model_network"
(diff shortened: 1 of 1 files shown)


nvidianz (approved):

yhwen (commented):
There are many items defined here are serving general application purpose. Some of them are still being used in clara-train. We should not simply remove them.

YuanTingHsieh:
I think these constants are NOT serving for "general" purpose.
They only serves for "clara-train" now.

We might want to discuss and come up with a principle on what can we put in the app_constant.py file.


I don't think we should put anything inside here.

YuanTingHsieh:
I think these constants are NOT serving for "general" purpose.
They only serves for "clara-train" now.

We might want to discuss and come up with a principle on what can we put in the app_constant.py file.


I don't think we should put anything inside here.

PR #430, opened 2022-04-22
Update hello-numpy-sag from app to job structure and update docs

YuanTingHsieh:
- Update from app to job
- Fix some docs error, more details need to be added by @nvkevlu


nvkevlu (approved):
command name may need to be changed, but that can be done later with other additions too. Thanks

PR #432, opened 2022-04-22
Update commands

nvkevlu:
- add clone and list jobs to FLAdminAPI
- remove usage of password
- add shutdown system command to set shutdown in overseer


IsaacYangSLA (approved):
LGTM on the shutdown portion.

YuanTingHsieh:
Can you run unit tests and integration tests locally and see what's the issue?

YuanTingHsieh (changes requested):
Let's make sure the unit tests and integration tests are passing

IsaacYangSLA:
In POC mode, we are using the DummyOverseerAgent, which does not require oversseer to operate.  #406 is to use the actual HttpOverseerAgent.  It may not apply this this case.

YuanTingHsieh (approved):
Thanks! LGTM, @IsaacYangSLA note that this PR might has minor conflicts with your ongoing PR.

PR #436, opened 2022-04-25
Clean up job scheduler, resource manager, and resource consumers

YuanTingHsieh:
- Move JobScheduler, ResourceManager, ResourceConsumer from apis into app_common
- Allow empty resource spec in meta.json


nvflare/lighter/project.yml
```diff
@@ -76,7 +76,7 @@ builders:
       components:
         server:
           - id: job_scheduler  # This id is reserved by system.  Do not change it.
-            path: nvflare.apis.impl.job_scheduler.DefaultJobScheduler
+            name: DefaultJobScheduler
```
IsaacYangSLA, line 79: I still prefer our current decision to use path in stead of name for all components and classes.  Not sure if we are going back to use name.

IsaacYangSLA (changes requested):

nvflare/lighter/project.yml
YuanTingHsieh: no problem, then I will change to path

PR #438, opened 2022-04-25
Download first to avoid race condition

YuanTingHsieh:
Download CIFAR10 dataset first to avoid race condition when multiple clients are running on the same machine


IsaacYangSLA (approved):
LGTM.

PR #443, opened 2022-04-26
Add to FLAdminAPI, fix FLAdminAPI runner

nvkevlu:
- Update FLAdminAPI Runner to work with provisioned startup package and fed_admin.json instead of individual args
- add abort_job command into FLAdminAPI
- add auth for delete_job and clone_job

### DIFF
--- a/nvflare/fuel/hci/client/api.py
+++ b/nvflare/fuel/hci/client/api.py
@@ -488,8 +488,10 @@ class AdminAPI(AdminAPISpec):
         elif len(entries) == 1:
             self.set_command_result(None)
             ent = entries[0]
-            ent.handler(args, self)
+            return_result = ent.handler(args, self)
             result = self.get_command_result()
+            if return_result:
+                return return_result
             if result is None:
                 return {"status": APIStatus.ERROR_RUNTIME, "details": "Client did not respond"}
             return result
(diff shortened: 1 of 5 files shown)


YuanTingHsieh (changes requested):
Please make sure conflicts are resolved and integration tests are passing.
Thanks!

nvkevlu:
Cancelled these tests after seeing them take over 2 hours and not complete. They did pass on my local, so not sure why these did not. It seems to be cuda memory issue?

YuanTingHsieh:
Hi @nvkevlu ,

Thanks for the clarification.
Yes, the CI machine has CUDA OOM error because someone is running on it.
Got it, if it works on your machine then it is good.

nvflare/fuel/hci/client/api.py
```diff
@@ -488,8 +488,10 @@ def do_command(self, command):
         elif len(entries) == 1:
             self.set_command_result(None)
             ent = entries[0]
-            ent.handler(args, self)
+            return_result = ent.handler(args, self)
```
YuanTingHsieh, line 491: when will this be used?

what client command has a return result?

YuanTingHsieh, line 491: when will this be used?

what client command has a return result?

nvkevlu: this is for the client side commands, so the File transfer, but now HA, the result is returned.

YuanTingHsieh: Got it, the "Admin client" side commands.

YuanTingHsieh (approved):
LGTM

PR #445, opened 2022-04-26
FLARE-136: Fixed the exception in list_jobs when study is None

nvidianz:


nvflare/private/fed/server/job_cmds.py
```diff
@@ -193,8 +193,8 @@ def _send_summary_list(self, conn: Connection, jobs: List[Job]):
             table.add_row(
                 [
                     job.meta.get("job_id"),
-                    job.meta.get("name"),
-                    job.meta.get("study_name"),
+                    job.meta.get("name") or "",
+                    job.meta.get("study_name") or "",
```
IsaacYangSLA, line 197: Maybe job.meta.get("study_name", "") is better.

nvidianz: Definitely better. Updated.

IsaacYangSLA (approved):
LGTM.  A small side note that `not empty string` is True.

nvidianz:
> LGTM. A small side note that `not empty string` is True.

Yeah. That's the desired behavior.

PR #453, opened 2022-04-27
Enhance Job_runner logging.

yhwen:
Add log info for deploy and start the app to clients.

### DIFF
--- a/nvflare/private/fed/server/job_runner.py
+++ b/nvflare/private/fed/server/job_runner.py
@@ -104,7 +104,7 @@ class JobRunner(FLComponent):
         replies = admin_server.send_requests(requests, timeout_secs=admin_server.timeout)
         return replies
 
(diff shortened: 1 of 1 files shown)


nvflare/private/fed/server/job_runner.py
```diff
@@ -123,6 +125,8 @@ def _start_run(self, run_number, job: Job, client_sites: list, fl_ctx: FLContext
         if not replies:
             raise RuntimeError("Failed to start the App to the clients")
 
+        display_sites = ",".join(list(client_sites.keys()))
```
YuanTingHsieh, line 128: is this arg `client_sites` a list or a dict?
The type hinting is saying it is a list

yhwen: Fixed the type hint.

IsaacYangSLA (approved):
LGTM.

PR #455, opened 2022-04-28
Fix integration tests

YuanTingHsieh:
- Clean up integration tests
- Remove "PIPE" to avoid "bad file descriptor" error

### DIFF
--- a/nvflare/fuel/hci/client/api.py
+++ b/nvflare/fuel/hci/client/api.py
@@ -210,7 +210,6 @@ class AdminAPI(AdminAPISpec):
 
                 thread = threading.Thread(target=self._login_sp)
                 thread.start()
-                thread.join()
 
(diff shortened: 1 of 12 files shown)


nvflare/fuel/hci/client/api.py
```diff
@@ -210,6 +210,7 @@ def _set_primary_sp(self, sp: SP):
 
                 thread = threading.Thread(target=self._login_sp)
                 thread.start()
+                thread.join()
```
nvkevlu, line 213: At first I thought this was good because it allows for the init to finish after login was successful. However, there was the reason that the login_sp was in a thread to begin with, and that is the callback for the overseer agent should be quick and not wait according to Isaac.

nvidianz (approved):

PR #460, opened 2022-04-28
Add back HA test cases

YuanTingHsieh:
- Add back HA test cases into CI/CD

### DIFF
--- a/.github/workflows/premerge.yml
+++ b/.github/workflows/premerge.yml
@@ -31,10 +31,8 @@ jobs:
       - run: pip install -r requirements-dev.txt
       - run: |
           export PYTHONPATH=$(pwd)
-          rm -rf /tmp/snapshot-storage
           cd tests/integration_test
           ./run_integration_tests.sh
-          rm -rf /tmp/snapshot-storage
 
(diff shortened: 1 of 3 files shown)


nvkevlu (approved):

.github/workflows/premerge.yml
```diff
@@ -31,10 +31,8 @@ jobs:
       - run: pip install -r requirements-dev.txt
       - run: |
           export PYTHONPATH=$(pwd)
-          rm -rf /tmp/snapshot-storage
           cd tests/integration_test
           ./run_integration_tests.sh
-          rm -rf /tmp/snapshot-storage
```
nvidianz, line 37: Should we create an unique folder for this test run so multiple tests can run simultaneously?

nvidianz (approved):

.github/workflows/premerge.yml
YuanTingHsieh: Good suggestion.
This folder is either configured in project.yaml and use provision command for setup.

Or we do string substitute in fed server/client JSON file to replace those values configured in the poc folder

PR #467, opened 2022-05-02
Support abs path and sort list_jobs result

nvidianz:


nvflare/private/fed/server/job_cmds.py
```diff
@@ -111,7 +111,9 @@ def list_jobs(self, conn: Connection, args: List[str]):
                     conn.append_error("No jobs matching the searching criteria")
                     return
 
-                if len(filtered_jobs) == 1 or parsed_args.d:
+                filtered_jobs.sort(key=lambda job: job.meta.get("submit_time", 0.0))
```
YuanTingHsieh, line 114: for "submit_time", might want to use the constant variable defined in job_def.py?

nvflare/fuel/hci/client/file_transfer.py
```diff
@@ -302,8 +302,7 @@ def submit_job(self, args, api: AdminAPISpec):
         # zip the data
         data = zip_directory_to_bytes(self.upload_dir, folder_name)
 
-        rel_path = os.path.relpath(full_path, self.upload_dir)
-        folder_name = remove_leading_dotdot(rel_path)
+        folder_name = split_path(full_path)[1]
```
YuanTingHsieh, line 305: What if the full_path is "/home/someuser/somefolder/job"?

Then folder_name becomes "someuser"?

YuanTingHsieh: Upon reading: https://docs.python.org/3/library/os.path.html#os.path.split

I figure it will return the last piece.

nvflare/fuel/hci/zip_utils.py
```diff
@@ -39,12 +40,29 @@ def _path_join(base: str, *parts: str) -> str:
...
 
 
-def remove_leading_dotdot(path):
+def remove_leading_dotdot(path: str) -> str:
     while path.startswith("../"):
         path = path[3:]
     return path
 
 
+def split_path(path: str) -> (str, str):
```
YuanTingHsieh, line 49: Do we need to consider if the path is only "/"?

YuanTingHsieh (changes requested):
Please fix this to pass the CI integration tests, thanks!

nvflare/fuel/hci/zip_utils.py
chesterxgchen: it would be better just write a simply unit tests

nvflare/private/fed/server/job_cmds.py
nvidianz: Done

nvflare/fuel/hci/zip_utils.py
```diff
@@ -157,13 +179,13 @@ def convert_legacy_zip(zip_data: bytes) -> bytes:
...
         with ZipFile(writer, "w") as out_zip:
-            out_zip.writestr(meta_path, META_DATA.format(Study.DEFAULT_STUDY_NAME, folder_name))
+            out_zip.writestr(meta_path, META_DATA.format(Study.DEFAULT_STUDY_NAME, folder_name, folder_name))
             # Push everything else to a sub folder with the same name:
             # hello-pt/README.md -> hello-pt/hello-pt/README.md
             for info in info_list:
                 name = info.filename
                 content = in_zip.read(name)
-                path = _path_join(folder_name, name)
+                path = folder_name + "/" + name
```
YuanTingHsieh, line 188: Should we either use os.path.join or os.path.sep ?

nvflare/fuel/hci/zip_utils.py
```diff
@@ -39,12 +40,29 @@ def _path_join(base: str, *parts: str) -> str:
...
+def split_path(path: str) -> (str, str):
+    """Split a path into prefix and folder name
+
+    Args:
+        path: Path to split
+
+    Returns: A tuple of (prefix, folder_name)
+    """
+
+    if path.endswith("/"):
```
YuanTingHsieh, line 58: Same here, but can be in a follow up improvement PR since we have other places using this "/"

nvidianz: In ZIP file, path is always separated by slashes, no back slashes.

YuanTingHsieh (approved):
LGTM.

We can think of how to support more OS later.

PR #468, opened 2022-05-02
Changed the job scheduler to use event.

yhwen:
Changed the Job Scheduler to use event base add / remove job. Updated the JobSchedulerSpec to delete the remove_job().

### DIFF
--- a/nvflare/app_common/job_schedulers/job_scheduler.py
+++ b/nvflare/app_common/job_schedulers/job_scheduler.py
@@ -15,7 +15,9 @@
 import threading
 from typing import Dict, List, Optional, Tuple
 
+from nvflare.apis.event_type import EventType
 from nvflare.apis.fl_component import FLComponent
+from nvflare.apis.fl_constant import FLContextKey
 from nvflare.apis.fl_context import FLContext
 from nvflare.apis.job_def import Job
 from nvflare.apis.job_scheduler_spec import DispatchInfo, JobSchedulerSpec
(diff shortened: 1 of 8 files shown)


nvflare/private/fed/server/server_engine.py
```diff
@@ -525,17 +525,20 @@ def register_aux_message_handler(self, topic: str, message_handle_func):
...
+                for t in self.get_clients():
+                    targets.append(t.name)
+            if targets:
+                return self.run_manager.aux_runner.send_aux_request(
+                    targets=targets, topic=topic, request=request, timeout=timeout, fl_ctx=fl_ctx
+                )
+            else:
+                return {}
+        except:
+            self.logger.error(f"Failed to send the aux_message: {topic}")
```
YuanTingHsieh, line 541: it might be good to have exception information here:

```
except Exception as e:
    self.logger.error(f"Failed to send the aux_message: {topic} with exception: {e}.")
```

nvflare/private/fed/server/job_runner.py
```diff
@@ -189,6 +189,7 @@ def run(self, fl_ctx: FLContext):
                     if ready_job:
```
YuanTingHsieh, line 189: for this "run" function,
if the job_manager is None in L181,
this function will then be doing nothing, it is not getting new values of job_manager.

nvflare/app_common/job_schedulers/job_scheduler.py
```diff
@@ -116,6 +118,17 @@ def __init__(
         self.scheduled_jobs = []
         self.lock = threading.Lock()
 
+    def handle_event(self, event_type: str, fl_ctx: FLContext):
+        if event_type == EventType.JOB_STARTED:
+            with self.lock:
+                job_id = fl_ctx.get_prop(FLContextKey.CURRENT_JOB_ID)
+                self.scheduled_jobs.append(job_id)
```
yanchengnv, line 125: Should check against whether the job_id is already in self.scheduled_jobs, and append it only if the job_id is not in it.

nvflare/private/fed/app/fl_conf.py
```diff
@@ -230,6 +240,7 @@ def __init__(
         )
 
         self.components = {}  # id => component
+        self.handlers = []
```
YuanTingHsieh, line 243: what are these handlers for?

Do we have an example of how these are used?

yanchengnv (changes requested):

nvflare/private/fed/server/server_engine.py
yhwen: done.

nvflare/private/fed/app/fl_conf.py
yhwen: This is to enable client side also define the components in fed_client.json. We don't have the exact usage yet.

nvflare/app_common/job_schedulers/job_scheduler.py
yhwen: done

yanchengnv (approved):

PR #482, opened 2022-05-04
Update install requires

IsaacYangSLA:


setup.py
```diff
@@ -82,10 +82,12 @@
...
         "grpcio",
         "google-api-python-client",
-        "cryptography",
+        "cryptography>=36.0.0",
         "tenseal==0.3.0",
         "gunicorn",
-        "flask"
+        "flask",
+        "PyJWT",
+        "minio==7.1.6"
```
YuanTingHsieh, line 90: This is not required if users are using file system storage or other storages they implemented

YuanTingHsieh (changes requested):

setup.py
IsaacYangSLA: I agree.  However, this requirement is tricky as we might also argue that HE, gunicorn or even PyYAML are not required if users run nvflare differently.  Because we don't have optional import/installation in dev-2.1, this is required if our goal is still to install all dependencies for all components.

nvkevlu (approved):

YuanTingHsieh (approved):

IsaacYangSLA:
Thanks.  Will merge it now.

PR #485, opened 2022-05-05
Fix list resource manager and add tests

YuanTingHsieh:
- Update ListResourceManager
  - When checking, add those resources into "reserved resources" and return a token
  - For each reserved resource, check expiration time periodically
  - Add unit tests

- Call resource_consumer.consume before start app

### DIFF
--- a/nvflare/apis/resource_manager_spec.py
+++ b/nvflare/apis/resource_manager_spec.py
@@ -29,8 +29,6 @@ class ResourceManagerSpec(ABC):
     def check_resources(self, resource_requirement: dict, fl_ctx: FLContext) -> (bool, Optional[str]):
         """Checks whether the specified resource requirement can be satisfied.
 
-        This is to respond to the Scheduler's request to check resource availabilities
-
         Args:
             resource_requirement: a dict that specifies resource requirement
             fl_ctx: the FLContext
@@ -45,9 +43,7 @@ class ResourceManagerSpec(ABC):
 
     @abstractmethod
     def cancel_resources(self, resource_requirement: dict, token: str, fl_ctx: FLContext):
-        """Cancels resources reserved if any.
-
-        This happens when the Scheduler check and Dispatcher decided to not dispatch the job to the client.
+        """Cancels reserved resources if any.
 
         Args:
             resource_requirement: a dict that specifies resource requirement
@@ -61,10 +57,7 @@ class ResourceManagerSpec(ABC):
 
     @abstractmethod
     def allocate_resources(self, resource_requirement: dict, token: str, fl_ctx: FLContext) -> dict:
-        """Allocate resources.
-
-        This is called by the Dispatcher on the client after the Server deploys a job to the client.
-        The Deployer calls this method to allocate/obtain resources for the requirement (committed when check_resource).
+        """Allocates resources.
 
         Note:
             resource requirements and resources may be different things.
@@ -81,7 +74,7 @@ class ResourceManagerSpec(ABC):
 
     @abstractmethod
     def free_resources(self, resources: dict, token: str, fl_ctx: FLContext):
-        """Free the resources.
+        """Frees resources.
 
         Args:
             resources: resources to be freed
--- a/nvflare/app_common/resource_managers/list_resource_manager.py
+++ b/nvflare/app_common/resource_managers/list_resource_manager.py
@@ -13,20 +13,23 @@
 # limitations under the License.
 
 import uuid
+from collections import deque
+from threading import Lock
 from typing import Dict, List, Optional
 
+from nvflare.apis.fl_component import FLComponent
 from nvflare.apis.fl_context import FLContext
 from nvflare.apis.resource_manager_spec import ResourceManagerSpec
 
 
-class ListResourceManager(ResourceManagerSpec):
+class ListResourceManager(ResourceManagerSpec, FLComponent):
     """Manage a list of resource units.
 
     For example:
 
-        - require 2, resource is [0, 1, 2, 3, 4, 5] -> check if things in list is available => return [0,1]
-        - require 3 => return [2, 3, 4]
-        - free 2, require 3 => return [0, 1, 5]
+        - require 2, current resources is [0, 1, 2, 3, 4, 5] => return [0,1]
+          after allocation the current resources become [2, 3, 4, 5]
+        - require 3, current resources [2, 3, 4, 5] => return [2, 3, 4]
 
     """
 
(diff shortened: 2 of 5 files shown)


nvflare/apis/resource_manager_spec.py
```diff
@@ -81,7 +74,7 @@ def allocate_resources(self, resource_requirement: dict, token: str, fl_ctx: FLC
 
     @abstractmethod
     def free_resources(self, resources: dict, token: str, fl_ctx: FLContext):
-        """Free the resources.
+        """Frees resources.
```
chesterxgchen, line 77: frees ? should be just "free"

YuanTingHsieh: Why should this be "free"?

I follow this: https://google.github.io/styleguide/pyguide.html#383-functions-and-methods

chesterxgchen: I see. OK

nvflare/app_common/resource_managers/list_resource_manager.py
```diff
@@ -12,66 +12,138 @@
...
-        - require 3 => return [2, 3, 4]
-        - free 2, require 3 => return [0, 1, 5]
+        - require 2, current resources is [0, 1, 2, 3, 4, 5] => return [0,1]
+          after allocation the current resources become [2, 3, 4, 5]
+        - require 3, current resources [2, 3, 4, 5] => return [2, 3, 4]
 
     """
 
-    def __init__(self, resources: Dict[str, List]):
+    def __init__(self, resources: Dict[str, List], expiration_period: int = 600):
```
chesterxgchen, line 38: is there a central configuration to define all default values ? or we just hard-code the default value directly ?

nvflare/app_common/resource_managers/list_resource_manager.py
```diff
@@ -12,66 +12,138 @@
...
     def check_resources(self, resource_requirement: dict, fl_ctx: FLContext) -> (bool, Optional[str]):
         if not isinstance(resource_requirement, dict):
             raise TypeError(f"resource_requirement should be of type dict, but got {type(resource_requirement)}.")
+
         check_result = True
-        for k in resource_requirement:
-            if k in self.resources:
-                if len(self.resources[k]) < resource_requirement[k]:
+        token = None
+        with self.lock:
```
chesterxgchen, line 100: L100 - L110, 
1)   should we refactor this code into a function that can easily add unit test to test logic
for example: 
def   check_available_resources (resource_requirement : dict)  -> Boolean : 
        ...
        return check_result
2) check_result seems to be a "global" flag that covers all type of resource requirements
      suppose we have two types of resources : gpu, cpu
      if the gpu is not available check_results = False, then cpu will get the same Check_result = False. 
      Is this intended ?  should the check_result be a dictionary, that express something like
           check_result['gpu'] = False
           check_result['cpu'] = True 
? 

if this make sense, the above function signature would be 
def   check_available_resources (resource_requirement : dict)  -> dict : 

and the overall function signature : probably need to change to 
def check_resources(self, resource_requirement: dict, fl_ctx: FLContext) -> (dict, Optional[str]):

nvflare/app_common/resource_managers/list_resource_manager.py
```diff
@@ -12,66 +12,138 @@
...
+                        self.log_debug(fl_ctx, f"Resource {k} is not enough.")
+                        break
+                else:
                     check_result = False
+                    self.log_debug(fl_ctx, f"Missing {k} in resources.")
                     break
-        return check_result, str(uuid.uuid4())
+
+        # reserve resource only when check is True
+        if check_result:
```
chesterxgchen, line 113: same comment here: 
    if we assume cpu resource is available, but gpu resource is NOT available,  the check_result is still TRUE. 
     The following logic will allocate the resources for both cpu and gpu.

nvflare/app_common/resource_managers/list_resource_manager.py
YuanTingHsieh: From what I know, we don't have a "central" config.

nvflare/app_common/resource_managers/list_resource_manager.py
YuanTingHsieh: For the current logic, if CPU resource is available but GPU is NOT available, then "check_result" will be False.

nvflare/app_common/resource_managers/list_resource_manager.py
YuanTingHsieh: 1. => sure I can do that.

2. => Yes, it is intended, you need to have ALL the resource available for the check to pass.
=> it depends on what we want to do with this ListResourceManager, since the user can write their own ResourceManager, I prefer to keep this class perform this way for simplicity

chesterxgchen: Ok, make sense

nvflare/app_common/resource_managers/list_resource_manager.py
```diff
@@ -12,66 +12,142 @@
...
+            if token and token in self.reserved_resources:
+                reserved_resources, _ = self.reserved_resources.pop(token)
+                for k in reserved_resources:
+                    for i in reserved_resources[k]:
+                        self.resources[k].append(i)
+            else:
+                self.log_debug(fl_ctx, f"Token {token} is not related to any reserved resources.")
         return None
 
     def allocate_resources(self, resource_requirement: dict, token: str, fl_ctx: FLContext) -> dict:
```
chesterxgchen, line 140: both cancel and allocation resources are taken resources from reserved resources. The only difference is one directly return it (use it), another put it back to the general resource pool, It might make sense to write a common function that can be used for both : take_reserved_resource

chesterxgchen (approved):
LGTM

tests/unit_test/app_common/resource_managers/list_resource_manager_test.py
```diff
@@ -0,0 +1,186 @@
...
+        )
+
+    def new_context(self):
+        return self.fl_ctx_mgr.new_context()
+
+    def fire_event(self, event_type: str, fl_ctx: FLContext):
+        pass
+
+
+CHECK_TEST_CASES = [
```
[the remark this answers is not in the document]
YuanTingHsieh: sure, let me make that change in a separate PR

nvflare/app_common/resource_managers/list_resource_manager.py
YuanTingHsieh: Actually, I found that only two lines are duplicated:

```
if token and token in self.reserved_resources:
    reserved_resources, _ = self.reserved_resources.pop(token)
```

And since the condition is needed, I don't think we need a common function that has only one line.

tests/unit_test/app_common/resource_managers/list_resource_manager_test.py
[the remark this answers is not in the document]
YuanTingHsieh: What do you suggest here?

Something like

"Each entry is: (resources, resource_requirement, expected_check_result, expected_reserved_resources)"?

PR #486, opened 2022-05-05
To support Tensor used in the model weights for cross_validation.

yhwen:
To work around a Python internal issue which caused a digest error between process data communication when using the Tensor in the model weights.

### DIFF
--- a/nvflare/private/fed/server/server_commands.py
+++ b/nvflare/private/fed/server/server_commands.py
@@ -22,7 +22,6 @@ from nvflare.apis.fl_constant import AdminCommandNames, FLContextKey, ServerComm
 from nvflare.apis.fl_context import FLContext
 from nvflare.apis.shareable import Shareable
 from nvflare.apis.utils.fl_context_utils import get_serializable_data
-from nvflare.private.fed.utils.numproto import proto_to_bytes
 from nvflare.widgets.widget import WidgetID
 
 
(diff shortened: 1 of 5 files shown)


holgerroth (approved):
LGTM

nvflare/private/fed/server/server_commands.py
```diff
@@ -134,7 +135,7 @@ def process(self, data: Shareable, fl_ctx: FLContext):
             ServerCommandKey.SHAREABLE: shareable,
             ServerCommandKey.FL_CONTEXT: copy.deepcopy(get_serializable_data(fl_ctx).props),
         }
-        return data
```
YuanTingHsieh, line 137: What about other CommandProcessor's return result?

Should they also return pickle.dumps(data)?

If so, is there a cleaner way to do this return?

nvflare/private/fed/server/server_engine.py
```diff
@@ -67,6 +67,18 @@
...
+class ClientConnection:
+    def __init__(self, client):
+        self.client = client
+
+    def send(self, data):
+        data = pickle.dumps(data)
+        self.client.send(data)
+
+    def recv(self):
+        return self.client.recv()
```
YuanTingHsieh, line 79: Since we wrap send side using pickle.dumps,
Do we want to have pickle.loads here in the receive side?

holgerroth (approved):
LGTM. I tested with the cifar10 example, and now cross-site validation runs successfully.

yanchengnv (approved):
LGTM

PR #487, opened 2022-05-05
sync clients in runner_proess.

yhwen:
Sync up the job clients in runner_process. All the controllers no need to call sync_up any more.

### DIFF
--- a/nvflare/app_common/workflows/cross_site_model_eval.py
+++ b/nvflare/app_common/workflows/cross_site_model_eval.py
@@ -128,7 +128,6 @@ class CrossSiteModelEval(Controller):
 
         # If the list of participating clients is not provided, include all clients currently available.
         if not self._participating_clients:
-            engine.sync_clients_from_main_process()
             clients = engine.get_clients()
             self._participating_clients = [c.name for c in clients]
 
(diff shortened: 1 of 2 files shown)


nvflare/private/fed/app/server/runner_process.py
```diff
@@ -129,6 +129,7 @@ def start_server_app(server, args, app_root, run_number, snapshot):
         if not isinstance(server.engine, ServerEngine):
             raise TypeError(f"server.engine must be ServerEngine. Got type:{type(server.engine).__name__}")
         server.engine.create_parent_connection(int(args.conn))
+        server.engine.sync_clients_from_main_process()
```
yanchengnv, line 132: What does sync_clients_from_main_process do? Does it only keep clients applicable to the job?

nvflare/app_common/workflows/cross_site_model_eval.py
```diff
@@ -128,7 +128,6 @@ def start_controller(self, fl_ctx: FLContext):
 
         # If the list of participating clients is not provided, include all clients currently available.
         if not self._participating_clients:
-            engine.sync_clients_from_main_process()
             clients = engine.get_clients()
```
yanchengnv, line 131: Does this call guarantee that only clients applicable to the job are returned?

yanchengnv (changes requested):

nvflare/private/fed/app/server/runner_process.py
yhwen: Yes, sync_clients_from_main_process only keep the clients running the job.

nvflare/app_common/workflows/cross_site_model_eval.py
yhwen: Yes, this only include the clients running the current job.

yanchengnv (approved):

PR #491, opened 2022-05-06
Add unittest cases on lighter Participant class

IsaacYangSLA:


YuanTingHsieh (approved):
LGTM.
It might be also good to add some valid/positive test cases.

PR #496, opened 2022-05-09
Ensure the job exit for abort_job.

yhwen:


nvflare/private/fed/server/job_runner.py
```diff
@@ -231,6 +231,8 @@ def stop_run(self, run_number: str, fl_ctx: FLContext):
                 job_manager.set_status(job.job_id, RunStatus.FINISHED_ABORTED, fl_ctx)
                 del self.running_jobs[run_number]
                 self.fire_event(EventType.JOB_ABORTED, fl_ctx)
+            else:
+                raise RuntimeError(f"Job run: {run_number} does not exist.")
```
yanchengnv, line 235: What's the user experience here? Is this exception caught somewhere and displays right message back to user?

I also see that you have two lists: running_processes and running_jobs. Could they become out of sync? Should they be consolidated into running_jobs, and have the running process info be part of the running job info?

yanchengnv (changes requested):

nvkevlu:
From the user's perspective, they may not know what "Job run" is, so maybe we should stick with "Job ID"

yanchengnv (approved):

PR #502, opened 2022-05-09
Enchanced the file storage write process.

yhwen:


nvidianz (approved):

nvflare/app_common/storages/filesystem_storage.py
```diff
@@ -26,18 +27,20 @@
 
 
 def _write(path: str, content):
+    tmp_name = "_" + str(uuid.uuid4())
```
YuanTingHsieh, line 30: Note: a minor improvement could be have just
`tmp_name = path + "_" + str(uuit.uuid4())`

and use tmp_name afterwards

YuanTingHsieh (approved):

PR #505, opened 2022-05-09
use makedirs to auto make intermediate folders if needed

YuanTingHsieh:
- Use `makedirs` so if intermediate folder is missed, it will also make those folders.


nvkevlu (approved):
LGTM

chesterxgchen (approved):

PR #509, opened 2022-05-11
Enhance filesystem_storage file writting.

yhwen:


nvflare/app_common/storages/filesystem_storage.py
```diff
@@ -27,20 +27,20 @@
 
 
 def _write(path: str, content):
-    tmp_name = "_" + str(uuid.uuid4())
+    tmp_path = "_" + str(uuid.uuid4())
```
nvidianz, line 30: Should it be `tmp_path = path + "_" + str(uuid.uuid4())` ?

yhwen: good catch. fixed,

nvidianz (approved):

PR #520, opened 2022-05-12
Rewrote job validation and added test cases

nvidianz:


yhwen (commented):
This is a question for discussion. If a site has "resource_spec" defined, but it does not appear in the deploy_map in the meta.json. Is this job valid?

tests/unit_test/data/jobs/duplicate_clients/meta.json
```diff
@@ -0,0 +1,22 @@
...
+    },
+    "site-2": {
+      "gpu": 1
+    }
+  },
+  "deploy_map": {
+    "sag1": [
+      "server",
+      "site-a",
+      "site-b"
```
nvkevlu, line 16: is it intentional that "resource_spec" has "site-1" and "site-2" while "deploy_map" has "site-a" and "site-b"?

nvflare/fuel/hci/server/file_transfer.py
```diff
@@ -37,6 +32,7 @@
 from nvflare.fuel.hci.conn import Connection
 from nvflare.fuel.hci.reg import CommandModule, CommandModuleSpec, CommandSpec
 from nvflare.fuel.hci.zip_utils import convert_legacy_zip, unzip_all_from_bytes, zip_directory_to_bytes
+from nvflare.private.fed.server.job_validator import JobValidator
```
YuanTingHsieh, line 35: Should we import nvflare.private from nvflare.fuel package?

nvflare/private/fed/server/job_validator.py
```diff
@@ -0,0 +1,150 @@
...
+
+class JobValidator:
+    """Job validator"""
+
+    def __init__(self, fl_ctx: FLContext):
+        self.fl_ctx = fl_ctx
+        engine = fl_ctx.get_engine()
+        self.study_manager = engine.get_component(SystemComponents.STUDY_MANAGER)
+        if not self.study_manager:
+            raise ValueError("Server configuration error: no study manager is configured")
```
YuanTingHsieh, line 36: this means by the time of creating this component,
The engine is in the fl_ctx and the study manager is in the engine.

@yhwen could this assumption be satisfied?

nvflare/private/fed/server/job_validator.py
```diff
@@ -0,0 +1,150 @@
...
+
+        Returns:
+            Tuple[bool, str, dict]: (is_valid, error_message, meta)
+        """
+
+        meta = {}
+        try:
+            with ZipFile(BytesIO(job_data), "r") as zf:
+                meta_file = job_name + "/meta.json"
+                meta_data = zf.read(meta_file)
```
YuanTingHsieh, line 53: If meta_file does not exist, what will happen?

nvflare/private/fed/server/job_validator.py
```diff
@@ -0,0 +1,150 @@
...
+                meta = json.loads(meta_data)
+                if JobMetaKey.JOB_FOLDER_NAME not in meta:
+                    meta[JobMetaKey.JOB_FOLDER_NAME.value] = job_name
```
YuanTingHsieh, line 56: What would happen if `meta[JobMetaKey.JOB_FOLDER_NAME.value]` has a value different than the submitted job_folder name?

nvflare/private/fed/server/job_validator.py
```diff
@@ -0,0 +1,150 @@
...
+                zip_file.getinfo(zip_folder)
+            except KeyError:
+                raise ValueError(f"App {app} in deploy_map doesn't exist for job {job_name}")
+            client_list.extend([site for site in deployment if site != "server"])
+
+        # Make sure an app is not deployed to multiple clients
+        if client_list:
+            duplicates = [site for site, count in collections.Counter(client_list).items() if count > 1]
+            if duplicates:
+                raise ValueError(f"More than one app are running on clients {duplicates} for job {job_name}")
```
YuanTingHsieh, line 134: Suggest change the error message since no job/app is running at the moment that it is being checked.

nvflare/private/fed/server/job_validator.py
```diff
@@ -0,0 +1,150 @@
...
+        client_list = []
+        for app, deployment in deploy_map.items():
+            zip_folder = job_name + "/" + app + "/config/"
+            try:
+                zip_file.getinfo(zip_folder)
+            except KeyError:
+                raise ValueError(f"App {app} in deploy_map doesn't exist for job {job_name}")
+            client_list.extend([site for site in deployment if site != "server"])
+
+        # Make sure an app is not deployed to multiple clients
```
YuanTingHsieh, line 130: We also need to make sure there can't be two app deploy to "server"

nvflare/private/fed/server/job_validator.py
```diff
@@ -0,0 +1,150 @@
...
+            mandatory_set = set(mandatory_clients)
+            all_clients = self._get_all_clients(meta)
+            if all_clients and not mandatory_set.issubset(all_clients):
+                diff = mandatory_set - all_clients
+                raise ValueError(f"Mandatory clients {diff} are not in the deployment for job {job_name}")
+
+        # Validating all apps exists
+        client_list = []
+        for app, deployment in deploy_map.items():
+            zip_folder = job_name + "/" + app + "/config/"
```
YuanTingHsieh, line 123: Each app target to be deployed to client should have "config_fed_client.json" inside config folder. 

Each app target to be deployed to server should have "config_fed_server.json" inside config folder.

tests/unit_test/data/jobs/duplicate_clients/sag1/config/config_fed_client.json
```diff
@@ -0,0 +1,17 @@
+{
```
YuanTingHsieh, line 1: just wondering if there is a way that we don't need to duplicate all these config_fed_client.json / config_fed_server.json

YuanTingHsieh (changes requested):
Thanks for the effort, I leave a few comments

nvflare/fuel/hci/server/file_transfer.py
nvidianz: This code is totally out of place. It requires some refactoring of the command module to get this working correctly. There is a ticket for that already.

nvflare/private/fed/server/job_validator.py
nvidianz: This validator is constructed when running `submit_job` command. The study manager should be available by that time.

nvflare/private/fed/server/job_validator.py
nvidianz: One will be created if meta.json is missing.

nvflare/private/fed/server/job_validator.py
nvidianz: This is the case when meta.json has a different `job_folder_name` than the actual folder name. The job should still work. Should we disallow it?

nvflare/private/fed/server/job_validator.py
nvidianz: Added the checks.

nvflare/private/fed/server/job_validator.py
nvidianz: Fixed.

nvflare/private/fed/server/job_validator.py
nvidianz: Changed.

nvflare/private/fed/server/job_validator.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+class JobValidator:
+    """Job validator"""
+
+    def __init__(self, fl_ctx: FLContext):
+        self.fl_ctx = fl_ctx
+        engine = fl_ctx.get_engine()
+        self.study_manager = engine.get_component(SystemComponents.STUDY_MANAGER)
+        if not self.study_manager:
```
chesterxgchen, line 38: study manager is not only needed (per decision today). we should remove this check and exception

nvflare/private/fed/server/job_validator.py
chesterxgchen: the read method won't create file, correct ? are you sure the meta.json not exists, it will auto create one ?

nvflare/private/fed/server/job_validator.py
```diff
@@ -0,0 +1,170 @@
...
+        meta = {}
+        try:
+            with ZipFile(BytesIO(job_data), "r") as zf:
+                meta_file = job_name + "/meta.json"
+                meta_data = zf.read(meta_file)
+                meta = json.loads(meta_data)
+                if JobMetaKey.JOB_FOLDER_NAME not in meta:
+                    meta[JobMetaKey.JOB_FOLDER_NAME.value] = job_name
+
+                study_name = meta.get(JobMetaKey.STUDY_NAME.value)
```
chesterxgchen, line 61: don't need study name anymore

nvflare/private/fed/server/job_validator.py
```diff
@@ -0,0 +1,170 @@
...
+
+        except ValueError as e:
+            return False, str(e), meta
+
+        return True, "", meta
+
+    def _validate_study(self, study_name: str, meta: dict) -> None:
+
+        # Default or empty study is always valid
+        if not study_name or study_name == Study.DEFAULT_STUDY_NAME:
```
chesterxgchen, line 78: don't need study

nvflare/private/fed/server/job_validator.py
chesterxgchen: should we separate Job folder key and job_name key ?

nvflare/private/fed/server/job_validator.py
```diff
@@ -0,0 +1,170 @@
...
+                f"Study {study_name} has ended. "
+                f"The end time is {datetime.utcfromtimestamp(study.end_time).isoformat()}"
+            )
+
+    def _validate_deployment(self, meta: dict, zip_file: ZipFile) -> None:
+
+        job_name = meta.get(JobMetaKey.JOB_FOLDER_NAME)
+        deploy_map = meta.get(JobMetaKey.DEPLOY_MAP)
+        if not deploy_map:
+            raise ValueError(f"deploy_map is missing in meta.json for job {job_name}")
```
chesterxgchen, line 112: how to handle deploy_map less situation

tests/unit_test/data/jobs/duplicate_clients/sag1/config/config_fed_client.json
chesterxgchen: always wondering if we should introduce pyhocon, which has include function

nvflare/private/fed/server/job_validator.py
nvidianz: Will remove it in next PR so we keep a commit with studies.

YuanTingHsieh (approved):
Good for now. We can remove study in follow up PRs

nvflare/private/fed/server/job_validator.py
nvidianz: deploy_map is a required filed in meta.json. It can contain empty site list but the map itself can't be empty.

PR #524, opened 2022-05-12
Heartbeat jobs

yhwen:


yanchengnv (changes requested):

nvflare/private/fed/client/communicator.py
```diff
@@ -405,6 +405,9 @@ def send_heartbeat(self, servers, task_name, token, ssid, client_name):
                 while retry > 0:
                     try:
                         self.logger.debug(f"Send {task_name} heartbeat {token}")
+                        run_numbers = engine.get_all_run_numbers()
```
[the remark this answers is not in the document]
yhwen: Yes, that will be an empty list.

nvflare/private/fed/server/fed_server.py
```diff
@@ -634,9 +634,20 @@ def Heartbeat(self, request, context):
             if self.admin_server:
                 self.admin_server.client_heartbeat(token)
 
-            summary_info = fed_msg.FederatedSummary()
+            display_jobs = ",".join(self._sync_client_jobs(request, client_name))
```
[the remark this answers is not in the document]
yhwen: Changed to "job_list".

yanchengnv:
Something to consider:
- Is it possible to send the "run numbers to abort" back to the client as the response of the heartbeat, and the client aborts these runs directly? This will make the server side logic simpler, and makes the whole process more efficient and possibly more reliable.
- Please add some log info messages when such conditions are detected.

nvflare/private/fed/server/fed_server.py
```diff
@@ -634,9 +634,20 @@ def Heartbeat(self, request, context):
...
             return summary_info
 
+    def _sync_client_jobs(self, request, client_name):
+        client_jobs = request.jobs
+        server_jobs = self.engine.run_processes.keys()
+        jobs_need_abort = list(set(client_jobs).difference(server_jobs))
+        with self.engine.new_context() as fl_ctx:
+            job_runner = self.engine.job_runner
+            for run_number in jobs_need_abort:
+                job_runner.abort_client_run(self.engine, run_number, [client_name], fl_ctx)
```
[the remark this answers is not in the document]
yhwen: changed the implementation respond the abort jobs to Heartbeat call. Client use that info to abort the extra jobs.

nvflare/private/fed/client/communicator.py
```diff
@@ -391,7 +391,7 @@ def quit_remote(self, servers, task_name, token, ssid, fl_ctx: FLContext):
                     time.sleep(3)
         return server_message
 
-    def send_heartbeat(self, servers, task_name, token, ssid, client_name):
+    def send_heartbeat(self, servers, task_name, token, ssid, client_name, engine):
```
chesterxgchen, line 394: please add argument signature

yhwen: done

nvflare/private/fed/server/fed_server.py
```diff
@@ -634,9 +633,24 @@ def Heartbeat(self, request, context):
...
+                self.logger.info(
+                    f"These jobs: {display_runs} are not running on the server. "
+                    f"Ask client: {client_name} to abort these runs."
+                )
             return summary_info
 
+    def _sync_client_jobs(self, request):
+        client_jobs = request.jobs
+        server_jobs = self.engine.run_processes.keys()
+        jobs_need_abort = list(set(client_jobs).difference(server_jobs))
```
chesterxgchen, line 652: what happened when server has more jobs than client ?

nvflare/private/fed/client/communicator.py
```diff
@@ -391,7 +391,7 @@ def quit_remote(self, servers, task_name, token, ssid, fl_ctx: FLContext):
                     time.sleep(3)
         return server_message
 
-    def send_heartbeat(self, servers, task_name, token, ssid, client_name):
+    def send_heartbeat(self, servers, task_name, token, ssid, client_name, engine: ClientEngine):
```
chesterxgchen, line 394: I was thinking not just to the newly added arguments, but whole method such as
servers: list, task_name: str, token : str, ssid: str, client_name :string 
and return type

chesterxgchen: I guess this probably doesn't just apply this methods, but all others, you can choose to do it later if you wish

chesterxgchen (approved):

nvflare/private/fed/server/fed_server.py
chesterxgchen: this will be handled in separate PR and dealing it later

yhwen:
Yan and I already discussed on how to proper handle the abort client extra jobs through the Heartbeat call response. Since he is not online, merge this PR for the build.

nvflare/private/fed/server/fed_server.py
```diff
@@ -634,9 +634,24 @@ def Heartbeat(self, request, context):
...
+                del summary_info.abort_jobs[:]
+                summary_info.abort_jobs.extend(abort_runs)
+                display_runs = ",".join(abort_runs)
+                self.logger.info(
+                    f"These jobs: {display_runs} are not running on the server. "
+                    f"Ask client: {client_name} to abort these runs."
+                )
             return summary_info
 
+    def _sync_client_jobs(self, request):
```
YuanTingHsieh, line 649: suggest to change this method name to reflect that it returns "job_to_abort"

nvflare/private/fed/server/fed_server.py
```diff
@@ -634,9 +634,24 @@ def Heartbeat(self, request, context):
...
+        client_jobs = request.jobs
+        server_jobs = self.engine.run_processes.keys()
```
yanchengnv, line 651: This is not quite right, since engine.run_processes return ALL processes. What's needed here are the processes corresponding to jobs running on this particular client! There could be cases (though extremely rare) that the client is running job A and the server is also running job A but for different clients. In this case the job A on the client still should be aborted.

The server already keeps jobs under each client (I think), and server_jobs here should be the jobs for this client.

yanchengnv (commented):
The server_jobs logic isn't quite right.

PR #527, opened 2022-05-13
Init pre-merge scripts

YanxuanLiu:
Init pre-merge CI scripts for blossom-ci


tests/ci_scripts/run_premerge_cpu.sh
```diff
@@ -0,0 +1,117 @@
+#!/bin/bash
+#
+# Copyright (c) 2020-2022, NVIDIA CORPORATION. All rights reserved.
```
pxLi, line 3: `2020-2022` -> `2022`

tests/ci_scripts/run_premerge_gpu.sh
```diff
@@ -0,0 +1,34 @@
+#!/bin/bash
+#
+# Copyright (c) 2020-2022, NVIDIA CORPORATION. All rights reserved.
```
pxLi, line 3: `2020-2022` -> `2022`

pxLi (commented):
So generally we may just need to put the script here

```
pip install -r requirements-dev.txt
./runtest.sh
```

and

```
pip install -r requirements-dev.txt
export PYTHONPATH=\$(pwd)
rm -rf /tmp/snapshot-storage
pushd ${testFolder}
./run_integration_tests.sh
popd
rm -rf /tmp/snapshot-storage
```

instead of copying the content of the original test scripts

pxLi:
also can you replace
`tests/ci_scripts/run_premerge_*.sh`
to 
`ci/run_premerge_*.sh`

we would better to not share the dev tests folder

pxLi:
convert to draft for now since we need more discussion

tests/ci_scripts/run_premerge_gpu.sh
```diff
@@ -0,0 +1,34 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+set -ex
+
+BUILD_TYPE=all
```
pxLi, line 20: currently we do not build_type here until developers require complicated test combination support

could just remove line 20-29

tests/ci_scripts/run_premerge_cpu.sh
```diff
@@ -0,0 +1,30 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+set -ex
+
+
```
pxLi, line 20: nit: remove empty line

tests/ci_scripts/run_premerge_gpu.sh
```diff
@@ -0,0 +1,28 @@
+#!/bin/bash
+#
+# Copyright (c) 2020-2022, NVIDIA CORPORATION. All rights reserved.
```
pxLi, line 3: nit: also init year should be time when it was added `2022`

if modified in the future, then we will update it as 2022-20XX

tests/ci_scripts/run_premerge_gpu.sh
```diff
@@ -0,0 +1,28 @@
...
+#
+
+set -ex
+
+## Integration Tests
+pip install -r requirements-dev.txt
+testFolder = 'tests/integration_test'
+export PYTHONPATH=$PWD
+rm -rf /tmp/snapshot-storage
+cd ${testFolder}
```
pxLi, line 25: nit: could try `pushd` and `popd` to join and leave the current dir and wrap the following script
if you prefer `cd`, then `cd -` to return back to $OLDPWD
in this way, we could always have a consistent workspace start point for all other commands

pxLi (commented):
some nits

pxLi:
one more enhancement we can add is the python envs for different steps, so we always have clean pip env for all tests within the same script

YanxuanLiu:
@pxLi and me are working on the pre-merge workflow of NVFlare. 

We'd like to create two pre-merge shell scripts: 
one for CPU stages `run_premerge_cpu.sh`, it executes existing test scripts, that depends on GPU machines, like Integration Tests `tests/integration_test/run_integration_tests.sh`
the other one for GPU stages `run_premerge_gpu.sh`, it executes other scripts not depend on GPU, like Unit tests `runtest.sh`

I also created a folder called `ci` in the root path of NVFlare. The CI scripts above are stored here.

**The purposes of the two scripts:**
1. Flexible and easy to maintain. In Jenkinsfile, we run the two scripts only, and we don't care the content of the scripts. Therefore, we don't need to modify jenkinsfile frequently, and NVFlare developers can modify the scripts depending on testing requirements.
2. Run CPU and GPU stages concurrently. Jenkins will assign K8s node to execute `run_premerge_cpu.sh`, and GPU nodes to execute `run_premerge_gpu.sh`

@chesterxgchen @YuanTingHsieh 
Hope to get your advice about the solution. Especially we need to agree on the path structure and script content.

YuanTingHsieh:
I think you mean 

`run_premerge_cpu.sh` -> Run scripts that does not need GPU (for example Unit test `runtest.sh`)
`run_premerge_gpu.sh` -> Run scripts that need GPU (for example integration tests `tests/integration_test/run_integration_tests.sh`)

The folder structure:

```
ci/
   run_premerge_cpu.sh
   run_premerge_gpu.sh
```

I think this is good.

pxLi:
@chesterxgchen 
this PR is more like a simple prototype, please help review the idea here. 

If things look good, we will start working on it along w/ internal blossom pipeline. 
The future plan is,
1. Enable blossom-ci as an optional PR check (manual trigger required), and let it run for a few weeks
2. Switch self-hosted runner CI as optional check, and blossom-ci as required one
3. remove self-hosted runners and related workflow files to meet security requirements
4. blossom-ci for premerge only

chesterxgchen:
at this point, this seems to be good. 
While working on the code base, I noticed that ./unittest.sh  seems to be includes license check, style check, set up minio and run unit tests and tear down minio 
We will probably split the code with different functions we split the step into several steps. So that when the test failed, we can quickly learn we failed at style format, license, or tests. before user has to dig into the detailed logs.

ci/run_premerge_cpu.sh
```diff
@@ -0,0 +1,29 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+set -ex
+
+## Unit Tests
+pip install -r requirements-dev.txt
+PYTHONPATH=$PWD
```
pxLi, line 22: should export to keep it consistent w/ other script

pxLi (commented):
Generally LGTM as a init PR

Created a follow-up here for pip env support https://github.com/NVIDIA/NVFlare/issues/578

ci/run_premerge_cpu.sh
YanxuanLiu: Done~

pxLi (approved):

PR #529, opened 2022-05-13
Fix client not online deploy error

yhwen:
Fixed the issue if a client in the deploy_map not online, the job failed to deploy and run issue.

### DIFF
--- a/nvflare/private/fed/server/job_runner.py
+++ b/nvflare/private/fed/server/job_runner.py
@@ -82,13 +82,14 @@ class JobRunner(FLComponent):
                     if p in sites:
                         client_sites.append(p)
 
-            self._deploy_clients(app_data, app_name, run_number, client_sites, engine)
-            display_sites = ",".join(client_sites)
-            self.log_info(
-                fl_ctx,
-                f"Application {app_name} deployed to the clients: {display_sites} for run:{run_number}",
-                fire_event=False,
-            )
+            if client_sites:
+                self._deploy_clients(app_data, app_name, run_number, client_sites, engine)
+                display_sites = ",".join(client_sites)
+                self.log_info(
+                    fl_ctx,
+                    f"Application {app_name} deployed to the clients: {display_sites} for run:{run_number}",
+                    fire_event=False,
+                )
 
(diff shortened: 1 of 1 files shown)


nvkevlu (approved):

nvflare/private/fed/server/job_runner.py
```diff
@@ -77,18 +77,19 @@ def _deploy_job(self, job: Job, sites: dict, fl_ctx: FLContext) -> str:
                         fl_ctx, f"Application {app_name} deployed to the server for run:{run_number}", fire_event=False
                     )
                     if not success:
-                        raise RuntimeError("Failed to deploy the App to the server")
+                        raise RuntimeError(f"Failed to deploy the App: {app_name} to the server")
```
chesterxgchen, line 80: is it possible to tell user why the deployment failed, networking, configuration, etc. etc. 
instead generic deployment failed ?  

Some instructive message, so user knows what failed, and he has next thing to do: fix network, start server, change configuration, fix code etc.

nvflare/private/fed/server/job_runner.py
```diff
@@ -101,7 +102,8 @@ def _deploy_clients(self, app_data, app_name, run_number, client_sites, engine):
         message.set_header(RequestHeader.APP_NAME, app_name)
         replies = self._send_to_clients(admin_server, client_sites, engine, message)
         if not replies:
-            raise RuntimeError("Failed to deploy the App to the clients")
+            display_sites = ",".join(client_sites)
```
chesterxgchen, line 105: ditto

chesterxgchen (approved):
I think we need to gradually improve the error message to give user more detailed failure message, and instructive. For now, I am ok with approve the improvement

nvflare/private/fed/server/job_runner.py
yhwen: How the app deployment works is loading the App data from the Job, and sending thee data to the client to unzip to the workspace. At this point, the Job has passed the validation, the job data structure and configuration meta.json already been validated. So the deployment failure only means the App failed to be sent to the client.

PR #530, opened 2022-05-13
address issues from QA by adding to docs, removing delete_job

nvkevlu:
- remove delete_job for now
- Address ticket to explain relationship between job and app
- Address ticket about adding documentation to describe abort tasks/jobs

### DIFF
--- a/docs/user_guide/job.rst
+++ b/docs/user_guide/job.rst
@@ -23,13 +23,25 @@ Many underlying assumptions about the operation of the system changed, with the 
 system manage the rest, instead of before when the admin uploaded apps, set the run number, deployed, then started the app
 at the server and client sites.
 
-Jobs now contain all of the apps and the information of which apps to deploy to which sites as the "deploy_map" inside
-a meta.json that should be included with a job to be submitted.
+Jobs now contain all of the :ref:`apps<application>` and the information of which apps to deploy to which sites as the
+:ref:`deploy_map<deploy_map>` inside a meta.json that should be included with a job to be submitted::
+
+    JOB_FOLDER:
+        - meta.json
+        - APP_FOLDER_FOR_SERVER    (required to have config/config_fed_server.json)
+        - APP_FOLDER_FOR_CLIENT1   (required to have config/config_fed_client.json)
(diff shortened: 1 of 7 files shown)


nvflare/private/fed/server/job_cmds.py
```diff
@@ -45,13 +45,13 @@ def get_spec(self):
...
-                    handler_func=self.delete_job,
-                    authz_func=self.authorize_job,
-                ),
+                # CommandSpec(
+                #     name="delete_job",
+                #     description="delete a job",
+                #     usage="delete_job job_id",
+                #     handler_func=self.delete_job,
+                #     authz_func=self.authorize_job,
+                # ),
```
YuanTingHsieh, line 54: We can also remove this along with delete_job function since they are in previous commits if we ever want to bring them back.

docs/user_guide/job.rst
```diff
@@ -83,6 +94,22 @@ support the job’s specified resource requirements. Since resource is a generic
 as a resource - NVIDIA FLARE 2.1.0 itself does not define any specific resources. Instead, NVIDIA FLARE provides a general
 framework for resource definition and interpretation.
 
+.. _deploy_map:
+
+Deploy Map
+==========
+The ``deploy_map`` is a map of which apps in the job being uploaded will be deployed to which FL client sites. Back in
+NVIDIA FLARE before 2.1.0, the admin command "deploy_app" was used to manually perform app deployment with the option
+to specify which sites to deploy to. Because the JobRunner now automatically picks up and handles the deployment and
```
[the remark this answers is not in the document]
nvkevlu: I think the context helps those who have used NVFlare before to understand the concept more quickly, and it also provides a setup to introduce what this does more easily.

nvflare/private/fed/server/job_cmds.py
nvkevlu: We are planning on addressing it 2.1.1, so this will be modified soon regardless

YuanTingHsieh (approved):
Thanks for the effort!

PR #531, opened 2022-05-13
Change tmp path under tmp/nvflare

YuanTingHsieh:
- Change `/tmp` to a common root `/tmp/nvflare/`


IsaacYangSLA (approved):
Thanks.  LGTM.

PR #533, opened 2022-05-13
Remove Study-related info from lighter/apis/poc/setup.py/requirements…

IsaacYangSLA:
….txt

Need PR#532.


IsaacYangSLA:
@nvkevlu , sorry that I don't know how to update the ui html file.

nvkevlu:
> @nvkevlu , sorry that I don't know how to update the ui html file.

I can update that

nvflare/apis/impl/job_def_manager.py
```diff
@@ -63,25 +62,14 @@ def __init__(self, reviewer_name, fl_ctx: FLContext):
...
-                return True
-            else:
-                self.result.append(job_from_meta(meta))
-        return True
+        approvals = meta.get(JobMetaKey.APPROVALS)
+        if approvals and self.reviewer_name in approvals:
+            # already reviewed
+            return True
+        else:
+            self.result.append(job_from_meta(meta))
```
YuanTingHsieh, line 72: missing a `return True` in the end

YuanTingHsieh (changes requested):
LGTM, can you address the comment and rebase the code thanks!

nvflare/apis/impl/job_def_manager.py
IsaacYangSLA: Added.  Thanks.

YuanTingHsieh (approved):
Thanks, LGTM

PR #535, opened 2022-05-13
Change the default value of ignore_result_error in SAG workflows

YuanTingHsieh:
- Change the default value of `ignore_result_error` from `True` to `False`


holgerroth (approved):
makes sense

PR #536, opened 2022-05-13
remove study from docs and provisioning UI to match project.yml updates

nvkevlu:


docs/examples/run_example.rst
```diff
@@ -0,0 +1,3 @@
+With the admin client command prompt successfully connected and logged in, enter the command below.  Pay close
+attention to what happens in each of four terminals. You can see how the admin submits the job to the server and how
+the :class:`JobRunner <nvflare.private.fed.server.job_runner.JobRunner>` on the server automatically picks up the job to deploy and start the run.
```
YuanTingHsieh, line 3: Seems like this file is not finished, I guess this file is for another PR?

nvkevlu: it was actually missed in a previous PR, I made some edits to one of the comments to split this out, so I used references, but this file did not make it into the commit because there is an entry in .gitignore for run_

YuanTingHsieh (approved):
Thanks, LGTM

PR #539, opened 2022-05-14
README update

chesterxgchen:
update readme. 
add and extend Quick Start to the front page


nvkevlu:
Thank you for putting together the diagram and the new sections. There are still some issues that we probably need to address though. The README.md is in the NVFlare package but is not actually a part of the rst built docs for html, and index.rst is the main page for that, so what you intended to do may not be the result from these docs changes. Also, we may want to copy edit the image as there are some typos ("feederated computing: task orchestratioon", "configuration mnanagement") and maybe we can make the capitalization standardized (if you have the file to generate this image, I can help edit this). Finally, it looks like the branch you pulled to work on this did not include some updates to the docs that may have been merged afterwards, so it undoes some changes like adding information about 2.1.0 features and to removing commands for 2.0, but only for some files, so maybe there were some versioning conflicts somewhere.

chesterxgchen:
I should have closed this PR, the commit is corrupted.  The intend is used this as discussion point. The diagram also changed quite a bit and I have fixed the typos.

chesterxgchen:
Close the PR for now.

PR #546, opened 2022-05-16
Job validator addition checks

chesterxgchen:
Main changes
1) improve jpb_validator logics and restructure the code to separate into small validation functions
  * validate meta.json exists or not
  * validate deploy_map
  * validate min_clients, including the int within (1, maxsize), min clients < deployed clients
  * validate resource_spec to make sure the specified resources in dict data type.

2) update unit test utils to make sure zip file can easily add different meta.json for test
  * add several unit tests to cover above test cases

3) remove one zip directory to file function from zip_utils.py, it is unused. It with incorrect method signature ( I guess never tested)

4) Update convert_legacy_zip to handle cases where meta.json doesn't contains job_name: we will update the job name if not exits.


nvflare/private/fed/server/job_validator.py
```diff
@@ -44,80 +46,125 @@ def validate(self, job_name: str, job_data: bytes) -> Tuple[bool, str, dict]:
...
 
         except ValueError as e:
             return False, str(e), meta
 
         return True, "", meta
 
-    def _validate_deployment(self, meta: dict, zip_file: ZipFile) -> None:
+    def _validate_meta(self, job_name: str, zf: ZipFile):
+        meta_file = f"{job_name}/{META}"
+        logging.info(f"validate file {meta_file} exists for job {job_name}")
```
nvidianz, line 63: DEBUG level maybe good enough for this message.

nvflare/private/fed/server/job_validator.py
```diff
@@ -44,80 +46,125 @@ def validate(self, job_name: str, job_data: bytes) -> Tuple[bool, str, dict]:
...
+        meta = {}
 
-        job_name = meta.get(JobMetaKey.JOB_FOLDER_NAME)
-        deploy_map = meta.get(JobMetaKey.DEPLOY_MAP)
-        if not deploy_map:
-            raise ValueError(f"deploy_map is missing in meta.json for job {job_name}")
+        if zf.namelist().__contains__(meta_file):
```
nvidianz, line 66: `if meta_file in zf.namelist()` may be better.

YuanTingHsieh (changes requested):
Need to fix for integration tests.

When upload an application as a job, the meta.json will be auto-generated.
In here: https://github.com/NVIDIA/NVFlare/blob/dev-2.1/nvflare/fuel/hci/zip_utils.py#L179

Then job validator validate will throw an error

nvflare/private/fed/server/job_validator.py
chesterxgchen: ok, will change

YuanTingHsieh (changes requested):
as commented

nvflare/fuel/hci/zip_utils.py
```diff
@@ -13,8 +13,10 @@
 # limitations under the License.
 
 import io
+import json
 import os
 from zipfile import ZipFile
+from nvflare.apis.job_def import JobMetaKey
 
 META_NAME = "meta.json"
 # A format string for the dummy meta.json
```
YuanTingHsieh, line 22: maybe this `META_DATA` needs to update: "name" needs to change to `JobMetaKey.JOB_FOLDER_NAME.value`

nvflare/fuel/hci/zip_utils.py
```diff
@@ -165,25 +157,38 @@ def convert_legacy_zip(zip_data: bytes) -> bytes:
...
-                info.filename = path
-                out_zip.writestr(info, content)
+            if meta:
+                out_zip.writestr(meta_path, json.dumps(meta))
+                out_zip.comment = in_zip.comment  # preserve the comment
+                for item in in_zip.infolist():
+                    if item.filename != meta_path:
+                        out_zip.writestr(item, in_zip.read(item.filename))
+            else:
+                out_zip.writestr(meta_path, META_DATA.format(folder_name, folder_name))
```
YuanTingHsieh, line 184: this line does not guarantee we have `JobMetaKey.JOB_FOLDER_NAME` in meta.json

nvflare/fuel/hci/zip_utils.py
chesterxgchen: right that's simpler than what  I did

nvflare/fuel/hci/zip_utils.py
chesterxgchen: since I check if the job_folder_name exist or not, the meta will be None or not None. 
if the name doesn't exists, the meta.json exists, the meta will be not None. so it will flow to the else part.

nvflare/private/fed/server/job_validator.py
```diff
@@ -44,80 +46,130 @@ def validate(self, job_name: str, job_data: bytes) -> Tuple[bool, str, dict]:
...
-                # Validating resource_spec
-                resource_spec = meta.get(JobMetaKey.RESOURCE_SPEC.value)
-                if resource_spec and not isinstance(resource_spec, dict):
-                    raise ValueError(f"Invalid resource_spec for job {job_name}")
+                meta = self._validate_meta(job_name, zf)
+                site_list = self._validate_deploy_map(job_name, meta, zf)
+                clients = self._get_all_clients(site_list)
+                self._validate_min_clients(job_name, meta, clients)
+                self._validate_resource(job_name, meta)
+                self._validate_mandatory_clients(job_name, meta, clients)
```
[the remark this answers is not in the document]
chesterxgchen: this is old comment, we don't want to add here.

nvflare/fuel/hci/zip_utils.py
chesterxgchen: let me try one more time.

chesterxgchen: no, the template already has name in it. this is to cover the case meta.json exists but not have job name. 

Maybe I did not follow your comments

nvflare/fuel/hci/zip_utils.py
```diff
@@ -165,25 +157,38 @@ def convert_legacy_zip(zip_data: bytes) -> bytes:
...
         meta_path = _path_join(folder_name, META_NAME)
         if next((info for info in info_list if info.filename == meta_path), None):
             # Already in job layout
-            return zip_data
+            meta_data = in_zip.read(meta_path)
+            meta = json.loads(meta_data)
+            if JobMetaKey.JOB_FOLDER_NAME not in meta:
+                meta[JobMetaKey.JOB_FOLDER_NAME.value] = folder_name
+            else:
+                return zip_data
```
YuanTingHsieh, line 174: 1. if meta.json exist already, 
    you insert `JobMetaKey.JOB_FOLDER_NAME` if it is not there.
    
  This logic is fine.

nvflare/fuel/hci/zip_utils.py
```diff
@@ -165,25 +157,38 @@ def convert_legacy_zip(zip_data: bytes) -> bytes:
...
     with ZipFile(reader, "r") as in_zip:
         info_list = in_zip.infolist()
         folder_name = info_list[0].filename.split("/")[0]
         meta_path = _path_join(folder_name, META_NAME)
         if next((info for info in info_list if info.filename == meta_path), None):
             # Already in job layout
-            return zip_data
+            meta_data = in_zip.read(meta_path)
+            meta = json.loads(meta_data)
+            if JobMetaKey.JOB_FOLDER_NAME not in meta:
```
YuanTingHsieh, line 170: not sure if this needs to be `JobMetaKey.JOB_FOLDER_NAME.value`

nvflare/fuel/hci/zip_utils.py
YuanTingHsieh: 2. When meta.json does not exist, this `meta` will be an empty dict so it enters this else statement.
    In L184 you are trying to generate a meta.json.
    I am saying this meta.json you generated here does NOT have `JobMetaKey.JOB_FOLDER_NAME`
    it only has "name", you can see this in L24 of this file
    So you need to change that `META_DATA` string to have `JobMetaKey.JOB_FOLDER_NAME`
    otherwise deploy will fail.

nvflare/fuel/hci/zip_utils.py
```diff
@@ -13,13 +13,17 @@
...
 import os
 from zipfile import ZipFile
 
+from nvflare.apis.job_def import JobMetaKey
+
 META_NAME = "meta.json"
 # A format string for the dummy meta.json
 META_DATA = """{{
   "name": "{}",
+  "job_folder_name": "{}",
```
YuanTingHsieh, line 26: note that these values might change `name, job_folder_name, resource_spec, deploy_map`
we could improve them in another PR to use constants.

chesterxgchen: let me change them now

tests/unit_test/private/fed/server/job_validator_test.py
```diff
@@ -90,7 +93,7 @@ def setup_class(cls):
         cls.validator = JobValidator(cls.fl_ctx)
 
     def test_valid_app(self):
-        self._assert_valid("valid_app")
+          self._assert_valid("valid_app_wo_meta")
```
YuanTingHsieh, line 96: format issue

PR #559, opened 2022-05-18
Fixed the missing client after running the system for a while.

yhwen:


chesterxgchen:
so the issue is due to some network interruption causing heatbeat to fail, you just retry ?

yhwen:
> so the issue is due to some network interruption causing heatbeat to fail, you just retry ?

network issue, or some other exception caused the heartbeat fail, then do a retry.

chesterxgchen (approved):
LGTM

PR #566, opened 2022-05-18
Sort the jobs in scheduler based on submit time

YuanTingHsieh:
- Sort the jobs in scheduler based on submit time

### DIFF
--- a/nvflare/app_common/job_schedulers/job_scheduler.py
+++ b/nvflare/app_common/job_schedulers/job_scheduler.py
@@ -19,7 +19,7 @@ from nvflare.apis.event_type import EventType
 from nvflare.apis.fl_component import FLComponent
 from nvflare.apis.fl_constant import FLContextKey
 from nvflare.apis.fl_context import FLContext
-from nvflare.apis.job_def import Job
+from nvflare.apis.job_def import Job, JobMetaKey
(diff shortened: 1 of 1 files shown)


nvflare/app_common/job_schedulers/job_scheduler.py
```diff
@@ -170,13 +170,18 @@ def handle_event(self, event_type: str, fl_ctx: FLContext):
...
-            self.log_debug(
-                fl_ctx,
-                f"Skipping schedule job because scheduled_jobs ({len(self.scheduled_jobs)}) "
-                f"is greater than max_jobs ({self.max_jobs})",
-            )
-            return None, None
+        with self.lock:
+            self.log_debug(fl_ctx, f"Current scheduled_jobs is {self.scheduled_jobs}")
+            if len(self.scheduled_jobs) >= self.max_jobs:
+                self.log_debug(
```
chesterxgchen, line 176: it might be wrap the L174 to 182 into a function then code would be much easier to read
something like 
     def exceceed_max_jobs(...)  -> Boolean
          exceed_limit = len(self.scheduled_jobs) >= self.max_jobs
          if exceed_limit : 
                  log.debug (...) 
           return exeed_limit

       then in main program
                the code would be
                if exceed_max_job(..) : 
                       return None, None

chesterxgchen (approved):

PR #567, opened 2022-05-18
add check to throw exception if Python version is >= 3.9

nvkevlu:


nvflare/private/fed/app/client/client_train.py
```diff
@@ -34,6 +34,8 @@
 
 
 def main():
+    if sys.version_info >= (3, 9):
+        raise RuntimeError("Python 3.9 and above are not yet supported. Please use Python 3.8 or 3.7.")
```
YuanTingHsieh, line 38: should we also add version should >= 3.7 check?

YuanTingHsieh (approved):
LGTM, we could put these into a helper function like check_python_version or something in a common place and import from there.

PR #571, opened 2022-05-19
runtests scripts refactoring

chesterxgchen:
1) refactoring runtests.sh to stream line different functions
   check license
   check imports, style, format, etc.
   fix import format, style
   run unit tests with option to turn on coverage test and unit test reports

2) give option to only run sub directories of styles, unit tests.

This make it possible to combine all tests (unit tests and integration tests) under the same command, We might re-name the command from runtests to something more general.

Our code base did not pass pylint and pytype checks, I have to disable both of them for now.
pylint usage may need legal approval due to GPL v2 license. I commented out the dependency and related codes.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -6,3 +6,5 @@ black==22.3.0
 click==8.1.2
 pytest-xdist==2.5.0
 pytest-cov==3.0.0
+# wait for approval
+# pylint!=2.13  # https://github.com/PyCQA/pylint/issues/5969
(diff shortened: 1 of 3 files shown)


YuanTingHsieh:
One thing: `./runtest.sh -u` is complaining `./runtest.sh: line 172: ``mypy_check`': not a valid identifier`
is this expected?

and also notice `./runtest.sh -u` is trying to run integration tests as well.

chesterxgchen:
thats not expected, let me check

chesterxgchen:
I believe I address the issues. 
also I updated dry run and clean

YuanTingHsieh (approved):
Thanks for the effort.

PR #572, opened 2022-05-19
fix issue with FLAdminAPI stalling

nvkevlu:
There was an issue where occasionally, the FLAdminAPIRunner will finish init and submit_job can be called before api.server_sess_active = True, causing an error of 'Exception: API session is inactive'

This is due to timing, because FLAdminAPI automatically sends another call to the server to register the commands after a successful login before setting self.server_sess_active = True.

To fix, FLAdminAPIRunner's init will now wait for self.server_sess_active = True.


holgerroth (approved):
LGTM

PR #576, opened 2022-05-19
 improve server error logging 

chesterxgchen:
try to capture more error information when communication failed : receive and send 
It will tell us why the communication failed, or when it fails what it was doing

### DIFF
--- a/nvflare/private/fed/server/server_engine.py
+++ b/nvflare/private/fed/server/server_engine.py
@@ -219,7 +219,7 @@ class ServerEngine(ServerEngineInternalSpec):
         try:
             return conn.recv()
         except BaseException as e:
-            self.logger.error(f"Received unexpected reply from Client for Job: {job_id}, run_number: {run_number}: {e}")
+            self.logger.error(f"Received unexpected reply from child process for Job: {job_id}, run_number: {run_number}: {e}")
             raise e
 
     def _send(self, conn, data, job_id, run_number):
@@ -227,7 +227,7 @@ class ServerEngine(ServerEngineInternalSpec):
             return conn.send(data)
         except BaseException as e:
             self.logger.error(
-                f"Failed to deliver data {data}  to client(s) for Job: {job_id}, run_number: {run_number}: {e}"
+                f"Failed to deliver data {data}  to child process for Job: {job_id}, run_number: {run_number}: {e}"
             )
             raise e
 
--- a/nvflare/private/fed/server/server_engine.py
+++ b/nvflare/private/fed/server/server_engine.py
@@ -215,6 +215,22 @@ class ServerEngine(ServerEngineInternalSpec):
             self.engine_info.status = MachineStatus.STARTED
             return ""
 
+    def _recv(self, conn, job_id, run_number):
+        try:
+            return conn.recv()
+        except BaseException as e:
+            self.logger.error(f"Received unexpected reply from Client for Job: {job_id}, run_number: {run_number}: {e}")
+            raise e
+
+    def _send(self, conn, data, job_id, run_number):
+        try:
+            return conn.send(data)
+        except BaseException as e:
+            self.logger.error(
+                f"Failed to deliver data {data} for Job: {job_id}, run_number: {run_number} to client(s): {e}"
+            )
+            raise e
+
     def _listen_command(self, listen_port, run_number):
         address = ("localhost", int(listen_port))
         listener = Listener(address, authkey="parent process secret password".encode())
@@ -225,13 +241,13 @@ class ServerEngine(ServerEngineInternalSpec):
             job_id = self.run_processes.get(run_number).get(RunProcessKey.JOB_ID)
             try:
                 if conn.poll(0.1):
-                    received_data = conn.recv()
+                    received_data = self._recv(conn, job_id, run_number)
                     command = received_data.get(ServerCommandKey.COMMAND)
                     data = received_data.get(ServerCommandKey.DATA)
 
                     if command == ServerCommandNames.GET_CLIENTS:
                         return_data = {ServerCommandKey.CLIENTS: clients, ServerCommandKey.JOB_ID: job_id}
-                        conn.send(return_data)
+                        self._send(conn, job_id, run_number, return_data)
                     elif command == ServerCommandNames.AUX_SEND:
                         targets = data.get("targets")
                         topic = data.get("topic")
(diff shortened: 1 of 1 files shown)


nvflare/private/fed/server/server_engine.py
```diff
@@ -224,13 +240,13 @@ def _listen_command(self, listen_port, run_number):
...
                 if conn.poll(0.1):
-                    received_data = conn.recv()
+                    received_data = self._recv(conn, job_id, run_number)
                     command = received_data.get(ServerCommandKey.COMMAND)
                     data = received_data.get(ServerCommandKey.DATA)
 
                     if command == ServerCommandNames.GET_CLIENTS:
                         return_data = {ServerCommandKey.CLIENTS: clients, ServerCommandKey.JOB_ID: job_id}
-                        conn.send(return_data)
+                        self._send(conn, job_id, run_number, return_data)
```
YuanTingHsieh, line 249: send signature is `def _send(self, conn, data, job_id, run_number):`

nvflare/private/fed/server/server_engine.py
```diff
@@ -240,9 +256,9 @@ def _listen_command(self, listen_port, run_number):
                         replies = self.aux_send(
                             targets=targets, topic=topic, request=request, timeout=timeout, fl_ctx=fl_ctx
                         )
-                        conn.send(replies)
-            except BaseException:
-                self.logger.warning("Failed to process the child process command.")
+                        self._send(conn, replies)
```
YuanTingHsieh, line 259: your send signature is `def _send(self, conn, data, job_id, run_number):`

YuanTingHsieh (changes requested):
As commented

ZiyueXu77:
Got this error:
```
2022-05-20 00:29:48,362 - root - ERROR - FL server execution exception: 'str' object has no attribute 'get'
  File "/nvflare/nvflare/private/fed/app/server/runner_process.py", line 134, in start_server_app
    server.engine.sync_clients_from_main_process()
  File "/nvflare/nvflare/private/fed/server/server_engine.py", line 564, in sync_clients_from_main_process
    clients = return_data.get(ServerCommandKey.CLIENTS)
AttributeError: 'str' object has no attribute 'get'
2022-05-20 00:29:48,363 - FederatedServer - INFO - Server app stopped.
```

nvflare/private/fed/server/server_engine.py
chesterxgchen: woops. let me change

nvflare/private/fed/server/server_engine.py
chesterxgchen: sorry, forgot I change the signature. thanks for catch it

YuanTingHsieh:
> Got this error:
> 
> ```
> 2022-05-20 00:29:48,362 - root - ERROR - FL server execution exception: 'str' object has no attribute 'get'
> Traceback (most recent call last):
>   File "/nvflare/nvflare/private/fed/app/server/runner_process.py", line 134, in start_server_app
>     server.engine.sync_clients_from_main_process()
>   File "/nvflare/nvflare/private/fed/server/server_engine.py", line 564, in sync_clients_from_main_process
>     clients = return_data.get(ServerCommandKey.CLIENTS)
> AttributeError: 'str' object has no attribute 'get'
> 2022-05-20 00:29:48,363 - FederatedServer - INFO - Server app stopped.
> ```

Can you elaborate? I don't think this related to this PR though?

ZiyueXu77:
> 

I believe it is related to the function name change, after the update it is working now

nvflare/private/fed/server/server_engine.py
```diff
@@ -214,6 +214,22 @@ def start_app_on_server(self, run_number: str, job_id: str = None, job_clients=N
             self.engine_info.status = MachineStatus.STARTED
             return ""
 
+    def _recv(self, conn, job_id, run_number):
+        try:
+            return conn.recv()
+        except BaseException as e:
+            self.logger.error(f"Received unexpected reply from Client for Job: {job_id}, run_number: {run_number}: {e}")
```
IsaacYangSLA, line 221: I think this is confusing.  conn is localhost for IPC.  In server, that's server child process.  @yhwen , can you help clarify this?

IsaacYangSLA (changes requested):

nvflare/private/fed/server/server_engine.py
yhwen: That's correct. This communication is between the parent and client process on the same server. Not between the client and server. When the exception happens, we already logged the message showing the communication and let it try again.

The original issue caused the error is because the child process got terminated. We need to investigate that root caused the child process crash.

chesterxgchen: Ok, I can change the wording, of the message, but we need to log the error at where it happens. so we know what's going on.

chesterxgchen: I update the message, please check again @IsaacYangSLA @yhwen

nvflare/private/fed/server/server_engine.py
```diff
@@ -215,6 +215,22 @@ def start_app_on_server(self, run_number: str, job_id: str = None, job_clients=N
             self.engine_info.status = MachineStatus.STARTED
             return ""
 
+    def _recv(self, conn, job_id, run_number):
```
IsaacYangSLA, line 218: I think the main purpose of job_id and run_number arguments are for the log msg.  The underlying recv function does not need job_id and run_number.
Maybe using a few try/except inside _listen_command is better.  It's just my personal opinion.

chesterxgchen: that's correct, So you prefer not wrap this into a try/except clause, but how do you distinguishing different errors as the main _listen_command has several _conn.recv_ conn.send and conn.aux_send methods
for example, conn.aux_send also do the similar try/expcet/raise to capture local error message.  

If might hurt the performance bit, but comparing to the training/transfer data, this should be nothing, I thought it would give a much better context of the error. 

That being said, I am willing to cancel PR if you all felt this is not necessary @YuanTingHsieh @yhwen @nvidianz

YuanTingHsieh: I am fine with this PR.

I think the server engine code needs to be cleaned up in 2.2 so maybe these will be changed at that time.

chesterxgchen: If don't make this PR, we have to do another PR to capture the exception cause, currently the exception is swallowed. did not show the real cause of the exception

YuanTingHsieh:
Can you do a rebase and fix the unit test issue thanks!

chesterxgchen:
@YuanTingHsieh  thanks, I was in a hurry, did not prepare the unit tests. I will do that first, and do the rebase

chesterxgchen:
Actually, its a bit tricky to mock the socket connection failure. Let me close this ticket and just add another PR to address the exception swallow. I might want to tackle this later

PR #579, opened 2022-05-20
added 30 seconds delay to terminate the child process.

yhwen:


nvflare/private/fed/server/server_engine.py
```diff
@@ -176,6 +176,8 @@ def heartbeat_to_parent(self):
             except BaseException:
                 # The parent process can not be reached. Terminate the child process.
```
chesterxgchen, line 177: should you capture the exception here for logging purpose, so one know that's happening caused the exeception

except BaseException as e: 
    logging.error ( ' ... ')
    raise e;

nvflare/private/fed/server/server_engine.py
```diff
@@ -176,6 +176,8 @@ def heartbeat_to_parent(self):
...
                 break
+        # delay some time for the wrap up process before the child process self terminate.
```
chesterxgchen, line 179: why 30 seconds ? why not 15 or 46 seconds ?

nvflare/private/fed/server/server_engine.py
yhwen: No, this exception is expected. It is used to indicate the corresponding job on the parent side already terminate. It's not an error case.

nvflare/private/fed/server/server_engine.py
yhwen: We can't have delay too long. Otherwise, it may cause a confusion that a ghost child process may still be running when the job is done. If too short, it may result to a controller not be able to finish the abort wrap up process. I think 30 seconds for that is reasonable.

IsaacYangSLA (approved):
30 sec is pretty safe.  I think it gives enough time for child process to clean up all pending operations.

PR #580, opened 2022-05-20
Fix handling of config exception during job running

YuanTingHsieh:
Fix handling of the following exception scenario:

- server has started the app, but the app code requires library that server did not install so execution exception happens
  => set job run status to `FINISHED: EXECUTION_EXCEPTION`
- client has started the app, but the app code requires library that client did not install so execution exception happens
  => job would stay RUNNING status since server side success
        since client side has started but execution exception happens, so the resource is freed.

### DIFF
--- a/nvflare/private/fed/app/client/worker_process.py
+++ b/nvflare/private/fed/app/client/worker_process.py
@@ -72,12 +72,11 @@ def main():
 
     print("starting the client .....")
 
-    deployer = None
-    command_agent = None
-
     startup = os.path.join(args.workspace, "startup")
     SecurityContentService.initialize(content_folder=startup)
 
+    deployer = None
+    command_agent = None
     federated_client = None
     try:
         token_file = os.path.join(args.workspace, EngineConstant.CLIENT_TOKEN_FILE)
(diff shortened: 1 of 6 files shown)


nvflare/private/fed/app/client/worker_process.py
```diff
@@ -173,19 +168,15 @@ def main():
 
     except BaseException as e:
         traceback.print_exc()
-        print("FL client execution exception: " + str(e))
+        print(f"FL client execution exception: {e}")
```
chesterxgchen, line 171: should we use logger ?

chesterxgchen (approved):

nvflare/private/fed/server/job_runner.py
```diff
@@ -215,7 +215,16 @@ def _job_complete_process(self, fl_ctx: FLContext):
...
+                            if run_number in engine.execution_exception_run_processes:
+                                run_process = engine.execution_exception_run_processes[run_number]
+                                # stop client run
+                                client_sites = run_process.get(RunProcessKey.PARTICIPANTS)
+
+                                replies = self.abort_client_run(engine, run_number, client_sites, fl_ctx)
+                                if not replies:
+                                    self.log_error(
+                                        fl_ctx, f"Failed to send abort command to clients for run: {run_number}"
+                                    )
```
yhwen, line 227: Why do we need this? When the ServerRunner end the job running, it already sent the ReservedTopic.END_RUN to all the running clients to end the client run. We should not need this any more.

yhwen: Checked with Yuan-ting on this. This logic is to handle the case the server app did't actually start due to the config error case. We could not use the event handling to notify client to abort run yet. The handle logic is correct.

yhwen (approved):

nvflare/private/fed/app/client/worker_process.py
```diff
@@ -173,19 +168,15 @@ def main():
...
+        raise e
```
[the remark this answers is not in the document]
YuanTingHsieh: by raising this exception, the return code of this process will not be 0.

PR #581, opened 2022-05-20
add tests for running through FLAdminAPI commands

nvkevlu:
Adds a test to execute a chain of FLAdminAPI commands. This adds 30s to the tests and covers most of the FLAdminAPI commands.

Fixes some issues with commands in FLAdminAPI.

Testing clone_job is a bit tricky because the cloned job will be picked up to run and the abort right now does not stop it properly if it is in the SUBMITTED state even with some modifications to job_cmds.py (simply setting the status to FINISHED:ABORTED does not work because it still gets picked up and then the status becomes RUNNING again, but this is another issue from these tests).

### DIFF
--- a/tests/integration_test/admin_controller.py
+++ b/tests/integration_test/admin_controller.py
@@ -25,6 +25,7 @@ from nvflare.fuel.hci.client.fl_admin_api_constants import FLDetailKey
 from nvflare.fuel.hci.client.fl_admin_api_spec import TargetType
 from nvflare.ha.dummy_overseer_agent import DummyOverseerAgent
 from nvflare.ha.overseer_agent import HttpOverseerAgent
+from tests.integration_test.test_fladminapi import run_admin_api_tests
 
 
 def process_logs(logs, run_state):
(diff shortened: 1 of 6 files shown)


tests/integration_test/admin_controller.py
```diff
@@ -450,6 +453,71 @@ def execute_actions(self, site_launcher, actions):
...
+                    print(self.admin_api.get_active_sp().get("details"))
+                    print("\nList SP:")
+                    print(self.admin_api.list_sp().get("details"))
+                    print("\nCommand: get_available_apps_to_upload")
+                    print(self.admin_api.get_available_apps_to_upload())
+                    print("\nList Jobs:")
+                    list_jobs_return_message = self.admin_api.list_jobs().get("details").get("message")
+                    print(list_jobs_return_message)
+                    first_job = list_jobs_return_message.split()[14]
+                    print("\nCommand: ls server -a .")
```
YuanTingHsieh, line 475: can these prints be changed to assert so it will fail automatically if something is not expected?

tests/integration_test/admin_controller.py
```diff
@@ -450,6 +453,71 @@ def execute_actions(self, site_launcher, actions):
                         client_ids = list(site_launcher.client_properties.keys())
                     for cid in client_ids:
                         site_launcher.start_client(cid)
+
+            elif command == "test":
+                if args[0] == "admin_commands":
+                    print(("\n" + "*" * 120) * 20)
```
YuanTingHsieh, line 459: another suggestion can you create a function something like

`admin_api_test(admin_api)`

And move these logic inside the function, and we just pass self.admin_api as argument

nvkevlu: done

YuanTingHsieh (approved):
LGTM, some enhancements can be made to replace those "print" with asserts.

tests/integration_test/test_fladminapi.py
```diff
@@ -0,0 +1,81 @@
...
+    print(("\n" + "*" * 120) * 20)
+    print("\n" + "=" * 40)
+    print("\nRunning through tests of admin commands:")
+    print("\n" + "=" * 40)
+    print("\nCommand: set_timeout")
+    print(admin_api.set_timeout(11).get("details").get("message"))
+    print("\nActive SP:")
+    print(admin_api.get_active_sp().get("details"))
+    print("\nList SP:")
+    print(admin_api.list_sp().get("details"))
```
YuanTingHsieh, line 27: Like these prints can we find a way to replace it with assert?

PR #590, opened 2022-05-21
hide download commands with issues that will be worked on in next rel…

nvkevlu:
hide download commands with issues that will be worked on in next release


IsaacYangSLA (approved):
LGTM.

PR #592, opened 2022-05-23
Fixed the snapshot log path info.

yhwen:


nvflare/private/fed/server/server_engine.py
```diff
@@ -652,7 +652,10 @@ def persist_components(self, fl_ctx: FLContext, completed: bool):
 
             self.server.snapshot_location = self.snapshot_persistor.save(snapshot=snapshot)
         finally:
-            self.logger.info(f"persist the snapshot to: {self.server.snapshot_location}")
+            if not completed:
+                self.logger.info(f"persist the snapshot to: {self.server.snapshot_location}")
+            else:
+                self.logger.info(f"The snapshot: {self.server.snapshot_location} has been removed.")
```
YuanTingHsieh, line 658: `self.snapshot_persistor.save(snapshot=snapshot)` this is not guaranteed to finish successfully in the try block.

So the print statement in this finally block is not stating the fact.

I think we need to move the print inside the try block.

YuanTingHsieh (changes requested):
Thanks for the effort, one issue as commented.

nvflare/private/fed/server/server_engine.py
yhwen: done.

PR #593, opened 2022-05-23
update provisioning ui to generate updated project.yml 

nvkevlu:
after args and uri_root changes, update generated project.yml to be consistent


IsaacYangSLA (approved):
Thanks for the update.

PR #599, opened 2022-05-25
Support @ALL in site lists of deploy_map

nvidianz:


nvflare/private/fed/server/job_meta_validator.py
```diff
@@ -0,0 +1,201 @@
...
+
+        if not site_list:
+            raise ValueError(f"No site is specified in deploy_map for job {job_name}")
+
+        if ALL_SITES.casefold() in (site.casefold() for site in site_list):
+            # if @ALL is specified, no other site can be in the list
+            if len(site_list) > 1:
+                raise ValueError(f"No other site can be specified if @ALL is used for job {job_name}")
+            else:
+                site_list = [ALL_SITES]
```
chesterxgchen, line 100: else is unnecessary, as ALL_SITES is already in the site_list

nvflare/private/fed/server/job_meta_validator.py
```diff
@@ -0,0 +1,201 @@
...
+            if duplicates:
+                raise ValueError(f"Multiple apps to be deployed to following sites {duplicates} for job {job_name}")
+
+        return site_list
+
+    def _validate_app(self, job_name: str, meta: dict, zip_file: ZipFile) -> None:
+
+        deploy_map = meta.get(JobMetaKey.DEPLOY_MAP.value)
+
+        for app, deployment in deploy_map.items():
```
chesterxgchen, line 112: should we use deployments: plural to indicate it is not single site, but multiple sites

chesterxgchen (approved):
minor comments, change if you like

nvflare/private/fed/server/job_meta_validator.py
nvidianz: Good idea.

nvflare/private/fed/server/job_meta_validator.py
nvidianz: The purpose of this line is to use canonical form (all caps) in site_list so comparison in other code is easier.

nvkevlu (approved):

yhwen (approved):

PR #600, opened 2022-05-25
add function to clean and install pip deps

### LINKED ISSUE #578: [CICD] Independent PIP env for multiple tests
follow-up of https://github.com/NVIDIA/NVFlare/pull/527/files

It would be nice to have a ENV context for each tests with in the same script.

So we can avoid potential conflicts.

YanxuanLiu:
fix #578 

1. maintain pip requirements with pipenv (already installed in docker image)
2. add args to shell scripts to enable running partial tests


pxLi (approved):
LGTM as a good start point. Please fill the PR desc well

ci/run_premerge_gpu.sh
```diff
@@ -15,15 +15,57 @@
...
+integration_test() {
+    echo "Run integration test..."
+    init_pipenv requirements-dev.txt
+    testFolder="tests/integration_test"
+    rm -rf /tmp/snapshot-storage
+    pushd ${testFolder}
+    pipenv run ./run_integration_tests.sh
+    popd
+    rm -rf /tmp/snapshot-storage
+}
```
pxLi, line 53: please also add remove_pipenv as a good context patten to follow even there is only one test in this file for now

pxLi (changes requested):
some small nit

ci/run_premerge_gpu.sh
YanxuanLiu: added remove_pipenv in the end

PR #604, opened 2022-05-25
Fix resource manager

YuanTingHsieh:
- Add `fire_event` method into `ClientEngine`
- Add firing `EventType.SYSTEM_START` and `EventType.SYSTEM_END`
- Get rid of `fl_ctx` in method signature, since `free_resources` using a dummy FLContext and that is not meaningful


yhwen (changes requested):
Discussed with @YuanTingHsieh. Better still keep to use the event firing approach. We will add the SYSTEM_START and SYSTEM_END event on the client,

YuanTingHsieh:
will do a rebase once https://github.com/NVIDIA/NVFlare/pull/607 is merged

nvflare/apis/resource_manager_spec.py
```diff
@@ -26,12 +24,11 @@ def consume(self, resources: dict):
 
 class ResourceManagerSpec(ABC):
     @abstractmethod
-    def check_resources(self, resource_requirement: dict, fl_ctx: FLContext) -> (bool, Optional[str]):
+    def check_resources(self, resource_requirement: dict) -> (bool, Optional[str]):
```
yhwen, line 27: It's still better to add fl_ctx: FLContext back into the API. Because we are providing an API spec for people to write their own ResourceManager implementation. We are not using the fl_ctx in own standard implementation. But other people could make use of this for their own purpose. By providing the fl_ctx: FLContext in the API signature, it provides a lot more flexibility.

nvflare/apis/resource_manager_spec.py
```diff
@@ -42,21 +39,20 @@ def check_resources(self, resource_requirement: dict, fl_ctx: FLContext) -> (boo
         pass
 
     @abstractmethod
-    def cancel_resources(self, resource_requirement: dict, token: str, fl_ctx: FLContext):
+    def cancel_resources(self, resource_requirement: dict, token: str):
```
yhwen, line 42: same fl_ctx.

nvflare/apis/resource_manager_spec.py
```diff
@@ -42,21 +39,20 @@ def check_resources(self, resource_requirement: dict, fl_ctx: FLContext) -> (boo
...
-            fl_ctx: the FLContext
 
         Note:
             If check_resource didn't return a token, then don't need to call this method
         """
         pass
 
     @abstractmethod
-    def allocate_resources(self, resource_requirement: dict, token: str, fl_ctx: FLContext) -> dict:
+    def allocate_resources(self, resource_requirement: dict, token: str) -> dict:
```
yhwen, line 55: same fl_ctx

nvflare/apis/resource_manager_spec.py
```diff
@@ -65,20 +61,18 @@ def allocate_resources(self, resource_requirement: dict, token: str, fl_ctx: FLC
...
-            fl_ctx: the FLContext
 
         Returns:
             A dict of allocated resources
         """
         pass
 
     @abstractmethod
-    def free_resources(self, resources: dict, token: str, fl_ctx: FLContext):
+    def free_resources(self, resources: dict, token: str):
```
yhwen, line 71: same fl_ctx

nvflare/private/fed/client/client_executor.py
```diff
@@ -360,16 +359,16 @@ def _wait_child_process_finish(self, client, run_number, allocated_resource, tok
...
             child_process = self.run_processes.get(run_number, {}).get(RunProcessKey.CHILD_PROCESS)
         if child_process:
             child_process.wait()
             return_code = child_process.returncode
-            self.logger.info(f"process finished with execution code: {return_code}")
+            self.logger.info(f"run ({run_number}): child worker process finished with execution code: {return_code}")
 
         if allocated_resource:
-            resource_manager.free_resources(resources=allocated_resource, token=token, fl_ctx=FLContext())
+            resource_manager.free_resources(resources=allocated_resource, token=token)
```
yhwen, line 371: Here, instead of using:
 fl_ctx=FLContext()

we can change it to:
fl_ctx = client.engine.new_context()

yhwen (commented):
Overall the PR looks good. Just add the fl_ctx back into the API signature, then we make the APIs more flexible to use.

yhwen (approved):

PR #611, opened 2022-05-26
Added proper system shutdown to avoid random Bad file descriptor errors.

yhwen:
Add proper actions to close the ThreadPool when shutdown the system. This is to avoid the random bad file descriptor errors.

### DIFF
--- a/nvflare/private/fed/client/admin_msg_sender.py
+++ b/nvflare/private/fed/client/admin_msg_sender.py
@@ -169,3 +169,6 @@ class AdminMessageSender(Sender):
         else:
             channel = grpc.insecure_channel(**channel_dict)
         return channel
+
+    def close(self):
+        self.pool.close()
--- a/nvflare/private/fed/server/fed_server.py
+++ b/nvflare/private/fed/server/fed_server.py
@@ -215,7 +215,8 @@ class BaseServer(ABC):
     def fl_shutdown(self):
         self.shutdown = True
         self.close()
-        self.executor.shutdown()
+        if self.executor:
+            self.executor.shutdown()
 
 
 class FederatedServer(BaseServer, fed_service.FederatedTrainingServicer, admin_service.AdminCommunicatingServicer):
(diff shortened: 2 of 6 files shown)


nvflare/private/fed/client/admin_msg_sender.py
```diff
@@ -169,3 +169,6 @@ def _set_up_channel(self, channel_dict):
         else:
             channel = grpc.insecure_channel(**channel_dict)
         return channel
+
+    def close(self):
+        self.pool.close()
```
chesterxgchen, line 174: should you do 
    pool.close()
    pool.join()
https://stackoverflow.com/questions/38271547/when-should-we-call-multiprocessing-pool-join

nvflare/private/fed/server/server_engine.py
```diff
@@ -779,7 +783,9 @@ def server_shutdown(server, touch_file):
     try:
         server.fl_shutdown()
         server.admin_server.stop()
+        time.sleep(3.0)
     finally:
+        server.status = ServerStatus.SHUTDOWN
         sys.exit(2)
```
YuanTingHsieh, line 789: if we call this sys.exit, will the process just end?

nvflare/private/fed/server/server_engine.py
```diff
@@ -391,6 +393,8 @@ def restart_server(self) -> str:
 
         touch_file = os.path.join(self.args.workspace, "restart.fl")
         _ = self.executor.submit(lambda p: server_shutdown(*p), [self.server, touch_file])
+        while self.server.status != ServerStatus.SHUTDOWN:
+            time.sleep(1.0)
         return ""
```
YuanTingHsieh, line 398: so we will not get to return here since we call sys.exit in `server_shutdown`?

nvflare/private/fed/server/fed_server.py
```diff
@@ -213,6 +215,7 @@ def remove_dead_clients(self):
     def fl_shutdown(self):
         self.shutdown = True
         self.close()
+        self.executor.shutdown()
```
YuanTingHsieh, line 218: should have `if self.executor`

YuanTingHsieh (changes requested):
Thanks for the effort. Minor stuffs as commented

yhwen:
The sys.exit() only cause the thread to end, not the whole process.

It will still return. Because server_shutdown only cause the other thread to end.

PR #612, opened 2022-05-27
fix shutdown_system command

nvkevlu:
api here is not FLAdminAPI but AdminAPI so check_status is not available and do_command needs to be used instead.

Fixes this, and improves exception message to be more clear of what the error is for.


nvkevlu:
Not really sure what is going on here, I would not expect the changes in this PR to impact the tests at all, yet something is failing.

PR #616, opened 2022-05-27
add HA integration tests with 2 servers

nvkevlu:


tests/integration_test/data/single_app_as_job/test_ha_2_servers_after_first_round.yml
```diff
@@ -0,0 +1,14 @@
...
+cleanup: True
+
+
+tests:
+  - app_name: np_sag
+    validators:
+      - path: tests.integration_test.validators.LogResultValidator
+        args: {
+          expected_in_result: "Round 1 started",
+        }
```
YuanTingHsieh, line 13: I think it is better to also add the following into validators to make sure it does restart and it does finish
```
      - path: tests.integration_test.validators.NumpySAGResultValidator
        args: { expected_result: [ [ 4, 5, 6 ], [ 7, 8, 9 ], [ 10, 11, 12 ] ] }
```

tests/integration_test/data/single_app_as_job/test_ha_2_servers_before_first_round.yml
```diff
@@ -0,0 +1,12 @@
...
+apps_root_dir: ./data/apps
+single_app_as_job: True
+cleanup: True
+
+
+tests:
+  - app_name: np_sag
+    validators:
+      - path: tests.integration_test.validators.LogResultValidator
+        args: { expected_in_result: "Round 0 started" }
```
YuanTingHsieh, line 11: I think it is better to also add the following into validators to make sure it does restart and it does finish

      - path: tests.integration_test.validators.NumpySAGResultValidator
        args: { expected_result: [ [ 4, 5, 6 ], [ 7, 8, 9 ], [ 10, 11, 12 ] ] }

tests/integration_test/data/single_app_as_job/test_ha_2_servers_after_first_round.yml
```diff
@@ -0,0 +1,14 @@
...
+single_app_as_job: True
+cleanup: True
+
+
+tests:
+  - app_name: np_sag
+    validators:
+      - path: tests.integration_test.validators.LogResultValidator
+        args: {
+          expected_in_result: "Round 1 started",
```
YuanTingHsieh, line 12: should be `Round 2 started`? since we abort the job after round 1

YuanTingHsieh (changes requested):
Thanks for the effort, I leave some comments.

tests/integration_test/data/single_app_as_job/test_ha_2_servers_after_first_round.yml
nvkevlu: The first round is actually round 0, so it aborts after round 0 finishes. I was trying to do more with validations, but in practice, it doesn't really matter because the logs seem to be persisted across servers for the job now, and it is hard to do more advanced validation without finding timing of events and comparing that.

YuanTingHsieh (approved):
Thanks, LGTM

PR #623, opened 2022-05-28
Fix abort job when server execution exception

YuanTingHsieh:
- Fix abort job when server execution exception


chesterxgchen (approved):
LGTM

PR #624, opened 2022-05-29
fixed jobs run race condition issue.

yhwen:
add description for Yuhong to explain why this solve the race condition: 
there are two race condition points
1) use token_file when create child process
2)  start_run

### DIFF
--- a/nvflare/private/fed/app/client/worker_process.py
+++ b/nvflare/private/fed/app/client/worker_process.py
@@ -15,6 +15,7 @@
 """Provides a command line interface for a federated client trainer."""
 
 import argparse
+import logging
 import os
 import sys
 import threading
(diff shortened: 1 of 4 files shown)


nvflare/private/fed/app/client/worker_process.py
```diff
@@ -53,6 +54,12 @@ def main():
     parser = argparse.ArgumentParser()
     parser.add_argument("--workspace", "-m", type=str, help="WORKSPACE folder", required=True)
     parser.add_argument("--startup", "-w", type=str, help="startup folder", required=True)
+    parser.add_argument("--token", "-t", type=str, help="token", required=True)
```
chesterxgchen, line 57: why we adding additional argument, is this needed for fix racing condition. We are in code freeze, i.e. only changes needed to relate to critical fixes

chesterxgchen (changes requested):
you should split the PR into two 1) one for argument parsing improvement 2) race condition fix
We are in code freeze: i.e. unless it is critical bugs, we have to wait. 

Lets keep the code freeze means code freeze.

nvflare/private/fed/app/client/worker_process.py
chesterxgchen: thank Yuhong for explaining to me that the change is actually to replace token_file, which is also a race condition (share processed can access the same token file)

chesterxgchen (approved):

PR #630, opened 2022-06-01
Fix a situation where server start success but client start failed

YuanTingHsieh:
When this specific situation happens:
     server and client deploy success, and  server start success but client start failed
     
 The `check_status server` will show APP NAME as "?".
 
 ```
 
 > check_status server
Engine status: started
-------------------------------------------------------------------
| RUN_NUMBER                           | APP NAME                 |
-------------------------------------------------------------------
| 903172d7-655c-4180-b5e4-72d3ab1ca13c | cifar10_fedavg_he        |
| 5d94d8a6-2152-4169-900e-114030415d70 | cifar10_fedavg_he        |
| 247c9d96-2acf-4dc3-90c1-c429df69fca2 | cifar10_fedavg_stream_tb |
| 178cee01-4423-48a9-bde0-6f006bdd0ddc | cifar10_fedavg           |
| c92b5e7a-46c0-4dea-a1c2-70679ceb3576 | cifar10_fedprox          |
| d1952fda-dd66-4553-8e3f-b9b2378367b5 | cifar10_fedopt           |
| 597b7ae5-b065-41d7-8bec-07ea888027ec | cifar10_fedavg_stream_tb |
| 5bbd3cd7-f398-41b7-af82-d8459eb3dd3b | ?                        |
-------------------------------------------------------------------

 ```
 
 While `list_job` show
 
 ```
 
 | 5bbd3cd7-f398-41b7-af82-d8459eb3dd3b | cifar10_fedavg           | FAILED_TO_RUN      | 2022-05-31T16:16:09.636671+00:00 |
 
 ```


nvflare/private/fed/server/job_runner.py
```diff
@@ -259,9 +257,8 @@ def run(self, fl_ctx: FLContext):
                                 self.running_jobs[run_number] = ready_job
                                 job_manager.set_status(ready_job.job_id, RunStatus.RUNNING, fl_ctx)
                             except Exception as e:
-                                run_number = fl_ctx.get_prop(FLContextKey.JOB_RUN_NUMBER)
                                 if run_number:
-                                    self._delete_run(run_number, list(client_sites.keys()), fl_ctx)
+                                    self._stop_run(run_number, fl_ctx)
```
yhwen, line 261: self._stop_run() will update the job status to RunStatus.FINISHED_ABORTED first, then it will update it to RunStatus.FAILED_TO_RUN. Better change it to update only once.

YuanTingHsieh: thanks for the review,

`self._stop_run` will NOT update the job status.
It is `self.stop_run` that updates the job status to `RunStatus.FINISHED_ABORTED`

YuanTingHsieh: if we want to use `self.stop_run`, when should I update the job status?

yhwen: better to use self.stop_run(), because it takes care of the  self.running_jobs scenario. just no need to do two times of the status update.

YuanTingHsieh: then maybe I should add a line after this function to fire JOB_ABORTED event?

yhwen: The issue for using self._stop_run is in case the job added into self.running_jobs, then it will become a ghost never got removed.

PR #635, opened 2022-06-01
make additions in docs to address various issues

nvkevlu:
Add to docs:

- mention that secure content cannot be altered
- add more on run_number -> job_id
- add item about IP change for FL server issue to FAQ
- abort_task usage

### DIFF
--- a/docs/examples/hello_pt.rst
+++ b/docs/examples/hello_pt.rst
@@ -240,7 +240,7 @@ run the following commands in the fl_admin to shutdown the system (while inputti
 
 In order to stop all processes, run ``./stop_fl.sh``.
 
-All artifacts from the FL run can be found in the server run folder you created with ``set_run_number``.  In this exercise, the folder is ``run_1``.
+All artifacts from the FL run can be found in the server folder for the job id.
 
 Congratulations!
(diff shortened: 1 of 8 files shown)


docs/examples/hello_pt.rst
```diff
@@ -240,7 +240,7 @@ run the following commands in the fl_admin to shutdown the system (while inputti
 
 In order to stop all processes, run ``./stop_fl.sh``.
 
-All artifacts from the FL run can be found in the server run folder you created with ``set_run_number``.  In this exercise, the folder is ``run_1``.
+All artifacts from the FL run can be found in the server folder for the job id.
```
YuanTingHsieh, line 243: It can be found in the server folder using the "run_number",
it just happens that our run number right now is job id,
do we need to explain this in details?

If not this is good to me.

YuanTingHsieh: @kkersten what do you think?

nvkevlu: I think this is probably okay for now since it is better to keep things simple in this example. We can revise mentions of run_number / job_id once we do have a distinction between these in the future.

chesterxgchen (approved):

PR #637, opened 2022-06-02
Enhanced the child_process logging.

yhwen:


nvflare/private/fed/app/client/worker_process.py
```diff
@@ -105,37 +104,42 @@ def main():
     deployer = None
     command_agent = None
     federated_client = None
+
+    startup = args.startup
```
chesterxgchen, line 108: L108 to L122 should have wrap into a function, logging_setup or something.

yhwen: done

chesterxgchen (approved):

PR #639, opened 2022-06-03
Add missing import

YuanTingHsieh:
- Missing import Optional in this file


nvidianz (approved):
Maybe we should add pylint to the runtest.sh

yhwen (approved):
That function didn't have usage.

PR #645, opened 2022-06-06
Fix issue #638 by removing DistributionBuilder

### LINKED ISSUE #638: Remove external 'zip' dependency from DistributionBuilder (zip_password=True)
Hi there, first off I want to acknowledge your efforts towards making an incredible tool, thank you!

The issue: I faced the exception *FileNotFoundError: [Errno 2] No such file or directory: 'zip'* when trying to specify `zip_password: true` for the DistributionBuilder args while running `provision -p project.yml`, and tracked the cause to `zip` not being installed on the local machine, which was a problem since installing additional software is not allowed on that machine.

## Replicate
```bash
# Launch a docker container with image python:3.8 running interactive bash
docker run -it --rm python:3.8 bash

# === Within docker container ===
pip install nvflare
provision # create example project.yml

# Set zip_password to true
sed -i 's/zip_password: false/zip_password: true/' project.yml

provision -p project.yml
```
We will find the following output:
```
Project yaml file: /project.yml.
Exception raised during provision.  Incomplete prod_n folder removed.
  File "/usr/local/lib/python3.8/site-packages/nvflare/lighter/spec.py", line 166, in provision
    b.build(study, ctx)
  File "/usr/local/lib/python3.8/site-packages/nvflare/lighter/impl/workspace.py", line 109, in build
    subprocess.run(run_args)
  File "/usr/local/lib/python3.8/subprocess.py", line 493, in run
    with Popen(*popenargs, **kwargs) as process:
  File "/usr/local/lib/python3.8/subprocess.py", line 858, in __init__
    self._execute_child(args, executable, preexec_fn, close_fds,
  File "/usr/local/lib/python3.8/subprocess.py", line 1704, in _execute_child
    raise child_exception_type(errno_num, err_msg, err_filename)
FileNotFoundError: [Errno 2] No such file or directory: 'zip'
```
Whereas it'll work if we install zip
```bash
# === Within docker container ===
apt-get update && apt-get install -y zip
provision -p project.yml
```

## Fix
The build function of the DistributionBuilder uses `shutil` to make the non-password protected zip but relies on `subprocess.run(['zip', ...])` in order to create password-protected files, since shutil doesn't support passwords.

I'd propose adding a package requirement of something like `pyminizip` or `pyzipper` to handle password protection.

https://github.com/NVIDIA/NVFlare/blob/c8d51bfb534c02eff874a760c47086c87cbff59f/nvflare/lighter/impl/workspace.py#L100-L113

IsaacYangSLA:
Fix #638

### DIFF
--- a/nvflare/lighter/impl/workspace.py
+++ b/nvflare/lighter/impl/workspace.py
@@ -15,10 +15,8 @@
 import os
 import pathlib
 import shutil
-import subprocess
 
 from nvflare.lighter.spec import Builder, Project
-from nvflare.lighter.utils import generate_password
 
 
 class WorkspaceBuilder(Builder):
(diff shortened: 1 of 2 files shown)


nvkevlu:
What if we take it out of the default project.yml but leave the code itself with a note in the docstring maybe that zip is required in the environment? That way, there is some sort of reference that makes it easier for people who do want to implement a way to package and distribute the startup kits.

nvkevlu (approved):
Thanks, LGTM

PR #646, opened 2022-06-06
Remove fl_ctx in init method of job meta validator

YuanTingHsieh:
- Remove fl_ctx in init method of job meta validator


nvidianz (approved):
LGTM

PR #650, opened 2022-06-07
Show better messages to users when starting overseer in POC mode

IsaacYangSLA:
Start overseer in debug mode in POC


IsaacYangSLA:
app.config is set by flask.  The value of "DEBUG" key is set to False as default.  When FLASK_ENV is set to development, value of DEBUG key is set to True. https://flask.palletsprojects.com/en/2.1.x/config/#DEBUG

PR #653, opened 2022-06-08
Added support for name in meta.json

nvidianz:


YuanTingHsieh:
can you also update this line to use JOB_NAME?

https://github.com/NVIDIA/NVFlare/blob/dev-2.1/nvflare/private/fed/server/cmd_utils.py#L155

YuanTingHsieh (approved):
thanks LGTM

PR #654, opened 2022-06-08
Replace run_number with job_id in most occurrences

nvkevlu:
Replaces run_number and strings with job_id. Areas that may still need attention:

nvflare/private/fed/server/job_runner.py:
- uses FLContextKey.JOB_RUN_NUMBER, appends with logic of run_number = job.job_id + ":" + str(count)

nvflare/private/fed/server/server_engine.py: these methods have both run_number and job_id
- start_app_on_server
- _start_runner_process

    
nvflare/private/fed/server/fed_server.py:
- _turn_to_hot
    
nvflare/apis/fl_constant.py:
- ReservedKey.RUN_NUM
    
nvflare/app_common/widgets/event_recorder.py:
- has "__run_num__"
    
nvflare/private/fed/protos/federated_pb2.py

### DIFF
--- a/docs/programming_guide/event_system.rst
+++ b/docs/programming_guide/event_system.rst
@@ -90,8 +90,8 @@ NVIDIA FLARE's system-defined event types are specified in :class:`nvflare.apis.
 .. csv-table::
    :header: Event, Description, Data Key, Data Type, Server, Client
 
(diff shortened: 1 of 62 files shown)


yhwen (approved):
The changes look good. The change scope is big, need thorough testing.

YuanTingHsieh:
I found the root cause of the failure, few places needs to change as well.

Let's just merge this.

PR #658, opened 2022-06-08
Remove overseer from poc

IsaacYangSLA:
Add message about POC limitation
Remove HA tests as they require poc supports


nvkevlu (approved):
Thanks, LGTM. Looks like we will need to do something about the tests for HA.

PR #662, opened 2022-06-14
Added back delete_job command, removed the delete_workspace command.

yhwen:
- Added back "delete_job" to delete the job and workspace.
- Removed the "delete_run" command.

### DIFF
--- a/nvflare/fuel/hci/client/fl_admin_api.py
+++ b/nvflare/fuel/hci/client/fl_admin_api.py
@@ -334,7 +334,7 @@ class FLAdminAPI(AdminAPI, FLAdminAPISpec):
         if not isinstance(job_id, str):
             raise APISyntaxError("job_id must be str but got {}.".format(type(job_id)))
         success, reply_data_full_response, reply = self._get_processed_cmd_reply_data(
-            AdminCommandNames.DELETE_RUN + " " + str(job_id)
+            AdminCommandNames.DELETE_WORKSPACE + " " + str(job_id)
         )
         if reply_data_full_response:
             if "can not be deleted" in reply_data_full_response:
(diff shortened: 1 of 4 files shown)


nvflare/fuel/hci/client/fl_admin_api.py
```diff
@@ -334,7 +334,7 @@ def delete_run(self, job_id: str) -> FLAdminAPIResponse:
         if not isinstance(job_id, str):
             raise APISyntaxError("job_id must be str but got {}.".format(type(job_id)))
         success, reply_data_full_response, reply = self._get_processed_cmd_reply_data(
-            AdminCommandNames.DELETE_RUN + " " + str(job_id)
+            AdminCommandNames.DELETE_WORKSPACE + " " + str(job_id)
         )
```
YuanTingHsieh, line 338: maybe instead of `delete_run`, we should have `delete_job` method in this FLAdminAPI?

@nvkevlu what do you think?

But that change can come in different PR.

nvkevlu: yes, definitely, I will have that in my PR.

nvflare/private/fed/server/training_cmds.py
```diff
@@ -45,12 +45,12 @@ def get_spec(self):
...
-                    name=AdminCommandNames.DELETE_RUN,
-                    description="delete a run",
-                    usage="delete_run job_id",
+                    name=AdminCommandNames.DELETE_WORKSPACE,
+                    description="delete the workspace of a job",
+                    usage="delete_workspace job_id",
                     handler_func=self.delete_job_id,
                     authz_func=self.authorize_set_job_id,
-                    visible=True,
+                    visible=False,
```
nvkevlu, line 53: why do we set visible=False here?

PR #665, opened 2022-06-14
make commands consistent with updates for FLAdminAPI and docs

nvkevlu:
Since delete_run is now delete_job and download commands have been replaced with download_job, this makes the FLAdminAPI and documentation consistent with the changes.

### DIFF
--- a/nvflare/fuel/hci/client/fl_admin_api.py
+++ b/nvflare/fuel/hci/client/fl_admin_api.py
@@ -330,22 +330,6 @@ class FLAdminAPI(AdminAPI, FLAdminAPISpec):
         )
 
     @wrap_with_return_exception_responses
-    def delete_run(self, job_id: str) -> FLAdminAPIResponse:
-        if not isinstance(job_id, str):
-            raise APISyntaxError("job_id must be str but got {}.".format(type(job_id)))
-        success, reply_data_full_response, reply = self._get_processed_cmd_reply_data(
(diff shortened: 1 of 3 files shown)


nvkevlu:
Another PR will be needed for updates to download_job to support URLs behind the scenes (return URL if the size of the download is larger than a certain threshold)

nvflare/fuel/hci/client/fl_admin_api.py
```diff
@@ -395,6 +379,23 @@ def list_jobs(self, options: str = None) -> FLAdminAPIResponse:
             APIStatus.ERROR_RUNTIME, {"message": "Runtime error: could not handle server reply."}, reply
         )
 
+    @wrap_with_return_exception_responses
+    def download_job(self, job_id: str) -> FLAdminAPIResponse:
+        if not job_id:
+            raise APISyntaxError("job_id is required but not specified.")
+        if not isinstance(job_id, str):
+            raise APISyntaxError("job_id must be str but got {}.".format(type(job_id)))
+        success, reply_data_full_response, reply = self._get_processed_cmd_reply_data("download_job " + job_id)
```
YuanTingHsieh, line 388: Some of these command names you use constants like `AdminCommandNames.DELETE_RUN`

Should we use constants for all these command names?

nvkevlu: made an update to address some of this for some that made sense, we can continue to clean up more in the future.

PR #670, opened 2022-06-15
Delete job command enhance

yhwen:
- not allowing delete job when job is running.
- give proper error message when delete a job not exist.
- If the download job size larger than 50GB, only provide a download URL.

### DIFF
--- a/nvflare/apis/impl/job_def_manager.py
+++ b/nvflare/apis/impl/job_def_manager.py
@@ -21,7 +21,7 @@ import tempfile
 import time
 import uuid
 from abc import ABC, abstractmethod
-from typing import Any, Dict, List
+from typing import Any, Dict, List, Optional
 
 from nvflare.apis.fl_context import FLContext
 from nvflare.apis.job_def import Job, JobDataKey, JobMetaKey, job_from_meta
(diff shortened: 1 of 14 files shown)


yhwen:
> 

Removed cli_copy..py

nvkevlu (approved):
Thanks, LGTM. Maybe in the future we could configure the MAX_DOWNLOAD_JOB_SIZE but for now this should be okay.

nvflare/apis/impl/job_def_manager.py
```diff
@@ -174,9 +177,11 @@ def _load_job_data_from_store(self, jid: str, temp_dir: str, fl_ctx: FLContext):
         unzip_all_from_bytes(data_bytes, job_id_dir)
         return job_id_dir
 
-    def get_content(self, jid: str, fl_ctx: FLContext) -> bytes:
+    def get_content(self, jid: str, fl_ctx: FLContext) -> Optional[bytes]:
         store = self._get_job_store(fl_ctx)
         stored_data = store.get_data(self.job_uri(jid))
+        if not stored_data:
```
YuanTingHsieh, line 183: Instead of changing the storage code, should we just add a try catch block here?


An enhancement would be to change storage's RuntimeError to something like NVFlareObjectNotFoundError so we don't catch other unwanted situations

nvflare/apis/impl/job_def_manager.py
```diff
@@ -134,10 +134,13 @@ def _validate_uploaded_content(self, uploaded_content) -> bool:
         """
         pass
 
-    def get_job(self, jid: str, fl_ctx: FLContext) -> Job:
+    def get_job(self, jid: str, fl_ctx: FLContext) -> Optional[Job]:
         store = self._get_job_store(fl_ctx)
         job_meta = store.get_meta(self.job_uri(jid))
-        return job_from_meta(job_meta)
```
YuanTingHsieh, line 140: Instead of changing the storage code, should we just add a try catch block here?

nvflare/apis/impl/job_def_manager.py
yhwen: when job not exist, return None make more sense than raise RuntimeError. Easier to distinguish with other real errors.

PR #671, opened 2022-06-15
Add download_job_url to project.yml and template

IsaacYangSLA:


nvflare/lighter/impl/static_file.py
```diff
@@ -28,6 +28,7 @@ def __init__(
         enable_byoc=False,
         config_folder="",
         app_validator="",
+        download_job_url="",
```
yhwen, line 31: do you need to make change when calling this constructor?

PR #674, opened 2022-06-15
Unregister client itself when client shutdown by stop_fl.sh

IsaacYangSLA:


nvflare/poc/client/startup/sub_start.sh
```diff
@@ -33,14 +33,15 @@ stop_fl() {
...
+  kill -SIGUSR2 $pid
   sleep 10
   kill -0 ${pid} 2> /dev/null 1>&2
   if [[ $? -ne 0 ]]; then
     echo "Process alreday terminated"
     return
   fi
   kill -9 $pid
-  rm -f $DIR/../pid.fl $DIR/../shutdown.fl $DIR/../restart.fl
+  rm -f $DIR/../pid.fl $DIR/../shutdown.fl $DIR/../restart.fl 2> /dev/null 1>&2
```
YuanTingHsieh, line 44: just find out there is a typo "alreday" should be "already"

IsaacYangSLA: Thanks.  Seems poc folder is lightly different from master template.

YuanTingHsieh (approved):
Thanks, LGTM

PR #681, opened 2022-06-16
Update example access result part and update documentation

YuanTingHsieh:
- Update "access result" part for hello-* examples
- Rename file `hello_numpy.rst` -> `hello_scatter_and_gather.rst`
- Remove `examples/federated_analysis/histograms_example.html`
- Fix several links and format issues in docs


examples/hello-numpy-cross-val/config/config_fed_server.json
```diff
@@ -32,6 +32,11 @@
       "id": "formatter",
       "path": "nvflare.app_common.np.np_formatter.NPFormatter",
       "args": {}
+    },
+    {
```
chesterxgchen, line 36: is this something new ?

YuanTingHsieh: no, this is not new stuff, it was specified in the text of the documentation, but somehow is not included in the example config.

chesterxgchen (approved):

PR #704, opened 2022-06-30
Fix blossom CI/CD tests

YuanTingHsieh:
- Split integration tests into TF, PT, numpy cases
- Numpy cases need only CPU
- TF and PT need GPU
- TF and PT needs to run in different container (environment)


ci/run_premerge_gpu_tf.sh
```diff
@@ -0,0 +1,72 @@
+#!/bin/bash
```
pxLi, line 1: may I ask why we dup most of code/script here? 

it would be nice to have an arg for func `integration_test()` at `ci/run_premerge_gpu.sh`
```
...
./run_integration_tests.sh -m ${1}
...
```

simply call the func
```
integration_test tensorflow
integration_test pytorch
```
when needed.

Could also extend the `BUILD_TYPE` args part to control which test/tests should be run

YuanTingHsieh: Thanks for the feedback, let me modify this.

nvkevlu (approved):

PR #708, opened 2022-07-06
Schema of user/site/org/resource

IsaacYangSLA:


nvflare/dashboard/application/models.py
```diff
@@ -0,0 +1,105 @@
...
+    title = db.Column(db.String(128), default="")
+    description = db.Column(db.String(512), default="")
+    app_location = db.Column(db.String(64), default="")
+    ha_mode = db.Column(db.Boolean, default=False)
+    starting_date = db.Column(db.String(64), default="")
+    end_date = db.Column(db.String(64), default="")
+    overseer = db.Column(db.String(64), default="")
+    server1 = db.Column(db.String(64), default="")
+    server2 = db.Column(db.String(64), default="")
+    root_cert = db.Column(db.String(2048), default="")
```
YuanTingHsieh, line 64: Could this server1 server2 be "servers"?
Or if we only have 2 servers change to primary_server, backup_server?

nvflare/dashboard/application/project.py
```diff
@@ -0,0 +1,103 @@
...
+            }
+        )
+    else:
+        return jsonify({"status": "unauthenticated"}), 401
+
+
+@app.route("/api/v1/dummy", methods=["GET"])
+def dummy_blob():
+    fileobj = Store.dummy()
+    return jsonify({"status": "ok"})
```
YuanTingHsieh, line 52: do we need this api?

nvflare/dashboard/application/models.py
IsaacYangSLA: We only support up to two servers for the project hosted by this dashboard instance.  I tried not to use primary or backup because those two servers are backup for each other.  When server1 is primary, server2 is the backup.  When server2 is primary, server1 is backup.

nvflare/dashboard/application/project.py
IsaacYangSLA: That's for testing and should be removed.

nvkevlu (approved):
LGTM and works with what I have.

PR #712, opened 2022-07-08
Clean up workflow implementations, use self._engine

YuanTingHsieh:
- Use `self._engine` as defined in `nvflare/apis/impl/controller.py`


yhwen (approved):
LGTM

PR #714, opened 2022-07-15
Docker compose

IsaacYangSLA:
Edit project.yml as before
After provision, one file (compose.yaml) and a folder (nvflare_compose) are created.
Users can do `docker compose build` and `docker compose up` (or `docker-compose build` and `docker-compose up` for old syntax) to launch the overseer, servers and clients in docker compose manner.

The admin client can reach that formation if the /etc/hosts include the ip and hostnames of those participants.

### DIFF
--- a/nvflare/lighter/impl/docker.py
+++ b/nvflare/lighter/impl/docker.py
@@ -1,0 +1,91 @@
+# Copyright (c) 2021-2022, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import copy
+import os
+import shutil
+
+import yaml
+
+from nvflare.lighter.spec import Builder
+
+
+class DockerBuilder(Builder):
+    def __init__(self, base_image="python:3.8", requirements_file="requirements.txt"):
+        """Build docker compose file."""
+        self.base_image = base_image
+        self.requirements_file = requirements_file
+
+    def _build_overseer(self, overseer, ctx):
+        protocol = overseer.props.get("protocol", "http")
+        default_port = "443" if protocol == "https" else "80"
+        port = overseer.props.get("port", default_port)
+        info_dict = copy.deepcopy(self.services["__overseer__"])
+        info_dict["volumes"] = [f"./{overseer.name}:/workspace"]
+        info_dict["ports"] = [f"{port}:{port}"]
+        info_dict["build"] = "nvflare_compose"
+        self.services[overseer.name] = info_dict
+
+    def _build_server(self, server, ctx):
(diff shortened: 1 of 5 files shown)


nvflare/lighter/impl/master_template.yml
```diff
@@ -475,3 +475,62 @@ docker_adm_sh: |
...
+          ]
+
+  volumes:
+    nvflare_svc_persist:
+
+dockerfile: |
+  RUN pip install -U pip
+  RUN pip install nvflare
+  COPY requirements.txt requirements.txt
+  RUN pip install -r requirements.txt
```
kkersten, line 536: The updated project.yml config for DockerBuilder has
`requirements_file: docker_compose_requirements.txt`
which is copied into the build dir as requirements.txt, and then into the container image and installed in the dockerfile here.

Is this file required?  It seems like the docker build would fail on the COPY and RUN pip install -r requirements.txt if it is not supplied.  I'm thinking of a situation where the user is using a custom base image that already has their requirements installed, rather than the python3.8 base image.  We may want to add the logic to make this optional.

nvflare/lighter/impl/docker.py
```diff
@@ -0,0 +1,91 @@
...
+        for server in servers:
+            self._build_server(server, ctx)
+        for client in project.get_participants_by_type("client", first_only=False):
+            self._build_client(client, ctx)
+        self.services.pop("__overseer__")
+        self.services.pop("__flserver__")
+        self.services.pop("__flclient__")
+        self.compose["services"] = self.services
+        with open(self.compose_file_path, "wt") as f:
+            yaml.dump(self.compose, f)
```
YuanTingHsieh, line 81: what is the purpose of writing out this self.compose object?

nvflare/lighter/impl/docker.py
```diff
@@ -0,0 +1,91 @@
...
+        self.template = ctx.get("template")
+        self.compose = yaml.safe_load(self.template.get("compose_yaml"))
+        self.services = self.compose.get("services")
+        self.compose_file_path = os.path.join(self.get_wip_dir(ctx), "compose.yaml")
+        overseer = project.get_participants_by_type("overseer")
+        self._build_overseer(overseer, ctx)
+        servers = project.get_participants_by_type("server", first_only=False)
+        for server in servers:
+            self._build_server(server, ctx)
+        for client in project.get_participants_by_type("client", first_only=False):
```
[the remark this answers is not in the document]
IsaacYangSLA: Default is to get the first only.  Setting it to False will return a list of all matched participants.

nvflare/lighter/impl/master_template.yml
IsaacYangSLA: Line 87 and below of docker.py handle that.
If requirements_file is not specified, default to requirements.txt.
If that file does not exist, an empty requirements.txt file is created in nvflare_compose folder in the provisioned result.
ADD requirements.txt requirements.txt will then either add the copied-version of user-specified file in nvflare_compose folder or add an empty requirements.txt.

nvflare/lighter/impl/docker.py
IsaacYangSLA: This is the main compose.yaml file.  It has the information to create/start overseer/servers/clients in docker compose up command.

nvflare/lighter/impl/master_template.yml
kkersten: Ahh, missed that.  The pip install -r of the empty file just returns.  Perfect. :)

YuanTingHsieh (approved):
LGTM

kkersten (approved):
LGTM

YuanTingHsieh (approved):
LGTM

PR #719, opened 2022-07-22
Add Collective communication and hello-mpi example

YuanTingHsieh:
- Add collective communication mechanisms
    - Collective communication constants
    - Collective communication command agent

- Add collective communication widget and functor to handle AllReduce, AllGather, Broadcast
    - Collective communication widget
    - Collective functor
    
- Add a hello-mpi example to demonstrate the capabilities
    - MPIExecutor
    - MPIController

### DIFF
--- a/examples/hello-mpi/app_client/custom/mpi_execute.py
+++ b/examples/hello-mpi/app_client/custom/mpi_execute.py
@@ -1,0 +1,64 @@
+# Copyright (c) 2021, NVIDIA CORPORATION.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+from nvflare.apis.collective_comm_constants import CollectiveCommRequestTopic, CollectiveCommShareableHeader
+from nvflare.apis.executor import Executor
+from nvflare.apis.fl_constant import FLContextKey, ReturnCode
+from nvflare.apis.fl_context import FLContext
+from nvflare.apis.shareable import Shareable, make_reply
+from nvflare.apis.signal import Signal
+
+
+class MPIExecutor(Executor):
+    def execute(
+        self,
+        task_name: str,
+        shareable: Shareable,
+        fl_ctx: FLContext,
+        abort_signal: Signal,
+    ) -> Shareable:
+        # retrieve model weights download from server's shareable
+        if abort_signal.triggered:
+            return make_reply(ReturnCode.TASK_ABORTED)
+
+        if task_name != "mpi_train":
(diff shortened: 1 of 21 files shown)


nvflare/private/fed/server/fed_server.py
```diff
@@ -596,28 +599,60 @@ def AuxCommunicate(self, request, context):
...
+            command_shareable = Shareable()
+            command_shareable.set_header(ServerCommandKey.PEER_FL_CONTEXT, shared_fl_context)
+            command_shareable.set_header(ServerCommandKey.TOPIC, topic)
+            command_shareable.set_header(ServerCommandKey.SHAREABLE, shareable)
+            data = {
+                ServerCommandKey.COMMAND: ServerCommandNames.AUX_COMMUNICATE,
+                ServerCommandKey.DATA: command_shareable,
+            }
+            if shareable.get_header(CollectiveCommShareableHeader.IS_COLLECTIVE_AUX, False):
+                with self.engine.lock:
```
yhwen, line 611: The COLLECTIVE_CONNECTION_WAITER has a timeout waiter wait for all the clients calling the AuxCommunicate, which will essentially block all the other server engine task execution because of the self.engine.lock. With a potential large self._collective_comm_timeout, it might cause the system issue?

nvflare/private/fed/server/fed_server.py
```diff
@@ -596,28 +599,60 @@ def AuxCommunicate(self, request, context):
...
+                        command_conn.send(data)
+                        while True:
+                            if command_waiter.received_requests == participate_clients:
+                                break
+                            if not command_waiter.receive_all.wait(self._collective_comm_timeout):
+                                # Known issue:: IF the number of threads is lower than number of clients (world_size)
+                                #       then this will be DEAD-LOCK
+                                # Assumption: all the client request need to be here to proceed
+                                data[CollectiveCommandKey.TIMEOUT] = f"Timeout for {self._collective_comm_timeout}"
+                                command_conn.send(data)
```
yhwen, line 630: if it's a timeout, not receiving the client AuxCommunicate update, should this send to the child process?

nvflare/private/fed/server/fed_server.py
YuanTingHsieh: I only use ``with self.engine.lock`` to get the command waiter.

You can see L617 to L625 is NOT inside this with block.

So it does not block other server engine task execution.

examples/hello-mpi/app_client/custom/mpi_execute.py
```diff
@@ -0,0 +1,64 @@
+# Copyright (c) 2021, NVIDIA CORPORATION.
```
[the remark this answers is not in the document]
YuanTingHsieh: I can add this example into the integration tests.

This code itself illustrate how they can use it.

nvflare/private/fed/server/fed_server.py
YuanTingHsieh: yes, the child process need to know things are bad so can respond to it

nvflare/private/fed/server/fed_server.py
yhwen: that's correct.

examples/hello-mpi/app_client/custom/mpi_execute.py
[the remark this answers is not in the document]
YuanTingHsieh: For unit test, can you elaborate on what do you want to see? For each controller / executor as long as they implement those functions/interfaces they are valid.

PR #723, opened 2022-07-22
Add two command options to add user/client

IsaacYangSLA:


nvflare/lighter/provision.py
```diff
@@ -25,6 +25,30 @@
...
+        args:
+        gpu_resource_key: gpu
+"""
+
+adding_user_error_msg = """
+name: $USER_EMAIL_ADDRESS
+org: $ORGANIZATION_NAME
+roles:
+    - $ROLE
+"""
```
YuanTingHsieh, line 50: I am thinking if we put these "example yaml" in separate files.
Then in the future if we are going to modify the content we don't need to change the code in this file.

nvflare/lighter/provision.py
```diff
@@ -86,6 +112,27 @@ def main():
...
+            print("The yaml file format is")
+            print(adding_user_error_msg)
+            exit(0)
+    if args.add_client:
+        try:
+            extra = load_yaml(os.path.join(current_path, args.add_client))
+            extra.update({"type": "client"})
+            participants.append(Participant(**extra))
+        except BaseException as e:
+            print("Error during adding client")
```
YuanTingHsieh, line 131: this message is not consistent with "** Error during adding user **"

suggest unify their style.

nvflare/lighter/provision.py
IsaacYangSLA: My thought was that the info for clients/users were still undecided yet.  After everything is finalized, this user-facing msg can be in a separate file.

nvflare/lighter/provision.py
IsaacYangSLA: Thanks.  Fixed.

YuanTingHsieh (approved):
LGTM, for the message we can discuss and finalize it later.

YuanTingHsieh:
@IsaacYangSLA the check says it "requires all commits to be signed".
Can you sign the commits, thanks

nvkevlu (approved):
Thanks, looks pretty good, maybe the message could be:

The yaml file format should be

instead of

The yaml file format is

PR #725, opened 2022-07-22
Helm chart

IsaacYangSLA:
Users can deploy overseer and servers to kubernetes cluster with Helm Chart
1. Need to edit ConfigMap of ingress tcp conf and DaemonSet of nginx ingress
2. overseer and server names must be a DNS-1035 label (lower case alphanumeric or '-')
3. machine running clients or consoles must set /etc/hosts for resolving overseer and server names.

### DIFF
--- a/nvflare/fuel/hci/server/hci.py
+++ b/nvflare/fuel/hci/server/hci.py
@@ -107,7 +107,7 @@ class AdminServer(socketserver.ThreadingTCPServer):
             server_key: server's private key file
             accepted_client_cns: list of accepted Common Names from client, if specified
         """
-        socketserver.TCPServer.__init__(self, (host, port), _MsgHandler, False)
+        socketserver.TCPServer.__init__(self, ("0.0.0.0", port), _MsgHandler, False)
 
         self.use_ssl = False
         if ca_cert and server_cert:
--- a/nvflare/lighter/impl/master_template.yml
+++ b/nvflare/lighter/impl/master_template.yml
@@ -64,7 +64,7 @@ readme_fs: |
   Please always safeguard the server.key.
 
 gunicorn_conf_py: |
-  bind="{~~hostname~~}:{~~port~~}"
+  bind="0.0.0.0:{~~port~~}"
   cert_reqs=2
   do_handshake_on_connect=True
   timeout=30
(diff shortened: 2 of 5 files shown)


nvflare/fuel/hci/server/hci.py
```diff
@@ -107,7 +107,7 @@ def __init__(
             server_key: server's private key file
             accepted_client_cns: list of accepted Common Names from client, if specified
         """
-        socketserver.TCPServer.__init__(self, (host, port), _MsgHandler, False)
+        socketserver.TCPServer.__init__(self, ("0.0.0.0", port), _MsgHandler, False)
```
YuanTingHsieh, line 110: why do we change this from "host" to "0.0.0.0"?

nvflare/private/fed/server/fed_server.py
```diff
@@ -174,7 +174,9 @@ def deploy(self, args, grpc_args=None, secure_train=False):
                 root_certificates=root_ca,
                 require_client_auth=True,
             )
-            self.grpc_server.add_secure_port(target, server_credentials)
+            port = target.split(":")[1]
+            tmp_target = f"0.0.0.0:{port}"
+            self.grpc_server.add_secure_port(tmp_target, server_credentials)
```
YuanTingHsieh, line 179: why are we ignoring the "host" specified in target?

nvflare/lighter/project.yml
```diff
@@ -87,6 +87,9 @@ builders:
     args:
       base_image: python:3.8
       requirements_file: docker_compose_requirements.txt
+  - path: nvflare.lighter.impl.helm_chart.HelmChartBuilder
+    args:
+      docker_image: localhost:32000/nvfl-min:0.0.1
```
YuanTingHsieh, line 92: how do we get this port and image name?

nvflare/lighter/impl/master_template.yml
```diff
@@ -533,4 +533,139 @@ dockerfile: |
...
+          system: overseer
+      spec:
+        volumes:
+          - name: workspace
+            hostPath:
+              path:
+              type: Directory
+        containers:
+          - name: overseer
+            image: nvflare-min:2.2.0
```
YuanTingHsieh, line 603: how do we get this image build, and this name?

nvflare/fuel/hci/server/hci.py
IsaacYangSLA: In k8s, the host name is controlled by the cluster.  Binding 0.0.0.0 is a working solution so the server can listen to the port.

nvflare/private/fed/server/fed_server.py
IsaacYangSLA: The same as the above.

nvflare/lighter/project.yml
IsaacYangSLA: The person who is going to run helm install knows where the image is.  This setting, localhost:32000, is a typical docker registry for microk8s.

nvflare/fuel/hci/server/hci.py
YuanTingHsieh: should we get rid of this "host" arg is it is not used?
@yhwen you see any issues removing this host arg?

nvflare/private/fed/server/fed_server.py
```diff
@@ -174,7 +174,9 @@ def deploy(self, args, grpc_args=None, secure_train=False):
...
             self.logger.info("starting secure server at %s", target)
         else:
             self.grpc_server.add_insecure_port(target)
```
YuanTingHsieh, line 182: Should we also change other places using this "target"?

nvflare/lighter/impl/master_template.yml
IsaacYangSLA: This will be replaced by the docker_image in helm chart builder.  Basically, anything running in k8s cluster must be docker images.  The person who controls the k8s, with kubectl or helm, knows where to pull the images and he needs to build the image and pushes it to the registry.

nvflare/private/fed/server/fed_server.py
YuanTingHsieh: got it, Should we also change other places using this "target"?

and maybe we want to change the "fed_server.json" to get rid of the host

nvflare/private/fed/server/fed_server.py
IsaacYangSLA: helm chart only runs in secure mode.

nvflare/lighter/impl/master_template.yml
YuanTingHsieh: got it, we will need some documentation updates to explain these stuff to the users.

nvflare/lighter/project.yml
YuanTingHsieh: got it, we will need some documentation updates to explain these stuff to the users.

nvflare/fuel/hci/server/hci.py
IsaacYangSLA: I think we should keep that arg right now.  The information could be used somewhere else.

YuanTingHsieh (approved):
Thanks, LGTM.

PR #744, opened 2022-08-03
Add two functions to retrieve roles and org from cert

IsaacYangSLA:


nvflare/fuel/hci/security.py
```diff
@@ -80,3 +80,41 @@ def get_certificate_common_name(cert: dict):
...
+
+    Returns: roles of provided cert
+
+    """
+    if cert is None:
+        return None
+
+    for sub in cert.get("subject", ()):
+        for key, value in sub:
+            if key == "unstructuredName":
```
YuanTingHsieh, line 99: should we use NameOID.ORGANIZATION_NAME to replace this string?

nvflare/fuel/hci/security.py
```diff
@@ -80,3 +80,41 @@ def get_certificate_common_name(cert: dict):
...
+
+    Returns: organization of provided cert
+
+    """
+    if cert is None:
+        return None
+
+    for sub in cert.get("subject", ()):
+        for key, value in sub:
+            if key == "organizationName":
```
YuanTingHsieh, line 118: should we use NameOID.UNSTRUCTURED_NAME to replace this ?

nvflare/fuel/hci/security.py
IsaacYangSLA: roles are stored under Subject.unstructuredName in the form of "role1|role2|role3|...".  Because this function is to get the roles of peer cert, the key will be "unstructuredName."
From openssl:
Subject: CN = admin@nvidia.com, O = nvidia, unstructuredName = super

nvflare/fuel/hci/security.py
IsaacYangSLA: The org is stored under Subject.organizationName.
Like the above comment, this function to get org, so we check "organizationName."

Subject: CN = admin@nvidia.com, O = nvidia, unstructuredName = super

nvflare/fuel/hci/security.py
IsaacYangSLA: In peer cert (ssl package), all keys are string.  The NameOID is only used in the cryptography package.

[peer cert](https://docs.python.org/3/library/ssl.html#ssl.SSLSocket.getpeercert)
[NameOID](https://cryptography.io/en/latest/x509/reference/#cryptography.x509.oid.NameOID)

YuanTingHsieh (approved):
LGTM

nvkevlu (dismissed):

nvkevlu (approved):

PR #751, opened 2022-08-07
nvflare poc  syntax update  and add NVFLARE_POC_WORKSPACE env variable

chesterxgchen:
1) Remove workspace option, using nvflare_poc.json config as replacement
Main idea behind it:
* -w workspace is a lot word to type
* when start/stop/clean poc, one need to always -w worksapce to make sure the workspace is correct, which is too much typing, nvflare poc --start/stop/clean is just much simpler.
* current default workspace is hard-coded to /tmp/nvflare/poc. One can't change it. so instead of using -w worksapce, we add the configuration to allow user to change it

* the optional env variable NVFLARE_POC_WORKSPACE="/tmp/nvflare/poc2"

if the NVFLARE_POC_WORKSPACEis specified, it will be used instead of default. This way, the workspace will be created at /tmp/nvflare/poc2 with above config using command nvflare poc --prepare

2) replace hard-coded "transfer" directory with upload_dir read from admin/startup/fed_admin.json config

### DIFF
--- a/nvflare/lighter/poc_commands.py
+++ b/nvflare/lighter/poc_commands.py
@@ -12,12 +12,15 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import json
 import os
 import random
 import sys
 
 from nvflare.lighter.poc import generate_poc
 from nvflare.lighter.service_constants import FlareServiceConstants as SC
+
+DEFAULT_WORKSPACE = "/tmp/nvflare/poc"
 
 
 def get_package_command(cmd_type: str, poc_workspace: str, package_dir) -> str:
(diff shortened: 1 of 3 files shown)


holgerroth (approved):
LGTM

holgerroth (approved):

nvflare/lighter/poc_commands.py
```diff
@@ -207,20 +224,29 @@ def is_poc(cmd_args) -> bool:
     )
 
 
+def get_poc_config_path() -> str:
+    user_home = os.path.expanduser("~")
+    return os.path.join(user_home, POC_CONFIG_FILE)
+
+
```
YuanTingHsieh, line 231: where is POC_CONFIG_FILE and is this func being used?

YuanTingHsieh (approved):
Thanks, LGTM

PR #761, opened 2022-08-11
Clean up fuel utils and client executor

YuanTingHsieh:
fuel/utils:

- Removed unused files/functions

client_executor.py:

- Removed unused files/functions
- Renamed `start_train\abort_train` to `start_app\abort_app`


nvflare/fuel/utils/dict_utils.py
```diff
@@ -12,18 +12,6 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
```
yhwen, line 14: This update(d, u) is a useful function. Better to keep it even we are not directly use it right now.

nvflare/private/fed/client/client_executor.py
```diff
@@ -24,138 +24,20 @@
...
-        """
-        pass
-
-    def cleanup(self):
-        """Cleanup."""
-        self.pipe.clear()
-
-
-class ProcessExecutor(ClientExecutor):
+class ProcessExecutor:
```
yhwen, line 31: We used to have another implementation ThreadExecutor, which implement the ClientExecutor APIs using multi-threads, instead of separate sub-process. Probably still keep this API architecture with more options?

yhwen (approved):
look good now.

PR #763, opened 2022-08-12
Add GPU resource consumer

YuanTingHsieh:
- Create a base class `BaseResourceManager`
- Add GPUResourceManager (it reads from site.json)
  -  it supports number of gpus and gpu memory (treat each GPU has same GPU memory)
- Add GPUResourceManager unit test
- Update GPUResourceConsumer to match with GPUResourceManager

### DIFF
--- a/nvflare/fuel/utils/gpu_utils.py
+++ b/nvflare/fuel/utils/gpu_utils.py
@@ -16,25 +16,59 @@ import subprocess
 from typing import List
 
 
-def get_host_gpu_ids() -> List:
+def use_nvidia_smi(query: str, report_format: str = "csv"):
     try:
         process = subprocess.Popen(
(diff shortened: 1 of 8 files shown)


nvflare/fuel/utils/gpu_utils.py
```diff
@@ -16,25 +16,53 @@
 from typing import List
 
 
-def get_host_gpu_ids() -> List:
+def use_nvidia_smi(
+    query: str,
+    report_format: str = "csv"
+):
     try:
         process = subprocess.Popen(
```
nvidianz, line 24: subprocess.run() may be better suited this case.

nvidianz (approved):

nvidianz (approved):

nvflare/fuel/utils/gpu_utils.py
YuanTingHsieh: Thanks fixed

PR #773, opened 2022-08-17
handle no gpu or no nvidia-smi installed cases

chesterxgchen:


nvflare/fuel/utils/gpu_utils.py
```diff
@@ -17,17 +17,22 @@
...
-        # 'GPU 0: NVIDIA GeForce RTX 3090 (UUID: GPU-xxxx-xxxx-xxxx-xxx)\n'
-        if result[0].startswith("GPU"):
-            gpus = result[0].split("\n")
-            gpu_ids = [int(gpu.split(":")[0].split(" ")[1]) for gpu in gpus[:-1]]
+    try:
+        process = subprocess.Popen(["nvidia-smi", "--list-gpus"], stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True)
+        result = process.communicate()
+        rc = process.returncode
+        if rc > 0:
+            raise Exception("Failed to get host gpu device Ids", result[0])
```
YuanTingHsieh, line 25: Do we want to catch this exception or not?
If no nvidia-smi -> return []
then if nvidia-smi exist but execution wrong -> what should be return?

YuanTingHsieh (commented):
leave a comment

YuanTingHsieh (approved):
thanks LGTM

PR #774, opened 2022-08-17
consolidate NVFLARE CLI

chesterxgchen:
1) add nvflare provision command ( add deprecate warning to provision command) 
2) add nvflare preflight_check command

### DIFF
--- a/nvflare/lighter/nvflare.py
+++ b/nvflare/lighter/nvflare.py
@@ -17,32 +17,43 @@ import os
 import sys
 
 from nvflare.lighter.poc_commands import def_poc_parser, handle_poc_cmd, is_poc
+from nvflare.lighter.provision import define_provision_parser, handle_provision
+from nvflare.tool.preflight_check import define_preflight_check_parser, check_packages
 
 
-def handle_provision_cmd(cmd_args):
-    print("handle provision command")
-    pass
+def is_provision(cmd_args) -> bool:
+    return (
+            hasattr(cmd_args, "add_user")
+            or hasattr(cmd_args, "add_client")
+            or hasattr(cmd_args, "project_file")
+            or hasattr(cmd_args, "ui_tool")
+    )
 
 
-#
-#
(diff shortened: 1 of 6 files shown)


nvflare/lighter/nvflare.py
```diff
@@ -17,45 +17,58 @@
...
+            handle_poc_cmd(prog_args)
+        elif is_provision(prog_args):
+            handle_provision(prog_args)
+        elif is_preflight_checker(prog_args):
+            check_packages(prog_args)
+        else:
+            prog_parser.print_help()
+    except Exception as e:
+        print("unable to handle command, please check syntax")
         prog_parser.print_help()
```
YuanTingHsieh, line 72: if we have this generic entry "nvflare cli" inside lighter, maybe we just move all the preflight check codes into the "nvflare/lighter" folder?

YuanTingHsieh: but that can be done in a separate PR.

nvflare/lighter/poc_commands.py
```diff
@@ -317,30 +336,37 @@ def is_poc(cmd_args) -> bool:
...
+        gpu_ids = get_gpu_ids(cmd_args.gpu, get_local_host_gpu_ids())
+        start_poc(poc_workspace, gpu_ids, excluded, white_list)
+    elif cmd_args.prepare_poc:
+        prepare_poc(cmd_args.number_of_clients, poc_workspace)
+    elif cmd_args.stop_poc:
+        stop_poc(poc_workspace, excluded, white_list)
+    elif cmd_args.clean_poc:
+        clean_poc(poc_workspace)
+    else:
+        raise Exception(f"unable to handle poc command:{cmd_args}")
```
YuanTingHsieh, line 372: should this be CLIException?

YuanTingHsieh: got it you catch it in nvflare.py

YuanTingHsieh (approved):
Thanks, LGTM

PR #781, opened 2022-08-21
Fix return message_list in admin_msg_sender

YuanTingHsieh:
- Should still return empty list when exception happens


yanchengnv (approved):
LGTM

PR #792, opened 2022-08-23
Simulator for 2.2.0 release

yhwen:


nvflare/apis/utils/fl_context_utils.py
```diff
@@ -36,6 +36,7 @@ def get_serializable_data(fl_ctx: FLContext):
 
 
 def generate_log_message(fl_ctx: FLContext, msg: str):
+    _identity_ = "identity"
```
nvidianz, line 39: Those are constants. They should be all caps and placed at module level.

nvflare/private/fed/app/deployer/simulator_deployer.py
```diff
@@ -0,0 +1,164 @@
...
+
+
+class SimulatorDeployer(ServerDeployer):
+    def __init__(self, ports=None):
+        super().__init__()
+        if ports:
+            self.open_ports = ports
+        else:
+            self.open_ports = get_open_ports(2)
+        self.admin_storage = tempfile.mkdtemp()
```
nvidianz, line 38: Should add a prefix here so it's easier to do find the folder and do cleanup.

nvflare/private/fed/app/deployer/simulator_deployer.py
```diff
@@ -0,0 +1,164 @@
...
+from ..server.server_train import create_admin_server
+from .base_client_deployer import BaseClientDeployer
+from .server_deployer import ServerDeployer
+
+
+class SimulatorDeployer(ServerDeployer):
+    def __init__(self, ports=None):
+        super().__init__()
+        if ports:
+            self.open_ports = ports
```
YuanTingHsieh, line 35: Maybe should add a check here saying the ports need to be of length 2?
Or for simplicity we just don't allow this arg

nvflare/private/fed/app/deployer/simulator_deployer.py
```diff
@@ -0,0 +1,164 @@
...
+        federated_client.register()
+        federated_client.start_heartbeat()
+        federated_client.run_manager = None
+        # servers = [{t["name"]: t["service"]} for t in client_config.get("servers")]
+        # admin_agent = self.create_admin_agent(
+        #     sorted(servers)[0],
+        #     federated_client,
+        #     args,
+        # )
+        # admin_agent.start()
```
YuanTingHsieh, line 85: suggest remove comments.

nvflare/private/fed/app/simulator/simulator_runner.py
```diff
@@ -0,0 +1,383 @@
...
+                        + ",".join([name for name in split_client_names[index]])
+                        + " -g "
+                        + gpus[index]
+                        + " -p "
+                        + ",".join([str(i) for i in self.deployer.open_ports])
+                    )
+                    _ = subprocess.Popen(shlex.split(command, True), preexec_fn=os.setsid, env=os.environ.copy())
+            else:
+                client_runner = SimulatorClientRunner(self.args, self.client_names, self.deployer)
+                client_runner.run()
```
YuanTingHsieh, line 230: In here you are using subprocess.Popen if self.args.gpu is active.

So inside your SimulatorClientRunner.run() you don't need to use subprocess.Popen anymore

nvflare/private/fed/app/simulator/simulator_runner.py
```diff
@@ -0,0 +1,383 @@
...
+        command = (
+            sys.executable
+            + " -m nvflare.private.fed.app.simulator.simulator_worker -o "
+            + self.args.workspace
+            + " --client "
+            + client.client_name
+            + " --port "
+            + str(open_port)
+        )
+        _ = subprocess.Popen(shlex.split(command, True), preexec_fn=os.setsid, env=os.environ.copy())
```
YuanTingHsieh, line 337: should not need this

nvflare/apis/utils/fl_context_utils.py
yhwen: These are only used here locally in the generate_log_message(). I think it's okay to be here. And they are all consistent with all the others.

nvflare/private/fed/app/deployer/simulator_deployer.py
yhwen: The tempfile.mkdtemp() creates the folder in the standard location, easy for cleanup. Plus the close() will do the cleanup as well.

nvflare/private/fed/app/deployer/simulator_deployer.py
yhwen: done

nvflare/private/fed/app/deployer/simulator_deployer.py
yhwen: done

nvflare/private/fed/app/simulator/simulator_runner.py
yhwen: This has to be in a separate child process. Pytorch / Ignite could not release thee GPU memory fully. Otherwise, we will  see GPU OOM over a period of time.

nvflare/private/fed/app/simulator/simulator_runner.py
yhwen: This is to split the clients onto different GPUs. Multi-GPU need to have sub-processes to run different clients.

nvflare/private/fed/app/simulator/simulator_worker.py
```diff
@@ -0,0 +1,221 @@
...
+        task_worker.run(args, conn)
+    finally:
+        stop_event.set()
+        conn.close()
+        AuditService.close()
+
+
+if __name__ == "__main__":
+    """
+    This is the main program of simulator worker process when running the NVFlare Simulator using multi-GPU.
```
YuanTingHsieh, line 216: even if single GPU we are using this script to run.

Suggest change the description

```
This is the main program of simulator worker process when running the NVFlare Simulator
```

nvflare/private/fed/app/simulator/simulator_runner.py
YuanTingHsieh: got it, so the self.args.gpu does NOT indicate whether to use GPU.

It actually means self.multi_gpu.
Then can we change the option to let user pass in multi_gpu?
And required users to also specified a GPU list if multi_gpu is enabled

YuanTingHsieh: This also means either way we are going to use subprocess.

I suggest we always use subprocess Popen in here to run client_run_process.py (or we rename it as simulator_worker.py):

   1. If user not specify, we launch 1 client_run_process.py.
   2. If the user specify multi-GPU we launch several client_run_processes.py.

For the "SimulatorClientRunner.run()" function we can just run things directly.
We don't need to do another subprocess Popen there.

yhwen: This has been changed. Only ClientTaskWorker run with subprocess.

nvflare/private/fed/app/simulator/simulator_worker.py
yhwen: done

nvflare/private/fed/app/simulator/simulator_runner.py
```diff
@@ -0,0 +1,383 @@
...
+            + " -m nvflare.private.fed.app.simulator.simulator_worker -o "
+            + self.args.workspace
+            + " --client "
+            + client.client_name
+            + " --port "
+            + str(open_port)
+        )
+        if gpu:
+            command += " --gpu " + str(gpu)
+        _ = subprocess.Popen(shlex.split(command, True), preexec_fn=os.setsid, env=os.environ.copy())
```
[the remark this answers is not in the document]
yhwen: we could construct the command in an array directly. but the command in a string is better readable when debug.

nvflare/private/fed/app/simulator/simulator_runner.py
```diff
@@ -0,0 +1,383 @@
...
+        self.simulator_root = os.path.join(self.args.workspace, SimulatorConstants.JOB_NAME)
+        if os.path.exists(self.simulator_root):
+            shutil.rmtree(self.simulator_root)
+
+        try:
+            data_bytes, job_name, meta = self.validate_job_data()
+
+            if not self.client_names:
+                self.client_names = self._extract_client_names_from_meta(meta)
+            if not self.client_names:
```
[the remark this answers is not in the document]
yhwen: Yes, if meta uses "@all", and provides no client name option, it will result in error message:
"Please provide the client names list, or the number of clients to run the simulator"

nvflare/private/fed/app/simulator/simulator.py
```diff
@@ -0,0 +1,52 @@
...
+    if sys.version_info >= (3, 9):
+        raise RuntimeError("Python versions 3.9 and above are not yet supported. Please use Python 3.8 or 3.7.")
+    if sys.version_info < (3, 7):
+        raise RuntimeError("Python versions 3.6 and below are not supported. Please use Python 3.8 or 3.7.")
+    args = parse_args()
+
+    simulator = SimulatorRunner(args)
+    if simulator.setup():
+        simulator.run()
+    os._exit(0)
```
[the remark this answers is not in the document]
yhwen: os._exit(0) calls the termination right away. It ensures and forces the program exit.

IsaacYangSLA (approved):
Approved based on others' approval.

PR #801, opened 2022-08-26
Add dashboard cli

IsaacYangSLA:


nvflare/dashboard/cli.py
```diff
@@ -0,0 +1,112 @@
...
+            entrypoint=["/usr/local/bin/python3", "nvflare/web/wsgi.py"],
+            detach=True,
+            auto_remove=True,
+            name="nvflare-dashboard",
+            ports={8443: port},
+            volumes={folder: {"bind": "/var/tmp/nvflare/web", "model": "rw"}},
+            environment=environment,
+        )
+    except docker.errors.APIError as e:
+        print(f"Either {dashboard_image} image does not exit or another nvflare-dashboard instance is still running.")
```
nvkevlu, line 59: is "does not exit" intended here? or does not exist?

IsaacYangSLA: Thanks.  That's my typo.  You are right.  It should read "does not exist."

nvkevlu:
CI is failing due to f strings with no placeholders in prints of ln 31 and ln 64

nvkevlu (approved):
Thanks, LGTM

nvflare/dashboard/cli.py
```diff
@@ -0,0 +1,112 @@
...
+        environment.update({"NVFL_CREDENTIAL": f"{answer}:{pwd}"})
+    if app_url and app_dest_filename:
+        images_folder = os.path.join(folder, "images")
+        if not os.path.exists(images_folder):
+            os.makedirs(images_folder)
+        dest_file = os.path.join(images_folder, app_dest_filename)
+        print(f"downloding {app_url} to {dest_file}")
+        with open(dest_file, "wb") as out_file:
+            content = requests.get(app_url, stream=True).content
+            out_file.write(content)
```
YuanTingHsieh, line 45: if the "app_url" is not up and running what will happen here, do we need to handle that?

nvflare/dashboard/cli.py
```diff
@@ -0,0 +1,112 @@
...
+import requests
+
+from nvflare.lighter.utils import generate_password
+
+
+def start(port, folder, app_url, app_dest_filename, dashboard_image):
+    if not folder:
+        folder = os.getcwd()
+    environment = dict()
+    if not os.path.exists(os.path.join(folder, "db.sqlite")):
```
YuanTingHsieh, line 29: what is this "db.sqlite"? why we don't need to generate password if it exists

setup.py
```diff
@@ -84,6 +84,7 @@
         "PyYAML==6.0",
         "six>=1.15.0",
         "tenseal==0.3.0",
+        "docker>=6.0",
```
YuanTingHsieh, line 87: no question regarding this change.

Just a general feedback, we really need to sort out the dependencies with a better package strategy.

nvflare/dashboard/cli.py
IsaacYangSLA: The dashboard can not run without that if that's what project admin requires.

However, that portion of code may be removed soon because we will have different way to get application images.

nvflare/dashboard/cli.py
IsaacYangSLA: This is the sqlite database file.  If it exists, that means project admin is created previously and stored in the database.  Project admin can login the dashboard with the email address and the randomly generated password.

setup.py
IsaacYangSLA: Agreed.  This requires some improvement.

PR #803, opened 2022-08-26
Fixed typos and return run status for simulator

yhwen:
- Fix typos
- Return run_status
- Clean up codes


nvflare/private/fed/app/simulator/simulator_runner.py
```diff
@@ -53,6 +53,8 @@ def __init__(self, args):
         self.deployer = SimulatorDeployer()
         self.client_names = []
 
+        self.args.set = []
```
YuanTingHsieh, line 56: Do we need this `self.args`?

nvflare/private/fed/app/simulator/simulator_runner.py
```diff
@@ -224,10 +226,13 @@ def run(self):
 
             executor.shutdown()
             server_thread.join()
+            run_status = 0
         except BaseException as error:
             self.logger.error(error)
+            run_status = 2
```
YuanTingHsieh, line 232: Why use `2` here? Why don't we use `1`

nvflare/private/fed/client/client_executor.py
```diff
@@ -141,7 +135,7 @@ def send_aux_command(self, shareable: Shareable, job_id):
 
     def cleanup(self):
         """Cleanup."""
-        self.pipe.clear()
+        pass
```
YuanTingHsieh, line 138: Just a note that part of the cleanup of this to the dev branch is in PR: https://github.com/NVIDIA/NVFlare/pull/761

nvflare/private/fed/app/simulator/simulator.py
```diff
@@ -48,5 +47,7 @@ def parse_args():
 
     simulator = SimulatorRunner(args)
     if simulator.setup():
-        simulator.run()
-    os._exit(0)
+        run_status = simulator.run()
+    else:
+        run_status = 1
+    os._exit(run_status)
```
YuanTingHsieh, line 53: why we use `os._exit` here? should it be `sys.exit`?

nvflare/private/fed/app/simulator/simulator_runner.py
yhwen: yes, need this to create the client. Share the same codes.

nvflare/private/fed/app/simulator/simulator_runner.py
yhwen: Use this to distinguish the error code.

nvflare/private/fed/app/simulator/simulator.py
yhwen: os._exit() to terminate right away and force exit. sys.exit is not guarantee.

YuanTingHsieh: But according to the python doc: 
"https://docs.python.org/3/library/os.html#os._exit"

Note The standard way to exit is sys.exit(n). [_exit()](https://docs.python.org/3/library/os.html#os._exit) should normally only be used in the child process after a [fork()](https://docs.python.org/3/library/os.html#os.fork).

nvflare/private/fed/app/simulator/simulator.py
```diff
@@ -48,5 +47,7 @@ def parse_args():
 
     simulator = SimulatorRunner(args)
     if simulator.setup():
-        simulator.run()
-    os._exit(0)
+        run_status = simulator.run()
+    else:
+        run_status = 1
```
YuanTingHsieh, line 52: should we log "setup" fail here so people know this exit code of 1 means setup failed.

yhwen: Detail error message already in the log. We will document the error code.

nvflare/private/fed/app/simulator/simulator.py
yhwen: sys.exit() will not guarantee the process to be terminated. It could result in the process to hang, for example if a resource is not fully released. os._exit() will force the process to be terminated.

YuanTingHsieh: Got it.

But another question here, if the resource is not fully released, should we terminate the process?

yhwen: we need to terminate the process.

nvflare/private/fed/client/client_executor.py
yhwen: Once merged with dev, this will be addressed with other PR.

PR #807, opened 2022-08-29
Update cert reading (get_identity_info) and component (resource manager)

IsaacYangSLA:
Three identity info is retrieved from get_identity_info, common_name, organization and role.
The existing get_certificate_common_name function is not removed for compatibility.
All users have at most one role.
Default client site resource manager changed to GPUResourceManager and args format changed, too.

### DIFF
--- a/nvflare/fuel/hci/security.py
+++ b/nvflare/fuel/hci/security.py
@@ -64,6 +64,33 @@ def make_session_token():
     return str(t)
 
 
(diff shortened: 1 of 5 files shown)


YuanTingHsieh (approved):
LGTM, just one question, we keep `get_certificate_common_name` to be compatible with what?
are we using this function somewhere else in the code?

yhwen (approved):

PR #808, opened 2022-08-29
Include org info in sub_start.sh, compose.yaml, helm chart and other …

IsaacYangSLA:
…related files.

org is set via the --set option group when starting python script


nvkevlu (dismissed):
Thanks, LGTM

nvkevlu (approved):
Thanks, LGTM

PR #813, opened 2022-08-31
Addressed several QA issues.

yhwen:
- Added message for FLCommunicationError.

### DIFF
--- a/nvflare/private/fed/client/fed_client_base.py
+++ b/nvflare/private/fed/client/fed_client_base.py
@@ -190,7 +190,8 @@ class FederatedClientBase:
 
             return task
         except FLCommunicationError as e:
-            self.logger.info(e)
+            if e:
+                self.logger.info(e)
 
     def push_execute_result(self, project_name, shareable: Shareable, fl_ctx: FLContext):
         """Submit execution results of a task to server.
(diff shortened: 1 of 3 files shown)


nvflare/private/fed/client/fed_client_base.py
```diff
@@ -190,7 +190,8 @@ def fetch_execute_task(self, project_name, fl_ctx: FLContext):
 
             return task
         except FLCommunicationError as e:
-            self.logger.info(e)
+            if e:
+                self.logger.info(e)
```
YuanTingHsieh, line 194: Since this exception is thrown, there must be something in there.

We should look for how we construct this `FLCommunicationError` and put relevant information there?

yhwen: Added the FLCommunicationError for different places.

YuanTingHsieh: thanks, then we should just log them out right?
so we don't need this "if" condition

YuanTingHsieh (changes requested):
Since we have the error messages, we should just log them out.
So we don't need these if conditions

nvflare/private/fed/client/fed_client_base.py
yhwen: done

YuanTingHsieh (approved):
thanks, LGTM

PR #816, opened 2022-09-01
Optimize the SimulatorRunner config.

yhwen:
Optimize the SimulatorRunner config and changed some simulator option names based on researchers feedback.


nvflare/private/fed/app/simulator/simulator.py
```diff
@@ -24,10 +24,10 @@
 def parse_args():
     parser = argparse.ArgumentParser()
     parser.add_argument("job_folder")
-    parser.add_argument("--workspace", "-o", type=str, help="WORKSPACE folder", required=True)
+    parser.add_argument("--workspace", "-w", type=str, help="WORKSPACE folder", required=True)
     parser.add_argument("--clients", "-n", type=int, help="number of clients")
-    parser.add_argument("--client_list", "-c", type=str, help="client names list")
```
holgerroth, line 29: I think this should be
```
    parser.add_argument("--n_clients", "-n", type=int, help="number of clients")
    parser.add_argument("--clients", "-c", type=str, help="client names list")
```

holgerroth (changes requested):
Still some inconsistency in the arguments. See comment. Otherwise looks good.

nvflare/private/fed/app/simulator/simulator.py
yhwen: fixed.

yhwen:
> is this PR is similar to to other PR

This PR is for 2.2 branch. The other one is for dev. Unfortunately we have to make it into 2 PR, and codes are different.

nvflare/private/fed/app/simulator/simulator_runner.py
```diff
@@ -43,23 +44,45 @@
...
-            for i in range(self.args.clients):
+        self.args = self._generate_args(
+            self.job_folder, self.workspace, self.clients, self.n_clients, self.threads, self.gpu
+        )
+
+        if self.args.clients:
+            self.client_names = self.args.clients.strip().split(",")
+        elif self.args.n_clients:
+            for i in range(self.args.n_clients):
                 self.client_names.append("site-" + str(i + 1))
```
YuanTingHsieh, line 86: what happens if users specify both "clients" and "n_clients"? 
Should we give a warning or disallow that?

nvflare/private/fed/app/simulator/simulator_runner.py
```diff
@@ -43,23 +44,45 @@
...
+        return args
 
     def setup(self):
-        if self.args.client_list:
-            self.client_names = self.args.client_list.strip().split(",")
-        elif self.args.clients:
-            for i in range(self.args.clients):
+        self.args = self._generate_args(
+            self.job_folder, self.workspace, self.clients, self.n_clients, self.threads, self.gpu
+        )
```
YuanTingHsieh, line 80: Should we just pass "required" args into the `SimulatorServerAppRunner` instead of everything (self.args)?

```
server_app_runner = SimulatorServerAppRunner()
        snapshot = None
        server_app_runner.start_server_app(
            self.services, self.args, app_server_root, self.args.job_id, snapshot, self.logger
        )
```

nvflare/private/fed/app/simulator/simulator_runner.py
yhwen: "n_clients" will take precedence.

nvflare/private/fed/app/simulator/simulator_runner.py
yhwen: These are all the args coming from the CLI. We need to pass them all to SimulatorServerAppRunner

holgerroth (approved):
Looks good!

nvflare/private/fed/app/simulator/simulator_runner.py
YuanTingHsieh: actually "clients" will take precedence, but I get it, we just need to make sure we have noted this in doc.

Thanks

PR #843, opened 2022-09-06
Fix get_connected_clients in fl_admin_api

### LINKED ISSUE #753: Admin API get_connected_client_list() includes the table heading "CLIENT" in results
When using the python admin api to `get_connected_client_list()`, the reply details includes the table header "CLIENT" with the actual results:
```
>>> adm.api.get_connected_client_list()
{'status': <APIStatus.SUCCESS: 'SUCCESS'>, 'details': {<FLDetailKey.CONNECTED_CLIENTS: 'connected_clients'>: ['CLIENT', 'site2', 'site1']}}
```

Straightforward fix --

There's a small mismatch in the column headers hardcoded in two files. The column header is defined as "Client" in the following (.upper gets called elsewhere): https://github.com/NVIDIA/NVFlare/blob/65848ff5ecdac29b26ca1a448f2106e06cda1d23/nvflare/private/fed/server/training_cmds.py#L589
however the admin api function is trying to filter out "CLIENT NAME" instead: https://github.com/NVIDIA/NVFlare/blob/99c199e2ed2b737a939114e23d4c4b765e71e0c3/nvflare/fuel/hci/client/fl_admin_api.py#L769-L776

Hope this helps!

YuanTingHsieh:
Fix #753

### DIFF
--- a/nvflare/fuel/hci/client/fl_admin_api.py
+++ b/nvflare/fuel/hci/client/fl_admin_api.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+
 import os
 import re
 import time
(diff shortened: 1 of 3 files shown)


nvkevlu (approved):
Only note is that the capitalization of the headers is now all lowercase, so the output will look different, but if that is okay then this LGTM, thanks

YuanTingHsieh:
Thanks for the review.

When a Table object is created, it always capitalize the headers, so the output will stay the same.

PR #852, opened 2022-09-07
add make sure it can include log.config file in 2.2 wheel [skip ci]

chesterxgchen:


IsaacYangSLA (approved):

setup.py
```diff
@@ -60,7 +60,7 @@
         ],
         exclude=["tests", "tests.*"],
     ),
-    package_data={"": ["*.yml", "*.html", "poc.zip"]},
+    package_data={"": ["log.config", "*.yml", "*.html", "poc.zip"]},
```
yhwen, line 63: maybe it's better to include the whole resource folder, instead of individual file?

PR #862, opened 2022-09-08
Re-factor hello-xgboost codes

YuanTingHsieh:
- Use a single Executor instead of Learner and LearnerExecutor
- Use gpu_hist instead of hist
- Use absolute path for data
- Add more app to showcase how to handle different data paths on each site


nvflare/app_opt/xgboost/executor.py
```diff
@@ -17,11 +17,12 @@
 from nvflare.apis.fl_constant import FLContextKey, ReturnCode
 from nvflare.apis.fl_context import FLContext
 from nvflare.apis.shareable import Shareable, make_reply
-from nvflare.app_opt.xgboost.learner_base import XGBFedLearnerBase, XGBoostParams
 from nvflare.utils.import_utils import optional_import
 
+from .executor_base import XGBExecutorBase, XGBoostParams
```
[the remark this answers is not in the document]
YuanTingHsieh: fixed

examples/hello-xgboost/app_site1/config/config_fed_client.json
```diff
@@ -0,0 +1,33 @@
...
+          "train_data": "{data_root}higgs.train.csv.1?format=csv&label_column=0",
+          "test_data": "{data_root}higgs.test.csv.1?format=csv&label_column=0",
+          "num_rounds": 100,
+          "early_stopping_round": 2,
+          "xgboost_params": {
+            "max_depth": 8,
+            "eta": 0.1,
+            "objective": "binary:logistic",
+            "eval_metric": "auc",
+            "tree_method": "gpu_hist",
```
nvidianz, line 23: I think it's better to keep this as "hist". GPU shouldn't be required to run an example.

nvidianz (approved):

examples/hello-xgboost/app_site1/config/config_fed_client.json
YuanTingHsieh: let's add description in README to explain this.

PR #866, opened 2022-09-08
Fixed some log issues and a timing issue for CTR-C kill the simulator.

yhwen:
Fix issues:

1. Several log message syntax issues.
2. empty log message when error happened
3. log file consistent format
4. passing the parent pid to child process, instead of fetching from the child process


YuanTingHsieh (approved):
Thanks, LGTM

PR #867, opened 2022-09-08
Remove CORS because setting baseUrl to "" works

IsaacYangSLA:
Set download attachment filename to include quote


nvkevlu (dismissed):
LGTM, thanks

nvkevlu (approved):
thanks, LGTM

PR #869, opened 2022-09-08
Increase download_count for user and client whenever blob is downloaded

IsaacYangSLA:


nvkevlu (approved):
LGTM, thanks

nvkevlu (approved):
LGTM, thanks

PR #879, opened 2022-09-13
2.2_Branch: fix POC command bug [skip ci]

chesterxgchen:
Issue is caused by unpack_archive only to current directory not dest poc directory

### DIFF
--- a/nvflare/lighter/poc.py
+++ b/nvflare/lighter/poc.py
@@ -39,8 +39,8 @@ def unpack_poc(dest_poc_folder) -> bool:
         import tempfile
 
         with tempfile.TemporaryDirectory() as tmp_dir:
-            shutil.unpack_archive(poc_zip_path, extract_dir=tmp_dir.name)
-            copy_from_src(os.path.join(tmp_dir.name, "poc"), dest_poc_folder)
+            shutil.unpack_archive(poc_zip_path, extract_dir=tmp_dir)
+            copy_from_src(os.path.join(tmp_dir, "poc"), dest_poc_folder)
 
         return True
     except shutil.ReadError:
--- a/nvflare/lighter/poc.py
+++ b/nvflare/lighter/poc.py
@@ -32,11 +32,16 @@ def clone_client(num_clients: int, dest_poc_folder):
     shutil.rmtree(src_folder)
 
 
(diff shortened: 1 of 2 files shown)


nvflare/lighter/poc.py
```diff
@@ -32,11 +32,17 @@ def clone_client(num_clients: int, dest_poc_folder):
     shutil.rmtree(src_folder)
 
 
-def unpack_poc() -> bool:
+def unpack_poc(dest_poc_folder) -> bool:
     file_dir_path = pathlib.Path(__file__).parent.absolute()
     poc_zip_path = file_dir_path.parent / "poc.zip"
     try:
-        shutil.unpack_archive(poc_zip_path)
+        import tempfile
```
IsaacYangSLA, line 39: Suggest:

try:
    with tempfile.TemporaryDirectory() as tmp_dir:
        ....
except shutil.ReadError:


Current codes will skip temp_dir.cleanup() when shutil.ReadError happens.

nvflare/lighter/poc.py
```diff
@@ -32,11 +32,16 @@ def clone_client(num_clients: int, dest_poc_folder):
...
+
+        with tempfile.TemporaryDirectory() as tmp_dir:
+            shutil.unpack_archive(poc_zip_path, extract_dir=tmp_dir.name)
```
IsaacYangSLA, line 42: tmp_dir is the name of the temp directory, so it's a string.  I think you don't need tmp_dir.name.

IsaacYangSLA:
```
$ python3
Python 3.10.4 (main, Jun 29 2022, 12:14:53) [GCC 11.2.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import tempfile
>>> with tempfile.TemporaryDirectory() as tmp_dir:
...   print(tmp_dir)
...   print(tmp_dir.name)
... 
/tmp/tmpyk0leofm
  File "<stdin>", line 3, in <module>
AttributeError: 'str' object has no attribute 'name'
```

IsaacYangSLA (approved):
Thanks.  LGTM.

YuanTingHsieh (approved):

PR #885, opened 2022-09-13
accelerations for tree based federated xgboost

eordentlich:


nvflare/app_common/learners/xgboost_tree_fed_learner.py
```diff
@@ -101,6 +101,8 @@ def initialize(self, parts: dict, fl_ctx: FLContext):
         self.set_lr()
         # load data, this is task-specific
         self.load_data(fl_ctx)
+        self.bst = None
+        self.global_model_as_dict = None
```
YuanTingHsieh, line 105: do we need `self.tree_method = tree_method` in the init function?

YuanTingHsieh: I mean the argument of the function.

eordentlich: Yes.  The conflict resolution got a little messed up, but missed this one in the previous cleanup.

Can-Zhao:
Both cyclic and bagging work well and fast on my machine with cpu and gpu.

nvflare/apis/impl/controller.py
```diff
@@ -66,7 +66,7 @@ def _check_inputs(task: Task, fl_ctx: FLContext, targets: Union[List[Client], Li
 
 
 class Controller(Responder, ControllerSpec, ABC):
-    def __init__(self, task_check_period=0.5):
+    def __init__(self, task_check_period=0.01):
```
ZiyueXu77, line 69: will this change impact other pipelines / shall we change it for all? - may need comments from Chester / Yan

nvflare/app_common/workflows/scatter_and_gather.py
```diff
@@ -234,14 +234,14 @@ def control_flow(self, abort_signal: Signal, fl_ctx: FLContext) -> None:
 
                 self.fire_event(AppEventType.BEFORE_SHAREABLE_TO_LEARNABLE, fl_ctx)
                 self._global_weights = self.shareable_gen.shareable_to_learnable(aggr_result, fl_ctx)
-                fl_ctx.set_prop(AppConstants.GLOBAL_MODEL, self._global_weights, private=True, sticky=True)
+                fl_ctx.set_prop(AppConstants.GLOBAL_MODEL, self._global_weights, private=True, sticky=False)
```
ZiyueXu77, line 237: same here, scatter_and_gather is used much more often for other pipelines, not sure how much impact sticky has @YuanTingHsieh

nvflare/apis/impl/controller.py
eordentlich: This was temp change before this param could be set from config files.  Will revert to original default.

nvflare/app_common/workflows/scatter_and_gather.py
YuanTingHsieh: let merge this and I will do some testing on other examples to see if this is fine or not.

Thanks

YuanTingHsieh:
I will fix the style issues in a later PR, thanks for the algo. update

PR #890, opened 2022-09-14
Fed stats new logic [skip ci]

chesterxgchen:
1. If any feature of a client's dataset does not meet the min_count requirement, the feature will not be counted in, but we will count other features for this client
    2. If ALL feature of a client's dataset does not meet the min_count requirement, the client will not be counted in  ( this is consequence of #1) 

    3. if  for any given feature, min_client is not received,  The global aggregation for this feature will not computed, all client's local stats for this feature  will not be used. 

    4. if for all features, the min_clients are not received, the job is aborted.


holgerroth (approved):
LGTM

PR #891, opened 2022-09-14
Fix hello-xgboost in secure mode

YuanTingHsieh:
- Fix _get_certificates function return True in the end
- Fix get server address in executor

### DIFF
--- a/nvflare/app_opt/xgboost/executor_base.py
+++ b/nvflare/app_opt/xgboost/executor_base.py
@@ -15,6 +15,7 @@
 import os
 from abc import ABC, abstractmethod
 
+from nvflare.apis.event_type import EventType
 from nvflare.apis.executor import Executor
 from nvflare.apis.fl_constant import FLContextKey, ReturnCode
 from nvflare.apis.fl_context import FLContext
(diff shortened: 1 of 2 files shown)


nvflare/app_opt/xgboost/executor_base.py
```diff
@@ -81,12 +73,21 @@ def __init__(
...
 
     @abstractmethod
     def xgb_train(self, params: XGBoostParams, fl_ctx: FLContext) -> Shareable:
         """Main XGBoost training method"""
         pass
 
+    def handle_event(self, event_type: str, fl_ctx: FLContext):
+        if event_type == EventType.START_RUN:
+            engine = fl_ctx.get_engine()
+            sp = engine.client.overseer_agent.get_primary_sp()
```
nvidianz, line 86: Is overseer_agent required in non-HA mode?

nvidianz (approved):

nvflare/app_opt/xgboost/executor_base.py
YuanTingHsieh: It should be there for POC mode, let me check for simulator mode

nvidianz (approved):

PR #892, opened 2022-09-14
Fix app validator

YuanTingHsieh:
- Server app just need to have config_fed_server.json
- Client app just need to have config_fed_client.json

### DIFF
--- a/nvflare/private/fed/app/default_app_validator.py
+++ b/nvflare/private/fed/app/default_app_validator.py
@@ -16,18 +16,27 @@ import os
 from typing import Dict, Tuple
 
 from nvflare.apis.app_validation import AppValidationKey, AppValidator
+from nvflare.apis.fl_constant import SiteType, WorkspaceConstants
 
 
 class DefaultAppValidator(AppValidator):
+    def __init__(self, site_type: str, config_folder_inside_app="config"):
+        self._site_type = site_type
+        self._config_folder = config_folder_inside_app
+
     def validate(self, app_folder: str) -> Tuple[str, Dict]:
         result = dict()
         app_root = os.path.abspath(app_folder)
-        if not os.path.exists(os.path.join(app_root, "config")):
+        if not os.path.exists(os.path.join(app_root, self._config_folder)):
             return "Missing config folder inside app folder.", {}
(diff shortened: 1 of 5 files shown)


nvflare/private/fed/app/default_app_validator.py
```diff
@@ -16,18 +16,27 @@
...
 class DefaultAppValidator(AppValidator):
+    def __init__(self, site_type: str, config_folder_inside_app="config"):
+        self._site_type = site_type
+        self._config_folder = config_folder_inside_app
+
     def validate(self, app_folder: str) -> Tuple[str, Dict]:
         result = dict()
         app_root = os.path.abspath(app_folder)
-        if not os.path.exists(os.path.join(app_root, "config")):
+        if not os.path.exists(os.path.join(app_root, self._config_folder)):
```
yanchengnv, line 30: It is not good to repeat these path manipulation code. We should consolidate to use the Workspace object. But this will require changes to the signature of the AppValidator to take a Workspace object and a job id.  Let's not do it at this time.

I don't think it is necessary to make the config folder name configurable. On the contrary, we should standardize everything.

yanchengnv (approved):
Looks good. We can improve it more later.

nvflare/private/fed/app/default_app_validator.py
```diff
@@ -16,18 +16,27 @@
 from typing import Dict, Tuple
 
 from nvflare.apis.app_validation import AppValidationKey, AppValidator
+from nvflare.apis.fl_constant import SiteType, WorkspaceConstants
 
 
 class DefaultAppValidator(AppValidator):
+    def __init__(self, site_type: str, config_folder_inside_app="config"):
```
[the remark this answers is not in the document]
YuanTingHsieh: thanks let me change it

nvflare/private/fed/app/default_app_validator.py
YuanTingHsieh: I think there was a reason saying someone don't have the config folder or they want to put the config in the app root directly.

But I agree, We should just enforce that structure in NVFlare.

PR #902, opened 2022-09-20
Update admin api

YuanTingHsieh:
- Update AdminAPI


nvkevlu (approved):
Thanks for cleaning up the logic, LGTM

PR #914, opened 2022-09-22
Fixes several problems found during authz testing [skip ci]

nvidianz:
1. Fixed clone_job, not copying all fields.
2. Fixed the deploy error when one site rejects the job.
3. Admin API errors are logged now, instead of causing FOBS error
4. Fixed Workspace decomposer problem caused by name->site_name change.


nvflare/private/fed/server/server_engine.py
```diff
@@ -726,11 +726,16 @@ def check_client_resources(self, resource_reqs) -> Dict[str, Tuple[bool, str]]:
         for r in replies:
             site_name = self.get_client_name_from_token(r.client_token)
             if r.reply:
-                resp = fobs.loads(r.reply.body)
-                result[site_name] = (
-                    resp.get_header(ShareableHeader.CHECK_RESOURCE_RESULT, False),
-                    resp.get_header(ShareableHeader.RESOURCE_RESERVE_TOKEN, ""),
-                )
+                error_code = r.reply.get_header(MsgHeader.RETURN_CODE, ReturnCode.OK)
```
yanchengnv, line 729: This is good fix. There is a similar case (Line 584) in aux_send(). Please fix it.

yanchengnv (changes requested):
Looks really good. Just need one more fix for aux_send.

nvflare/private/fed/server/job_cmds.py
```diff
@@ -344,12 +355,15 @@ def clone_job(self, conn: Connection, args: List[str]):
             with engine.new_context() as fl_ctx:
                 data_bytes = job_def_manager.get_content(job_id, fl_ctx)
 
+                job_meta = {str(k): job.meta[k] for k in job.meta.keys() & CLONED_META_KEYS}
+
```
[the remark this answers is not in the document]
nvidianz: Yep. & is to get the intersection of 2 sets. When cloning jobs, we don't want to copy things like deploy_details.

nvflare/private/fed/server/job_runner.py
```diff
@@ -338,10 +338,16 @@ def run(self, fl_ctx: FLContext):
...
+                                else:
+                                    deployable_clients = client_sites
+
                                 self._start_run(
                                     job_id=job_id,
                                     job=ready_job,
-                                    client_sites=client_sites,
+                                    client_sites=deployable_clients,
                                     fl_ctx=fl_ctx,
                                 )
```
YuanTingHsieh, line 352: Maybe we need to check if a job is deployable at scheduler time.

Cause we need to also check if require_clients is in deployable_clients
and if deployable_clients are larger than min_clients

YuanTingHsieh: got it, make sense to me, agree to keep this for now.

nvflare/private/fed/server/server_engine.py
nvidianz: That error is also serialized so it doesn't actually cause a FOBS error. I added logging when error reply is received.

YuanTingHsieh (approved):
thanks, LGTM

yhwen (dismissed):

nvflare/private/fed/server/server_engine.py
```diff
@@ -581,12 +581,16 @@ def aux_send(self, targets: [], topic: str, request: Shareable, timeout: float,
             client_name = self.get_client_name_from_token(r.client_token)
             if r.reply:
                 try:
-                    results[client_name] = fobs.loads(r.reply.body)
-                except BaseException:
+                    shareable = fobs.loads(r.reply.body)
```
yanchengnv, line 584: We should do exactly the same thing as line 733: check return_code first. Only if the RC is OK, should we do fobs.loads().

yanchengnv (changes requested):

nvflare/private/fed/server/server_engine.py
nvidianz: Done.

yanchengnv (approved):

PR #923, opened 2022-09-23
README.rst for app_opt/xgboost module [skip ci]

nvidianz:


nvkevlu:
rst files should be in the docs structure in order to get built properly, so this will have to be moved. We still need to reorganize the docs anyway so we can put this anywhere now and move this later.

nvflare/app_opt/xgboost/README.rst
```diff
@@ -0,0 +1,267 @@
...
+          "site-1b"
+        ],
+        "app_site2_gpu0": [
+          "site-2a"
+        ],
+        "app_site2_gpu1": [
+          "site-2b"
+        ]
+      },
+      "min_clients": 4
```
YuanTingHsieh, line 218: i need to check if this is "min_clients" or "mandatory_clients"

YuanTingHsieh (approved):
good explanation for the app_opt/xgboost code.
We can do unify documentation in later PRs.

nvflare/app_opt/xgboost/README.rst
YuanTingHsieh: should be "mandatory_clients"

PR #924, opened 2022-09-23
Fix mgpu executor

yhwen:
Fixed the PTMultiProcessExecutor error. Create the fobs_initialize() for all the processes.

### DIFF
--- a/nvflare/apis/utils/decomposers/flare_decomposers.py
+++ b/nvflare/apis/utils/decomposers/flare_decomposers.py
@@ -111,7 +111,7 @@ class WorkspaceDecomposer(Decomposer):
         return Workspace
 
     def decompose(self, target: Workspace) -> Any:
-        return [target.root_dir, target.name, target.config_folder]
+        return [target.root_dir, target.site_name, target.config_folder]
 
     def recompose(self, data: Any) -> Workspace:
         return Workspace(data[0], data[1], data[2])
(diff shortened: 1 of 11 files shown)


yanchengnv (approved):
LGTM

nvflare/private/fed/app/client/worker_process.py
```diff
@@ -93,6 +93,7 @@ def main():
     if os.path.exists(restart_file):
         os.remove(restart_file)
 
+    fobs_initialize()
```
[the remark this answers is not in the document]
yhwen: No, this is not only for the check_parent_alive(). This is the client app running process. It needs to initialize the fobs.

YuanTingHsieh (approved):
Thanks LGTM

nvflare/apis/utils/decomposers/flare_decomposers.py
```diff
@@ -111,7 +111,7 @@ def supported_type():
         return Workspace
 
     def decompose(self, target: Workspace) -> Any:
-        return [target.root_dir, target.name, target.config_folder]
+        return [target.root_dir, target.site_name, target.config_folder]
```
YuanTingHsieh, line 114: Note this change is the same as in #914

yhwen: that's fine. PR merge will take care of this.

PR #933, opened 2022-09-24
convert to simulator with latest performance updates [skip ci]

ZiyueXu77:
Update readme with:
- simulator mode
- new result curves after the method update
Update the tensorboard curve script to comply with simulator mode workspaces


YuanTingHsieh (approved):
Thanks, LGTM

PR #941, opened 2022-09-26
Fix broken mprocess executor

yhwen:
. Fixed the broken multi-gpu executor
. Enable the multi-gpu executor with simulator run.

### DIFF
--- a/nvflare/app_common/executors/multi_process_executor.py
+++ b/nvflare/app_common/executors/multi_process_executor.py
@@ -33,6 +33,7 @@ from nvflare.fuel.common.multi_process_executor_constants import CommunicateData
 from nvflare.fuel.utils.class_utils import ModuleScanner
 from nvflare.fuel.utils.component_builder import ComponentBuilder
 from nvflare.fuel.utils.network_utils import get_open_ports
+from nvflare.private.fed.simulator.simulator_app_runner import SimulatorClientRunManager
 
 
 class WorkerComponentBuilder(ComponentBuilder):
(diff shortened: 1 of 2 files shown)


nvflare/app_common/executors/multi_process_executor.py
```diff
@@ -33,6 +33,7 @@
 from nvflare.fuel.utils.class_utils import ModuleScanner
 from nvflare.fuel.utils.component_builder import ComponentBuilder
 from nvflare.fuel.utils.network_utils import get_open_ports
+from nvflare.private.fed.simulator.simulator_app_runner import SimulatorClientRunManager
```
YuanTingHsieh, line 36: I think code from app_common can't depend on private?
We could set field in fl_ctx to avoid this import.

nvflare/app_common/executors/multi_process_executor.py
```diff
@@ -145,14 +146,29 @@ def _initialize_multi_process(self, fl_ctx: FLContext):
...
             self.open_ports = get_open_ports(self.num_of_processes * 3)
-
+            client_name = fl_ctx.get_identity_name()
+            job_id = fl_ctx.get_job_id()
+
+            engine = fl_ctx.get_engine()
+            if isinstance(engine, SimulatorClientRunManager):
+                simulator_engine = True
+            else:
+                simulator_engine = False
```
YuanTingHsieh, line 156: we should use a field/key in fl_ctx instead of getting the engine and use class name

nvflare/app_common/executors/multi_process_executor.py
```diff
@@ -145,14 +146,29 @@ def _initialize_multi_process(self, fl_ctx: FLContext):
...
             command = (
                 self.get_multi_process_command()
                 + " -m nvflare.private.fed.app.client.sub_worker_process"
```
YuanTingHsieh, line 159: But then this line is arguable.
If this "nvflare/app_common/executors/multi_process_executor.py" rely on "nvflare.private.fed.app.client.sub_worker_process", we should put this class in private?

nvflare/app_common/executors/multi_process_executor.py
yhwen: changed the way to determine simulator mode. removed this import.

nvflare/app_common/executors/multi_process_executor.py
yhwen: changed the way to determine simulator mode. removed this import.

nvflare/app_common/executors/multi_process_executor.py
yhwen: changed the way to determine simulator mode. removed this import.

nvflare/private/fed/app/client/sub_worker_process.py
```diff
@@ -161,15 +171,26 @@ def main():
...
-    run_manager = ClientRunManager(
-        client_name=args.client_name,
-        job_id=job_id,
-        workspace=workspace,
-        client=None,
-        components=data[CommunicationMetaData.COMPONENTS],
-        handlers=data[CommunicationMetaData.HANDLERS],
-        conf=None,
-    )
+    if args.simulator_engine.lower() == "true":
```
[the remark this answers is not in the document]
yhwen: SimulatorClientRunManager is already the subclass of ClientRunManager. The root issue is that the sub_worker_process is called from the component MultiProcessExecutor from the app, which stays the same regardless running in simulator mode or the real-world application mode.

PR #957, opened 2022-09-28
rename metric, metrics to static, statics for codes and doc [skip ci]

chesterxgchen:
rename metric, metrics to static, statics for codes and doc.


kkersten (approved):
Looks good.

YuanTingHsieh (approved):

PR #962, opened 2022-09-28
Fix simulator package issue [skip ci]

YuanTingHsieh:
- Move "log.config" to simulator module level
- Include "log.config" in wheel
- Change hardcoded values inside simulator to use pre-defined constants


yhwen:
This PR moved the log.config from “simulator/resource” folder into the “simulator” folder directly. It’s better to separate the non Python codes into a separate “resource” folder, not mixing the config files and Python source codes together.

YuanTingHsieh:
If we put log.config in the simulator/resource folder, we will need to add an __init__.py there to make the packager to pick up this log.config file, that is not good.

And to the point of "mixing the config files and Python source codes together."
If the log.config is inside the nvflare root folder, we are already mixing configs with Python source codes.

PR #971, opened 2022-09-30
Simulator max clients

yhwen:
Add the option to set max_clients for running simulator.


tests/unit_test/private/fed/app/deployer/simulator_deployer_test.py
```diff
@@ -42,6 +42,7 @@ def _create_parser(self):
         parser.add_argument("--clients", "-n", type=int, help="number of clients", required=True)
         parser.add_argument("--threads", "-t", type=int, help="number of running threads", required=True)
         parser.add_argument("--gpu", "-gpu", type=str, help="list of GPUs")
+        parser.add_argument("-m", "--max_clients", type=int, default=100, help="max number of clients")
 
         parser.add_argument("--set", metavar="KEY=VALUE", nargs="*")
```
YuanTingHsieh, line 48: instead of having this _create_parser in test, it better just import and use the one from 
`nvflare/private/fed/app/simulator/simulator.py` ?

nvflare/private/fed/app/simulator/simulator.py
```diff
@@ -28,6 +28,7 @@ def define_simulator_parser(simulator_parser):
     simulator_parser.add_argument("-c", "--clients", type=str, help="client names list")
     simulator_parser.add_argument("-t", "--threads", type=int, help="number of parallel running clients")
     simulator_parser.add_argument("-gpu", "--gpu", type=str, help="list of GPU Device Ids, comma separated")
+    simulator_parser.add_argument("-m", "--max_clients", type=int, default=100, help="max number of clients")
```
YuanTingHsieh, line 31: We might want to change this "threads" argument name?
Suggest to `n_parallel_clients`

tests/unit_test/private/fed/app/deployer/simulator_deployer_test.py
yhwen: done

nvflare/private/fed/app/simulator/simulator.py
yhwen: Simulator already have been officially released. Should not change this to break the user facing argument name.

yhwen:
> Although I approved the PR, I think YT's proposal is good one, using thread, implies the your use thread implementation, when you change the implementation this becomes hard to explain, especially in Muti-GPU cases, where each thread actually start a new process.
> 
> calling the number of parallel clients can avoid such confusion

Consulted this with Holger. He prefers to keep the "--threads" as it is from the user perspective.

PR #987, opened 2022-10-05
Moved the custom folder into local folder.

yhwen:


YuanTingHsieh:
Do we need to change the corresponding "start.sh"/"sub_start.sh"? @IsaacYangSLA

yhwen:
> Do we need to change the corresponding "start.sh"/"sub_start.sh"? @IsaacYangSLA

No need to change.

PR #989, opened 2022-10-05
Fix server engine abort_app_on_server

YuanTingHsieh:
it should return empty string when success

### DIFF
--- a/nvflare/private/fed/server/server_engine.py
+++ b/nvflare/private/fed/server/server_engine.py
@@ -344,7 +344,6 @@ class ServerEngine(ServerEngineInternalSpec):
 
         self.logger.info("Abort the server app run.")
 
-        status_message = ""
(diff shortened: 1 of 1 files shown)


nvflare/private/fed/server/server_engine.py
```diff
@@ -362,7 +361,7 @@ def abort_app_on_server(self, job_id: str) -> str:
                 self.run_processes.pop(job_id)
```
yhwen, line 362: This needs to be changed to:

        finally:
            with self.lock:
                if job_id in self.run_processes:
                    self.run_processes.pop(job_id)

YuanTingHsieh: thanks, fixed.

yhwen (approved):

PR #992, opened 2022-10-05
more additions and updates to the docs [skip ci]

nvkevlu:


kkersten (commented):
We had discussed removing the Appendix from the main docs page in favor of posting changelog and any breaking changes on the github.  This makes it easier to stay current with point releases.

The cert creation instructions are a holdover from very early versions.  I think we can get rid of this since we do it in the cert builder as a reference example.

I vote for removing docs/appendix/ and docs/appendix.rst altogether.  Appendix was already dropped from the main toc.

nvkevlu:
Sounds good to me, I have removed the Appendix sections from the docs.

kkersten (approved):
LGTM

PR #996, opened 2022-10-07
Switch to XGBoost Communicator API

rongou:
In https://github.com/dmlc/xgboost/pull/8257 the `Rabit` API was switched to `Communicator`.


YuanTingHsieh (approved):
Thanks, LGTM

PR #999, opened 2022-10-08
Clean up authz

YuanTingHsieh:
We only have "role" now.


yhwen (approved):
LGTM

nvflare/fuel/sec/authz.py
```diff
@@ -22,35 +23,27 @@
...
+        if not isinstance(user, Person):
+            raise ValueError(f"user need to be of type Person but got {type(user)}")
+        if submitter and not isinstance(submitter, Person):
+            raise ValueError(f"submitter need to be of type Person but got {type(submitter)}")
         self.right = right
         self.user = user
         self.submitter = submitter
         self.attrs = {}
-        if not submitter:
+        if submitter is None:
```
nvidianz, line 46: The original one is better. It can handle empty string.

YuanTingHsieh: I added the checks in L40-41, so we should be good.

PR #1001, opened 2022-10-08
Fix info collect commands

YuanTingHsieh:
- Fix `get_run_info`, `get_app_run_info`, `show_stats`, `get_errors` and related info collector commands
- Add check of `if server_runner` in `AbortCommand`

### DIFF
--- a/nvflare/private/fed/server/fed_server.py
+++ b/nvflare/private/fed/server/fed_server.py
@@ -15,11 +15,10 @@
 import logging
 import os
 import shutil
-import threading
 import time
 from abc import ABC, abstractmethod
 from concurrent import futures
-from threading import Lock
+from threading import Lock, Thread
 from typing import List, Optional
 
 import grpc
@@ -101,6 +100,7 @@ class BaseServer(ABC):
 
         self.grpc_server = None
         self.admin_server = None
+        self.lock = Lock()
         self.snapshot_lock = Lock()
         self.fl_ctx = FLContext()
         self.platform = None
@@ -111,7 +111,7 @@ class BaseServer(ABC):
         self.abort_signal = None
         self.executor = None
 
-        self.lock = threading.Lock()
+        self.logger = logging.getLogger(self.__class__.__name__)
 
     def get_all_clients(self):
         return self.client_manager.get_clients()
@@ -184,9 +184,7 @@ class BaseServer(ABC):
             self.logger.info("starting insecure server at %s", target)
         self.grpc_server.start()
 
-        # return self.start()
-        cleanup_thread = threading.Thread(target=self.client_cleanup)
-        # heartbeat_thread.daemon = True
+        cleanup_thread = Thread(target=self.client_cleanup)
         cleanup_thread.start()
 
     def client_cleanup(self):
@@ -247,10 +245,7 @@ class FederatedServer(BaseServer, fed_service.FederatedTrainingServicer, admin_s
             handlers: A list of handler
             args: arguments
             secure_train: whether to use secure communication
-            enable_byoc: whether to enable custom components
         """
-        self.logger = logging.getLogger("FederatedServer" + args.name)
-
         BaseServer.__init__(
             self,
             project_name=project_name,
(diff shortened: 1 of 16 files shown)


nvflare/private/fed/server/fed_server.py
```diff
@@ -403,14 +397,12 @@ def GetTask(self, request, context):
             self.logger.debug(f"Fetch task requested from client: {client.name} ({client.get_token()})")
             token = client.get_token()
 
-            # engine = fl_ctx.get_engine()
             shared_fl_ctx = fobs.loads(proto_to_bytes(request.context["fl_context"]))
             job_id = str(shared_fl_ctx.get_prop(FLContextKey.CURRENT_RUN))
-            # fl_ctx.set_peer_context(shared_fl_ctx)
+            run_processes = self.engine.get_run_processes()
```
yhwen, line 402: Move this run_processes = self.engine.get_run_processes() outside the "with self.lock:"  will loose the protection of the engine run_processes change. As the run_processes could be changed other job running thread.

nvflare/private/fed/server/fed_server.py
```diff
@@ -492,12 +484,13 @@ def SubmitUpdate(self, request, context):
                 response_comment = "Ignored the submit from invalid client. "
                 self.logger.info(response_comment)
             else:
+                run_processes = self.engine.get_run_processes()
```
yhwen, line 487: same issue as the self.engine.get_run_processes() outside the "with self.lock:"

nvflare/private/fed/server/server_engine.py
```diff
@@ -253,14 +254,24 @@ def wait_for_complete(self, job_id):
...
+                        run_process_info = self.run_processes.pop(job_id)
+                        return_code = run_process_info[RunProcessKey.CHILD_PROCESS].poll()
+                        # if process exit but with Execution exception
+                        if return_code and return_code != 0:
+                            self.execution_exception_run_processes[job_id] = run_process_info
                 self.engine_info.status = MachineStatus.STOPPED
                 break
 
+    def get_run_processes(self) -> dict:
+        with self.lock:
```
yhwen, line 267: lock on this single statement does not do much protection.

nvflare/private/fed/server/server_engine.py
```diff
@@ -253,14 +254,24 @@ def wait_for_complete(self, job_id):
...
+            result = self.run_processes
+        return result
+
+    def get_execution_exception_run_processes(self) -> dict:
+        with self.lock:
```
yhwen, line 272: same here.

nvflare/private/fed/server/server_engine.py
```diff
@@ -149,11 +147,14 @@ def get_engine_info(self) -> EngineInfo:
 
         return self.engine_info
 
-    def get_run_info(self) -> RunInfo:
+    def get_run_info(self) -> dict:
```
yhwen, line 150: Better to return a well defined RunInfo object, instead of free dict here, even though the RunInfo does not contain too much info yet.

yhwen (commented):
I am not clear what the issue this PR is trying to address. The change potentially has big impacts especially for some race conditions. Need to be very careful.

nvflare/private/fed/server/server_commands.py
```diff
@@ -174,16 +166,17 @@ def process(self, data: Shareable, fl_ctx: FLContext):
...
 
-        shareable = data.get(ReservedKey.SHAREABLE)
-        shared_fl_ctx = data.get(ReservedKey.SHARED_FL_CONTEXT)
-        client = shareable.get_header(ServerCommandKey.FL_CLIENT)
-        fl_ctx.set_peer_context(shared_fl_ctx)
-        contribution_task_name = shareable.get_header(ServerCommandKey.TASK_NAME)
-        task_id = shareable.get_cookie(FLContextKey.TASK_ID)
         server_runner = fl_ctx.get_prop(FLContextKey.RUNNER)
-        server_runner.process_submission(client, contribution_task_name, task_id, shareable, fl_ctx)
+        if server_runner:
```
yhwen, line 170: What issue is this trying to address? We should not have this kind of scenario. If the server_runner has not been created, the client won't be able to get task to run and submit result. 
If the run already completed, the server parent process should not send request to here.

PR #1008, opened 2022-10-11
Fix shutdown command when authz failed

YuanTingHsieh:
- When the clients auth failed, we should not call `engine.remove_clients`


nvflare/private/fed/server/training_cmds.py
```diff
@@ -119,10 +120,16 @@ def _shutdown_app_on_clients(self, conn: Connection) -> bool:
...
+        clients_to_be_removed = set(clients)
+        for r in replies:
+            if r.reply.get_header(MsgHeader.RETURN_CODE) == ReturnCode.ERROR:
+                clients_to_be_removed.remove(r.client_token)
+
+        if clients_to_be_removed:
+            err = engine.remove_clients(clients_to_be_removed)
+            if err:
+                conn.append_error(err)
+                return False
```
yhwen, line 132: This logic is not correct. when clients_to_be_removed != clients, it means some clients responded error. It needs to return False. 
In the extreme case, all the clients have error return code, the clients_to_be_removed will be empty. With the current logic, it will return True.

YuanTingHsieh: yes, thanks for the review, I just update this PR.

PR #1011, opened 2022-10-12
Fixed the ha_admin_cmds.

yhwen:


YuanTingHsieh (approved):
LGTM for now, we should think of other ways to get the overseer agent functionality later.

instead of getting it like `ctx.get_api().service_finder.overseer_agent`

PR #1012, opened 2022-10-12
Changed the default max_jobs to 4.

yhwen:


YuanTingHsieh:
I am good with whatever numbers here, we might need to document this default number and where to change it in the documentation so users would know, @nvkevlu do we have this section right now?

nvkevlu:
We don't really have a section on this but I can add something about this.

PR #1014, opened 2022-10-12
Use randomly generated string as SECRET_KEY if users do not set it

IsaacYangSLA:
Truncate project's short name to 16 chars if its length exceeds 16.


nvkevlu (approved):
Thanks, LGTM

PR #1021, opened 2022-10-13
set the peer_ctx to private.

yhwen:


yhwen:
> what does private here mean ?

Private means only keep locally, not going to share with remote side (server / client)

yhwen:
> @yanchengnv can probably review this. seems ok to me, but I am not sure the impact.

The same fix has been in 2.2.0 release. https://github.com/NVIDIA/NVFlare/pull/880

yanchengnv (approved):
LGTM

PR #1023, opened 2022-10-13
Enhance job run status

yhwen:
Added update_job_run_status to update the proper job run status.

Description:
At the end of ServerAppRunner application job run, it will check if the job run has FLContextKey.FATAL_SYSTEM_ERROR been reported. If it does, then post the job execution exception status to the parent process to update the job FINISHED:FINISHED_EXECUTION_EXCEPTION outcome. Otherwise, the job outcome will be FINISHED:COMPLETED.

system panic and server job exception will mark as FLContextKey.FATAL_SYSTEM_ERROR.

### DIFF
--- a/nvflare/private/fed/server/job_runner.py
+++ b/nvflare/private/fed/server/job_runner.py
@@ -285,10 +285,10 @@ class JobRunner(FLComponent):
                     with self.lock:
                         job = self.running_jobs.get(job_id)
                         if job:
-                            execution_exception_run_processes = engine.execution_exception_run_processes
-                            if job_id in execution_exception_run_processes:
+                            exception_run_processes = engine.exception_run_processes
+                            if job_id in exception_run_processes:
                                 self.log_info(fl_ctx, f"Try to abort run ({job_id}) on clients.")
-                                run_process = execution_exception_run_processes[job_id]
+                                run_process = exception_run_processes[job_id]
 
                                 # stop client run
                                 client_sites = run_process.get(RunProcessKey.PARTICIPANTS)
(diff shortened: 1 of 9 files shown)


nvflare/private/fed/server/server_engine.py
```diff
@@ -254,11 +266,7 @@ def wait_for_complete(self, job_id):
             except BaseException:
                 with self.lock:
                     if job_id in self.run_processes:
-                        run_process_info = self.run_processes.pop(job_id)
-                        return_code = run_process_info[RunProcessKey.CHILD_PROCESS].poll()
-                        # if process exit but with Execution exception
-                        if return_code and return_code != 0:
-                            self.execution_exception_run_processes[job_id] = run_process_info
+                        self.run_processes.pop(job_id)
                 self.engine_info.status = MachineStatus.STOPPED
```
YuanTingHsieh, line 270: The original logic covers the situation where an exception happens in the "finally" block in `nvflare/private/fed/server/server_app_runner.py`

Also, runner_process will throw exception as well: `https://github.com/NVIDIA/NVFlare/blob/dev/nvflare/private/fed/app/server/runner_process.py#L129`

This new logic does not account for those situations.

yhwen: This already covered in ServerAppRunner.start_server_app() finally block.  No need for the old logic any more.

nvflare/private/fed/server/job_runner.py
```diff
@@ -301,6 +301,7 @@ def _job_complete_process(self, fl_ctx: FLContext):
                             fl_ctx.set_prop(FLContextKey.CURRENT_JOB_ID, job.job_id)
                             self.fire_event(EventType.JOB_COMPLETED, fl_ctx)
                             self.log_debug(fl_ctx, f"Finished running job:{job.job_id}")
+                    engine.remove_exception_process(job_id)
```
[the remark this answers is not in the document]
yhwen: Once job outcome has been updated, no need to keep the job's execution error info any more.

nvflare/private/fed/server/server_engine.py
yhwen: rolled back the wait_for_complete logic.

PR #1024, opened 2022-10-13
fix typing, fix issue of abort_task error

nvkevlu:


nvflare/private/fed/server/cmd_utils.py
```diff
@@ -51,6 +51,10 @@ def authorize_client_operation(self, conn: Connection, args: List[str]) -> PreAu
         return PreAuthzReturnCode.REQUIRE_AUTHZ
 
     def authorize_abort_client_operation(self, conn: Connection, args: List[str]) -> PreAuthzReturnCode:
+        if len(args) < 2:
+            conn.append_error("missing job_id (syntax is: abort_task job_id <client-name>)")
+            return PreAuthzReturnCode.ERROR
+
```
YuanTingHsieh, line 57: If this function only designed for "abort_task" of client, then this error message is good, but I suggest to change the name of the function to `authorize_abort_client_task` or something.

If this function is generic purpose, then this error message is not appropriate in here.

nvkevlu: renamed the function to authorize_abort_client_task

YuanTingHsieh (approved):
Thanks, LGTM

PR #1036, opened 2022-10-16
prelight_check update [skip ci]

chesterxgchen:
1) update preflight_check how to fix message based TME suggestion
2) remove package_root option and replace page with package_path ( check one package at a time) 
3) move stop_dry_run to the very end based on return code, as currently it will kill the process before the report is printout 
4) stop dry_run use pkill -9 command which always print out "Killed" at console. to avoid confusion, add print statement, "print dry run process" 
5) check_dry_run() give the wrong reports, change the logic

### DIFF
--- a/nvflare/tool/package_checker/check_rule.py
+++ b/nvflare/tool/package_checker/check_rule.py
@@ -86,16 +86,16 @@ class CheckOverseerRunning(CheckRule):
         if err:
             return CheckResult(
                 f"Can't connect to overseer ({overseer_agent_args['overseer_end_point']}): {err}",
-                "1) Please check if overseer is up or certificates are correct." +
-                "2) Please check if overseer hostname in project.yml is available." +
-                "3) if running in local machine, check if overseer defined in project.yml is defined in /etc/hosts"
+                "1) Please check if overseer is up or certificates are correct."
+                + "2) Please check if overseer hostname in project.yml is available."
+                + "3) if running in local machine, check if overseer defined in project.yml is defined in /etc/hosts",
             )
         elif not check_response(resp):
             return CheckResult(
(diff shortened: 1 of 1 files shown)


nvflare/tool/package_checker/package_checker.py
```diff
@@ -51,12 +51,19 @@ def should_be_checked(self) -> bool:
...
 
-    def stop_dry_run(self):
-        process = run_command_in_subprocess(f"pkill -9 -f '{self.package_path}'")
+    def stop_dry_run(self, force: bool = False):
+        # todo: add gracefully shutdown command
+        print("killing dry run process")
+        if force:
+            cmd = f"pkill -9 -f '{self.package_path}'"
+        else:
+            cmd = f"pkill -9 -f '{self.package_path}'"
```
YuanTingHsieh, line 60: these 2 commands are the same, is this intended?

nvflare/tool/package_checker/package_checker.py
```diff
@@ -77,35 +84,50 @@ def check(self):
...
-
+                ret_code = self.check_dry_run(timeout=self.dry_run_timeout)
+            return ret_code
         except Exception as e:
             self.add_report(
                 "Package Error",
                 f"Exception happens in checking: {e}, this package is not in correct format.",
                 "Please download a new package.",
             )
+            return ret_code
```
YuanTingHsieh, line 95: suggest return True/False instead of `ret_code`
Since the typehint in method signature suggest "bool"

YuanTingHsieh: Or we need to change the type hint to return "int"

YuanTingHsieh (approved):
Thanks, LGTM

PR #1038, opened 2022-10-16
Check nvidia-smi exists before using it

chesterxgchen:
avoid user confusion with failed to find nvidia-smi message
1) check if the nvidia-smi before using it, avoid file not exception
2) remove output "Failed to get gpu device IDs, assume no gpu device." message as it seems to be confuse user.

### DIFF
--- a/nvflare/fuel/utils/gpu_utils.py
+++ b/nvflare/fuel/utils/gpu_utils.py
@@ -38,9 +38,7 @@ def use_nvidia_smi(query: str, report_format: str = "csv"):
 
 def _parse_gpu_mem(result: str = None, unit: str = "MiB") -> List:
     gpu_memory = []
-    if not result:
-        print("Failed to get gpu memory, assume no gpu device.")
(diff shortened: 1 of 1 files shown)


nvflare/fuel/utils/gpu_utils.py
```diff
@@ -28,8 +33,6 @@ def use_nvidia_smi(query: str, report_format: str = "csv"):
             raise Exception(f"Failed to call nvidia-smi with query {query}", result.stderr)
         else:
             return result.stdout.splitlines()
-    except FileNotFoundError as e:
-        print(f"Failed to call nvidia-smi: {e}")
     return None
```
YuanTingHsieh, line 38: _parse_gpu_mem also print this "failed" msg we can either remove in this Pr or the later one

YuanTingHsieh (dismissed):

YuanTingHsieh (approved):
LGTM

PR #1040, opened 2022-10-16
Display proper not client authorized message [gpu]

yhwen:


nvflare/private/fed/server/sys_cmd.py
```diff
@@ -32,10 +32,13 @@ def _parse_replies(conn, replies):
...
-                resources = json.loads(r.reply.body)
-            except BaseException as e:
-                resources = f"Bad replies: {secure_format_exception(e)}"
+            if isinstance(r.reply.body, str) and r.reply.body:
+                resources = r.reply.body
+            else:
+                try:
+                    resources = json.loads(r.reply.body)
+                except BaseException as e:
+                    resources = f"Bad replies: {secure_format_exception(e)}"
```
YuanTingHsieh, line 41: The new code's behavior is diff. VS the old code.

Consider the case where `r.reply.body = '{"hello": "motor"}' `

The old code resources going to be a dict `{"hello": "motor"}`
The new code resources will be a string `'{"hello": "motor"}' `

Is this what we want?

nvflare/private/fed/server/training_cmds.py
```diff
@@ -316,17 +319,22 @@ def _process_scope_replies(self, table, conn, replies):
...
-                    else:
-                        conn.append_error(f"bad response from client {client_name}: expect dict but got {type(body)}")
-                except BaseException as e:
-                    self.logger.error(f"Bad reply from client: {secure_format_exception(e)}")
-                    conn.append_error(f"bad response from client {client_name}: {secure_format_exception(e)}")
+                if isinstance(r.reply.body, str) and r.reply.body:
+                    self._add_scope_info(table, client_name, r.reply.body, "")
+                else:
+                    try:
+                        body = json.loads(r.reply.body)
```
YuanTingHsieh, line 326: Same situation here as commented above,

Consider the case where r.reply.body = '{"hello": "motor"}'

The old code body going to be a dict {"hello": "motor"}
The new code body will be a string '{"hello": "motor"}'

nvflare/private/fed/server/sys_cmd.py
yhwen: Intended to handle the client authorization error case. Should not simply check the "str" response body type. Changed to use the proper error returnCode.

nvflare/private/fed/server/training_cmds.py
yhwen: Same as above.

YuanTingHsieh (approved):
LGTM

PR #1057, opened 2022-10-18
Fix sqlalchemy version to avoid flask-sqlalchemy __all__ issue.

IsaacYangSLA:


nvkevlu (approved):
Thanks, LGTM

setup.py
```diff
@@ -77,6 +77,7 @@
         "Flask==2.1.2",
         "Flask-JWT-Extended==4.4.3",
         "Flask-SQLAlchemy==2.5.1",
+        "SQLAlchemy==1.4.31",
```
YuanTingHsieh, line 80: not sure if we need to update the "requirements-dev.txt"

PR #1058, opened 2022-10-18
Updated getting started guide to use FL Simulator [skip ci]

kkersten:
Updated getting started guide to use FL Simulator


kkersten:
> This looks very good. The only things left are the link to the container toolkit installation and maybe adding a sample docker run or docker exec command for users to run the docker container that they just built so we cover all the VDR feedback.
> 
> We can either merge this first then do that in another PR, or if it is easy for you to push one more commit with those changes on top of this PR, we can merge it with those additions here.

Good call.  I will add this in another commit before we merge.

nvkevlu (approved):
Thank you, this looks great

PR #1091, opened 2022-10-25
Fixed the abort_job HA failed to remove snapshot issue.

yhwen:


nvflare/private/fed/server/server_engine.py
```diff
@@ -370,13 +370,18 @@ def abort_app_on_server(self, job_id: str) -> str:
...
-                if job_id in self.run_processes:
-                    self.run_processes.pop(job_id)
+            threading.Thread(target=self._remove_run_processes, args=[job_id]).start()
 
         self.engine_info.status = MachineStatus.STOPPED
         return ""
 
+    def _remove_run_processes(self, job_id):
+        # wait for the run process to gracefully terminated, and ensure to remove from run_processes.
+        time.sleep(60.0)
```
YuanTingHsieh, line 380: Note that this does not guarantee the run process has gracefully terminated though.

We just sleep 60 seconds and assume that.

nvflare/private/fed/server/server_commands.py
```diff
@@ -81,7 +88,12 @@ def process(self, data: Shareable, fl_ctx: FLContext):
         if server_runner:
             server_runner.abort(fl_ctx)
             # wait for the runner process gracefully abort the run.
-            time.sleep(3.0)
+            engine = fl_ctx.get_engine()
+            start_time = time.time()
+            while engine.engine_info.status != MachineStatus.STOPPED:
+                time.sleep(1.0)
+                if time.time() - start_time > 30.0:
+                    break
```
YuanTingHsieh, line 96: If this period is longer than 30 seconds, does this mean something is wrong?

Should we return a warning message?

nvflare/private/fed/server/server_engine.py
yhwen: correct. The abort command is sent. We can just wait 60 seconds assuming it shutdown self. Then terminate it any way.

nvflare/private/fed/server/server_commands.py
yhwen: abort_job command only needs to respond the command been delivered to the child process. No need to send back warning message.

PR #1095, opened 2022-10-26
Move s3_storage [skip ci]

YuanTingHsieh:
move file s3_storage.py from app_common to app_opt


YuanTingHsieh:
@chesterxgchen OK let me change it

PR #1097, opened 2022-10-26
Remove ByeCommand

YuanTingHsieh:
This command is not used anymore


YuanTingHsieh:
I think this command is not used, @yhwen can you confirm?

YuanTingHsieh:
We are not using these 2 commands anywhere now.
I personally prefer to remove them.
But I will leave them there for now.

PR #1103, opened 2022-11-01
FOBS redesign with enum and data class support

nvidianz:
1. Added generic decomposer for dict sub-classes, like Shareable, Learnable.
2. Moved doc from Data Class from README to code to treat this feature as internal.
3. Removed all decomposers for enum classes since Enums are handled by FOBS automatically.


nvflare/fuel/utils/fobs/fobs.py
```diff
@@ -55,22 +61,59 @@ def register(decomposer: Union[Decomposer, Type[Decomposer]]) -> None:
...
+    """
+
+    for enum_type in enum_types:
+        if not issubclass(enum_type, Enum):
+            raise TypeError(f"Can't register class {enum_type}, which is not a subclass of Enum")
+        decomposer = EnumTypeDecomposer(enum_type)
+        register(decomposer)
+
+
+def auto_register_enum_types(enabled=True) -> None:
```
yhwen, line 106: Why do we need to have this _enum_auto_register True or False? Instead of always True?

nvidianz: This flag is just a safety measure so we can go back to the old behavior with a single switch. We may remove this flag in a later release.

nvflare/fuel/utils/fobs/decomposer.py
```diff
@@ -58,3 +58,40 @@ def recompose(self, data: Any) -> T:
             The reconstructed object
         """
         pass
+
+
+class DataClassDecomposer(Decomposer):
+    """Generic decomposers for data classes"""
```
YuanTingHsieh, line 64: might be good to explain more or add an example of what is classified as a DataClass

YuanTingHsieh: I see in README you add it, I think it might be good to also put some of those in docstring

YuanTingHsieh:
LGTM in general

yhwen (dismissed):

nvflare/fuel/utils/fobs/README.rst
```diff
@@ -154,6 +156,53 @@ by MessagePack, a decomposer is included in `fobs` module so no need to further
...
+auto-registering the generic decomposer for this enum type.
+
+The auto-registering of enum decomposers can be disabled like this,
+
+::
+
+    fobs.auto_register_enum_types(False)
+
+
+Data Classes
```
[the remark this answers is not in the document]
nvidianz: Made data class an internal concept and removed the instructions from README.

nvflare/fuel/utils/fobs/decomposer.py
nvidianz: Moved the doc from README to code so this feature is treated as internal.

PR #1104, opened 2022-11-01
Update build and package steps

YuanTingHsieh:
- Move install require from setup.py to setup.cfg
- Use setup.cfg as default for requirements.txt
- Add build_system for pyproject.toml


IsaacYangSLA (approved):
Thanks.  This is a good improvement over our legacy codes.

PR #1142, opened 2022-11-15
Added warnings for vulnerable OpenSSL versions

nvidianz:


nvflare/fuel/sec/cve_checker.py
```diff
@@ -0,0 +1,84 @@
...
+        return (3, 0, 0) <= triple <= (3, 0, 6)
+
+    # OpenSSL 1.1.1 and 1.0.2 are not affected by this issue.
+    if triple[0] == 1:
+        return triple != (1, 1, 1) and triple != (1, 0, 2)
+
+    return False
+
+
+def warn_openssl(version: str = ssl.OPENSSL_VERSION):
```
yhwen, line 63: Instead of parsing the ssl.OPENSSL_VERSION string, it's better to use the ssl.OPENSSL_VERSION_INFO, which is a openssl version tuple.

nvidianz: It's hard to run tests with that tuple so I decided to use the string. Also, the tuple is not well-documented, don't really know what the 5 numbers mean.

yhwen (approved):

PR #1145, opened 2022-11-15
Create quickstart page for outside links, fix provisioning section [skip ci]

nvkevlu:


docs/real_world_fl/overview.rst
```diff
@@ -70,42 +70,9 @@ name in the workspace folder created where provision.py is run.
    signed by :class:`SignatureBuilder<nvflare.lighter.impl.signature.SignatureBuilder>` so the system will detect if any
    of the files have been altered and may not run.
```
holgerroth, line 72: Where are the instructions now on how to generate password protected startup kits?

nvkevlu: I will need to add a section for that

docs/programming_guide/provisioning_system.rst
```diff
@@ -283,6 +283,62 @@ A new builder to write 'gateway.conf' can be implemented as follows (for referen
                   f.write(f"name = {gw.name}\n")
                   f.write(f"port = {port}\n")
 
+.. _distribution_builder:
+
+Case 4: adding a builder for enabling the creation of zip archives for the startup kits
+---------------------------------------------------------------------------------------
+DistributionBuilder was included in NVIDIA FLARE before version 2.2.1 but has been removed from the
+default builders. You can add it as a builder in project.yml if you want to zip the startup kits::
```
holgerroth, line 291: Why copy the class code here? Would be better to show how it looks in the .yml file.

nvkevlu: Since this code has been removed in 2.2.1, the user would have to include it in order to make things work. I added an example for what needs to be included in the .yml file too.

holgerroth: I see.

holgerroth (approved):

PR #1164, opened 2022-11-28
Fixed a race condition and client privacy.json issue.

yhwen:
. Fixed a server parent process race condition.
. Fixed a client privacy.json component scan issue.


yanchengnv (approved):
LGTM

PR #1169, opened 2022-11-30
Fix signing jobs in POC

IsaacYangSLA:
This fix is for handling job signing in POC mode.  Basically, all jobs should not be signed and should not be verified in POC as there are no private keys/certificates.

### DIFF
--- a/nvflare/fuel/hci/client/file_transfer.py
+++ b/nvflare/fuel/hci/client/file_transfer.py
@@ -30,7 +30,7 @@ from nvflare.fuel.utils.zip_utils import split_path, unzip_all_from_bytes, zip_d
 from nvflare.lighter.utils import load_private_key_file, sign_folders
 from nvflare.security.logging import secure_format_exception, secure_log_traceback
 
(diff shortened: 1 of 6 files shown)


nvflare/fuel/hci/client/file_transfer.py
```diff
@@ -325,9 +325,11 @@ def upload_folder(self, args, ctx: CommandContext):
 
         # sign folders and files
         api = ctx.get_api()
-        client_key_file_path = api.client_key
-        private_key = load_private_key_file(client_key_file_path)
-        sign_folders(full_path, private_key, api.client_cert)
+        if api.poc_key != "admin":
```
yhwen, line 328: Better to extract "admin" to a constant, since it could be changed in AdminAPI.

also extra line 341 need to be removed:
        api = ctx.get_api()

IsaacYangSLA: Thanks.  Fixed in the update.

yhwen (approved):
looks good.

PR #1196, opened 2022-12-22
fix simulator missing client heartbeat call.

yhwen:
fix simulator missing client heartbeat call for 2.2 branch.


yhwen:
> @yhwen 2.2 branch is for 2.2.0. I think the 2.2.x (x > 0) is on a tag or different branch @IsaacYangSLA

We confirmed that the 2.2 branch has been changed for 2.2.3 release already.

PR #1204, opened 2022-12-31
MLFlowReceiver 

chesterxgchen:
NVFLARE's Tensorboard Streaming feature allows data scientists to record the metrics from different clients to server and displayed via Tensorboard. Although this is very useful, data scientists may choose to use other experimentation tracking tools such as MLFlow, Weights&Biases in some cases. 
The goal of this project is to support MLFLow as alternative experimental tools. This make it easier to integrate with Azure cloud where the MLFLow is default tracking tool.. 
 

Requirements
Allow existing code that is written for Tensorboard experiment tracking be directly used for MLFLow without rewrite with MLFlow APIs.  Allow support both MLFlow and tensorflow streaming
Similarly, the code is written for MLFLow APIs, should be able to switch to other tracking receiver such as Tensorboard
This is Part 1 of the PR, in including the mainly the code changes to support MLFLow Receiver with Tensorboard Sender 
1) Updated AnalyticsDataType to include MLFLow Data Types
2) Updated AnalyticsSender to includes sender (Tracker) metadata as well as broader data types
3) MLFLowReciever 
4) hello-pt-tb-mlflow examples ( only with configuration change) with notebook

### DIFF
--- a/nvflare/app_common/widgets/streaming.py
+++ b/nvflare/app_common/widgets/streaming.py
@@ -169,9 +169,6 @@ class AnalyticsSender(Widget):
         if self.engine:
             self.engine = None
 
-    def get_initial_kwargs(self):
-        return {}
-
 
 class AnalyticsReceiver(Widget, ABC):
     def __init__(self, events: Optional[List[str]] = None):
--- a/nvflare/fuel/utils/json_scanner.py
+++ b/nvflare/fuel/utils/json_scanner.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import copy
+import logging
 import traceback
 from abc import ABC, abstractmethod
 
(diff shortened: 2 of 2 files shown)


nvflare/app_common/widgets/streaming.py
```diff
@@ -157,6 +169,9 @@ def close(self):
         if self.engine:
             self.engine = None
 
+    def get_initial_kwargs(self):
```
yhwen, line 172: Why do we need this?

nvflare/fuel/utils/json_scanner.py
```diff
@@ -95,6 +96,9 @@ def _do_scan(self, node: Node):
         try:
             node.processor.process_element(node)
         except BaseException as e:
+            if not is_secure():
+                print(traceback.format_exc())
```
yhwen, line 100: Should we change this to use logger?

nvflare/fuel/utils/json_scanner.py
```diff
@@ -90,11 +92,15 @@ def __init__(self, json_data: dict, location=None):
...
         self.location = location
         self.data = json_data
+        self.logger = logging.getLogger("JsonScanner")
 
     def _do_scan(self, node: Node):
         try:
             node.processor.process_element(node)
         except BaseException as e:
+            if not is_secure():
+                self.logger.error(traceback.format_exc())
```
yhwen, line 102: I think you don't need to do a logging here. Simply raise ConfigError here, then the Configer will handle the error logging already.

yhwen: We discussed the purpose of this. Keep this implementation.

yhwen (approved):

PR #1207, opened 2023-01-03
PSI initial PR

chesterxgchen:
**PSI Initial PR**
are under the following assumptions 
1) PR is only works in Linux env. it doesn't work for Mac at the moment
2) PR will has failure when there is certain ratio of items size of PSI server and client ( underline openmined psi issue) 

The PSI is designed to have the following components and Alogrithms. 
**Algorithms:** 
  * Pair-wise Client interaction (via server) are used leverage 2-party client PSI
  * Both Forward pass and Backpass are leverage parallel executions (see design for details)

**System design:** 

**Client Side:** 

PSIExecutor –  FLARE Executor for PSI
ClientExecutor – FLARE Interface to handle real client execution ( invoke method ClientExectuor.client_exec()) 
DhPSIExecutor – an Implementation of ClientExector Interface for openmined PSI
PSI – a user facing FLComponent, PSIExecutor will need to delegate to user to implement. 
PsiClient –  wrapper for the client part of two-party PSI algorithm
PsiServer – wrapper for the server part of the two-party PSI algorithm

**Server side:** 
   PSIController – FLARE Controller for PSI 
   PSIWorkflow – PSI Workflow, one for each PSI Algorithm

Although one can configure the PSI server side operation via PSIController and PSIWorkflow, we added 
```
class DhPSIController(PSIController)
```
to simply the configuration 

**Example,** 
We added an basic example, other examples might be added later

**Client-Server Interaction sequence diagram**
https://github.com/chesterxgchen/NVFlare/blob/psi2/nvflare/app_common/psi/README.md

### DIFF
--- a/nvflare/app_common/executors/psi/__init__.py
+++ b/nvflare/app_common/executors/psi/__init__.py
@@ -1,0 +1,13 @@
+# Copyright (c) 2021-2022, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
--- a/nvflare/app_common/psi/__init__.py
+++ b/nvflare/app_common/psi/__init__.py
@@ -1,0 +1,13 @@
+# Copyright (c) 2021-2022, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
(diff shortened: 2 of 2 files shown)


nvflare/app_common/app_constant.py
```diff
@@ -191,3 +191,34 @@ class StatisticsConstants(AppConstants):
     }
 
     PRE_RUN_RESULT = "fed_stats_pre_run_result"
+
+
+class PSIConst(AppConstants):
+    PSI_TASK_KEY = "PSI_TASK_KEY"
```
holgerroth, line 197: We might not need the `PSI_` prefix as the class name already specifies "PSI".

nvflare/app_common/psi/psi_workflow_spec.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class PSIWorkflow(InitFinalArgsComponent, ABC):
+    """
+    PSIWorkflow is an interface for different PSI algorithms
+    for example, DDH-Based PSI, Homomorphic-based PSI etc.
+    """
+
+    @abstractmethod
+    def pre_workflow(self, abort_signal: Signal) -> bool:
```
holgerroth, line 27: Consider calling these methods "preprocess," "run," and "postprocess".

IsaacYangSLA (changes requested):
LGTM.  However, when we build the wheel, the build process ignores any folder that has no __init__.py file.  I believe some of the new folders in this PR do not have that file.  Can you help add them?

IsaacYangSLA (approved):
LGTM.

PR #1236, opened 2023-01-20
Only terminate the local session in the worker_process, not logout th…

yhwen:
Only terminate the local session in the worker_process, not logout the client.


YuanTingHsieh (approved):
Thanks, LGTM

PR #1250, opened 2023-01-25
Revert flake8 and isort version

YuanTingHsieh:
- Our jenkins CI docker container has to be upgraded, then we can advance these versions


IsaacYangSLA (approved):
Agree.  We should keep CI as stable as possible.  It's better not to advance dependencies.

PR #1286, opened 2023-02-02
Create pull_request_template.md [skip ci]

YuanTingHsieh:
- Add a template for PR so it is more descriptive and make people's life easier


IsaacYangSLA (approved):
Thanks, @YuanTingHsieh .
Not sure if the first line (Fixes # .) always applies.  I think it's only for item 1 and 2 unless we require all PR must have an issue filed first.
Other than that, it's good.

PR #1295, opened 2023-02-03
Re-organize example folder structure

YuanTingHsieh:
Re-organize the example structure to be:

- hello-world
- advanced
- tutorial

Not touch cifar10 in this PR just because it has some pending PRs


kkersten:
This looks like a very good first pass.  I vote for merging this and iterating from here.

Looks like the `hello-monai-bundle/README.md` is going away, instead linking directly to the example in `../integration/monai/README.md`.  Seems like the right move, and we can get rid of the top-level `hello-monai-bundle` dir.

Where should we keep the `hello-pt*` examples? The `hello-pt` example makes sense in `basic`, and `hello-pt-tb-mlflow` probably makes sense as a tutorial (when ready).  What about `hello-pt-tb`?  also `basic` or does it merge with `hello-pt-tb-mlflow`?

holgerroth:
Why not call it `hello_world` instead of `basic`?

YuanTingHsieh:
@kkersten good question, I will move `hello-pt-tb` into advanced

kkersten:
Agree with @holgerroth on `hello_world` rather than `basic`.

holgerroth (approved):

PR #1296, opened 2023-02-03
Downgrade protobuf version for compatibility

YuanTingHsieh:
- As pointed out in https://github.com/NVIDIA/NVFlare/discussions/772, the latest Tensorflow and tensorboard wants protobuf < 3.20


YuanTingHsieh:
will do this in #1370

PR #1314, opened 2023-02-07
Update documentation

### LINKED ISSUE #1182: [BUG] Command "nvflare" not found
Unfortunately I can't run nvflare on Ubuntu because it tells me that the command (`nvflare simulator -h`) is not found. Even if I use the Docker solution. Inside the docker it also shows "command not found". Thus unable to run local experiments.

I tried the MWE from https://nvflare.readthedocs.io/en/2.2.1/getting_started.html

Used OS: Ubuntu 18.04.5 LTS

YuanTingHsieh:
Fixes #1182 .

### Description

- Add a Dockerfile: docs/resources/Dockerfile.doc
- Add "copy" button to code-blocks
- Fix documentation issues found when building docs
- Fix example code link follow #1295 

### Types of changes

- [x] Documentation updated.

### DIFF
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -1,0 +1,5 @@
+##########
+Quickstart
+##########
+
+See :ref:`getting_started`.
--- a/docs/quickstart.rst
+++ b/docs/quickstart.rst
@@ -1,5 +1,0 @@
-##########
-Quickstart
-##########
-
-See :ref:`quickstart`.
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -76,6 +76,7 @@ extensions = [
     "sphinx.ext.autodoc",
     "sphinx.ext.viewcode",
     "sphinx.ext.autosectionlabel",
+    "sphinx_copybutton",
 ]
 
 autoclass_content = "both"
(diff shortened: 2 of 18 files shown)


docs/quickstart.rst
```diff
@@ -1,5 +0,0 @@
-##########
-Quickstart
-##########
-
-See :ref:`quickstart`.
```
nvkevlu, line 5: Maybe this could be: See :ref:`getting_started`.

This page is here because there were pages that linked to quickstart from outside the docs here, and it went to a broken page. If we keep this and have a basic redirect, users that land here can find their way to the getting started guide.

nvkevlu:
Just the one comment on the deletion of the quickstart page, but otherwise looks great to me. Thanks!

docs/quickstart.rst
YuanTingHsieh: @nvkevlu got it, in that case, maybe we should update those documentation to point to getting started page?

nvkevlu: The problem is some of those were published and linked outside of our control, so on our end, if someone does arrive at the page, it is better to exist and have a link pointing to something rather than the page does not exist error.

YuanTingHsieh: got it, let me add it back.

nvkevlu (approved):
Thanks, LGTM

PR #1344, opened 2023-02-11
Add support for AWS cloud launch scripts (dashboard/server/client)

IsaacYangSLA:
### Description

This PR enables generating AWS cloud scripts for dashboard, server and client.
The command for all of them is similar to azure, just replace azure with aws on the --cloud option.
For example, to launch a server in AWS, type ./start.sh --cloud aws

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/dashboard/application/__init__.py
+++ b/nvflare/dashboard/application/__init__.py
@@ -23,6 +23,7 @@ jwt = JWTManager()
 
 
 def init_app():
+    os.makedirs("/var/tmp/nvflare/dashboard", exist_ok=True)
     static_folder = os.environ.get("NVFL_DASHBOARD_STATIC_FOLDER", "static")
     app = Flask(__name__, static_url_path="", static_folder=static_folder)
     app.config.from_object("nvflare.dashboard.config.Config")
@@ -41,4 +42,6 @@ def init_app():
             email = credential.split(":")[0]
             pwd = credential.split(":")[1]
             Store.seed_user(email, pwd)
+    with open(os.path.join("/var/tmp/nvflare/dashboard", ".db_init_done"), "ab") as f:
+        f.write(bytes())
     return app
--- a/nvflare/dashboard/application/blob.py
+++ b/nvflare/dashboard/application/blob.py
@@ -154,6 +154,12 @@ def gen_server(key, first_server=True):
                 "t",
                 exe=True,
             )
+            _write(
+                os.path.join(dest_dir, get_csp_start_script_name("aws")),
+                utils.sh_replace(get_csp_template("aws", "svr", template), {"server_name": entity.name}),
+                "t",
+                exe=True,
+            )
         signatures = utils.sign_all(dest_dir, deserialize_ca_key(project.root_key))
         json.dump(signatures, open(os.path.join(dest_dir, "signature.json"), "wt"))
 
@@ -256,6 +262,12 @@ def gen_client(key, id):
         _write(
             os.path.join(dest_dir, get_csp_start_script_name("azure")),
             get_csp_template("azure", "cln", template),
+            "t",
+            exe=True,
+        )
+        _write(
+            os.path.join(dest_dir, get_csp_start_script_name("aws")),
+            get_csp_template("aws", "cln", template),
             "t",
             exe=True,
         )
--- a/nvflare/dashboard/cli.py
+++ b/nvflare/dashboard/cli.py
@@ -43,7 +43,7 @@ def start(args):
         environment["NVFL_DASHBOARD_PP"] = passphrase
     if args.cred:
         environment.update({"NVFL_CREDENTIAL": args.cred})
-    elif not os.path.exists(os.path.join(folder, "db.sqlite")):
+    elif not os.path.exists(os.path.join(folder, ".db_init_done")):
         need_email = True
         while need_email:
             answer = input(
@@ -121,9 +121,6 @@ def cloud(args):
     template = utils.load_yaml(os.path.join(lighter_folder, "impl", "master_template.yml"))
     cwd = os.getcwd()
     csp = args.cloud
-    if csp not in supported_csp:
-        print(f"Currently support either azure or aws, but {csp} is requested.")
-        exit(1)
     dest = os.path.join(cwd, f"{csp}_start_dsb.sh")
     _write(
         dest,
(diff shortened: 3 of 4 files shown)


nvflare/dashboard/application/blob.py
```diff
@@ -259,6 +265,12 @@ def gen_client(key, id):
             "t",
             exe=True,
         )
+        _write(
```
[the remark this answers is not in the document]
IsaacYangSLA: This is to generate the actual startup kits, so just lots of files to download.  For now, it seems a good match, but we can find other name in the future.

nvflare/dashboard/cli.py
```diff
@@ -117,7 +121,10 @@ def cloud(args):
     template = utils.load_yaml(os.path.join(lighter_folder, "impl", "master_template.yml"))
     cwd = os.getcwd()
     csp = args.cloud
-    dest = os.path.join(cwd, f"{csp}_start.sh")
+    if csp not in supported_csp:
+        print(f"Currently support either azure or aws, but {csp} is requested.")
```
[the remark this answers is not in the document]
IsaacYangSLA: This is another protection.  The csp is already checked in handle_dashboard.  If csp is not in supported, the cloud function will not be called.

nvflare/dashboard/cli.py
```diff
@@ -157,18 +167,18 @@ def define_dashboard_parser(parser):
         "--passphrase", help="Passphrase to encrypt/decrypt root CA private key.  !!! Do not share it with others. !!!"
     )
     parser.add_argument("-e", "--env", action="append", help="additonal environment variables: var1=value1")
+    parser.add_argument("--cred", help="set credential directly in the form of USER_EMAIL:PASSWORD")
```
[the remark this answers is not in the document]
IsaacYangSLA: This is for proj_admin, not for AWS.

nvflare/lighter/impl/master_template.yml
```diff
@@ -397,8 +397,11 @@ start_cln_sh: |
       azure)
         $DIR/azure_start.sh ${all_arguments}
         ;;
+      aws)
+        $DIR/aws_start.sh ${all_arguments}
+        ;;
       *)
-        echo "Only on-prem or Azure is currently supported."
+        echo "Only on-prem or azure or aws is currently supported."
```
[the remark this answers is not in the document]
IsaacYangSLA: This section tied to the above case, which covers azure and aws.  Changing this to supported_csp also means the above case statement will be dynamic.  I think it seems not quite logic.

nvflare/lighter/impl/master_template.yml
```diff
@@ -436,8 +439,11 @@ start_svr_sh: |
...
```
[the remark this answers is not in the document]
IsaacYangSLA: This section tied to the above case, which covers azure and aws.  Changing this to supported_csp also means the above case statement will be dynamic.  I think it seems not quite logic.

nvflare/lighter/impl/master_template.yml
```diff
@@ -1551,3 +1557,399 @@ adm_notebook: |
   "nbformat": 4,
   "nbformat_minor": 5
   }
+
+aws_start_svr_sh: |
```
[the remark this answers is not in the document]
IsaacYangSLA: Sure, I will put the refactoring as one task in the future.

nvflare/dashboard/application/blob.py
```diff
@@ -154,6 +154,12 @@ def gen_server(key, first_server=True):
                 "t",
                 exe=True,
             )
+            _write(
+                os.path.join(dest_dir, get_csp_start_script_name("aws")),
+                utils.sh_replace(get_csp_template("aws", "svr", template), {"server_name": entity.name}),
+                "t",
+                exe=True,
+            )
```
YuanTingHsieh, line 162: I notice we put this piece of code under "if not project.ha_mode".
So it means the cloud deployment needs to be done in POC mode?

YuanTingHsieh: Or how does it work?

nvflare/lighter/impl/master_template.yml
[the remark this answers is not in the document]
YuanTingHsieh: I like the re-factoring idea as well.
It would be better to have smaller files maybe organized in directories instead of having them all in one place.

nvflare/dashboard/application/blob.py
IsaacYangSLA: In cloud deployment, users should use what cloud service provider supports for high availability.  Therefore, we don't want to interfere with those features.

YuanTingHsieh: got it, thanks for the explanation

nvflare/dashboard/cli.py
```diff
@@ -157,18 +168,18 @@ def define_dashboard_parser(parser):
...
-    if has_no_arguments():
-        print("Add -h option to see usage")
-        exit(0)
     if args.stop:
         stop()
     elif args.start:
         start(args)
-    elif args.cloud == "azure":
+    elif args.cloud in supported_csp:
         cloud(args)
```
YuanTingHsieh, line 180: I notice we already check "args.cloud in supported_csp" here,
in that case should we move the warning message here?

nvflare/dashboard/application/__init__.py
```diff
@@ -23,6 +23,7 @@
 
 
 def init_app():
+    os.makedirs("/var/tmp/nvflare/dashboard", exist_ok=True)
```
YuanTingHsieh, line 26: For this temp folder, why we put in "/var/tmp" but not "/tmp"?
What is the pros and cons?

If so, should we also change other "/tmp/nvflare" paths to "/var/tmp/nvflare"?

nvflare/dashboard/cli.py
IsaacYangSLA: Good suggestion.  Let me move that warning here.

nvflare/dashboard/application/__init__.py
IsaacYangSLA: /var/tmp folder is not removed after reboot while /tmp is.  Because the db file should not be erased after reboot, it's better inside /var/tmp/.

YuanTingHsieh: got it, thanks for the explanation

PR #1347, opened 2023-02-13
Add warnings to ModuleScanner

YuanTingHsieh:
### Description

If a file requires an optional dependencies, then during our configuration process, the ModuleScanner should not go silent.

- It would be good to inform the users what package are missing if they want to use that file.
- Also remove unused methods
- filter out deprecation package so it won't print unnecessary messages.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/fuel/utils/class_utils.py
+++ b/nvflare/fuel/utils/class_utils.py
@@ -20,6 +20,8 @@ from typing import List
 
 from nvflare.security.logging import secure_format_exception
 
+DEPRECATED_PACKAGES = ["nvflare.app_common.pt"]
+
 
 def get_class(class_path):
     module_name, class_name = class_path.rsplit(".", 1)
@@ -81,6 +83,8 @@ class ModuleScanner:
 
             for module_info in pkgutil.walk_packages(path=package.__path__, prefix=package.__name__ + "."):
                 module_name = module_info.name
+                if any(module_name.startswith(deprecated_package) for deprecated_package in DEPRECATED_PACKAGES):
+                    continue
                 if module_name.startswith(base):
                     if not self.exclude_libs or (".libs" not in module_name):
                         if any(module_name.startswith(base + "." + name + ".") for name in self.module_names):
--- a/nvflare/fuel/utils/class_utils.py
+++ b/nvflare/fuel/utils/class_utils.py
@@ -14,6 +14,7 @@
 
 import importlib
 import inspect
+import logging
 import pkgutil
 from typing import List
 
@@ -57,17 +58,6 @@ def instantiate_class(class_path, init_params):
     return instance
 
 
-def get_object_method(obj, method_name):
-    op = getattr(obj, method_name, None)
-    if op is None or not callable(op):
-        return None
-    return op
-
-
-def get_instance_method(instance, method_name):
-    return get_object_method(instance, method_name)
-
-
 class ModuleScanner:
     def __init__(self, base_pkgs: List[str], module_names: List[str], exclude_libs=True):
         """Scanner to look for and load specified module names.
@@ -80,6 +70,8 @@ class ModuleScanner:
         self.base_pkgs = base_pkgs
         self.module_names = module_names
         self.exclude_libs = exclude_libs
+
+        self._logger = logging.getLogger(self.__class__.__name__)
         self._class_table = {}
         self._create_classes_table()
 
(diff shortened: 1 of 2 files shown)


nvflare/fuel/utils/class_utils.py
```diff
@@ -80,24 +70,34 @@ def __init__(self, base_pkgs: List[str], module_names: List[str], exclude_libs=T
...
     def _create_classes_table(self):
         for base in self.base_pkgs:
             package = importlib.import_module(base)
 
-            for importer, modname, ispkg in pkgutil.walk_packages(path=package.__path__, prefix=package.__name__ + "."):
-
-                if modname.startswith(base):
-                    if not self.exclude_libs or (".libs" not in modname):
-                        if any(modname.startswith(base + "." + name + ".") for name in self.module_names):
+            for module_info in pkgutil.walk_packages(path=package.__path__, prefix=package.__name__ + "."):
```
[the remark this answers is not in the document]
YuanTingHsieh: Following the documentation: https://docs.python.org/3/library/pkgutil.html#pkgutil.walk_packages

It returns `ModuleInfo` class.

And `ModuleInfo` class contains of (module_finder, name, ispkg): https://docs.python.org/3/library/pkgutil.html#pkgutil.ModuleInfo

I think we don't assume `ModuleInfo` is always a 3 items tuple is better.

nvflare/fuel/utils/class_utils.py
```diff
@@ -80,24 +70,34 @@ def __init__(self, base_pkgs: List[str], module_names: List[str], exclude_libs=T
...
+                                    if (
+                                        not name.startswith("_")
+                                        and inspect.isclass(obj)
+                                        and obj.__module__ == module_name
+                                    ):
+                                        self._class_table[name] = module_name
+                            except (ModuleNotFoundError, RuntimeError) as e:
+                                self._logger.warning(
+                                    f"Try to import module {module_name}, but failed: {e}. "
+                                    f"Please ignore this if you are not using that file."
```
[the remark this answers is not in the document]
YuanTingHsieh: Thanks, let me improve this.

nvflare/fuel/utils/class_utils.py
```diff
@@ -80,24 +72,36 @@ def __init__(self, base_pkgs: List[str], module_names: List[str], exclude_libs=T
...
             package = importlib.import_module(base)
 
-            for importer, modname, ispkg in pkgutil.walk_packages(path=package.__path__, prefix=package.__name__ + "."):
-
-                if modname.startswith(base):
-                    if not self.exclude_libs or (".libs" not in modname):
-                        if any(modname.startswith(base + "." + name + ".") for name in self.module_names):
+            for module_info in pkgutil.walk_packages(path=package.__path__, prefix=package.__name__ + "."):
+                module_name = module_info.name
+                if any(module_name.startswith(deprecated_package) for deprecated_package in DEPRECATED_PACKAGES):
```
yhwen, line 86: Why added the DEPRECATED_PACKAGES? This will result component initialization could not find the module. For example:  PTFileModelLocator.

nvflare/fuel/utils/class_utils.py
```diff
@@ -80,24 +72,36 @@ def __init__(self, base_pkgs: List[str], module_names: List[str], exclude_libs=T
...
-                                        self._class_table[name] = modname
-                            except (ModuleNotFoundError, RuntimeError):
+                                    if (
+                                        not name.startswith("_")
+                                        and inspect.isclass(obj)
+                                        and obj.__module__ == module_name
+                                    ):
+                                        self._class_table[name] = module_name
+                            except (ModuleNotFoundError, RuntimeError) as e:
+                                self._logger.warning(
```
yhwen, line 101: This will result to a lot warning messages when starting the server, client, and also the admin. This may not be the behaviors we want.

yhwen (commented):
We should run the system testing, with the server / client / applications with the code changes.

nvflare/fuel/utils/class_utils.py
YuanTingHsieh: Because we have moved pt module from app common to app opt in PR #1348

nvflare/fuel/utils/class_utils.py
YuanTingHsieh: It will only print warnings about the package that can't be imported, which should not be in app common in the first place. 

Like HE, PSI.

I will move them to app opt as well.

PR #1356, opened 2023-02-14
FCI TCP Driver [skip ci]

nvidianz:
Merged in TCP driver for FCI (Flare Communication Interface)

### Description

1. Added TCP Driver
2. Added unit tests for DriveManager and Communicator

### Types of changes

- [ ] Documentation updated.


nvflare/fuel/f3/drivers/base_driver.py
```diff
@@ -0,0 +1,66 @@
...
+import threading
+from typing import Dict, Optional
+
+from nvflare.fuel.f3.connection import Connection, ConnState
+from nvflare.fuel.f3.drivers.driver import ConnectorInfo, Driver
+
+log = logging.getLogger(__name__)
+
+
+# noinspection PyAbstractClass
```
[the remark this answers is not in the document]
nvidianz: To get rid of PyCharm warning for non-implemented Abstract methods.

nvflare/fuel/f3/drivers/base_driver.py
```diff
@@ -0,0 +1,66 @@
...
+        conn.state = ConnState.CLOSED
+        self._notify_monitor(conn)
+
+        with self.conn_lock:
+            if not self.connections.pop(conn.name, None):
+                log.debug(f"{conn.name} is already removed from driver")
+
+    def close_all(self):
+        with self.conn_lock:
+            for name in sorted(self.connections.keys()):
```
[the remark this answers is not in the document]
nvidianz: Just easier to debug if connections are all closed in the same order.

nvflare/fuel/f3/drivers/base_driver.py
```diff
@@ -0,0 +1,66 @@
...
+                conn = self.connections[name]
+                log.debug(f"Closing connection: {self.get_name()}:{conn}")
+                conn.close()
+
+    def _notify_monitor(self, conn: Connection):
+        if not self.conn_monitor:
+            log.error(f"Connection monitor not registered for driver {self.get_name()}")
```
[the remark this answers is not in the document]
nvidianz: Should be error. It's a programming error not setting up a monitor.

nvflare/fuel/f3/drivers/base_driver.py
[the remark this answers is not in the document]
YuanTingHsieh: Why do we want to get rid of that warning? 
We should implement all the abstract methods in derived classes

nvflare/fuel/f3/drivers/socket_conn.py
```diff
@@ -0,0 +1,170 @@
...
+        self.closing = False
+        self.conn_props = self._get_socket_properties()
+
+    def get_conn_properties(self) -> dict:
+        return self.conn_props
+
+    def close(self):
+        self.closing = True
+        if self.sock:
+            self.sock.close()
```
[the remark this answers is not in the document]
nvidianz: The ConnManager in SFM layer handles the exception so driver doesn't have to worry about it.

nvflare/fuel/f3/drivers/base_driver.py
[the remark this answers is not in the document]
nvidianz: The BaseDriver is still an abstract class shared by all drivers. Python doesn't allow ABC to be used in subclasses so Pycharm gives warnings.

nvflare/fuel/f3/drivers/socket_conn.py
```diff
@@ -0,0 +1,170 @@
...
+            log.info(f"Connection {self.name} is closed by peer")
+        except BaseException as ex:
+            if self.closing:
+                log.debug(f"Connection {self.name} is closed")
+            else:
+                log.error(f"Connection {self.name} is closed due to error: {ex}")
+
+    def read_frame_loop(self):
+
+        while not self.closing:
```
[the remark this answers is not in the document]
nvidianz: It's meant to go forever as long as connection is up. When connection is closed, the read will throw exception and it will get out of the loop.

nvflare/fuel/f3/drivers/socket_conn.py
```diff
@@ -0,0 +1,170 @@
...
+
+        fileno = 0
+        try:
+            peer = self.sock.getpeername()
+            if isinstance(peer, tuple):
+                peer_addr = f"{peer[0]}:{peer[1]}"
+            else:
+                fileno = self.sock.fileno()
+                peer_addr = f"{peer}:{fileno}"
+        except OSError as ex:
```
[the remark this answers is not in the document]
nvidianz: This is to handle a MacOS bug. It throws Invalid Argument error but everything works fine so it should be ignored. Peer address is for information only.

nvflare/fuel/f3/drivers/socket_conn.py
```diff
@@ -0,0 +1,170 @@
...
+            peer_addr = "N/A"
+            log.debug(f"getpeername() error: {ex}")
+
+        conn_props[DriverParams.PEER_ADDR.value] = peer_addr
+
+        local = self.sock.getsockname()
+        if isinstance(local, tuple):
```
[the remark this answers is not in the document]
nvidianz: Good idea. Made some changes.

nvflare/fuel/f3/drivers/socket_conn.py
```diff
@@ -0,0 +1,170 @@
...
+        return conn_props
+
+
+class ConnectionHandler(BaseRequestHandler):
+    def handle(self):
+
+        # noinspection PyUnresolvedReferences
+        conn_props = {DriverParams.LOCAL_ADDR.value: self.server.local_addr}
+        if self.client_address:
+            if isinstance(self.client_address, tuple):
```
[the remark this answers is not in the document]
nvidianz: Actually, the code here is indeed duplicate and is removed.

nvflare/fuel/f3/drivers/socket_conn.py
```diff
@@ -0,0 +1,170 @@
...
+                peer_addr = f"{self.client_address[0]}:{self.client_address[1]}"
+            else:
+                peer_addr = self.client_address
+        else:
+            # ThreadingUnixStreamServer doesn't set client_address
+            # noinspection PyUnresolvedReferences
+            peer_addr = self.server.path
+        conn_props[DriverParams.PEER_ADDR.value] = peer_addr
+
+        # noinspection PyUnresolvedReferences
```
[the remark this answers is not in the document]
nvidianz: Python doesn't hve the full definition of the class so PyCharm complains about Unresovled References. This comment gets rid of the warning.

nvflare/fuel/f3/drivers/tcp_driver.py
```diff
@@ -0,0 +1,106 @@
...
+from nvflare.fuel.f3.drivers.base_driver import BaseDriver
+from nvflare.fuel.f3.drivers.driver import ConnectorInfo, Driver
+from nvflare.fuel.f3.drivers.driver_params import DriverCap, DriverParams
+from nvflare.fuel.f3.drivers.net_utils import get_ssl_context, get_tcp_urls
+from nvflare.fuel.f3.drivers.socket_conn import ConnectionHandler, SocketConnection
+
+log = logging.getLogger(__name__)
+
+
+class TcpStreamServer(ThreadingTCPServer):
```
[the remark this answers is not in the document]
nvidianz: TCP is a stream protocol vs UDP a datagram protocol. This is just to emphasize the data comes in as byte stream, with no boundaries.

nvflare/fuel/f3/drivers/tcp_driver.py
```diff
@@ -0,0 +1,106 @@
...
+
+    TCPServer.allow_reuse_address = True
+
+    def __init__(self, driver: "Driver", connector: ConnectorInfo):
```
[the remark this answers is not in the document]
nvidianz: The quote is not needed here. Removed.

It's only needed for forward reference: https://peps.python.org/pep-0484/#forward-references

nvflare/fuel/f3/drivers/tcp_driver.py
```diff
@@ -0,0 +1,106 @@
...
+        if self.ssl_context:
+            self.socket = self.ssl_context.wrap_socket(self.socket, server_side=True)
+
+        try:
+            self.server_bind()
+            self.server_activate()
+        except BaseException as ex:
+            log.error(f"{os.getpid()}: Error binding to  {host}:{port}: {ex}")
+            self.server_close()
+            raise
```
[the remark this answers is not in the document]
nvidianz: raise without exception just raises the last exception, same as 

raise ex

nvflare/fuel/f3/drivers/tcp_driver.py
```diff
@@ -0,0 +1,106 @@
...
+        super().__init__()
+        self.server = None
+
+    @staticmethod
+    def supported_transports() -> List[str]:
+        return ["tcp", "stcp"]
+
+    @staticmethod
+    def capabilities() -> Dict[str, Any]:
+        return {DriverCap.HEARTBEAT.value: False, DriverCap.SUPPORT_SSL.value: True}
```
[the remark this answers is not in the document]
nvidianz: The TCP protocol doesn't have heartbeat. This is just telling SFM the driver doesn't have this capability and SFM needs to handle the heartbeat.

Websocket/GRPC has built-in heartbeat and this value would be true.

nvflare/fuel/f3/drivers/base_driver.py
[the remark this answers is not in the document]
nvidianz: Tried that and it worked.

nvflare/fuel/f3/drivers/socket_conn.py
```diff
@@ -0,0 +1,170 @@
...
+    def send_frame(self, frame: BytesAlike):
+        try:
+            self.sock.sendall(frame)
+        except BaseException as ex:
+            raise CommError(CommError.ERROR, f"Error sending frame: {ex}")
+
+    def read_loop(self):
+        try:
+            self.read_frame_loop()
+        except CommError:
```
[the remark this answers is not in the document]
nvidianz: Added check for CLOSED.

yhwen (dismissed):

nvflare/fuel/f3/drivers/base_driver.py
```diff
@@ -0,0 +1,66 @@
...
+
+    def close_connection(self, conn: Connection):
+        log.debug(f"Connection removed: {self.get_name()}:{conn}")
+
+        conn.state = ConnState.CLOSED
+        self._notify_monitor(conn)
+
+        with self.conn_lock:
+            if not self.connections.pop(conn.name, None):
+                log.debug(f"{conn.name} is already removed from driver")
```
YuanTingHsieh, line 53: Can the connection name be `""`?

In that case should we use `if self.connections.pop(conn.name, None) is not None:`?

And should we use `log.debug(f"{conn.name} is successfully removed from driver")` ?

nvflare/fuel/f3/sfm/conn_manager.py
```diff
@@ -206,7 +206,7 @@ def send_message(self, endpoint: Endpoint, app_id: int, headers: Headers, payloa
 
     def register_message_receiver(self, app_id: int, receiver: MessageReceiver):
         if self.receivers.get(app_id):
-            raise CommError(CommError.BAD_DATA, f"Receiver for app {app_id} is already registered")
+            raise CommError(CommError.BAD_CONFIG, f"Receiver for app {app_id} is already registered")
```
YuanTingHsieh, line 209: When should we use `CommError.BAD_CONFIG` VS `CommError.BAD_DATA`?

nvflare/fuel/f3/drivers/base_driver.py
```diff
@@ -0,0 +1,66 @@
...
+    def close_all(self):
+        with self.conn_lock:
+            for name in sorted(self.connections.keys()):
+                conn = self.connections[name]
+                log.debug(f"Closing connection: {self.get_name()}:{conn}")
+                conn.close()
+
+    def _notify_monitor(self, conn: Connection):
+        if not self.conn_monitor:
+            log.error(f"Connection monitor not registered for driver {self.get_name()}")
```
YuanTingHsieh, line 64: if this is a programming error, should we raise Exception?

nvflare/fuel/f3/sfm/conn_manager.py
nvidianz: BAD_CONFIG is for errors happened locally and BAD_DATA is for errors in the data received from remote endpoint.

nvflare/fuel/f3/drivers/base_driver.py
nvidianz: Connection name can't be None.

"Already removed" is a corner case so I want that printed instead of the normal case.

nvflare/fuel/f3/drivers/tcp_driver.py
```diff
@@ -0,0 +1,106 @@
...
+
+    def listen(self, connector: ConnectorInfo):
+        self.connector = connector
+        self.server = TcpStreamServer(self, connector)
+        self.server.serve_forever()
+
+    def connect(self, connector: ConnectorInfo):
+        self.connector = connector
```
YuanTingHsieh, line 77: Why does `TCPDriver` keeps changing its `self.connector`?
In every listen and connect method.
If we need it, do we need to lock it since these methods can be called in diff. threads?
If we don't need it, should we just remove it?

nvflare/fuel/f3/drivers/base_driver.py
nvidianz: It's an error for production use. In tests, you may not care about the connection state. So it's better to log an error.

nvflare/fuel/f3/drivers/tcp_driver.py
nvidianz: Connector is needed in the driver but it's not available at driver construction time so it's passed in through those 2 methods.

Those methods are only called from one thread so lock is not needed.

nvflare/fuel/f3/drivers/socket_conn.py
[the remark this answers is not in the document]
YuanTingHsieh: So the bug is coming from the Python socket: https://docs.python.org/3/library/socket.html#socket.socket.getpeername ?

nvflare/fuel/f3/drivers/base_driver.py
YuanTingHsieh: But we should assume users are not doing testing right?

nvflare/fuel/f3/drivers/tcp_driver.py
YuanTingHsieh: Will we re-use the driver class?

If we re-used the driver class, it's `self.connector` will keep changing?

nvflare/fuel/f3/drivers/base_driver.py
nvidianz: Not for users, for developers. The error and exception have the same effect, they result in errors in the log.

nvflare/fuel/f3/drivers/tcp_driver.py
nvidianz: Yes, the driver can be reused for different connectors if implemented properly but it's not reused in current implementation for simplicity.

YuanTingHsieh: Got it, we might need to take care of this `self.connector` if we want to reuse this in the future.
For now since we are not reusing the driver, it is OK.

nvflare/fuel/f3/drivers/base_driver.py
YuanTingHsieh: I agree "log error" or "log exception" is similar.

But "log error" is different from "raising Exception"?

nvflare/fuel/f3/drivers/base_driver.py
YuanTingHsieh: In that case,
should the logic be

```
           if self.connections.pop(conn.name, None) is None:
                log.debug(f"{conn.name} is already removed from driver")
```

nvflare/fuel/f3/drivers/socket_conn.py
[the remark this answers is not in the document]
nvidianz: Yes, only on MacOS.

nvflare/fuel/f3/drivers/base_driver.py
nvidianz: "not x" and "x is None" should be identical in this use case. The former is more Pythonic.

nvflare/fuel/f3/drivers/base_driver.py
nvidianz: The only difference is that raising an exception will cause the connection to end and the error is logged once. While in current implementation, the connection still works and the error will be logged twice when state changes to CONNECTED and DISCONNECTED. I think latter is better.

nvflare/fuel/f3/sfm/conn_manager.py
```diff
@@ -139,7 +139,7 @@ def stop(self):
                 connector.stopping = True
                 connector.driver.shutdown()
 
-        self.conn_mgr_executor.shutdown(True)
+        self.conn_mgr_executor.shutdown(False)
```
YuanTingHsieh, line 142: I noticed you shutdown the socket now, so maybe this can stay "True"

nvflare/fuel/f3/drivers/base_driver.py
YuanTingHsieh: If the driver is written wrong / programming error, we should just ends the program?

nvflare/fuel/f3/drivers/base_driver.py
YuanTingHsieh: IMO here (since we use `pop(something, None)`) use `is None` or `is not None` is more Pythonic, but I am OK with `not x`

nvflare/fuel/f3/sfm/conn_manager.py
nvidianz: Yeah. Changed to true.

nvflare/fuel/f3/drivers/base_driver.py
nvidianz: Let's talk about this and deal with it in another PR.

YuanTingHsieh (approved):
Looks good to me.

The following points can be improved later:
- improve the "self.connector" when we want to support reusing driver 
- improve the handling of missing "conn_monitor" in driver (raise exception rather than log)

PR #1367, opened 2023-02-15
Fixed the issue could not locate the components due to app_opt move.

yhwen:
### Description

Fixed the issue the component could not be located due to "app_opt" folder move.

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh:
This PR is trying to fix if users want to use name to refer to components that are originally in nvflare.app_common.pt:

For example:

```
"name": "PTFileModelPersistor"
```

YuanTingHsieh (approved):
By doing this will allow users to use "name" to refer to classes inside "app_opt" folder.

Note that, if a module requires certain library but the user forgot to install that library, the user will get the same error "ConfigError: Class XXX not found" just as before.

PR #1369, opened 2023-02-16
Change warning to debug for ModuleScanner

YuanTingHsieh:
### Description

Follows PR #1367 . 

For users' convenience, we decide to try to import all the modules in app_opt folder by default.
So if users have installed the required libraries, they can use `"name": "SomeClassInAppOpt"` in the configuration files.

Since we try to import all the modules in app_opt folder, it will print too much warnings that might scare/annoy users.
So we change the level to "debug" for a better user experience.

Also added pt_by_name app to integration test to cover this part.

### Types of changes

- [ ] Documentation updated.


nvflare/fuel/utils/class_utils.py
```diff
@@ -96,13 +111,31 @@ def _create_classes_table(self):
...
+            The module name if found.
+        """
+        if class_name not in self._class_table:
+            return None
+        scan_result = self._class_table[class_name]
+        module_name = scan_result.modules[0]
+        if len(scan_result.modules) != 1:
+            self._logger.warning(
+                f"More than 1 module contains class {class_name}. The first module: {module_name} is used."
+            )
```
yhwen, line 140: The intention of this is to address the duplicate class name used in different module. But the solution is half baked. Simply giving the warning for this could easily cause confusion, because a wrong class instance been created. 

Instead, it's better to let the configuration fail and let the user know the situation. At the same time, we can provide the solution for choose which one to use. For example, when duplicate class name detected, we store the "module_name.class_name" as the key in the _class_table. So user can use that unique name in the configuration.

nvflare/fuel/utils/class_utils.py
```diff
@@ -96,13 +112,35 @@ def _create_classes_table(self):
...
+    @staticmethod
+    def _get_class_table_from_scan_result(scan_result_table: Dict[str, _ModuleScanResult]) -> Dict[str, str]:
+        class_table = {}
+        for class_name, scan_result in scan_result_table.items():
+            if len(scan_result.modules) != 1:
+                for module_name in scan_result.modules:
+                    class_table[f"{module_name}.{class_name}"] = module_name
+            else:
+                class_table[class_name] = scan_result.modules[0]
+        return class_table
```
yhwen, line 135: This logic is correct. But I think we can optimize it to include this operation within the one step of  following operations, instead of doing it in another loop.
                                    
scan_result.add_module(module_name)
scan_result_table[name] = scan_result

YuanTingHsieh: thanks, simplified.

yhwen (approved):

PR #1381, opened 2023-02-16
Add limit to number of jobs in list_jobs and options to flare_api

nvkevlu:
### Description

Adds option to the list_jobs command to limit the maximum number of jobs returned with the -m option (the default is now 5 jobs returned). The FLARE API now has options to use the new limit for the number of jobs as well as the filters for id_prefix and name_prefix.

Note: Due to the way how FLAdminAPI commands take options, users can use any of these options of list_jobs. The filters for id_prefix and name_prefix already existed, and -n was already used for the name prefix, so I chose -m for the maximum jobs limit. I looked into how to add more custom filters, but no matter what I could come up with, nothing made more sense than just filtering with the results obtained from the list_jobs command (passing in functions and the necessary arguments gets very awkward, and the returned result should be very easy to work with and Python allows for the easy manipulation of it).

### Types of changes

- [ ] Documentation updated.


nvflare/fuel/flare_api/flare_api.py
```diff
@@ -260,21 +261,38 @@ def get_job_meta(self, job_id: str) -> dict:
             raise InternalError("server failed to return job meta")
         return job_meta
 
-    def list_jobs(self, detailed: bool = False, all: bool = False) -> List[dict]:
+    def list_jobs(
+        self, detailed: bool = False, all: bool = False, limit: int = 5, id_prefix: str = None, name_prefix: str = None
```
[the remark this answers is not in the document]
nvkevlu: This was the original approach I was trying to take, but looking into the actual existing implementation with the current filters on the server side (to not make changes to existing functionality) and experimenting with actually using the function, I made the decision to instead just support the existing filters.

It was very hard to figure out a good way to put the filter into list_jobs because it was almost trivial to do additional filtering with python to the result of list_jobs, and the code was more readable, functional, and logical than any way to package it into list_jobs. You could also use a sorting function similarly, which I realized can also be easily performed in a single line with the result for list_jobs (but really hard to elegantly do inside list_jobs), so perhaps we could document a couple examples of how this can be done with the result of list_jobs instead.

nvflare/fuel/flare_api/flare_api.py
```diff
@@ -275,6 +280,18 @@ def list_jobs(self, detailed: bool = False, all: bool = False) -> List[dict]:
...
+        if name_prefix:
+            if not isinstance(name_prefix, str):
+                raise InvalidArgumentError("name_prefix must be str but got {}.".format(type(name_prefix)))
+            else:
+                command = command + " -n " + name_prefix
+        if id_prefix:
+            if not isinstance(id_prefix, str):
+                raise InvalidArgumentError("id_prefix must be str but got {}.".format(type(id_prefix)))
+            else:
+                command = command + " " + id_prefix
```
YuanTingHsieh, line 294: To be consistent should we also check other arguments passed in to make sure they have the correct type?

nvkevlu: I can add checks for other args

nvkevlu: May do this in future PR though, going through more systematically. For now just getting this PR merged.

PR #1387, opened 2023-02-17
Job run integrate with FCI Cellnet

yhwen:
### Description

Enable the job run to work with FCI Cellnet, including all the admin commands.

### Types of changes

- [ ] Documentation updated.


nvflare/apis/fl_exception.py
```diff
@@ -16,12 +16,14 @@
...
         Args:
             exception: grpc.RpcError when trying to register gprc channel
         """
         super().__init__()
-        # Copy all the gRPC exception properties into FLCommunicationError instance.
-        self.__dict__.update(exception.__dict__)
+        # Copy all the exception properties into FLCommunicationError instance.
+        if exception:
+            self.__dict__.update(exception.__dict__)
+        self.message = message
```
YuanTingHsieh, line 29: Where are we using this?

If we modify the "__init__" method here we need to also change all the places that use this class.

nvflare/apis/utils/decomposers/flare_decomposers.py
```diff
@@ -30,6 +30,7 @@
 from nvflare.apis.workspace import Workspace
 from nvflare.fuel.utils import fobs
 from nvflare.fuel.utils.fobs.decomposer import Decomposer, DictDecomposer
+from nvflare.private.admin_defs import Message
```
YuanTingHsieh, line 33: In the package of "nvflare/apis", we should not import from nvflare.private

nvflare/apis/utils/decomposers/flare_decomposers.py
```diff
@@ -63,15 +64,11 @@ def register():
     if register.registered:
         return
 
+    from nvflare.private.fed.server.run_manager import RunInfo
```
YuanTingHsieh, line 67: In the package of "nvflare/apis", we should not import from nvflare.private

nvflare/app_common/job_schedulers/job_scheduler.py
```diff
@@ -155,7 +155,7 @@ def _try_job(self, job: Job, fl_ctx: FLContext) -> (int, Optional[Dict[str, Disp
         job_participants = [fl_ctx.get_identity_name(default="server")]
         job_participants.extend(applicable_sites)
 
-        fl_ctx.set_prop(FLContextKey.CURRENT_JOB_ID, job.job_id, private=True, sticky=False)
+        fl_ctx.set_prop(FLContextKey.CURRENT_JOB_ID, job.job_id, private=True)
```
YuanTingHsieh, line 158: Do we want to change this line?

nvflare/private/fed/cmi.py
```diff
@@ -0,0 +1,252 @@
...
+from nvflare.apis.fl_constant import FLContextKey, ReturnCode
+from nvflare.apis.fl_context import FLContext
+from nvflare.apis.shareable import ReservedHeaderKey, Shareable, make_reply
+from nvflare.fuel.f3.cellnet.cell import FQCN, Cell, Message, MessageHeaderKey
+from nvflare.fuel.f3.cellnet.cell import ReturnCode as CellReturnCode
+from nvflare.fuel.f3.cellnet.cell import new_message
+from nvflare.private.defs import CellMessageHeaderKeys
+
+
+class CellMessageInterface(FLComponent):
```
YuanTingHsieh, line 27: should this be `CellMessageInterface(FLComponent, ABC):`?
Since it has abstract method that needs to be defined

nvflare/private/fed/cmi.py
```diff
@@ -0,0 +1,252 @@
...
+                shareable.set_peer_props(public_props)
+
+    def _filter_incoming_request(self, message: Message):
+        self._filter_incoming_message(message)
+        fl_ctx = self.engine.new_context()
+        peer_ctx = message.get_prop(self.PROP_KEY_PEER_CTX)
+        assert isinstance(fl_ctx, FLContext)
+        if peer_ctx:
+            fl_ctx.set_peer_context(peer_ctx)
+        message.set_prop(self.PROP_KEY_FL_CTX, fl_ctx)
```
YuanTingHsieh, line 90: When we get a message generated by "new_cmi_message" it should already has the prop `PROP_KEY_FL_CTX`? 

So we don't need to set it again?

Can this JobCellMessenger get message that are not created by `new_cmi_message`?

nvflare/private/fed/cmi.py
```diff
@@ -0,0 +1,252 @@
...
+
+    def _filter_incoming_message(self, message: Message):
+        public_props = message.get_header(self.HEADER_KEY_PEER_PROPS)
+        if public_props:
+            peer_ctx = self._make_peer_ctx(public_props)
+            message.set_prop(self.PROP_KEY_PEER_CTX, peer_ctx)
+        shareable = message.payload
+        if isinstance(shareable, Shareable):
+            if public_props:
+                shareable.set_peer_props(public_props)
```
YuanTingHsieh, line 81: It feels weird to modify the message payload if you are a messenger.

nvflare/private/fed/server/job_runner.py
```diff
@@ -344,67 +345,66 @@ def run(self, fl_ctx: FLContext):
...
-                                    }
-                                else:
-                                    deployable_clients = client_sites
-
-                                self._start_run(
-                                    job_id=job_id,
-                                    job=ready_job,
-                                    client_sites=deployable_clients,
-                                    fl_ctx=fl_ctx,
+                        # with self.lock:
```
YuanTingHsieh, line 348: If we remove this lock, do we need to make sure the job_manager will function correctly if multiple threads calling it?

nvflare/private/fed/server/job_runner.py
```diff
@@ -430,17 +430,20 @@ def restore_running_job(self, run_number: str, job_id: str, job_clients, snapsho
...
-        with self.lock:
-            self._stop_run(job_id, fl_ctx)
-            job = self.running_jobs.get(job_id)
-            if job:
-                self.log_info(fl_ctx, f"Stop the job run: {job_id}")
-                fl_ctx.set_prop(FLContextKey.CURRENT_JOB_ID, job.job_id)
-                job_manager.set_status(job.job_id, RunStatus.FINISHED_ABORTED, fl_ctx)
+        self._stop_run(job_id, fl_ctx)
+
+        # with self.lock:
```
YuanTingHsieh, line 435: if we change this lock location, do we need to make sure the job_manager will function correctly if multiple threads calling it?

nvflare/private/fed/server/job_runner.py
```diff
@@ -344,67 +345,66 @@ def run(self, fl_ctx: FLContext):
...
-                                    )
-
-                                self.fire_event(EventType.JOB_ABORTED, fl_ctx)
-                                self.log_error(
-                                    fl_ctx, f"Failed to run the Job ({ready_job.job_id}): {secure_format_exception(e)}"
+                            job_manager.set_status(ready_job.job_id, RunStatus.RUNNING, fl_ctx)
+                        except BaseException as e:
+                            if job_id:
+                                if job_id in self.running_jobs:
+                                    del self.running_jobs[job_id]
```
YuanTingHsieh, line 393: this deletion needs to be inside a lock?

nvflare/private/fed/server/job_runner.py
```diff
@@ -344,67 +345,66 @@ def run(self, fl_ctx: FLContext):
...
+                                self._stop_run(job_id, fl_ctx)
```
YuanTingHsieh, line 394: need to make sure this _stop_run if not put inside a lock, will still function correctly when multiple threads arrive here

nvflare/private/fed/server/run_manager.py
```diff
@@ -13,18 +13,19 @@
...
 from nvflare.apis.fl_constant import MachineStatus
 from nvflare.apis.fl_context import FLContext, FLContextManager
 from nvflare.apis.server_engine_spec import ServerEngineSpec
 from nvflare.apis.workspace import Workspace
+from nvflare.private.aux_runner import AuxRunner
 from nvflare.private.event import fire_event
 from nvflare.private.fed.utils.fed_utils import create_job_processing_context_properties
 
 from .client_manager import ClientManager
-from .server_aux_runner import ServerAuxRunner
```
YuanTingHsieh, line 29: If we are not using `ServerAuxRunner` anymore we will also need to remove its related code/docstrings

nvflare/private/fed/app/client/worker_process.py
```diff
@@ -129,17 +134,17 @@ def main():
         client_app_runner.start_run(app_root, args, config_folder, federated_client, secure_train)
 
     except BaseException as e:
-        logger = logging.getLogger("worker_process")
-        logger.error(f"FL client execution exception: {secure_format_exception(e)}")
+        if logger:
+            logger.error(f"FL client execution exception: {secure_format_exception(e)}")
```
YuanTingHsieh, line 138: it is possible the error happens during 
```
conf = FLClientStarterConfiger(
            workspace=workspace,
            args=args,
            kv_list=args.set,
        )
        conf.configure()

```

nvflare/private/fed/server/run_manager.py
```diff
@@ -107,5 +110,36 @@ def fire_event(self, event_type: str, fl_ctx: FLContext):
     def add_handler(self, handler: FLComponent):
         self.handlers.append(handler)
 
+    def get_cell(self):
+        return self.cell
+
+    def validate_clients(self, client_names: List[str]) -> Tuple[List[Client], List[str]]:
+        return self._get_all_clients_from_inputs(client_names)
```
YuanTingHsieh, line 117: This function should be in `ServerEngine`, instead of in here.

nvflare/private/fed/server/server_app_runner.py
```diff
@@ -66,13 +64,13 @@ def start_server_app(self, workspace: Workspace, server, args, app_root, job_id,
...
         finally:
-            self.update_job_run_status(server)
+            # self.update_job_run_status(server)
             server.status = ServerStatus.STOPPED
             server.engine.engine_info.status = MachineStatus.STOPPED
             server.stop_training()
 
     def sync_up_parents_process(self, args, server):
-        server.engine.create_parent_connection(int(args.conn))
+        # server.engine.create_parent_connection(int(args.conn))
```
YuanTingHsieh, line 73: we should remove the commented out codes

nvflare/private/fed/server/server_command_agent.py
```diff
@@ -12,60 +12,101 @@
...
-            except Exception as e:
-                self.logger.error(
-                    f"IPC Communication error on the port: {self.listen_port}: {secure_format_exception(e)}."
-                )
+        self.cell.register_request_cb(
+            channel=CellChannel.AUX_COMMUNICATION,
+            topic="*",
+            cb=self.aux_communicate,
+        )
+        self.logger.info(f"ServerCommandAgent cell start: {self.cell.get_fqcn()}")
```
YuanTingHsieh, line 56: This is not "start" anymore, we are registering the callback into the CELL.
The Cell is already started somewhere else.

Suggest change method name and change log message.

nvflare/private/fed/server/server_command_agent.py
```diff
@@ -12,60 +12,101 @@
...
+
+        assert isinstance(request, CellMessage), "request must be CellMessage but got {}".format(type(request))
+
+        command_name = request.get_header(MessageHeaderKey.TOPIC)
+        data = fobs.loads(request.payload)
+
+        token = request.get_header(CellMessageHeaderKeys.TOKEN, None)
+        client = None
+        if token:
+            client = self.engine.server.client_manager.clients.get(token)
```
YuanTingHsieh, line 68: It feels weird to accessing the instance variable like this, we should introduce some getting methods?

nvflare/private/fed/server/server_command_agent.py
```diff
@@ -12,60 +12,101 @@
...
+        self.cell.register_request_cb(
+            channel=CellChannel.AUX_COMMUNICATION,
+            topic="*",
+            cb=self.aux_communicate,
+        )
+        self.logger.info(f"ServerCommandAgent cell start: {self.cell.get_fqcn()}")
+
+    def execute_command(self, request: CellMessage) -> CellMessage:
+
+        assert isinstance(request, CellMessage), "request must be CellMessage but got {}".format(type(request))
```
YuanTingHsieh, line 60: we should remove assert code if not in test code

nvflare/private/fed/server/server_command_agent.py
```diff
@@ -12,60 +12,101 @@
...
 import logging
-import threading
 
-from nvflare.apis.fl_constant import ServerCommandKey
+from nvflare.apis.fl_constant import FLContextKey, ServerCommandKey
+from nvflare.apis.fl_context import FLContext
+from nvflare.apis.shareable import ReservedHeaderKey
+from nvflare.apis.utils.fl_context_utils import get_serializable_data
+from nvflare.fuel.f3.cellnet.cell import Cell
+from nvflare.fuel.f3.cellnet.cell import Message as CellMessage
```
YuanTingHsieh, line 23: Message class is defined in f3 message, I suggest don't import it from cellnet

Should be: `from nvflare.fuel.f3.message import Message`

nvflare/private/fed/server/server_engine_internal_spec.py
```diff
@@ -206,15 +203,26 @@ def build_component(self, config_dict):
...
     def get_client_from_name(self, client_name: str) -> Client:
-        """Gets the registered client from client_name.
+        """Get the registered client token from client_name.
 
         Args:
             client_name: client name
 
-        Returns:
-            The registered client.
+        Returns: registered client
```
YuanTingHsieh, line 226: No need to change the docstring.

nvflare/private/fed/server/server_engine_internal_spec.py
```diff
@@ -206,15 +203,26 @@ def build_component(self, config_dict):
...
+        """Get the registered client name from communication token.
+
+        Args:
+            token: communication token
+
+        Returns:
+            Client name
+        """
+        pass
+
```
YuanTingHsieh, line 217: This function already defined in `ServerEngineSpec`

No need to define again.

nvflare/private/fed/server/server_engine_internal_spec.py
```diff
@@ -39,7 +37,6 @@ def __init__(self):
 
 
 class ServerEngineInternalSpec(ServerEngineSpec, ABC):
-    @abstractmethod
     def get_engine_info(self) -> EngineInfo:
```
YuanTingHsieh, line 40: this method should be an abstractmethod

nvflare/private/fed/server/server_engine_internal_spec.py
```diff
@@ -237,41 +245,33 @@ def ask_to_stop(self):
...
         Returns:
-            Component stats of the server
+            Server components errors.
 
         """
         pass
 
     @abstractmethod
-    def get_errors(self, job_id) -> dict:
+    def reset_errors(self, job_id) -> str:
```
YuanTingHsieh, line 274: this method is not needed in the spec, please remove it.

nvflare/private/fed/server/server_runner.py
```diff
@@ -335,8 +336,8 @@ def _try_to_get_task(self, client, fl_ctx, timeout=1.0, retry_interval=0.005):
 
                 time.sleep(retry_interval)
 
-            # ask client to retry
-            return "", "", None
+        # ask client to retry
+        return "", "", None
```
YuanTingHsieh, line 340: Is this intended?

nvflare/private/fed/server/sys_cmd.py
```diff
@@ -131,7 +130,7 @@ def _process_replies(self, conn, replies):
             else:
                 conn.append_string(": No replies")
 
-    def report_resources(self, conn: Connection, args: List[str]):
+    def report_resources(self, conn: Connection, args: [str]):
```
YuanTingHsieh, line 133: we should keep the type hints in this file

nvflare/private/fed/server/training_cmds.py
```diff
@@ -320,7 +320,7 @@ def _process_client_status_replies(self, conn, replies):
             else:
                 table.add_row([client_name, app_name, job_id, "No Reply"])
 
-    def _add_scope_info(self, table, site_name, scope_names: List[str], default_scope: str):
+    def _add_scope_info(self, table, site_name, scope_names: [str], default_scope: str):
```
YuanTingHsieh, line 323: let's keep the type hint

nvflare/private/fed/client/communicator.py
```diff
@@ -136,7 +136,7 @@ def client_registration(self, client_name, servers, project_name):
                 token = result.get_header(CellMessageHeaderKeys.TOKEN)
                 ssid = result.get_header(CellMessageHeaderKeys.SSID)
                 if not token and not self.should_stop:
-                    time.sleep(5)
+                    time.sleep(2)
```
YuanTingHsieh, line 139: we might want to make this configurable

nvflare/private/fed/client/client_engine_executor_spec.py
```diff
@@ -63,6 +64,17 @@ def get_widget(self, widget_id: str) -> Widget:
...
+        """To validate the target names.
+
+        Args:
+            target_names: input target names
+
+        Returns: (valid_names, invalid_names)
+
+        """
+        pass
+
```
YuanTingHsieh, line 77: What does it mean to have `validate_clients` in `client_engine_executor_spec`?

Should we rename to `validate_targets` or something else?

nvflare/private/fed/client/client_app_runner.py
```diff
@@ -89,9 +97,17 @@ def create_run_manager(self, args, conf, federated_client, workspace):
         )
 
     def start_command_agent(self, args, client_runner, federated_client, fl_ctx):
+        start = time.time()
+        while federated_client.cell is None:
+            time.sleep(0.1)
+            if time.time() - start > 60.0:
+                raise RuntimeError("No cell created. Failed to start the command_agent for ClientAppRunner.")
```
YuanTingHsieh, line 104: these randomly pick values should be configurable somewhere

nvflare/private/fed/client/client_app_runner.py
```diff
@@ -33,14 +34,20 @@ def __init__(self) -> None:
 
     def start_run(self, app_root, args, config_folder, federated_client, secure_train):
         client_runner = self.create_client_runner(app_root, args, config_folder, federated_client, secure_train)
+        federated_client.set_client_runner(client_runner)
+        start = time.time()
+        while federated_client.communicator.cell is None:
+            time.sleep(1.0)
+            if time.time() - start > 60.0:
```
YuanTingHsieh, line 41: here as well

nvflare/private/fed/app/server/runner_process.py
```diff
@@ -129,8 +133,27 @@ def main():
...
+def kill_child_processes(parent_pid, sig=signal.SIGTERM):
+    try:
+        parent = psutil.Process(parent_pid)
+    except psutil.NoSuchProcess:
+        return
+    children = parent.children(recursive=True)
+    for process in children:
+        process.send_signal(sig)
+
+
```
YuanTingHsieh, line 151: we are not using these two methods, why add them back

nvflare/apis/fl_exception.py
yhwen: This has already been changed.

nvflare/apis/utils/decomposers/flare_decomposers.py
yhwen: Will move the decomposers used by the platform to a separate private_decomposers.

nvflare/apis/utils/decomposers/flare_decomposers.py
yhwen: Will move out.

nvflare/app_common/job_schedulers/job_scheduler.py
yhwen: Yes, otherwise there's a conflict.

nvflare/private/fed/cmi.py
yhwen: Done

nvflare/private/fed/cmi.py
yhwen: When getting an incoming request, it only guarantee a request Message. The Message does not have to be generated by `new_cmi_message`.

nvflare/private/aux_runner.py
```diff
@@ -126,44 +100,62 @@ def _process_request(self, topic: str, request: Shareable, fl_ctx: FLContext) ->
...
+        job_id = fl_ctx.get_prop(FLContextKey.CURRENT_RUN)
+        cmi = JobCellMessenger(self.engine, job_id)
+
+        # validate targets
+        request.set_peer_props(fl_ctx.get_all_public_props())
+        request.set_header(ReservedHeaderKey.TOPIC, topic)
+
+        target_names = []
+        for t in targets:
+            if isinstance(t, str):
```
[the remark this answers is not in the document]
yhwen: There's execution flow interruption. It can not be simply wrapped into a method.

nvflare/private/fed/cmi.py
yhwen: `public_props` are special message headers in the message exchange. We are not modifying any other message payload.

nvflare/private/fed/app/client/worker_process.py
```diff
@@ -98,13 +100,15 @@ def main():
...
 
+    logger = None
     try:
         # start parent process checking thread
         thread = threading.Thread(target=check_parent_alive, args=(parent_pid, stop_event))
         thread.start()
 
         conf = FLClientStarterConfiger(
             workspace=workspace,
+            args=args,
```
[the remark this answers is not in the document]
yhwen: We can make this enhancement in the future. That may require quite a few changes.

nvflare/private/fed/server/training_cmds.py
```diff
@@ -130,12 +130,12 @@ def _shutdown_app_on_clients(self, conn: Connection) -> bool:
             # means some clients can not be shutdown
             result = False
 
-        if clients_to_be_removed:
-            # needs to remove all the clients that can be removed
-            err = engine.remove_clients(clients_to_be_removed)
-            if err:
-                conn.append_error(err)
-                result = False
+        # if clients_to_be_removed:
```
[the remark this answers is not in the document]
yhwen: Done

nvflare/private/fed/cmi.py
```diff
@@ -0,0 +1,252 @@
...
+                    cell.queue_message(channel=channel, topic=topic, message=cell_msg, targets=target_fqcns)
+                else:
+                    cell.fire_and_forget(
+                        channel=channel,
+                        topic=topic,
+                        message=cell_msg,
+                        targets=target_fqcns,
+                    )
+                return {}
+        except BaseException as e:
```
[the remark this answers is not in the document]
yhwen: Changed this to throw the exception and let the outer calling place the handle the exception.

nvflare/private/fed/cmi.py
```diff
@@ -0,0 +1,252 @@
...
+                if cell_replies:
+                    for k, v in cell_replies.items():
+                        assert isinstance(v, Message)
+                        rc = v.get_header(MessageHeaderKey.RETURN_CODE, ReturnCode.OK)
+                        client_name = FQCN.get_root(k)
+                        if rc == CellReturnCode.OK:
+                            result = v.payload
+                            if not isinstance(result, Shareable):
+                                self.logger.error(f"reply of {channel}:{topic} must be dict but got {type(result)}")
+                                result = make_reply(ReturnCode.ERROR)
```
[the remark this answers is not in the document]
yhwen: The request handling party replied the message. From the message exchange perspective, the message reply is OK. However, the reply message payload is not the expected Shareable. That's still treated as an ERROR.

nvflare/private/fed/server/job_runner.py
yhwen: The self.lock is not totally removed. Just reduced the lock block scope. Made it less impact other codes executing.

nvflare/private/fed/server/job_runner.py
yhwen: Yes. The purpose of the lock is only to protect the self.running_jobs change. Reduced the scope to only necessary.

nvflare/private/fed/server/job_runner.py
yhwen: good catch. Added it.

nvflare/private/fed/server/job_runner.py
yhwen: _stop_run only need to send the stop message / command to the child process. no need to be placed in the lock block.

nvflare/private/fed/server/run_manager.py
yhwen: The removal will be included in the follow up PR.

nvflare/private/fed/app/client/worker_process.py
yhwen: When this happened, the logger has not been initialized, we could not use the logger to log the error. Instead, the exception will be raised. the Pyhton will print out the clear program stack trace.

nvflare/private/fed/server/server_app_runner.py
yhwen: Removed.

nvflare/private/fed/server/server_command_agent.py
yhwen: From the ServerCommandAgent perspective, it's still the start. Changed the log message to register_request_cb.

nvflare/private/fed/server/server_command_agent.py
yhwen: Changed.

nvflare/private/fed/server/server_command_agent.py
yhwen: wrapped in _get_client()

nvflare/private/fed/server/server_command_agent.py
yhwen: Done

nvflare/private/fed/server/server_engine_internal_spec.py
yhwen: Done

nvflare/private/fed/server/server_engine_internal_spec.py
yhwen: Removed.

nvflare/private/fed/server/server_engine_internal_spec.py
yhwen: Added.

nvflare/private/fed/server/server_engine_internal_spec.py
yhwen: This will b needed in next PR when adding the reset_error command.

nvflare/private/fed/server/server_runner.py
yhwen: Yes, this results to SpecialTaskName.TRY_AGAIN

nvflare/private/fed/server/run_manager.py
yhwen: Refactored. Move the get_all_clients_from_inputs() to client_manager.

nvflare/private/fed/server/sys_cmd.py
yhwen: The List[str] and [str] essentially the same. It's a string list.

nvflare/private/fed/server/training_cmds.py
yhwen: The List[str] and [str] essentially the same. It's a string list.

nvflare/private/fed/client/communicator.py
yhwen: I really don't see the meaningful reason for making this configurable. It's within the platform managed process. But made it configurable any way.

nvflare/private/fed/client/client_engine_executor_spec.py
yhwen: extracted the validate_targets to EngineSpec. Both server and client engine require this method.

nvflare/private/fed/client/client_app_runner.py
yhwen: Added configurable.

nvflare/private/fed/client/client_app_runner.py
yhwen: Added configurable.

nvflare/private/fed/app/server/runner_process.py
yhwen: cleaned up.

nvflare/apis/utils/decomposers/flare_decomposers.py
yhwen: Moved.

nvflare/apis/utils/decomposers/flare_decomposers.py
yhwen: Moved.

nvflare/private/fed/app/client/worker_process.py
YuanTingHsieh: the old logic will log error out securely in that case, we might want to keep the old logic

nvflare/private/fed/server/sys_cmd.py
YuanTingHsieh: From type hinting tool static analysis perspective (for example "mypy"),
Only List[str] is valid for the check.

otherwise we get 
`error: Bracketed expression "[...]" is not valid as a type  [valid-type]
`

nvflare/private/fed/server/training_cmds.py
YuanTingHsieh: From type hinting tool static analysis perspective (for example "mypy"),
Only List[str] is valid for the check.

otherwise we get
error: Bracketed expression "[...]" is not valid as a type  [valid-type]

nvflare/private/fed/server/run_manager.py
```diff
@@ -33,10 +35,10 @@ def __init__(self, job_id, app_path):
         self.job_id = job_id
         self.start_time = time.time()
         self.app_path = app_path
-        self.status = MachineStatus.STOPPED
+        self.status = MachineStatus.STOPPED.value
 
 
-class RunManager:
+class RunManager(EngineSpec):
```
YuanTingHsieh, line 41: Just a side note I think we need to clean up these names make server and client side names more consistent.

For client side we have ClientEngine inherit `ClientEngineInternalSpec`
ClientRunManager inherit `ClientEngineExecutorSpec`.

On server side we have ServerEngine inherit `ServerEngineInternalSpec`
ServerRunManager inherit `EngineSpec`


And then the `ServerEngineInternalSpec` inherits `ServerEngineSpec` inherits `EngineSpec`.
`ClientEngineInternalSpec` inherits `ClientEngineSpec` inherits `EngineSpec`.
`ClientEngineExecutorSpec` inherits both `ClientEngineSpec` and `EngineSpec`

^ we need some clean up for these complicated relationships/naming.
Maybe some middle layers can be removed.

nvflare/private/fed/client/client_engine.py
```diff
@@ -53,7 +53,6 @@ def __init__(self, client, client_name, args, rank, workers=5):
         Args:
```
YuanTingHsieh, line 53: why `ClientEngine` don't need `validate_targets`?

nvflare/private/fed/client/client_run_manager.py
```diff
@@ -113,11 +120,10 @@ def new_context(self) -> FLContext:
         return self.fl_ctx_mgr.new_context()
 
     def send_task_result(self, result: Shareable, fl_ctx: FLContext) -> bool:
-        try:
-            self.client.push_results(result, fl_ctx)  # push task execution results
+        push_result = self.client.push_results(result, fl_ctx)  # push task execution results
+        if push_result[0] == CellReturnCode.OK:
```
YuanTingHsieh, line 124: This assume only one service.
For now it is fine, if anytime we do add another service need to change this.
I personally think we should just change all codes to assume we have 1 service.

nvflare/private/fed/utils/decomposers/private_decomposers.py
```diff
@@ -0,0 +1,36 @@
...
+
+from nvflare.fuel.utils import fobs
+from nvflare.private.admin_defs import Message
+
+
+def register():
+    if register.registered:
+        return
+
+    from nvflare.private.fed.server.run_manager import RunInfo
```
YuanTingHsieh, line 29: should we put this import upfront?

nvflare/private/fed/client/client_app_runner.py
```diff
@@ -28,19 +29,26 @@
...
+        federated_client.set_client_runner(client_runner)
+        start = time.time()
+        while federated_client.communicator.cell is None:
+            time.sleep(1.0)
+            if time.time() - start > self.timeout:
+                raise RuntimeError("No cell created for communicator. Failed to start the ClientAppRunner.")
         federated_client.status = ClientStatus.STARTED
         client_runner.run(app_root, args)
 
+        federated_client.communicator.cell.stop()
```
YuanTingHsieh, line 47: Instead of accessing the instance variable hierarchically, should we define a function/method in FL client called `stop_cell` and call that here?

nvflare/private/fed/server/server_engine.py
```diff
@@ -570,101 +477,50 @@ def send_aux_request(self, targets: [], topic: str, request: Shareable, timeout:
...
+            channel=CellChannel.SERVER_PARENT_LISTENER,
+            topic=ServerCommandNames.GET_CLIENTS,
+            request=request,
+        )
+        data = fobs.loads(return_data.payload)
+        if data.get(ServerCommandKey.CLIENTS):
+            return data
+        else:
+            time.sleep(1.0)
+            return self._get_clients_data(job_id)
```
YuanTingHsieh, line 497: If for whatever reason we can't get the clients from the return_data's payload,
this function never ends?

nvflare/private/fed/server/server_engine.py
```diff
@@ -693,13 +549,10 @@ def persist_components(self, fl_ctx: FLContext, completed: bool):
...
-                    job_id = return_data.get(ServerCommandKey.JOB_ID)
-                    job_clients = return_data.get(ServerCommandKey.CLIENTS)
-                    fl_ctx.set_prop(FLContextKey.JOB_INFO, (job_id, job_clients))
+                return_data = self._get_clients_data(job_id)
+                job_id = return_data.get(ServerCommandKey.JOB_ID)
+                job_clients = return_data.get(ServerCommandKey.CLIENTS)
+                fl_ctx.set_prop(FLContextKey.JOB_INFO, (job_id, job_clients))
             else:
                 (job_id, job_clients) = job_info
             snapshot.set_component_snapshot(
```
YuanTingHsieh, line 558: do we need the `self.parent_conn_lock` anymore?

nvflare/private/fed/server/server_commands.py
```diff
@@ -293,6 +294,33 @@ def process(self, data: Shareable, fl_ctx: FLContext):
...
 
 
+class ResetErrorsCommand(CommandProcessor):
+    """To implement the show_errors command."""
+
+    def get_command_name(self) -> str:
+        """To get the command name.
+
+        Returns: ServerCommandNames.GET_ERRORS
+
```
YuanTingHsieh, line 304: Returns:
    `ServerCommandNames.RESET_ERRORS`

nvflare/private/fed/server/server_commands.py
```diff
@@ -293,6 +294,33 @@ def process(self, data: Shareable, fl_ctx: FLContext):
...
+    def process(self, data: Shareable, fl_ctx: FLContext):
+        """Called to process the abort command.
+
+        Args:
+            data: process data
+            fl_ctx: FLContext
+
+        Returns: Engine run_info
+
+        """
```
YuanTingHsieh, line 317: nothing is returned here

nvflare/private/fed/server/server_commands.py
```diff
@@ -317,6 +345,30 @@ def process(self, data: Shareable, fl_ctx: FLContext):
         return None
 
 
+class HeartbeatCommand(CommandProcessor):
+    """To implement the ShutdownCommand."""
```
YuanTingHsieh, line 349: should this be 
`HeartbeatCommand` ?

nvflare/private/fed/server/server_commands.py
```diff
@@ -317,6 +345,30 @@ def process(self, data: Shareable, fl_ctx: FLContext):
...
+
+    def get_command_name(self) -> str:
+        """To get the command name.
+
+        Returns: AdminCommandNames.SHUTDOWN
+
+        """
+        return ServerCommandNames.HEARTBEAT
```
YuanTingHsieh, line 357: Returns `ServerCommandNames.HEARTBEAT`

nvflare/private/fed/server/server_commands.py
```diff
@@ -317,6 +345,30 @@ def process(self, data: Shareable, fl_ctx: FLContext):
...
+    def process(self, data: Shareable, fl_ctx: FLContext):
+        """Called to process the Shutdown command.
+
+        Args:
+            data: process data
+            fl_ctx: FLContext
+
+        Returns: Shutdown command message
+
+        """
```
YuanTingHsieh, line 368: not shutdown command, return nothing

nvflare/private/fed/server/server_command_agent.py
```diff
@@ -12,60 +12,105 @@
...
+
+        if command:
+            with self.engine.new_context() as new_fl_ctx:
+                reply = command.process(data=data, fl_ctx=new_fl_ctx)
+                if reply is not None:
+                    return_message = new_cell_message({}, fobs.dumps(reply))
+                    return_message.set_header(MessageHeaderKey.RETURN_CODE, ReturnCode.OK)
+                else:
+                    return_message = make_reply(ReturnCode.PROCESS_EXCEPTION, "No process results", fobs.dumps(None))
+                return return_message
```
YuanTingHsieh, line 81: if no command, what will be returned?
Should we have some responses that contains error information?

nvflare/private/fed/cmi.py
```diff
@@ -0,0 +1,248 @@
...
+    ) -> dict:
+        """Send request to the job cells of other target sites.
+        Args:
+            targets (list): list of client names that the request will be sent to
+            channel (str): channel of the request
+            topic (str): topic of the request
+            request (Shareable): request
+            timeout (float): how long to wait for result. 0 means fire-and-forget
+            fl_ctx (FLContext): the FL context
+            bulk_send: whether to bulk send this request
```
YuanTingHsieh, line 172: this `bulk_send` arg is only checked if `timeout` is 0.
So maybe we improve the docstring to say this is only checked in fire-and-forget, on the other note only checked in non-blocking situation.

nvflare/private/fed/server/run_manager.py
yhwen: EngineSpec defines the operations which needed for both server and client sides, like validate_targets(). Server / Client have its own spec to define the operations which needed for parent and child process. The interface hierarchy is needed. We may name the interface a little differently, but keep the naming for now.

nvflare/private/fed/client/client_engine.py
yhwen: There's no use case for the client parent process to validate_targets().

nvflare/private/fed/client/client_run_manager.py
yhwen: We can make the changes for that in a separate effort. That will require changes to a lot of places. The current implementation is based on the clients can talk to multiple servers. That seems an over-engineering, since we don't really have that requirement.

nvflare/private/fed/utils/decomposers/private_decomposers.py
yhwen: RunInfo has a circulate import issue before. Extract the RunInfo to a separate run_info.py solved the problem.

nvflare/private/fed/client/client_app_runner.py
yhwen: refactored

nvflare/private/fed/server/server_engine.py
yhwen: Added timeout control.

nvflare/private/fed/server/server_engine.py
yhwen: no need any more. Removed.

nvflare/private/fed/server/server_commands.py
yhwen: RESET_ERRORS does not return value. Removed.

nvflare/private/fed/server/server_commands.py
yhwen: Cleaned.

nvflare/private/fed/server/server_commands.py
yhwen: cleaned.

nvflare/private/fed/server/server_commands.py
yhwen: cleaned

nvflare/private/fed/cmi.py
yhwen: Added docstring explanation.

nvflare/private/fed/server/sys_cmd.py
yhwen: Added back

YuanTingHsieh (approved):
Thanks for the PR. Overall LGTM.
Minor improvements we can do in the future

PR #1390, opened 2023-02-18
Async HTTP Driver for FCI [skip ci]

nvidianz:
Merged async HTTP driver to dev

### Description

1. Merged http driver to dev
2. Added test cases for it
3. Put __version__ back to nvflare/__init__.py

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/3rdParty/websockets.LICENSE.txt
+++ b/3rdParty/websockets.LICENSE.txt
@@ -1,0 +1,26 @@
+BSD License: https://pypi.org/project/websockets/
+
+Copyright (c) Aymeric Augustin and contributors
+
+Redistribution and use in source and binary forms, with or without
+modification, are permitted provided that the following conditions are met:
+
+    * Redistributions of source code must retain the above copyright notice,
+      this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above copyright notice,
+      this list of conditions and the following disclaimer in the documentation
+      and/or other materials provided with the distribution.
+    * Neither the name of the copyright holder nor the names of its contributors
+      may be used to endorse or promote products derived from this software
+      without specific prior written permission.
+
+THIS SOFTWARE IS PROVIDED BY THE COPYRIGHT HOLDERS AND CONTRIBUTORS "AS IS" AND
+ANY EXPRESS OR IMPLIED WARRANTIES, INCLUDING, BUT NOT LIMITED TO, THE IMPLIED
+WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A PARTICULAR PURPOSE ARE
+DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT HOLDER OR CONTRIBUTORS BE LIABLE
+FOR ANY DIRECT, INDIRECT, INCIDENTAL, SPECIAL, EXEMPLARY, OR CONSEQUENTIAL
+DAMAGES (INCLUDING, BUT NOT LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR
+SERVICES; LOSS OF USE, DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER
+CAUSED AND ON ANY THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY,
+OR TORT (INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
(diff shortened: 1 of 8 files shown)


nvflare/fuel/f3/drivers/aio_http_driver.py
```diff
@@ -0,0 +1,192 @@
...
+        self.conn_props = self._get_socket_properties()
+
+    def get_conn_properties(self) -> dict:
+        return self.conn_props
+
+    def close(self):
+        self.closing = True
+        self.aio_context.run_coro(self.websocket.close())
+
+    def send_frame(self, frame: BytesAlike):
```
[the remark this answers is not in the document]
nvidianz: This name is part of the interface.

It turns out the queue is not needed for websocket. Removed. The queue is added for flow control. Websocket already does that.

nvflare/fuel/f3/drivers/aio_http_driver.py
```diff
@@ -0,0 +1,192 @@
...
+
+    @staticmethod
+    async def _reader(conn: WsConnection):
+        while not conn.closing:
+            # Reading from websocket and call receiver CB
+            frame = await conn.websocket.recv()
+            conn.process_frame(frame)
+
+    @staticmethod
+    async def _writer(conn: WsConnection):
```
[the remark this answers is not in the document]
nvidianz: The queue is removed.

setup.cfg
```diff
@@ -32,6 +32,7 @@ install_requires =
     six>=1.15.0
     msgpack==1.0.3
     docker>=6.0
+    websockets
```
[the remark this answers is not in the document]
nvidianz: Fixed.

nvflare/fuel/f3/drivers/aio_http_driver.py
```diff
@@ -0,0 +1,184 @@
...
+        self.closing = False
+        self.secure = secure
+        self.conn_props = self._get_socket_properties()
+
+    def get_conn_properties(self) -> dict:
+        return self.conn_props
+
+    def close(self):
+        self.closing = True
+        self.aio_context.run_coro(self.websocket.close())
```
YuanTingHsieh, line 52: we don't need to call `result` for this one?
When should we call result when should we not call?

nvflare/fuel/f3/drivers/aio_http_driver.py
```diff
@@ -0,0 +1,184 @@
...
+
+    def send_frame(self, frame: BytesAlike):
+        self.aio_context.run_coro(self._async_send_frame(frame))
```
YuanTingHsieh, line 55: we don't need to call `result` for this one?
When should we call result when should we not call?

nvflare/fuel/f3/drivers/aio_http_driver.py
nvidianz: result() will block till the result is back. No need to wait for result here.

nvflare/fuel/f3/drivers/aio_http_driver.py
nvidianz: This one can't call result(). It causes deadlock of SFM handshake. This is kind of a bug in SFM. We need to fix it so non-block send_frame is not required.

nvflare/fuel/f3/drivers/aio_http_driver.py
```diff
@@ -0,0 +1,184 @@
...
+        self.close_all()
+
+        if self.stop_event:
+            self.stop_event.set_result(None)
+
+    @staticmethod
+    def get_urls(scheme: str, resources: dict) -> (str, str):
+        secure = resources.get(DriverParams.SECURE)
+        if secure:
+            scheme = "https"
```
yhwen, line 124: If user define the scheme as "http", but secure == True, this will implicitly change the scheme to "https". Will this cause the confusion to the user? Maybe explicitly provides the warning message to the user?

nvidianz: We have to 2 ways to specify the secure connection, using "https" or "secure=True". Those are all normal use-cases. a warning probably causes more confusion.

yhwen: what happens when user set scheme to "https" and "secure" to "False"?

nvidianz: It will be secure.

3rdParty/websockets.LICENSE.txt
```diff
@@ -0,0 +1,24 @@
+Copyright (c) Aymeric Augustin and contributors
+
```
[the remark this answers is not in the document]
nvidianz: According to Pypi, it's BSD. Updated the license file.
https://pypi.org/project/websockets/

PR #1399, opened 2023-02-22
Fixed Close CB bug and added socket cleanup [skip ci]

nvidianz:
### Description

1. Fixed the bug that causes SFM closed_cb not invoked.
2. Added cleanup code to remove useless socket files

### Types of changes

- [ ] Documentation updated.


yhwen (dismissed):

nvflare/fuel/f3/drivers/uds_driver.py
```diff
@@ -125,3 +124,13 @@ def get_socket_path(params: dict) -> str:
             socket_path = "/" + socket_path
 
         return socket_path
+
+    @staticmethod
+    def remove_socket_file(socket_path: str) -> bool:
+        try:
+            os.unlink(socket_path)
+        except OSError as error:
+            log.debug(f"Removing {socket_path} : {error}")
```
YuanTingHsieh, line 133: This log here, should we say `f"Error happens when removing {socket_path} : {error}"`

nvflare/fuel/f3/drivers/uds_driver.py
```diff
@@ -86,6 +84,7 @@ def connect(self, connector: ConnectorInfo):
         self.add_connection(connection)
         connection.read_loop()
         self.close_connection(connection)
+        self.remove_socket_file(socket_path)
```
YuanTingHsieh, line 87: Will both the "connector" side and "listener" side try to remove the same socket file at the same time?

nvflare/fuel/f3/drivers/uds_driver.py
```diff
@@ -65,14 +65,12 @@ def listen(self, connector: ConnectorInfo):
...
-            os.unlink(socket_path)
-        except OSError:
-            if os.path.exists(socket_path):
-                raise CommError(CommError.ERROR, f"Can't remove existing socket file: {socket_path}")
+        if not self.remove_socket_file(socket_path):
+            raise CommError(CommError.ERROR, f"Can't remove existing socket file: {socket_path}")
 
         self.server = SocketStreamServer(socket_path, self, connector)
         self.server.serve_forever()
+        self.remove_socket_file(socket_path)
```
YuanTingHsieh, line 73: Will both the "connector" side and "listener" side try to remove the same socket file at the same time?

nvflare/fuel/f3/drivers/uds_driver.py
nvidianz: The socket file can only be removed when both sides close the socket so one side is going to fail. This shouldn't be treated as an error.

nvflare/fuel/f3/drivers/uds_driver.py
nvidianz: It may happen. Only last one closes the connection can remove it. This could be either side so both sides try to close. Only one side will succeed.

nvflare/fuel/f3/drivers/uds_driver.py
nvidianz: It may happen. Only the last one who closes the connection can remove the file, which can be either side. So both sides try to remove the file. Only one will succeed.

nvflare/fuel/f3/drivers/aio_grpc_driver.py
```diff
@@ -120,7 +124,7 @@ async def generate_output(self):
                 yield item
```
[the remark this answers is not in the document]
nvidianz: item is retrieved from output queue in the previous line.

nvflare/fuel/f3/drivers/aio_grpc_driver.py
```diff
@@ -120,7 +124,7 @@ async def generate_output(self):
...
         except Exception as ex:
             if self.closing:
-                self.logger.debug(f"{self}: connected closed by {type(ex)}: {ex}")
+                self.logger.debug(f"{self}: connection closed by {type(ex)}: {ex}")
             else:
                 self.logger.error(f"{self}: generate_output exception {type(ex)}: {ex}")
```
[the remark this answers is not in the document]
nvidianz: No. The exceptions are all logged. Just changed level to debug if closing to clean up the log.

YuanTingHsieh (approved):
LGTM.

PR #1402, opened 2023-02-23
update README for hello-pt on model initialization [skip ci]

chesterxgchen:
Fixes # .

### Description
 hello-pt example demonstrated new client-side model initialization feature. We update the readme to discuss the different global model initialization approaches. Also explained, in high-level, the reason for using client-side model initialization and pro and cons of the side-side and client-side model initialization approaches 

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/examples/hello-world/hello-pt/README.md
+++ b/examples/hello-world/hello-pt/README.md
@@ -41,7 +41,8 @@ The advantage of server-side model initialization is that the model only needs t
 (the server), and then distributed to all clients, ensuring that all clients have the same initial model. 
 However, it is important to note that server-side model initialization may present potential security risks 
 if custom Python code is run on the server. An alternative approach for server-side initialization is to use 
-a predefined model file as the initialization model. The ScatterAndGather controller is using persistor to reads / init model from server-side. 
+a predefined model file as the initialization model. The ScatterAndGather controller is using persistor to reads / init
+model from server-side. 
 
 In this example, we are using **client-side** model initialization approach. 
 
@@ -89,8 +90,11 @@ Looking at the job workflow, we have defined three workflows in config_fed_serve
   ]
 ```
 
-Once the global model is initialized, it is set to the fl_ctx as property and then pass to the 
-next controller (Scatter and Gather) in the training step. 
+Once the global model is initialized, it is set to the fl_ctx as sticky property and then pass to the 
+next controller (ScatterAndGather) in the training step. The sticky property allows properties pass cross controllers. 
+The ScatterAndGather still leverage _persistor_ to load the initial global model, but since there is no model file 
+or server-side initialized model, the ScatterAndGather then try to load the model from fl_ctx's _"global_model"_ property, 
+which is initialized from the client-side and set by the previous controller in the workflow. 
  
 
 


examples/hello-world/hello-pt/README.md
```diff
@@ -30,3 +30,66 @@ $ ls /tmp/nvflare/simulate_job/
 app_server  app_site-1  app_site-2  log.txt
 
 ```
+
+### 4. Global Model Initialization Approaches
+
+There are multiple approaches to initialize a global model, either on FL server side or on the client side. Users can decide which model initialization approaches depends on their own use cases.
```
YuanTingHsieh, line 36: suggest paste contents in either ChatGPT or grammarly for grammar fixes.

examples/hello-world/hello-pt/README.md
```diff
@@ -30,3 +30,66 @@ $ ls /tmp/nvflare/simulate_job/
...
+
+There are multiple approaches to initialize a global model, either on FL server side or on the client side. Users can decide which model initialization approaches depends on their own use cases. 
+
+When the model is initialized at the FL **server-side**, the initialized global model will then be broadcast to all the FL clients.
+Clients will take this initial model and start training. The benefits of the server-side model initialization 
+allows the model to only initialize once in one place (server) and then distributed to all clients, 
+so all clients have the same initial model. The potential issue with server-side model initialization might
+involve security risk of running custom-python code on server. leveraging a predefined model file as initialization model can be another server-side approach. 
+
+The ScatterAndGather controller is using server-side model initialization.
```
YuanTingHsieh, line 44: The `ScatterAndGather` controller is using `persistor` which reads / init model from server side that is correct.

But in this example, we are using `InitializeGlobalWeights` and then save it somewhere that can be loaded by this `persistor` so the next `ScatterAndGather` will pick it up.

Simply put this line here might be confusing to users

YuanTingHsieh: @holgerroth do we know if the weights that are loaded / init by `InitializeGlobalWeights` are actually being used in `ScatterAndGather`?

holgerroth: Yes, in this example, the persistor doesn't initialize the model:

examples/hello-world/hello-pt/README.md
```diff
@@ -30,3 +30,68 @@ $ ls /tmp/nvflare/simulate_job/
...
+        "name": "CrossSiteModelEval",
+        "args": {
+          "model_locator_id": "model_locator"
+        }
+      }
+  ]
+```
+
+Once the global model is initialized, it is set to the fl_ctx as property and then pass to the 
+next controller (Scatter and Gather) in the training step.
```
YuanTingHsieh, line 93: Should we just make it `ScatterAndGather`?

One final note here, we said it upfront that `The ScatterAndGather controller is using persistor to reads / init model from server-side.`

The real scenario is `ScatterAndGather` will first try to use persistor to reads / init model from server-side. 

Then it will look for fl_ctx.get_prop(AppConstants.GLOBAL_WEIGHTS)

holgerroth (approved):
Looks good.

YuanTingHsieh (approved):
Thanks, LGTM

YuanTingHsieh (approved):

PR #1403, opened 2023-02-23
fix new_insecure_session

nvkevlu:
Fixes issue with starting insecure session with flare_api by adding the username for the call to Session.

### Types of changes

- [ ] Documentation updated.


kkersten (approved):
This fixes the issue.  Looks good.

PR #1416, opened 2023-02-24
fix default order of jobs in list_jobs command

nvkevlu:
Fixes the default ordering of jobs being returned in the list_jobs command to be the last submitted jobs instead of the first submitted jobs.

### Types of changes

- [ ] Documentation updated.


nvidianz (approved):

nvidianz (commented):
Waiting for more changes.

nvkevlu:
New changes take into account Chester's comments about not changing the existing behavior and only filtering with -m included.

nvflare/private/fed/server/job_cmds.py
```diff
@@ -288,11 +288,15 @@ def list_jobs(self, conn: Connection, args: List[str]):
...
+                help="List all jobs, default is filtered to jobs submitted by the same user",
             )
             parser.add_argument("-n", help="Filter by job name prefix")
             parser.add_argument(
-                "-m", default=5, help="Limit maximum number of jobs returned to the specified number, default is 5"
+                "-m",
+                nargs="?",
+                const=5,
+                type=int,
+                help="Limit maximum number of jobs returned to the specified number, default is 5 if flag is set but there is no integer provided for the specified limit",
```
nvidianz, line 299: Should try to avoid using nargs="?". This doesn't follow getopts convention and it can be ambiguous. For example, this probably throws an exception,

   list_jobs -m job_id

nvkevlu: Modified to not use nargs="?".

nvidianz (approved):

PR #1420, opened 2023-02-24
Hide cell cmds

yanchengnv:
F3/Cellnet introduces many new admin commands for testing and analyzing the cell performance of the cell network. However these commands are not ready to be publicized. This PR makes such commands invisible or disabled:
- If a command could cause interruptive changes to the system (e.g. stop_net, stop_cell, etc), it is disabled
- Otherwise the command is made invisible to end user. The command is still accessible to the user.

Changes are in the following areas:
- Admin server is changed to send invisible commands to the client with visibility info. Internal commands are not sent.
- Admin client hides invisible commands from the user, but they are available to the user if the user knows the command name.
- The NetManager (which implements these commands) can be used in "for_test" mode. In this mode, all commands are enabled and visible. This is only used for independent cellnet testing.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/fuel/f3/cellnet/net_manager.py
+++ b/nvflare/fuel/f3/cellnet/net_manager.py
@@ -29,12 +29,11 @@ def _to_int(s: str):
 
 
 class NetManager(CommandModule):
-    def __init__(self, agent: NetAgent, for_test=False):
+    def __init__(self, agent: NetAgent, diagnose=False):
         self.agent = agent
-        self.for_test = for_test
+        self.diagnose = diagnose
 
     def get_spec(self) -> CommandModuleSpec:
-        visible = self.for_test
         return CommandModuleSpec(
             name="cellnet",
             cmd_specs=[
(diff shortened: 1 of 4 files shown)


nvflare/fuel/f3/cellnet/net_manager.py
```diff
@@ -29,10 +29,12 @@ def _to_int(s: str):
 
 
 class NetManager(CommandModule):
-    def __init__(self, agent: NetAgent):
+    def __init__(self, agent: NetAgent, for_test=False):
```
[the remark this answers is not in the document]
yanchengnv: Good idea. Will change.

nvflare/fuel/f3/cellnet/net_manager.py
```diff
@@ -29,8 +29,9 @@ def _to_int(s: str):
 
 
 class NetManager(CommandModule):
-    def __init__(self, agent: NetAgent):
+    def __init__(self, agent: NetAgent, diagnose=False):
```
yhwen, line 32: How do we turn this on in a live system?

yanchengnv: We don't, since it will turn on some dangerous commands like stop_net, stop_cell, etc.

yhwen: We need to make it configurable in server/admin.py to turn it on / off when creating the NetManager? Otherwise, we may not be able to use this?

nvflare/fuel/hci/client/cli.py
```diff
@@ -237,7 +237,7 @@ def do_help(self, arg):
                 self.write_string("Server Commands")
                 self.write_string("---------------")
                 for cmd_name in server_cmds:
-                    self._show_one_command(cmd_name, self.api.server_cmd_reg)
+                    self._show_one_command(cmd_name, self.api.server_cmd_reg, show_invisible=True)
```
YuanTingHsieh, line 240: why do we want to show invisible commands here?

YuanTingHsieh: My question here should be, will we somehow show invisible commands even if the command name not 100% match?

yanchengnv: This is used when user types "help cmd", and the cmd help info is shown only if cmd name matches.
Invisible only means to hide the command from the command list shown to the user. The command is still active, and the user can use it if he knows the name.

nvflare/fuel/f3/cellnet/net_manager.py
yanchengnv: Commands are invisible but still available to users.

nvidianz (approved):

nvflare/fuel/f3/cellnet/net_manager.py
yanchengnv: For FLARE, NetManager will always run in non-diagnose mode, which will disable dangerous commands (stop_cell, stop_net), and make other useful commands invisible (but still accessible).

The NetManager will be used in diagnose mode only in the independent cellnet testing tools (QAT stuff).

PR #1423, opened 2023-02-24
Changed the fetch_task fetch_again without delay.

yhwen:
Fixes # .

### Description

Changed the client fetch_task fetch_again behavior to without delay. Client Runner handles the task_fetch_interval.

### Types of changes

- [ ] Documentation updated.


nvidianz (approved):

yanchengnv (approved):
This is good. Since client_runner already handles task fetch interval, there is no need to add this extra delay.

PR #1440, opened 2023-02-28
Enhance the job run process not to kill its own process, instead let …

yhwen:
Fixes # .

Fix the random job need to wait for 3600 seconds to shutdown, then start the 2nd job in CI issue.

### Description

Instead of monitor the job finish and kill its own process, send the stop() to the Runner to stop the main loop. Then let the process shutdown to MPM to manage.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/private/fed/app/simulator/simulator_runner.py
+++ b/nvflare/private/fed/app/simulator/simulator_runner.py
@@ -69,7 +69,7 @@ class SimulatorRunner(FLComponent):
         self.ask_to_stop = False
 
         self.simulator_root = None
-        self.services = None
+        self.server = None
         self.deployer = SimulatorDeployer()
         self.client_names = []
         self.federated_clients = []
@@ -197,7 +197,7 @@ class SimulatorRunner(FLComponent):
 
             # Deploy the FL server
             self.logger.info("Create the Simulator Server.")
-            simulator_server, self.services = self.deployer.create_fl_server(self.args)
+            simulator_server, self.server = self.deployer.create_fl_server(self.args)
             # self.services.deploy(self.args, grpc_args=simulator_server)
 
             self.logger.info("Deploy the Apps.")
@@ -328,12 +328,12 @@ class SimulatorRunner(FLComponent):
         if self.setup():
             try:
                 self.create_clients()
-                self.services.engine.run_processes[SimulatorConstants.JOB_NAME] = {
+                self.server.engine.run_processes[SimulatorConstants.JOB_NAME] = {
                     RunProcessKey.LISTEN_PORT: None,
                     RunProcessKey.CONNECTION: None,
                     RunProcessKey.CHILD_PROCESS: None,
                     RunProcessKey.JOB_ID: SimulatorConstants.JOB_NAME,
-                    RunProcessKey.PARTICIPANTS: self.services.engine.client_manager.clients,
+                    RunProcessKey.PARTICIPANTS: self.server.engine.client_manager.clients,
                 }
 
                 self.logger.info("Deploy and start the Server App.")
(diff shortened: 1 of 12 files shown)


nvidianz (approved):

nvflare/private/fed/runner.py
```diff
@@ -0,0 +1,22 @@
...
+# limitations under the License.
+
+
+class Runner:
+    def start(self):
+        """Method call at the start of the Runner process."""
+        pass
+
+    def stop(self):
+        """Method call at the end of the Runner process."""
```
YuanTingHsieh, line 22: If only `stop` method is needed, suggest

```
class Runner(ABC):
    @abstractmethod
    def stop(self):
        """Stops the runner."""
        pass
```

nvflare/private/fed/simulator/simulator_app_runner.py
```diff
@@ -40,6 +40,9 @@ def create_run_manager(self, args, conf, federated_client, workspace):
 
 
 class SimulatorServerAppRunner(ServerAppRunner):
+    def __init__(self, server) -> None:
+        super().__init__(server)
+
```
YuanTingHsieh, line 45: if `SimulatorServerAppRunner` share the same init method as parent class, maybe we don't need this part.

nvflare/private/fed/server/server_app_runner.py
```diff
@@ -74,3 +79,6 @@ def sync_up_parents_process(self, args, server):
 
     def update_job_run_status(self, server):
         server.engine.update_job_run_status()
+
+    def stop(self):
+        self.server.engine.asked_to_stop = True
```
YuanTingHsieh, line 84: I prefer have some kind of method introduced instead of directly access other class's variable.

nvflare/private/fed/server/server_app_runner.py
```diff
@@ -42,7 +43,11 @@ def _set_up_run_config(workspace: Workspace, server, conf):
     server.handlers = conf.handlers
 
 
-class ServerAppRunner:
+class ServerAppRunner(Runner):
+    def __init__(self, server) -> None:
+        super().__init__()
+        self.server = server
```
YuanTingHsieh, line 49: If we have this in init method, maybe other method of this class don't need to take "server" as input argument anymore?

nvflare/private/fed/app/simulator/simulator_runner.py
```diff
@@ -398,7 +398,7 @@ def start_server_app(self):
             False,
             None,
         )
-        server_app_runner = SimulatorServerAppRunner()
+        server_app_runner = SimulatorServerAppRunner(self.services)
```
YuanTingHsieh, line 401: since in this class `self.services` just mean 1 FL server, I prefer we rename it to self.server or self.fl_server something like that, but that can be done in a separate PR

nvflare/private/fed/runner.py
yhwen: changed.

nvflare/private/fed/simulator/simulator_app_runner.py
yhwen: This maybe not necessary. Does not hurt.

nvflare/private/fed/server/server_app_runner.py
yhwen: just a simple line, maybe a Python way.

nvflare/private/fed/server/server_app_runner.py
yhwen: changed.

nvflare/private/fed/app/simulator/simulator_runner.py
yhwen: Done.

YuanTingHsieh (approved):

PR #1447, opened 2023-03-01
Fixed the issue job status not updated to exception when controller e…

yhwen:
Fixes # .

Fixed the issue when the controller raise exception, but job status not updated as execution exception.

### Description

Updated the job status as FINISHED:EXECUTION_EXCEPTION when controller raise exception.

### Types of changes

- [ ] Documentation updated.


nvflare/private/fed/server/fed_server.py
```diff
@@ -336,12 +336,13 @@ def _listen_command(self, request: Message) -> Message:
...
-            if execution_error:
-                with self.lock:
-                    run_process_info = self.engine.run_processes.get(job_id)
+            with self.lock:
+                run_process_info = self.engine.run_processes.get(job_id)
+                if execution_error:
                     self.engine.exception_run_processes[job_id] = run_process_info
-                    reply = make_cellnet_reply(F3ReturnCode.OK, "", None)
-                    return reply
+                run_process_info[RunProcessKey.PROCESS_FINISHED] = True
```
YuanTingHsieh, line 343: This `ServerCommandNames.UPDATE_RUN_STATUS` will only be called when a job is finished?

nvflare/private/fed/server/fed_server.py
```diff
@@ -336,12 +336,13 @@ def _listen_command(self, request: Message) -> Message:
             return make_cellnet_reply(F3ReturnCode.OK, "", fobs.dumps(return_data))
         elif command == ServerCommandNames.UPDATE_RUN_STATUS:
             execution_error = data.get("execution_error")
-            if execution_error:
-                with self.lock:
-                    run_process_info = self.engine.run_processes.get(job_id)
+            with self.lock:
```
YuanTingHsieh, line 339: This `ServerCommandNames.UPDATE_RUN_STATUS` will only be called when a job is finished?

nvflare/private/fed/server/server_engine.py
```diff
@@ -206,6 +206,16 @@ def remove_exception_process(self, job_id):
 
     def wait_for_complete(self, job_id, process):
         process.wait()
+        run_process_info = self.run_processes.get(job_id)
+        process_finished = run_process_info.get(RunProcessKey.PROCESS_FINISHED, False)
```
YuanTingHsieh, line 210: this line `self.run_processes.get(job_id)` can return None.
then the following line will error.

nvflare/private/fed/server/fed_server.py
yhwen: Yes

nvflare/private/fed/server/server_engine.py
yhwen: we only have one place to pop the "job_id" which is after the code here. We won't have the case of None.

nvflare/private/fed/server/fed_server.py
yhwen: self.engine.run_processes[job_id] already in the run_process_info variable. The developer read the codes should know that.

nvflare/private/fed/server/server_engine.py
YuanTingHsieh: We have other places that will pop this job_id, for example: https://github.com/NVIDIA/NVFlare/blob/dev/nvflare/private/fed/server/server_engine.py#L319-L324

yhwen: Added the run_process_info None check.

nvflare/private/fed/server/server_engine.py
```diff
@@ -206,13 +206,24 @@ def remove_exception_process(self, job_id):
...
-                # # if process exit but with Execution exception
-                # if return_code and return_code != 0:
-                #     self.exception_run_processes[job_id] = run_process_info
+        run_process_info = self.run_processes.get(job_id)
+        if run_process_info:
+            process_finished = run_process_info.get(RunProcessKey.PROCESS_FINISHED, False)
+            # Wait for the job process to finish UPDATE_RUN_STATUS process
+            start_time = time.time()
+            while not process_finished:
+                if time.time() - start_time > 5.0:
```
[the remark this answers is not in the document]
yhwen: We can schedule this task to a particular release to address. We can make use of the Config_Server for this refactor.

nvflare/private/fed/server/server_engine.py
```diff
@@ -206,13 +206,24 @@ def remove_exception_process(self, job_id):
...
+                    self.logger.error(f"Job:{job_id} UPDATE_RUN_STATUS didn't finish fast enough.")
+                    break
+                time.sleep(0.1)
+
+            with self.lock:
+                if job_id in self.run_processes:
+                    run_process_info = self.run_processes.pop(job_id)
```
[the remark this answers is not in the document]
yhwen: Done

PR #1450, opened 2023-03-02
Randomize resource group to avoid duplicate resource group names [skip ci]

IsaacYangSLA:
### Description

Randomize the resource group names so users can run dashboard more than one time and the resource names will be be duplicate.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/lighter/impl/master_template.yml
+++ b/nvflare/lighter/impl/master_template.yml
@@ -1305,7 +1305,7 @@ azure_start_dsb_sh: |
   fi
   }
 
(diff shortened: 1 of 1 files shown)


IsaacYangSLA:
This is for dashboard.  The last one is for clients.  It may be likely that users launch multiple clients under the same subscription, thus client scripts should consider such use case.  I was not 100% sure about the dashboard case.  But since people may still run dashboard multiple times, this PR should help avoid resource group name collision.

PR #1456, opened 2023-03-06
Check if resource group exists.

IsaacYangSLA:
### Description

In Azure cloud scripts, add `az group exists` to find if the resource group name exists.  If it exists, the resource group will not be created.  Instead, the same resource group is reused for following operations.

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh (approved):
Thanks LGTM.
Just need to make sure our documentation mention this behavior

PR #1458, opened 2023-03-06
Fix admin API issues and support optional messages

yanchengnv:
- Fix an issue of "The response of do_command APIStatus is not aligned with its meta status."
- Changed to use secure logging for exceptions in cellnet
- Set API Status properly for admin API
- Added support of "optional" messages. This will be used for generating clean log messages during shutdown process
- Changed watching thread to daemon in dummy overseer agent, to avoid unnecessary delay in shutdown process

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh:
@yanchengnv can you try sign your commits?

PR #1466, opened 2023-03-08
Fix job runner multiple start issue

YuanTingHsieh:
In the old codes, when we start more than 1 server, all of the servers will start JobRunner by default and it keeps reading and updating the job storage at the same time which might cause racing and other issues.

We change "run" of JobRunner to wait for server state to become HOT

### Description

- JobRunner will not proceed until server turns to HOT state

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/private/fed/server/fed_server.py
+++ b/nvflare/private/fed/server/fed_server.py
@@ -271,9 +271,8 @@ class FederatedServer(BaseServer):
         self.tokens = None
         self.round_started = time.time()
 
-        self.tokens = dict()
-        for token, client in self.get_all_clients().items():
-            self.tokens[token] = self.task_meta_info(client.name)
+        with self.lock:
+            self.reset_tokens()
 
         self.cmd_modules = cmd_modules
 
--- a/nvflare/private/fed/server/fed_server.py
+++ b/nvflare/private/fed/server/fed_server.py
@@ -205,7 +205,7 @@ class BaseServer(ABC):
         return client
 
     def notify_dead_client(self, client):
-        """Called to do further processing of the dead client.
+        """Called to do further processing of the dead client
 
         Args:
             client: the dead client
@@ -638,8 +638,14 @@ class FederatedServer(BaseServer):
             self.logger.error(f"FL server execution exception: {secure_format_exception(e)}")
         finally:
             self.engine.update_job_run_status()
+            self.stop_run_engine_cell()
 
         self.engine.engine_info.status = MachineStatus.STOPPED
+
+    def stop_run_engine_cell(self):
+        # self.cell.stop()
+        # mpm.stop()
+        pass
 
     def deploy(self, args, grpc_args=None, secure_train=False):
         super().deploy(args, grpc_args, secure_train)
@@ -697,11 +703,11 @@ class FederatedServer(BaseServer):
         with self.snapshot_lock:
             fl_snapshot = self.snapshot_persistor.retrieve()
             if fl_snapshot:
-                for job_id, snapshot in fl_snapshot.run_snapshots.items():
+                for run_number, snapshot in fl_snapshot.run_snapshots.items():
                     if snapshot and not snapshot.completed:
                         # Restore the workspace
                         workspace_data = snapshot.get_component_snapshot(SnapshotKey.WORKSPACE).get("content")
-                        dst = os.path.join(self.workspace, WorkspaceConstants.WORKSPACE_PREFIX + str(job_id))
+                        dst = os.path.join(self.workspace, WorkspaceConstants.WORKSPACE_PREFIX + str(run_number))
                         if os.path.exists(dst):
                             shutil.rmtree(dst, ignore_errors=True)
 
@@ -710,9 +716,10 @@ class FederatedServer(BaseServer):
 
                         job_id = snapshot.get_component_snapshot(SnapshotKey.JOB_INFO).get(SnapshotKey.JOB_ID)
                         job_clients = snapshot.get_component_snapshot(SnapshotKey.JOB_INFO).get(SnapshotKey.JOB_CLIENTS)
-                        self.logger.info(f"Restore the previous snapshot. JOB ID: {job_id}")
+                        self.logger.info(f"Restore the previous snapshot. Run_number: {run_number}")
                         with self.engine.new_context() as fl_ctx:
                             self.engine.job_runner.restore_running_job(
+                                run_number=run_number,
                                 job_id=job_id,
                                 job_clients=job_clients,
                                 snapshot=snapshot,
--- a/nvflare/private/fed/server/fed_server.py
+++ b/nvflare/private/fed/server/fed_server.py
@@ -19,8 +19,9 @@ import threading
 import time
 from abc import ABC, abstractmethod
 from threading import Lock
-from typing import List, Optional
+from typing import Dict, List, Optional
 
+from nvflare.apis.client import Client
 from nvflare.apis.event_type import EventType
 from nvflare.apis.fl_component import FLComponent
 from nvflare.apis.fl_constant import (
@@ -87,13 +88,11 @@ class BaseServer(ABC):
 
         self.heart_beat_timeout = heart_beat_timeout
         self.handlers = handlers
-        # self.cmd_modules = cmd_modules
 
         self.client_manager = ClientManager(
             project_name=self.project_name, min_num_clients=self.min_num_clients, max_num_clients=self.max_num_clients
         )
 
-        # self.grpc_server = None
         self.cell = None
         self.admin_server = None
         self.lock = Lock()
@@ -110,7 +109,12 @@ class BaseServer(ABC):
 
         self.logger = logging.getLogger(self.__class__.__name__)
 
-    def get_all_clients(self):
+    def get_all_clients(self) -> Dict[str, Client]:
+        """Get the list of registered clients.
+
+        Returns:
+            A dict of {client_token: client}
+        """
         return self.client_manager.get_clients()
 
     @abstractmethod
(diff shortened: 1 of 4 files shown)


nvidianz (dismissed):

nvflare/private/fed/server/fed_server.py
```diff
@@ -265,16 +223,14 @@ def __init__(
         self.tokens = None
         self.round_started = time.time()
 
-        with self.lock:
-            self.reset_tokens()
+        self.tokens = dict()
+        for client in self.get_all_clients().keys():
```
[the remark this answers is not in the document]
YuanTingHsieh: key is token, value is Client object, let me change this.

nvflare/private/fed/app/deployer/server_deployer.py
```diff
@@ -62,70 +61,65 @@ def create_fl_server(self, args, secure_train=False):
...
         # We only deploy the first server right now .....
-        first_server = sorted(self.server_config)[0]
-        heart_beat_timeout = first_server.get("heart_beat_timeout", 600)
+        first_server_config = sorted(self.server_config)[0]
+        heart_beat_timeout = first_server_config.get("heart_beat_timeout", 600)
 
         if self.host:
-            target = first_server["service"].get("target", None)
-            first_server["service"]["target"] = self.host + ":" + target.split(":")[1]
+            target = first_server_config["service"].get("target", None)
```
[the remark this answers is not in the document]
YuanTingHsieh: I remove self.host as we are not using it.

nvflare/private/fed/server/job_runner.py
```diff
@@ -290,10 +293,9 @@ def _delete_run(self, job_id, client_sites: List[str], fl_ctx: FLContext):
     def _job_complete_process(self, fl_ctx: FLContext):
         engine = fl_ctx.get_engine()
         job_manager = engine.get_component(SystemComponents.JOB_MANAGER)
-        while not self.ask_to_stop:
+        while not self.ask_to_stop and isinstance(engine.server.server_state, HotState):
```
yanchengnv: I don't think we need server state to be hot here since it is trying to complete jobs that are already running.

nvflare/private/fed/server/job_runner.py
```diff
@@ -328,28 +330,32 @@ def _save_workspace(self, fl_ctx: FLContext):
         shutil.rmtree(run_dir)
 
     def run(self, fl_ctx: FLContext):
+        """Starts job runner."""
         engine = fl_ctx.get_engine()
-
```
yanchengnv, line 332: It is okay to start the thread here.

nvflare/private/fed/server/job_runner.py
```diff
@@ -328,28 +330,32 @@ def _save_workspace(self, fl_ctx: FLContext):
...
-        thread = threading.Thread(target=self._job_complete_process, args=[fl_ctx])
-        thread.start()
-
         job_manager = engine.get_component(SystemComponents.JOB_MANAGER)
         if job_manager:
-            while not self.ask_to_stop:
-                # approved_jobs = job_manager.get_jobs_by_status(RunStatus.APPROVED, fl_ctx)
+            while not isinstance(engine.server.server_state, HotState):
```
yanchengnv: We shouldn't have to wait for the server to become hot here.

nvflare/private/fed/server/job_runner.py
```diff
@@ -328,28 +330,32 @@ def _save_workspace(self, fl_ctx: FLContext):
...
 
                     if ready_job:
-                        # with self.lock:
                         client_sites = {k: v for k, v in sites.items() if k != "server"}
                         job_id = None
                         try:
-                            self.log_info(fl_ctx, f"Got the job:{ready_job.job_id} from the scheduler to run")
+                            self.log_info(fl_ctx, f"Got the job: {ready_job.job_id} from the scheduler to run")
                             fl_ctx.set_prop(FLContextKey.CURRENT_JOB_ID, ready_job.job_id)
                             job_id, failed_clients = self._deploy_job(ready_job, sites, fl_ctx)
```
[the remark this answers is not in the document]
YuanTingHsieh: yes it is the same, but revert changes to get rid of these clean up

yanchengnv (changes requested):
Please address issues in the review.

nvflare/private/fed/server/job_runner.py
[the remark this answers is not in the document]
YuanTingHsieh: thanks, modify the logic a bit, if not HOT then continue, otherwise the loop condition the same as before.

YuanTingHsieh: thanks, changed.

nvflare/private/fed/server/fed_server.py
```diff
@@ -418,8 +408,8 @@ def reset_tokens(self):
         This function is not thread-safe.
         """
         self.tokens = dict()
-        for client in self.get_all_clients().keys():
-            self.tokens[client] = self.task_meta_info(client.name)
```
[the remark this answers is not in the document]
YuanTingHsieh: yes, this code was wrong, so I modify it, I will replace the thing in front with this function.

nvflare/private/fed/server/job_runner.py
YuanTingHsieh: The job_complete_process requires job_manager to be True,
So I think it is better to move the start after the check

nvflare/private/fed/server/fed_server.py
```diff
@@ -270,16 +271,14 @@ def __init__(
         self.tokens = None
         self.round_started = time.time()
 
-        with self.lock:
-            self.reset_tokens()
+        self.tokens = dict()
+        for token, client in self.get_all_clients().items():
```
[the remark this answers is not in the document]
YuanTingHsieh: agree, actually the lock should be in client_manager, but I will keep this code the same as before in this PR.

We can fix that in another PR.

nvflare/private/fed/server/job_runner.py
YuanTingHsieh: thanks, fixed.

nvflare/private/fed/server/fed_server.py
```diff
@@ -323,17 +319,13 @@ def _register_cellnet_cbs(self):
...
     def _listen_command(self, request: Message) -> Message:
-
-        assert isinstance(request, Message), "request must be CellMessage but got {}".format(type(request))
-
         job_id = request.get_header(CellMessageHeaderKeys.JOB_ID)
         command = request.get_header(MessageHeaderKey.TOPIC)
         data = fobs.loads(request.payload)
 
         if command == ServerCommandNames.GET_CLIENTS:
             if job_id in self.engine.run_processes:
```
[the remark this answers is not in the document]
YuanTingHsieh: this is true, OK I will open a ticket for this.

YuanTingHsieh:
@yanchengnv @chesterxgchen thanks for your review!
I fixed the issue according to the comments.
Please review again.

nvflare/private/fed/server/fed_server.py
```diff
@@ -705,17 +686,16 @@ def overseer_callback(self, overseer_agent):
...
-            self.server_state = self.server_state.handle_sd_callback(sp, fl_ctx)
+
+        with self.lock:
+            with self.engine.new_context() as fl_ctx:
+                self.server_state = self.server_state.handle_sd_callback(sp, fl_ctx)
 
         if isinstance(self.server_state, Cold2HotState):
-            server_thread = threading.Thread(target=self._turn_to_hot)
-            server_thread.start()
+            self._turn_to_hot()
```
yanchengnv, line 695: So now instead of running _turn_to_hot in a thread, you call it directly here. What could be the ramifications?
The turn_to_hot logic is complex and could take a while to finish. If running in a thread, then this overseer_callback will return quickly. This is potentially dangerous, since the CB could be called again by the overseer agent, which will create yet another thread to do state change.

So I think this is a good change to ensure that we only do one state change at a time.

yanchengnv (approved):
LGTM. Please test extensively.

nvflare/private/fed/server/job_runner.py
```diff
@@ -328,28 +330,32 @@ def _save_workspace(self, fl_ctx: FLContext):
...
-        thread.start()
-
         job_manager = engine.get_component(SystemComponents.JOB_MANAGER)
         if job_manager:
+            thread = threading.Thread(target=self._job_complete_process, args=[fl_ctx])
+            thread.start()
+
             while not self.ask_to_stop:
-                # approved_jobs = job_manager.get_jobs_by_status(RunStatus.APPROVED, fl_ctx)
+                if not isinstance(engine.server.server_state, HotState):
```
yanchengnv, line 341: Now you changed to put the check into the loop - good.

You could put the check as an "if" condition for rest of the logic to reuse the existing sleep(1.0). This is just coding style thing.

nvflare/private/fed/server/fed_server.py
YuanTingHsieh: Yes, Yan you are very correct.

I observed that when running the old code, this function gets called multiple times.
And since this method also modify the state, it is like a bunch of threads keep modifying the state (from Cold to Cold2Hot, or from Cold2Hot to Hot).
So a racing condition there.

I agree it is better to ensure only do one state change at a time.

yhwen: There's a potential issue with using the lock here. The overseer calls the overseer_callback very quickly, the default is 6 seconds. Using lock could potentially cause multiple threads piling up waiting for execution.

yanchengnv: Good point!

In this case, the overseer_callback should simply return if it is currently working on state change. This means that we only work on one state change at a time.

PR #1467, opened 2023-03-08
Randomize security group in AWS dashboard and client scripts

IsaacYangSLA:
### Description

Security groups can not be reused because the rules attached to them are specific for the VM.  Therefore, this PR randomizes the security group names to avoid reuse.

### Types of changes

- [ ] Documentation updated.


nvkevlu (approved):
Thanks, LGTM

PR #1473, opened 2023-03-09
Fixed exception in list_jobs

nvidianz:
### Description

Fixed the bug that causes exception in list_jobs in Flare API when there is no job.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/apis/fl_context.py
+++ b/nvflare/apis/fl_context.py
@@ -298,7 +298,7 @@ class FLContextManager(object):
 
         # set permanent props
         ctx.put(key=ReservedKey.ENGINE, value=self.engine, private=True, sticky=False)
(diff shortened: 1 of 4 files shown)


nvflare/fuel/flare_api/flare_api.py
```diff
@@ -300,7 +300,7 @@ def list_jobs(
                 command = command + " " + id_prefix
         result = self._do_command(command)
```
YuanTingHsieh, line 301: we prob. should also update above as default is list all jobs (the same behavior as in 2.2)

YuanTingHsieh: but this PR LGTM

PR #1474, opened 2023-03-09
Fix fl context prop

YuanTingHsieh:
### Description

Some props stickiness are inconsistent.
We change them to be the same.

### Types of changes

- [ ] Documentation updated.


holgerroth (approved):
Looks good.

PR #1476, opened 2023-03-09
Fixed the simulator threads option for multi_gpu case.

yhwen:
Fixes # .

Fixed the simulator threads option issue when using multi-gpu

### Description

- When using multi_gpu simulator, set the threads to 1,
- when using single gpu in the --gpu option, use the default value 1, or use the user choose threads number. 

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh (approved):
Thanks, LGTM.

nvidianz (approved):

PR #1480, opened 2023-03-09
Configuration exception handling

chesterxgchen:
Fixes # .  Both customer reported  issue and FLARE-516 
### Description
**Problem**
When user use embedded component in configuration, i.e. component's argument is another component or class. such as
```
   "comp1": { 
      "path": "module1.for.comp1.Compt1", 
       "args" : {
            "path": "module2.for.comp2.Comp2", 
             args: {
                "a": "b"
             }
        }
   }

```
Where comp2 is the embedded component. The original component_builder looks the dictionary argument 
```
"args" : {
            "path": "module2.for.comp2.Comp2", 
             args: {
                "a": "b"
             }
        }
```
Couldn't sure this is normal dictionary arg or its embedded component configuration. So the Component_builder just try to build the component ( instantiate the class for Comp2). If the class Comp2 is instantiated, then ok. if failed, the component_builder is considered it is normal dictionary arg, so swallow the exception. 

This logic is flawed as it did not consider the cases where 
   * The configuration error for Comp2 arguments causing Comp2 class instantiation failure
   * The initialization error for Comp2 causing Comp2 class instantiation failure

as result, the user get very strange configuration error message or random exception which is not related the root cause of the exception. 

**Approaches**
This PR attempt to fix this with following approaches: 
* Try to identity the embedded configuration is a component/class before try to instantiate 
    * if the instantiation failed, we raise the exception for user ( this will cover both args mis-config and initiation error) 
    * The class/component identification can be done by checking 
               * must has a "path" or "name"
                * must have a valid class path
  * for non-class config, they must be normal dict

**Changes**
1) change the component_builder with above logics
2) add unit tests to cover both positive and negative of component (with/without embedded component) creations 
    Unit tests handle both < 3.9.x and 3.10.x python message differences. 

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/examples/advanced/cifar10/cifar10-sim/jobs/cifar10_central/cifar10_central/config/config_fed_server.json
+++ b/examples/advanced/cifar10/cifar10-sim/jobs/cifar10_central/cifar10_central/config/config_fed_server.json
@@ -16,8 +16,7 @@
       "args": {
         "model": {
           "path": "pt.networks.cifar10_nets.ModerateCNN",
-	      "args": {
-          }
+	      "args": {}
         }
       }
     },
--- a/nvflare/fuel/utils/component_builder.py
+++ b/nvflare/fuel/utils/component_builder.py
@@ -28,9 +28,7 @@ class ComponentBuilder:
         pass
 
     def is_class_config(self, config_dict: dict) -> bool:
-        # the dict has both path/name or args, it's likely a component,
-        # test if it has a valid class path
-        if "args" in config_dict and ("path" in config_dict or "name" in config_dict):
+        def has_valid_class_path():
             try:
                 _ = self.get_class_path(config_dict)
                 # we have valid class path
@@ -38,6 +36,11 @@ class ComponentBuilder:
             except ConfigError:
                 # this is not a valid class path
                 return False
+
+        # regardless it has args or not. if path/name and valid class path, very likely we have
+        # class config.
+        if ("path" in config_dict or "name" in config_dict) and has_valid_class_path():
+            return True
         else:
             return False
 
--- a/tests/unit_test/fuel/utils/component_builder_test.py
+++ b/tests/unit_test/fuel/utils/component_builder_test.py
@@ -14,6 +14,7 @@
 import re
 
 import pytest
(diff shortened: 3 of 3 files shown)


yanchengnv (approved):
Yeah, I have been thinking about similar approach.

yanchengnv:
There is a unit test error. Please fix.

holgerroth (approved):
Nice. It works as expected on my side.

IsaacYangSLA:
One question.  In the resources.json.default file, one component does not have args.  Because args is required for component, will this PR think it's not a component?

IsaacYangSLA:
Inside resources.json.default of server startup kit local folder
`        {
            "id": "job_store",
            "path": "nvflare.app_common.storages.filesystem_storage.FilesystemStorage"
        }
`

examples/advanced/cifar10/cifar10-sim/jobs/cifar10_central/cifar10_central/config/config_fed_server.json
```diff
@@ -16,7 +16,8 @@
       "args": {
         "model": {
           "path": "pt.networks.cifar10_nets.ModerateCNN",
-	      "args": {}
+	      "args": {
+          }
```
YuanTingHsieh, line 20: can we undo changes to this file? thanks

nvflare/fuel/utils/component_builder.py
```diff
@@ -28,6 +27,20 @@ def get_module_scanner(self):
         """
         pass
 
+    def is_class_config(self, config_dict: dict) -> bool:
+        # the dict has both path/name or args, it's likely a component,
+        # test if it has a valid class path
+        if "args" in config_dict and ("path" in config_dict or "name" in config_dict):
```
YuanTingHsieh, line 33: the comment string should be "the dict has both args and path/name"

What if the SUB component does not specify args?

So maybe somewhere in documentation we need to say you need to specify "args" for it to be treat as a SUB component?

YuanTingHsieh (approved):

PR #1481, opened 2023-03-09
add wait_for_system_shutdown [skip ci]

chesterxgchen:
Fixes # . FLARE-1097

### Description
**Problem**
POC shutdown show a lot errors from communication. 
* This is partly due to FCI layer trying to re-connect and failed error during shutdown
* Another part is that the process got killed without completely shutdown. 

 
**Solution Approach**
  The 1st part of problem is handled by different PR
  The 2nd part of the problem is addressed in this PR. 

  The shutdown.fl trigger the "kill" process, we would like to delay the "kill" by wait for system shutdown completely. 
  We add wait_for_system_shutdown() by checking for system status changes every 0.1 second, but with timeout = 30 seconds. 
  This should reduce some number of errors. 

**Changes**
   1. Some README update
   2. handle Answer "N" prompt for nvflare poc --prepare  ( it was not properly handled) 
   3. implement wait_for_system_shutdown() 

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/fuel/flare_api/flare_api.py
+++ b/nvflare/fuel/flare_api/flare_api.py
@@ -265,7 +265,6 @@ class Session(SessionSpec):
     def list_jobs(
         self,
         detailed: bool = False,
-        all: bool = False,
         limit: Optional[int] = None,
         id_prefix: str = None,
         name_prefix: str = None,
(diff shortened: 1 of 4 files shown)


nvflare/lighter/poc_commands.py
```diff
@@ -209,6 +209,9 @@ def stop_poc(poc_workspace: str, excluded=None, white_list=None):
 
         print("shutdown NVFLARE")
         sess.api.do_command("shutdown all")
+
+        wait_for_system_shutdown(sess)
+
```
YuanTingHsieh, line 214: Why if this session wait for system to shutdown will have less error messages?
This session should have nothing to do with the messages on either server or client side?

nvflare/lighter/poc_commands.py
```diff
@@ -221,6 +225,24 @@ def stop_poc(poc_workspace: str, excluded=None, white_list=None):
...
+    timeout = 30
+    start = time.time()
+    duration = 0
+    cnt = 0
+    while status == "started" and duration < timeout:
+        sys_info = sess.get_system_info()
+        status = sys_info.server_info.status
+        curr = time.time()
+        duration = curr - start
+        if cnt % 25 == 0:
```
YuanTingHsieh, line 240: maybe make these numbers 25, 30 to be passed in as argument.

YuanTingHsieh (approved):

PR #1487, opened 2023-03-10
fix list_job in flare api

yanchengnv:
### Description

Fixed flare api's list_job to be consistent with the command behavior. Specifically, removed the "all" arg. To get all jobs, just do not specify the "limit" arg (or set to None).

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh (approved):
Thanks, LGTM

PR #1489, opened 2023-03-10
protect server state against multiple state changes

yanchengnv:
### Description

Overseer CB could happen multiple times in quick succession, whereas server state change could take time to complete. To protect server state's integrity, we do not want to handle these state change requests concurrently since they may mess up each other. We also don't need to pile up the requests to process them sequentially. This is because the latest state change request counts. So if there are multiple such requests during the state change processing, we only need to deal with the last one after the current state change processing is done and ignore all others.

This PR uses a simple flag to indicate whether we are busy with state change. If so simply ignore any other state change requests. Since state change requests repeat themselves, we will be able to catch the wanted change in the next request.

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh (approved):
Thanks, changes LGTM

PR #1491, opened 2023-03-11
simulator create the clients in parallel.

yhwen:
Fixes # .

### Description

It takes time to create the communication Cell for the client. Small number of client, it's around 0.5 ~ 0.9 seconds. With large number of clients, the creating time could be up to 2 ~ 3 seconds. So instead of creating the simulator clients sequentially, change to create the clients parallel. 

### Types of changes

- [ ] Documentation updated.


nvflare/private/fed/app/simulator/simulator_runner.py
```diff
@@ -286,18 +288,30 @@ def split_clients(self, clients: [], gpus: []):
...
 
+        clients_created_waiter.wait()
         self.logger.info("Set the client status ready.")
         self._set_client_status()
 
+    def create_client(self, client_name, client_count_lock, clients_created_waiter):
+        client, self.client_config, self.deploy_args, self.build_ctx = self.deployer.create_fl_client(
+            client_name, self.args
+        )
+        self.federated_clients.append(client)
```
YuanTingHsieh, line 306: there might be chances multiple threads arrive here at the same time?
we should use a lock for "append"?

nvflare/private/fed/app/simulator/simulator_runner.py
```diff
@@ -286,18 +288,30 @@ def split_clients(self, clients: [], gpus: []):
...
-            )
-            self.federated_clients.append(client)
-            app_root = os.path.join(self.simulator_root, "app_" + client_name)
-            app_custom_folder = os.path.join(app_root, "custom")
-            sys.path.append(app_custom_folder)
+            threading.Thread(
+                target=self.create_client, args=[client_name, client_count_lock, clients_created_waiter]
+            ).start()
 
+        clients_created_waiter.wait()
```
YuanTingHsieh, line 298: we could either use this event wait mechanism, or we can have a for loop to join all these threads.

YuanTingHsieh:
Note that this PR only minimize the "SETUP" time of simulator, the rest of the time in dev is still much slower than version 2.2

nvflare/private/fed/app/simulator/simulator_runner.py
yhwen: The List append() is thread safe. No need to use additional lock here. 

https://stackoverflow.com/questions/6319207/are-lists-thread-safe

nvflare/private/fed/app/simulator/simulator_runner.py
yhwen: The threading.Event wait() is efficient to use here. Also the waiting logic is fully self contained within the class.

yhwen:
> I am not so sure about number of threads is linearly proportional to the number of clients. For small number of clients, this should be fine, for large number clients. this means a lot of threads were created about the same time. Maybe we should have up-to N threads to be proportional to N clients. After N threads, we create a thread pool.
> 
> But on the other hands, each thread only run 2-3 seconds, do we need such trouble to do the thread pool.
> 
> Like to hear what other think.

The problem we are trying to address is the speed up here. I think we do not need to introduce another artificial

yhwen:
> I am not so sure about number of threads is linearly proportional to the number of clients. For small number of clients, this should be fine, for large number clients. this means a lot of threads were created about the same time. Maybe we should have up-to N threads to be proportional to N clients. After N threads, we create a thread pool.
> 
> But on the other hands, each thread only run 2-3 seconds, do we need such trouble to do the thread pool.
> 
> Like to hear what other think.

Changed to use a threadpool to control how many clients creating at the same time.

yhwen:
> Note that this PR only minimize the "SETUP" time of simulator, the rest of the time in dev is still much slower than version 2.2

My test profiling show only the delay are caused by the client cell creating. After changing to parallel, the running are pretty much compatible with 2.2 now.

YuanTingHsieh (approved):
Thanks, this part LGTM

PR #1495, opened 2023-03-13
Random forest minor fixes

ZiyueXu77:
Fixes # .

### Description

Fixed two suggestions from QA

### Types of changes

- [ ] Documentation updated.


ZiyueXu77:
@YuanTingHsieh , I think the suggestion from QA is good in handling large dataset path like HIGGS - putting the root as a variable for the bash script

YuanTingHsieh (approved):

YuanTingHsieh:
@ZiyueXu77 I agree if we are asking user to modify it we better just pass it.

However, for some examples that need these path in config fed client.json / config_fed_server.json I think we should just have a default there, that is also better for automation.

PR #1503, opened 2023-03-14
fix list_jobs detail handling

yanchengnv:
### Description

The old FLAdminAPI didn't support the "-d" option of the "list_jobs" command. This PR fixes it by using the returned "meta" info.

This PR fixes the bug reported in NVBug 4005332.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/fuel/hci/client/fl_admin_api.py
+++ b/nvflare/fuel/hci/client/fl_admin_api.py
@@ -396,11 +396,10 @@ class FLAdminAPI(AdminAPI, FLAdminAPISpec):
             command = command + " " + options
         success, _, reply = self._get_processed_cmd_reply_data(command)
         if success:
-            if reply.get("data"):
-                for data in reply["data"]:
-                    if data["type"] == "table":
-                        details = data["rows"]
-                        return FLAdminAPIResponse(APIStatus.SUCCESS, details, reply)
+            meta = reply.get("meta")
+            if meta:
+                jobs_list = meta.get("jobs", [])
+                return FLAdminAPIResponse(APIStatus.SUCCESS, jobs_list, reply)
         return FLAdminAPIResponse(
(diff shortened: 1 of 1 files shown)


nvflare/fuel/hci/client/fl_admin_api.py
```diff
@@ -396,11 +396,10 @@ def list_jobs(self, options: str = None) -> FLAdminAPIResponse:
...
         if success:
-            if reply.get("data"):
-                for data in reply["data"]:
-                    if data["type"] == "table":
-                        details = data["rows"]
-                        return FLAdminAPIResponse(APIStatus.SUCCESS, details, reply)
+            meta = reply.get("meta")
+            if meta:
+                jobs_list = meta.get("jobs", [])
+                return FLAdminAPIResponse(APIStatus.SUCCESS, jobs_list, reply)
```
YuanTingHsieh, line 402: This LGTM, I just went through the code and think we could have some clean-ups regarding fl_admin_api/admin_api/flare_api, but that can be done later.

YuanTingHsieh (approved):

nvflare/fuel/hci/client/fl_admin_api.py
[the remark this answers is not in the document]
yanchengnv: The old code simply did not handle the "-d" case - it was only looking for a table in the result. But the "-d" result was not a table, instead it was a list of JSON strings, and the code treated as RUN TIME ERROR.

This simply changes to use the "meta" structure that handles both cases.

BTW, this old fl_admin_api will be sunset soon.

yanchengnv: We'll sunset fl_admin_api in future. flare_api is the way to go.

PR #1505, opened 2023-03-14
silent abort message logging

yanchengnv:
### Description

When aborting a job, the server sends abort commands to all clients and wait for their responses. However since some of the job cells may already be gone, the server will timeout and creates an error log message.

This PR makes the abort request "optional", and create the timeout message in DEBUG level.

This PR fixes NVBug 4027037.

This PR also added some minor improvements:
- Updated the help info of the "list_jobs" command with the "-r" option
- Added/fixed some docstrings.
- Shorten the wait time of the abort command to 2.0 secs from 10 secs

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/fuel/f3/cellnet/cell.py
+++ b/nvflare/fuel/f3/cellnet/cell.py
@@ -1119,6 +1119,7 @@ class Cell(MessageReceiver, EndpointMonitor):
         Args:
             target_msgs: messages to be sent
             timeout: timeout value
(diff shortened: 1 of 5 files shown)


nvidianz (approved):

yanchengnv:
> probably rename optional to something else

The "optional" already has a defined meaning in cellnet. This PR simply makes other places to use this feature.

PR #1507, opened 2023-03-14
not creating internal listener for the job cell.

yhwen:
Fixes # .

### Description

Change to not create the internal listener for the job cells. No need to create them.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/fuel/f3/cellnet/cell.py
+++ b/nvflare/fuel/f3/cellnet/cell.py
@@ -603,7 +603,7 @@ class Cell(MessageReceiver, EndpointMonitor):
                 self.logger.debug(f"{self.my_info.fqcn}: removing agent {a}")
                 self.agents.pop(a, None)
 
-    def create_internal_listener(self):
+    def make_internal_listener(self):
         """
         Create the internal listener for child cells of this cell to connect to.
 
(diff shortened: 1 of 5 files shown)


nvflare/fuel/f3/cellnet/cell.py
```diff
@@ -603,7 +603,7 @@ def drop_agents(self):
                 self.logger.debug(f"{self.my_info.fqcn}: removing agent {a}")
                 self.agents.pop(a, None)
 
-    def create_internal_listener(self):
+    def make_internal_listener(self):
```
yanchengnv, line 606: Thanks for catching the name conflict between the method name and instance variable name!

yanchengnv (approved):
Thanks for catching the name conflict between the method name and instance variable name!
Looks good to me.
Removing the creation of unnecessary internal listeners not only saves time, but also saves port numbers!

nvflare/private/fed/client/fed_client_base.py
```diff
@@ -181,9 +181,11 @@ def _create_cell(self, location, scheme):
         if self.args.job_id:
             fqcn = FQCN.join([self.client_name, self.args.job_id])
             parent_url = self.args.parent_url
+            create_internal_listener = False
         else:
             fqcn = self.client_name
             parent_url = None
+            create_internal_listener = True
```
YuanTingHsieh, line 188: When should we create internal listener?
When that cell know it has a child?

PR #1508, opened 2023-03-14
Remove things in __init__.py in app_opt

YuanTingHsieh:
### Description

- Remove things in `__init__.py` in app_opt/he and app_opt/pt
- Fix imports for other files

### Types of changes

- [ ] Documentation updated.


nvflare/app_common/pt/pt_fed_utils.py
```diff
@@ -18,5 +18,7 @@
     f"This module: {__file__} is deprecated. Please use nvflare.app_opt.pt.", category=FutureWarning, stacklevel=2
```
holgerroth, line 18: Should the warning say `nvflare.app_opt.pt.model_persistence_format_manager`?

YuanTingHsieh: previously this `model_persistence_format_manager` is put inside `pt_fed_utils.py`
but right now in app_opt I separate them, what do you think should be the correct warning?

holgerroth: I see. Please ignore my comment. The current warning should be fine then.

PR #1520, opened 2023-03-15
Throw exception when connection monitor is not registered

nvidianz:
### Description

Changed error handling when connection monitor is not registered in FCI to throw an exception. In the current version, it only prints an error.

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh (approved):
thanks, LGTM

PR #1523, opened 2023-03-15
fix job status and speed up fed event end_run

yanchengnv:
### Description

- NVBug 4029429. Consolidated API status for authorization error.
- NVBug 4007603. Modified the server's job status update/waiting logic
- NVBug 4029376. The error was caused by the race condition that the job cell tries to retrieve client list from the server before its connection to the server is fully established. Changed to keep trying until result is received or timed out. Note: there is a similar case on the client side that Yuhong will address.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/private/fed/server/server_engine.py
+++ b/nvflare/private/fed/server/server_engine.py
@@ -59,6 +59,7 @@ from nvflare.private.scheduler_constants import ShareableHeader
 from nvflare.security.logging import secure_format_exception
 from nvflare.widgets.info_collector import InfoCollector
 from nvflare.widgets.widget import Widget, WidgetID
+from nvflare.fuel.f3.cellnet.defs import MessageHeaderKey, ReturnCode as CellMsgReturnCode
 
 from .client_manager import ClientManager
 from .job_runner import JobRunner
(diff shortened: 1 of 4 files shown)


nvflare/private/fed/server/server_engine.py
```diff
@@ -488,28 +493,49 @@ def send_aux_request(
...
     def sync_clients_from_main_process(self):
-        return_data = self._get_clients_data(self.args.job_id)
-        clients = return_data.get(ServerCommandKey.CLIENTS)
-        self.client_manager.clients = clients
-
-    def _get_clients_data(self, job_id):
+        # repeatedly ask the parent process to get participating clients until we receive the result
+        # or timed out after 30 secs (already tried 30 times).
         start = time.time()
+        max_wait = 30.0
```
yhwen, line 499: probably no need to use max_wait for 30 seconds. This should be relative fast.

yanchengnv: fast is fine since it won't wait that long - as soon as the result is retrieved, it will break out.

yhwen: that's fine.

yhwen (dismissed):

nvflare/widgets/fed_event.py
```diff
@@ -138,32 +130,26 @@ def _post(self):
...
+                if len(self.in_events) > 0:
                     continue
+                else:
+                    break
             else:
-                sleep_time = min(sleep_time * 2, 1)
+                sleep_time = self.regular_interval
                 continue
 
+            self.logger.info(f"========== got fed event: {event_type}")
```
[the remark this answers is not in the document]
yanchengnv: Oh no, need to remove this.

yhwen (approved):

PR #1524, opened 2023-03-15
Optimize the get_all_clients, move to the training process beginning.

yhwen:
Fixes # .

### Description

Move the get all_clients to the beginning of client job start. wrap in the sync_up_parents_process().

### Types of changes

- [ ] Documentation updated.


yhwen:
> Although logics LGTM, but not sure why you need to do this ?

It's for optimization purpose. After this, no need to check if get the clients already. It's cleaner.

PR #1526, opened 2023-03-16
Add config_type to distinguish

YuanTingHsieh:
### Description

For `PTFedOptModelShareableGenerator` we do want to treat those args as dictionary but NOT nested components.

This PR adds a "config_type" to distinguish between a dictionary should be built as component or just pass in as a regular dictionary. 

### Types of changes

- [] Non-breaking change (fix or new feature that would not break existing functionality).
- [ ] Documentation updated.


IsaacYangSLA (approved):
Thanks.  This is a good PR.

PR #1527, opened 2023-03-16
Added Decomposers for HE Classes [skip ci]

nvidianz:
### Description

Added decomposer for CKKSVector class for HE examples.

### Types of changes

- [ ] Documentation updated.


nvflare/app_opt/he/he_decomposers.py
```diff
@@ -0,0 +1,44 @@
+# Copyright (c) 2021-2023, NVIDIA CORPORATION.  All rights reserved.
```
holgerroth, line 1: Should remove the he_ prefix in the filename to follow the scheme of the other files in this folder.

nvidianz: Removed the prefix.

nvidianz:
> 

We should. That's what FOBS is designed to do.

holgerroth:
> 

Actually, we need to manually call serialize() in order to support saving the global model with `torch.save()`.

holgerroth (approved):
Looks good!

PR #1529, opened 2023-03-16
Raise Exception when error code = EXECUTION_RESULT_ERROR in Controller error handling

chesterxgchen:
…ing.

Fixes #  FLARE-1068 ( 4007688) 

### Description
   **Problems**
   CommonController.handle_client_errors() 
   when client return errors, either we need to abort the job, or we need to report error. 
   Current implementation only logging error when not abort the job. 
   This result the controller logs something like CompletionStatus.OK even the executor completely failed. 

  **approach to resolve**
   Here we change it to raise the ValueError() so system indicate something is wrong. 

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/.github/workflows/blossom-ci.yml
+++ b/.github/workflows/blossom-ci.yml
@@ -56,6 +56,7 @@ jobs:
       yanchengnv,\
       ZiyueXu77,\
       Can-Zhao,\
+      guopengf,\
       ', format('{0},', github.actor)) && github.event.comment.body == '/build'
     steps:
       - name: Check if comment is issued by authorized person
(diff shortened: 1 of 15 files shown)


nvflare/app_common/workflows/common_controller.py
```diff
@@ -43,4 +43,4 @@ def handle_client_errors(self, rc: str, client_task: ClientTask, fl_ctx: FLConte
             )
             self.log_error(fl_ctx, f"Execution failed for {client_name}")
         else:
-            self.log_error(fl_ctx, f"Execution result is not received for {client_name}")
+            raise ValueError(f"Execution result is not received for {client_name}")
```
yanchengnv, line 46: What is the CommonController for?

YuanTingHsieh (approved):

PR #1534, opened 2023-03-16
Support direct cell message

yanchengnv:
### Description

This PR is intended to boost overall performance of the simulator. It does so in two areas:
- Use in-memory calls for comms between two cells within the same process. This is 2X faster than even local tcp.
- Reduce number of unnecessary threads. Instead of creating them at component init time, create them only on as-needed basis. This strategy is applied to fed_event and cellnet. In case of fed event, the "poster" thread is created only when fed events are received. Most jobs don't use fed events.  In case of cellnet, the threads for handling queued messages are also made on as-needed basis. Since we do not use the queue message mechanism in FLARE, this will save two extra threads per cell.  All these will have a huge saving on number of threads when simulating large number of clients.

The removing of fed event's "poster" thread will also reduce the overall shutdown time.

### Types of changes

- [ ] Documentation updated.


nvflare/fuel/f3/cellnet/cell.py
```diff
@@ -505,10 +527,13 @@ def is_backbone_ready(self):
 
     def _set_bb_for_client_root(self):
```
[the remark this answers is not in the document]
yanchengnv: backbone

nvflare/fuel/f3/cellnet/cell.py
```diff
@@ -1238,18 +1293,26 @@ def fire_and_forget(
...
         if isinstance(targets, str):
             targets = [targets]
 
         message.set_header(MessageHeaderKey.OPTIONAL, optional)
         with self.bulk_lock:
+            if self.bulk_checker is None:
+                self.logger.info(f"{self.my_info.fqcn}: starting bulk_checker")
+                self.bulk_checker = threading.Thread(target=self._check_bulk)
+                self.bulk_checker.start()
+                self.logger.info(f"{self.my_info.fqcn}: started bulk_checker")
```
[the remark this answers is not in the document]
yanchengnv: check bulk messages. Bulk message is a message that contains many other messages that are queued. Currently FLARE is not using this mechanism yet.

nvflare/fuel/f3/cellnet/cell.py
```diff
@@ -175,6 +176,10 @@ def log_messaging_error(logger, log_text: str, cell, msg: Union[Message, None],
         logger.debug(log_text)
         if log_except:
             logger.debug(secure_format_traceback())
+    elif log_warn:
```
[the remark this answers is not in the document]
yanchengnv: Sounds like a good idea. Will make the change.

nvflare/widgets/fed_event.py
```diff
@@ -104,6 +102,11 @@ def _receive(self, topic: str, request: Shareable, fl_ctx: FLContext) -> Shareab
             return make_reply(ReturnCode.BAD_REQUEST_DATA)
 
         with self.in_lock:
+            if self.poster is None:
+                # create the poster thread now
+                self.poster = threading.Thread(target=self._post, args=())
+                self.poster.start()
```
YuanTingHsieh, line 108: should we change the line in `__init__.py` to `self.poster = None`

nvflare/widgets/fed_event.py
```diff
@@ -104,6 +102,11 @@ def _receive(self, topic: str, request: Shareable, fl_ctx: FLContext) -> Shareab
             return make_reply(ReturnCode.BAD_REQUEST_DATA)
 
         with self.in_lock:
+            if self.poster is None:
+                # create the poster thread now
+                self.poster = threading.Thread(target=self._post, args=())
```
nvidianz, line 107:
```suggestion
                self.poster = threading.Thread(target=self._post, name="poster", args=())
```

nvflare/fuel/f3/cellnet/cell.py
```diff
@@ -1238,18 +1293,26 @@ def fire_and_forget(
...
+            raise RuntimeError(f"{self.get_fqcn()}: bulk message is not enabled!")
+
         if isinstance(targets, str):
             targets = [targets]
 
         message.set_header(MessageHeaderKey.OPTIONAL, optional)
         with self.bulk_lock:
+            if self.bulk_checker is None:
+                self.logger.info(f"{self.my_info.fqcn}: starting bulk_checker")
+                self.bulk_checker = threading.Thread(target=self._check_bulk)
```
nvidianz, line 1306:
```suggestion
                self.bulk_checker = threading.Thread(target=self._check_bulk, name="bulk_checker“)
```

nvflare/fuel/f3/cellnet/cell.py
```diff
@@ -1273,6 +1336,11 @@ def _receive_bulk_message(self, request: Message):
         target_msgs = request.payload
         assert isinstance(target_msgs, list)
         with self.bulk_msg_lock:
+            if self.bulk_processor is None:
+                self.logger.debug(f"{self.my_info.fqcn}: starting bulk message processor")
+                self.bulk_processor = threading.Thread(target=self._process_bulk_messages)
```
nvidianz, line 1341:
```suggestion
                self.bulk_processor = threading.Thread(target=self._process_bulk_messages， name="bulk_proc")
```

nvidianz (approved):

nvflare/fuel/f3/cellnet/cell.py
```diff
@@ -160,7 +161,7 @@ def __init__(self, targets: List[str]):
         self.received_replies = {}
 
 
-def log_messaging_error(logger, log_text: str, cell, msg: Union[Message, None], log_except=False):
+def log_messaging_error(logger, log_text: str, cell, msg: Union[Message, None], log_except=False, log_warn=False):
```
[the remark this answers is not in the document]
yanchengnv: good idead

nvflare/fuel/f3/cellnet/cell.py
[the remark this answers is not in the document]
yanchengnv: Changed to use a simpler solution: logger.log(log_level...).

nvflare/widgets/fed_event.py
yanchengnv: Changed as suggested.

nvflare/fuel/f3/cellnet/cell.py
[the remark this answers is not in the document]
yanchengnv: Changed as suggested.

nvflare/widgets/fed_event.py
yanchengnv: Changed as suggested.

nvflare/fuel/f3/cellnet/cell.py
yanchengnv: Changed as suggested

nvflare/fuel/f3/cellnet/cell.py
yanchengnv: Changed as suggested

nvflare/fuel/f3/cellnet/cell.py
```diff
@@ -1837,14 +1904,30 @@ def get_sub_cell_names(self) -> Tuple[List[str], List[str]]:
 
         Returns: fqcns of child cells, fqcns of top-level client cells
         """
-        children = []
-        clients = []
+        children_dict = {}
+        clients_dict = {}
```
YuanTingHsieh, line 1908: just a note, we could also use set here since we only care about keys.

YuanTingHsieh (approved):
LGTM

PR #1537, opened 2023-03-17
Add a controller lock to prevent racing condition

YuanTingHsieh:
### Description

- Add a controller_lock to make sure _check_tasks, process_task_request, process_submission does not interfere with each other
- Fix unit test
- Add new unit test

### Types of changes

- [ ] Documentation updated.


nvflare/apis/impl/controller.py
```diff
@@ -169,125 +172,126 @@ def process_task_request(self, client: Client, fl_ctx: FLContext) -> Tuple[str,
...
 
         client_task_to_send = None
-        with self._task_lock:
-            self.logger.debug("self._tasks: {}".format(self._tasks))
-            for task in self._tasks:
-                if task.completion_status is not None:
-                    # this task is finished (and waiting for the monitor to exit it)
-                    continue
+        with self._controller_lock:
+            with self._task_lock:
```
nvidianz, line 176: Double locking should be avoided if all possible. It can easily cause dead-lock. In this case, I don't see the need for the controller_lock.

YuanTingHsieh: Thanks for the review, I simplified the logic

nvflare/apis/impl/controller.py
```diff
@@ -822,6 +834,10 @@ def _monitor_tasks(self):
             time.sleep(self._task_check_period)
 
     def _check_tasks(self):
+        with self._controller_lock:
+            self._do_check_tests()
```
yanchengnv, line 838: Should be _do_check_tasks, "tests".

yanchengnv (changes requested):
Please rename the method do_check_tests to do_check_tasks.

nvflare/apis/impl/controller.py
YuanTingHsieh: Thanks for finding this, fixed

yanchengnv (approved):
LGTM

PR #1542, opened 2023-03-17
Allow execute individual workflow

chesterxgchen:
Fixes # .FLARE-87, NVBugs 115662

Description

Problem or Feature ask
1. This is a long standing ask from customer for ability to be able to run cross-site validation on demand after train 
Use case is upload cross site validation config to run it after the fact or run for all Runs (including previous runs) 
2. run evaluation/cross-validation without training 
3. provide the global model and then only run-cross validation

Analysis
There are two parts to this requests: 
1) Related to workflow: The task is more general than simple cross-site validation re-run. Can we execute individual workflow ? As training and evaluation are two workflows. The task is to selectively run selected workflows as needed.

2) related to the model, we need to be able to use the model provided or saved from previous training run during validation workflow. 

This PR only addresses this 1st one. 

Changes
Allow one to execute individual workflow.
*add optional config element "target_workflow_ids", such as
```
{
  "format_version": 2,
  "task_data_filters": [],
  "task_result_filters": [],
  "components": [   <skip rest > ],
  "workflows": [
    {
      "id": "scatter_and_gather",
      "path": "nvflare.app_common.workflows.scatter_and_gather.ScatterAndGather",
        <skip rest>
    },
    {
      "id": "cross_site_model_eval",
      "path": "nvflare.app_common.workflows.cross_site_model_eval.CrossSiteModelEval",
      <skip rest>
    }
  ],
"target_workflow_ids": ["cross_site_model_eval"]
}
```
target_workflows_ids specify the selected workflows and order the executions based on the array order. In case re-run cross site validation, only one workflow  is needed for the target_worklow_ids. 

### Types of changes

- [ ] Documentation updated.


nvflare/private/fed/server/server_runner.py
```diff
@@ -92,57 +96,69 @@ def __init__(self, config: ServerRunnerConfig, job_id: str, engine: ServerEngine
...
+        else:
+            target_workflows = []
+            for wf_id in wf_ids:
+                target_wfs = [wf for wf in self.config.workflows if wf.id == wf_id]
+                target_workflows.extend(target_wfs)
+        return target_workflows
+
+    def _execute_run(self):
+        target_wf = self._get_target_workflow()
+        for wf in target_wf:
```
YuanTingHsieh, line 157: the previous code using "while self.current_wf_index < len(self.config.workflows):" for a reason.

If you check code: https://github.com/NVIDIA/NVFlare/blob/dev/nvflare/private/fed/server/server_runner.py#L482-L487

You see we will persist the `current_wf_index`.
And when HA resumes, we will only execute the workflows that has not been finished.

YuanTingHsieh: Please make sure to take HA into consideration, thanks

YuanTingHsieh (changes requested):
Please make sure the original resume logic work.

holgerroth:
Do we need "target_workflow_ids"? One could just delete the workflow that shouldn't be executed.

yanchengnv:
We need to be clear on how this is intended to be used. If the user wants to be able to independently evaluate a model trained previously, then a few things have to happen:
- The model has to be trained in a job. Once this job is done, the model is saved in the workspace of this job;
- If the user wants to be able to validate that model in separate jobs, the user has to arrange to make that model available in these new jobs.

I guess you want to be able to share the same job config, and only change the target_workflow_ids when submitting as a new job?

Perhaps we should discuss how this thing works from user's perspective first?

yanchengnv:
Do we still want this? Unclear to me.

nvflare/private/fed/server/server_runner.py
```diff
@@ -38,8 +39,9 @@ def __init__(
         workflows: [],
         task_data_filters: dict,
         task_result_filters: dict,
-        handlers=None,
-        components=None,
+        handlers: Optional[List] = None,
+        components: Optional[Dict] = None,
+        target_wf_ids: Optional[List[str]] = None,
```
YuanTingHsieh, line 44: code-wise LGTM, prefer "selected_wf_ids" over "target_wf_ids"

The reason is that we have "target" in Task and Controller class and that means a different thing.

PR #1549, opened 2023-03-20
Fixed the shared object issue in the controller task return.

yhwen:
Fixes # .

Fixed the simulator delay with large number of clients run.

### Description

A shared memory object issue in the controller task return caused the concurrent clients get_task request got the same task to run. This caused the simulator clients ran for multiple times.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/apis/impl/controller.py
+++ b/nvflare/apis/impl/controller.py
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+import copy
 import threading
 import time
 from abc import ABC
@@ -287,7 +287,7 @@ class Controller(Responder, ControllerSpec, ABC):
                 task.last_client_task_map[client.name] = client_task_to_send
                 task.client_tasks.append(client_task_to_send)
                 self._client_task_map[client_task_to_send.id] = client_task_to_send
-            return task_name, client_task_to_send.id, task_data
+            return task_name, client_task_to_send.id, copy.deepcopy(task_data)
 
     def handle_exception(self, task_id: str, fl_ctx: FLContext) -> None:
         """Called to cancel one task as its client_task is causing exception at upper level.
--- a/nvflare/private/fed/server/server_commands.py
+++ b/nvflare/private/fed/server/server_commands.py
@@ -15,6 +15,7 @@
 """FL Admin commands."""
 
 import copy
+import logging
 import time
 from abc import ABC, abstractmethod
 from typing import List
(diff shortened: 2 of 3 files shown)


nvflare/apis/impl/controller.py
```diff
@@ -287,7 +287,7 @@ def process_task_request(self, client: Client, fl_ctx: FLContext) -> Tuple[str,
                 task.last_client_task_map[client.name] = client_task_to_send
                 task.client_tasks.append(client_task_to_send)
                 self._client_task_map[client_task_to_send.id] = client_task_to_send
-            return task_name, client_task_to_send.id, task_data
+            return task_name, client_task_to_send.id, copy.deepcopy(task_data)
```
IsaacYangSLA, line 290: This may be using lots of memory as task_data is a shareable.

yhwen: I am aware of this. Ideally, we need only the copy the headers part of the shareable. But we don't that enabled in the Sharable. We can enhance that in the future.

YuanTingHsieh: The upstream (who called this method) is trying to set header to this task_data's shareable.
Resulting in multiple threads trying to write/set same key to different values at the same time.

By returning a full copy, we don't write/set the same object.

nvflare/private/fed/server/server_commands.py
```diff
@@ -166,7 +164,10 @@ def process(self, data: Shareable, fl_ctx: FLContext):
         shared_fl_ctx.set_public_props(copy.deepcopy(get_serializable_data(fl_ctx).get_all_public_props()))
         shareable.set_header(key=FLContextKey.PEER_CONTEXT, value=shared_fl_ctx)
 
-        # return fobs.dumps(data)
+        self.logger.info(
+            f"return task to client.  client_name:{client.name}  task_id:{task_id}  "
+            f"sharable_header_task_id: {shareable.get_header(key=FLContextKey.TASK_ID)}"
+        )
```
YuanTingHsieh, line 170: do we want this log at info level anymore?

nvflare/private/fed/server/server_commands.py
```diff
@@ -209,6 +203,7 @@ def process(self, data: Shareable, fl_ctx: FLContext):
         task_id = data.get_cookie(FLContextKey.TASK_ID)
         server_runner = fl_ctx.get_prop(FLContextKey.RUNNER)
         server_runner.process_submission(client, contribution_task_name, task_id, data, fl_ctx)
+        self.logger.info(f"submit_update process. client_name:{client.name}   task_id:{task_id}")
```
YuanTingHsieh, line 206: do we want this log at info level anymore?

yhwen: same as above.

nvflare/private/fed/server/server_commands.py
YuanTingHsieh: sounds good, I am OK to both situations.

PR #1556, opened 2023-03-20
Fixed -m option in list_jobs [skip ci]

nvidianz:
### Description

Changed -m in list_jobs to show most recent jobs, regardless of the -r option.

### Types of changes

- [ ] Documentation updated.


nvflare/private/fed/server/job_cmds.py
```diff
@@ -327,7 +327,10 @@ def list_jobs(self, conn: Connection, args: List[str]):
                 filtered_jobs.sort(key=lambda job: job.meta.get(JobMetaKey.SUBMIT_TIME.value, 0.0), reverse=reverse)
```
YuanTingHsieh, line 327: Maybe default we should have "reverse" equal to True and keep the following line unchanged?

YuanTingHsieh: let's say the order from old to latest: job1 job2 job3 job4,

if specified nothing we want:
job1
job2
job3
job4

if specified -m 1 we want:
job4

-m 2:
job3
job4

nvflare/private/fed/server/job_cmds.py
```diff
@@ -327,7 +327,10 @@ def list_jobs(self, conn: Connection, args: List[str]):
...
 
                 if max_jobs_listed:
-                    filtered_jobs = filtered_jobs[:max_jobs_listed]
+                    if reverse:
```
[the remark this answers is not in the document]
nvidianz: It always returns most recent jobs so -r -m 2 will return

Job4
Job3

PR #1558, opened 2023-03-21
Fix cell timing

yanchengnv:
### Description

- Once cell is created, there is a delay to get it fully set into the engine. If an aux message is sent before this happens, it would result in an error (NVBug 4035152). The fix is to wait for the cell in the engine to become available before sending the message.
- In fed_client_base, it was waiting for client_runner to be created before proceeding.  However the check interval was set to be too high (0.5 sec), and during this time, the cell won't set to the client_runner's engine, even if it's already created long before that. The fix is to reduce the check interval to a very small number (0.005). This makes the runner almost immediately available. 
- Fixed a typo in the spelling of heart beat (it was hear_beat)
- Fixed is_cell_reachable and is_cell_connected methods in cell.py to consider local cells. If the target cell is a local cell (i.e. within the same process), then these two methods return True. This fix is necessary to support simulator where one process contains many cells.
- Improved the stats category names to to help differentiate request and reply of the same message channel/topic.
- Added type checking to flare_api for list_jobs (NVBug 4035678)
- Fixed list_job command on server to return empty list if jobs are not found.
- Make messages in client_command channel option to suppress unnecessary error logs (NVBug 4037340, 4037082)
- Changed "invalid client" to "unknown client" in some error text to avoid API status computation error in "cat" command (NVBug 4037160)

### Types of changes

- [ ] Documentation updated.


nvflare/apis/shareable.py
```diff
@@ -137,3 +138,19 @@ def make_reply(rc, headers=None) -> Shareable:
...
             reply.set_header(k, v)
     return reply
+
+
+def make_copy(source: Shareable) -> Shareable:
+    """
+    Make a copy from the source.
+    The content (non-headers) will be kept intact. Headers will be deep-copied into the new instance.
+    """
+    assert isinstance(source, Shareable)
```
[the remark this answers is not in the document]
yanchengnv: I usually do this for the ease of code completion by the IDE.

nvflare/private/fed/server/fed_server.py
```diff
@@ -636,7 +636,7 @@ def authentication_check(self, request: Message, state_check):
         if client_ssid != self.server_state.ssid:
             # return make_reply(ReturnCode.AUTHENTICATION_ERROR, "Request from invalid client SSID",
             #                   fobs.dumps(None))
-            error = "Request from invalid client SSID"
+            error = "Request from unknown client SSID"
```
YuanTingHsieh, line 639: Should we make it more detailed? 
"Request has a different SSID than the server side's SSID"

nvflare/private/fed/server/server_command_agent.py
```diff
@@ -73,7 +73,7 @@ def execute_command(self, request: CellMessage) -> CellMessage:
                 if not client:
                     return make_reply(
                         ReturnCode.AUTHENTICATION_ERROR,
-                        "Request from invalid client: missing client token",
+                        "Request from client: missing client token",
```
YuanTingHsieh, line 76: "Request missing client login token or has an invalid token"

nvflare/private/fed/server/message_send.py
```diff
@@ -74,6 +74,7 @@ def send_requests(
         if job_id:
             fqcn = FQCN.join([client.name, job_id])
             channel = CellChannel.CLIENT_COMMAND
+            optional = True
```
YuanTingHsieh, line 77: Why is this optional?
I think we have messages sent via this channel that are required?

nvflare/private/fed/client/client_executor.py
```diff
@@ -227,6 +227,7 @@ def check_status(self, job_id):
                     channel=CellChannel.CLIENT_COMMAND,
                     topic=AdminCommandNames.CHECK_STATUS,
                     request=request,
+                    optional=True,
```
YuanTingHsieh, line 230: How do we decide what commands are optional?

yanchengnv: Commands are not optional. The messages are marked optional for the sole purpose of suppressing unnecessary log cell error messages (they are instead logged at DEBUG level).

nvflare/private/fed/server/fed_server.py
yanchengnv: Not for now. This change is simply to avoid the word "invalid client" because of the API status determination issue.

nvflare/private/fed/server/message_send.py
yanchengnv: Because all messages for CLIENT COMMAND channel should be optional to suppress unnecessary CELL ERROR messages.

nvflare/private/fed/server/server_command_agent.py
yanchengnv: Looks like it could be either case: the token is missing, or client couldn't found from the token.
Could be changed to "missing or invalid token".

nvflare/private/fed/client/client_executor.py
YuanTingHsieh: got it, thanks for the clarification

nvflare/private/fed/server/fed_server.py
YuanTingHsieh: got it, we might need to enhance this message later.

yhwen (approved):

PR #1559, opened 2023-03-21
update comments and exception messages[skip ci]

chesterxgchen:
Fixes #  4035778, FLARE-1149

### Description
1. Update PSI comment and Exception message
2.  local PSI load_items() is called repeated due to the nature of PSI pair-wise intersection calculation. This may result unexpected consequences if the random.sample() is used in the load_items.  We can item cache at executor level, so the load_items() is only called once, to avoid the issue. 

### Types of changes

- [ ] Documentation updated.


nvkevlu (approved):
LGTM, thanks

nvidianz (approved):

PR #1561, opened 2023-03-21
Copy headers only in process_task_request

YuanTingHsieh:
### Description

A more efficient way of https://github.com/NVIDIA/NVFlare/pull/1549

### Types of changes

- [ ] Documentation updated.


yhwen (commented):
We can not simply change from deepcopy to make_copy. The simulator get_task exception has to change, otherwise it will run into another over execution issue.

PR #1563, opened 2023-03-21
Fix abort job with only connected clients

YuanTingHsieh:
### Description

When we abort a job, we should just send abort job messages to the clients that are connected.
The clients that are connected at job launching time might be already dead/dropped.

### Types of changes

- [ ] Documentation updated.


yhwen (approved):
LGMT

PR #1565, opened 2023-03-22
Controller no deepcopy

yhwen:
Fixes # .

### Description

Optimized the controller not to use deepcopy task_data. Fixed the simulator get_task exception handling.

### Types of changes

- [ ] Documentation updated.


nvflare/private/fed/app/simulator/simulator_worker.py
```diff
@@ -79,13 +82,24 @@ def do_one_task(self, client):
...
-                self.logger.info(f"Finished one task run for client: {client.client_name}")
-
-                # if any client got the END_RUN event, stop the simulator run.
-                if client_runner.end_run_fired or client_runner.asked_to_stop:
-                    stop_run = True
-                    self.logger.info("End the Simulator run.")
+                run_task_tries = 0
+                while True:
+                    interval, task_processed = client_runner.fetch_and_run_one_task(fl_ctx)
+                    self.logger.info(f"Finished one task run for client: {client.client_name}")
```
yanchengnv, line 88: Maybe should include "interval" and "task_processed" in the log message so we know whether it is retry or the task is really done.

yanchengnv (changes requested):
Please improve the log message.

nvflare/private/fed/app/simulator/simulator_worker.py
yhwen: Added

nvflare/private/fed/app/simulator/simulator_worker.py
```diff
@@ -79,13 +82,27 @@ def do_one_task(self, client):
...
+                        self.logger.info("End the Simulator run.")
+                        break
+                    else:
+                        if task_processed:
+                            break
+                        else:
+                            run_task_tries += 1
+                            if run_task_tries >= FETCH_TASK_RUN_RETRY:
+                                break
+                            time.sleep(0.5)
```
YuanTingHsieh, line 105: just a note, we have all these random time.sleep throughout our code base, we need to enhance them in the future (have a file that has all these values) so people at least know where to find/modify these.

nvflare/private/fed/app/simulator/simulator_worker.py
```diff
@@ -79,13 +82,27 @@ def do_one_task(self, client):
...
+                        stop_run = True
+                        self.logger.info("End the Simulator run.")
+                        break
+                    else:
+                        if task_processed:
+                            break
+                        else:
+                            run_task_tries += 1
+                            if run_task_tries >= FETCH_TASK_RUN_RETRY:
+                                break
```
YuanTingHsieh, line 104: when we reach the max attempt of `if run_task_tries >= FETCH_TASK_RUN_RETRY:`
and break the loop.

What would user see?
Can we let user know we stop working because `FETCH_TASK_RUN_RETRY = 3` has reached?

yhwen: Retry has the same effect as break out from the loop, in terms of application workflow. The purpose of the retry is to let the client try a couple of more time if getting no task to execute, for example: TRY_AGAIN task. If break out from the loop, the client will actually do the same of fetch_and_run.

nvflare/private/fed/app/simulator/simulator_worker.py
yhwen: Will address this in future.

YuanTingHsieh (approved):

PR #1571, opened 2023-03-22
Fix SAG client result error handling

YuanTingHsieh:
### Description

The return code should explain the failing reasons.
And return code other than OK is treated as failed

### Types of changes

- [ ] Documentation updated.


holgerroth (approved):
Makes sense!

PR #1580, opened 2023-03-24
update POC and tutorial storage locations [skip ci]

chesterxgchen:
Fixes # .

### Description

Issue, the default storage locations for job and snapshot are 
/tmp/nvflare/jobs-storage
/tmp/nvflare/snapshots-storage

This is true for both production and POC mode. 
In order to isolate POC run from the production run, we like to move the location from /tmp/nvflare to /tmp/nvflare/poc the default POC workspace. If NVFLARE_POC_WORKSPCAE is set to different location, we will update the storage location as well. 
This way, when we cleanup the POC workspace. all the related storages are also cleaned

Similarly for tutorial setup, if we provisioned and run FL system in production mode for  notebooks, we would like put the storage inside the provision workspace, so we can cleanup after use. 

Chanages
1. add utility function to update storage location for resources.json
2. update POC command to update resources.json when workspace is not default
3. update default POC resources storage to /tmp/nvflare/poc/xxx-storage
4. update tu_pres_start.sh to update resources.json

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/app_common/workflows/scatter_and_gather.py
+++ b/nvflare/app_common/workflows/scatter_and_gather.py
@@ -336,34 +336,18 @@ class ScatterAndGather(Controller):
         # Raise errors if bad peer context or execution exception.
         if rc and rc != ReturnCode.OK:
             if self.ignore_result_error:
-                self.log_error(
+                self.log_warning(
                     fl_ctx,
                     f"Ignore the train result from {client_name} at round {self._current_round}. Train result error code: {rc}",
                 )
                 return False
             else:
-                if rc in [ReturnCode.MISSING_PEER_CONTEXT, ReturnCode.BAD_PEER_CONTEXT]:
-                    self.system_panic(
-                        f"Peer context is bad or missing. ScatterAndGather exiting at round {self._current_round}.",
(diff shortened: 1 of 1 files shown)


nvflare/lighter/utils.py
```diff
@@ -146,3 +146,29 @@ def sh_replace(src, mapping_dict):
...
+
+def update_storage_locations(local_dir: str,
+                             workspace: str,
+                             default_resource_name: str = "resources.json.default",
+                             job_storage_name: str = "jobs-storage",
+                             snapshot_storage_name: str = "snapshot-storage"):
+    default_resource = f"{local_dir}/{default_resource_name}"
+    target_resource = f"{local_dir}/resources.json"
+    job_storage = f"{workspace}/{job_storage_name}"
+    snapshot_storage = f"{workspace}/{snapshot_storage_name}"
```
YuanTingHsieh, line 159: logic LGTM, we have these strings defined in

```
./nvflare/apis/fl_constant.py
```

But lighter should be before apis, so we can think about how to deal with these common strings later.

YuanTingHsieh (approved):

PR #1608, opened 2023-03-29
Fixes several shutdown related issues

nvidianz:
Fixes FLARE-1170

### Description

1. Added an exit function to stop communicator in case upper layer doesn't do it. This gets rid of the "Broken pipe" errors.
2. Moved setting up close_cb for communicator.stop closer to start() to reduce chances of close_cb not registering.
3. Changed default internal protocol from UDS to TCP. We observed some hiccups in UDS connection.

### Types of changes

- [ ] Documentation updated.


nvflare/private/fed/app/simulator/simulator_worker.py
```diff
@@ -187,6 +187,7 @@ def _create_client_cell(self, federated_client, root_url, parent_url):
             parent_url=parent_url,
         )
         cell.start()
+        mpm.add_cleanup_cb(cell.stop)
```
YuanTingHsieh, line 190: why do we make this change?
this still after cell.start?

nvflare/private/fed/server/fed_server.py
```diff
@@ -381,13 +381,12 @@ def create_job_cell(self, job_id, root_url, parent_url, secure_train, server_con
 
         cell.start()
         net_agent = NetAgent(cell)
+        mpm.add_cleanup_cb(net_agent.close)
+        mpm.add_cleanup_cb(cell.stop)
```
YuanTingHsieh, line 385: why do we make this change?
this still after cell.start?

nvflare/private/fed/app/simulator/simulator_worker.py
```diff
@@ -187,6 +187,7 @@ def _create_client_cell(self, federated_client, root_url, parent_url):
             parent_url=parent_url,
         )
         cell.start()
```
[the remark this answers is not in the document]
nvidianz: Yes. We can't guarantee everything is cleaned up. We just try our best. The process is ending anyway so this doesn't affect function. It just causes some weird log messages.

An exit_func is added in Communicator so all the connections are properly closed even if there is no cleanup CB.

nvflare/fuel/f3/communicator.py
```diff
@@ -48,15 +61,25 @@ def start(self):
...
 
         Raises:
             CommError: If any error encountered while shutting down
         """
+        if self.stopped:
+            return
+
         self.conn_manager.stop()
-        log.debug(f"Communicator for local endpoint: {self.local_endpoint.name} has stopped")
+        self.stopped = True
```
yanchengnv, line 76: should this line be before self.conn_manager.stop()?

yanchengnv (changes requested):
See my comments

nvflare/fuel/f3/communicator.py
nvidianz: If stop() throws exception, this will allow exit_func to try it again.

PR #1611, opened 2023-03-30
Don't submit update with task data from old SSID

YuanTingHsieh:
### Problem

When in HA mode:

Start 2 servers 2 clients
Submit a job so server1 and 2 clients is running it.
The clients pull tasks from this old job.
Then in between I shutdown the primary server (server1), so the second server starts and resume the job.
And then those two clients trigger switch-ssid logic and aborted their old task (also setting return code to aborted)
Then those two clients still sent back the result of the tasks.
The second server receives those incorrect results.

### Description

In this PR, we add a check on client side to NOT send updates if the SSID has changed.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/private/fed/client/communicator.py
+++ b/nvflare/private/fed/client/communicator.py
@@ -23,7 +23,7 @@ from nvflare.apis.fl_constant import ReturnCode as ShareableRC
 from nvflare.apis.fl_constant import ServerCommandKey, ServerCommandNames
 from nvflare.apis.fl_context import FLContext
 from nvflare.apis.fl_exception import FLCommunicationError
-from nvflare.apis.shareable import ReservedHeaderKey, Shareable
+from nvflare.apis.shareable import Shareable
 from nvflare.apis.utils.fl_context_utils import get_serializable_data
 from nvflare.fuel.f3.cellnet.cell import FQCN, Cell
 from nvflare.fuel.f3.cellnet.defs import MessageHeaderKey, ReturnCode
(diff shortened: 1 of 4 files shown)


nvflare/private/fed/client/communicator.py
```diff
@@ -195,25 +195,26 @@ def pull_task(self, servers, project_name, token, ssid, fl_ctx: FLContext):
             size = len(task.payload)
             task.payload = fobs.loads(task.payload)
             task_name = task.payload.get_header(ServerCommandKey.TASK_NAME)
+            task.payload.add_cookie(ReservedHeaderKey.SSID, ssid)
             if task_name not in [SpecialTaskName.END_RUN, SpecialTaskName.TRY_AGAIN]:
                 self.logger.info(
                     f"Received from {project_name} server "
-                    f" ({size} Bytes). getTask: {task_name} time: {end_time - start_time} seconds"
+                    f" ({size} Bytes). pull_task: {task_name} time: {end_time - start_time} seconds"
```
yanchengnv, line 202: Let's not change these words - people have been looking for these words in their work.

nvflare/private/fed/client/communicator.py
```diff
@@ -195,25 +195,26 @@ def pull_task(self, servers, project_name, token, ssid, fl_ctx: FLContext):
...
                 )
         elif return_code == ReturnCode.AUTHENTICATION_ERROR:
-            self.logger.warning("get_task request authentication failed.")
+            self.logger.warning("pull_task request authentication failed.")
             time.sleep(5.0)
             return None
         else:
             task = None
-            self.logger.warning(f"Failed to get_task from {project_name} server. Will try it again.")
+            self.logger.warning(f"Failed to pull_task from {project_name} server. Will try it again.")
```
yanchengnv, line 210: Let's not change this.

nvflare/private/fed/client/communicator.py
```diff
@@ -195,25 +195,26 @@ def pull_task(self, servers, project_name, token, ssid, fl_ctx: FLContext):
             size = len(task.payload)
             task.payload = fobs.loads(task.payload)
             task_name = task.payload.get_header(ServerCommandKey.TASK_NAME)
+            task.payload.add_cookie(ReservedHeaderKey.SSID, ssid)
```
yanchengnv, line 198: Change to use fl_ctx, not cookie, which is a kludge.

nvflare/private/fed/client/communicator.py
```diff
@@ -235,6 +236,10 @@ def submit_update(
 
         # shareable.add_cookie(name=FLContextKey.TASK_ID, data=task_id)
         shareable.set_header(FLContextKey.TASK_NAME, execute_task_name)
+        task_ssid = shareable.get_cookie(ReservedHeaderKey.SSID)
```
yanchengnv, line 239: Change to use fl_ctx.

yanchengnv (changes requested):
See my comments

nvflare/private/fed/client/communicator.py
YuanTingHsieh: changed

nvflare/private/fed/client/communicator.py
YuanTingHsieh: thanks, fixed.

nvflare/private/fed/client/communicator.py
YuanTingHsieh: thanks, fixed.

nvflare/private/fed/client/communicator.py
YuanTingHsieh: thanks, fixed using fl_ctx.

yanchengnv (approved):
LGTM

PR #1613, opened 2023-03-30
Job status management enhancement

yhwen:
Fixes # .

### Description

- Allow abort_job in SUBMITTED status to ABORTED
- Only when config to use HttpOverseerAgent then running the system in HA mode
- When not in HA mode, don't generate snapshot, also don't restore jobs from the snapshot
- When starting the server in None HA mode, update the previous remaining RUNNING jobs to FINISHED:ABANDONED
- check and ensure only job in SUBMITTED status to be able to deploy
- check and ensure only job in DISPATCHED status to be able to start run

### Types of changes

- [ ] Documentation updated.


nvflare/private/fed/server/job_cmds.py
```diff
@@ -407,12 +407,20 @@ def abort_job(self, conn: Connection, args: List[str]):
...
             job_id = conn.get_prop(self.JOB_ID)
             with engine.new_context() as fl_ctx:
-                message = job_runner.stop_run(job_id, fl_ctx)
-                if message:
-                    conn.append_error(message, meta=make_meta(MetaStatusValue.INTERNAL_ERROR, message))
-                else:
-                    conn.append_string("Abort signal has been sent to the server app.")
+                job_manager = engine.get_component(SystemComponents.JOB_MANAGER)
+                job = job_manager.get_job(job_id, fl_ctx)
+                if job.meta.get(JobMetaKey.STATUS) in [RunStatus.SUBMITTED, RunStatus.DISPATCHED]:
```
yanchengnv, line 412: I wonder whether we need DISPATCHED status. Does it serve us any good? What if the system crashed when the job is left at DISPATCHED status? If the job is still in SUBMITTED status, it will get a chance to be rescheduled.

nvflare/private/fed/server/job_cmds.py
```diff
@@ -407,12 +407,20 @@ def abort_job(self, conn: Connection, args: List[str]):
         try:
             job_id = conn.get_prop(self.JOB_ID)
             with engine.new_context() as fl_ctx:
-                message = job_runner.stop_run(job_id, fl_ctx)
-                if message:
-                    conn.append_error(message, meta=make_meta(MetaStatusValue.INTERNAL_ERROR, message))
-                else:
-                    conn.append_string("Abort signal has been sent to the server app.")
+                job_manager = engine.get_component(SystemComponents.JOB_MANAGER)
+                job = job_manager.get_job(job_id, fl_ctx)
```
yanchengnv, line 411: You shouldn't need to make this call. You could simply do:

job = conn.get_prop(self.JOB).

This is because the job authorizer already did job_manager.get_job() for you.

yanchengnv (changes requested):
Please address raised issues.

nvflare/private/fed/server/job_cmds.py
yhwen: DISPATCHED status is a transition state after the job deployed to all the parties. If for some reasons the job got deployed but could not get into RUNNIG state, we want to have manual investigation into this situation, instead of leaving it to SUBMITTED, then may recursively get into this again.

nvflare/private/fed/server/job_cmds.py
yhwen: changed.

nvflare/private/fed/server/job_cmds.py
```diff
@@ -407,12 +407,20 @@ def abort_job(self, conn: Connection, args: List[str]):
...
+                if job.meta.get(JobMetaKey.STATUS) in [RunStatus.SUBMITTED, RunStatus.DISPATCHED]:
+                    job_manager.set_status(job.job_id, RunStatus.FINISHED_ABORTED, fl_ctx)
+                    conn.append_string(f"Abort the job: {job_id} before run.")
```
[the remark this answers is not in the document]
yhwen: This already handled in the following:

conn.append_success("", make_meta(MetaStatusValue.OK))

nvflare/private/fed/server/job_cmds.py
```diff
@@ -407,12 +407,20 @@ def abort_job(self, conn: Connection, args: List[str]):
...
                     conn.append_success("", make_meta(MetaStatusValue.OK))
+                    return
+                else:
```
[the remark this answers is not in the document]
yhwen: This already handled. For those FINISHED:xxx job, it will result in the following error message:

Error: Job xxxxx is not running.

nvflare/private/fed/server/job_cmds.py
```diff
@@ -407,12 +407,20 @@ def abort_job(self, conn: Connection, args: List[str]):
...
             with engine.new_context() as fl_ctx:
-                message = job_runner.stop_run(job_id, fl_ctx)
-                if message:
-                    conn.append_error(message, meta=make_meta(MetaStatusValue.INTERNAL_ERROR, message))
-                else:
-                    conn.append_string("Abort signal has been sent to the server app.")
+                job_manager = engine.get_component(SystemComponents.JOB_MANAGER)
+                job = job_manager.get_job(job_id, fl_ctx)
+                if job.meta.get(JobMetaKey.STATUS) in [RunStatus.SUBMITTED, RunStatus.DISPATCHED]:
+                    job_manager.set_status(job.job_id, RunStatus.FINISHED_ABORTED, fl_ctx)
```
YuanTingHsieh, line 413: will we have other threads calling job_manager.set_status to the same job at the same time?

nvflare/private/fed/server/job_runner.py
```diff
@@ -385,6 +385,9 @@ def run(self, fl_ctx: FLContext):
                     )
 
                     if ready_job:
+                        reload_job = job_manager.get_job(ready_job.job_id, fl_ctx)
+                        if reload_job.meta.get(JobMetaKey.STATUS) != RunStatus.SUBMITTED:
+                            continue
```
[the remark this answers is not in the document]
yhwen: added

nvflare/private/fed/server/job_runner.py
```diff
@@ -385,6 +385,9 @@ def run(self, fl_ctx: FLContext):
                     )
 
                     if ready_job:
+                        reload_job = job_manager.get_job(ready_job.job_id, fl_ctx)
+                        if reload_job.meta.get(JobMetaKey.STATUS) != RunStatus.SUBMITTED:
```
YuanTingHsieh, line 389: What do you think if we wrap this two lines as a simple method like
`_check_job_status(job_manager, job_id, job_run_status)`
or add it in job_manager class.

And then in here we just call `if _check_job_status(job_manager, ready_job.job_id, RunStatus.SUBMITTED)`

nvflare/private/fed/server/job_runner.py
```diff
@@ -417,6 +420,10 @@ def run(self, fl_ctx: FLContext):
                             else:
                                 deployable_clients = client_sites
 
+                            job = job_manager.get_job(ready_job.job_id, fl_ctx)
+                            if job.meta.get(JobMetaKey.STATUS) != RunStatus.DISPATCHED:
```
YuanTingHsieh, line 424: What do you think if we wrap this two lines as a simple method like
`_check_job_status(job_manager, job_id, job_run_status)`
or add it in job_manager class.

And then here we just call `if _check_job_status(job_manager, ready_job.job_id, RunStatus.DISPATCHED)`

nvflare/private/fed/server/job_runner.py
```diff
@@ -470,6 +477,24 @@ def restore_running_job(self, run_number: str, job_id: str, job_clients, snapsho
...
+    def update_unfinished_jobs(self, fl_ctx: FLContext):
+        engine = fl_ctx.get_engine()
+        job_manager = engine.get_component(SystemComponents.JOB_MANAGER)
+        all_jobs = []
+        dispatched_jobs = job_manager.get_jobs_by_status(RunStatus.DISPATCHED, fl_ctx)
+        all_jobs.extend(dispatched_jobs)
+        running_jobs = job_manager.get_jobs_by_status(RunStatus.RUNNING, fl_ctx)
+        all_jobs.extend(running_jobs)
+
+        for job in all_jobs:
```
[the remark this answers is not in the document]
yhwen: This update_unfinished_jobs() will only be called in non-HA mode.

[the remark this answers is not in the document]
YuanTingHsieh: Regarding question 2 and 3:

I think this method only called in non-HA mode
in nvflare/private/fed/server/fed_server.py: _turn_to_hot when ha is False.

nvflare/private/fed/server/job_cmds.py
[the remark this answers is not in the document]
yanchengnv: No, we shouldn't rely on job_runner.stop_run() to tell us that, especially when the status is already FINISHED:XXX, we shouldn't even have to call it.

The returned message is a str for the user to read, but hard for flare_api to digest - it only reads stuff in meta.

We should explicitly check FINISHED: status, and then:
conn.append_string(f"Job {job_id} is not running", meta=make_meta(MetaStatusValue.OK, "job not running"))
The important thing here is MetaStatusValue.OK; otherwise flare_api would raise INTERNAL_ERROR exception.

yanchengnv (changes requested):
The job status handling need to address FINISHED status explicitly.
Please resolve other unresolved issues too.

nvflare/private/fed/server/job_runner.py
```diff
@@ -470,6 +477,24 @@ def restore_running_job(self, run_number: str, job_id: str, job_clients, snapsho
...
+            try:
+                job_manager.set_status(job.job_id, RunStatus.ABANDONED, fl_ctx)
+                self.logger.info(f"Update the previous running job: {job.job_id} to FAILED_TO_FINISH.")
```
YuanTingHsieh, line 492: suggest change to

`self.log_info(fl_ctx, f"Update the previous running job: {job.job_id} to {RunStatus.ABANDONED}.")`

nvflare/private/fed/server/job_runner.py
```diff
@@ -470,6 +477,24 @@ def restore_running_job(self, run_number: str, job_id: str, job_clients, snapsho
...
+            except Exception as e:
+                self.log_error(
+                    fl_ctx, f"Failed to update the job: {job.job_id} to FAILED_TO_FINISH: {secure_format_exception(e)}."
+                )
```
YuanTingHsieh, line 496: suggest use 

``self.log_exception(fl_ctx, f"Failed to update the job: {job.job_id} to {RunStatus.ABANDONED}.")``

nvflare/private/fed/server/job_cmds.py
```diff
@@ -407,12 +407,20 @@ def abort_job(self, conn: Connection, args: List[str]):
         try:
             job_id = conn.get_prop(self.JOB_ID)
             with engine.new_context() as fl_ctx:
-                message = job_runner.stop_run(job_id, fl_ctx)
-                if message:
-                    conn.append_error(message, meta=make_meta(MetaStatusValue.INTERNAL_ERROR, message))
-                else:
-                    conn.append_string("Abort signal has been sent to the server app.")
+                job_manager = engine.get_component(SystemComponents.JOB_MANAGER)
```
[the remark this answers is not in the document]
yhwen: changed to use the same.

nvflare/private/fed/server/job_cmds.py
yhwen: we do have other places update the job status as well in JobRunner. We already add the protection to handle this as much as we can.

nvflare/private/fed/server/job_runner.py
yhwen: yeah, that's a little cleaner. changed it.

nvflare/private/fed/server/job_runner.py
yhwen: changed it.

nvflare/private/fed/server/job_runner.py
yhwen: changed.

nvflare/private/fed/server/job_runner.py
yhwen: changed.

nvflare/private/fed/server/job_cmds.py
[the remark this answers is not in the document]
yhwen: done.

nvflare/private/fed/server/job_runner.py
```diff
@@ -385,8 +385,8 @@ def run(self, fl_ctx: FLContext):
                     )
 
                     if ready_job:
-                        reload_job = job_manager.get_job(ready_job.job_id, fl_ctx)
-                        if reload_job.meta.get(JobMetaKey.STATUS) != RunStatus.SUBMITTED:
+                        if self._check_job_status(job_manager, ready_job.job_id, RunStatus.SUBMITTED, fl_ctx):
```
yanchengnv, line 388: It took me a moment to understand that "_check_job_status" actually checks whether the job is NOT in the specified status. We should rename this function to something like _job_not_in_status(), so that the code can be more readable.

yanchengnv (changes requested):
Please change the func name _check_job_status to be more readable.

nvflare/private/fed/server/job_runner.py
yanchengnv: Let's fix it later so we can build a RC tonight.

yanchengnv (approved):
Let's change the func name later.

PR #1614, opened 2023-03-30
Make consistent the error message for shutdown_system without auth

nvkevlu:
There were inconsistencies in the error messages for shutdown_system depending on whether the admin was part of the overseer's org. With this change, the error messages should be consistent.

### Types of changes

- [ ] Documentation updated.


yanchengnv (changes requested):
See my comments.

nvflare/ha/ha_admin_cmds.py
```diff
@@ -15,78 +15,20 @@
...
-    else:
-        return {"status": APIStatus.SUCCESS, "details": "Set state to shutdown in overseer."}
-
-
 class HACommandModule(CommandModule):
     """Command module with commands for management in relation to the high availability framework."""
 
     def __init__(self, overseer_agent: OverseerAgent):
-        self.overseer_agent = overseer_agent
+        self.overseer_agent = overseer_agent  # we should not need self.overseer_agent anymore
```
yanchengnv, line 31: Why not? Since this is provided, we should use it in all methods, instead of getting it from service_finder.

nvflare/ha/ha_admin_cmds.py
```diff
@@ -15,78 +15,20 @@
...
         self.logger = logging.getLogger(self.__class__.__name__)
```
YuanTingHsieh, line 32: just a note, can be done in another PR later, these 2 things are not used we can remove them both.

nvflare/ha/ha_admin_cmds.py
```diff
@@ -97,50 +39,53 @@ def get_spec(self):
...
                     usage="shutdown_system",
-                    handler_func=shutdown_system,
+                    handler_func=self.shutdown_system,
                 ),
             ],
         )
 
-    def list_sp(self, args, api):
-        """List service provider information based on the last heartbeat from the overseer.
+    def list_sp(self, args, ctx: CommandContext):
```
YuanTingHsieh, line 65: Just a note, can be done in another PR later.
These methods don't need to be instance method as they are not using "self"

nvflare/private/fed/server/cmd_utils.py
```diff
@@ -130,6 +133,16 @@ def validate_command_targets(self, conn: Connection, args: List[str]) -> str:
         conn.set_prop(self.TARGET_CLIENTS, all_clients)
         return ""
 
+    def must_be_project_admin(self, conn: Connection, args: List[str]):
+        if args[0] == AdminCommandNames.ADMIN_CHECK_STATUS and len(args) == 1:
```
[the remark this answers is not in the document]
nvkevlu: for AdminCommandNames.ADMIN_CHECK_STATUS, authz_func=self.must_be_project_admin in training_cmds.py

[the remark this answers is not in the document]
yanchengnv: Oh, I see we need to extend the args with "server" in it so the handler will see "check_status server". Nice detail!

PR #1615, opened 2023-03-30
fixed the job status for config error.

yhwen:
Fixes # .

### Description

Update the job status to EXECUTION_EXCEPTION if there's config error, the job could not run.

### Types of changes

- [ ] Documentation updated.


nvflare/private/fed/app/server/runner_process.py
```diff
@@ -148,4 +148,5 @@ def main():
     This is the program when starting the child process for running the NVIDIA FLARE server runner.
     """
     # main()
-    mpm.run(main_func=main)
+    rc = mpm.run(main_func=main)
```
YuanTingHsieh, line 151: What is the return code if MPM can't finished cleanup in given time?

yhwen: Added additional FINISHED_ABNORMAL for this particular can't finished cleanup in given time scenario. The return code is -9.

YuanTingHsieh: got it, sounds good.

PR #1616, opened 2023-03-30
Removing UDS

nvidianz:
Fixes FLARE-1209

### Description

Removed UDS drivers and use TCP for all internal connections.

### Types of changes

- [ ] Documentation updated.


nvflare/private/fed/server/fed_server.py
```diff
@@ -155,10 +155,12 @@ def deploy(self, args, grpc_args=None, secure_train=False):
             credentials = {}
         parent_url = None
 
-        if scheme == "uds":
-            listen_target = target
+        parts = target.split(":")
```
YuanTingHsieh, line 158: Why are we making this change?

Do you have examples of different "target" ?

nvidianz: It's kind of future proof. If there is no port, we don't know what to do so we try the original target. We are talking about supporting Azure queues, I don't know how is there destination, probably don't have a port.

YuanTingHsieh: @IsaacYangSLA can you help comment on this^
other than this question changes LGTM

PR #1617, opened 2023-03-30
fix controller dead client handling; added stats pool to_dict

yanchengnv:
### Description

The PR addresses the following issues:
- In controller.py, there is a bug in setting the dead client report time. The report time should only record the first time the report is received, not resetting it repeatedly; otherwise the dead client logic will never be triggered.
- Also changed to use config service for dead client/job processing parameters.
- Added stats pool to_dict logic to enable pool stats to be written to file.
- Modified flare_api to use a generic way to detect job completion.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/apis/impl/controller.py
+++ b/nvflare/apis/impl/controller.py
@@ -24,6 +24,7 @@ from nvflare.apis.fl_context import FLContext
 from nvflare.apis.responder import Responder
 from nvflare.apis.shareable import Shareable, make_copy
 from nvflare.apis.signal import Signal
+from nvflare.fuel.utils.config_service import ConfigService
 from nvflare.security.logging import secure_format_exception
 from nvflare.widgets.info_collector import GroupInfoCollector, InfoCollector
 
@@ -36,6 +37,12 @@ from .task_manager import TaskCheckStatus, TaskManager
 _TASK_KEY_ENGINE = "___engine"
 _TASK_KEY_MANAGER = "___mgr"
 _TASK_KEY_DONE = "___done"
+
+# wait this long since client death report before treating the client as dead
+_CONFIG_VAR_DEAD_CLIENT_GRACE_PERIOD = "dead_client_grace_period"
+
+# wait this long since job schedule time before starting to check dead clients
+_CONFIG_VAR_DEAD_CLIENT_CHECK_LEAD_TIME = "dead_client_check_lead_time"
 
 
 def _check_positive_int(name, value):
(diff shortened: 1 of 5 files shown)


nvflare/apis/impl/controller.py
```diff
@@ -979,10 +988,12 @@ def _check_dead_clients(self):
     def _client_still_alive(self, client_name):
         now = time.time()
         report_time = self._dead_client_reports.get(client_name, None)
+        grace_period = ConfigService.get_float_var(name=_CONFIG_VAR_DEAD_CLIENT_GRACE_PERIOD, default=30.0)
```
YuanTingHsieh, line 991: just a note, in the future it might be good if we have a common place to keep these default values for timing so that people can just check that

nvflare/fuel/utils/stats_utils.py
```diff
@@ -311,3 +311,23 @@ def new_message_size_pool(name: str, description="", marks=None) -> HistPool:
     if not marks:
         marks = (0.01, 0.1, 1, 10, 50, 100, 200, 500, 800, 1000)
     return HistPool(name=name, description=description, marks=marks, unit="MB")
+
+
+VALID_HIST_MODES = [StatsMode.COUNT, StatsMode.PERCENT, StatsMode.AVERAGE, StatsMode.MAX, StatsMode.MIN]
+
+
+def parse_hist_mode(mode: str) -> str:
```
YuanTingHsieh, line 319: do we have any usages of this method?

nvflare/fuel/flare_api/flare_api.py
```diff
@@ -431,11 +431,7 @@ def monitor_job(
             if not job_status:
                 raise InternalError(f"missing status in job {job_id}")
 
-            if job_status in [
-                RunStatus.FINISHED_EXECUTION_EXCEPTION.value,
-                RunStatus.FINISHED_COMPLETED.value,
-                RunStatus.FINISHED_ABORTED,
-            ]:
+            if job_status.startswith("FINISHED"):
```
YuanTingHsieh, line 434: If not using `RunStatus` anymore needs to remove the import otherwise the style tests failed.

nvflare/apis/impl/controller.py
yanchengnv: Yes, I'm thinking the same thing. We'll need to centrally manage all such control vars so they can be easily documented and explained to other people.

nvflare/fuel/flare_api/flare_api.py
yanchengnv: Ah, good catch. Will change.

nvflare/fuel/utils/stats_utils.py
yanchengnv: Another good catch! It was taken out from _parse_mode from net_manager.py, and intended to be used by the StatsViewer tool (to be checked in separate PR). I need to remove _parse_mode from net_manager and use this func.

nvflare/fuel/flare_api/flare_api.py
yanchengnv: Fixed

nvflare/fuel/utils/stats_utils.py
yanchengnv: Changed net_manager to use this func.

YuanTingHsieh (approved):
Thanks, LGTM

PR #1618, opened 2023-03-30
Fix various notebook bugs [skip ci]

chesterxgchen:
### Description

There were several issues with notebooks including 
1) using Server1 instead of localhost for server name
2) another bugs

we modify the code to fix those issues.  During test, we also noticed there is some issues with running tu_api_notebooks.sh (where we start/run/stop pieces are together). So for now, we remove that scripts
Also, the shutdown is not equallently kill, a proper shutdown ( gracefully) will be needed in another PR

### Types of changes

- [ ] Documentation updated.


examples/tutorials/flare_api.ipynb
```diff
@@ -160,8 +159,31 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "b162ce0e-e219-475d-a014-9a7a23145239",
+   "metadata": {
+    "tags": []
+   },
+   "outputs": [],
+   "source": [
     "! tree ../hello-world/hello-numpy-sag {provision_dir}/{username}/transfer/"
```
YuanTingHsieh, line 187: is this tree command viewing `../hello-world/hello-numpy-sag`? 
or `{provision_dir}/{username}/transfer/`?

when you pass two params into tree, what does it do?

nvflare/lighter/provision.py
```diff
@@ -63,7 +63,6 @@ def handle_provision(args):
     current_path = os.getcwd()
     custom_folder_path = os.path.join(current_path, args.custom_folder)
     sys.path.append(custom_folder_path)
-    print("\nPath list (sys.path) for python codes loading: {} \n".format(sys.path))
```
YuanTingHsieh, line 66: why remove this line?
although not sure if this is needed anymore @IsaacYangSLA can you comment on this

YuanTingHsieh (approved):

nvidianz (approved):

PR #1619, opened 2023-03-30
Do not shutdown job runner when server turn to cold state

YuanTingHsieh:
### Description

If we have 2 servers, and keep promoting each other.
p1 -> p2 -> p1 (job runner dead)

We will shutdown the job runner then server is not functioning.

After HOT turn to COLD.
We should NOT shutdown job runner, but instead have it standby.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/private/fed/server/fed_server.py
+++ b/nvflare/private/fed/server/fed_server.py
@@ -509,18 +509,19 @@ class FederatedServer(BaseServer):
             if self.admin_server:
                 self.admin_server.client_heartbeat(token, client_name)
 
-            abort_runs = self._sync_client_jobs(request, token)
             reply = self._generate_reply(
                 headers={CellMessageHeaderKeys.MESSAGE: "Heartbeat response"}, payload=None, fl_ctx=fl_ctx
             )
-            if abort_runs:
-                reply.set_header(CellMessageHeaderKeys.ABORT_JOBS, abort_runs)
+            if isinstance(self.server_state, HotState):
+                abort_runs = self._sync_client_jobs(request, token)
+                if abort_runs:
+                    reply.set_header(CellMessageHeaderKeys.ABORT_JOBS, abort_runs)
 
-                display_runs = ",".join(abort_runs)
-                self.logger.debug(
-                    f"These jobs: {display_runs} are not running on the server. "
-                    f"Ask client: {client_name} to abort these runs."
-                )
+                    display_runs = ",".join(abort_runs)
+                    self.logger.debug(
+                        f"These jobs: {display_runs} are not running on the server. "
+                        f"Ask client: {client_name} to abort these runs."
+                    )
             return reply
 
     def _sync_client_jobs(self, request, client_token):
@@ -532,7 +533,7 @@ class FederatedServer(BaseServer):
         # also check jobs that are running on server but not on the client
         jobs_on_server_but_not_on_client = list(set(server_jobs).difference(client_jobs))
         if jobs_on_server_but_not_on_client:
-            # should this job be running on the client?
+            # notify all the participating clients these jobs are not running on server anymore
             for job_id in jobs_on_server_but_not_on_client:
                 job_info = self.engine.run_processes[job_id]
                 participating_clients = job_info.get(RunProcessKey.PARTICIPANTS, None)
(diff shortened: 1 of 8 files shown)


yanchengnv:
This is a hole in our system. Please do not merge this PR until we have a thorough discussion with the team.

nvflare/private/fed/server/fed_server.py
```diff
@@ -808,6 +808,7 @@ def stop_training(self):
 
     def fl_shutdown(self):
         self.engine.stop_all_jobs()
+        self.engine.stop_job_runner()
```
yhwen, line 811: self.engine.stop_all_jobs() will shutdown all the running jobs. This should not be called when the server turning to Cold, otherwise the running won't be able to switch to new SP and keep running. Instead, we need to work on a new function to stop all the running job processes, but however, not to impact the snapshot storage and job status storage.

yhwen (changes requested):

nvflare/private/fed/server/job_runner.py
```diff
@@ -528,4 +528,3 @@ def stop_all_runs(self, fl_ctx: FLContext):
             self.stop_run(job_id, fl_ctx)
 
         self.log_info(fl_ctx, "Stop all the running jobs.")
-        self.ask_to_stop = True
```
yhwen, line 531: Add back here the:

self.ask_to_stop = True

So the server_engine stop_all_jobs() does not have to call two of the job_runner methods.

self.job_runner.stop_all_runs(fl_ctx)
self.job_runner.stop()

nvflare/private/fed/server/server_engine.py
```diff
@@ -779,6 +780,12 @@ def start_client_job(self, job_id, client_sites):
     def stop_all_jobs(self):
         fl_ctx = self.new_context()
         self.job_runner.stop_all_runs(fl_ctx)
+        self.job_runner.stop()
+
+    def pause_server_jobs(self):
+        running_jobs = list(self.run_processes.keys())
+        for job_id in running_jobs:
+            self.abort_app_on_server(job_id)
```
yhwen, line 788: This will cause the JobRunner to stop the job run and update the job status from the Cold server. Need to add the following to remove the job from the JobRunner running_jobs:

            with self.lock:
                del self.job_runner.running_jobs[job_id]

nvflare/private/fed/server/fed_server.py
```diff
@@ -509,18 +509,19 @@ def client_heartbeat(self, request: Message) -> Message:
...
             )
-            if abort_runs:
-                reply.set_header(CellMessageHeaderKeys.ABORT_JOBS, abort_runs)
-
-                display_runs = ",".join(abort_runs)
-                self.logger.debug(
-                    f"These jobs: {display_runs} are not running on the server. "
-                    f"Ask client: {client_name} to abort these runs."
-                )
+            if isinstance(self.server_state, HotState):
```
yhwen, line 515: No need to make the change for this HotState check. The "self._handle_state_check(state_check, fl_ctx)" before already ensure only HotState can reach here.

yhwen (changes requested):

nvflare/private/fed/server/fed_server.py
YuanTingHsieh: "handle_state_check" only sets a prop inside fl_ctx.

Cold state can also reach here.

nvflare/private/fed/server/job_runner.py
YuanTingHsieh: Ok will do

nvflare/private/fed/server/fed_server.py
yhwen: You are right. But the proper fix is to do something like this:

    def _handle_state_check(self, state_check, fl_ctx: FLContext):
        if state_check.get(ACTION) in [NIS, ABORT_RUN]:
            fl_ctx.set_prop(FLContextKey.COMMUNICATION_ERROR, state_check.get(MESSAGE), sticky=False)
            return False
        return True

And:

            if self._handle_state_check(state_check, fl_ctx):
                return

yhwen (approved):
LGTM now

PR #921, opened 2022-09-22
Update dashboard cli to remove dashboard_image option [skip ci]

IsaacYangSLA:

PR #1622, opened 2023-03-31
[WIP] Support dashboard local deployment and image/version option

parkeraddison:
### Description

This PR includes two feature additions which both have the same impediment.

**Image option:**  
If we start a project with NVFlare v2.x.x, then a new breaking version v3.x.x comes along, I want to continue to host my dashboard with the v2.x.x image. I notice this was abandoned in #921.

**Local deployment:**  
If I'm already in a Docker container, the current dashboard CLI forces me to use docker-in-docker or docker-out-of-docker which both have security implications I'd like to avoid. I want to be able to launch the dashboard without spinning up a new container.

### Types of changes

- [ ] Documentation updated.

### Work done so far
* Added -i/--image option to dashboard/cli.py. If not present, then the dashboard Flask server is launched locally. If present, defaults to latest nvflare/nvflare image and accepts any other image name.
* Adjusted cli and app init to rely on NVFL_WEB_ROOT and NVFL_WEB_PORT environment variables where needed for local support.

### Blockers
The dashboard expects static files at `nvflare/dashboard/application/static/`. This directory seems to _only_ exist within the nvflare/nvflare image. I can't locate the static files in the git repo nor in the installed nvflare python package files. Where do the static files exist?

At the moment, this limitation means that the only supported image is nvflare/nvflare and if you deploy it locally you must do so from a container using the nvflare/nvflare image.

### Additional comments
Ideally the image tags under nvflare/nvflare would reference the version number, e.g. nvflare/nvflare:2.2.5.

### DIFF
--- a/nvflare/dashboard/cli.py
+++ b/nvflare/dashboard/cli.py
@@ -19,6 +19,7 @@ import subprocess
 import sys
 
 import docker
+import nvflare
 from nvflare.apis.utils.format_check import name_check
 from nvflare.dashboard.application.blob import _write
 from nvflare.lighter import utils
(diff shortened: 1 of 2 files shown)


YuanTingHsieh:
@parkeraddison thanks for the contribution!

@IsaacYangSLA is our PIC for dashboard and he will be reviewing and discussing any questions with you.

Thanks.

nvflare/dashboard/cli.py
```diff
@@ -58,47 +59,59 @@ def start(args):
...
-    print(f"{folder} on host mounted to /var/tmp/nvflare/dashboard in container")
-    if environment:
-        print(f"environment vars set to {environment}")
+    if args.image:
+        try:
+            client = docker.from_env()
+        except docker.errors.DockerException:
+            print("Unable to communicate to docker daemon/socket.  Please make sure your docker is up and running.")
+            exit(0)
+        dashboard_image = args.image
```
IsaacYangSLA, line 68: I think once the variable dashboard_image is set by the args.image, the following logic can stay the same.  Is there any reason to indent the following codes?

IsaacYangSLA: @parkeraddison , let me explain our current design.  When starting dashboard locally, users will not be able to see web pages as we only include those html/css/js files in the container images.  To include this option with similar user experiences, this PR might not be sufficient.  Would you like to add some information on the document side?   The file related to dashboard design is https://github.com/NVIDIA/NVFlare/blob/dev/docs/user_guide/dashboard_api.rst.

parkeraddison: That makes sense. Yup, I noticed that those static files are only in the nvflare image...

Maybe I should change the default behavior to use the nvflare image, and instead add an option to the command to launch without using Docker? e.g. --no-image or --no-container  or --local? That way the default behavior remains unchanged.

Then you're right that some additional documentation would be needed to explain that a dashboard can be launched without spinning up a new Docker container *but only if the command is already being ran inside of a container with the static files from the nvflare image already*.

The reason I still think this would be useful to support is because on our team we're already in favor of using Docker for dependency management rather than requiring a specific version of Python and using virtual environments. So, just like with all of the other services (server, client site, and consoles), we just have the host machine spin up a container with nvflare installed which in turn launches the service. Ideally launching the dashboard could follow the same process of not requiring nvflare to be installed on the host machine and allowing the dashboard to be launched within a container.

Or, would it be useful/possible to have an NVFlare CLI as a separate PyPI package with more lenient Python version support, and the CLI simply runs daemon-ish commands such as launching the dashboard/services in other containers?

YuanTingHsieh: @parkeraddison 
I just review our design and understand your pain point now.

I think what you propose is good.
So for `nvflare dashboard --start` we keep the default behavior.

I agree we can add an option say `nvflare dashboard --start --no-image`.

In the implementation part, instead of using the if/else and re-indent the original method.
We could introduce another methods.
Say we rename original to `start_with_docker_image` or something.
And another method we call `start_without_docker_image`.
In there you can put your "else" logic.

You can also introduce new stop method like `stop_without_docker_image` or something if you want.
@IsaacYangSLA ^ what do you think about this?

IsaacYangSLA: @YuanTingHsieh , thanks.  If the original behavior does not change in the PR, and additional option will launch the dashboard locally, then it's OK to merge this PR.
@parkeraddison , once your PR is ready for merge, can you remove [WIP] from the PR title?  Thanks for the contribution.

PR #1625, opened 2023-03-31
Fixed job could not run when overseer is offline.

yhwen:
Fixes # .

### Description

Fixed the issue when overseer is offline, the client job could not start to run issue.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/private/fed/client/client_app_runner.py
+++ b/nvflare/private/fed/client/client_app_runner.py
@@ -38,12 +38,14 @@ class ClientAppRunner(Runner):
         self.timeout = time_out
         self.client_runner = None
 
-    def start_run(self, app_root, args, config_folder, federated_client, secure_train):
+    def start_run(self, app_root, args, config_folder, federated_client, secure_train, sp):
         self.client_runner = self.create_client_runner(app_root, args, config_folder, federated_client, secure_train)
+
         federated_client.set_client_runner(self.client_runner)
-        while federated_client.communicator.cell is None:
-            self.logger.info("Waiting for the client job cell to be created ....")
-            time.sleep(1.0)
+        federated_client.set_primary_sp(sp)
+
+        with self.client_runner.engine.new_context() as fl_ctx:
+            self.start_command_agent(args, federated_client, fl_ctx)
 
         self.sync_up_parents_process(federated_client)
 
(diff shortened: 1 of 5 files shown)


nvflare/private/fed/client/client_app_runner.py
```diff
@@ -103,16 +104,11 @@ def create_run_manager(self, args, conf, federated_client, workspace):
...
-            time.sleep(1.0)
-
+    def start_command_agent(self, args, federated_client, fl_ctx):
         # Start the command agent
-        # self.command_agent = CommandAgent(federated_client, int(args.listen_port), client_runner)
-        self.command_agent = CommandAgent(federated_client, client_runner)
+        self.command_agent = CommandAgent(federated_client)
         self.command_agent.start(fl_ctx)
-        client_runner.command_agent = self.command_agent
+        self.command_agent.register_cell_cb()
```
yanchengnv, line 111: Why do you register_cell_cb in here? It is already called in command_agent.start().

nvflare/private/fed/client/client_app_runner.py
```diff
@@ -38,12 +37,14 @@ def __init__(self, time_out=60.0) -> None:
...
         federated_client.set_client_runner(self.client_runner)
-        while federated_client.communicator.cell is None:
-            self.logger.info("Waiting for the client job cell to be created ....")
-            time.sleep(1.0)
+        federated_client.set_primary_sp(sp)
+
+        with self.client_runner.engine.new_context() as fl_ctx:
+            self.start_command_agent(args, federated_client, fl_ctx)
 
         self.sync_up_parents_process(federated_client)
```
yanchengnv, line 49: This call could raise exception when it fails to get clients info from server root. What's going to happen to the system if this exception occurs? Will the job cell be shutdown?

nvflare/private/fed/client/client_app_runner.py
yhwen: Removed the duplicate calls.

nvflare/private/fed/client/client_app_runner.py
yhwen: If exception happens, the client job process will simply terminate. Error info in the log.

PR #1627, opened 2023-03-31
update the abort_job status after the job complete

yhwen:
Fixes # .

### Description

Fixed an issue the aborted job has no workspace from download_job.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/apis/job_def.py
+++ b/nvflare/apis/job_def.py
@@ -116,6 +116,7 @@ class Job:
         self.submit_time = None
 
         self.run_record = None  # job id, dispatched time/UUID, finished time, completion code (normal, aborted)
+        self.run_aborted = False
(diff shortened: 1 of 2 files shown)


yanchengnv (changes requested):
Don't understand how this PR could fix the aborted job's workspace problem.

yhwen:
> I am not sure about the problem this PR tries to fix. It seems to say that an aborted job has no workspace and hence cannot be downloaded. But the code changes seem to have nothing to do with workspace.

This PR is to ensure after the running job is completed, the running workspace will be saved to the job storage. So when user use the "download_job" command, it will be able to get the workspace included.

yanchengnv (approved):
LGTM

PR #1635, opened 2023-04-01
Scripts refactoring and notebooks bug fixes/update [skip ci]

chesterxgchen:
Fixes # .

### Description
1. refactoring the code to share common used wait for start and wait for shutdown, so it can be used by both POC and prod code. 
2. fix bugs 
3. update both hello-world and flare API notebooks to add start scripts experience. 

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/lighter/utils.py
+++ b/nvflare/lighter/utils.py
@@ -16,17 +16,13 @@ import json
 import os
 import random
 import shutil
-import time
 from base64 import b64decode, b64encode
-from typing import List
 
 import yaml
 from cryptography.hazmat.backends import default_backend
 from cryptography.hazmat.primitives import hashes, serialization
 from cryptography.hazmat.primitives.asymmetric import padding
 
-from nvflare.fuel.flare_api.api_spec import JobNotFound, NoConnection
-from nvflare.fuel.flare_api.flare_api import Session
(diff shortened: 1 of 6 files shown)


nvflare/lighter/utils.py
```diff
@@ -205,3 +209,117 @@ def update_storage_locations(
     json_object = json.dumps(resources, indent=4)
     with open(target_resource, "w") as outfile:
         outfile.write(json_object)
+
+
+def shutdown_system(prod_dir: str, username: str = "admin", secure_mode: bool = False, timeout_in_sec: int = 30):
```
YuanTingHsieh, line 214: Should we put these methods here in "lighter"?

"lighter" should be just for setting up the "flame" or "flare".

These methods maybe should be in "nvflare/tool" or "nvflare/utils".

In addition, I also think flare api is not fuel, as it is higher level stuff.
Other stuffs inside fuel is low level stuff used by flare.

yhwen: I also feel that this should not be in the nvflare/lighter level. It should be within the nvflare tools layer or the nvflare api layer.

yhwen (commented):
Added a minor comment. We can also refactor that later.

yhwen (approved):

PR #1642, opened 2023-04-03
remove the _ensure_daemon_process_shutdown which caused restart fail.

yhwen:
Fixes # .

### Description

remove the _ensure_daemon_process_shutdown in client_train.py which caused restart fail.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/private/fed/app/client/client_train.py
+++ b/nvflare/private/fed/app/client/client_train.py
@@ -18,7 +18,6 @@ import argparse
 import os
 import sys
 import time
-from pathlib import Path
 
 from nvflare.apis.event_type import EventType
 from nvflare.apis.fl_constant import JobConstants, SiteType, WorkspaceConstants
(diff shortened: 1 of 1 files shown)


yanchengnv:
I'm not sure I know enough about this simple change, but will it cause other side effect that may break other features (like true shutdown)? Yuan-Ting and Yuhong please double check.

YuanTingHsieh:
@yanchengnv yes, I discussed with Yuhong on this part.

The PR is a revert of https://github.com/NVIDIA/NVFlare/pull/1578

So I added these logic because when the exception happens I want the child parent process to died immediately.
(by creating this shutdown.fl file so that "./sub_start.sh" will NOT re-start it again)


BUT our "restart client" command is actually creating a "restart.fl" and depends on "./start.sh" to restart the child process.
So added these new logic will cause the "./sub_start.sh" to exit right away thus the restart can't work.

By reverting this change, restart is fixed.

And since our "./sub_start.sh" has a maximum restart tries of 5.
If exceptions happen, we will restart 5 times at most and failed.

We should improve these behavior in future releases.

YuanTingHsieh (approved):

PR #1657, opened 2023-04-05
fix abort_job in old FLAdminAPI

yanchengnv:
Fixes # .

### Description

The abort_job command has been improved to work for all job statuses (running, submitted but not running, finished). However, the old FLAdminAPI could only abort jobs that are running. This PR fixes this.

The fix makes use of the meta info sent from the server, instead of looking for certain string. This approach is cleaner and more accurate.

### Types of changes

- [ ] Documentation updated.


nvkevlu (approved):
LGTM, thanks

YuanTingHsieh (approved):

PR #1663, opened 2023-04-06
[2.3] remove apt_opt in the hello-world and tutorial nvflare install lskip ci]

chesterxgchen:
### Description

The previous changes added the nvflare[apt_op]>= 2.3.0rc3 from nvflare>= 2.3.0rc3. This change includes all the optional dependencies in the initial installation. The intent was avoid to adding additional requirements when running examples.

This actually causing Tensorflow2 example to fail as the it install pytorch and cudnn which is incompatable with Tensorflow2.

In this PR, we remove here and ask user to pip install -r requirements for the individual examples

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh (approved):
This change is fine, it might not solve the issue though

PR #1678, opened 2023-04-10
Clean up AdminAPI

YuanTingHsieh:
### Description

- Remove login/login_with_poc function in `FLAdminAPI` as it conflicts with auto_login
- Our `AdminAPI` required `ServiceFinder` and it can't be None
- Make `auto_login_timeout` and `auto_login_interval` as arguments instead of hard-code
- Fix code style

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/fuel/hci/client/api.py
+++ b/nvflare/fuel/hci/client/api.py
@@ -44,6 +44,9 @@ _CMD_TYPE_UNKNOWN = 0
 _CMD_TYPE_CLIENT = 1
 _CMD_TYPE_SERVER = 2
 
+MAX_AUTO_LOGIN_TRIES = 300
+AUTO_LOGIN_INTERVAL = 1.0
+
 
 class ResultKey(object):
 
@@ -335,8 +338,7 @@ class AdminAPI(AdminAPISpec):
         session_event_cb=None,
         session_timeout_interval=None,
         session_status_check_interval=None,
-        auto_login_timeout: float = 5,
-        auto_login_interval: float = 1,
+        auto_login_max_tries: int = 5,
     ):
         """Underlying API to keep certs, keys and connection information and to execute admin commands through do_command.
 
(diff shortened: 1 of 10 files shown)


yanchengnv:
I am not sure I understand the need of this PR. Let's discuss this first.

nvkevlu:
Looks very good to me. My only suggestion is to add "in seconds" to the auto_login_timeout and auto_login_interval descriptions so it is clearer to the user what unit it is.

nvflare/fuel/hci/client/api.py
```diff
@@ -349,8 +352,10 @@ def __init__(
             poc: Whether to enable poc mode for using the proof of concept example without secure communication.
             debug: Whether to print debug messages, which can help with diagnosing problems. False by default.
             session_event_cb: the session event callback
-            session_timeout_interval: if specified, automatically close the session after inactive for this long
-            session_status_check_interval: how often to check session status with server
+            session_timeout_interval: if specified, automatically close the session after inactive for this long, unit is second
+            session_status_check_interval: how often to check session status with server, unit is second
+            auto_login_timeout: will keep trying to auto-login within this interval, unit is second
+            auto_login_interval: how often to try to auto-login, unit is second
```
yanchengnv, line 358: I don't think we need to allow auto_login_interval to be configurable by the user to prevent stressing out the system (e.g. it's set to 0.0001). It is okay to set to 1.0 sec (or even larger).
Instead of auto_login_timeout, we could do auto_login_max_tries=5
Also need to validate it: must be > 0, and < a preset max (e.g. 300)

nvflare/fuel/hci/client/api.py
```diff
@@ -462,8 +469,15 @@ def _handle_sp_address_change(self, host: str, port: int, ssid: str):
             self.new_ssid = ssid
 
     def _try_auto_login(self):
-        resp = {}
-        for i in range(5):
+        start_time = time.time()
+        while True:
+            if time.time() - start_time >= self.auto_login_timeout:
```
yanchengnv, line 474: Change to test max number of tries.

yanchengnv (changes requested):
Please improve the API's auto login control args.

nvflare/fuel/hci/client/api.py
YuanTingHsieh: thanks, fixed as suggested

nvflare/fuel/hci/client/api.py
YuanTingHsieh: thanks, fixed as suggested

yanchengnv (approved):
LGTM. Please make sure that all places that use FLAdminAPI is updated since you changed the position of overseer_agent.

PR #1685, opened 2023-04-13
Docs fixes links [skip ci]

nvkevlu:
Fixes the minor issues on the overview.rst page with the links and cleans up the unused rows in the examples README table.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/fuel/f3/qat/time_comm.py
+++ b/nvflare/fuel/f3/qat/time_comm.py
@@ -1,0 +1,112 @@
+# Copyright (c) 2023, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+import os
+
+from nvflare.fuel.f3.stats_pool import CsvRecordReader
+
+"""
+This tool can be used to compute the total time spent on communication for a job.
+
+NOTE: if all processes (server and clients) are run on the same host, then the computed results are accurate.
+If processes are run on different hosts, then these hosts must be synchronized with NTP (Network Time Protocol).
+
+Before starting this tool, you must collect the stats_pool_records.csv files of all the processes into a folder. 
+These files must all have the suffix of ".csv". For FL clients, these files are located in their workspaces.
+For FL server, you need to download the job first (using admin console or flare api) and then find it in the downloaded 
+workspace of the job.
+
+Since these files have the same name in their workspaces, you must rename them when copying into the same folder.
+You can simply use the client names for clients and "server" for server file.
+
+Once you have all the csv files in the same folder, you can start this tool with the following args:
+
+    -d: the directory that contains the csv files. Required.
+    -o: the output file that will contain the result. Optional.
+
+If the output file name is not specified, it will be default to "comm.txt".
+The result is printed to the screen and written to the output file.
+
+The output file will be placed into the same folder that contains the csv files. 
+Do not name your output file with the suffix ".csv"!
+
+"""
+
+
+def _print(data: str, out_file):
+    print(data)
+    if out_file is not None:
+        out_file.write(data + "\n")
+
+
+def _compute_time(file_name: str, pool_name: str, out_file):
+    result = 0.0
+    max_time = 0.0
+    min_time = 1000
+    count = 0
+
+    _print(f"Processing record file: {file_name}", out_file)
+    reader = CsvRecordReader(file_name)
+    for rec in reader:
+        if rec.pool_name != pool_name:
+            continue
+        count += 1
+        result += rec.value
+        if max_time < rec.value:
+            max_time = rec.value
+        if min_time > rec.value:
+            min_time = rec.value
+
+    _print(f"    Max={max_time};  Min={min_time};  Avg={result/count}; Count={count}; Total={result}", out_file)
+    return result
+
+
+def main():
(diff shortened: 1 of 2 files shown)


examples/README.md
```diff
@@ -106,30 +106,30 @@ When you open a notebook, select the kernel `nvflare_example` using the dropdown
...
+|----------------------------------------------------------------------------------------------------------------------------------------|------------------------------------------------------------------------------------------------------------------------------------------------------------------------|------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| [Federated Statistics Overview](./advanced/federated-statistics/README.md)                                                             | -                                                                                                                                                                      | -                      | Discuss the overall federated statistics features.                                                                                                              |
```
YuanTingHsieh, line 118: Should we have this entry inside a table?
I am thinking maybe we just don't have this row?

examples/README.md
```diff
@@ -106,30 +106,30 @@ When you open a notebook, select the kernel `nvflare_example` using the dropdown
...
+| [Federated Statistics for Medical Imaging](./advanced/federated-statistics/image_stats/README.md)                                      | [Image Histograms](./advanced/federated-statistics/image_stats.ipynb)                                                                                                  | see requirements.txt   | Example of gathering local image histogram to compute the global dataset histograms.                                                                            |
```
YuanTingHsieh, line 119: instead of "Federated Statistics for Medical Imaging",
what about just "Federated Statistics for Images"
Since we might use other images there?

examples/README.md
```diff
@@ -106,30 +106,30 @@ When you open a notebook, select the kernel `nvflare_example` using the dropdown
...
+| Example                                                                                                                                | Framework    | Notes                                     | Summary                                                                                                                                                         |
+|----------------------------------------------------------------------------------------------------------------------------------------|--------------|-------------------------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| [NVFlare + MONAI integration](../integration/monai/README.md)                                                                          | MONAI        | -                                         | For an example of using NVIDIA FLARE to train a 3D medical image analysis model using federated averaging (FedAvg) and MONAI Bundle, see [here](../integration/monai/examples/README.md). |
```
YuanTingHsieh, line 111: I am debating with myself here as well,
Should we have this entry here in examples/README?
Since this is already inside integration/monai folder
@chesterxgchen @holgerroth what do you think?

examples/README.md
```diff
@@ -106,30 +106,30 @@ When you open a notebook, select the kernel `nvflare_example` using the dropdown
...
+| [Federated Learning with Differential Privacy for BraTS18 segmentation](./advanced/brats18/README.md)                                  | MONAI        | see requirements.txt                      | Illustrates the use of differential privacy for training brain tumor segmentation models using federated learning.                                              |
+| [Federated Learning for Prostate Segmentation from Multi-source Data](./advanced/prostate/README.md)                                   | MONAI        | see requirements.txt                      | Example of training a multi-institutional prostate segmentation model using [FedAvg](https://arxiv.org/abs/1602.05629), [FedProx](https://arxiv.org/abs/1812.06127), and [Ditto](https://arxiv.org/abs/2012.04221). |
```
YuanTingHsieh, line 113: Do we need the "Notes" column?
it only said "see requirements.txt"
but nothing other than dependencies is there

examples/README.md
```diff
@@ -106,30 +106,30 @@ When you open a notebook, select the kernel `nvflare_example` using the dropdown
...
+| Example                                                                               | Framework     | Notes                     | Summary                                                                                                                                                         |
+|---------------------------------------------------------------------------------------|---------------|---------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------|
+| [NLP-NER](./advanced/nlp-ner/README.md)                                               | see details   | -                         | Illustrates both [BERT](https://github.com/google-research/bert) and [GPT-2](https://github.com/openai/gpt-2) models from [Hugging Face](https://huggingface.co/) ([BERT-base-uncased](https://huggingface.co/bert-base-uncased), [GPT-2](https://huggingface.co/gpt2)) on a Named Entity Recognition (NER) task using the [NCBI disease dataset](https://pubmed.ncbi.nlm.nih.gov/24393765/).  |
```
YuanTingHsieh, line 135: Do we need the "Notes" column?

examples/README.md
```diff
@@ -106,30 +106,30 @@ When you open a notebook, select the kernel `nvflare_example` using the dropdown
...
 |----------------------------------------------------------------------------------------------------------------------------------------|--------------|-----------------------------------------------------|---------------------------|-----------------------------------------------------------------------------------------------------------------------------------------------------------------|
 | [Hello PyTorch with TensorBoard Streaming](./advanced/experiment-tracking/tensorboard-streaming/README.md)                             | PyTorch      | -                                                   | Also requires tensorboard | Example building upon [Hello PyTorch](./hello-world/hello-pt/README.md) showcasing the [TensorBoard](https://tensorflow.org/tensorboard) streaming capability from the clients to the server.  |
```
YuanTingHsieh, line 130: Do we need the "Notes" column?
Do we need the "Notebook" column?

examples/README.md
nvkevlu: Some people had strong opinions we should have things formatted in a table, so we converted everything to table. If we have a good way to use columns to categorize and distinguish how the different entries are different, that would be great, but for some things it was not obvious to me the best way to do that.

examples/README.md
nvkevlu: In general, adding columns in the beginning was to make the table appear more like a table, since two or even three did not feel to me like enough to warrant the format of a table. So to me, the need of these columns is a balance of that with the possibility of us using these columns.

examples/README.md
nvkevlu: this works for me

examples/README.md
YuanTingHsieh: I guess no harm duplicating stuff here.

examples/README.md
YuanTingHsieh: @chesterxgchen can you comment on do we want this here?
I guess it is more of subjective preference, we can ask product or our users.

YuanTingHsieh (approved):

PR #1687, opened 2023-04-13
Support pool stats file creation

yanchengnv:
Fixes # .

### Description

This PR supports pool stats file creation at the end of the job. Two files are created: the **stats_pool_summary.json** contains all the stats pool summary data; the **stats_pool_records.csv** contains raw stats records generated during the job execution.

The two files could be used by researchers for post-job performance analysis. 

The pool_stats_summary.json is always created; whereas the records file is only created when raw records are kept. This is because there could be huge number of raw records that could take big amount of memory. 

To use this feature, the user needs to enable records saving in the meta.json of the job, by adding the following section:

"stats_pool_config": {
    "save_pools": ["request_processing", "request_response"]
  }

The value of "save_pools" is a list of pool names that you want to capture their records. The special name "*" means all pools. 

The stats files are kept in the workspace of the job.

The records file is in CSV format so it is not limited to 2GB. To parse this file, you can use the CsvRecordHandler.read_records() function. It returns a dict like this:
{
   "pool_name1": {
        "category_name1": [
              [timestamp, value],
              [timestamp, value],
              ...
         ],
         "category_name2": [
              [timestamp, value],
              [timestamp, value],
              ...
         ]
   },
   ....
}

For easier maintenance, consolidated stats_utils.py into stats_pool.py.

### Types of changes

- [ ] Documentation updated.


nvflare/private/fed/utils/fed_utils.py
```diff
@@ -214,3 +215,26 @@ def fobs_initialize():
     flare_decomposers.register()
     common_decomposers.register()
     private_decomposers.register()
+
+
+def set_stats_pool_config_for_job(workspace: Workspace, job_id: str):
+    job_meta = get_job_meta_from_workspace(workspace, job_id)
+    stats_config = job_meta.get("stats_pool_config")
```
YuanTingHsieh, line 222: instead of using hard-coded string here "stats_pool_config"

We could add this into ./nvflare/apis/job_def.py "JobMetaKey"

yanchengnv: Yes, we should.

nvflare/fuel/f3/stats_pool.py
```diff
@@ -12,15 +12,22 @@
...
 # limitations under the License.
 
+import csv
+import json
 import threading
 
 from nvflare.fuel.utils.stats_utils import CounterPool, HistPool, new_message_size_pool, new_time_pool
 
 
 class StatsPoolManager:
```
YuanTingHsieh, line 22: A general question, since this file solely depends on `nvflare.fuel.utils.stats_utils`.
And does not need things from f3, should we move this file into nvflare/fuel/utils?

And since this file is about "StatsPoolManager" but not "StatsPool" should we rename it to "stats_pool_manager.py"?

yanchengnv: Maybe. I do feel that the stats_pool and stats_utils could be merged into one thing. 
Let's do this later.

nvflare/fuel/f3/stats_pool.py
```diff
@@ -33,17 +40,36 @@ def _check_name(cls, name, scope):
                 return name
         raise ValueError(f"pool '{name}' is already defined")
 
+    @classmethod
+    def set_hist_pool_config(cls, config: dict):
+        if not isinstance(config, dict):
+            raise ValueError(f"config data must be dict but got {type(config)}")
+        for k, v in config.items():
+            cls.hist_pool_config[k.lower()] = v
```
YuanTingHsieh, line 48: One minor thing here, I know we only use this "stats_pool_config" right now for hist_pool only.
But we could potentially use it for other type of pools.

Maybe we change this method name to more general "set_stats_pool_config"?
and the class variable rename from hist_pool_config to stats_pool_config

nvflare/fuel/f3/stats_pool.py
```diff
@@ -33,17 +40,36 @@ def _check_name(cls, name, scope):
...
+
+    @classmethod
+    def _keep_hist_records(cls, name):
+        name = name.lower()
+        save_pools_list = cls.hist_pool_config.get("save_pools", None)
```
YuanTingHsieh, line 53: we might consider to make a constant variable for this "save_pools" string.

nvflare/fuel/f3/stats_pool.py
```diff
@@ -117,3 +155,71 @@ def from_dict(cls, d: dict):
...
+
+        json_object = json.dumps(recs_dict, indent=4)
+        if not json_object:
+            return
+        with open(csv_file_name, "w") as f:
+            writer = csv.writer(f)
+
+            # num_pools
+            for pool_name, cat_recs in recs_dict.items():
+                assert isinstance(cat_recs, dict)
```
YuanTingHsieh, line 180: Are we dumping records in "recs_dict" twice?
Once in json and once in csv?

nvflare/fuel/f3/stats_pool.py
```diff
@@ -117,3 +157,71 @@ def from_dict(cls, d: dict):
...
+
+    @classmethod
+    def dump_records(cls, csv_file_name: str):
+        recs_dict = cls.get_records()
+        if not recs_dict:
+            return
+
+        json_object = json.dumps(recs_dict, indent=4)
+        if not json_object:
+            return
```
YuanTingHsieh, line 176: Do we need these lines here?

YuanTingHsieh (commented):
Overall logic LGTM, minor comments added.

nvflare/fuel/f3/stats_pool.py
```diff
@@ -117,3 +157,68 @@ def from_dict(cls, d: dict):
...
+    @classmethod
+    def dump_summary(cls, file_name: str):
+        stats_dict = cls.to_dict()
+        json_string = json.dumps(stats_dict, indent=4)
+        with open(file_name, "w") as f:
+            f.write(json_string)
+
+    @classmethod
+    def dump_records(cls, csv_file_name: str):
+        recs_dict = cls.get_records()
```
yhwen, line 170: It makes sense to dump_summary at the end of the RUN. But for dump_records(), should we dump / write the record every time when we get a record, instead of appending / accumulating them, then do a huge dump at the end?

yhwen (commented):
LGTM

PR #1691, opened 2023-04-16
Fix SKLearn executor

YuanTingHsieh:
### Description

- Follows PR https://github.com/NVIDIA/NVFlare/pull/1688 
- Fixed learner init method
- I will write another PR to unify this SKLearnExecutor with LearnerExecutor

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/app_opt/sklearn/sklearn_executor.py
+++ b/nvflare/app_opt/sklearn/sklearn_executor.py
@@ -23,67 +23,67 @@ from nvflare.apis.fl_constant import FLContextKey, ReturnCode
 from nvflare.apis.fl_context import FLContext
 from nvflare.apis.shareable import Shareable, make_reply
 from nvflare.apis.signal import Signal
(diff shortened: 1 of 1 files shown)


nvflare/app_opt/sklearn/sklearn_executor.py
```diff
@@ -23,44 +23,98 @@
...
-        self.writer = None
         self.learner_id = learner_id
+        self.learner = None
+        self.train_task = train_task
         self.local_model_path = None
         self.global_model_path = None
-        self.learner = None
+        self.client_id = None
+        self.writer = None
         self.fl_ctx = None
```
holgerroth, line 55: Do we need a self.fl_ctx? It seems like this could be easily out of sync with the actual context.

YuanTingHsieh: Very good point, I will address this in the next refactoring PR

ZiyueXu77 (approved):

PR #1694, opened 2023-04-17
Make CsvRecordHandler thread safe

yanchengnv:
Fixes # .

### Description

CSV's writerow method is not thread safe. When it is called from different threads at the same time, its behavior is unpredictable. This PR put the call into a lock clause to ensure thread safety. Also added sanity check of inputs.

To make it easy to read stats record file (csv format), added CsvRecordReader.
Added the time_comm tool to compute total communication time of a job. This tool is implemented with the use of CsvRecordReader.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/fuel/f3/stats_pool.py
+++ b/nvflare/fuel/f3/stats_pool.py
@@ -519,25 +519,22 @@ class CsvRecordHandler(RecordWriter):
     @staticmethod
     def read_records(csv_file_name: str):
         pools = {}
-        with open(csv_file_name) as f:
-            reader = csv.reader(f)
-            for row in reader:
-                if len(row) < 4:
-                    raise ValueError(f"'{csv_file_name}' is not a valid stats pool record file: row too short")
-                pool_name = row[0]
-                cat_name = row[1]
-                report_time = float(row[2])
-                value = float(row[3])
+        reader = CsvRecordReader(csv_file_name)
+        for rec in reader:
+            pool_name = rec.pool_name
+            cat_name = rec.category
+            report_time = rec.report_time
+            value = rec.value
 
-                cats = pools.get(pool_name)
-                if not cats:
-                    cats = {}
-                    pools[pool_name] = cats
-                recs = cats.get(cat_name)
-                if not recs:
-                    recs = []
-                    cats[cat_name] = recs
-                recs.append((report_time, value))
+            cats = pools.get(pool_name)
+            if not cats:
+                cats = {}
+                pools[pool_name] = cats
+            recs = cats.get(cat_name)
+            if not recs:
+                recs = []
+                cats[cat_name] = recs
+            recs.append((report_time, value))
         return pools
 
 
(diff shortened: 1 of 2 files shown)


nvflare/fuel/f3/stats_pool.py
```diff
@@ -501,34 +501,66 @@ class CsvRecordHandler(RecordWriter):
...
     def write(self, pool_name: str, category: str, value: float, report_time: float):
+        if not pool_name.isascii():
+            raise ValueError(f"pool_name {pool_name} contains non-ascii chars")
+        if not category.isascii():
+            raise ValueError(f"category {category} contains non-ascii chars")
+
         row = [pool_name, category, report_time, value]
-        self.writer.writerow(row)
+        with self.lock:
+            self.writer.writerow(row)
```
yhwen, line 514: What will happen if reading the csv file and write at the same time, since lock only applies to the write()? 
Probably it will be okay. Just wondering, is it worth to explore using the logger to write to the csv file? Logger should handle the thread reading / writing internally already.

yanchengnv: We don't have this use case. The writing happens during job execution. Reading happens after the job is done for post analysis. We can't access the csv file before the job is done since the file isn't even closed, so the text could be incomplete.

yanchengnv: I don't think we should use logger to write csv files. Even though CSV stands for "comma separated value", the actual separator doesn't have to be commas. Python's csv library can handle all cases, including escaping when value itself contains separators.

holgerroth (approved):
Looks good.

nvflare/fuel/f3/stats_pool.py
```diff
@@ -95,9 +97,15 @@ def from_dict(d: dict):
         b.count = d.get(_KEY_COUNT, 0)
         b.total = d.get(_KEY_TOTAL, 0)
         m = d.get(_KEY_MIN)
-        b.min = m if m else None
+        if isinstance(m, str):
+            b.min = None
+        else:
+            b.min = m
```
YuanTingHsieh, line 103: If "m" is a string, then we set b.min = None?
If not a string, then b.min set to m.

Question here, what type do we support for "m", any type?
If min = "5" (a string) then we just set b.min = None?
Or should we raise Exception in that case?

nvflare/fuel/f3/stats_pool.py
```diff
@@ -95,9 +97,15 @@ def from_dict(d: dict):
...
         x = d.get(_KEY_MAX)
-        b.min = x if x else None
+        if isinstance(x, str):
+            b.max = None
+        else:
+            b.max = x
```
YuanTingHsieh, line 108: same question here as the min value

nvflare/fuel/f3/stats_pool.py
yanchengnv: No, if you look at the code of to_dict, the only possible types are str (representing None) or float.

YuanTingHsieh: got it, then we are assuming, the dict pass in here is a dict generated from "to_dict" method.

yhwen (approved):

PR #1695, opened 2023-04-17
Passing the PYTHONPATH to the child process.

yhwen:
Fixes # .

When using sys.path.append or sys.path.insert doesn't seem to work with the simulator. I was able to reproduce the same issue also using just a python script, i.e. running python test.py. I cannot import the required modules inside the simulator run. However using export PYTHONPATH before running the script works fine. Interestingly, importing the module outside the SimulatorRunner works fine. The sys.path is not correctly passed on to subprocesses created by the simulator.

### Description

Pass the PYTHONPATH to the child process for SImulatorRunner.

### Types of changes

- [ ] Documentation updated.


nvflare/private/fed/app/simulator/simulator_runner.py
```diff
@@ -532,7 +532,9 @@ def do_one_task(self, client, num_of_threads, gpu, lock, timeout=60.0):
         )
         if gpu:
             command += " --gpu " + str(gpu)
-        _ = subprocess.Popen(shlex.split(command, True), preexec_fn=os.setsid, env=os.environ.copy())
+        new_env = os.environ.copy()
+        new_env["PYTHONPATH"] = ":".join(sys.path)[1:]
```
holgerroth, line 536: Is this linux specific? Could it be done so different platforms are supported?

yhwen: Changed to use os.pathsep to work platform independent.

yanchengnv: Don't understand what problem this PR tries to solve.
Why should we explicitly add everything on sys.path to PYTHONPATH?

nvflare/private/fed/app/simulator/simulator_runner.py
```diff
@@ -532,7 +532,10 @@ def do_one_task(self, client, num_of_threads, gpu, lock, timeout=60.0):
         )
         if gpu:
             command += " --gpu " + str(gpu)
-        _ = subprocess.Popen(shlex.split(command, True), preexec_fn=os.setsid, env=os.environ.copy())
+        new_env = os.environ.copy()
+        new_env["PYTHONPATH"] = os.pathsep.join(sys.path)[1:]
+        os.
```
yanchengnv, line 537: This is syntax error.

yanchengnv (changes requested):
There is syntax error.

nvflare/private/fed/app/simulator/simulator_runner.py
yhwen: fixed the typo.

nvflare/private/fed/app/simulator/simulator_runner.py
yhwen: Holger and one customer reported an issue running Simulator. 

When using sys.path.append or sys.path.insert doesn't seem to work with the simulator. I was able to reproduce the same issue also using just a python script, i.e. running python test.py. I cannot import the required modules inside the simulator run. However using export PYTHONPATH before running the script works fine. Interestingly, importing the module outside the SimulatorRunner works fine. The sys.path is not correctly passed on to subprocesses created by the simulator.

nvflare/private/fed/app/simulator/simulator_runner.py
```diff
@@ -532,7 +532,9 @@ def do_one_task(self, client, num_of_threads, gpu, lock, timeout=60.0):
         )
         if gpu:
             command += " --gpu " + str(gpu)
-        _ = subprocess.Popen(shlex.split(command, True), preexec_fn=os.setsid, env=os.environ.copy())
+        new_env = os.environ.copy()
+        new_env["PYTHONPATH"] = os.pathsep.join(sys.path)[1:]
```
holgerroth, line 536: Why do we need to index with `[1:]`?

holgerroth (approved):
Looks good. It works when tested in my environment.

holgerroth (approved):

nvflare/private/fed/app/simulator/simulator_runner.py
```diff
@@ -532,7 +532,12 @@ def do_one_task(self, client, num_of_threads, gpu, lock, timeout=60.0):
         )
         if gpu:
             command += " --gpu " + str(gpu)
-        _ = subprocess.Popen(shlex.split(command, True), preexec_fn=os.setsid, env=os.environ.copy())
+        new_env = os.environ.copy()
+        if not sys.path[0]:
+            new_env["PYTHONPATH"] = os.pathsep.join(sys.path[1:])
```
holgerroth, line 537: Shouldn't is be `os.pathsep.join(sys.path)[1:]`

holgerroth (approved):

PR #1702, opened 2023-04-19
Fix SAG docstring

YuanTingHsieh:
### Description

- Enhance the SAG's docstring

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/apis/impl/bcast_manager.py
+++ b/nvflare/apis/impl/bcast_manager.py
@@ -61,10 +61,6 @@ class BcastTaskManager(TaskManager):
             else:
                 clients_responded += 1
 
-        if clients_responded >= len(task.targets):
-            # all clients have responded!
-            return True, TaskCompletionStatus.OK
-
         # if min_responses is 0, need to have all client tasks responded
         if task.props[_KEY_MIN_RESPS] == 0 and clients_not_responded > 0:
             return False, TaskCompletionStatus.IGNORED
@@ -79,6 +75,10 @@ class BcastTaskManager(TaskManager):
         if min_resps_received_time is None:
             min_resps_received_time = time.time()
             task.props[_KEY_MIN_RESPS_RCV_TIME] = min_resps_received_time
+
+        if clients_responded >= len(task.targets):
+            # all clients have responded!
+            return True, TaskCompletionStatus.OK
 
         # see whether we have waited for long enough
         if time.time() - min_resps_received_time >= task.props[_KEY_WAIT_TIME_AFTER_MIN_RESPS]:
--- a/nvflare/apis/impl/bcast_manager.py
+++ b/nvflare/apis/impl/bcast_manager.py
@@ -61,6 +61,10 @@ class BcastTaskManager(TaskManager):
             else:
                 clients_responded += 1
 
+        if clients_responded >= len(task.targets):
+            # all clients have responded!
+            return True, TaskCompletionStatus.OK
+
         # if min_responses is 0, need to have all client tasks responded
         if task.props[_KEY_MIN_RESPS] == 0 and clients_not_responded > 0:
             return False, TaskCompletionStatus.IGNORED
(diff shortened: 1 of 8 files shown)


nvflare/private/fed/app/simulator/simulator_runner.py
```diff
@@ -207,6 +207,12 @@ def setup(self):
                 self.logger.error("Please provide the number of threads or provide gpu options to run the simulator.")
                 return False
 
+            if meta.get(JobMetaKey.MIN_CLIENTS, 0) > len(self.client_names):
+                self.logger.error(
+                    "The number of simulating clients can not be smaller than the minimum number of clients specified in job's meta."
+                )
+                return False
```
yhwen, line 214: I am not sure if we should enforce this. The main purpose of the Simulator is to help the researcher develop / debug the FL application. It does not really has the NVFlare job concept, or job scheduling. To enforce the min_clients requirement to run the simulator seems to be too strict.

yanchengnv: Right. We should not enforce the job deployment criteria here.

nvflare/apis/impl/bcast_manager.py
```diff
@@ -80,6 +76,10 @@ def check_task_exit(self, task: Task) -> Tuple[bool, TaskCompletionStatus]:
             min_resps_received_time = time.time()
             task.props[_KEY_MIN_RESPS_RCV_TIME] = min_resps_received_time
 
+        if clients_responded >= len(task.targets):
```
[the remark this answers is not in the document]
holgerroth: Otherwise, broadcast_and_wait continues because all connected clients have already responded. So, the min_clients arg in SaG is ignored.

holgerroth:
I tested your PR with monai examples, and it is working now as expected.

nvflare/apis/impl/bcast_manager.py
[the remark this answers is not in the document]
yanchengnv: This chunk of code already exists, but in the front. Why do we need to move to here after two other condition checks? So if all clients already responded, the condition is detected by this chunk of code, up front.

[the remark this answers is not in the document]
holgerroth: Oh. I see. I thought we still supported clients joining later on as in the earlier version. If not, it makes sense just to continue if all clients responded.

nvflare/app_common/workflows/scatter_and_gather.py
```diff
@@ -41,10 +41,10 @@ def _check_non_neg_int(data: Any, name: str):
 class ScatterAndGather(Controller):
     def __init__(
         self,
-        min_clients: int = 1,
+        min_clients: int = 1000,
         num_rounds: int = 5,
         start_round: int = 0,
-        wait_time_after_min_received: int = 10,
+        wait_time_after_min_received: int = 0,
```
holgerroth, line 47: Should we be changing this default value?

YuanTingHsieh: I think better the default is 0, so default it will not wait.

YuanTingHsieh: or you like 10 better?

holgerroth: The idea here is that even when setting `min_clients < len(targets)`, some clients can still contribute to the global model if they contribute within `wait_time_after_min_received`. It's a useful feature in real-world FL deployment. It's also about backward compatibility at this stage. Should we change the default behavior here? Either way, I don't have a strong feeling about this. @yanchengnv to comment.

yanchengnv: I'm not sure either. But it's good to keep default the original value, since there might be people that are relying on this.

PR #1705, opened 2023-04-19
Protect learner against unsafe jobs

yanchengnv:
Fixes # .

### Description

This PR implements learner protection feature requested by Roche.

The Learner API is heavily used to implement FL deep-learning applications. Filters are used to detect bad task data from the server and bad data generate by the learner. Such filters have been working well for Roche as a way of data privacy protection. 

However, a Learner has 3 other methods: initialize, finalize, and abort.  These methods are called outside of the filter chains, and hence they could still escape from filter's detections.

This PR implements a solution, as follows:

- Lazy initialize. The initialize method is not called until a task is received and filtered. 
- Unsafe Job Detection. Once an unsafe job is detected in a filter or by the learner itself, it will raise the UnsafeJobError exception
- The ClientRunner catches this exception and marks the executor to be unsafe
- Once the executor is marked unsafe, finalize and abort methods will no longer be called.

### Types of changes

- [ ] Documentation updated.


nvflare/private/fed/client/client_runner.py
```diff
@@ -193,10 +204,21 @@ def _process_task(self, task: TaskAssignment, fl_ctx: FLContext) -> Shareable:
         if filter_list:
```
YuanTingHsieh, line 204: I see our filters code (pre VS post) have a lot of duplicates, we should re-factor them in next PR.

nvflare/private/fed/client/client_runner.py
```diff
@@ -193,10 +204,21 @@ def _process_task(self, task: TaskAssignment, fl_ctx: FLContext) -> Shareable:
...
             task_data = task.data
             for f in filter_list:
+                filter_name = f.__class__.__name__
                 try:
                     task_data = f.process(task_data, fl_ctx)
+                except UnsafeJobError:
+                    self.log_exception(fl_ctx, f"UnsafeJobError from Task Data Filter {filter_name}")
```
YuanTingHsieh, line 211: I think it might be good to have the reason print out:

```
except UnsafeJobError as e:
    self.log_exception(fl_ctx, f"UnsafeJobError ({secure_format_exception(e)}) from Task Data Filter {filter_name}")
```

nvflare/private/fed/client/client_runner.py
```diff
@@ -305,10 +336,20 @@ def _process_task(self, task: TaskAssignment, fl_ctx: FLContext) -> Shareable:
 
         if filter_list:
             for f in filter_list:
+                filter_name = f.__class__.__name__
                 try:
                     reply = f.process(reply, fl_ctx)
+                except UnsafeJobError:
+                    self.log_exception(fl_ctx, f"UnsafeJobError from Task Result Filter {filter_name}")
```
YuanTingHsieh, line 343: I think it might be good to have the reason print out:

```
except UnsafeJobError as e:
    self.log_exception(fl_ctx, f"UnsafeJobError ({secure_format_exception(e)}) from Task Data Filter {filter_name}")
```

nvflare/private/fed/client/client_runner.py
```diff
@@ -278,8 +300,17 @@ def _process_task(self, task: TaskAssignment, fl_ctx: FLContext) -> Shareable:
                 fl_ctx=fl_ctx,
                 msg=f"submit result: {ReturnCode.EXECUTION_RESULT_ERROR}",
             )
+        except UnsafeJobError:
+            self.log_exception(fl_ctx, f"UnsafeJobError from executor {executor_name}")
+            executor.unsafe = True
```
YuanTingHsieh, line 305: I think it might be good to have the reason print out:

```
except UnsafeJobError as e:
    self.log_exception(fl_ctx, f"UnsafeJobError ({secure_format_exception(e)}) from executor {executor_name}")
```

yhwen, line 305: Who / which function is raising the UnsafeJobError? The "unsafe" flag should be set whenever this UnsafeJobError exception is raised, instead of setting it in the exception catch.

nvflare/app_common/executors/learner_executor.py
```diff
@@ -71,7 +76,9 @@ def initialize(self, fl_ctx: FLContext):
 
     def execute(self, task_name: str, shareable: Shareable, fl_ctx: FLContext, abort_signal: Signal) -> Shareable:
         self.log_info(fl_ctx, f"Client trainer got task: {task_name}")
-
+        if not self.is_initialized:
+            self.is_initialized = True
+            self.initialize(fl_ctx)
```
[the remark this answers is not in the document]
yanchengnv: If executor fails to initialize, the task's return code will be EXECUTION_ERROR, which will cause the whole job to be aborted.

Right now, since is_unsafe is only set during task execution (which happens after initialize), there is no need to check this flag in the initialize function.

HOWEVER, we probably should make this a general feature, meaning that this flag could be set by any component at any time. Then you are right - we should check this flag in initialize. Yuhong seems to be suggesting too.

In this case, we should keep is_unsafe flag as a global property, not specific to executors - there could be many other components that could benefit from this flag.

nvflare/private/fed/client/client_runner.py
yanchengnv: Good idea!

nvflare/private/fed/client/client_runner.py
yanchengnv: Agreed. See my comments in response to Chester's.

nvflare/app_common/executors/learner_executor.py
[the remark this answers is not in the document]
yanchengnv: Changed to set unsafe_job as a sticky prop in FL Context. When any component (other than filters and executors) detects the job is unsafe, it can call fl_ctx.set_job_is_unsafe() to set the prop. 

The client_runner has been modified to end the current job if the unsafe_job prop is set.

nvflare/private/fed/client/client_runner.py
yanchengnv: Implemented.

nvflare/private/fed/client/client_runner.py
yanchengnv: Done.

yanchengnv: Anyone can detect a job is unsafe. The exception UnsafeJobError is mainly used by filters and executor, since this is how our customers are already doing (raise exception to end job).

nvflare/private/fed/client/client_runner.py
yanchengnv: Done

yanchengnv:
These are great points! This PR is not a general solution that solves all issues. This is only intended to solve the Learner pattern requested by customer. We'll need to improve the solution over time.

YuanTingHsieh (approved):
This PR is good for the purpose of protecting against "bad" learner implementation.

For those good points raise by @chesterxgchen we should think and do some other stuff to solve them.

PR #1708, opened 2023-04-20
Fixed a race condition issue for HA when job is about to end_run.

yhwen:
Fixes # .

Fixed the issue Server resumes the job that is nearly completed after switchover but client run have been shutdown

### Description

- Notify the server job process earlier that server has been turned to Cold state, so the client request won't result in END_RUN.
- If the job run snapshot has been removed (completed the run), but the job status has not been updated to FINISHED, those jobs will be updated to FINISHED_ABNORMAL when the server turning to Hot.

### Types of changes

- [ ] Documentation updated.


nvflare/private/fed/server/fed_server.py
```diff
@@ -729,9 +729,7 @@ def _check_server_state(self, overseer_agent):
             self._turn_to_hot()
 
         elif isinstance(self.server_state, Hot2ColdState):
-            self._turn_to_cold()
-
-        self._notify_state_change(old_state_name)
```
YuanTingHsieh, line 734: Why we don't need to notify all the other state changes?

nvflare/private/fed/server/fed_server.py
```diff
@@ -795,6 +794,9 @@ def _turn_to_hot(self):
                                     snapshot=snapshot,
                                     fl_ctx=fl_ctx,
                                 )
+                            restored_job_ids.append(job_id)
+            with self.engine.new_context() as fl_ctx:
+                self.engine.job_runner.update_abnormal_finished_jobs(restored_job_ids, fl_ctx=fl_ctx)
```
YuanTingHsieh, line 799: A question:

1. For the "restored" jobs, do they have same job ID as before?
    - If they do, this logic will mark all those jobs as "FINISHED_ABNORMAL"

YuanTingHsieh: Ok, I found they do have the same ID as before.

But since we try to "restore jobs first", then we mark things that are not restored to be "FINISHED_ABNORMAL"

nvflare/private/fed/server/job_runner.py
```diff
@@ -492,14 +492,26 @@ def restore_running_job(self, run_number: str, job_id: str, job_clients, snapsho
...
+    def update_abnormal_finished_jobs(self, running_job_ids, fl_ctx: FLContext):
+        engine = fl_ctx.get_engine()
+        job_manager = engine.get_component(SystemComponents.JOB_MANAGER)
+        all_jobs = self._get_all_running_jobs(job_manager, fl_ctx)
+
+        for job in all_jobs:
+            if job.job_id not in running_job_ids:
+                try:
+                    job_manager.set_status(job.job_id, RunStatus.FINISHED_ABNORMAL, fl_ctx)
+                    self.logger.info(f"Update the previous running job: {job.job_id} to FINISHED_ABNORMAL.")
```
YuanTingHsieh, line 504: I think we can modify all print that contain status to be

`self.logger.info(f"Update the previous running job: {job.job_id} to {RunStatus.FINISHED_ABNORMAL}.")`

nvflare/private/fed/server/fed_server.py
yhwen: Only when the server is Hot state it has the running job process, and notify to turn to Cold.

YuanTingHsieh: got it, then we could rename it to `_notify_state_change_to_cold` or something, but that can be done in another PR

YuanTingHsieh (approved):
LGTM

PR #1709, opened 2023-04-20
Add cuml backend to SVM example

ZiyueXu77:
Fixes # .

### Description

Add cuml backend to SVM example. This is done by adding an alternative backend option and library import.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/examples/advanced/sklearn-svm/jobs/sklearn_svm_base/app/custom/svm_learner.py
+++ b/examples/advanced/sklearn-svm/jobs/sklearn_svm_base/app/custom/svm_learner.py
@@ -19,7 +19,7 @@ from sklearn.metrics import roc_auc_score
 from nvflare.apis.fl_context import FLContext
 from nvflare.app_common.abstract.learner_spec import Learner
 from nvflare.app_opt.sklearn.data_loader import load_data_for_range
-
+from nvflare.fuel.utils.import_utils import optional_import
 
 class SVMLearner(Learner):
     def __init__(
@@ -33,6 +33,18 @@ class SVMLearner(Learner):
     ):
         super().__init__()
         self.backend = backend
+        if self.backend == "sklearn":
+            self.svm_lib, flag = optional_import(module="sklearn.svm")
+            if not flag:
+                self.log_error(fl_ctx, "Can't import sklearn.svm")
+                return
+        elif self.backend == "cuml":
+            self.svm_lib, flag = optional_import(module="cuml.svm")
+            if not flag:
+                self.log_error(fl_ctx, "Can't import cuml.svm")
+                return
+        else:
+            self.system_panic(f"backend SVM library {self.backend} unknown!", fl_ctx)
 
         self.data_path = data_path
         self.train_start = train_start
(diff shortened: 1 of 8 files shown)


examples/advanced/sklearn-svm/jobs/sklearn_svm_base/app/custom/svm_learner.py
```diff
@@ -14,6 +14,7 @@
 
 from typing import Optional, Tuple
 
+from cuml.svm import SVC as SVC_gpu
```
holgerroth, line 17: Do we want to depend on both cuml and sklearn here? Shouldn't it be an optional import?

ZiyueXu77: right, let me do the import inside

ZiyueXu77: updated

examples/advanced/sklearn-svm/jobs/sklearn_svm_base/app/custom/svm_learner.py
```diff
@@ -64,7 +66,16 @@ def train(self, curr_round: int, global_param: Optional[dict], fl_ctx: FLContext
...
+            if self.backend == "sklearn":
+                from sklearn.svm import SVC
+
+                self.svm = SVC(kernel=self.kernel)
+            elif self.backend == "cuml":
+                from cuml.svm import SVC as SVC_gpu
+
+                self.svm = SVC_gpu(kernel=self.kernel)
+            else:
+                self.system_panic("backend SVM library unknown!", fl_ctx)
```
holgerroth, line 78: f"backend SVM library {self.backend} unknown!"

examples/advanced/sklearn-svm/jobs/sklearn_svm_base/app/custom/svm_learner.py
```diff
@@ -79,7 +90,16 @@ def train(self, curr_round: int, global_param: Optional[dict], fl_ctx: FLContext
     def validate(self, curr_round: int, global_param: Optional[dict], fl_ctx: FLContext) -> Tuple[dict, dict]:
         # local validation with global support vectors
         # fit a standalone SVM with the global support vectors
-        svm_global = SVC(kernel=self.kernel)
+        if self.backend == "sklearn":
+            from sklearn.svm import SVC
```
holgerroth, line 94: It might be better to use the `optional_import` once as [here](https://github.com/NVIDIA/NVFlare/blob/793dcaefe23123c290edca21f400ed03d83c381c/nvflare/app_opt/xgboost/histogram_based/controller.py#L87) instead of importing twice in the same function

ZiyueXu77: updated

examples/advanced/sklearn-svm/utils/prepare_job_config.py
```diff
@@ -187,6 +194,7 @@ def _update_meta(meta: dict, args):
 def _update_client_config(config: dict, args, site_name: str, site_indices):
     # update client config
     # data path and training/validation row indices
+    config["components"][0]["args"]["backend"] = args.backend
```
YuanTingHsieh, line 197: We might be able to enhance this later, instead of the hard-coding "0" index here.
But we can do that in a separate PR.

YuanTingHsieh (approved):
Thanks, the changes based on current code structure LGTM.

In the future when we re-factor Learner/LearnerExecutor we might change some of the method.

examples/advanced/sklearn-svm/utils/prepare_job_config.py
ZiyueXu77: Agree, we can recognize this component with its name, will enhance in a separate PR

holgerroth (approved):

examples/advanced/sklearn-svm/jobs/sklearn_svm_base/app/custom/svm_learner.py
YuanTingHsieh: updated code LGTM.

PR #1713, opened 2023-04-22
provision refactoring

chesterxgchen:
### Description
This PR is pre-requisite for work for POC upgrade task. 
In this PR, we made three code refactoring changes, these changes is intended
1) make better organization of codes so that we can re-use the code 
2) abstract some properties such as server name, or subject in certification into methods, such change make it possible to overwrite the method if we like to change server_name or overwrite the subject 

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/lighter/impl/cert.py
+++ b/nvflare/lighter/impl/cert.py
@@ -79,7 +79,7 @@ class CertBuilder(Builder):
             self.persistent_state["root_pri_key"] = serialize_pri_key(self.pri_key).decode("ascii")
 
     def _build_write_cert_pair(self, participant, base_name, ctx):
-        subject = participant.subject
+        subject = self.get_subject(participant)
         if self.persistent_state and subject in self.persistent_state:
             cert = x509.load_pem_x509_certificate(
                 self.persistent_state[subject]["cert"].encode("ascii"), default_backend()
@@ -125,7 +125,7 @@ class CertBuilder(Builder):
 
     def get_pri_key_cert(self, participant):
         pri_key, pub_key = self._generate_keys()
-        subject = participant.subject
+        subject = self.get_subject(participant)
         subject_org = participant.org
         if participant.type == "admin":
             role = participant.props.get("role")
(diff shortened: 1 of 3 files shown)


nvflare/lighter/provision.py
```diff
@@ -94,55 +95,81 @@ def handle_provision(args):
...
-            f"Configuration error: Expect 2 or 1 server to be provisioned.  {project_full_path} contains {n_servers} servers."
-        )
-        return
-
-    builders = list()
-    for b in project_dict.get("builders"):
-        path = b.get("path")
-        args = b.get("args")
-        builders.append(instantiate_class(path, args))
+        raise (f"Configuration error: Expect 2 or 1 server to be provisioned. project contains {n_servers} servers.")
```
YuanTingHsieh, line 147: should raise an Exception?

nvflare/lighter/impl/static_file.py
```diff
@@ -127,9 +130,9 @@ def _build_server(self, server, ctx):
         ctx["admin_port"] = admin_port
         fed_learn_port = server.props.get("fed_learn_port", 8002)
         ctx["fed_learn_port"] = fed_learn_port
-        ctx["server_name"] = server.name
-        server_0["service"]["target"] = f"{server.name}:{fed_learn_port}"
-        server_0["admin_host"] = server.name
+        ctx["server_name"] = self.get_server_name(server)
+        server_0["service"]["target"] = f"{self.get_server_name(server)}:{fed_learn_port}"
+        server_0["admin_host"] = self.get_server_name(server)
```
YuanTingHsieh, line 135: instead of wrapping "server.name" into a method,

suggest just adding type hint `server: Participant`

nvflare/lighter/impl/cert.py
```diff
@@ -125,7 +125,7 @@ def build(self, project, ctx):
 
     def get_pri_key_cert(self, participant):
         pri_key, pub_key = self._generate_keys()
-        subject = participant.subject
+        subject = self.get_subject(participant)
```
YuanTingHsieh, line 128: instead of wrapping "participant.subject" into a method,

suggest just adding type hint `participant: Participant`

nvflare/lighter/impl/cert.py
```diff
@@ -79,7 +79,7 @@ def _build_root(self, subject, subject_org):
             self.persistent_state["root_pri_key"] = serialize_pri_key(self.pri_key).decode("ascii")
 
     def _build_write_cert_pair(self, participant, base_name, ctx):
-        subject = participant.subject
+        subject = self.get_subject(participant)
```
YuanTingHsieh, line 82: instead of wrapping "participant.subject" into a method,

suggest just adding type hint `participant: Participant`

IsaacYangSLA (approved):
Thanks.  LGTM.

PR #1717, opened 2023-04-26
Add scheme in fed_server.json and fed_client.json

IsaacYangSLA:
### Description

Add scheme option in project.yml to support different communication driver.  Currently supporting grpc only.
Add codes in dashboard to use project-wide scheme.  Currently, default to grpc only.
Prepare dashboard model to include scheme.

### Types of changes

- [ ] Documentation updated.


IsaacYangSLA:
Added the purposes of this PR.  Also, changed the wording in the comment about the available option of scheme in project.yml so users will not set scheme to unsupported protocol.

yanchengnv:
Since we haven't sorted out the stability/usability issues of non-grpc drivers, we need to hold off this PR for now.

PR #1722, opened 2023-05-02
Api parity

yanchengnv:
Fixes # .

### Description

This PR adds more functions to FLARE API to make it consistent with the old FLAdminAPI. For detailed changes, see https://docs.google.com/document/d/1RVcBWZnDG_bRYzFuw0KhlHcnxqg6-c2SSdvQ0tq8eS4/edit#heading=h.21o8y0jdhltm

This PR also fixed a few minor issues:
- Changed to use predefined constants to replace hard-coded string values
- When processing client replies, add client_token to ClientReply. This is to avoid the call to engine.get_client_name_by_token, which could return empty string under certain race condition (the client just disconnected before the reply is processed).
- Changed the set_timeout to become a local command and new timeout value is only effective for the current API session.
- Added job ID creation and validation function based on UUID4. This is necessary to fix abort_task bug where it is impossible to differentiate job ID from site name.

### Types of changes

- [ ] Documentation updated.


nvflare/apis/job_def.py
```diff
@@ -189,3 +189,21 @@ def job_from_meta(meta: dict) -> Job:
         required_sites=meta.get(JobMetaKey.MANDATORY_CLIENTS, []),
     )
     return job
+
+
+def new_job_id() -> str:
+    return str(uuid.uuid4())
+
+
+def is_valid_job_id(jid: str) -> bool:
```
YuanTingHsieh, line 198: Since we are adding this new method "is_valid_job_id", can we also add unit tests for this?

Put a test inside "tests/unit_test/apis/job_def_test.py"

Could be:

```
def test_is_valid_job_id():
    assert is_valid_job_id("site-1") == False
    assert is_valid_job_id("a5464dr") == True
```

nvflare/apis/job_def.py
```diff
@@ -189,3 +189,21 @@ def job_from_meta(meta: dict) -> Job:
...
+        return False
+
+    try:
+        val = uuid.UUID(jid, version=4)
+    except ValueError:
+        return False
+
+    # If the jid string is a valid hex code, but an invalid uuid4,the UUID.__init__ will convert it to a
+    # valid uuid4. This is bad for validation purposes.
+    return val.hex == jid.replace("-", "")
```
YuanTingHsieh, line 209: maybe we can add an example jid string that "is a valid hex code, but an invalid uuid4" in test

nvflare/fuel/flare_api/api_spec.py
```diff
@@ -205,6 +227,256 @@ def delete_job(self, job_id: str):
 
     @abstractmethod
     def get_system_info(self) -> SystemInfo:
+        """Get general info of the FLARE system"""
+        pass
+
+    def get_client_job_status(self, client_names: List[str] = None) -> List[dict]:
```
YuanTingHsieh, line 233: Since this is a "spec", or "abstract class".
It might be better if we add `@abstractmethod` to these method definitions
SO we enforced whoever inherits this class need to implement these methods.

nvflare/fuel/flare_api/api_spec.py
```diff
@@ -205,6 +227,256 @@ def delete_job(self, job_id: str):
...
+            client_names: names of the clients to get status info
+
+        Returns: A list of jobs running on the clients. Each job is described by a dict of: id, app name and status.
+        If there are multiple jobs running on one client, the list contains one entry for each job for that client.
+        If no FL clients are connected or the server failed to communicate to them, this method returns None.
+
+        """
+        pass
+
+    def restart(self, target_type: str, client_names: Optional[List[str]] = None) -> dict:
```
YuanTingHsieh, line 246: Since this is a "spec", or "abstract class".
It might be better if we add `@abstractmethod` to these method definitions
SO we enforced whoever inherits this class need to implement these methods.

nvflare/fuel/hci/client/api.py
```diff
@@ -455,6 +456,19 @@ def __init__(
         self.service_finder.start(self._handle_sp_address_change)
         self._start_session_monitor()
 
+    def set_command_timeout(self, timeout: float):
+        if not isinstance(timeout, float):
```
YuanTingHsieh, line 460: note that, this means

set_command_timeout(0) will failed since

```
>>> isinstance(0, float)
False

```

nvflare/fuel/hci/cmd_arg_utils.py
```diff
@@ -68,3 +71,82 @@ def get_usage(self) -> str:
...
+def process_targets_into_str(targets: List[str]) -> str:
+    if not isinstance(targets, list):
+        raise SyntaxError("targets is not a list.")
+    if not all(isinstance(t, str) for t in targets):
+        raise SyntaxError("all targets in the list of targets must be strings.")
+    for t in targets:
+        try:
+            validate_required_target_string(t)
+        except SyntaxError:
+            raise SyntaxError("each target in targets must be a string of only valid characters and no spaces.")
```
YuanTingHsieh, line 85: suggest include what target failed and include the error message raised from validate_required_target_string:

```
except SyntaxError as e:
    raise SyntaxError(f"target: {t} is invalid: {e}")
```

nvflare/fuel/hci/cmd_arg_utils.py
```diff
@@ -68,3 +71,82 @@ def get_usage(self) -> str:
...
+
+
+def validate_path_string(path: str) -> str:
+    """Returns the path string if it is valid."""
+    if not isinstance(path, str):
+        raise SyntaxError("path is not str.")
+    if not re.match("^[A-Za-z0-9-._/]*$", path):
+        raise SyntaxError("unsupported characters in path {}".format(path))
+    if path.startswith("/"):
+        raise SyntaxError("absolute path is not allowed")
```
YuanTingHsieh, line 116: suggest use built-in method:

`os.path.isabs(path)`

So potentially we could support more platforms in the future

nvflare/fuel/hci/cmd_arg_utils.py
```diff
@@ -68,3 +71,82 @@ def get_usage(self) -> str:
...
+    if not isinstance(file, str):
+        raise SyntaxError("file is not str.")
+    if not re.match("^[A-Za-z0-9-._/]*$", file):
+        raise SyntaxError("unsupported characters in file {}".format(file))
+    if file.startswith("/"):
+        raise SyntaxError("absolute path for file is not allowed")
+    paths = file.split("/")
+    for p in paths:
+        if p == "..":
+            raise SyntaxError(".. in file path is not allowed")
```
YuanTingHsieh, line 135: since this part duplicated with "validate_path_string" can we reuse that method here?

nvflare/private/fed/server/cmd_utils.py
```diff
@@ -134,11 +140,9 @@ def validate_command_targets(self, conn: Connection, args: List[str]) -> str:
         return ""
 
     def must_be_project_admin(self, conn: Connection, args: List[str]):
-        if args[0] == AdminCommandNames.ADMIN_CHECK_STATUS and len(args) == 1:
-            args.extend(["server"])
         role = conn.get_prop(ConnProps.USER_ROLE, "")
-        if role != "project_admin":
-            conn.append_error("Not authorized as project_admin.", meta=make_meta(MetaStatusValue.NOT_AUTHORIZED))
+        if role not in ["project_admin", "super"]:
```
YuanTingHsieh, line 144: why do we need to introduce another super user "super"?

nvflare/fuel/hci/proto.py
```diff
@@ -107,27 +118,29 @@ def append_table(self, headers: List[str], name=None) -> Table:
         self.data.append({ProtoKey.TYPE: ProtoKey.TABLE, ProtoKey.ROWS: t.rows})
         return t
 
-    def append_string(self, data: str, meta: dict = None):
-        self.data.append({ProtoKey.TYPE: ProtoKey.STRING, ProtoKey.DATA: data})
+    def add_meta(self, meta: dict):
```
YuanTingHsieh, line 121: minor personal preference: "update_meta"

nvflare/private/fed/server/cmd_utils.py
```diff
@@ -214,15 +220,19 @@ def _process_replies_to_string(self, conn: Connection, replies) -> str:
...
+                client_name = r.client_name
                 response += "client:" + client_name
                 if r.reply:
                     response += " : " + r.reply.body + "\n"
+                    client_replies[client_name] = r.reply.body
                 else:
                     response += " : No replies\n"
+                    client_replies[client_name] = MetaStatusValue.NO_REPLY
+
+        conn.add_meta({MetaKey.CLIENT_STATUS: client_replies})
```
YuanTingHsieh, line 237: Most of the case we see:
```
meta = {MetaKey.STATUS: MetaStatusValue.xxx}
```
In here we specially see:
```
meta = {MetaKey.CLIENT_STATUS: {client1: MetaStatusValue.xxx, client2: MetaStatusValue.xxx}}
```

I am not sure if we want to mix the usage here.

nvflare/ha/ha_admin_cmds.py
```diff
@@ -70,60 +70,73 @@ def list_sp(self, args, ctx: CommandContext):
...
+        return {ProtoKey.STATUS: APIStatus.SUCCESS, ProtoKey.DETAILS: str(sp), ProtoKey.META: sp.__dict__}
 
     def promote_sp(self, args, ctx: CommandContext):
         overseer_agent = self.overseer_agent
         if len(args) != 2:
-            return {"status": APIStatus.ERROR_SYNTAX, "details": "usage: promote_sp example1.com:8002:8003"}
+            return {
+                ProtoKey.STATUS: APIStatus.ERROR_SYNTAX,
+                ProtoKey.DETAILS: "usage: promote_sp example1.com:8002:8003",
+                ProtoKey.META: {MetaKey.STATUS: MetaStatusValue.SYNTAX_ERROR},
```
YuanTingHsieh, line 89: should we also use "make_meta" method here?

nvflare/ha/ha_admin_cmds.py
```diff
@@ -70,60 +70,73 @@ def list_sp(self, args, ctx: CommandContext):
...
         resp = overseer_agent.promote_sp(sp_end_point)
-        if json.loads(resp.text).get("Error"):
+        err = json.loads(resp.text).get("Error")
+        if err:
             return {
-                "status": APIStatus.ERROR_RUNTIME,
-                "details": "Error: {}".format(json.loads(resp.text).get("Error")),
+                ProtoKey.STATUS: APIStatus.ERROR_RUNTIME,
+                ProtoKey.DETAILS: f"Error: {err}",
+                ProtoKey.META: {MetaKey.STATUS: MetaStatusValue.INTERNAL_ERROR, MetaKey.INFO: err},
```
YuanTingHsieh, line 99: should we also use "make_meta" method here?

nvflare/ha/ha_admin_cmds.py
```diff
@@ -70,60 +70,73 @@ def list_sp(self, args, ctx: CommandContext):
...
+            info = f"Promoted endpoint: {sp_end_point}. Synchronizing with overseer..."
             return {
-                "status": APIStatus.SUCCESS,
-                "details": "Promoted endpoint: {}. Synchronizing with overseer...".format(sp_end_point),
+                ProtoKey.STATUS: APIStatus.SUCCESS,
+                ProtoKey.DETAILS: info,
+                ProtoKey.META: {
+                    MetaKey.STATUS: MetaStatusValue.OK,
+                    MetaKey.INFO: info,
+                },
```
YuanTingHsieh, line 109: should we also use "make_meta" method here?

nvflare/private/fed/server/training_cmds.py
```diff
@@ -201,47 +196,39 @@ def restart(self, conn: Connection, args: List[str]):
...
-                response = self._restart_clients(conn, tokens)
+                response = self._restart_clients(conn)
                 conn.append_string(response)
                 # check with Isaac - no need to wait!
                 # time.sleep(5)
 
             err = engine.restart_server()
             if err:
-                conn.append_error(err)
+                conn.append_error(err, meta={MetaKey.SERVER_STATUS: MetaStatusValue.ERROR, MetaKey.INFO: err})
```
YuanTingHsieh, line 219: should we also use "make_meta" method here?

nvflare/private/fed/server/training_cmds.py
```diff
@@ -201,47 +196,39 @@ def restart(self, conn: Connection, args: List[str]):
...
             else:
-                conn.append_string("Server scheduled for restart")
+                conn.append_string("Server scheduled for restart", meta={MetaKey.SERVER_STATUS: MetaStatusValue.OK})
```
YuanTingHsieh, line 221: should we also use "make_meta" method here?

yanchengnv: No, we can't here since make_meta is used to make the final status meta.

nvflare/private/fed/server/training_cmds.py
yanchengnv: Yes

nvflare/ha/ha_admin_cmds.py
yanchengnv: Yes

nvflare/ha/ha_admin_cmds.py
yanchengnv: Yes

nvflare/ha/ha_admin_cmds.py
yanchengnv: Yes

nvflare/private/fed/server/cmd_utils.py
yanchengnv: Different here. MetaKey.STATUS specifies the overall status of the request; whereas MetaKey.CLIENT_STATUS is a piece of meta information in the response.

nvflare/fuel/hci/proto.py
yanchengnv: update_meta seems to be better.

nvflare/private/fed/server/cmd_utils.py
yanchengnv: Yeah, this was from old days that the admin console was generic and used the role name "super" in non-secure mode. I guess we could change it to "project_admin" since the admin console now is FLARE specific.

nvflare/fuel/hci/cmd_arg_utils.py
yanchengnv: Probably so.

nvflare/fuel/hci/cmd_arg_utils.py
```diff
@@ -68,3 +71,82 @@ def get_usage(self) -> str:
...
+
+def validate_path_string(path: str) -> str:
+    """Returns the path string if it is valid."""
+    if not isinstance(path, str):
+        raise SyntaxError("path is not str.")
+    if not re.match("^[A-Za-z0-9-._/]*$", path):
+        raise SyntaxError("unsupported characters in path {}".format(path))
+    if path.startswith("/"):
+        raise SyntaxError("absolute path is not allowed")
+    paths = path.split("/")
```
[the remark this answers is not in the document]
yanchengnv: Yup

nvflare/fuel/hci/cmd_arg_utils.py
yanchengnv: yup

nvflare/fuel/hci/cmd_arg_utils.py
yanchengnv: right

nvflare/fuel/hci/client/api.py
yanchengnv: maybe we should introduce unset_command_timeout() method.

nvflare/fuel/flare_api/api_spec.py
yanchengnv: yes

nvflare/apis/job_def.py
yanchengnv: good idea

nvflare/fuel/flare_api/api_spec.py
```diff
@@ -61,6 +65,24 @@ class AuthorizationError(BaseException):
...
+class NoClientsAvailable(BaseException):
+    pass
+
+
+class ClientsStillRunning(BaseException):
+    pass
+
+
+class InvalidTarget(BaseException):
+    pass
```
yhwen, line 77: Instead of making each one an individual class, maybe we can combine them into one class, and add a "class type" field to indicate what class it is. In this case, we don't have to add a new class for each new type / scenario we need to process.

nvflare/fuel/hci/cmd_arg_utils.py
```diff
@@ -68,3 +71,82 @@ def get_usage(self) -> str:
         usage_output = buffer.getvalue().split("\n", 1)[1]
         buffer.close()
         return usage_output
+
+
+def process_targets_into_str(targets: List[str]) -> str:
+    if not isinstance(targets, list):
+        raise SyntaxError("targets is not a list.")
+    if not all(isinstance(t, str) for t in targets):
+        raise SyntaxError("all targets in the list of targets must be strings.")
```
yhwen, line 80: We are raising the same SyntaxError here. Instead of call "raise SyntaxError()" multiple places, we can generate the error message from different place, and then call the "raise SyntaxError(error_message)" in one place. This maybe a little cleaner.

nvflare/fuel/hci/cmd_arg_utils.py
```diff
@@ -68,3 +71,82 @@ def get_usage(self) -> str:
...
+
+
+def validate_required_target_string(target: str) -> str:
+    """Returns the target string if it exists and is valid."""
+    if not target:
+        raise SyntaxError("target is required but not specified.")
+    if not isinstance(target, str):
+        raise SyntaxError("target is not str.")
+    if not re.match("^[A-Za-z0-9._-]*$", target):
+        raise SyntaxError("target must be a string of only valid characters and no spaces.")
```
yhwen, line 96: same error_message and "raise SyntaxError(error_message)" as above.

nvflare/fuel/flare_api/api_spec.py
yanchengnv: implemented as suggested

nvflare/apis/job_def.py
yanchengnv: added test case as suggested

nvflare/apis/job_def.py
yanchengnv: done as suggested

nvflare/fuel/flare_api/api_spec.py
yanchengnv: done as suggested

nvflare/fuel/hci/cmd_arg_utils.py
yanchengnv: done as suggested

nvflare/fuel/hci/cmd_arg_utils.py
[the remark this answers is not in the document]
yanchengnv: done as suggested

nvflare/fuel/hci/cmd_arg_utils.py
yanchengnv: done as suggested

nvflare/fuel/hci/proto.py
yanchengnv: renamed to update_meta

nvflare/ha/ha_admin_cmds.py
yanchengnv: done as suggested

nvflare/fuel/flare_api/api_spec.py
yanchengnv: Not sure. The API has been designed to use exception for different error conditions.

nvflare/fuel/hci/cmd_arg_utils.py
yanchengnv: This was from old code (just moved from FLAdminAPI to here to be reused). It's probably not worth the time to do what you suggested.

nvflare/ha/ha_admin_cmds.py
yanchengnv: done as suggested

nvflare/private/fed/server/cmd_utils.py
yanchengnv: changed to use project_admin only.

nvflare/fuel/flare_api/api_spec.py
yanchengnv: Maybe we could use one exception (e.g. FlareApiError) for all cases. This exception accepts a status_code.
It does get tedious to add a new exception type and a new status code every time we have a new error condition.

tests/unit_test/apis/job_def_test.py
```diff
@@ -0,0 +1,25 @@
+# Copyright (c) 2022, NVIDIA CORPORATION.  All rights reserved.
```
YuanTingHsieh, line 1: Since we create this file in 2023, we should put 2023 here instead of 2022.

nvflare/private/fed/server/cmd_utils.py
YuanTingHsieh: got it thanks

nvflare/fuel/hci/cmd_arg_utils.py
YuanTingHsieh: The current code:

```
except SyntaxError:
            raise SyntaxError("each target in targets must be a string of only valid characters and no spaces.")
```

Users can NOT know which target is having issue, as we swallow individual SyntaxError and just raise this overall one.

YuanTingHsieh (commented):
Mostly good now, just 2 more comments.

tests/unit_test/apis/job_def_test.py
yanchengnv: Not sure this is the case. Is copyright thing about when the code was put in?

nvflare/fuel/hci/cmd_arg_utils.py
yanchengnv: This has been fixed already.

tests/unit_test/apis/job_def_test.py
YuanTingHsieh: I read some things online, they say it is when this piece of code is created, our other files are created in 2021 2020 or 2022 which is fine.

I refer to this: https://opensource.google/documentation/reference/copyright#the_year

nvflare/fuel/flare_api/api_spec.py
yanchengnv: Just checked the codebase, unfortunately some of the defined exception types have been used in app code and example notebooks. If we change to use one exception type, all those places would have to change.

So for now let's stay with this.

tests/unit_test/apis/job_def_test.py
yanchengnv: I see. Changed as suggested.

yhwen (approved):

PR #1742, opened 2023-05-16
Include exception type in error logs

taleinat:
### Description

```python
>>> int("a")

>>> try:
	int("a")
except ValueError as exc:
	print(str(exc))

```

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/security/logging.py
+++ b/nvflare/security/logging.py
@@ -130,4 +130,4 @@ def secure_format_exception(e: BaseException) -> str:
     if is_secure():
         return str(type(e))
     else:
-        return str(e)
+        return f"{str(type(e))}: {str(e)}"


taleinat:
>  do you mind to submit another PR for the dev branch ?

Sure.

taleinat:
Upon further consideration, I think it makes sense to also include the exception's traceback if `is_secure()` is false.

We just ran into a case where the traceback would be where the truly useful information was in such an error.

What do you think?

YuanTingHsieh:
@taleinat we have "secure_format_traceback" method here, do you think that serves your need?
Or you suggest have "secure_format_exception" absorb "secure_format_traceback" method?

taleinat:
Thanks for the pointer @YuanTingHsieh! Yes I think `secure_format_traceback()` or `secure_log_traceback()` would be good here. I'll update this PR.

taleinat:
> Thanks for the pointer @YuanTingHsieh! Yes I think `secure_format_traceback()` or `secure_log_traceback()` would be good here.
> 
> I'll create another PR.

Actually, I see that on the dev branch `secure_log_traceback()` has been added for configuration errors, which is where we found it to be missing. And that's already included in v2.3, but I don't see it on the v2.2 branch. So apparently no extra PR is needed.

YuanTingHsieh:
@taleinat can you fix the unit test failure?
I believe it is cause by this one: https://github.com/NVIDIA/NVFlare/blob/dev/tests/unit_test/fuel/utils/component_builder_test.py#L79-L82 

You can run test locally using ./runtest.sh

Thanks

YuanTingHsieh (changes requested):
please fix the unit test thanks!

YuanTingHsieh:
Thanks, Changes move to #1785, I fix the unit test there

taleinat:
Thanks @YuanTingHsieh, I couldn't get to fixing these before today...

PR #1743, opened 2023-05-16
Don't catch BaseException

taleinat:
### Description

`BaseException` includes `SystemExit` and `KeyboardInterrupt`, so catching it can interfere with proper process termination. It also includes `GeneratorExit` which can cause some very strange behaviors.

Catching `BaseException` should only be necessary in highly exceptional cases. Clearly, though, in this codebase it is used wherever `except Exception` would normally be used, i.e. to catch any type of "normal" exception (what a lovely oxymoron!).

I reviewed some of the places where this replacement is done in this PR, but not all of them.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/examples/advanced/cifar10/cifar10-real-world/figs/plot_tensorboard_events.py
+++ b/examples/advanced/cifar10/cifar10-real-world/figs/plot_tensorboard_events.py
@@ -127,7 +127,7 @@ def main():
             for k in xsite_keys:
                 try:
                     xsite_data[k].append(xsite_results["site-1"][k]["val_accuracy"])
-                except BaseException as e:
+                except Exception as e:
                     raise ValueError(f"No val_accuracy for {k} in {xsite_file}!")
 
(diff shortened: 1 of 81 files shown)


yanchengnv:
We do need a complete review on all such cases. For most places, we should use Exception.  There might be very few places that we do need to catch everything, including KeyboardInterrupt. But these should not be the norm.

YuanTingHsieh:
Thanks for the PR!

@yhwen please confirm this change is OK with nvflare/private codes.
@IsaacYangSLA please confirm this change is OK with nvflare/dashboard and nvflare/lighter.

Other parts LGTM.

IsaacYangSLA:
OK on dashboard/overseer/lighter.

nvidianz (approved):

yhwen (approved):

yhwen (approved):

PR #1749, opened 2023-05-18
Include exception type in logged error messages

taleinat:
This is a copy of PR #1742, for the dev branch instead of main.

### Description

```python
>>> int("a")

>>> try:
	int("a")
except ValueError as exc:
	print(str(exc))

```

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh:
@taleinat thanks for the submission, please fix the unit test on another PR and port the changes here thanks!

YuanTingHsieh (changes requested):
please fix the unit test thanks!

PR #1750, opened 2023-05-18
Added GPU group support for simulator.

yhwen:
Fixes # .

### Description

Add support to enable multi-gpu training for Simulator. Provide the GPU group option to the format like: "-gpu [0,1],[1,2],[3]"

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/private/fed/app/simulator/simulator_runner.py
+++ b/nvflare/private/fed/app/simulator/simulator_runner.py
@@ -175,7 +175,11 @@ class SimulatorRunner(FLComponent):
                 return False
 
             if self.args.gpu:
-                gpus = self.args.gpu.split(",")
+                gpus, success = self.split_gpus(self.args.gpu)
+                if not success:
+                    self.logger.error(f"GPUs group list in wrong format. Please use format like: [0,1],[1,2] ...")
+                    return False
+
                 host_gpus = [str(x) for x in (get_host_gpu_ids())]
                 if host_gpus and not set(gpus).issubset(host_gpus):
                     wrong_gpus = [x for x in gpus if x not in host_gpus]
@@ -369,7 +373,7 @@ class SimulatorRunner(FLComponent):
                 #     client.start_heartbeat(interval=2)
 
                 if self.args.gpu:
-                    gpus = self.args.gpu.split(",")
+                    gpus, _ = self.split_gpus(self.args.gpu)
                     split_clients = self.split_clients(self.federated_clients, gpus)
                 else:
                     gpus = [None]
@@ -394,6 +398,29 @@ class SimulatorRunner(FLComponent):
         else:
             run_status = 1
         return run_status
+
+    def split_gpus(self, gpus) -> ([str], bool):
+        gpus = gpus.replace(" ", "")
+        if not (gpus.startswith("[") and gpus.endswith("]")):
+            return None, False
+        result = []
+        index = 0
+        while index < len(gpus):
+            if gpus[index] != "[":
+                return None, False
+            a = ""
+            while True:
+                index += 1
+                if gpus[index] == "]":
+                    break
+                a += gpus[index]
+            result.append(a)
+            index += 1
+            if index < len(gpus) and gpus[index] != ",":
+                return None, False
+            else:
+                index += 1
+        return result, True
 
     def client_run(self, clients, gpu):
         client_runner = SimulatorClientRunner(self.args, clients, self.client_config, self.deploy_args, self.build_ctx)
--- a/tests/unit_test/private/fed/app/simulator/simulator_runner_test.py
+++ b/tests/unit_test/private/fed/app/simulator/simulator_runner_test.py
@@ -86,3 +86,31 @@ class TestSimulatorRunner:
         runner = SimulatorRunner(job_folder="", workspace="")
         split_names = runner.split_clients(client_names, gpus)
         assert sorted(split_names) == sorted(expected_split_names)
+
+    @pytest.mark.parametrize(
+        "gpus, expected_gpus",
+        [
+            ("[0,1],[1, 2]", ["0,1", "1,2"]),
+            ("[0,1],[3]", ["0,1", "3"]),
+            ("[0,1],[ 3 ]", ["0,1", "3"]),
+        ],
+    )
+    def test_split_gpus_success(self, gpus, expected_gpus):
+        runner = SimulatorRunner(job_folder="", workspace="")
+        split_gpus, _ = runner.split_gpus(gpus)
+        assert split_gpus == expected_gpus
+
+    @pytest.mark.parametrize(
+        "gpus",
+        [
+            "[0,1],3",
+            "0,1,2",
+        ],
+    )
+    def test_split_gpus_fail(self, gpus):
+        runner = SimulatorRunner(job_folder="", workspace="")
+        split_gpus, success = runner.split_gpus(gpus)
+        assert split_gpus is None
+        assert success is False
+
+


nvflare/private/fed/app/simulator/simulator_runner.py
```diff
@@ -175,7 +175,11 @@ def setup(self):
                 return False
 
             if self.args.gpu:
-                gpus = self.args.gpu.split(",")
+                gpus, success = self.split_gpus(self.args.gpu)
+                if not success:
+                    self.logger.error(f"GPUs group list in wrong format. Please use format like: [0,1],[1,2] ...")
```
YuanTingHsieh, line 180: need to change to `self.logger.error("GPUs group list in wrong format. Please use format like: [0,1],[1,2] ...")`
otherwise style check will fail

nvflare/private/fed/app/simulator/simulator_runner.py
```diff
@@ -395,6 +399,29 @@ def simulator_run_main(self):
...
+                    break
+                a += gpus[index]
+            result.append(a)
+            index += 1
+            if index < len(gpus) and gpus[index] != ",":
+                return None, False
+            else:
+                index += 1
+        return result, True
+
```
YuanTingHsieh, line 424: Suggest move this function outside of the class, since it does not need anything from class.

tests/unit_test/private/fed/app/simulator/simulator_runner_test.py
```diff
@@ -86,3 +86,31 @@ def test_split_names(self, client_names, gpus, expected_split_names):
...
+        split_gpus, _ = runner.split_gpus(gpus)
+        assert split_gpus == expected_gpus
+
+    @pytest.mark.parametrize(
+        "gpus",
+        [
+            "[0,1],3",
+            "0,1,2",
+        ],
+    )
```
YuanTingHsieh, line 109: suggest add test cases of :

`"[]"` => is this valid?
`"[a]"` => is this valid?
`"[01,3]"` => should be invalid
`"[0,1,2]extra"` => should be invalid

YuanTingHsieh (changes requested):
Thanks, logic LGTM.
Suggest move method outside class and add more test cases.

holgerroth:
> Why force into 2-GPU groups ?
> 
> Why not just use what POC did (which is originally copied from Holger's logics)
> 
> --gpu 0 1 2 3 4 5
> 
> Depends on number of the clients, we auto assign the GPUs to the corresponding clients
> 
> for example in above gpus, with 3 clients
> 
> site-1 : 0, 3 site-2 : 1, 4 site-3: 2, 5
> 
> with 2 clients
> 
> site-1: 0, 2, 4 site-2: 1, 3, 5

This could be an alternative way to configure the behavior. I also like the possibility to assign different number of gpus to different clients, e.g., "[0, 1],[2],[3,4,5]".

tests/unit_test/private/fed/app/simulator/simulator_runner_test.py
yhwen: split_gpus() is just a function to split the gpu list. It does not check / dictate the gpu name format. The gpu name could be potentially anything, depending on each different system. The first 3 cases are valid. The last one is invalid. I will add these into the unit test cases.

nvflare/private/fed/app/simulator/simulator_runner.py
yhwen: I am not quite sure about this. This is a unique function only used by this class.

yhwen:
> > Why force into 2-GPU groups ?
> > Why not just use what POC did (which is originally copied from Holger's logics)
> > --gpu 0 1 2 3 4 5
> > Depends on number of the clients, we auto assign the GPUs to the corresponding clients
> > for example in above gpus, with 3 clients
> > site-1 : 0, 3 site-2 : 1, 4 site-3: 2, 5
> > with 2 clients
> > site-1: 0, 2, 4 site-2: 1, 3, 5
> 
> This could be an alternative way to configure the behavior. I also like the possibility to assign different number of gpus to different clients, e.g., "[0, 1],[2],[3,4,5]".

The gpu group provides a more flexible way to assign the client to different gpus. Also with the simulator, it has the option to run large number of clients, but with small number of gpus.

nvflare/private/fed/app/simulator/simulator_runner.py
```diff
@@ -395,6 +399,29 @@ def simulator_run_main(self):
             run_status = 1
         return run_status
 
+    def split_gpus(self, gpus) -> ([str], bool):
+        gpus = gpus.replace(" ", "")
+        if not (gpus.startswith("[") and gpus.endswith("]")):
```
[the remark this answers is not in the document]
yanchengnv: Maybe we could use a different notation: instead of [1,2,3], we could use 1:2:3. This is much easier to parse - you don't need to worry about pairing up brackets. Just do split on ":".

nvflare/private/fed/app/simulator/simulator_runner.py
yhwen: Moved the split_gpus() to fed_utils.

nvflare/private/fed/app/simulator/simulator_runner.py
```diff
@@ -395,6 +399,29 @@ def simulator_run_main(self):
             run_status = 1
         return run_status
 
+    def split_gpus(self, gpus) -> ([str], bool):
```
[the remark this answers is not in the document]
yhwen: > why the function returns ( List[str], bool) ? instead of ( List[Tuple], bool) ?
> 
> essentially "[1,2],[3], [4,5,6]" --> split_gpus() -> [(1,2), (3,), (4,5,6)]

Just need to have the list of GPU IDs in string. I don't have the usage case for a Tuple.

yanchengnv (commented):
See my comments

nvflare/private/fed/app/simulator/simulator_runner.py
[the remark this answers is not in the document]
yhwen: I did think of using format 1:2:3 first. Only reason I didn't adopt it is this may looks like not so natural for a list / group of GPUs. Definitely the parsing for that would be much easier. -- But, maybe 1:2:3 format is also acceptable.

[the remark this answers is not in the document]
yanchengnv: The bracket syntax is fine. But we should make it backward compatible.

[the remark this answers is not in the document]
yhwen: Made it backward compatible, accepts "[0,1],3", or "0,1,2".

nvflare/private/fed/utils/fed_utils.py
```diff
@@ -236,3 +236,38 @@ def create_stats_pool_files_for_job(workspace: Workspace, job_id: str, prefix=No
         err = f"Failed to create stats pool summary file {summary_file}: {secure_format_exception(e)}"
     StatsPoolManager.close()
     return err
+
+
+def split_gpus(gpus) -> ([str], bool):
```
yanchengnv, line 241: This should work, but parsing seems to be tedious. Consider using something like this:


def find_char(s, ch):
    return [i for i, c in enumerate(s) if c == ch]


def split_gpus(gpus) -> [str]:
    s = gpus.replace(" ", "")
    lefts = find_char(s, '[')
    rights = find_char(s, ']')
    if len(lefts) != len(rights):
        raise ValueError('brackets not paired')

    offset = 0
    for i in range(len(lefts)):
        if lefts[i] > rights[i]:
            raise ValueError('brackets not properly paired')
        p = s[lefts[i]+1-offset:rights[i]-offset]
        p = p.replace(',', ':')
        s = s.replace(s[lefts[i]-offset:rights[i]+1-offset], p)
        offset += 2

    result = s.split(',')
    for i in range(len(result)):
        result[i] = result[i].replace(':', ',')
    print(f"{gpus} ==> {result}")
    return result

yanchengnv: Somehow the indentation got messed up, but you should be able to see it. The idea is to first find all pairing brackets. Then for the content in each pair, replace "," with ":"; and then put the replaced content back to the original list. This would reduce the total str size by 2 for each pair. After this process, the groups are only separated by ",".  Then you replace ":" with "," for each group.

This works for all your test cases, and it also works for syntax like "1:2:3, [4,5], [3:2], 1, 5".
It will also catch improperly paired brackets like:
"[1,2,3"
"]1, 2, 3["

tests/unit_test/private/fed/app/simulator/simulator_runner_test.py
```diff
@@ -86,3 +87,34 @@ def test_split_names(self, client_names, gpus, expected_split_names):
...
+        "gpus, expected_gpus",
+        [
+            ("[0,1],[1, 2]", ["0,1", "1,2"]),
+            ("[0,1],[3]", ["0,1", "3"]),
+            ("[0,1],[ 3 ]", ["0,1", "3"]),
+            ("[02,1],[ a ]", ["02,1", "a"]),
+            ("[]", [""]),
+            ("[0,1],3", ["0,1", "3"]),
+            ("0,1,2", ["0", "1", "2"]),
+        ],
```
yanchengnv, line 101: Suggest to add some negative cases:

"[2,1], ]1,2,3["
"[1,3,4"

yhwen: All these negative cases already handled, and the unit tests covered in test_split_gpus_fail()

nvflare/private/fed/utils/fed_utils.py
yhwen: I am not sure if "replace(',', ':')" is a good idea. This is a split gpus string function. In theory the GPU ID / name could be also contain ":" character as well, depending on different system. So maybe change the content of provided GPU name may break it.

yanchengnv: In that case, we could use a different char than ":", say "#" or "@". Idea is to make it easier for split with comma later.

yanchengnv: I also updated the code to protect against bad start or end of a group and nested brackets.

tests/unit_test/private/fed/app/simulator/simulator_runner_test.py
```diff
@@ -86,3 +87,34 @@ def test_split_names(self, client_names, gpus, expected_split_names):
...
+        splitted_gpus, _ = split_gpus(gpus)
+        assert splitted_gpus == expected_gpus
+
+    @pytest.mark.parametrize(
+        "gpus",
+        [
+            "[0,1],3]",
+            "0,1,[2",
+            "[0,1]extra"
+        ],
```
yanchengnv, line 113: what about:
[0, 1, [2, 3]]

yhwen: This is invalid. GPU group "[]" can not be included in another group.

nvflare/private/fed/utils/fed_utils.py
yanchengnv, line 241: Perhaps it's better to return a message instead if "bool" when the gpus string is invalid so help user know what's wrong.

yhwen: Changed.

nvflare/private/fed/utils/fed_utils.py
```diff
@@ -236,3 +240,34 @@ def create_stats_pool_files_for_job(workspace: Workspace, job_id: str, prefix=No
...
```
yanchengnv, line 245: Now since this func returns list of str, please change the signature to -> [str]

yanchengnv (changes requested):
Need to change func signature of split_gpus.

nvflare/private/fed/utils/fed_utils.py
yhwen: done

yanchengnv (approved):
LGTM

yanchengnv (approved):
LGTM

nvflare/private/fed/utils/fed_utils.py
```diff
@@ -236,3 +240,34 @@ def create_stats_pool_files_for_job(workspace: Workspace, job_id: str, prefix=No
         err = f"Failed to create stats pool summary file {summary_file}: {secure_format_exception(e)}"
     StatsPoolManager.close()
     return err
+
+
+def split_gpus(gpus) -> [str]:
```
YuanTingHsieh, line 245: One comment here, not sure if this type hint works or we need: `List[str]`.
But since we are not checking for type hints now in CI, we can ignore this.

YuanTingHsieh (dismissed):
Thanks, LGTM, one minor comment.

nvflare/private/fed/app/simulator/simulator_runner.py
```diff
@@ -175,7 +175,12 @@ def setup(self):
                 return False
 
             if self.args.gpu:
-                gpus = self.args.gpu.split(",")
+                try:
+                    gpus = split_gpus(self.args.gpu)
+                except ValueError as e:
+                    self.logger.error(f"GPUs group list option in wrong format. Error: {e}")
+                    return False
+
```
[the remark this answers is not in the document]
yhwen: Fixed. Also fixed an issue when checking the available GPU ID.

nvflare/private/fed/app/simulator/simulator_runner.py
```diff
@@ -175,7 +175,12 @@ def setup(self):
                 return False
 
             if self.args.gpu:
-                gpus = self.args.gpu.split(",")
+                try:
+                    gpus = split_gpus(self.args.gpu)
```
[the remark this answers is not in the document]
yhwen: changed.

nvflare/private/fed/app/simulator/simulator_runner.py
```diff
@@ -175,27 +175,33 @@ def setup(self):
...
-                if len(gpus) > len(self.client_names):
+                if len(gpu_groups) > len(self.client_names):
                     self.logger.error(
                         f"The number of clients ({len(self.client_names)}) must be larger than or equal to "
-                        f"the number of GPUS: ({len(gpus)})"
+                        f"the number of GPU groups: ({len(gpu_groups)})"
                     )
                     return False
-                if len(gpus) > 1:
+                if len(gpu_groups) > 1:
```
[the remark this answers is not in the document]
yhwen: The number of thread applies to each GPU group. Updated the log message.

yanchengnv (dismissed):
LGTM

holgerroth (approved):
Looks good.

PR #1755, opened 2023-05-19
Upgrade POC: changing POC mode to Local Prod Mode, but keep POC command name. 

chesterxgchen:
## Description
POC command was created to allow user experiment federated settings in local machines with client, server and admin console. The POC is a special kind of deployment mode, which is insecure mode.  We also developed “nvflare poc” CLI commands which make it easier to experiments NVFLARE. 

#### Goals:
The goals of this PR are the followings: 
* Preserve all the existing POC commands, –prepare/start/stop/clean as well as all the existing argument options 
-n number of clients
-qpus
-p package
-ex package 
etc. 

* Leverage existing documentations for POC so has smallest impact to the end-user
* Bring user one step closer to real deployment structures (startup kits) 
* In additional we will make few changes to allow enable
   add -d –docker option. This will allow FL Server and clients are running in dockers (user needs to provide docker image)
   admin console will still running in sub_process
  This would make it easier for user to experiment containerized env. in the localhost. 

* Allow user to specify examples directory to setup symlink with --prepare-example -e <examples_dir> or --prepare -e <example_dir>
  if the -e <examples_dir> is specified, the default NVFlare/examples will not be used even NVFLARE_HOME is specified. 
  User can overwrite the the existing examples/transfer directory with new customized code if needed

* Allow user to specify Homomorphic encryption with -he option. This will help user to set default Homomorphic encryption builder automatically
* Allow user to specific participating client names -c client-1 client-2  ( instead of default site-1, site-2 etc.)
* Allow user to provide his own customization project yaml ( with some limitations)  with -i /path/to/project.yml ( with some limitation)
* change ValueErrors with CLIException which will not print out the -help messages. 
* the resulting project.yml generated from **nvflare poc --prepare**  options can be directly taken to production and 
  using nvflare provision -p project.yml to re-produce production provision. 

#### Current POC drawbacks
* Not using project.yml → POC → Real Deployment gap, can’t test customized project.yml 
* POC is insecure → FLARE API needs to handle both insecure and secure sessions
* Technical debt → maintain both POC and Production runtime modes
* User confusion → POC is similar to real deployment, but not the same
* Can’t enable Homomorphic encryption 

#### POC upgrade proposal
* No longer support old POC mode
* Keep the POC name, but implementation will be production local, non-HA mode
* All server and clients are in localhost (just like POC mode)
* Secure mode, non-HA, only use DummyOverseerAgent 
* Using default project.yml as base if none supplied 
* change default server1 → Server
* Allow user to supply project.yml, all server and client name will be kept as it is, but the actually hostname is only localhost
The design doc can be found here: 
https://docs.google.com/document/d/1J0AjsHYHt3JhdZrS83BbBpTwvtaYaf52qF0zgqIU33g/edit#

### Types of changes

- [X] New tests added to cover the changes.
- [X] Quick tests passed locally by running `./runtest.sh`.
- [ ] Documentation updated.

### DIFF
--- a/nvflare/lighter/poc_commands.py
+++ b/nvflare/lighter/poc_commands.py
@@ -62,29 +62,29 @@ def client_gpu_assignments(clients: List[str], gpu_ids: List[int]) -> Dict[str, 
     return gpu_assignments
 
 
-def get_package_command(cmd_type: str, prod_dir: str, package_dir, global_packages: Dict) -> str:
+def get_service_command(cmd_type: str, prod_dir: str, service_dir, service_config: Dict) -> str:
     cmd = ""
-    admin_package = global_packages.get(SC.FLARE_PROJ_ADMIN, SC.FLARE_PROJ_ADMIN)
+    admin_dir_name = service_config.get(SC.FLARE_PROJ_ADMIN, SC.FLARE_PROJ_ADMIN)
     if cmd_type == SC.CMD_START:
-        if not global_packages.get(SC.IS_DOCKER_RUN):
-            if package_dir == admin_package:
-                cmd = get_cmd_path(prod_dir, package_dir, "fl_admin.sh")
+        if not service_config.get(SC.IS_DOCKER_RUN):
+            if service_dir == admin_dir_name:
+                cmd = get_cmd_path(prod_dir, service_dir, "fl_admin.sh")
             else:
-                cmd = get_cmd_path(prod_dir, package_dir, "start.sh")
+                cmd = get_cmd_path(prod_dir, service_dir, "start.sh")
         else:
-            if package_dir == admin_package:
-                cmd = get_cmd_path(prod_dir, package_dir, "fl_admin.sh")
+            if service_dir == admin_dir_name:
+                cmd = get_cmd_path(prod_dir, service_dir, "fl_admin.sh")
             else:
-                cmd = get_cmd_path(prod_dir, package_dir, "docker.sh -d")
+                cmd = get_cmd_path(prod_dir, service_dir, "docker.sh -d")
 
     elif cmd_type == SC.CMD_STOP:
-        if not global_packages.get(SC.IS_DOCKER_RUN):
-            cmd = get_stop_cmd(prod_dir, package_dir)
+        if not service_config.get(SC.IS_DOCKER_RUN):
+            cmd = get_stop_cmd(prod_dir, service_dir)
         else:
-            if package_dir == admin_package:
-                cmd = get_stop_cmd(prod_dir, package_dir)
+            if service_dir == admin_dir_name:
+                cmd = get_stop_cmd(prod_dir, service_dir)
             else:
-                cmd = f"docker stop {package_dir}"
+                cmd = f"docker stop {service_dir}"
 
     else:
         raise CLIException("unknown cmd_type :", cmd_type)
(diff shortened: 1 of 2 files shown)


IsaacYangSLA (dismissed):
LGTM.  My understanding after reviewing is the poc will generate a complete startup kits for every participant.  There are two additional builders to modify participant's info in place.  Basically, pointing server/overseer to localhost.

nvflare/lighter/impl/local_cert.py
```diff
@@ -0,0 +1,22 @@
+# Copyright (c) 2021, NVIDIA CORPORATION.  All rights reserved.
```
YuanTingHsieh, line 1: Please change to 2023

nvflare/lighter/impl/local_static_file.py
```diff
@@ -0,0 +1,71 @@
+# Copyright (c) 2021, NVIDIA CORPORATION.  All rights reserved.
```
YuanTingHsieh, line 1: Please change to 2023

nvflare/lighter/poc_commands.py
```diff
@@ -78,57 +97,359 @@ def get_stop_cmd(poc_workspace: str, service_dir_name: str):
...
+            if os.path.islink(dst):
+                os.unlink(dst)
+            if os.path.isdir(dst):
+                shutil.rmtree(dst, ignore_errors=True)
+
+            print(f"link examples from {src} to {dst}")
+            os.symlink(src, dst)
+    else:
+        if os.path.isdir(dst):
+            shutil.rmtree(dst, ignore_errors=True)
```
YuanTingHsieh, line 157: I thought remove will remove all examples in "dst"?

nvflare/lighter/poc_commands.py
```diff
@@ -78,57 +97,359 @@ def get_stop_cmd(poc_workspace: str, service_dir_name: str):
...
+
+
+def get_packages(project_config):
+    packages = {
+        SC.FLARE_SERVER: get_fl_server_name(project_config),
+        SC.FLARE_PROJ_ADMIN: get_proj_admin(project_config),
+        SC.FLARE_CLIENTS: get_fl_client_names(project_config),
+        SC.IS_DOCKER_RUN: is_docker_run(project_config),
+    }
+    return packages
```
YuanTingHsieh, line 284: Should we call this get_service_config?
Or what does a "package" mean?

nvflare/lighter/poc_commands.py
```diff
@@ -78,57 +97,359 @@ def get_stop_cmd(poc_workspace: str, service_dir_name: str):
...
+
+def add_he_builder(use_he: bool, project_config: OrderedDict):
+    if use_he:
+        he_builder = {
+            "path": "nvflare.lighter.impl.he.HEBuilder",
+            "args": {},
+        }
+        project_config["builders"].append(he_builder)
+
+    return project_config
```
YuanTingHsieh, line 339: Why do we want HE as default here?
When we get other builders, do we also want to add them here as an option?

nvflare/lighter/poc_commands.py
YuanTingHsieh: Yes, for get_packages I would expect it might returning a list of directories, or a list of objects.

"global_packages" to me is also not that clear.
We actually mean "service_config"?

or call this method "get_service_config_from_project"?

YuanTingHsieh (approved):
Thanks, LGTM

PR #1758, opened 2023-05-22
FCI Stream API

nvidianz:
### Description

This PR contains following Stream API definitions,
1. Byte Stream API
2. BLOB Streaming API
3. File Streaming API
4. Object Streaming API

### Types of changes

- [ ] Documentation updated.


nvflare/fuel/f3/cellnet/cell.py
```diff
@@ -625,7 +626,7 @@ def drop_connectors(self):
                     try:
                         self.communicator.remove_connector(connector.handle)
                         self.communicator.remove_endpoint(cell_name)
-                    except:
+                    except Exception:
```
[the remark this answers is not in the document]
nvidianz: The log_except will log exception so "as e" is not needed.

nvflare/fuel/f3/cellnet/cell.py
```diff
@@ -1972,3 +1972,154 @@ def _is_my_sub(self, candidate_info: FqcnInfo) -> int:
...
+    def register_stream_cb(self, channel: str, topic: str, stream_cb: Callable, *args, **kwargs):
+        """
+        Register a callback for reading stream. The stream_cb must have the following signature,
+            stream_cb(future: StreamFuture, stream: Stream, resume: bool, *args, **kwargs) -> int
+                future: The future represents the ongoing streaming. It's done when streaming is complete.
+                stream: The stream to read the receiving data from
+                resume: True if this is a restarted stream
+                It returns the offset to resume from if this is a restarted stream
+
+        The resume_cb returns the offset to resume from:
```
IsaacYangSLA, line 2006: When this resume_cb is called?

nvflare/fuel/f3/stream_cell.py
```diff
@@ -0,0 +1,209 @@
...
+from nvflare.fuel.f3.message import Message
+from nvflare.fuel.f3.streaming.blob_streamer import BlobStreamer
+from nvflare.fuel.f3.streaming.byte_receiver import ByteReceiver
+from nvflare.fuel.f3.streaming.byte_streamer import ByteStreamer
+from nvflare.fuel.f3.streaming.file_streamer import FileStreamer
+from nvflare.fuel.f3.streaming.object_streamer import ObjectStreamer
+from nvflare.fuel.f3.streaming.stream_types import ObjectIterator, ObjectStreamFuture, Stream, StreamError, StreamFuture
+
+
+class StreamCell:
```
yhwen, line 27: The StreamCell class holds 4 different kinds of Streamer, each performs separate independent type of functions, ByteStream, BlobStream, FileStream and ObjectStream. Every time when creating a StreamCell instance, it will create other unrelated streamer. Probably it's better to separate the different types of StreamCell, or use the factory pattern to create different type of streamer as needed.

Also is it possible to abstract the Streamer class, and make the ByteStreamer, BlobStreamer, FileStreamer and ObjectStreamer extend from it? Conceptually I think that's supposed to be. But right now they are all separately without any relationship.

nvflare/fuel/f3/streaming/tools/file_receiver.py
```diff
@@ -0,0 +1,84 @@
...
+
+        name = future.result()
+        print(f"Time elapsed: {(time.time() - start):.3f} seconds")
+        print(f"File {name} is sent")
+        self.file_received += 1
+
+        return name
+
+
+if __name__ == "__main__":
```
yhwen, line 67: Do we need this "__main__" function? Looks like it's only for the testing purpose. If so we can implement this testing function through the unit test.

nvflare/fuel/f3/streaming/tools/sender.py
```diff
@@ -0,0 +1,53 @@
...
+    def __init__(self, url: str):
+        core_cell = Cell(TX_CELL, url, secure=False, credentials={})
+        self.stream_cell = StreamCell(core_cell)
+        core_cell.start()
+
+    def send(self, blob: bytes) -> StreamFuture:
+        return self.stream_cell.send_blob(TEST_CHANNEL, TEST_TOPIC, RX_CELL, Message(None, blob))
+
+
+if __name__ == "__main__":
```
yhwen, line 44: same "__main__" testing function.

nvflare/fuel/f3/streaming/tools/utils.py
```diff
@@ -0,0 +1,51 @@
...
+        buf[buf_len : buf_len + temp_len] = temp[0:temp_len]
+        buf_len += temp_len
+        n += 1
+        if buf_len >= size:
+            break
+
+    return buf
+
+
+def setup_log(level):
```
yhwen, line 42: why do we need a separate logger set up? NVFlare already has one.

nvflare/fuel/f3/cellnet/cell.py
```diff
@@ -905,39 +866,39 @@ def register_request_cb(self, channel: str, topic: str, cb, *args, **kwargs):
...
-        self.error_handler_reg.set(channel, topic, _CB(cb, args, kwargs))
+        self.error_handler_reg.set(channel, topic, Callback(cb, args, kwargs))
 
     def _filter_outgoing_request(self, channel: str, topic: str, request: Message) -> Union[None, Message]:
         cbs = self.out_req_filter_reg.find(channel, topic)
         if not cbs:
             return None
         for _cb in cbs:
-            assert isinstance(_cb, _CB)
+            assert isinstance(_cb, Callback)
```
YuanTingHsieh, line 901: If we need this check on production code, we need to change assert to ```if not isinstance(_cb, Callback): raise ValueError```

nvflare/fuel/f3/cellnet/cell.py
```diff
@@ -1112,7 +1073,7 @@ def _send_target_messages(
                 self.logger.debug(f"{self.my_info.fqcn}: invoking outgoing request filters")
                 assert isinstance(req_filters, list)
                 for f in req_filters:
-                    assert isinstance(f, _CB)
+                    assert isinstance(f, Callback)
```
YuanTingHsieh, line 1076: If we need this check on production code, we need to change assert to ```if not isinstance(_cb, Callback): raise ValueError```

nvflare/fuel/f3/cellnet/cell.py
```diff
@@ -1477,20 +1438,19 @@ def _process_request(self, origin: str, message: Message) -> Union[None, Message
             self.logger.debug(f"{self.my_info.fqcn}: invoking incoming request filters")
             assert isinstance(req_filters, list)
             for f in req_filters:
-                assert isinstance(f, _CB)
+                assert isinstance(f, Callback)
```
YuanTingHsieh, line 1441: If we need this check on production code, we need to change assert to ```if not isinstance(_cb, Callback): raise ValueError```

If we don't need, we should remove these assert statements.

nvflare/private/fed/cmi.py
```diff
@@ -66,7 +65,7 @@ def __init__(
         self.cell.add_incoming_reply_filter(channel="*", topic="*", cb=self._filter_incoming_message)
 
     def new_cmi_message(self, fl_ctx: FLContext, headers=None, payload=None):
-        msg = new_message(headers, payload)
+        msg = Message(headers, payload)
```
YuanTingHsieh, line 68: Can we make this explicit `msg = Message(headers=headers, payload=payload)`?

nvflare/fuel/f3/cellnet/cell.py
```diff
@@ -215,7 +176,7 @@ def send(self):
             f"{self.cell.get_fqcn()}: bulk sender {self.target} sending bulk size {len(messages_to_send)}"
         )
         tms = [m.to_dict() for m in messages_to_send]
-        bulk_msg = new_message(payload=tms)
+        bulk_msg = Message(None, tms)
```
YuanTingHsieh, line 179: Can we make this explicit `msg = Message(headers=None, payload=tms)`?

nvflare/fuel/f3/streaming/tools/file_sender.py
```diff
@@ -0,0 +1,80 @@
...
+import threading
+import time
+
+from nvflare.fuel.f3.cellnet.cell import Cell, CellAgent
+from nvflare.fuel.f3.message import Message
+from nvflare.fuel.f3.stream_cell import StreamCell
+from nvflare.fuel.f3.streaming.tools.utils import RX_CELL, TEST_CHANNEL, TEST_TOPIC, TX_CELL, setup_log
+
+
+class FileSender:
```
YuanTingHsieh, line 25: If all these classes and scripts in "f3/streaming/tools" folder are for testing purpose, we can move them into unit test folder?

nvflare/fuel/f3/streaming/blob_streamer.py
```diff
@@ -0,0 +1,111 @@
...
+            while True:
+                buf = stream.read(chunk_size)
+                if not buf:
+                    break
+
+                length = len(buf)
+                if self.size > 0:
+                    self.buffer[buf_size : buf_size + length] = buf
+                else:
+                    self.buffer += buf
```
YuanTingHsieh, line 86: Will this self.buffer OOM?
When the model / message is very big, which streamer should we use?

nvflare/fuel/f3/cellnet/cell.py
nvidianz: Keyword should be only used when there are many arguments. Using keyword arg everywhere defeats the purpose of positional arg.

nvflare/fuel/f3/cellnet/cell.py
nvidianz: assert is only for dev only. No need to enforce the type here. It will throw exception when it's used.

nvflare/fuel/f3/cellnet/cell.py
nvidianz: Same as above.

nvflare/fuel/f3/cellnet/cell.py
nvidianz: Assert can be useful for developers. It's used everywhere. Let's leave it here for now.

nvflare/fuel/f3/cellnet/registry.py
```diff
@@ -0,0 +1,51 @@
...
+
+class Registry:
+    def __init__(self):
+        self.reg = {}  # channel/topic => _CB
+
+    @staticmethod
+    def _item_key(channel: str, topic: str) -> str:
+        return f"{channel}:{topic}"
+
+    def set(self, channel: str, topic: str, items):
```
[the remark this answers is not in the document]
nvidianz: Currently, only Callback is used as items but it can store anything in the future. Added hint "Any".

nvflare/fuel/f3/cellnet/registry.py
```diff
@@ -0,0 +1,51 @@
...
+        key = self._item_key(channel, topic)
+        self.reg[key] = items
+
+    def append(self, channel: str, topic: str, items):
+        key = self._item_key(channel, topic)
+        item_list = self.reg.get(key)
```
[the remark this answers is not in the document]
nvidianz: items can be anything. Currently, it's an Callback object. The append should just append a single item to the list.

nvflare/fuel/f3/streaming/stream_utils.py
```diff
@@ -0,0 +1,41 @@
...
+        view = memoryview(buffer)
+
+    return view
+
+
+def gen_stream_id():
+    global lock, stream_count, start_time
+    with lock:
+        stream_count += 1
+    return "SID%16d" % (start_time + stream_count)
```
IsaacYangSLA, line 41: Minor suggestion.  Change "SID%16d" % (start_time + stream_count) to f"SID{start_time + stream_count:16.0f}"

IsaacYangSLA (approved):
Just integrated with current dev branch and tested it with numpy-sag jobs with models ~= 1GB without issues.

PR #1766, opened 2023-05-25
Add changes to support mlflow experiment tracking

nvkevlu:
Add changes to support mlflow experiment tracking.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/examples/advanced/experiment-tracking/README.md
+++ b/examples/advanced/experiment-tracking/README.md
@@ -1,8 +1,12 @@
 # Experiment Tracking Overview
 
-Please make sure you set up virtual environment and follows [example root readme](../../README.md)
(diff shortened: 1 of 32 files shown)


chesterxgchen (changes requested):
The PR is too large. 

can we break into two and possible three PRs
1) Examples
2) MLFLow 
3) Log Writer and Data Type Mapping

Other comments
1) why in the example directories named with "-streaming" ? 
    we already have experiment-tracking in the parent directory, we can simply name the mlflow-streaming => mlflow, tensorboard-streaming => tensorboard
2) why do we have two tb_receiver.py files in different directories

nvkevlu:
addressing the comments and breaking this up into multiple PRs

PR #1768, opened 2023-05-25
Make changes for groundwork for experiment tracking

nvkevlu:
PR without examples or MLflow yet, to lay the groundwork for upcoming experiment tracking updates.

This PR renames some components and refactors to keep things updated. TBWriter is used instead of AnalyticsSender, but the methods in AnalyticsSender are kept with descriptions that they are legacy methods.

### Types of changes

- [ ] Documentation updated.


nvflare/app_common/widgets/streaming.py
```diff
@@ -50,117 +53,126 @@ def send_analytic_dxo(comp: FLComponent, dxo: DXO, fl_ctx: FLContext, event_type
...
         Args:
             tag (str): Data identifier.
             scalar (float): Value to send.
             global_step (optional, int): Global step value.
             **kwargs: Additional arguments to pass to the receiver side.
         """
-        self._add(tag=tag, value=scalar, data_type=AnalyticsDataType.SCALAR, global_step=global_step, kwargs=kwargs)
+        self.add(tag=tag, value=scalar, data_type=AnalyticsDataType.SCALAR, global_step=global_step, **kwargs)
 
     def add_scalars(self, tag: str, scalars: dict, global_step: Optional[int] = None, **kwargs):
```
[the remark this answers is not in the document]
nvkevlu: I mention in the docstring that these are legacy methods, I guess keeping it is good if someone is already using it from before and we don't want to force them to change it to make things work.

tests/unit_test/apis/analytix_test.py
```diff
@@ -64,32 +64,35 @@
...
-    ),
+    # (
+    #     "tag",
+    #     1.0,
+    #     AnalyticsDataType.SCALAR,
+    #     [1],
+    #     TypeError,
+    #     f"expect kwargs to be an instance of dict, but got {type(list())}.",
+    # ),
 ]
```
YuanTingHsieh, line 75: Why we comment this out?
if not applicable we should remove.
if applicable we should fix the test.

runtest.sh
```diff
@@ -89,7 +89,7 @@ function check_license() {
     folders_to_check_license="nvflare examples tests integration research"
     echo "checking license header in folder: $folders_to_check_license"
     (grep -r --include "*.py" --exclude-dir "*protos*" -L \
-    "\(# Copyright (c) \(2021\|2022\|2023\), NVIDIA CORPORATION.  All rights reserved.\)\|\(This file is released into the public domain.\)" \
+    "\(# Copyright (c) \(2021\|2022\|2023\|2021-2023\|2022-2023\), NVIDIA CORPORATION.  All rights reserved.\)\|\(This file is released into the public domain.\)" \
```
YuanTingHsieh, line 92: I think we only need to specify the "create" year.
So it's easier for us to maintain, otherwise every year we need to modify all headers.

nvflare/app_opt/tracking/tb/tb_writer.py
```diff
@@ -0,0 +1,62 @@
+# Copyright (c) 2022, NVIDIA CORPORATION.  All rights reserved.
```
YuanTingHsieh, line 1: year should be 2023

YuanTingHsieh: ah.. this file is copy from old file, never mind.

nvflare/app_common/widgets/streaming.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2021, NVIDIA CORPORATION.  All rights reserved.
+# Copyright (c) 2021-2023, NVIDIA CORPORATION.  All rights reserved.
```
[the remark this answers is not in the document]
nvkevlu: I can address these issues in the next PR since this is merged already

PR #1770, opened 2023-05-26
Add DXOExchanger and FLModelExchanger

YuanTingHsieh:
### Description

- Makes heartbeat_timeout of PipeMonitor optional (so we can enable either side NOT writing HEARTBEAT)
- Adds nvflare/app_common/dex/dxo_exchanger.py
- Adds nvflare/app_common/dex/dxo_file_exchanger.py
- Adds nvflare/app_common/dex/fl_model_exchanger.py
- Adds nvflare/fuel/utils/pipe/pickle_file_accessor.py
- Adds unit tests

### Types of changes

- [ ] Documentation updated.


nvflare/app_common/dex/dxo_exchanger.py
```diff
@@ -0,0 +1,115 @@
...
+
+from nvflare.apis.dxo import DXO
+from nvflare.fuel.utils.pipe.file_accessor import FileAccessor
+from nvflare.fuel.utils.pipe.file_pipe import FilePipe
+from nvflare.fuel.utils.pipe.pickle_file_accessor import PickleFileAccessor
+from nvflare.fuel.utils.pipe.pipe import Message
+from nvflare.fuel.utils.pipe.pipe_monitor import PipeMonitor, Topic
+
+
+class DXOExchanger:
```
yanchengnv, line 28: This should be called DXOFileExchanger based on the abstract class DXOExchanger, which should use Pipe object.

nvflare/app_common/dex/dxo_exchanger.py
```diff
@@ -0,0 +1,115 @@
...
+        file_pipe.open(self._pipe_name, self._pipe_role)
+        self.pipe_monitor = PipeMonitor(
+            file_pipe,
+            read_interval=self._get_poll_interval,
+            heartbeat_interval=self._heartbeat_interval,
+            heartbeat_timeout=self._heartbeat_timeout,
+        )
+        self.pipe_monitor.start()
+
+    def put(self, data_id: str, data: DXO):
```
yanchengnv, line 78: This put() method should:
- has timeout param
- return a bool (whether the msg is read by peer).

nvflare/app_common/dex/dxo_exchanger.py
```diff
@@ -0,0 +1,115 @@
...
+                raise RuntimeError("the other end is aborted")
+            elif msg.topic in [Topic.END, Topic.PEER_GONE]:
+                raise RuntimeError(f"received {msg.topic}: {msg.data} while waiting for result for {req_topic}")
+            elif msg.topic == req_topic and data_id is None:
+                return msg.data
+            elif msg.topic == req_topic and msg.msg_id == data_id:
+                return msg.data
+            time.sleep(self._get_poll_interval)
+
+    def get(self, data_id: Optional[str] = None, timeout: Optional[float] = None):
```
yanchengnv, line 106: This data_id seems to be topic. It that's the case, then just use the term "topic", instead of "data_id".

nvflare/app_common/dex/dxo_exchanger.py
```diff
@@ -0,0 +1,115 @@
...
+            read_interval=self._get_poll_interval,
+            heartbeat_interval=self._heartbeat_interval,
+            heartbeat_timeout=self._heartbeat_timeout,
+        )
+        self.pipe_monitor.start()
+
+    def put(self, data_id: str, data: DXO):
+        if self.pipe_monitor is None:
+            raise RuntimeError("PipeMonitor is not initialized.")
+        req = Message.new_request(self._file_accessor_name, data, data_id)
```
yanchengnv, line 81: The topic should not be self._file_accessor_name, which should not be exposed at all.  
We should get rid of the term "data_id" and replace it with "topic".

yanchengnv (changes requested):
The concept of "data_id" is confusing and seems to be misused/mixed with the concept of "topic". We need to correct these.
Should define a DXOExchanger as an abstract base class, and extend it to create a DXOFileExchanger.
We will need DXOMemoryExchanger later.

nvflare/app_common/dex/dxo_exchanger.py
```diff
@@ -0,0 +1,115 @@
...
+
+
+class DXOExchanger:
+    def __init__(
+        self,
+        pipe_role: str,
+        pipe_name: str = "pipe",
+        get_poll_interval: float = 0.1,
+        heartbeat_interval: float = 0.5,
+        heartbeat_timeout: float = 0.0,
```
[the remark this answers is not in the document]
YuanTingHsieh: thanks, changed.

nvflare/app_common/dex/dxo_exchanger.py
YuanTingHsieh: thanks, changed.

nvflare/app_common/dex/dxo_exchanger.py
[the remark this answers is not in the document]
YuanTingHsieh: thanks, changed to put_request/reply,  get_request/reply

YuanTingHsieh: thanks, changed to put_request/reply,  get_request/reply

nvflare/app_common/dex/dxo_exchanger.py
YuanTingHsieh: thanks, changed.

nvflare/app_common/dex/dxo_exchanger.py
YuanTingHsieh: thanks, changed.

nvflare/app_common/dex/dxo_exchanger.py
```diff
@@ -0,0 +1,139 @@
...
+        pipe_role: str,
+        pipe_name: str = "pipe",
+        get_poll_interval: float = 0.1,
+        heartbeat_interval: float = 1.0,
+        heartbeat_timeout: float = 30.0,
+    ):
+        """Exchanges DXO.
+
+        Args:
+            get_poll_interval (float): how often to check if the other side has sent file
```
yanchengnv, line 57: Change the text to: "how often to check if the other side has sent **data**", since the pipe doesn't have to be file based.

nvflare/app_common/dex/dxo_exchanger.py
```diff
@@ -0,0 +1,139 @@
...
+    pass
+
+
+class DXOExchanger(ABC):
+    def __init__(
+        self,
+        pipe_role: str,
+        pipe_name: str = "pipe",
+        get_poll_interval: float = 0.1,
+        heartbeat_interval: float = 1.0,
```
yanchengnv, line 51: This may be too aggressive. 5 secs should be enough.

nvflare/app_common/dex/dxo_exchanger.py
```diff
@@ -0,0 +1,139 @@
...
+        self._heartbeat_timeout = heartbeat_timeout
+        self._topic = "dxo"
+        self.pipe_monitor = None
+
+    @abstractmethod
+    def create_pipe(self, *args, **kwargs) -> Pipe:
+        """Creates a Pipe."""
+        pass
+
+    def initialize(self, *args, **kwargs):
```
yanchengnv, line 77: Please add docstring

nvflare/app_common/dex/dxo_exchanger.py
```diff
@@ -0,0 +1,139 @@
...
+        if self.pipe_monitor is None:
+            raise RuntimeError("PipeMonitor is not initialized.")
+        start = time.time()
+        while True:
+            msg: Message = self.pipe_monitor.get_next()
+            if not msg:
+                if timeout and time.time() - start > timeout:
+                    self.pipe_monitor.notify_abort()
+                    raise ExchangeTimeoutException(f"get file timeout after {timeout} secs")
+            elif msg.topic == Topic.HEARTBEAT:
```
yanchengnv, line 120: Don't need this since HEARTBEAT is handled by pipe_monitor internally.

nvflare/app_common/dex/dxo_file_exchanger.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+class DXOFileExchanger(DXOExchanger):
+    def create_pipe(self, data_exchange_path: str, file_accessor: Optional[FileAccessor] = None) -> FilePipe:
+        data_exchange_path = os.path.abspath(data_exchange_path)
+        file_pipe = FilePipe(data_exchange_path)
+        if file_accessor is not None:
+            file_pipe.set_file_accessor(file_accessor)
+        else:
+            file_pipe.set_file_accessor(PickleFileAccessor())
```
yanchengnv, line 31: Using pickle could cause some criticism from other people, though this should be perfect safe here...

yanchengnv (changes requested):
Very nicely done! Some comments:
- Please add docstrings to all public methods
- Consider another file accessor other than pickle since it is very controversial

nvflare/app_common/dex/dxo_exchanger.py
YuanTingHsieh: thanks, fixed.

nvflare/app_common/dex/dxo_exchanger.py
YuanTingHsieh: thanks, fixed.

nvflare/app_common/dex/dxo_exchanger.py
YuanTingHsieh: thanks, fixed.

nvflare/app_common/dex/dxo_file_exchanger.py
YuanTingHsieh: Got it, removed.

nvflare/app_common/dex/fl_model_exchanger.py
```diff
@@ -0,0 +1,77 @@
...
+# limitations under the License.
+
+from typing import Optional, Tuple
+
+from nvflare.app_common.abstract.fl_model import FLModel
+from nvflare.app_common.dex.dxo_exchanger import DXOExchanger
+from nvflare.app_common.utils.fl_model_utils import FLModelUtils
+
+
+class FLModelExchanger:
```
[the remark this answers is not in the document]
yanchengnv: We could call this class a different name so it won't be confused with a subclass of DXOExchanger.

nvflare/app_common/dex/fl_model_exchanger.py
```diff
@@ -0,0 +1,77 @@
...
+    def __init__(self, exchanger: DXOExchanger):
+        self.exchanger = exchanger
+
+    def send_request(self, model: FLModel, timeout: Optional[float] = None) -> Tuple[bool, str]:
```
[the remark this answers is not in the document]
yanchengnv: No, this is not the highest-level API.  This API is still part of the "basic mechanism". User-facing APIs need to be built on top of this.

The destination is always "the other side", which is configured with "pipe_role" (or side) and "pipe_name", which is a path (file based) or URL (IPC based) or queue id (memory based).

send_request is to send the FLModel object to the peer.  Pipe is just the underlying mechanism for transmitting the data and it doesn't matter whether it's file based or not.

I guess we could send FLModel directly without having to convert to DXO, because the underlying pipe doesn't really care what type of data it is. In fact, if you look at DXOExchanger, there is no DXO-specific processing at all. So it could be simply called DataExchanger, and the type of input/output should be type "Any" (replace DXO with Any). The place that serialize the payload is in FileAccessor object.  We won't need dxo_file_exchanger any more.

The DataExchanger could be used for any data, and the pipe takes care of converting the data to/from the right data objects (e.g. FLModel). I like this!

Then this fl_model_exchanger (or a different name) could be re-purposed to be the user-facing API. For the ease of adoption, perhaps we should have two sub-APIs: one for the sender side (FLARE side), one for the external app side. FLARE side only cares about send_model and get_reply; external app cares about get_model, submit_metrics, and submit_model.

YuanTingHsieh:
As we discussed, we will have the following design:

NVFlare Client Executor side:
    Add LauncherExecutor which directly use Pipe.
      
3rd party / client original code changes:
    - Either use ModelExchanger 
    - Or use easy ml-to-fl "decorators"

Close this PR.

PR #1771, opened 2023-05-26
Add mlflow components

nvkevlu:
Adds the MLflowReceiver and MLflowWriter for integrating with MLflow.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/app_opt/tracking/mlflow/mlflow_receiver.py
+++ b/nvflare/app_opt/tracking/mlflow/mlflow_receiver.py
@@ -237,7 +237,7 @@ class MLflowReceiver(AnalyticsReceiver):
     def buffer_data(self, data: AnalyticsData, record_origin: str) -> None:
         """Buffer the data to send later.
 
-        A buffer for each data_type is in each site_buffer, all of which are in self.buffer 
+        A buffer for each data_type is in each site_buffer, all of which are in self.buffer
 
         Args:
             data (AnalyticsData): Data.
--- a/nvflare/app_opt/tracking/mlflow/mlflow_receiver.py
+++ b/nvflare/app_opt/tracking/mlflow/mlflow_receiver.py
@@ -27,7 +27,7 @@ from nvflare.apis.dxo import from_shareable
 from nvflare.apis.fl_context import FLContext
 from nvflare.apis.shareable import Shareable
 from nvflare.app_common.tracking.track_exception import ExpTrackingException
-from nvflare.app_common.tracking.tracker_types import LogWriterName, TrackConst
+from nvflare.app_common.tracking.tracker_types import ANALYTIC_EVENT_TYPE, LogWriterName, TrackConst
 from nvflare.app_common.widgets.streaming import AnalyticsReceiver
 
 
@@ -70,7 +70,7 @@ class MLflowReceiver(AnalyticsReceiver):
                                        traffic to the MLflow tracking server, which in some cases can actually cause more latency.
         """
         if events is None:
-            events = ["fed.analytix_log_stats"]
+            events = ["fed."+ANALYTIC_EVENT_TYPE]
         super().__init__(events=events)
         self.artifact_location = artifact_location
         self.fl_ctx = None
@@ -83,7 +83,7 @@ class MLflowReceiver(AnalyticsReceiver):
         self.run_ids = {}
         self.buffer = {}
         self.time_start = 0
-        self.time_taken = 0
+        self.time_since_flush = 0
         self.buff_flush_time = buffer_flush_time
 
         if self.tracking_uri:
@@ -218,9 +218,6 @@ class MLflowReceiver(AnalyticsReceiver):
         if not data:
             return
 
-        # mlflow_client = self.get_mlflow_client(record_origin)
-        # run_id = self.get_run_id(record_origin)
-
         if data.data_type == AnalyticsDataType.TEXT:
             # not currently supported
             pass
@@ -232,7 +229,9 @@ class MLflowReceiver(AnalyticsReceiver):
             pass
         else:
             self.buffer_data(data, record_origin)
-            self.flush_buffer(record_origin)
+            self.time_since_flush += timeit.default_timer() - self.time_start
+            if self.time_since_flush >= self.buff_flush_time:
+                self.flush_buffer(record_origin)
 
     def buffer_data(self, data: AnalyticsData, record_origin: str) -> None:
         """Buffer the data to send later.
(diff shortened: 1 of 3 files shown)


chesterxgchen (commented):
LGTM

nvflare/app_opt/tracking/mlflow/mlflow_writer.py
```diff
@@ -0,0 +1,106 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Dict, Optional
+
+from nvflare.apis.analytix import AnalyticsDataType
+from nvflare.app_common.tracking.log_writer import LogWriter
+from nvflare.app_common.tracking.tracker_types import LogWriterName
```
YuanTingHsieh, line 19: Since we are using the terms "tracking" everyone,
Does it make sense to have "LogWriter" -> "Tracker"
"LogWriterName" -> "TrackerName"?

nvflare/app_opt/tracking/mlflow/mlflow_writer.py
```diff
@@ -0,0 +1,106 @@
...
+from nvflare.app_common.widgets.streaming import ANALYTIC_EVENT_TYPE
+
+
+class MLflowWriter(LogWriter):
+    def __init__(self, event_type: str = ANALYTIC_EVENT_TYPE):
+        """MLflowWriter mimics the usage of mlflow.
+
+        For a client, replacing the import of mlflow with MLflowWriter will send log records to
+        the receiver on the server side instead of
```
YuanTingHsieh, line 28: Suggest changing to the following:
```
Users can replacing the import of mlflow with MLflowWriter.
Then they will use MLflowWriter the same as they would use mlflow.
MLflowWriter will send log records to the receiver.
```

Note that the receiver does NOT necessary need to be on the server side, it could be on the client side as well.

nvflare/app_opt/tracking/mlflow/mlflow_receiver.py
```diff
@@ -0,0 +1,324 @@
...
+        run_tags["job_id"] = self.get_job_id_tag(group_id=run_group_id)
+        run_tags["run_name"] = run_name
+        return run_tags
+
+    def get_job_id_tag(self, group_id: str) -> str:
+        job_id = self.fl_ctx.get_job_id()
+        if job_id == "simulate_job":
+            # Since all jobs run in the simulator have the same job_id of "simulate_job", use group_id instead
+            job_id = group_id
+        return job_id
```
YuanTingHsieh, line 164: Should we do this part here?

If the all jobs run in the simulator have the same job_id of "simulate_job", then that is simulator's issue/implementation.

We should just use job_id here IMO.

If we want simulator to use new job name or custom job name, we should modify simulator?

nvflare/app_opt/tracking/mlflow/mlflow_receiver.py
```diff
@@ -0,0 +1,324 @@
...
+        if self.time_start == 0:
+            self.time_start = timeit.default_timer()
+
+        dxo = from_shareable(shareable)
+        data = AnalyticsData.from_dxo(dxo, receiver=LogWriterName.MLFLOW)
+        if not data:
+            return
+
+        # mlflow_client = self.get_mlflow_client(record_origin)
+        # run_id = self.get_run_id(record_origin)
```
YuanTingHsieh, line 222: suggest remove commented lines

nvflare/app_opt/tracking/mlflow/mlflow_receiver.py
```diff
@@ -0,0 +1,324 @@
...
+        if data_type == AnalyticsDataType.METRIC:
+            return AnalyticsDataType.METRICS
+        elif data_type == AnalyticsDataType.PARAMETER:
+            return AnalyticsDataType.PARAMETERS
+        elif data_type == AnalyticsDataType.TAG:
+            return AnalyticsDataType.TAGS
+        else:
+            return data_type
+
+    def flush_buffer(self, record_origin, force_flush: bool = False):
```
YuanTingHsieh, line 276: Suggest remove the "force_flush" option to be consistent with general "flush" method definition.

Suggets move the check for whether to flush or not logic in "save" method:


```
self.time_taken += timeit.default_timer() - self.time_start
if self.time_taken >= self.buff_flush_time:
    flush_buffer()
```

nvflare/app_opt/tracking/mlflow/mlflow_receiver.py
```diff
@@ -0,0 +1,324 @@
...
+            pass
+        elif data.data_type == AnalyticsDataType.MODEL:
+            # not currently supported
+            pass
+        elif data.data_type == AnalyticsDataType.IMAGE:
+            # not currently supported
+            pass
+        else:
+            self.buffer_data(data, record_origin)
+            self.flush_buffer(record_origin)
```
YuanTingHsieh, line 235: move it here essentially.

nvflare/app_opt/tracking/mlflow/mlflow_receiver.py
```diff
@@ -0,0 +1,324 @@
...
+            events (_type_, optional): The event the receiver is listening to. By default, it listens to "fed.analytix_log_stats".
+            buffer_flush_time (int, optional): The time in seconds between deliveries of event data to the MLflow tracking server. The
+                                       data is buffered and then delivered to the MLflow tracking server in batches, and
+                                       the buffer_flush_time controls the frequency of the sending. By default, the buffer
+                                       flushes every second. You can reduce the time to a fraction of a second if you prefer
+                                       less delay. Keep in mind that reducing the buffer_flush_time will potentially cause high
+                                       traffic to the MLflow tracking server, which in some cases can actually cause more latency.
+        """
+        if events is None:
+            events = ["fed.analytix_log_stats"]
```
YuanTingHsieh, line 73: This "analytix_log_stats" we can use the constants we defined?

I am not sure if we want to have this default to be this or not.
This default only works IF client side has a Writer and a ConvertToFedevent component with prefix "fed."

nvflare/app_opt/tracking/mlflow/mlflow_writer.py
nvkevlu: For this, I think it makes sense that LogWriter is a type of Tracker, so I think keeping this for now is okay, but we can change it if people think it makes more sense.

nvflare/app_opt/tracking/mlflow/mlflow_receiver.py
nvkevlu: This is to make the UI of mlflow have different run names, which is pretty important if there are multiple separate runs. I think in principle what you say makes sense but for now maybe keep this as is. I may further refine how this is as well later.

nvflare/app_opt/tracking/mlflow/mlflow_receiver.py
nvkevlu: Since we need something to be the default, I think for now, this is okay. I changed it to use a constant.

YuanTingHsieh (approved):
Thanks, LGTM.

The JOB ID/RUN thing might need bigger changes across multiple classes if we decide to do that.

PR #1772, opened 2023-05-26
Add examples for MLflow

nvkevlu:
Adds examples for MLflow experiment tracking and jupyter notebook.

### Types of changes

- [ ] Documentation updated.


chesterxgchen (changes requested):
reduce duplicated code

chesterxgchen (changes requested):
reduce duplicated code

nvflare/app_common/widgets/streaming.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2021-2023, NVIDIA CORPORATION.  All rights reserved.
+# Copyright (c) 2021, NVIDIA CORPORATION.  All rights reserved.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
```
YuanTingHsieh, line 4: For `AnalyticsSender` I noticed we deprecate some methods here.

Maybe we can write a decorator of "deprecated" or similar.
And we can pass those deprecation information to that decorator, and it will prints out automatically when user use that method.

For example:

```
@deprecated("This method is deprecated, please use :py:class:`TBWriter <nvflare.app_opt.tracking.tb.tb_writer.TBWriter>` instead")
def add_scalar(self, tag: str, scalar: float, global_step: Optional[int] = None, **kwargs):
      """Sends a scalar."""
```

examples/advanced/experiment-tracking/mlflow/jobs/hello-pt-tb-mlflow/app/config/config_fed_server.json
```diff
@@ -0,0 +1,89 @@
...
+      "path": "nvflare.app_common.widgets.validation_json_generator.ValidationJsonGenerator",
+      "args": {}
+    },
+    {
+      "id": "tb_analytics_receiver",
+      "name": "TBAnalyticsReceiver"
+    },
+    {
+      "id": "mlflow_receiver_with_tracking_uri",
+      "path": "nvflare.app_opt.tracking.mlflow.mlflow_receiver.MLflowReceiver",
```
YuanTingHsieh, line 49: So we are having only one TB writer/sender/tracker on the client side.

BUT we have TBAnalyticsReceiver and MLflowReceiver on server side?
How does this work?

For this I suggest we either use names for both or paths for both?

examples/advanced/experiment-tracking/mlflow/jobs/hello-pt-mlflow/app/config/config_fed_server.json
```diff
@@ -0,0 +1,89 @@
...
+      "path": "nvflare.app_common.widgets.validation_json_generator.ValidationJsonGenerator",
+      "args": {}
+    },
+    {
+      "id": "tb_analytics_receiver",
+      "name": "TBAnalyticsReceiver"
+    },
+    {
+      "id": "mlflow_receiver_with_tracking_uri",
+      "path": "nvflare.app_opt.tracking.mlflow.mlflow_receiver.MLflowReceiver",
```
YuanTingHsieh, line 49: Same here.

Does this mean we will have 2 files written out on the server side?

examples/advanced/experiment-tracking/mlflow/jobs/hello-pt-mlflow/app/config/config_fed_server.json
```diff
@@ -0,0 +1,89 @@
...
+    },
+    {
+      "id": "json_generator",
+      "path": "nvflare.app_common.widgets.validation_json_generator.ValidationJsonGenerator",
+      "args": {}
+    },
+    {
+      "id": "tb_analytics_receiver",
+      "path": "nvflare.app_opt.tracking.tb.tb_receiver.TBAnalyticsReceiver"
+    },
```
YuanTingHsieh, line 46: Question: We only configure "MLflowWriter" in the client config.

So for the same record,
Will server side saves once using "TBAnalyticsReceiver",
And save once using "MLflowReceiver"?

nvkevlu: Yes, I believe that they both independently handle the events from the writer.

chesterxgchen (approved):
LGTM

PR #1781, opened 2023-06-01
Allow additional Job Configuration Options

chesterxgchen:
### Description
**Goals:**
In addition to the current JSON configuration for jobs, system are able to choose alternative configurations formats such as  Pyhocon (python HOCON, considered it as superset of json) and Omegaconf ( considered it as superset of YAML) 

This will allow user to add

Comment support ( inline, block)
nested configuration structure
ENV variable substitution
variable substitution 
Multi-line strings ( for pyhocon) 
optional overwrite from env. Variables (pyhocon) 
Dictionary, list merge
Include directive support (pyhocon) 
Default support API
dot notation for access nested elements

**Non-goals:**
Replace current configuration. 

### Design Changes: 
1 re-design the configuration to support the new configuration alternatives
  * for given config input_file path such xxx.json. The system is actually ignore the ".json" extension and search based on the lookup order for files that has the names. The lookup order is [.json, .json.default, .conf, .conf.default, .yml, .yml.default]
  * The first found match file will be consider the real configuration file. 
  * The .json and .json.default will be considered JSON ConfigFormat, .conf, .conf.default will be Pyhocon format and similar yml goes to OmegaConf. 
  * Currently only Json and Pyhocon are implemented, OmegaConf will be added in later PRs. 
  * There are several benefts for this change: 
  * User can choose to use any configuration format in any one of the config files. All he needs to do is provide the corresponding config file for the given name but different extension. 
  * New Configuration Format can be added, if desired, by adding new ConfigLoader, Config

2. Example: 
    PSI example config fed_client_config.json is changed fed_client_config.conf. Similarly for fed_server_config.conf
    Since the configuration is very simple, change to pyhocon from json doesn't show much benefits. 
    We might show other benefits with other examples

### design doc
for detailed discussion and design, please look at  [design doc](https://docs.google.com/document/d/1t_1BjgQ6OCxw2LpaIReuXhtvIBnh7NDU7a0amGY23c8)

 

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/fuel/utils/config_factory.py
+++ b/nvflare/fuel/utils/config_factory.py
@@ -37,9 +37,9 @@ class ConfigFactory:
             search_dirs = [str(parent_dir)]
         file_path = os.path.splitext(pathlib.Path(init_file_path).name)[0]
         for search_dir in search_dirs:
-            for fmt, ext in ConfigFormat.ordered_search_extensions():
-                logger.debug(f"search format {fmt.name} with ext {ext},file:{file_path}, search dirs = {search_dirs}")
-                for root, dirs, files in os.walk(search_dir):
+            logger.debug(f"search file:{file_path} basename, search dirs = {search_dirs}")
+            for root, dirs, files in os.walk(search_dir):
+                for fmt, ext in ConfigFormat.ordered_search_extensions():
                     file = f"{file_path}{ext}"
                     if file in files:
                         config_file_path = os.path.join(root, file)


yanchengnv:
I have some suggestions:

- We should abstract loading logic into something like ConfigLoader. Then we could develop different types of loaders to basically load config file content to dict. We shouldn't use "if" statements on format.
- Instead of using a env var to determine the config format, perhaps we should decide the loader dynamically based on the existence of config files. For example, we can check for ".json" first, then ".conf", ".ymal", etc. We may not be able to require all files to be based on the same format since they are defined by different people. For example, config files in "local" are defined by site admin; whereas jobs are defined by users.

PR #1785, opened 2023-06-01
[main] include exception type

YuanTingHsieh:
### Description

- Include exception type
- Fix unit test for https://github.com/NVIDIA/NVFlare/pull/1742 by @taleinat 

By Tal:

```python
>>> int("a")

>>> try:
	int("a")
except ValueError as exc:
	print(str(exc))

```

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/security/logging.py
+++ b/nvflare/security/logging.py
@@ -130,4 +130,4 @@ def secure_format_exception(e: Exception) -> str:
     if is_secure():
         return str(type(e))
     else:
-        return f"{str(type(e))}: {str(e)}"
+        return f"{type(e).__name__}: {str(e)}"
--- a/nvflare/security/logging.py
+++ b/nvflare/security/logging.py
@@ -130,4 +130,4 @@ def secure_format_exception(e: Exception) -> str:
     if is_secure():
         return str(type(e))
     else:
-        return str(e)
+        return f"{str(type(e))}: {str(e)}"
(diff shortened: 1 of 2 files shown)


nvflare/security/logging.py
```diff
@@ -130,4 +130,4 @@ def secure_format_exception(e: Exception) -> str:
     if is_secure():
         return str(type(e))
     else:
-        return str(e)
+        return f"{str(type(e))}: {str(e)}"
```
taleinat, line 133: I've recently noticed this doesn't print what I expected, and should be fixed thus:
```suggestion
        return f"{type(e).__name__}: {str(e)}"
```

tests/unit_test/fuel/utils/component_builder_test.py
```diff
@@ -77,9 +77,9 @@ def test_component_failure(self):
 
         # the failure message changes since 3.10
         if is_python_greater_than_309():
-            msg = "Class nvflare.app_common.np.np_model_locator.NPModelLocator has parameters error: NPModelLocator.__init__() got an unexpected keyword argument 'xyz'."
+            msg = "Class nvflare.app_common.np.np_model_locator.NPModelLocator has parameters error: <class 'TypeError'>: NPModelLocator.__init__() got an unexpected keyword argument 'xyz'."
```
taleinat, line 80:
```suggestion
            msg = "Class nvflare.app_common.np.np_model_locator.NPModelLocator has parameters error: TypeError: NPModelLocator.__init__() got an unexpected keyword argument 'xyz'."
```

taleinat: Similarly in the other tests.

nvflare/security/logging.py
YuanTingHsieh: thanks, I will change accordingly

taleinat (approved):
LGTM

PR #1791, opened 2023-06-02
Add missing cert chain verification on submitted job signatures

IsaacYangSLA:
### Description

Current implementation on job signing includes
- sign all files at each folder in the submitted job before job submission
- add submitter's certificate and the signatures from previous step at job submission
- verify signatures of all files at each folder before schedule the submitted job with submitter's certificate

The above implementation does not verify the submitter's certificate against the root CA of this project.  That is, the certificate chain is not built and is not verified.

This PR adds the missing part so the complete certificate chain is built and verified.

### Types of changes

- [ ] Documentation updated.


nvflare/private/fed/client/training_cmds.py
```diff
@@ -124,7 +125,8 @@ def process(self, req: Message, app_ctx) -> Message:
         if secure_train and not from_hub_site:
             workspace = Workspace(root_dir=engine.args.workspace, site_name=client_name)
             app_path = workspace.get_app_dir(job_id)
-            if not verify_folder_signature(app_path):
+            root_ca_path = os.path.join(workspace.get_startup_kit_dir(), "rootCA.pem")
```
YuanTingHsieh, line 128: not related to this PR,
For these "rootCA.pem", "server.crt", "client.crt" maybe we can define in a file these names as constants?

IsaacYangSLA: Thanks.  Will do that in the future PR.

yanchengnv (commented):
See my comments. There are also some alert in our test cases.

PR #1793, opened 2023-06-02
Fix deprecated

YuanTingHsieh:
### Description

To support 3.8/3.9/3.10 without introducing new dependencies, we write our own deprecated decorator

- Add deprecated
- Add unit test

### Types of changes

- [ ] Documentation updated.


nvkevlu (approved):
Thank you, this LGTM

PR #1796, opened 2023-06-05
Allow additional Config Options Part 1 : Config, ConfigLoader, ConfigFormat and ConfigFactory

chesterxgchen:
### Description
**Goals:**
In addition to the current JSON configuration for jobs, system are able to choose alternative configurations formats such as Pyhocon (python HOCON, considered it as superset of json) and Omegaconf ( considered it as superset of YAML)

This will allow user to add

Comment support ( inline, block)
nested configuration structure
ENV variable substitution
variable substitution
Multi-line strings ( for pyhocon)
optional overwrite from env. Variables (pyhocon)
Dictionary, list merge
Include directive support (pyhocon)
Default support API
dot notation for access nested elements

Non-goals:
Replace current configuration.

Design Changes:
1. new existing configuration or even constant changes. 
2. For file-configuration, we are going to search based on the order of ".json", ".conf", ".yml" extensions. The first found config file base name such <filename>.ext will be used as the configuration. 
3. We can mixed different configuration formats as one wishes. 
4. This PR introduce the Config (with JsonConfig, PyhoconConfig, OmegaConf), ConfigFormat, ConfigLoader, ConfigFactory. 
5. OmegaConf is not yet implemented, right now is just placeholder. 

for detailed discussion and design, please look at [design doc](https://docs.google.com/document/d/1t_1BjgQ6OCxw2LpaIReuXhtvIBnh7NDU7a0amGY23c8)

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/fuel/utils/config.py
+++ b/nvflare/fuel/utils/config.py
@@ -38,6 +38,16 @@ class ConfigFormat(Enum):
 
 
 class Config(ABC):
+    def __init__(self, fmt: ConfigFormat):
+        self.format = fmt
+
+    def get_format(self) -> ConfigFormat:
+        """returns the current config objects ConfigFormat
+        Returns:
+            return ConfigFormat
+        """
+        return self.format
+
     @abstractmethod
     def get_native_conf(self):
         """Return the original underline config object representation if you prefer to use it directly
@@ -47,15 +57,6 @@ class Config(ABC):
 
         Returns: Any,
             return native config objects
-        """
-
-        pass
-
-    @abstractmethod
-    def get_format(self):
-        """returns the current config objects ConfigFormat
-        Returns:
-            return ConfigFormat
         """
 
         pass
@@ -98,19 +99,21 @@ class Config(ABC):
 
 
 class ConfigLoader(ABC):
-    @abstractmethod
-    def load_config(
-        self, file_path: str, default_file_path: Optional[str] = None, overwrite_config: Optional[Dict] = None
-    ) -> Config:
-        """configuration from default_file_path will be the default config if specified
-            configuration from file_path will be the merge with default config overwrite the same key
-            configuration from overwrite_config if provided will be the merge with config overwrite the same key
+    def __init__(self, fmt: ConfigFormat):
+        self.format = fmt
 
+    def get_format(self) -> ConfigFormat:
+        """returns the current configLoader's ConfigFormat
+        Returns:
+            return ConfigFormat
+        """
+        return self.format
+
+    @abstractmethod
+    def load_config(self, file_path: str) -> Config:
+        """load configuration from config file: file_path
         Args:
             file_path: file path for configuration to be loaded
-            default_file_path: file path for default configuration
-            overwrite_config: dict config that will  overwrite the final config if provided
-
         Returns:
             return Config
         """
--- a/nvflare/fuel/utils/config_factory.py
+++ b/nvflare/fuel/utils/config_factory.py
@@ -25,6 +25,11 @@ from nvflare.fuel.utils.pyhocon_loader import PyhoconLoader
 
 class ConfigFactory:
     logger = logging.getLogger(__qualname__)
+    _fmt2Loader = {
+        ConfigFormat.JSON: JsonConfigLoader,
+        ConfigFormat.PYHOCON: PyhoconLoader,
+        ConfigFormat.OMEGACONF: OmegaConfLoader,
+    }
 
     @staticmethod
     def search_config_format(
@@ -54,12 +59,11 @@ class ConfigFactory:
         ext2fmt_map = ConfigFormat.config_ext_formats()
         for search_dir in search_dirs:
             logger.debug(f"search file:{file_basename} basename, search dirs = {search_dirs}")
-            for name in os.listdir(search_dir):
-                for ext in ext2fmt_map:
-                    fmt = ext2fmt_map[ext]
-                    file = f"{file_basename}{ext}"
-                    if file == name:
-                        return file
+            for ext in ext2fmt_map:
+                fmt = ext2fmt_map[ext]
+                file = os.path.join(search_dir, file_basename, ext)
+                if os.path.isfile:
+                    return fmt, file
 
         return None, None
 
@@ -101,13 +105,4 @@ class ConfigFactory:
         """
         if config_format is None:
             return None
-        if config_format == ConfigFormat.JSON:
-            return JsonConfigLoader()
-        elif config_format == ConfigFormat.OMEGACONF:
-            return OmegaConfLoader()
-        elif config_format == ConfigFormat.PYHOCON:
-            return PyhoconLoader()
-        else:
-            raise NotImplementedError(
-                f"configuration format {config_format.name} with file ext {config_format.value} is not implemented"
-            )
+        return ConfigFactory._fmt2Loader.get(config_format)
--- a/nvflare/fuel/utils/json_config_loader.py
+++ b/nvflare/fuel/utils/json_config_loader.py
@@ -22,15 +22,12 @@ from nvflare.security.logging import secure_format_exception
 
 class JsonConfig(Config):
     def __init__(self, conf: Dict, file_path: Optional[str] = None):
+        super(JsonConfig, self).__init__(ConfigFormat.JSON)
         self.conf = conf
-        self.format = ConfigFormat.JSON
         self.file_path = file_path
 
     def get_native_conf(self):
         return self.conf
-
-    def get_format(self):
-        return self.format
 
     def get_location(self) -> Optional[str]:
         return self.file_path
@@ -44,20 +41,11 @@ class JsonConfig(Config):
 
 class JsonConfigLoader(ConfigLoader):
     def __init__(self):
-        self.format = ConfigFormat.JSON
+        super(JsonConfigLoader, self).__init__(fmt=ConfigFormat.JSON)
         self.logger = logging.getLogger(self.__class__.__name__)
 
-    def load_config(
-        self, file_path: str, default_file_path: Optional[str] = None, overwrite_config: Optional[Dict] = None
-    ) -> Config:
-
+    def load_config(self, file_path: str) -> Config:
         conf_dict = self._from_file(file_path)
-        if default_file_path:
-            default_conf_dict = self._from_file(default_file_path)
-            self._dict_merge(default_conf_dict, conf_dict)
-        if overwrite_config:
-            self._dict_merge(conf_dict, overwrite_config)
-
         return JsonConfig(conf_dict, file_path)
 
     def load_config_from_str(self, config_str: str) -> Config:
@@ -78,10 +66,3 @@ class JsonConfigLoader(ConfigLoader):
             except Exception as e:
                 self.logger.error("Error loading config file {}: {}".format(path, secure_format_exception(e)))
                 raise e
-
-    def _dict_merge(self, target_dict: dict, overwrite_dict: dict):
-        for k, v in overwrite_dict.items():
-            if k in target_dict and isinstance(target_dict[k], dict) and isinstance(overwrite_dict[k], dict):
-                self._dict_merge(target_dict[k], overwrite_dict[k])
-            else:
-                target_dict[k] = overwrite_dict[k]
--- a/nvflare/fuel/utils/omegaconf_loader.py
+++ b/nvflare/fuel/utils/omegaconf_loader.py
@@ -11,18 +11,15 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from typing import Dict, Optional
 
 from nvflare.fuel.utils.config import Config, ConfigFormat, ConfigLoader
 
 
 class OmegaConfLoader(ConfigLoader):
     def __init__(self):
-        self.format = ConfigFormat.OMEGACONF
+        super(OmegaConfLoader, self).__init__(ConfigFormat.OMEGACONF)
 
-    def load_config(
-        self, file_path: str, default_file_path: Optional[str] = None, overwrite_config: Optional[Dict] = None
-    ) -> Config:
+    def load_config(self, file_path: str) -> Config:
         raise NotImplementedError
 
     def load_config_from_str(self, config_str: str) -> Config:
--- a/nvflare/fuel/utils/pyhocon_loader.py
+++ b/nvflare/fuel/utils/pyhocon_loader.py
@@ -23,15 +23,12 @@ from nvflare.fuel.utils.config import Config, ConfigFormat, ConfigLoader
 
 class PyhoconConfig(Config):
     def __init__(self, conf: ConfigTree, file_path: Optional[str] = None):
+        super(PyhoconConfig, self).__init__(ConfigFormat.PYHOCON)
         self.conf = conf
-        self.format = ConfigFormat.PYHOCON
         self.file_path = file_path
 
     def get_native_conf(self):
         return self.conf
-
-    def get_format(self):
-        return self.format
 
     def get_location(self) -> Optional[str]:
         return self.file_path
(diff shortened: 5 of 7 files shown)


nvflare/fuel/utils/config.py
```diff
@@ -0,0 +1,142 @@
...
+    @classmethod
+    def config_exts(cls) -> str:
+        search_sequence: List[ConfigFormat] = cls.search_order()
+        extensions = []
+        for fmt in search_sequence:
+            exts = cls.get_extensions(fmt.name)
+            if exts:
+                extensions.extend(exts)
+
+        return "|".join(iter(extensions))
```
YuanTingHsieh, line 62: why do we need this method?

nvflare/fuel/utils/config.py
```diff
@@ -0,0 +1,142 @@
...
+    @abstractmethod
+    def get_format(self):
+        pass
+
+    @abstractmethod
+    def get_location(self) -> Optional[str]:
+        # return the real file path of the reel configuration.
+        # For example, if initial config_fed_json.json was initially input file path. but the real file path is
+        # config_fed_json.conf. the get location will be return config_fed_json.conf
+        pass
```
YuanTingHsieh, line 83: Can we improve the docstring?

This method is not clear to me, as the implementation always return "self.file_path"

YuanTingHsieh: If get_location return None, what does that mean?

nvflare/fuel/utils/config.py
```diff
@@ -0,0 +1,142 @@
...
+    @abstractmethod
+    def to_conf(self, element: Dict) -> str:
+        pass
+
+    @abstractmethod
+    def get_int(self, key: str, default=None) -> Optional[int]:
+        pass
+
+    @abstractmethod
+    def get_float(self, key: str, conf=None, default=None) -> Optional[float]:
```
YuanTingHsieh, line 98: should the signature be consistent with other "get" method?

nvflare/fuel/utils/config.py
```diff
@@ -0,0 +1,142 @@
...
+    @abstractmethod
+    def get_conf(self):
+        # Return the original underline config object representation if you prefer to use it directly
+        # Pyhocon → ConfigTree
+        # JSON → Dict
+        # OMEGACONF → Conf
+        pass
+
+    @abstractmethod
+    def get_format(self):
```
YuanTingHsieh, line 75: should this return ConfigFormat?

nvflare/fuel/utils/config.py
```diff
@@ -0,0 +1,142 @@
...
+    def get_str(self, key: str, default=None) -> Optional[str]:
+        pass
+
+    @abstractmethod
+    def get_list(self, key: str, default=None) -> Optional[List]:
+        pass
+
+    @abstractmethod
+    def get_config(self, key: str, default=None):
+        pass
```
YuanTingHsieh, line 115: What is the difference "get_config" VS "get_conf" VS "get_location"?


Say we have a "config_fed_server.json"

get_config will return "config_fed_server.json"?
get_conf will return a dictionary
get_location will return "config_fed_server.json"?

so get_config and get_location is the same thing?

nvflare/fuel/utils/config_factory.py
```diff
@@ -0,0 +1,96 @@
...
+    @staticmethod
+    def search_config_format(
+        init_file_path, search_dirs: Optional[List[str]] = None
+    ) -> (Optional[ConfigFormat], Optional[str]):
+        # we ignore the original extension
+        logger = ConfigFactory.logger
+        if not search_dirs:
+            parent_dir = pathlib.Path(init_file_path).parent
+            search_dirs = [str(parent_dir)]
+        file_path = os.path.splitext(pathlib.Path(init_file_path).name)[0]
```
YuanTingHsieh, line 38: the comment "# we ignore the original extension" should be above this line ^

YuanTingHsieh: suggest renaming file_path to maybe file_name as it is not path anymore.

nvflare/fuel/utils/config.py
```diff
@@ -0,0 +1,142 @@
...
+    def ordered_search_extensions(cls) -> List:
+        search_sequence: List[ConfigFormat] = cls.search_order()
+        extensions = []
+        for fmt in search_sequence:
+            exts = cls.get_extensions(fmt.name)
+            if exts:
+                for ext in exts:
+                    extensions.append((fmt, ext))
+
+        return extensions
```
YuanTingHsieh, line 47: Why do we put these methods in ConfigFormat but not ConfigFactory?

nvflare/fuel/utils/json_config_loader.py
```diff
@@ -0,0 +1,144 @@
...
+        return json.dumps(element)
+
+    def get_int(self, key: str, default=None) -> Optional[int]:
+        value = self.conf.get(key, default)
+        try:
+            return int(value) if value is not None else None
+        except (TypeError, ValueError):
+            raise ConfigError("{key} has type '{type}' rather than 'int'".format(key=key, type=type(value).__name__))
+
+    def get_float(self, key: str, conf=None, default=None) -> Optional[float]:
```
YuanTingHsieh, line 52: method signature should be consistent with other get

nvflare/fuel/utils/json_config_loader.py
```diff
@@ -0,0 +1,144 @@
...
+        value = self.conf.get(key, default)
+        if value:
+            if isinstance(value, Dict):
+                return JsonConfig(value, self.file_path)
+            else:
+                raise ConfigError(
+                    "{key} has type '{type}' rather than 'Dict' or 'JsonConfig' ".format(
+                        key=key, type=type(value).__name__
+                    )
+                )
```
YuanTingHsieh, line 104: if value is Dict, things are good,
if value is of type "JsonConfig", this code will raise ConfigError

nvflare/fuel/utils/omegaconf_loader.py
```diff
@@ -0,0 +1,32 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from typing import Dict, Optional
+
+from nvflare.fuel.utils.config import Config, ConfigFormat, ConfigLoader
+
+
+class OmegaConfLoader(ConfigLoader):
```
YuanTingHsieh, line 19: maybe we add this file in another follow up PR.

nvflare/fuel/utils/pyhocon_loader.py
```diff
@@ -0,0 +1,128 @@
...
+        return self._convert_conf_item(self.conf)
+
+    def to_conf(self, element: Dict) -> str:
+        config = CF.from_dict(element)
+        return HOCONConverter.to_hocon(config)
+
+    def get_int(self, key: str, default=None) -> Optional[int]:
+        return self.conf.get_int(key, default)
+
+    def get_float(self, key: str, conf=None, default=None) -> Optional[float]:
```
YuanTingHsieh, line 50: suggest change it to be consistent with other get

nvflare/fuel/utils/config.py
```diff
@@ -0,0 +1,192 @@
...
+    @classmethod
+    def config_exts(cls) -> str:
+        search_sequence: List[ConfigFormat] = cls.search_order()
+        extensions = []
+        for fmt in search_sequence:
+            exts = cls.get_extensions(fmt.name)
+            if exts:
+                extensions.extend(exts)
+
+        return "|".join(iter(extensions))
```
yanchengnv, line 50: Why joining into a str, instead of returning the list?

nvflare/fuel/utils/config.py
```diff
@@ -0,0 +1,192 @@
...
+            element: dict
+
+        Returns: str string representation of the configuration in given format
+
+        """
+
+        pass
+
+    @abstractmethod
+    def get_int(self, key: str, default=None) -> Optional[int]:
```
yanchengnv, line 109: Why do we need these get_xxx() methods?

I thought the conf loader will only return a dict. It will be up to another layer (i.e. ConfigService) to do all these get_xxx() stuff.

nvflare/fuel/utils/config.py
```diff
@@ -0,0 +1,192 @@
...
+        pass
+
+    @abstractmethod
+    def get_list(self, key: str, default=None) -> Optional[List]:
+        pass
+
+    @abstractmethod
+    def get_config(self, key: str, default=None):
+        """
+           This method returns nvflare.fuel.utils.Config object
```
yanchengnv, line 131: Not sure I get this. What is nvflare.fuel.utils.Config object? Isn't it the same as the "self"?

nvflare/fuel/utils/config_factory.py
```diff
@@ -0,0 +1,134 @@
...
+        """
+        logger = ConfigFactory.logger
+        if not search_dirs:
+            parent_dir = pathlib.Path(init_file_path).parent
+            search_dirs = [str(parent_dir)]
+        # we ignore the original extension
+        file_basename = os.path.splitext(pathlib.Path(init_file_path).name)[0]
+        for search_dir in search_dirs:
+            logger.debug(f"search file:{file_basename} basename, search dirs = {search_dirs}")
+            for root, dirs, files in os.walk(search_dir):
```
yanchengnv, line 66: Should not walk thru all sub folders, otherwise a wrong file could be found. Instead, only look at the files directly in the dir.
Also the extension sequence is not right - see my comment before about non-default should be searched first before searching default cross ext types.

nvflare/fuel/utils/config_factory.py
```diff
@@ -0,0 +1,134 @@
...
+
+    @staticmethod
+    def has_config(init_file_path: str, search_dirs: Optional[List[str]] = None) -> bool:
+        _, real_file_path = ConfigFactory.search_config_format(init_file_path, search_dirs)
+        return real_file_path is not None
+
+    @staticmethod
+    def match_config(parent, init_file_path, match_fn) -> bool:
+        # we ignore the original extension
+        basename = os.path.splitext(init_file_path)[0]
```
yanchengnv, line 91: What is basename has "." in it?

yanchengnv (changes requested):
Looks good overall. Some changes needed:
- The search order of non-default vs. default
- I don't think get_xxx() methods should be part of the Config. Var handling should be handled by another layer. This PR should simply support loading of different formats and return dict. It's up to another layer how to determine the value of a var based on the config and other factors.

nvflare/fuel/utils/config.py
yanchengnv: You could also directly use the list in your error message. It will print quite nicely as a list.

nvflare/fuel/utils/config.py
yanchengnv: But I am not sure this is needed. If every loader simply returns a dict, the nested handling is only on dict. There is a dict_utils already. We could add more funcs there if needed. We don't need every Config type to do it.

nvflare/fuel/utils/config_factory.py
yanchengnv: Right, splitext doesn't know ".json.default" should be treated as one ext.
I wonder maybe we should change the treatment of default files. Instead of xxx.json.default, why don't we change it to default.xxx.json? This will make it much easier for you, right?
This is totally under our control. Of course we will need to change the Provision to produce default files with the new names.

yanchengnv: Also xxx.json.default is anti-pattern - technically the ext here is ".default". Many tools rely on ext to work.

nvflare/fuel/utils/config.py
YuanTingHsieh: ^ "the get_xxx just enforce the types or interpret the types."

So you mean for the same key say "num_of_rounds", we could use config.get_int("num_of_rounds") to get an integer representation and we could also use config.get_float("num_of_rounds") to get as float?

Then the ConfigService's get_xxx will try to 
1. get this key from environment variable
2. get this key from config_fed_server.json using "config.get_xxx"
3. get this key from resources.json using "config.get_xxx"
4. get this key from command line arguments

^ the order can be what we decide.
The first value return will be the value that is used.

Is this what we are thinking?

yanchengnv: Yes and no. Even though each config could provide its own get_xxx(), it's unnecessary since everything is loaded into dict already, and all vars could be retrieved from the dict, regardless where the dict is loaded from.

Secondly, as to how the var's value is determined, it could vary, depending on the nature of the var. What you laid out here is just one possible path. There are other cases.

Bottom line is that the var value determination logic is beyond a specific config loader.

nvflare/fuel/utils/config.py
```diff
@@ -0,0 +1,134 @@
...
+        Args:
+            config_str:
+        Returns:nvflare.fuel.utils.Config
+
+        """
+        raise NotImplementedError
+
+    def load_config_from_dict(self, config_dict: dict) -> Config:
+        """
+        Load Configuration based for given config dict.
```
YuanTingHsieh, line 127: Thanks for the update, the docstrings and API layer is clearer to me right now.

For the docstrings, can you follow the google docstring style here: https://google.github.io/styleguide/pyguide.html#383-functions-and-methods 

Thanks!

nvflare/fuel/utils/config.py
```diff
@@ -0,0 +1,134 @@
...
+        convert underline config object to dictionary
+        Returns:Dict
+
+        """
+        pass
+
+    @abstractmethod
+    def to_conf(self, element: Dict) -> str:
+        """
+            convert dict element to the str representation of the underline configuration.
```
YuanTingHsieh, line 82: Since it returns a string, suggest change method name to "to_conf_str" or "to_str" to be more clear?

tests/unit_test/fuel/utils/json_config_test.py
```diff
@@ -0,0 +1,86 @@
...
+        assert config is not None
+        conf = config.get_native_conf()
+        assert conf["a"]["a1"] == 200
+        assert conf.get("b") == 1
+        assert conf.get("c") == "hello"
+        assert conf.get("d") == [200, 400, 500]
+        assert conf.get("e1") == "Yes"
+        assert conf.get("e2") == "True"
+        assert conf.get("e3") == "NO"
+        assert conf.get("e4", None) is None
```
YuanTingHsieh, line 68: Would "assert conf == dicts" achieve the same effect?

nvflare/fuel/utils/config.py
```diff
@@ -0,0 +1,134 @@
...
+        configuration from overwrite_config if provided will be the merge with config overwrite the same key
+        Args:
+            file_path: file path for configuration to be loaded
+            default_file_path: file path for default configuration
+            overwrite_config: dict config that will  overwrite the final config if provided
+
+        Returns:nvflare.fuel.utils.Config
+        """
+
+        pass
```
YuanTingHsieh, line 112: Another approach:
What if we have "ConfigLoader" just load the config, so `def load_config(self, file_path)`.
And we let `ConfigFactory` to determine what config should overwrite the other?

tests/unit_test/fuel/utils/json_config_test.py
YuanTingHsieh: I re-read again and found your loader.load_config not only loads config from file it also loads from dict and it also merges them.
Given that implementation this test case is good.

nvflare/fuel/utils/config.py
```diff
@@ -0,0 +1,149 @@
...
+            config_str:
+        Returns:
+            return Config
+
+        """
+        raise NotImplementedError
+
+    def load_config_from_dict(self, config_dict: dict) -> Config:
+        """
+            Load Configuration based for given config dict.
```
YuanTingHsieh, line 141: Instead of starting in the next line, it should starts right after """.

For example:  `"""Loads Configuration based for given config dict.`

nvflare/fuel/utils/config.py
```diff
@@ -0,0 +1,149 @@
...
+
+        Returns:
+            return Config
+        """
+
+        pass
+
+    def load_config_from_str(self, config_str: str) -> Config:
+        """
+            Load Configuration based on the string representation of the underline configuration
```
YuanTingHsieh, line 128: Same here for the style.

nvflare/fuel/utils/config.py
```diff
@@ -0,0 +1,149 @@
...
+                ".yml.default": ConfigFormat.OMEGACONF,
+            }
+        )
+
+
+class Config(ABC):
+    @abstractmethod
+    def get_native_conf(self):
+        """
+            Return the original underline config object representation if you prefer to use it directly
```
YuanTingHsieh, line 44: docstring style

YuanTingHsieh (commented):
A few places docstring style needs to be corrected.

Otherwise the logic LGTM.

YuanTingHsieh (approved):
I still feel the "load_config" method of "ConfigLoader" should be keep simple.
I think it should not also do merge or replacing.

But we can discuss that further in part2 when I see more examples.

yanchengnv (changes requested):
The main issue is where the merge logic belongs. I don't think loaders should worry about it.

yanchengnv (approved):
LGTM

PR #1800, opened 2023-06-07
add METRICS_ONLY Transfer Type to FLModel

chesterxgchen:
Fixes # .

### Description

1) add TransferType.METRICS_ONLY and allows the FLMODEL to be used to send metrics with no model
2) make model optional in case of METRICS_ONLY

### Types of changes

- [ ] Documentation updated.


ZiyueXu77 (commented):
makes sense

holgerroth (approved):

PR #1806, opened 2023-06-12
Fix an issue in lighter utils CI tests

IsaacYangSLA:
### Description

During CI tests with pytest xdist, two test cases in lighter/util may fail due to damaged test data caused by race conditions.  This PR puts those two cases in a group so they will run one by one.

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh (approved):
Thanks, LGTM.
If we ever need to run them in parallel we can enhance it in the future

PR #1808, opened 2023-06-13
Use --dist loadgroup for pytest

YuanTingHsieh:
### Issue

Some tests can NOT be run in parallel, so we add "@pytest.mark.xdist_group".
However that only works if we run pytest using flag `--dist loadgroup`

### Description

- Add `--dist loadgroup` for running pytest.
(Follows: https://pytest-xdist.readthedocs.io/en/latest/distribution.html#running-tests-across-multiple-cpus)

### Types of changes

- [ ] Documentation updated.


nvkevlu (approved):
Thanks, LGTM

PR #1812, opened 2023-06-14
Make changes to fix MLflow experiment tracking

nvkevlu:
Make changes to fix MLflow experiment tracking:

- Fix issue with export path
- Fix step of TB send/receive
- Add log_text implementation for MLflow

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/fuel/flare_api/flare_api.py
+++ b/nvflare/fuel/flare_api/flare_api.py
@@ -62,12 +62,12 @@ _VALID_TARGET_TYPES = [TargetType.ALL, TargetType.SERVER, TargetType.CLIENT]
 
 class Session(SessionSpec):
     def __init__(self, username: str = None, startup_path: str = None, secure_mode: bool = True, debug: bool = False):
-        """Initializes a session with the NVFLARE system
+        """Initializes a session with the NVFLARE system.
 
         Args:
-            username: string of username to log in with
-            startup_path: path to the provisioned startup kit, which contains endpoint of the system
-            secure_mode: whether to run in secure mode or not
+            username (str): string of username to log in with
+            startup_path (str): path to the provisioned startup kit, which contains endpoint of the system
+            secure_mode (bool): whether to run in secure mode or not
         """
         assert isinstance(username, str), "username must be str"
         self.username = username
@@ -135,11 +135,7 @@ class Session(SessionSpec):
         self.overseer_agent = conf.overseer_agent
 
     def close(self):
-        """Close the session
-
-        Returns:
-
-        """
+        """Close the session."""
         self.api.close()
 
     def try_connect(self, timeout):
@@ -226,7 +222,7 @@ class Session(SessionSpec):
             raise JobNotFound(f"invalid job_id {job_id}")
 
     def clone_job(self, job_id: str) -> str:
-        """Create a new job by cloning a specified job
+        """Create a new job by cloning a specified job.
 
         Args:
             job_id: job to be cloned
@@ -244,14 +240,14 @@ class Session(SessionSpec):
         return job_id
 
     def submit_job(self, job_definition_path: str) -> str:
-        """Submit a predefined job to the NVFLARE system
+        """Submit a predefined job to the NVFLARE system.
 
         Args:
             job_definition_path: path to the folder that defines a NVFLARE job
 
         Returns: the job id if accepted by the system
 
-        If the submission fails, exception will be raised:
+        If the submission fails, an exception will be raised.
 
         """
         if not job_definition_path:
@@ -274,7 +270,7 @@ class Session(SessionSpec):
         return job_id
 
     def get_job_meta(self, job_id: str) -> dict:
-        """Get the meta info of the specified job
+        """Get the meta info of the specified job.
 
         Args:
             job_id: ID of the job
@@ -298,14 +294,15 @@ class Session(SessionSpec):
         name_prefix: str = None,
         reverse: bool = False,
     ) -> List[dict]:
-        """Get the job info from the server
+        """Get the job info from the server.
 
         Args:
-            detailed: True to get the detailed information for each job, False by default
-            limit: maximum number of jobs to show, with 0 or None to show all (defaults to None to show all)
-            id_prefix: if included, only return jobs with the beginning of the job ID matching the id_prefix
-            name_prefix: if included, only return jobs with the beginning of the job name matching the name_prefix
-            reverse: if specified, list jobs in the reverse order of submission times
+            detailed (bool): True to get the detailed information for each job, False by default
+            limit (int, optional): maximum number of jobs to show, with 0 or None to show all (defaults to None to show all)
+            id_prefix (str): if included, only return jobs with the beginning of the job ID matching the id_prefix
+            name_prefix (str): if included, only return jobs with the beginning of the job name matching the name_prefix
+            reverse (bool): if specified, list jobs in the reverse order of submission times
+
         Returns: a list of job metadata
 
         """
@@ -345,11 +342,10 @@ class Session(SessionSpec):
         return jobs_list
 
     def download_job_result(self, job_id: str) -> str:
-        """
-        Download result of the job
+        """Download result of the job.
 
         Args:
-            job_id: ID of the job
+            job_id (str): ID of the job
 
         Returns: folder path to the location of the job result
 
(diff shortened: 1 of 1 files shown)


nvflare/app_common/widgets/streaming.py
```diff
@@ -103,10 +103,11 @@ def handle_event(self, event_type: str, fl_ctx: FLContext):
...
-        global_step = kwargs.get("global_step", None)
-        if global_step:
+        if global_step is not None:
             if not isinstance(global_step, int):
                 raise TypeError(f"Expect global step to be an instance of int, but got {type(global_step)}")
+            kwargs[TrackConst.GLOBAL_STEP_KEY] = global_step
+        global_step = kwargs.get(TrackConst.GLOBAL_STEP_KEY, None)
         dxo = create_analytic_dxo(tag=tag, value=value, data_type=data_type, writer=self.get_writer_name(), **kwargs)
         with self.engine.new_context() as fl_ctx:
             send_analytic_dxo(self, dxo=dxo, fl_ctx=fl_ctx, event_type=self.event_type)
```
YuanTingHsieh, line 113: For this part, can you help adding unit tests to test the following:


1. to_dxo, from_dxo => make sure things are consistent
2. Adding different test cases to this "add" method, for example:
    - case 1: global_step = None, kwargs = None
    - case 2: global_step = 2, kwargs = None
    - case 3: global_step = None, kwargs = {"global_step": 3, "cool": 4}
    - case 4 (should not happen): global_step = 10, kwargs = {"global_step": 3, "extra_stuff": "hello"}

nvflare/app_opt/tracking/mlflow/mlflow_writer.py
```diff
@@ -79,21 +79,31 @@ def log_metrics(self, metrics: Dict[str, float], step: Optional[int] = None) ->
...
 
+    def log_text(self, text: str, artifact_file_path: str) -> None:
+        """Log text as an artifact under the current run.
+
+        Args:
+            text (str): String of text to log.
+            artifact_file_path (str): The run-relative artifact file path in posixpath format
+                to which the text is saved (e.g. “dir/file.txt”).
+        """
+        self.sender.add(tag="text", value=text, data_type=AnalyticsDataType.TEXT, path=artifact_file_path)
```
YuanTingHsieh, line 97: Can you help add these as test cases for the "sender add method"?

(log_text, set_tag, log_metrics) just make sure they all work.

chesterxgchen (changes requested):
few minor changes

nvflare/app_opt/tracking/mlflow/mlflow_receiver.py
```diff
@@ -219,8 +219,12 @@ def save(self, fl_ctx: FLContext, shareable: Shareable, record_origin: str):
             return
 
         if data.data_type == AnalyticsDataType.TEXT:
-            # not currently supported
-            pass
+            mlflow_client = self.get_mlflow_client(record_origin)
+            if not mlflow_client:
+                raise RuntimeError(f"mlflow client is None for site {record_origin}.")
+            run_id = self.get_run_id(record_origin)
+            if data.kwargs.get("path"):
```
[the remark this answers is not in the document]
nvkevlu: Done

nvflare/app_opt/tracking/mlflow/mlflow_writer.py
nvkevlu: Looked into this, and I think that to do this right, there may need to be additional infrastructure around mocking the engine to test for events firing as expected. The scope of this is a bit larger than I expected at first, so maybe we can do this in a future PR.

nvflare/app_common/widgets/streaming.py
nvkevlu: Looked into this, and I think that to do this right, there may need to be additional infrastructure around mocking the engine to test for events firing as expected. The scope of this is a bit larger than I expected at first, so maybe we can do this in a future PR.

YuanTingHsieh: @nvkevlu agree on the event firing part, that mechanism we can add later to the unit test framework.

For now, maybe we just add unit test up to the point of "create_analytic_dxo".
Just want to make sure these if-else statements including global_steps, kwargs is working as we expected.

Something like:

```

def mock_add(tag: str, value, data_type: AnalyticsDataType, global_step: Optional[int] = None, **kwargs):
    kwargs = kwargs if kwargs else {}
    if global_step is not None:
        if not isinstance(global_step, int):
            raise TypeError(f"Expect global step to be an instance of int, but got {type(global_step)}")
        kwargs[TrackConst.GLOBAL_STEP_KEY] = global_step
    global_step = kwargs.get(TrackConst.GLOBAL_STEP_KEY, None)
    dxo = create_analytic_dxo(tag=tag, value=value, data_type=data_type, writer=self.get_writer_name(), **kwargs)
    return dxo

class TestAnalyticsSender:
    def test_add(x, y, z):
        dxo = mock_add(x, y, z)
        assert dxo == expected_dxo
            
```

chesterxgchen (dismissed):

tests/unit_test/app_common/widgets/streaming_test.py
```diff
@@ -96,3 +98,93 @@ def test_invalid_send_analytic_dxo(self, comp, dxo, fl_ctx, expected_error, expe
...
+class TestAnalyticsSender:
+    @pytest.mark.parametrize(
+        "tag,value,data_type,global_step,kwargs,expected_dxo_data_kind,expected_dxo_data,expected_dxo_meta",
+        ANALYTICS_SENDER_TEST_CASES,
+    )
+    def test_add(
+        self, tag, value, data_type, global_step, kwargs, expected_dxo_data_kind, expected_dxo_data, expected_dxo_meta
+    ):
+        print(kwargs)
+        dxo = mock_add(tag=tag, value=value, data_type=data_type, global_step=global_step, kwargs=kwargs)
```
YuanTingHsieh, line 187: Thanks for adding the test.

Can you add test cases like:

```dxo = mock_add(tag=tag, value=value, data_type=data_type, global_step=global_step, **kwargs)```

Because kwargs should be any random dictionary

For example:
```
def cool_f(x, *args, **kwargs):
    if args:
        print(f"args is {args}")

    if kwargs:
        print(f"kwargs is {kwargs}")

input_dict = {"global_step": 100}

cool_f(x=1, a=3, b=10)
cool_f(x=2, global_step=100)
cool_f(x=5, **input_dict)

```

File: nvflare/app_common/widgets/streaming.py [hunk 5]
```diff
@@ -102,11 +102,12 @@ def handle_event(self, event_type: str, fl_ctx: FLContext):
             self.engine = fl_ctx.get_engine()
 
     def add(self, tag: str, value, data_type: AnalyticsDataType, global_step: Optional[int] = None, **kwargs):
-        kwargs = kwargs if kwargs else {}
-        global_step = kwargs.get("global_step", None)
-        if global_step:
+        kwargs = kwargs["kwargs"] if kwargs and kwargs["kwargs"] else {}
```
YuanTingHsieh, line 105: I think should be something like: `add(tag, value, data_type, global_step, step=10, hello=100)`
Not `add(tag, value, data_type, global_step, kwargs={"step":10, "hello": 100})`?

what do you think? which call approach you prefer?

chesterxgchen (commented):
Need to review the kwargs

nvflare/app_common/widgets/streaming.py
nvkevlu: I just changed this to make the test work, but it was because I didn't add the test case the right way. I am changing this back now.

tests/unit_test/app_common/widgets/streaming_test.py
nvkevlu: should be fixed now

tests/unit_test/app_common/widgets/streaming_test.py
```diff
@@ -96,3 +98,113 @@ def test_invalid_send_analytic_dxo(self, comp, dxo, fl_ctx, expected_error, expe
...
             create_analytic_dxo(tag=tag, value=value, data_type=data_type, step=step, writer=LogWriterName.TORCH_TB)
+
+
+def mock_add(tag: str, value, data_type: AnalyticsDataType, global_step: Optional[int] = None, **kwargs):
+    kwargs = kwargs if kwargs else {}
+    if global_step is not None:
+        if not isinstance(global_step, int):
+            raise TypeError(f"Expect global step to be an instance of int, but got {type(global_step)}")
+        kwargs[TrackConst.GLOBAL_STEP_KEY] = global_step
+    global_step = kwargs.get(TrackConst.GLOBAL_STEP_KEY, None)
```
YuanTingHsieh, line 109: Looks like we can remove this line as we are not using "global_step"

nvflare/app_common/widgets/streaming.py
YuanTingHsieh: Got it, thanks, LGTM now.

YuanTingHsieh (approved):
Thanks for the fixes and tests , looks good to me

PR #1816, opened 2023-06-16
Allow Alternative Config Option Part 2: Adding OmegaConf loader and Config

chesterxgchen:
### Description
This is part 2 the PR :  https://github.com/NVIDIA/NVFlare/pull/1796
In this PR, we made the following changes
1) refactoring the Config
2) add OmegaConfConfig and OmegaConf Loader (https://omegaconf.readthedocs.io/en/2.3_branch/) 
3) add unit tests 
4) add dependency 
5) add License for omegaconf. 
 

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/tests/unit_test/fuel/utils/pyhocon_config_test.py
+++ b/tests/unit_test/fuel/utils/pyhocon_config_test.py
@@ -11,7 +11,6 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-import json
 
 import pytest
 from pyhocon import ConfigFactory as CF


nvflare/fuel/utils/omegaconf_loader.py
```diff
@@ -11,19 +11,43 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+from typing import Dict, Optional
+
+from omegaconf import DictConfig, OmegaConf
```
yhwen, line 16: We may consider package this omegaconf_loader.py separately from the "main" NVFlare. This adds another additional third party dependency to NVFlare, which is a down side. This kind of "extra" to NVFlare could be endless.

YuanTingHsieh: We can discuss about this at team meeting.

As to how to add to optional dependencies, we add the code to "app_opt" (so they are not FUEL).
And we can either change one item or add a new entry in setup.cfg: https://github.com/NVIDIA/NVFlare/blob/dev/setup.cfg#L37

setup.cfg
```diff
@@ -33,6 +33,7 @@ install_requires =
     docker>=6.0
     websockets>=10.4
     pyhocon
+    omegaconf
```
YuanTingHsieh, line 36: As @yhwen mentioned, if we add them here, pyhocon and omegaconf becomes mandatory.
We should discuss on the pros and cons to make them mandatory.

nvflare/fuel/utils/config.py
```diff
@@ -38,8 +38,10 @@ def config_ext_formats(cls):
 
 
 class Config(ABC):
```
YuanTingHsieh, line 40: As our config/config service codes are growing larger, does it make sense to move it outside of `nvflare/fuel/utils`?

nvflare/fuel/utils/config.py
```diff
@@ -118,8 +120,6 @@ def load_config(self, file_path: str) -> Config:
             return Config
         """
 
-        pass
-
```
YuanTingHsieh, line 122: just curious, is having this one line `pass` clearer or not having it?
we have some places also got this pattern, we can update them as well.

tests/unit_test/fuel/utils/omegaconf_config_test.py
```diff
@@ -0,0 +1,88 @@
...
+        assert conf.c == "hi"
+        assert conf.d == [1, 2]
+        assert conf.get("e4", None) is None
+
+        a_dict = OmegaConf.to_container(conf.a, resolve=True)
+        assert a_dict == {"a1": 1, "a2": 2}
+        config = loader.load_config_from_dict(dicts)
+        assert config.get_format() == ConfigFormat.OMEGACONF
+        assert config is not None
+        assert config.to_dict() == dicts
```
YuanTingHsieh, line 83: it might be clearer if we can separate these into different test methods
like `test_load_config`, `test_load_config_from_dict`, `test_load_config_from_str`

nvflare/fuel/utils/config.py
```diff
@@ -59,43 +60,44 @@ def get_native_conf(self):
...
         Returns:
             return None if the config is not from file else return file path
 
         """
-        pass
+        return self.file_path
 
     @abstractmethod
-    def to_dict(self) -> Dict:
+    def to_dict(self, resolve: Optional[bool] = True) -> Dict:
```
YuanTingHsieh, line 75: if only OmegaConf need this "resolve" should we add it into the base class definition?

setup.cfg
```diff
@@ -46,20 +45,28 @@ SKLEARN =
...
+    pyhocon
+    omegaconf
 app_opt =
     %(HE)s
     %(PSI)s
     %(PT)s
     %(SKLEARN)s
     %(TRACKING)s
+core_opt =
+    %(CONFIG)s
```
YuanTingHsieh, line 58: since we put it inside "fuel_opt" should we just call this "fuel_opt"?

nvflare/fuel_opt/README.md
```diff
@@ -0,0 +1,31 @@
...
+That means there are dependencies are not required to be installed when user type
+ ```
+     pip install nvflare
+ ```
+* you will need to install the required python package separately or use
+
+```
+  pip install nvflare[core_opt]
+```
+to install optional dependency for "core" module.
```
YuanTingHsieh, line 13: Since we put it inside fuel_opt, should we call it fuel_opt?

nvflare/fuel_opt/README.md
```diff
@@ -0,0 +1,31 @@
...
+* to install optional dependency for application level module, use
+
+```
+  pip install nvflare[apt_opt]
+```
+to install all optional dependency, use
+
+```
+  pip install nvflare[all]
+```
```
YuanTingHsieh, line 31: this information is good to have.
but should we put it here or the README.md in the root ?

YuanTingHsieh (approved):
Thanks LGTM

YuanTingHsieh (approved):

YuanTingHsieh (approved):

PR #1820, opened 2023-06-21
Add missing cert chain verification on submitted job signatures

IsaacYangSLA:
### Description

This PR is created from cherry-picking 77daca3a2a2e99efaba4cb354a0f56f942a1718b

Add missing cert chain verification on submitted job signatures
Add unit test on verify_folder_signature

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/private/fed/client/training_cmds.py
+++ b/nvflare/private/fed/client/training_cmds.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import json
+import os
 from typing import List
 
 from nvflare.apis.workspace import Workspace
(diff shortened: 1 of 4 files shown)


nvflare/private/fed/client/training_cmds.py
```diff
@@ -117,6 +119,16 @@ def process(self, req: Message, app_ctx) -> Message:
         if not job_meta:
             return error_reply("missing job meta")
 
+        kv_list = parse_vars(engine.args.set)
+        secure_train = kv_list.get("secure_train", True)
+        from_hub_site = job_meta.get(JobMetaKey.FROM_HUB_SITE.value)
+        if secure_train and not from_hub_site:
+            workspace = Workspace(root_dir=engine.args.workspace, site_name=client_name)
+            app_path = workspace.get_app_dir(job_id)
```
YuanTingHsieh, line 127: Some of these lines are from HUB PR, according to @chesterxgchen we are not including HUB changes into 2.3

So we need to undo some of these lines, we can refer to https://github.com/NVIDIA/NVFlare/commit/77daca3a2a2e99efaba4cb354a0f56f942a1718b

YuanTingHsieh (changes requested):
Need to remove HUB new codes.

nvflare/private/fed/client/training_cmds.py
IsaacYangSLA: Got it.  Just force pushed one more commit.

YuanTingHsieh (approved):
Thanks, LGTM

PR #1829, opened 2023-06-27
Add test driver with FLARE API

nvkevlu:
Still debugging some tests, currently using the underlying api of FLARE API's session's do_command to get results that are the same as before, with wrapping in some cases to match previous FLAdminAPI response.

Ultimately, we will still need to determine whether to run all tests with both previous driver and FLARE API, or just some with this one, or which.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/tests/integration_test/src/action_handlers_flare_api.py
+++ b/tests/integration_test/src/action_handlers_flare_api.py
@@ -1,0 +1,175 @@
+# Copyright (c) 2022, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from __future__ import annotations
+
+import typing
+from nvflare.fuel.flare_api.flare_api import Session
+from nvflare.fuel.hci.client.api_status import APIStatus
+
+if typing.TYPE_CHECKING:
+    from .nvf_test_driver_flare_api import NVFFLAREAPITestDriver
+
+import time
+from abc import ABC, abstractmethod
+
+from tests.integration_test.src.utils import check_job_done_FLARE_API, run_admin_api_tests
+
+
+class _CmdHandler(ABC):
+    @abstractmethod
+    def handle(self, command_args: list, admin_controller: NVFFLAREAPITestDriver, admin_api: Session):
+        pass
+
+
+class _StartHandler(_CmdHandler):
+    def handle(self, command_args: list, admin_controller: NVFFLAREAPITestDriver, admin_api: Session):
+        if command_args[0] == "server":
+            if len(command_args) == 2:
+                # if server id is provided
+                server_ids = [command_args[1]]
+            else:
+                # start all servers
+                server_ids = list(admin_controller.site_launcher.server_properties.keys())
+            for sid in server_ids:
+                admin_controller.site_launcher.start_server(sid)
+        elif command_args[0] == "client":
+            if len(command_args) == 2:
+                # if client id is provided
+                client_ids = [command_args[1]]
+            else:
+                # start all clients
+                client_ids = list(admin_controller.site_launcher.client_properties.keys())
+            for cid in client_ids:
+                admin_controller.site_launcher.start_client(cid)
+        elif command_args[0] == "overseer":
+            admin_controller.site_launcher.start_overseer()
+        else:
+            raise RuntimeError(f"Target {command_args[0]} is not supported.")
+
+
+class _KillHandler(_CmdHandler):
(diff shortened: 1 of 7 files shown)


tests/integration_test/src/action_handlers_flare_api.py
```diff
@@ -0,0 +1,175 @@
...
+if typing.TYPE_CHECKING:
+    from .nvf_test_driver_flare_api import NVFFLAREAPITestDriver
+
+import time
+from abc import ABC, abstractmethod
+
+from tests.integration_test.src.utils import check_job_done_FLARE_API, run_admin_api_tests
+
+
+class _CmdHandler(ABC):
```
yhwen, line 30: why all of these classes named starting with "_"? I think it makes more sense to remove this and name them properly.

tests/integration_test/src/action_handlers_flare_api.py
```diff
@@ -0,0 +1,175 @@
...
+                client_ids = list(admin_controller.site_launcher.client_properties.keys())
+            for cid in client_ids:
+                admin_controller.site_launcher.start_client(cid)
+        elif command_args[0] == "overseer":
+            admin_controller.site_launcher.start_overseer()
+        else:
+            raise RuntimeError(f"Target {command_args[0]} is not supported.")
+
+
+class _KillHandler(_CmdHandler):
```
yhwen, line 62: better name it "StopHandler" ?

tests/integration_test/src/action_handlers_flare_api.py
```diff
@@ -0,0 +1,175 @@
...
+    def handle(self, command_args: list, admin_controller: NVFFLAREAPITestDriver, admin_api: Session):
+        pass
+
+
+class _TestDoneHandler(_CmdHandler):
+    def handle(self, command_args: list, admin_controller: NVFFLAREAPITestDriver, admin_api: Session):
+        admin_controller.test_done = True
+
+
+class _SubmitJobHandler(_CmdHandler):
```
yhwen, line 107: Probably it's better to build a AdminCommandHandler which can run different kinds of admin commands, instead of build a Handler for each individual admin command.

tests/integration_test/src/nvf_test_driver.py
```diff
@@ -317,6 +318,10 @@ def run_event_sequence(self, event_sequence):
                 stats = self._get_stats(target=TargetType.SERVER, job_id=self.job_id)
                 # update run_state
                 changed, run_state = _update_run_state(stats=stats, run_state=run_state, job_run_status=job_run_status)
+                if changed:
+                    print("****************************************************************")
+                    print("********************Updated RUN STATE***************************")
+                    print(run_state)
```
yhwen, line 324: Should use the assert to check and verify the "run_state" data. Should avoid using "print".

tests/integration_test/src/run_local_test.py
```diff
@@ -0,0 +1,14 @@
...
+
+if __name__ == "__main__":
+    test_driver = NVFFLAREAPITestDriver(
+    download_root_dir="/tmp/tmp_a5jpli4/admin/transfer", site_launcher=None, poll_period=1
+        )
+    test_driver.initialize_super_user(
+        workspace_root_dir="/tmp/tmp_a5jpli4", upload_root_dir="/workspace/repos/NVFlare_experiment2/tests/integration_test/data/apps", poc=True, super_user_name="admin"
+    )
+    
+    print(check_job_done_FLARE_API(job_id="f113af4e-b5bf-41f1-b773-a85a7c2d64af", admin_api=test_driver.super_admin_api))
```
yhwen, line 14: It's better to implement this test run "main" using the pytest. That would be easier to integrate with other testing / automatic system.

tests/integration_test/src/utils.py
```diff
@@ -197,6 +201,19 @@ def check_client_status_ready(response: dict) -> bool:
 
     return True
 
+def check_client_status_ready_FLARE_API(response: dict) -> bool:
```
yhwen, line 204: better call this "is_client_status_ready()" ?

tests/integration_test/src/utils.py
```diff
@@ -233,6 +250,18 @@ def check_job_done(job_id: str, admin_api: FLAdminAPI):
                             return True
     return False
 
+def check_job_done_FLARE_API(job_id: str, admin_api: Session):
```
yhwen, line 253: change to "is_job_done()" ?

tests/integration_test/src/utils.py
```diff
@@ -233,6 +250,18 @@ def check_job_done(job_id: str, admin_api: FLAdminAPI):
...
+    job_meta = get_job_meta_FLARE_API(admin_api, job_id=job_id)
+    print(job_meta)
+    job_run_status = job_meta.get("status")
+
+    print(job_run_status)
```
yhwen, line 258: Should not use "print()".

tests/integration_test/src/utils.py
```diff
@@ -437,3 +466,21 @@ def ensure_admin_api_logged_in(admin_api: FLAdminAPI, timeout: int = 60):
...
+        while time.time() - start_time <= timeout:
+            if admin_api.api.is_ready():
+                login_success = True
+                break
+            time.sleep(0.2)
+
+        if not login_success:
+            print(f"FLARE API failed to log in within {timeout} seconds: {admin_api.api.fsm.current_state}.")
+        else:
+            print("FLARE API Admin successfully logged into server.")
```
yhwen, line 483: Should not use "print()".

PR #1850, opened 2023-07-10
Fix SNI passing for AdminAPI (#1849)

### LINKED ISSUE #1849: [BUG] AdminAPI not passing SNI when connecting to the nvflare-server
**Describe the bug**

In our setup with NVFlare we noticed that at some case admin shell wouldn't be able to connect to server, even though the client running in same exact environment / network would be able to connect to our server. After some debugging it turned boiled down to these: 
- In our setup we proxy requests to appropriate server based on [SNI](https://en.wikipedia.org/wiki/Server_Name_Indication)
- Proxy wasn't getting any SNI indication when admin client was connecting

**To Reproduce**

I created a [sample repo](https://github.com/xander-aphe-hatschi/nvflare-sni-repro-case) to obtain logs of the connection initiation for both admin and client.

If we take a look at [client logs](https://github.com/xander-aphe-hatschi/nvflare-sni-repro-case/blob/main/client-connecting.log) and [admin logs](https://github.com/xander-aphe-hatschi/nvflare-sni-repro-case/blob/main/admin-connecting.log) first item missing would be `TLS client extension "server name" (id=0) ...` missing from the admin logs

**Expected behavior**

`AdminAPI` class should pass `SNI` when connecting to service

 - OS: Ubuntu 22.04
 - Python Version 3.10.9
 - NVFlare Version 2.3.0

**Additional context**

There's https://github.com/NVIDIA/NVFlare/pull/1850 that seems to fix this. I might need help figuring out how to test it through in test though (manual testing passes).

(Marked as bug just because feels like client/admin should be on-par in terms of being able to connect with the same network setup)

xander-aphe-hatschi:
Fixes #1849 .

### Description

Add a server_hostname SNI to the socket that is created

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh:
Changes LGTM, @yanchengnv @IsaacYangSLA @nvidianz do you see any risk if we pass this SNI?

IsaacYangSLA:
Thanks for creating this PR to fix the issue, @xander-aphe-hatschi .   The PR also looks good to me.

nvidianz (approved):

PR #1861, opened 2023-07-18
New cell fix

IsaacYangSLA:
### Description

This PR prepares the NewCell class for upper layer application.  It does not replace existing cell yet as the applications still import Cell.  Once all features are ready, we can switch to NewCell by removing import Cell and the comment of import NewCell as Cell.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/fuel/f3/cellnet/new_cell.py
+++ b/nvflare/fuel/f3/cellnet/new_cell.py
@@ -1,0 +1,159 @@
+# Copyright (c) 2023, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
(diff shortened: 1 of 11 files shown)


nvflare/fuel/f3/cellnet/new_cell.py
```diff
@@ -0,0 +1,159 @@
...
+        self.logger = logging.getLogger(self.__class__.__name__)
+        self.register_blob_cb("*", "*", self._process_reply)  # this should be one-time registration
+
+    def __getattr__(self, func):
+        def method(*args, **kwargs):
+            if not hasattr(self.core_cell, func):
+                raise AttributeError(f"'{func}' not in core_cell.")
+            return getattr(self.core_cell, func)(*args, **kwargs)
+
+        return method
```
nvidianz, line 73: Should we check the attr is a method here?

nvidianz (approved):

PR #1862, opened 2023-07-18
Fixes an End-of-Stream bug

nvidianz:
Fixes FLARE-1480

### Description

This fixes a bug that causes streaming ending prematurely and causes FOBS error.

### Types of changes

- [ ] Documentation updated.


IsaacYangSLA (approved):
Thanks for the fix.  The last_chunk bool is also included in the elements of task.buffers.  It signals the end of entire stream.

PR #1868, opened 2023-07-23
Support Roche requirements and fixes

yanchengnv:
Fixes # .

This PR implements the following:
- Enhanced UnsafeJob processing by proactively aborting the job on all clients when UnsafeJob error is returned.
- Implemented event-based unsafe component detection and processing (see https://docs.google.com/document/d/1g07KTFBwcQS1vJjzz1icYCcZwWeTLmN3na1DauwPLYA/edit)
- Define gRPC's keepalive_time_ms to 120000 as default value. This is to make gRPC's keepalive more frequent to avoid the problem that the network provider reset TCP after idling too long.
- Improved cancelled task handling in AIO
- Improved logger creation by naming with qualified class name. This will make it easier to configure loggers to different log level.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/apis/fl_exception.py
+++ b/nvflare/apis/fl_exception.py
@@ -33,3 +33,9 @@ class UnsafeJobError(Exception):
     """Raised when a job is detected to be unsafe"""
 
     pass
+
+
+class UnsafeComponentError(Exception):
+    """Raised when configuration parsing error happens."""
+
+    pass
--- a/nvflare/apis/fl_exception.py
+++ b/nvflare/apis/fl_exception.py
@@ -36,6 +36,6 @@ class UnsafeJobError(Exception):
 
 
 class UnsafeComponentError(Exception):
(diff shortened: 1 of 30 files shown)


nvflare/apis/fl_exception.py
```diff
@@ -33,3 +33,9 @@ class UnsafeJobError(Exception):
     """Raised when a job is detected to be unsafe"""
 
     pass
+
+
+class UnsafeComponentError(Exception):
+    """Raised when configuration parsing error happens."""
+
+    pass
```
YuanTingHsieh, line 41: What about """Raised when a component in the configuration is detected to be unsafe"""?

yanchengnv: Changed as suggested.

SYangster (approved):
Passes test plan, LGTM

PR #1870, opened 2023-07-24
switch check status and check meta for submit_job result

chesterxgchen:
### Description

In case of submit_job failure ( job folder missing or etc.), the existing FLARE API check the meta info first, which will show user "missing meta from result". This is confusion and misleading, as user will check the the job folder and found the meta.json does exists.  not sure what's the real cause the failure. 

   This PR switch the order the check return status and check meta. The first, the failure status is checked and real cause will be tell user what's the cause. 

   After that, check the meta. 

### Types of changes

- [ ] Documentation updated.


yanchengnv:
This PR seems to make sense, but I'm not totally sure about its effect on other commands.
The "missing meta from result" means that the response from the server does not have the "meta" data (which contains structured info for the client to process). This has nothing to do with the "meta.json" of the job.

In case of submit_job, there might be a bug in the processor of this command on the server that somehow failed to include the meta info in response.

PR #1876, opened 2023-07-26
Refactor xgboost examples

SYangster:
Building upon @YuanTingHsieh 's work
- Separated DataLoader from Executor to allow the use of other datasets
- Modified examples to work with integration auto generate tests
- Added tensorboard support to xgboost histogram examples
- General cleanup

### Types of changes

- [ ] Documentation updated.


tests/integration_test/src/utils.py
```diff
@@ -333,6 +333,8 @@ def generate_test_config_yaml_for_example(
     output_yamls = []
     os.makedirs(OUTPUT_YAML_DIR, exist_ok=True)
     for job in os.listdir(example.jobs_root_dir):
+        if "base" in job:
+            continue
```
YuanTingHsieh, line 337: Locally user will run "prepare_job_config.sh" that's correct.

But The automatic test stage in the integration tests will actually use "base" job directory.
So we can't add these two lines.

examples/advanced/xgboost/histogram-based/jobs/base/app/config/config_fed_client.json
```diff
@@ -1,16 +1,15 @@
...
-  "components": [],
   "executors": [
     {
       "tasks": [
         "train"
       ],
       "executor": {
         "id": "Executor",
-        "path": "higgs_executor.FedXGBHistogramHiggsExecutor",
+        "name": "FedXGBHistogramExecutor",
```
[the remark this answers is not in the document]
SYangster: The base jobs themselves currently don't work

tests/integration_test/src/utils.py
SYangster: Update: current integration test framework needs some modifications to support these base jobs; to be fixed in the future

nvflare/app_opt/xgboost/histogram_based/executor.py
```diff
@@ -188,6 +239,13 @@ def train(self, shareable: Shareable, fl_ctx: FLContext, abort_signal: Signal) -
             verbose_eval=self.verbose_eval,
         )
 
+        engine = fl_ctx.get_engine()
+        if engine.client.overseer_agent:
+            sp = engine.client.overseer_agent.get_primary_sp()
+            if sp and sp.primary is True:
+                self._server_address = sp.name
+        self.log_info(fl_ctx, f"server address is {self._server_address}")
```
YuanTingHsieh, line 247: since we duplicated this code section, can you wrap this into a method and the re-use it here?

YuanTingHsieh: the duplicated code also exists in `def handle_event(self, event_type: str, fl_ctx: FLContext):`

YuanTingHsieh (approved):
Thanks LGTM

PR #1881, opened 2023-07-27
Support admin client handlers and submit_job with custom data

yanchengnv:
Fixes # .

### Description

- Added event handler support to FL Admin Client. This is needed for flexible function extension on admin client.
- Added event types for admin client for login and command execution.
- Support custom props added by event handlers and send to server when executing a command.
- Modified submit_job command processing on Server to save custom props (if defined) as part of the job meta.
- Modified client runner of FL Client to support tasks received through aux message. This is to support client workflow controller (another PR).
- Check deploy policy to abort job when dead job report is received from clients

### Types of changes

- [ ] Documentation updated.


nvflare/fuel/hci/client/cli.py
```diff
@@ -112,6 +114,10 @@ def __init__(
 
         self._get_login_creds()
 
+        event_handlers = [self]
+        if handlers:
+            event_handlers.extend(handlers)
```
yhwen, line 119: better to check handlers are instances of EventHandler?

nvflare/fuel/hci/client/event.py
```diff
@@ -0,0 +1,61 @@
+# Copyright (c) 2023, NVIDIA CORPORATION.  All rights reserved.
```
yhwen, line 1: we already have "event.py" in the nvflare/private for the server and client event firing. Better to rename this differently to avoid confusion?

nvflare/fuel/hci/client/event.py
```diff
@@ -0,0 +1,61 @@
...
+class EventPropKey:
+
+    MSG = "msg"
+    USER_NAME = "user_name"
+    CMD_NAME = "cmd_name"
+    CMD_CTX = "cmd_ctx"
+    CUSTOM_PROPS = "custom_props"
+
+
+class EventContext(SimpleContext):
```
yhwen, line 43: What's the difference between the CommandContext and EventContext? I feel it could be hard for user choose which one to use. Should we combine them?

nvflare/private/fed/client/client_runner.py
```diff
@@ -76,6 +76,14 @@ def add_component(self, comp_id: str, component: object):
             self.handlers.append(component)
 
 
+class TaskInfo:
```
yhwen, line 79: The TaskInfo needs to built into the Task class. They are directly tied together.

nvflare/apis/signal.py
```diff
@@ -16,31 +16,49 @@
...
+        self._value = value
+        self._trigger_time = None
+        self._triggered = False
+
+    @property
+    def triggered(self):
+        if self._triggered:
+            return True
+        if self._parent:
+            return self._parent.triggered
```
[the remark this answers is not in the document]
yanchengnv: This is a recursive call already since "triggered" is a method marked as "property".

nvflare/fuel/hci/client/api.py
```diff
@@ -456,6 +436,13 @@ def __init__(
         self.service_finder.start(self._handle_sp_address_change)
         self._start_session_monitor()
 
+    def fire_event(self, event_type: str, ctx: SimpleContext):
+        if self.debug:
+            print(f"DEBUG: firing event {event_type}")
```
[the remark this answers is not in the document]
yanchengnv: Currently admin side does not use logger. Not sure the best way. Let's discuss and fix it later in a separate PR.

nvflare/fuel/hci/client/api.py
```diff
@@ -486,7 +480,14 @@ def _handle_sp_address_change(self, host: str, port: int, ssid: str):
     def _try_auto_login(self):
         resp = None
         for i in range(self.auto_login_max_tries):
-            self.fire_session_event(SessionEventType.TRYING_LOGIN, "Trying to login, please wait ...")
+            try:
+                self.fire_session_event(EventType.TRYING_LOGIN, "Trying to login, please wait ...")
+            except Exception as ex:
+                print(f"exception handling event {EventType.TRYING_LOGIN}: {secure_format_exception(ex)}")
```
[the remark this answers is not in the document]
yanchengnv: Currently admin side does not use logger. Not sure the best way. Let's discuss and fix it later in a separate PR.

nvflare/fuel/hci/client/cli.py
yanchengnv: Already checked in parsing the config.

nvflare/fuel/hci/client/event.py
yanchengnv: I think this is fine. These two are for different sides and never will be used together.

nvflare/fuel/hci/client/event.py
yanchengnv: Big difference. CommandContext is only used to ensure multiple commands can be run concurrently, whereas EventContext is used when an event is fired. Events could be fired without a command (e.g. before login).

nvflare/private/fed/client/client_runner.py
yanchengnv: Removed TaskInfo. Was used for keeping and managing task instances. Now it's simplified.

nvflare/fuel/hci/client/api.py
```diff
@@ -579,7 +580,12 @@ def _monitor_session(self, interval):
             msg = f"exception occurred: {secure_format_exception(e)}"
 
         self.server_sess_active = False
-        self.fire_session_event(SessionEventType.SESSION_CLOSED, msg)
+        try:
+            self.fire_session_event(EventType.SESSION_CLOSED, msg)
+        except Exception as ex:
+            if self.debug:
+                print(f"exception occurred handling event {EventType.SESSION_CLOSED}: {secure_format_exception(ex)}")
```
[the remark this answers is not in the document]
yanchengnv: Let's address this in a separate PR.

nvflare/apis/fl_context.py
```diff
@@ -170,6 +170,19 @@ def get_prop(self, key, default=None):
             else:
                 return default
 
+    def get_custom_prop(self, key: str, default=None):
+        props = self.get_prop(ReservedKey.CUSTOM_PROPS)
+        if not props:
+            return default
+        return props.get(key, default)
```
yhwen, line 177: This is duplicated with the function defined in job_def.py.

yhwen (commented):
Need to resolve the conflicts with main branch before merging.

yhwen (approved):
Approve the PR, will make further changes to the client_runner for the client controller integration.

yhwen (approved):

PR #1892, opened 2023-08-01
add Mac specific setup

chesterxgchen:
### Description

add mac specific setup (removing HE and openmind.psi dependency)

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh:
Changes is good for current situation.

@IsaacYangSLA do you know if there is a cleaner way? 
Like pip install can determine the OS automatically.

But we can do that as enhancements.

YuanTingHsieh (dismissed):

YuanTingHsieh (approved):

PR #1902, opened 2023-08-04
Fix fl model utils

YuanTingHsieh:
### Description

Use explicit argument name instead of kwargs

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/app_common/utils/fl_model_utils.py
+++ b/nvflare/app_common/utils/fl_model_utils.py
@@ -85,7 +85,6 @@ class FLModelUtils:
         meta[MetaKey.CURRENT_ROUND] = fl_model.current_round
         meta[MetaKey.TOTAL_ROUNDS] = fl_model.total_rounds
 
-        dxo.meta.update(meta)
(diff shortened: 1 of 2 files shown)


nvflare/app_common/utils/fl_model_utils.py
```diff
@@ -104,36 +103,42 @@ def from_shareable(
...
+            meta[MetaKey.JOB_ID] = fl_ctx.get_job_id()
+            meta[MetaKey.SITE_NAME] = fl_ctx.get_identity_name()
+
+        result = FLModel(
+            params_type=params_type,
+            params=params,
+            metrics=metrics,
+            current_round=current_round,
+            total_rounds=total_rounds,
+            meta=meta,
```
yhwen, line 140: Should we treat the current_round and total_rounds part of the "meta"?

YuanTingHsieh: I am neutral on this, @holgerroth @chesterxgchen what do you think?

PR #1906, opened 2023-08-06
Improve FLARE lightning API

chesterxgchen:
### Description
Another version of the FLARE Lightning API. Different from prev. version

Prev. version
``` 
    flare.patch(LitAutoEncoder)
    cli = LightningCLI( ... )

    print("--- train new model ---")
    cli.trainer.fit(cli.model, datamodule=cli.datamodule)

    print("--- test new model ---")
    cli.trainer.test(ckpt_path="best", datamodule=cli.datamodule)

    print("--- test global model ---")
    test_result = cli.trainer.test(cli.model.get_fl_module(), datamodule=cli.datamodule)
    output_model = flare.FLModel(params=cli.model.cpu().state_dict(), metrics=test_result[0])
    flare.send(output_model)
```

New version
```
    cli = LightningCLI(
          ....
        trainer_defaults={"callbacks": [
            ImageSampler(),
            flare.Callback(send_trigger=flare.SendTrigger.AFTER_TRAIN_AND_TEST)
        ], "max_epochs": 1},
       .....

    )
   
    print("--- test global model ---")
    cli.trainer.test(cli.model, datamodule=cli.datamodule)

    print("--- train new model ---")
    cli.trainer.fit(cli.model, datamodule=cli.datamodule)

    print("--- test new model ---")
    cli.trainer.test(ckpt_path="best", datamodule=cli.datamodule)
```

The new API requires One place change ( besides import) and one extra test for global model than original standalong code

The receive model and send model are hidden . 

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/apis/controller_spec.py
+++ b/nvflare/apis/controller_spec.py
@@ -23,7 +23,7 @@ from nvflare.apis.signal import Signal
 
 from .client import Client
 from .fl_context import FLContext
-from .shareable import Shareable
+from .shareable import ReservedHeaderKey, Shareable
 
 
 class TaskCompletionStatus(Enum):
@@ -104,7 +104,6 @@ class Task(object):
 
         if not isinstance(data, Shareable):
             raise TypeError("data must be an instance of Shareable, but got {}.".format(type(data)))
-
         if operator and not isinstance(operator, dict):
             raise TypeError(f"operator must be a dict but got {type(operator)}")
 
@@ -112,6 +111,8 @@ class Task(object):
         self.data = data  # task data to be sent to client(s)
         self.operator = operator
         self.cb_lock = threading.Lock()
+
+        data.set_header(ReservedHeaderKey.TASK_NAME, name)
 
         if props is None:
             self.props = {}
@@ -524,27 +525,5 @@ class ControllerSpec(ABC):
         Args:
             completion_status: the TaskCompletionStatus of the task
             fl_ctx: the FL context
-        """
-        pass
-
-    def abort_task(self, task: Task, fl_ctx: FLContext):
-        """Asks all clients to abort the execution of the specified task.
-
-        Args:
-            task: the task to be aborted
-            fl_ctx: the FL context
-
-        """
-        pass
-
-    def abort_all_tasks(self, fl_ctx: FLContext):
-        """Asks clients to abort the execution of all tasks.
-
-        NOTE: the server should send a notification to all clients, regardless of whether the server
-        has any standing tasks.
-
-        Args:
-            fl_ctx: the FL context
-
         """
         pass
--- a/nvflare/apis/event_type.py
+++ b/nvflare/apis/event_type.py
@@ -28,7 +28,6 @@ class EventType(object):
     END_WORKFLOW = "_end_workflow"
     ABORT_TASK = "_abort_task"
     FATAL_SYSTEM_ERROR = "_fatal_system_error"
-    FATAL_TASK_ERROR = "_fatal_task_error"
     JOB_DEPLOYED = "_job_deployed"
     JOB_STARTED = "_job_started"
     JOB_COMPLETED = "_job_completed"
--- a/nvflare/apis/fl_component.py
+++ b/nvflare/apis/fl_component.py
@@ -103,16 +103,6 @@ class FLComponent(StatePersistable):
         fl_ctx.set_prop(FLContextKey.EVENT_DATA, reason, private=True, sticky=False)
         self.fire_event(EventType.FATAL_SYSTEM_ERROR, fl_ctx)
 
-    def task_panic(self, reason: str, fl_ctx: FLContext):
-        """Signals a fatal condition that could cause the current task (on Client) to end.
-
-        Args:
-            reason (str): The reason for panic.
-            fl_ctx (FLContext): FLContext information.
-        """
-        fl_ctx.set_prop(FLContextKey.EVENT_DATA, reason, private=True, sticky=False)
-        self.fire_event(EventType.FATAL_TASK_ERROR, fl_ctx)
-
     def handle_event(self, event_type: str, fl_ctx: FLContext):
         """Handles events.
 
--- a/nvflare/apis/fl_constant.py
+++ b/nvflare/apis/fl_constant.py
@@ -101,6 +101,7 @@ class ReservedKey(object):
     JOB_DEPLOY_DETAIL = "__job_deploy_detail__"
     FATAL_SYSTEM_ERROR = "__fatal_system_error__"
     JOB_IS_UNSAFE = "__job_is_unsafe__"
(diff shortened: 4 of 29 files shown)


nvflare/app_opt/lightning/fl_callbacks.py
```diff
@@ -0,0 +1,99 @@
...
+
+    def on_train_end(self, trainer, pl_module):
+        if self.output_fl_model:
+            self.output_fl_model.params = pl_module.cpu().state_dict()
+        else:
+            self.output_fl_model = flare.FLModel(params=pl_module.cpu().state_dict())
+
+    def on_test_start(self, trainer, pl_module):
+        if pl_module:
+            self._receive_update_model(pl_module)
```
YuanTingHsieh, line 48: Note that this will change the behavior of the original "test" method.
But since we are already changing the behavior of "fit" method.
Maybe it does not matter.
If users choose to attach/use our hook in their Trainer, their fit/train/test method won't be the same as before.

nvflare/app_opt/lightning/fl_callbacks.py
```diff
@@ -0,0 +1,99 @@
...
+            except Exception as e:
+                self.model_delivered = False
+
+
+def test_loop_run_decorator(loop, cb):
+    func = loop.run
+
+    def wrapper(*args, **kwargs):
+        if cb.model_delivered:
+            return func(*args, **kwargs)
```
YuanTingHsieh, line 89: By the time we run this decorator, which is in the "setup" method.
Can this model_delivered be True?
If not maybe we just remove this if statement

nvflare/app_opt/lightning/fl_callbacks.py
```diff
@@ -0,0 +1,99 @@
...
+        else:
+            metrics = func(*args, **kwargs)
+            if loop.trainer.state.fn == TrainerFn.TESTING and metrics:
+                if cb.output_fl_model is None:
+                    cb.output_fl_model = flare.FLModel(metrics=metrics[0])
+                elif not cb.output_fl_model.metrics:
+                    cb.output_fl_model.metrics = metrics[0]
+            return metrics
```
YuanTingHsieh, line 97: If we don't have the "if statement", maybe these codes can be moved into hook "on_test_end"?

nvflare/app_opt/lightning/fl_callbacks.py
```diff
@@ -0,0 +1,99 @@
...
+import pytorch_lightning as pl
+from pytorch_lightning.trainer.states import TrainerFn
+
+import nvflare.client as flare
+from nvflare.app_common.abstract.fl_model import FLModel
+from nvflare.client.config import ConfigKey
+from nvflare.client.constants import ModelExchangeFormat
+
+
+class SendTrigger(IntEnum):
```
YuanTingHsieh, line 12: Not sure if Python 3.8 has IntEnum.

That leads to a question:
Are we saying we only support Python 3.9+? if so we are good with this code.
IF not, we need to change this.

nvflare/app_opt/lightning/fl_callbacks.py
```diff
@@ -0,0 +1,142 @@
...
+    def on_fit_start(self, trainer, pl_module):
+        # receive the global model and update the local model with global model
+        # the 1st time test() or fit() is called.
+        self._receive_update_model(pl_module)
+
+    def on_train_end(self, trainer, pl_module):
+        if self.output_fl_model:
+            self.output_fl_model.params = pl_module.cpu().state_dict()
+        else:
+            self.output_fl_model = flare.FLModel(params=pl_module.cpu().state_dict())
```
YuanTingHsieh, line 64: We should call self._check_and_send() here as well?

Otherwise, in your example, 
the first test will receive global model and test on it.
the train will use the cache global model and train on it. (<- we should submit here)
the second test will test on the trained model and in the end submit the trained model.

I think we should not assume the second test will always be there.

nvflare/app_opt/lightning/fl_callbacks.py
```diff
@@ -0,0 +1,157 @@
...
+def test_loop_run_decorator(loop, cb):
+    func = loop.run
+
+    def wrapper(*args, **kwargs):
+        if cb.metrics_captured or cb.send_mode == SendTrigger.AFTER_TRAIN:
+            try:
+                return func(*args, **kwargs)
+            except BaseException as e:
+                print(traceback.format_exc())
+                raise e
```
YuanTingHsieh, line 136: I think we should just `return func(*args, **kwargs)` here without try except.

Users can do try except block on the outside when calling trainer.test

nvflare/app_opt/lightning/fl_callbacks.py
```diff
@@ -0,0 +1,157 @@
...
+            if cb.output_fl_model is None:
+                cb.output_fl_model = flare.FLModel(metrics=result_metrics)
+            elif not cb.output_fl_model.metrics:
+                cb.output_fl_model.metrics = result_metrics
+
+    def _extract_metrics_from_tensor(metrics):
+        result_metrics = {}
+        for key, t in metrics.items():
+            result_metrics[key] = t.item()
+        return result_metrics
```
YuanTingHsieh, line 155: what is the type for this input "metrics"?

nvflare/app_opt/lightning/fl_callbacks.py
```diff
@@ -0,0 +1,157 @@
...
+                raise e
+        else:
+            metrics = func(*args, **kwargs)
+            _capture_metrics(metrics)
+            cb.metrics_captured = True
+            return metrics
+
+    def _capture_metrics(metrics):
+        result_metrics = _extract_metrics_from_tensor(metrics[0])
+        if loop.trainer.state.fn == TrainerFn.TESTING and metrics:
```
YuanTingHsieh, line 145: the "result_metrics = _extract_metrics_from_tensor(metrics[0])" should be put inside the if statement

nvflare/app_opt/lightning/fl_callbacks.py
YuanTingHsieh: If we doing this here, we should also be doing this try-except block for the else part to be consistent?

nvflare/app_opt/lightning/fl_callbacks.py
```diff
@@ -0,0 +1,157 @@
...
+            try:
+                return func(*args, **kwargs)
+            except BaseException as e:
+                print(traceback.format_exc())
+                raise e
+        else:
+            metrics = func(*args, **kwargs)
+            _capture_metrics(metrics)
+            cb.metrics_captured = True
+            return metrics
```
YuanTingHsieh, line 141: I mean then this part also need to have try-except?

YuanTingHsieh (commented):
code-wise is good to me.
Let's wait for other's feedback on user experience.

PR #1910, opened 2023-08-07
Fixed the admin tool requires the resources path.

yhwen:
Fixes # .

Fixed a bug the admin tool requires the resources path error.

### Description

Made the resources folder in the admin tool optional.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/fuel/hci/client/config.py
+++ b/nvflare/fuel/hci/client/config.py
@@ -44,8 +44,10 @@ class FLAdminClientStarterConfigurator(JsonConfigurator):
             sys.path.append(custom_dir)
 
         admin_config_file_path = workspace.get_admin_startup_file_path()
+        config_files = [admin_config_file_path]
         resources_file_path = workspace.get_resources_file_path()
-        config_files = [admin_config_file_path, resources_file_path]
+        if resources_file_path:
+            config_files.append(resources_file_path)
 
         JsonConfigurator.__init__(
             self,


nvflare/fuel/hci/client/config.py
```diff
@@ -44,8 +44,10 @@ def __init__(self, workspace: Workspace):
             sys.path.append(custom_dir)
 
         admin_config_file_path = workspace.get_admin_startup_file_path()
+        config_files = [admin_config_file_path]
         resources_file_path = workspace.get_resources_file_path()
-        config_files = [admin_config_file_path, resources_file_path]
+        if resources_file_path:
+            config_files.append(resources_file_path)
```
YuanTingHsieh, line 50: Make it optional looks good to me.

@nvkevlu and @IsaacYangSLA do we have place (AdminAPI or Provision) that required this resource file?

nvkevlu: Not required by AdminAPI side at least, so looks good to me

IsaacYangSLA: Provision does not generate resource files for admin startup kits.  That is, admin startup kit's local folder is empty.  However, provision still created that empty folder because I remember admin failed to start if it couldn't find that folder.

yhwen: This PR will make that resources folder as an optional.

PR #1921, opened 2023-08-10
Improved error handling and fixed memory leak

nvidianz:
Fixes FLARE-1455

### Description

1. Improved error handling. Always terminate streams on both sides.
2. Fixed a memory leak caused by not releasing task entries.
3. Not sending ACK when last chunk is received

### Types of changes

- [ ] Documentation updated.


nvflare/fuel/f3/streaming/byte_streamer.py
```diff
@@ -190,21 +185,24 @@ def _stop_task(self, task: TxTask, error: StreamError = None, notify=True):
         else:
             # Result is the number of bytes streamed
             task.stream_future.set_result(task.offset)
-        task.stop = True
+
+        self.tx_task_map.pop(task.sid)
```
yhwen, line 189: Need to catch the exception before this to ensure this pop to execute (place this in the finally block), otherwise it will result to memory leak again. For example, the self.cell.fire_and_forget() may raise exception.

nvflare/fuel/f3/streaming/byte_receiver.py
```diff
@@ -140,6 +145,26 @@ def register_callback(self, channel: str, topic: str, stream_cb: Callable, *args
...
+                    {
+                        StreamHeaderKey.STREAM_ID: task.sid,
+                        StreamHeaderKey.DATA_TYPE: StreamDataType.ERROR,
+                        StreamHeaderKey.ERROR_MSG: str(error),
+                    }
+                )
+                self.cell.fire_and_forget(STREAM_CHANNEL, STREAM_ACK_TOPIC, task.origin, message)
+
+        task.eos = True
+        self.rx_task_map.pop(task.sid)
```
yhwen, line 166: Same as other one, need to catch exception for the potential exception in self.cell.fire_and_forget().

yanchengnv, line 166: It might be safer to do self.rx_task_map.pop(task.sid, None), in case the item is already gone when you try to pop it.

yanchengnv: Agreed. The whole thing should be placed in a "try/except" block and place the "pop" in "finally". Basically we need to guarantee that the task is removed from the map no matter what happens.

nvflare/fuel/f3/streaming/byte_streamer.py
yanchengnv: Why do we have two places to touch the map? This could be vulnerable to race conditions.
Consider using a single "task manager" to manage task's lifecycle. You could use a separate thread that monitors the task status. All other places simply update task status. For example, the "stop_task" simply marks the task to be "stopped", and it's up to the "task monitor" to remove it from the map. The task object doesn't have to be removed right away, as long as it is removed. This can avoid the race condition that multiple places try to manipulate the same task object and the map.

yanchengnv (changes requested):
Please see my suggestion about task management.

nvflare/fuel/f3/streaming/byte_receiver.py
nvidianz: pop is moved to top and protected by a lock.

nvflare/fuel/f3/streaming/byte_streamer.py
nvidianz: The sending/receiving sides look very similar. There are 2 different maps: tx_task_map, rx_task_map, each map only have one place that create/remove entries. The earlier implementation uses a separate thread to manage the maps. The only reason for that was some ACK may come after the task has ended. The ACKs are just ignored now so no need for a different thread.

nvflare/fuel/f3/streaming/byte_receiver.py
nvidianz: Done. Also added lock to map operations.

nvflare/fuel/f3/streaming/byte_streamer.py
```diff
@@ -82,7 +83,7 @@ def send(self, channel: str, topic: str, target: str, headers: dict, stream: Str
 
     def _transmit_task(self, task: TxTask):
 
-        while not task.stop:
+        while True:
```
YuanTingHsieh, line 86: How do we know if this is done?
Or this method never ends, we use task_future object to terminate this?

nvidianz: When it reaches the EOS (End of Stream), the method returns at line 92.

nvidianz:
> Please see my suggestion about task management.

@yanchengnv The task is just part of the stream. Its life-cycle matches stream's exactly. So not sure what kind of management we need here. In an earlier implementation, task may lives after stream to catch the last few ACKs, which always arrive after stream is gone. But there is no value so those ACKs are simply ignored. Now task is just a member of the stream.

nvflare/fuel/f3/streaming/byte_receiver.py
```diff
@@ -37,6 +37,7 @@
 MAX_OUT_SEQ_CHUNKS = 16
 # 1/4 of the window size
 ACK_INTERVAL = 1024 * 1024 * 4
+READ_TIMEOUT = 60
```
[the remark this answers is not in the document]
nvidianz: This is passed to event.wait() so it's seconds.

nvflare/fuel/f3/streaming/byte_streamer.py
```diff
@@ -172,6 +169,9 @@ def _transmit(self, task: TxTask, final=False):
         task.stream_future.set_progress(task.offset)
 
     def _stop_task(self, task: TxTask, error: StreamError = None, notify=True):
+        with self.map_lock:
+            self.tx_task_map.pop(task.sid, None)
```
YuanTingHsieh, line 173: @chesterxgchen these parts for memory leak.

Previous code did NOT pop these so the reference of task was kept.
Thus Task data was kept, thus memory not getting garbage collected.

nvflare/fuel/f3/streaming/byte_receiver.py
```diff
@@ -140,6 +146,29 @@ def register_callback(self, channel: str, topic: str, stream_cb: Callable, *args
 
         self.registry.set(channel, topic, Callback(stream_cb, args, kwargs))
 
+    def stop_task(self, task: RxTask, error: StreamError = None, notify=True):
+
+        with self.map_lock:
+            self.rx_task_map.pop(task.sid, None)
```
YuanTingHsieh, line 152: @chesterxgchen these parts for memory leak.

Previous code did NOT pop these so the reference of task was kept.
Thus Task data was kept, thus memory not getting garbage collected.

yanchengnv (approved):
LGTM

PR #1858, opened 2023-07-17
Site authentication

yhwen:
Fixes # .

### Description

Added site-specific authorization security enhancement functions. Customers can provide handler to listen to EventType.AUTHORIZE_COMMAND_CHECK and use the security data available in the FL_Context to build their own authorization functions before any command executed on a particular site. 

For example, customer can build the handler use the AUTHORIZE_COMMAND_CHECK to check if the job is authorized to run on a particular client. The customized authorization can apply against the "check_resources", or "deploy_app" commands.

Also added more system events which will be used by CC integration.

### Types of changes

- [x] Non-breaking change (fix or new feature that would not break existing functionality).
- [ ] Breaking change (fix or new feature that would cause existing functionality to change).
- [ ] New tests added to cover the changes.
- [ ] Quick tests passed locally by running `./runtest.sh`.
- [ ] In-line docstrings updated.
- [ ] Documentation updated.

PR #1926, opened 2023-08-11
CC SDK 1.0 integration

IsaacYangSLA:
### Description

This PR depends on PR #1858, mainly on the events.
This PR adds a confidential computing components, which can be included in the resources.json in local folder.  This component does the followings

- all participants retrieve the GPU attestation token (cc token) via NVIDIA CC SDK
- clients send their own cc tokens to server during client registration step, server stores all cc tokens
- before job is deployed, server verifies all client cc tokens to ensure policy is met
- also server sends all cc tokens to all clients so every client can verify if the entire NVFlare system meets client own cc policy
- server deploys the job only when every participant cc token meets every participant's policy
- server will set the job to CANT SCHEDULE status if that above condition does not meet

### Types of changes

- [ ] Documentation updated.


IsaacYangSLA:
Will rebase to main once #1858 is merged

nvflare/app_opt/confidential_computing/cc_manager.py
```diff
@@ -0,0 +1,234 @@
...
+    def handle_event(self, event_type: str, fl_ctx: FLContext):
+        if event_type == EventType.SYSTEM_BOOTSTRAP:
+            try:
+                err = self._prepare_for_attestation(fl_ctx)
+            except:
+                self.log_exception(fl_ctx, "exception in attestation preparation")
+                err = "exception occurred"
+            if err:
+                self.log_critical(fl_ctx, err, fire_event=False)
+                sys.exit(-1)
```
yhwen, line 108: when handling the events, the handler should not call sys.exit(-1) during the error. You should just raise exception, for example: UnsafeComponentError.

nvflare/app_opt/confidential_computing/cc_manager.py
```diff
@@ -0,0 +1,234 @@
...
+    def _not_authorize_job(self, reason: str, fl_ctx: FLContext):
+        job_id = fl_ctx.get_prop(FLContextKey.CURRENT_JOB_ID, "")
+        self.log_error(fl_ctx, f"Job {job_id} is blocked: {reason}")
+        fl_ctx.set_prop(key=FLContextKey.AUTHORIZATION_REASON, value=reason)
+        fl_ctx.set_prop(key=FLContextKey.AUTHORIZATION_RESULT, value=False)
+
+    def _block_job(self, reason: str, fl_ctx: FLContext):
+        job_id = fl_ctx.get_prop(FLContextKey.CURRENT_JOB_ID, "")
+        self.log_error(fl_ctx, f"Job {job_id} is blocked: {reason}")
+        fl_ctx.set_prop(key=FLContextKey.JOB_BLOCK_REASON, value=reason)
```
yhwen, line 234: Do you need to add this here?

fl_ctx.set_prop(key=FLContextKey.AUTHORIZATION_RESULT, value=False)

yhwen (changes requested):

nvflare/app_opt/confidential_computing/cc_manager.py
IsaacYangSLA: Got it.  Thanks.

nvflare/app_opt/confidential_computing/cc_manager.py
yhwen: Probably we can combine the JOB_BLOCK_REASON check with the AUTHORIZATION_RESULT solution. But we can do that in a future enhancement.

nvflare/app_opt/confidential_computing/cc_helper.py
```diff
@@ -0,0 +1,137 @@
...
+
+    mapping = {GPU: Devices.GPU, CPU: Devices.CPU, NIC: Devices.NIC, OS: Devices.OS, DPU: Devices.DPU}
+
+
+class Env:
+
+    TEST = "test"
+    LOCAL = "local"
+    AZURE = "azure"
+    GCP = "gcp"
```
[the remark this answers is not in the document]
IsaacYangSLA: This is directly from Attestation SDK https://github.com/NVIDIA/nvtrust/blob/main/guest_tools/attestation_sdk/src/nv_attestation_sdk/attestation.py

They define the IntEnum, but for component arguments in the json, we need strings.  Therefore, our codes have to map strings into the IntEnum so the SDK can take the arguments.

docs/user_guide/confidential_computing.rst
```diff
@@ -0,0 +1,102 @@
+.. _confidential_computing:
+
+######################
+Confidential Computing
```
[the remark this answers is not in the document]
IsaacYangSLA: You are right.  The term CC is too broad.  Let me change it to the suggested term.

nvflare/app_opt/confidential_computing/cc_manager.py
```diff
@@ -0,0 +1,234 @@
...
+        self.participant_cc_info = {}  # used by the Server to keep tokens of all clients
+
+    def handle_event(self, event_type: str, fl_ctx: FLContext):
+        if event_type == EventType.SYSTEM_BOOTSTRAP:
+            try:
+                err = self._prepare_for_attestation(fl_ctx)
+            except:
+                self.log_exception(fl_ctx, "exception in attestation preparation")
+                err = "exception in attestation preparation"
+            if err:
```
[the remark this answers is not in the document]
IsaacYangSLA: Finally seems a better control flow.  Let me change it.

yhwen (approved):

PR #1934, opened 2023-08-15
Fixed the recursive FLComponents creation.

yhwen:
Fixes # .

### Description

- Fixed the recursive FLComponent creation missing the event handling issue.
- Fixed the client job process missing the proper FLcomponent handlers issue.
- Server Runner set the workflow ID in the FLContext before the initialize_run(fl_ctx).

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/private/fed/client/client_json_config.py
+++ b/nvflare/private/fed/client/client_json_config.py
@@ -102,8 +102,8 @@ class ClientJsonConfigurator(FedJsonConfigurator):
             self.current_exe.executor = self.authorize_and_build_component(element, config_ctx, node)
             return
 
-    def authorize_and_build_component(self, config_dict, config_ctx, node):
-        t = super().authorize_and_build_component(config_dict, config_ctx, node)
+    def build_component(self, config_dict):
+        t = super().build_component(config_dict)
         if isinstance(t, FLComponent):
             self.handlers.append(t)
         return t
--- a/nvflare/private/fed/client/client_runner.py
+++ b/nvflare/private/fed/client/client_runner.py
@@ -406,8 +406,7 @@ class ClientRunner(FLComponent):
         """
         default_task_fetch_interval = self.default_task_fetch_interval
         self.log_debug(fl_ctx, "fetching task from server ...")
-        temp_fl_ctx = copy.copy(fl_ctx)
-        task = self.engine.get_task_assignment(temp_fl_ctx)
+        task = self.engine.get_task_assignment(fl_ctx)
 
         if not task:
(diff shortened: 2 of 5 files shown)


nvflare/private/fed/client/client_json_config.py
```diff
@@ -102,8 +102,8 @@ def process_config_element(self, config_ctx: ConfigContext, node: Node):
             self.current_exe.executor = self.authorize_and_build_component(element, config_ctx, node)
             return
 
-    def authorize_and_build_component(self, config_dict, config_ctx, node):
-        t = super().authorize_and_build_component(config_dict, config_ctx, node)
+    def build_component(self, config_dict):
+        t = super().build_component(config_dict)
```
YuanTingHsieh, line 106: since it is main branch,
should we be using `authorize_and_build_component`?

nvflare/private/fed/server/server_json_config.py
```diff
@@ -138,8 +138,8 @@ def _get_all_workflows_ids(self):
             ids.append(t.id)
         return ids
 
-    def authorize_and_build_component(self, config_dict, config_ctx, node):
-        t = super().authorize_and_build_component(config_dict, config_ctx, node)
+    def build_component(self, config_dict):
+        t = super().build_component(config_dict)
```
YuanTingHsieh, line 142: since it is main branch,
should we be using `authorize_and_build_component`?

nvflare/private/fed/client/client_runner.py
```diff
@@ -406,7 +406,8 @@ def fetch_and_run_one_task(self, fl_ctx) -> (float, bool):
         """
         default_task_fetch_interval = self.default_task_fetch_interval
         self.log_debug(fl_ctx, "fetching task from server ...")
-        task = self.engine.get_task_assignment(fl_ctx)
+        temp_fl_ctx = copy.copy(fl_ctx)
+        task = self.engine.get_task_assignment(temp_fl_ctx)
```
YuanTingHsieh, line 410: why do we need this temp_fl_ctx?

YuanTingHsieh (changes requested):

nvflare/private/fed/server/server_json_config.py
yhwen: same as above.

nvflare/private/fed/client/client_runner.py
yhwen: The idea is to use this throw away the data added into the fl_ctx during the get_task_assignment() events firing. But thinking more, we should not do that in the platform level, instead the user components should take care of them. I will change it back.

YuanTingHsieh (approved):

Issue #1329, opened 2023-02-09
Access admin command history after logout


holgerroth:
**Is your feature request related to a problem? Please describe.**
When not using the admin console for some time, one will get automatically logged out. After logging in again, the previous commands are not available anymore with the up-arrow key.

**Describe the solution you'd like**
Access to admin command history after logout/login.

**Describe alternatives you've considered**
n/a

**Additional context**
n/a

PR #1938, opened 2023-08-15
Re-add cli persistent history

SYangster:
### Description

- Re-add persistent command history to admin cli
- Uses readline module

### Types of changes

- [ ] Documentation updated.


nvidianz (dismissed):

YuanTingHsieh:
@SYangster Does this solve #1329?  if so we can link this PR to the issue

nvflare/fuel/hci/client/cli.py
```diff
@@ -133,6 +140,15 @@ def __init__(
             session_status_check_interval=1800,  # check server for session status every 30 minutes
             event_handlers=event_handlers,
         )
+
+        nvflare_dir = os.path.join(os.path.expanduser("~"), ".nvflare")
```
yhwen, line 144: probably it's better to place this history file in the admin tool workspace folder, instead of the account "~/.nvflare" folder.

nvflare/fuel/hci/client/cli.py
```diff
@@ -370,6 +389,15 @@ def _do_default(self, line):
...
             return True
 
+    def preloop(self):
+        if readline and os.path.exists(self.cli_history_file):
+            readline.read_history_file(self.cli_history_file)
+
+    def postcmd(self, stop, line):
+        if readline:
+            readline.write_history_file(self.cli_history_file)
+        return stop
```
YuanTingHsieh, line 399: Why are we using postcmd here?
Can we use postloop?

SYangster: If it crashes before exiting the loop, then postloop() won't be called and history won't be saved

YuanTingHsieh (dismissed):
Thanks, LGTM

yhwen (dismissed):

YuanTingHsieh (approved):

PR #1942, opened 2023-08-17
SFM Heartbeat Support

nvidianz:
Fixes FLARE-952 and FLARE-1438

### Description

1. Added heartbeat support to SFM. The heartbeat interval is controlled by comm parameter heartbeat_interval.
2. Disabled WAIT_TIL in Cellnet to get around the clock drift issue.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/fuel/f3/drivers/aio_grpc_driver.py
+++ b/nvflare/fuel/f3/drivers/aio_grpc_driver.py
@@ -257,7 +257,7 @@ class AioGrpcDriver(BaseDriver):
 
     @staticmethod
     def capabilities() -> Dict[str, Any]:
-        return {DriverCap.HEARTBEAT.value: False, DriverCap.SUPPORT_SSL.value: True}
+        return {DriverCap.SEND_HEARTBEAT.value: True, DriverCap.SUPPORT_SSL.value: True}
 
     async def _start_server(self, connector: ConnectorInfo, aio_ctx: AioContext, conn_ctx: _ConnCtx):
         self.connector = connector
--- a/nvflare/fuel/f3/drivers/aio_grpc_driver.py
+++ b/nvflare/fuel/f3/drivers/aio_grpc_driver.py
@@ -39,7 +39,7 @@ from .net_utils import MAX_FRAME_SIZE, get_address, get_tcp_urls, ssl_required
 GRPC_DEFAULT_OPTIONS = [
     ("grpc.max_send_message_length", MAX_FRAME_SIZE),
     ("grpc.max_receive_message_length", MAX_FRAME_SIZE),
-    ("grpc.keepalive_time_ms", 120000),
+    ("grpc.keepalive_time_ms", 300000),
     ("grpc.http2.max_pings_without_data", 0),
 ]
 
@@ -257,7 +257,7 @@ class AioGrpcDriver(BaseDriver):
 
     @staticmethod
     def capabilities() -> Dict[str, Any]:
-        return {DriverCap.HEARTBEAT.value: True, DriverCap.SUPPORT_SSL.value: True}
+        return {DriverCap.HEARTBEAT.value: False, DriverCap.SUPPORT_SSL.value: True}
 
     async def _start_server(self, connector: ConnectorInfo, aio_ctx: AioContext, conn_ctx: _ConnCtx):
         self.connector = connector
--- a/nvflare/fuel/f3/drivers/aio_grpc_driver.py
+++ b/nvflare/fuel/f3/drivers/aio_grpc_driver.py
@@ -39,7 +39,7 @@ from .net_utils import MAX_FRAME_SIZE, get_address, get_tcp_urls, ssl_required
 GRPC_DEFAULT_OPTIONS = [
     ("grpc.max_send_message_length", MAX_FRAME_SIZE),
     ("grpc.max_receive_message_length", MAX_FRAME_SIZE),
-    ("grpc.keepalive_time_ms", 300000),
+    ("grpc.keepalive_time_ms", 120000),
     ("grpc.http2.max_pings_without_data", 0),
 ]
 
--- a/nvflare/fuel/f3/drivers/aio_http_driver.py
+++ b/nvflare/fuel/f3/drivers/aio_http_driver.py
@@ -103,7 +103,7 @@ class AioHttpDriver(BaseDriver):
 
     @staticmethod
     def capabilities() -> Dict[str, Any]:
-        return {DriverCap.HEARTBEAT.value: True, DriverCap.SUPPORT_SSL.value: True}
+        return {DriverCap.HEARTBEAT.value: False, DriverCap.SUPPORT_SSL.value: True}
 
     def listen(self, connector: ConnectorInfo):
         self._event_loop(Mode.PASSIVE, connector)
@@ -156,13 +156,17 @@ class AioHttpDriver(BaseDriver):
             scheme = "wss"
         else:
             scheme = "ws"
-        async with websockets.connect(f"{scheme}://{host}:{port}", ssl=self.ssl_context, max_size=MAX_FRAME_SIZE) as ws:
+        async with websockets.connect(
+            f"{scheme}://{host}:{port}", ssl=self.ssl_context, ping_interval=None, max_size=MAX_FRAME_SIZE
+        ) as ws:
             await self._handler(ws)
 
     async def _async_listen(self, host, port):
         self.ssl_context = net_utils.get_ssl_context(self.connector.params, True)
 
-        async with websockets.serve(self._handler, host, port, ssl=self.ssl_context, max_size=MAX_FRAME_SIZE):
+        async with websockets.serve(
+            self._handler, host, port, ssl=self.ssl_context, ping_interval=None, max_size=MAX_FRAME_SIZE
+        ):
             await self.stop_event
 
     async def _handler(self, websocket):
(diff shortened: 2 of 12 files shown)


nvflare/fuel/f3/drivers/aio_grpc_driver.py
```diff
@@ -39,7 +39,7 @@
 GRPC_DEFAULT_OPTIONS = [
     ("grpc.max_send_message_length", MAX_FRAME_SIZE),
     ("grpc.max_receive_message_length", MAX_FRAME_SIZE),
-    ("grpc.keepalive_time_ms", 120000),
+    ("grpc.keepalive_time_ms", 300000),
```
YuanTingHsieh, line 42: What should be a suitable number for this?
@chesterxgchen @IsaacYangSLA Do you remember the azure timeout stuff, what is that magic number

nvflare/fuel/f3/drivers/aio_http_driver.py
```diff
@@ -103,7 +103,7 @@ def supported_transports() -> List[str]:
 
     @staticmethod
     def capabilities() -> Dict[str, Any]:
-        return {DriverCap.HEARTBEAT.value: True, DriverCap.SUPPORT_SSL.value: True}
+        return {DriverCap.HEARTBEAT.value: False, DriverCap.SUPPORT_SSL.value: True}
```
YuanTingHsieh, line 106: Why is "DriverCap.HEARTBEAT.value: False"?
Who is sending the heartbeat now?

nvflare/fuel/f3/drivers/aio_grpc_driver.py
```diff
@@ -257,7 +257,7 @@ def supported_transports() -> List[str]:
 
     @staticmethod
     def capabilities() -> Dict[str, Any]:
-        return {DriverCap.HEARTBEAT.value: True, DriverCap.SUPPORT_SSL.value: True}
+        return {DriverCap.HEARTBEAT.value: False, DriverCap.SUPPORT_SSL.value: True}
```
YuanTingHsieh, line 260: Why is "DriverCap.HEARTBEAT.value: False"?
I thought GRPC will send heartbeat itself?

nvflare/fuel/f3/drivers/aio_http_driver.py
```diff
@@ -156,13 +156,17 @@ async def _async_connect(self, host, port):
             scheme = "wss"
         else:
             scheme = "ws"
-        async with websockets.connect(f"{scheme}://{host}:{port}", ssl=self.ssl_context, max_size=MAX_FRAME_SIZE) as ws:
+        async with websockets.connect(
+            f"{scheme}://{host}:{port}", ssl=self.ssl_context, ping_interval=None, max_size=MAX_FRAME_SIZE
+        ) as ws:
```
YuanTingHsieh, line 161: Do we have to disable this ping?

YuanTingHsieh (changes requested):
Thanks for the PR.
Please resolve the conflicts.

nvflare/fuel/f3/drivers/aio_grpc_driver.py
nvidianz: When this is false, SFM will send its own heartbeat for the connection. By doing this, we have heartbeat in both layers. This makes it easier to control heartbeat interval across all drivers.

nvflare/fuel/f3/drivers/aio_http_driver.py
nvidianz: The HeartbeatMonitor in SFM is sending the heartbeat now.

nvflare/fuel/f3/drivers/aio_grpc_driver.py
nvidianz: Changed to leave the value at 120000 (2 min).

nvflare/fuel/f3/sfm/heartbeat_monitor.py
```diff
@@ -0,0 +1,69 @@
...
+
+        for sfm_conn in self.conns.values():
+
+            driver = sfm_conn.conn.connector.driver
+            caps = driver.capabilities()
+            if caps and caps.get(DriverCap.HEARTBEAT.value, False):
+                continue
+
+            if self.curr_time - sfm_conn.last_activity > self.interval:
+                sfm_conn.send_heartbeat(Types.PING)
```
yhwen, line 68: Do we want to expand this to allow the sending of heartbeat along with additional data, for example, data with additional status info?

nvflare/fuel/f3/sfm/heartbeat_monitor.py
```diff
@@ -0,0 +1,69 @@
...
+from nvflare.fuel.f3.sfm.sfm_conn import SfmConnection
+
+log = logging.getLogger(__name__)
+
+HEARTBEAT_TICK = 5
+DEFAULT_HEARTBEAT_INTERVAL = 60
+
+
+class HeartbeatMonitor(Thread):
+    def __init__(self, conns: Dict[str, SfmConnection]):
```
yhwen, line 31: This requires creating all the connections first, then create the HeartbeatMonitor. Do we have the use case of creating additional connection after the HeartbeatMonitor created? In that case, we need the function like add_connection()?

nvflare/fuel/f3/drivers/aio_grpc_driver.py
YuanTingHsieh: Can we use another key/flag to determine whether to send SFM heartbeat or not?
Otherwise it is confusing to users as AIO GRPC driver DOES have the ability to send hearbeat.

nvidianz: How about SEND_HEARTBEAT?

nvflare/fuel/f3/sfm/heartbeat_monitor.py
nvidianz: Right now, the HearbeatMonitor is created before all connections. The `conns` will be updated if new connection is created.

nvflare/fuel/f3/drivers/aio_grpc_driver.py
YuanTingHsieh: That name should be fine.

We just need to make sure that this flag is not a driver capability.

As it is only related to SFM.

nvflare/fuel/f3/sfm/heartbeat_monitor.py
```diff
@@ -0,0 +1,69 @@
...
+        log.debug("Heartbeat monitor stopped")
+
+    def _check_heartbeat(self):
+
+        for sfm_conn in self.conns.values():
+
+            driver = sfm_conn.conn.connector.driver
+            caps = driver.capabilities()
+            if caps and caps.get(DriverCap.HEARTBEAT.value, False):
+                continue
```
YuanTingHsieh, line 65: We should use the new flag here right?

nvidianz: We can change name but it will still be in this dictionary. We don't want create yet another dictionary for the driver's configuration.

nvflare/fuel/f3/sfm/heartbeat_monitor.py
```diff
@@ -0,0 +1,69 @@
...
+        log.debug("Heartbeat monitor stopped")
+
+    def _check_heartbeat(self):
+
+        for sfm_conn in self.conns.values():
+
+            driver = sfm_conn.conn.connector.driver
+            caps = driver.capabilities()
+            if caps and not caps.get(DriverCap.SEND_HEARTBEAT.value, False):
+                continue
```
YuanTingHsieh, line 65: Just to confirm:
So if "DriverCap.SEND_HEARTBEAT.value" is True, we will not call sfm_conn.send_heartbeat right?
This means our current aio_grpc, aio_http, aio_tcp, tcp drivers they will NOT call sfm_conn.send_heartbeat.

YuanTingHsieh: In the end, we should just use SFM heartbeats, SO drivers do not need DriverCap.SEND_HEARTBEAT anymore.
But that can be done in follow-up PRs.

nvflare/fuel/f3/sfm/heartbeat_monitor.py
YuanTingHsieh: In the end, we should just use SFM heartbeats, SO drivers do not need DriverCap.SEND_HEARTBEAT anymore.
But that can be done in follow-up PRs.

nvflare/fuel/f3/sfm/heartbeat_monitor.py
```diff
@@ -0,0 +1,69 @@
...
+
+HEARTBEAT_TICK = 5
+DEFAULT_HEARTBEAT_INTERVAL = 60
+
+
+class HeartbeatMonitor(Thread):
+    def __init__(self, conns: Dict[str, SfmConnection]):
+        Thread.__init__(self)
+
+        self.thread_name = "hb_mon"
```
YuanTingHsieh, line 34: minor thing, it might be good to name this with sfm_hb_mon, otherwise hb_mon is not that clear.

YuanTingHsieh (approved):
Thanks, changes are good.

We should gradually get rid of "DriverCap.SEND_HEARTBEAT" in the end and just do SFM heartbeats.

nvflare/fuel/f3/sfm/heartbeat_monitor.py
nvidianz: Thread name should be short. This is good enough.

PR #1946, opened 2023-08-20
make sure the Job CLI support multi-config formats

chesterxgchen:
Fixes # .

Fix few cases where the Job CLI only handle one specific configuration format

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/tests/unit_test/apis/impl/controller_test.py
+++ b/tests/unit_test/apis/impl/controller_test.py
@@ -608,9 +608,7 @@ class TestCallback(TestController):
 
         task_name_out, _, data = controller.process_task_request(client, fl_ctx)
 
-        expected = Shareable()
-        expected["_test_data"] = client_name
-        assert data["_test_data"] == expected["_test_data"]
+        assert data["_test_data"] == client_name
         controller.cancel_task(task)
         assert task.completion_status == TaskCompletionStatus.CANCELLED
         launch_thread.join()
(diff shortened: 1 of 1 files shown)


nvflare/utils/cli_utils.py
```diff
@@ -160,26 +161,57 @@ def is_dir_empty(path: str):
...
-    fmt = ConfigFormat.JSON if to_json else ConfigFormat.PYHOCON
-    ext = ConfigFormat.extensions(fmt)[0]
-    if dst_path.endswith(ext):
+def hocon_to_string(target_fmt: ConfigFormat, dst_config: ConfigTree):
+    if target_fmt == ConfigFormat.JSON:
+        return HOCONConverter.to_json(dst_config)
+    elif target_fmt == ConfigFormat.PYHOCON:
+        return HOCONConverter.to_hocon(dst_config)
+    elif target_fmt == ConfigFormat.OMEGACONF:
+        from nvflare.fuel_opt.utils.omegaconf_loader import OmegaConfLoader
```
IsaacYangSLA, line 170: Not sure if the import statement should be here.  The PyhoconConfig is at the top.

IsaacYangSLA (approved):
LGTM.  This PR mainly adds support of OMEGACONF.  I think the OmegaConfLoader import location is OK, but for consistency, it should be at the top.

nvflare/utils/cli_utils.py
```diff
@@ -20,6 +20,7 @@
 from pyhocon import ConfigTree, HOCONConverter
```
YuanTingHsieh, line 20: not restricted to this PR.
We have all these "utils" file all over the place and tools/cli is all over the place, maybe we want to clean them up later.

YuanTingHsieh (approved):
Some other old tools might need to be updated as well so they can support multiple config formats.

PR #1947, opened 2023-08-21
Rewrite double quote to heredocs to avoid bash/zsh issues

IsaacYangSLA:
### Description

Fix the quotes that do not work in bash/zsh.
Fix an issue that does not include script common header in cloud launch scripts.

### Types of changes

- [ ] Documentation updated.


SYangster (approved):
tested the changes, works

PR #1960, opened 2023-08-28
Remove fobs calls

yhwen:
Fixes # .

### Description

Removed the no need extra Fobs.dumps() calls from the platform layer. Temporarily changed the Cell to use the CoreCell functions directly. @nvidianz will update it again to integrate with the Datum solution.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/fuel/f3/cellnet/cell.py
+++ b/nvflare/fuel/f3/cellnet/cell.py
@@ -20,7 +20,7 @@ from typing import Dict, List, Union
 from nvflare.apis.fl_constant import ServerCommandNames
 from nvflare.fuel.f3.cellnet.core_cell import CoreCell
 from nvflare.fuel.f3.cellnet.defs import MessageHeaderKey, MessageType, ReturnCode
-from nvflare.fuel.f3.cellnet.utils import make_reply, encode_payload
+from nvflare.fuel.f3.cellnet.utils import encode_payload, make_reply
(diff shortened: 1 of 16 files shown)


nvflare/fuel/f3/cellnet/cell.py
```diff
@@ -98,16 +99,20 @@ def fire_and_forget(
...
+        # if channel == CellChannel.SERVER_COMMAND and topic == ServerCommandNames.HANDLE_DEAD_JOB:
+        #     if isinstance(targets, list):
+        #         for target in targets:
+        #             self.send_blob(channel=channel, topic=topic, target=target, message=message)
+        #     else:
+        #         self.send_blob(channel=channel, topic=topic, target=targets, message=message)
+        # else:
+        #     self.core_cell.fire_and_forget(
+        #         channel=channel, topic=topic, targets=targets, message=message, optional=optional
+        #     )
```
YuanTingHsieh, line 111: The framework does not need to fobs.dump/fobs.load again since the CoreCell already handle that part.
But why we need to use CoreCell directly in here? this would mean that the streaming/LLM ability is DISABLED?

yhwen: As I mentioned in the description, this is a temporary change. Zhihong will continue to make the changes to integrate with Datum solution.

nvidianz (approved):

PR #1963, opened 2023-08-30
Add experiment tracking docs

nvkevlu:
Add to experiment tracking documentation and some changes to fix docs.

### Types of changes

- [ ] Documentation updated.


docs/programming_guide/experiment_tracking.rst
```diff
@@ -0,0 +1,131 @@
...
+Overview and Approaches
+***********************
+
+In a federated computing setting, the data is distributed across multiple devices or systems, and training is run
+on each device independently while preserving each client's data privacy.
+
+Assuming a federated system consisting of one server and many clients and the server coordinating the ML training of clients,
+we can interact with ML experiment tracking tools in two different ways: 
+
+    - Client-side experiment tracking: Each client will directly send the log metrics/parameters to the ML experiment tracking server or local file system
```
YuanTingHsieh, line 17: Can we add some examples here like:

the ML experiment tracking server (like MLFlow) or local file system (like tensorboard)

docs/programming_guide/experiment_tracking.rst
```diff
@@ -0,0 +1,131 @@
...
+    - Server-side experiment tracking: Clients will send the log metrics/parameters to FL server, and the FL server will send the metrics to ML experiment
+      tracking server or local file system
+
+Each approach will have its use cases and unique challenges. In NVFLARE, we developed a server-side approach:
```
YuanTingHsieh, line 21: This sentence might give the impression that NVFlare only supports server-side tracking.

We should say NVflare supports both client-side tracking and server-side tracking.

If we add our "Receivers" at the client side we could achieve client-side tracking with local file system.

If they want clients to send directly to some tracking server, they need to set up that tracking server and in their custom code import the corresponding api and call it themselves.

docs/programming_guide/experiment_tracking.rst
```diff
@@ -0,0 +1,131 @@
...
+      can be highly reduced. In some cases, such as in MLFLow, the events can be buffered in the server and sent to the tracking
+      server in batches, further reducing the traffic to the tracking server. The buffer may add additional latency, so you can
+      disable the buffering if you can set the buffer flush time to 0 assuming the tracking server can take the traffic.
+    - Another key benefit of using server-side experiment tracking is that we separate the metrics data collection 
+      from the metrics data delivery to the tracking server. Clients are only responsible for collecting metrics, and only the server needs to 
+      know about the tracking server. This allows us to have different tools for data collection and data delivery.
+      For example, if the client has training code with logging in Tensorboard syntax, without changing the code, the server can
+      receive the logged data and deliver the metrics to MLflow.
+    - Server-side experiment tracking also can organize different clients' results into different experiment runs so they can be easily
+      compared side-by-side.
```
YuanTingHsieh, line 36: Listed out pros and cons is good.

If possible, Could we also list client side experiment tracking pros and cons?

docs/programming_guide/experiment_tracking.rst
```diff
@@ -0,0 +1,131 @@
...
+      receive the logged data and deliver the metrics to MLflow.
+    - Server-side experiment tracking also can organize different clients' results into different experiment runs so they can be easily
+      compared side-by-side. 
+
+**************************************
+Tools, Sender, LogWriter and Receivers
+**************************************
+
+With the "experiment_tracking" examples in the advanced examples directory, you can see how to track and visualize
+experiments in real time and compare results by leveraging several experiment tracking tools:
```
YuanTingHsieh, line 43: Should we call them tools or frameworks or libraries?

docs/programming_guide/experiment_tracking.rst
```diff
@@ -0,0 +1,131 @@
...
+
+    - Tensorboard
```
YuanTingHsieh, line 45: It will be good to make each of these a link to their official website so people knows 100% what these are

docs/programming_guide/experiment_tracking.rst
```diff
@@ -0,0 +1,131 @@
...
+    - MLflow 
+    - Weights and Biases
+
+.. note::
+
+    The user needs to sign up at Weights and Biases to access the service, NVFlare can not provide access.
+
+In the Federated Learning phase, users can choose an API syntax that they are used to from one
+of above tools. NVFlare has developed components that mimic these APIs called
+:class:`LogWriters <nvflare.app_common.tracking.log_writer.LogWriter>`. All clients experiment logs
```
YuanTingHsieh, line 55: I think the LogWriters is firing "local" event, so we could also receive in different format at the client side.

We need the "ConvertToFedEvent" widget to convert local events to fed events so that our server side can receive it

YuanTingHsieh (commented):
Thanks. 

Overall very good and clear to me.

Suggest some changes

docs/programming_guide/experiment_tracking.rst
YuanTingHsieh: I have tested that before.

Only benefit is that you can log/receive cross-formats (for example TB->MLflow)

docs/programming_guide/experiment_tracking.rst
YuanTingHsieh: Then systems?

PR #1965, opened 2023-08-31
Refactor format_log_message with more readability

IsaacYangSLA:
### Description

This PR makes the log message from CoreCell more readable.  The new log message will look like `2023-08-30 16:51:28,587 - CoreCell - DEBUG - [ME=site-1.:ae98c47 O=server.:ae98c47 D=site-1.:ae98c47 F=server T=site-1.:ae98c47 CH=server_command TP=get_task SCH=? STP=?] trigger waiter - replies received from 1 targets for 26a7410a-e8d1-429c-8ae2-ddb0ac9bf9b6`

The fqcn of job cell is shortened so the uuid part is shown with last 7 characters with a leading `:`.  Other parts, if longer than 8, will be shortened in the similar way.

SCH and STP are Stream channel and Stream topic, respectively.

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh (approved):
LGTM

PR #1988, opened 2023-09-13
Support big job workspace

yanchengnv:
Fixes # .

### Description

- Save workspace without loading data into memory. 
- Load job data directly to file system when downloading to admin
- Improve download_job command with binary protocol to stream job files chunk by chunk, without loading into memory
- Unzip received files on admin client
- Fixed a bug in client_runner in sync runner processing

### Types of changes

- [ ] Documentation updated.


nvflare/private/fed/server/job_cmds.py
```diff
@@ -613,25 +631,33 @@ def _unzip_data(self, download_dir, job_data, job_id):
...
+        with engine.new_context() as fl_ctx:
+            job_def_manager.get_storage_for_download(job_id, download_dir, DATA, JOB_ZIP, fl_ctx)
+            job_def_manager.get_storage_for_download(job_id, download_dir, META, META_JSON, fl_ctx)
+            job_def_manager.get_storage_for_download(job_id, download_dir, WORKSPACE, WORKSPACE_ZIP, fl_ctx)
 
-        except Exception as e:
-            conn.append_error(f"Exception occurred trying to get job from store: {secure_format_exception(e)}")
-            return
+            self.download_folder(
+                conn, job_id, download_file_cmd_name=AdminCommandNames.DOWNLOAD_JOB_FILE, control_id=job_id
```
yhwen, line 662: The "self.download_folder()" will try to download all the files within the folder. Before the job run completes, the "workspace" binary file does not exist in the job store. But we still create the "wokspace.zip" symlink to that file. The advantage is whenever the "workspace" binary file generated in the job store, then it becomes available automatically. However, before the job run completes, the download_job will get an error for downloading this workspace.zip file, becuase it does not exist. Should we handle this scenario in the download_folder()?

Alternately, we may consider not to create the "workspace.zip" symlink before the job run workspace has been created. But we need to handle one way or the other.

yhwen (changes requested):

nvflare/private/fed/server/job_cmds.py
yanchengnv: Right. Let's use the alternative approach: if workspace doesn't exist yet, don't create the symlink.

yhwen (approved):

nvflare/fuel/hci/chunk.py
```diff
@@ -0,0 +1,209 @@
...
+HEADER_LEN = HEADER_STRUCT.size
+
+MARKER_DATA = 101
+MARKER_END = 102
+
+MAX_CHUNK_SIZE = 1024 * 1024
+
+
+def get_slice(buf, start: int, length: int):
+    return buf[start : start + length]
```
nvidianz, line 28: Use memoryview here could improve performance a little.

IsaacYangSLA: I think it's like a 'preference' or 'cosmetic issue.'

nvidianz (approved):

PR #1992, opened 2023-09-13
Add secure argument to Task and set secure argument in cell's call.

IsaacYangSLA:
### Description

This PR handles the new argument 'secure' in cell's methods.  And adds the new argument 'secure' in Task so client task control can set it in Task.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/apis/controller_spec.py
+++ b/nvflare/apis/controller_spec.py
@@ -77,6 +77,7 @@ class Task(object):
         result_received_cb=None,
         task_done_cb=None,
         operator=None,
+        secure=False,
     ):
         """Init the Task.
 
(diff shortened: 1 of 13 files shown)


yanchengnv (changes requested):
The changes are good, but not enough. You only changed server_engine_spec.py, there are concrete classes that implement this spec. All these classes need to be changed. You can do a search on "send_aux_request" and find these places.

IsaacYangSLA:
Thanks.  Let me go through all places.

yanchengnv (approved):
LGTM

YuanTingHsieh (approved):
LGTM

PR #1999, opened 2023-09-14
Support secure task in CCWF and comm improvement

yanchengnv:
Fixes # .

### Description

- Added support of secure task for CCWF. If the system is run in secure mode, all peer-to-peer task communications are encrypted.
- Improve the processing efficiency of cell's broadcast_request by encoding the payload before sending to multiple targets. This avoid multiple encoding my these targets.
- Changed cell's fire_and_forget to _fire_and_forget. This is to use the method routing.
- Use memoryview in binary file transfer between admin client and server. This to avoid bytes copy.
- Improved dead-client-cleanup in fed server to allow graceful shutdown.
- Improved abort job messages to avoid confusion.
- Allow multiple concurrent download of the same job. Clean up transfer folder after job downloaded.
- Support optional output folder name with download_job command, to allow a meaningful name for the job result folder. If not specified, default to job id.

### Types of changes

- [ ] Documentation updated.


nvflare/fuel/hci/client/file_transfer.py
```diff
@@ -339,59 +343,113 @@ def download_text_file(self, args, ctx: CommandContext):
...
+                        self.download_dir,
+                        f"{destination_name}__{tx_id}",
+                    )
+                    try:
+                        os.rename(tx_path, alt_destination_path)
+                        location = alt_destination_path
+                    except:
+                        location = tx_path
+                else:
+                    location = tx_path
```
yhwen, line 448: If download_job multiple times (sequentially), this will result the job and workspace been downloaded to the "transfer" folder multiple times, each with its own tx_id. This will cause confusion to the users. Most likely the users expect the latest download result to overwrite the previous contents.

yhwen (changes requested):

nvflare/fuel/hci/client/file_transfer.py
yanchengnv: I disagree. Just like regular download, if the same file is downloaded to your computer's Download folder, each time a new file is created with a different name (usually the name and a suffix).

yhwen: well, probably this is a little subjective. But I found this is very confusing, especially when running the "download_job" command multiple times. 

If we decide to make this change, we need to well document it. We are changing the user experience behavior.

nvflare/private/fed/server/job_cmds.py
```diff
@@ -165,12 +166,17 @@ def get_spec(self):
         )
 
     def authorize_job_file(self, conn: Connection, args: List[str]):
-        if len(args) < 2:
+        """
+        Args: cmd_name tx_id job_id file_name [end]
```
YuanTingHsieh, line 170: What does [end] here mean?

nvflare/private/fed/server/job_cmds.py
```diff
@@ -165,12 +166,17 @@ def get_spec(self):
...
+        """
+        if len(args) < 3:
             conn.append_error(
                 "syntax error: missing job_id", meta=make_meta(MetaStatusValue.SYNTAX_ERROR, "missing job_id")
             )
```
YuanTingHsieh, line 175: This error message does not match the check "if len(args) <3"
Users could be missing one of the following: tx_id, job_id or file_name.

nvflare/fuel/hci/client/file_transfer.py
yanchengnv: Yes this is subjective. On the other hand, it's not easy to support download of the same job at the same time. Both downloads are writing to temp folder, which is not a problem. But at the end of the download, only one of them can be renamed to the proper name, and the other rename will fail.  There is no way to know which download should be kept or removed. That's why I decided to keep both and let the user know exactly where the downloaded content is.

Yes, we'll need to document it.

nvflare/private/fed/server/job_cmds.py
YuanTingHsieh: I think tx_id should NOT be applicable here?
Given a job_id, whatever the tx_id might be, the file should be the same.

nvflare/private/fed/server/job_cmds.py
```diff
@@ -632,15 +638,37 @@ def _unzip_data(self, download_dir, job_data, job_id):
...
+
     def pull_file(self, conn: Connection, args: List[str]):
-        if len(args) != 3:
-            self.logger.error("syntax error: missing file name")
+        """
+        Args: cmd_name tx_id folder_name file_name [end]
+        """
+        if len(args) < 4:
+            self.logger.error("syntax error: missing tx_id folder_name file name")
             return
```
YuanTingHsieh, line 654: From the user's perspective, they should not worry about tx_id when downloading?

They should just "download_job_file [job_id] [file_name]"
Otherwise we making things more complicated for them?

nvflare/private/fed/server/fed_server.py
```diff
@@ -191,9 +191,20 @@ def deploy(self, args, grpc_args=None, secure_train=False):
         cleanup_thread.start()
 
     def client_cleanup(self):
+        last_remove_time = 0.0
+        remove_interval = 5.0
+        check_interval = 0.2
```
YuanTingHsieh, line 196: This is good improvement.

As we keep saying, later we need to have a common place for all these random constants sleep / check time interval.

nvflare/app_common/ccwf/server_ctl.py
```diff
@@ -87,6 +88,7 @@ def __init__(
             result_clients: clients to receive final results
             max_status_report_interval:
             progress_timeout:
+            private_p2p: whether peer-to-peer communication should be private
```
YuanTingHsieh, line 91: What does it mean for a peer-to-peer communication to be private VS public?

nvflare/apis/impl/task_controller.py
```diff
@@ -83,6 +83,7 @@ def broadcast_and_wait(
         request = task.data
         # apply task filters
         self.log_debug(fl_ctx, "firing event EventType.BEFORE_TASK_DATA_FILTER")
+        fl_ctx.set_prop(FLContextKey.TASK_DATA, task.data, sticky=False, private=True)
         self.fire_event(EventType.BEFORE_TASK_DATA_FILTER, fl_ctx)
```
YuanTingHsieh, line 87: Is this dangerous?
If we set the task.data in the fl_ctx, then any components can use fire_event to modify it.
Will that cause a racing issue?
Will that corrupt our task?

nvflare/private/fed/server/job_cmds.py
yanchengnv: Remember that the files are downloaded (pulled) one by one by the Admin Client with the command.  The "end" tells the server that this is the last file to be downloaded. The is needed so that the server can clean up the temporary download folder.

There could be multiple download transactions for the same job. The tx_id is the unique uuid for this download transaction. Of course the file_name is the same for all transactions (in reality the file_name is meta.josn, workspace.zip and data.zip). The tx_id is not used by authorize_job_file, but it is part of the command. Tx_id is important to ensure that each download session is treated separately (hence multiple transactions are happen at the same time).

nvflare/private/fed/server/job_cmds.py
yanchengnv: Good catch! Will fix.

nvflare/private/fed/server/job_cmds.py
yanchengnv: Correct! The user never sees tx_id. The "pull_file" is an internal command issued by the admin client's "pull_folder" command (which is mapped to the server's download_job command). So the user only sees "download_job job_id [result_name]" command.

nvflare/fuel/hci/client/file_transfer.py
yhwen: If we want to change this behavior, we should change the "tx_id" to be some number easier to understand, for example, incremental, 1, 2, 3... instead of using the uuid.

nvflare/private/fed/server/fed_server.py
yanchengnv: Without this, client_cleanup can be stuck for 15 seconds, which will cause the MPM's final cleanup to fail.

nvflare/app_common/ccwf/server_ctl.py
yanchengnv: I debated with myself what's the proper name for this arg, and decided on private_p2p. This means that when this flag is set, task communications between the two peers are encrypted. 

Private P2P communication is an additional level of protection on basic communication security (such as SSL). Each pair of peers have their own encryption keys to ensure that only they themselves can understand their messages, even if the messages may be relayed through other sites (e.g. server). Different pairs of peers have different keys. Currently, private P2P is enabled only when the system is in secure mode. This is because key exchange between peers requires both sides to have PKI certificates and keys, which requires the project to be provisioned in secure mode.

nvflare/apis/impl/task_controller.py
yanchengnv: This is a double-edged sword (like in many other places). Even handlers could have legitimate reason to process or just read the task data. I keep this here to be consistent with the existing behavior of server/client task interactions.

nvflare/fuel/hci/client/file_transfer.py
yanchengnv: I guess we could. It would make it look nicer.

nvflare/app_common/ccwf/server_ctl.py
YuanTingHsieh: Got it, maybe we could just add this explanation to the docstring as well?
Something like:
```
"""xxx

Args:
  xxx: yyy
  private_p2p: When this flag is set, task communications between the two peers are encrypted.

Note:
  Private P2P communication is an additional level of protection on basic communication security (such as SSL). 
  Each pair of peers have their own encryption keys to ensure that only they themselves can understand their messages, even if the messages may be relayed through other sites (e.g. server). 
  Different pairs of peers have different keys. Currently, private P2P is enabled only when the system is in secure mode. 
  This is because key exchange between peers requires both sides to have PKI certificates and keys, which requires the project to be provisioned in secure mode.

"""
```

nvflare/private/fed/server/job_cmds.py
YuanTingHsieh: got it, this is clear for "download_job job_id [result_name]" command.

So we DO NOT have a command for users to download just a specific file in a job?

yanchengnv: No we don't, and we never did.

nvflare/app_common/ccwf/server_ctl.py
yanchengnv: Yeah, I probably need to add docstrings for all args. They are all documented in the google doc, but it would be nice to include into docstrings.

nvflare/private/fed/server/job_cmds.py
YuanTingHsieh: got it, sounds good, all clear to me now.

nvflare/private/fed/server/job_cmds.py
```diff
@@ -165,12 +166,17 @@ def get_spec(self):
         )
 
     def authorize_job_file(self, conn: Connection, args: List[str]):
-        if len(args) < 2:
+        """
+        Args: cmd_name tx_id job_id file_name [end]
+        """
+        if len(args) < 3:
```
YuanTingHsieh, line 172: This needs to be len(args) <4? 
Since we need "cmd_name tx_id job_id file_name"
And we need to change the text in the CommandSpec up there in Line 159.
`usage=f"{AdminCommandNames.DOWNLOAD_JOB_FILE} tx_id job_id file_name",`

yanchengnv: Should be 4, yes. Will fix.

nvflare/app_common/ccwf/server_ctl.py
```diff
@@ -109,6 +111,13 @@ def __init__(
         self.participating_clients = participating_clients
         self.result_clients = result_clients
         self.result_clients_policy = result_clients_policy
+
+        # make private_p2p bool
+        if private_p2p:
```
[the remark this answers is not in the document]
yanchengnv: Need to make it a bool since it will be sent as part of config. I think I should validate it.

nvflare/fuel/hci/client/file_transfer.py
yanchengnv: Fixed - made the folder name ended with simple integers if necessary.

nvflare/app_common/ccwf/server_ctl.py
yanchengnv: Added docstrings.

nvflare/private/fed/server/job_cmds.py
yanchengnv: Fixed.

nvflare/app_common/ccwf/server_ctl.py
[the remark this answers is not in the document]
yanchengnv: Fixed.

nvflare/private/fed/server/job_cmds.py
yanchengnv: Fixed.

yhwen (approved):
LGTM

PR #2005, opened 2023-09-18
Swarm learning and cross-site-evaluation with cifar10

SYangster:
- adapted cifar10 pytorch example for swarm learning and client side cross-site evaluation
- created new swarm_cse_pt job template

### Types of changes

- [ ] Documentation updated.


SYangster:
> whats "cse" stands for ?

cross-site evaluation

examples/advanced/swarm_learning/README.md
```diff
@@ -0,0 +1,80 @@
...
+Our implementation of swarm learning leverages the newly added Client Controlled Workflow (CCWF) framework in FLARE, which allows for secure peer-to-peer communication between clients. The server is simply responsible for the job lifecycle management (health of client sites and monitoring of job status), while the clients are now responsible for training logic and aggregration management (where tasks are assigned via peer-to-peer communication).
+
+Algorithmically, swarm learning is identical to federated averaging with the main differences being that the server will no longer control the aggregation process and will not have access to any sensitive information, such as trained model weights.
+
+## Swarm Learning with FLARE
+
+In swarm learning, training and aggregation is done accoss multiple rounds. The starting client is responsible for the initial model which is loaded from the configured persistor. In each round:
+1. An aggregator client is randomly chosen from all clients.
+2. The training task is sent to all training clients on the current global model.
+3. The clients then send their training results to the designated client for aggregration.
```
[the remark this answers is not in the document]
yanchengnv: Step 1 already mentioned that the aggregator client is randomly selected. Maybe in step 3, you could say "to the selected client for aggregation".

examples/advanced/swarm_learning/pt/cifar10_data_utils.py
```diff
@@ -0,0 +1,181 @@
+# Copyright (c) 2023, NVIDIA CORPORATION.  All rights reserved.
```
yanchengnv, line 1: Are these code under "pt" the same as the centralized CIFAR10 example? If so, we should consolidate to reuse. This will also make it clear that all existing training code can be used without changes.

job_templates/swarm_cse_pt/config_fed_client.json
```diff
@@ -0,0 +1,94 @@
...
+      "executor": {
+        "path": "nvflare.app_common.ccwf.SwarmClientController",
+        "args": {
+          "learn_task_name": "train",
+          "learn_task_timeout": 120,
+          "persistor_id": "persistor",
+          "aggregator_id": "aggregator",
+          "shareable_generator_id": "shareable_generator",
+          "min_responses_required": 2,
+          "wait_time_after_min_resps_received": 120
```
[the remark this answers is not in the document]
yanchengnv: This is the standard behavior of original SAG to counter against potential communication problems. You wait for N secs after min replies received (if min != total), hoping to get more replies.

nvflare/app_common/ccwf/swarm_client_ctl.py
```diff
@@ -554,6 +554,9 @@ def do_learn_task(self, name: str, task_data: Shareable, fl_ctx: FLContext, abor
 
         self.log_info(fl_ctx, f"Round {current_round} started.")
 
+        learnable = self.persistor.load(fl_ctx)
+        fl_ctx.set_prop(AppConstants.GLOBAL_MODEL, learnable, private=True, sticky=True)
```
yanchengnv, line 558: Why do you need to do this?

yanchengnv (changes requested):
Try to consolidate training code with original CIFAR10.
Not sure why you need to load from persistor for global model of each round.

nvflare/app_common/ccwf/swarm_client_ctl.py
SYangster: Right this isn’t needed here, only for starting client. Making a fix since persistor.load() creates PTModelPersistenceFormatManager, so need to create this somewhere else since load() will only be called at initial site.

examples/advanced/swarm_learning/pt/cifar10_model_learner.py
```diff
@@ -0,0 +1,342 @@
...
+        aggregation_epochs: int = 1,
+        lr: float = 1e-2,
+        analytic_sender_id: str = "analytic_sender",
+        batch_size: int = 64,
+        num_workers: int = 0,
+    ):
+        """Simple CIFAR-10 Trainer.
+
+        Args:
+            train_idx_root: directory with site training indices for CIFAR-10 data.
```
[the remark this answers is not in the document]
SYangster: Will change to consolidate with original cifar10 model learner code. Original code uses CIFAR10_Idx to create subsets of dataset so if we want to reuse the model learner we have to use this as well.

job_templates/swarm_cse_pt/config_fed_client.json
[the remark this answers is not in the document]
SYangster: I'll lower to 30 sec

job_templates/swarm_cse_pt/config_fed_client.json
```diff
@@ -0,0 +1,88 @@
...
+    {
+      "id": "aggregator",
+      "path": "nvflare.app_common.aggregators.intime_accumulate_model_aggregator.InTimeAccumulateWeightedAggregator",
+      "args": {
+        "expected_data_kind": "WEIGHT_DIFF"
+      }
+    },
+    {
+      "id": "model_selector",
+      "name": "IntimeModelSelector",
```
YuanTingHsieh, line 79: suggest using "path" for additional clarity

job_templates/swarm_cse_pt/config_fed_server.json
```diff
@@ -0,0 +1,37 @@
...
+      "path": "pt.utils.cifar10_data_splitter.Cifar10DataSplitter",
+      "args": {
+          "split_dir": "/tmp/cifar10_splits",
+          "num_sites": 2,
+          "alpha": 1.0
+      }
+    },
+    {
+      "id": "json_generator",
+      "name": "ValidationJsonGenerator",
```
YuanTingHsieh, line 17: suggest using "path" for additional clarity

nvflare/app_common/ccwf/client_ctl.py
```diff
@@ -357,15 +357,17 @@ def execute(self, task_name: str, shareable: Shareable, fl_ctx: FLContext, abort
                 message_handle_func=self._process_end_workflow,
             )
 
+            learnable = self.persistor.load(fl_ctx)
+            fl_ctx.set_prop(AppConstants.GLOBAL_MODEL, learnable, private=True, sticky=True)
+
```
YuanTingHsieh, line 362: Why do we move this from "start_task_name" to "configure_task_name"?

nvflare/app_opt/pt/file_model_persistor.py
```diff
@@ -204,6 +204,11 @@ def load_model(self, fl_ctx: FLContext) -> ModelLearnable:
         else:
             # if no pretrained model provided, use the generated network weights from APP config
             # note that, if set "determinism" in the config, the init model weights will always be the same
+            self.log_info(
+                fl_ctx,
+                f"Unable to load model from {src_file_name}: Using default data instead.",
```
YuanTingHsieh, line 209: Suggest enhance this message for more clarity:

``f"Both source_ckpt_file_full_name and {AppConstants.CKPT_PRELOAD_PATH} are not provided. Using the default model weights initialized on the persistor side."``

nvflare/app_common/ccwf/client_ctl.py
SYangster: start_task_name is only for the starting client, and the persistor needs to be loaded to initialize the persistence_manager on all clients

YuanTingHsieh: got it, thanks for the explanation.

One question here:

1. If all the clients want to start from the same initialized weights (to avoid non-convergence), are we doing this?
2. Or are we saying each of the client start with its own initialized weights ( which means each client can have a different starting point)

yanchengnv: No. Only the final model is brodcast at the end of the training. There is no need to broadcast the initial model, since it is only used by the starting client.

The reason that the initial model needs to be loaded here is because some model persistors won't work later when saving result if it didn't load an initial model.

SYangster: Currently, the starting client does send the initial model weights via _scatter to all the clients:
swarm_client_ctl.py:386 `return self.send_learn_task(targets=targets, request=task_data, fl_ctx=fl_ctx)`
where task_data is the initial_model model weights from client_ctl.py:372`return self.start_workflow(initial_model, fl_ctx, abort_signal)`

YuanTingHsieh: So if I understand correctly:


At the very beginning,
All clients (c1, c2, c3, c4) after the "configure task" will set this "AppConstants.GLOBAL_MODEL" to what is loaded using the persistor (which is randomly initialized in pt's case) (m1, m2, m3, m4)

Then server will send "start_task" to the starting client (let's say c1), and the starting client (c1) will get this "AppConstants.GLOBAL_MODEL" as learnable (m1) and then BROADCAST to all participating training clients.

yanchengnv: Correct in the case of swarm learning. In case of cyclic, the starting client will train, and then sends its result to the next client, and so on.

YuanTingHsieh: thanks for all discussions/explanations, I am much clearer.

YuanTingHsieh (approved):

yanchengnv (approved):
The readme is very well written. Well done!

PR #2037, opened 2023-09-27
Complete the TODO places of generating peer fl context

IsaacYangSLA:
### Description

This PR replaces the statements generating peer fl context with the new utility function in the remaining places that require sending the peer fl context.

On large models, the utility function takes about 50ms to finish, compared to the previous statements which takes about 3~4 seconds.

### Types of changes

- [ ] Documentation updated.


yanchengnv:
There are more places that need fixing. Please search get_all_public_props across all files and you should see those places.

PR #2044, opened 2023-09-28
Optional flag and bug fixes

nvidianz:
Fixes FLARE-1594

1. Added optional flag.
2. Made streaming parameter configurable
3. Added info log for connection creation/closing
4. Fixed a bug when error received before future is created.

### Types of changes

- [ ] Documentation updated.


IsaacYangSLA (commented):
I am not sure if there is document describing what files and what names users can use to set values to override the defaults.

nvidianz:
> I am not sure if there is document describing what files and what names users can use to set values to override the defaults.

Where shall we document this? The parameters can be placed inside local/comm_config.json.

IsaacYangSLA:
> > I am not sure if there is document describing what files and what names users can use to set values to override the defaults.
> 
> Where shall we document this? The parameters can be placed inside local/comm_config.json.

I think you may need to update some rst file in docs folder so users know the syntax of that file and the meaning of important variables.

yanchengnv (approved):
LGTM

PR #2045, opened 2023-09-29
Cli job bug fix

chesterxgchen:
Fixes # .
  The previous code assume all path or name are class names, so it looking for the classes module path. with dot "."  
   like "modulepath.Classname", but when the path is not class path such as model.path = "{model_class_path}", it will fail. this PR fixed this by checking if the "dot" exists. if not. its not a class name anyway. 

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh (approved):
Thanks LGTM

PR #2046, opened 2023-09-29
Improve how to generate peer fl context

IsaacYangSLA:
### Description

This PR adds a utility function which generates peer fl context in a more efficient way and update a few locations with this function

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/private/fed/server/server_commands.py
+++ b/nvflare/private/fed/server/server_commands.py
@@ -14,7 +14,6 @@
 
 """FL Admin commands."""
 
-import copy
 import logging
 import time
 from abc import ABC, abstractmethod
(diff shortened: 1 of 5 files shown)


nvflare/private/fed/server/server_commands.py
```diff
@@ -180,8 +179,9 @@ def process(self, data: Shareable, fl_ctx: FLContext):
         shareable.set_header(key=FLContextKey.TASK_ID, value=task_id)
 
         shareable.set_header(key=ServerCommandKey.TASK_NAME, value=taskname)
-        shared_fl_ctx = FLContext()
-        shared_fl_ctx.set_public_props(copy.deepcopy(get_serializable_data(fl_ctx).get_all_public_props()))
+
+        # TODO
```
YuanTingHsieh, line 183: @IsaacYangSLA what is the TODO here about?

YuanTingHsieh (commented):
only one question otherwise LGTM

nvflare/private/fed/server/server_commands.py
yanchengnv: I think this PR just addressed this "TODO".

yanchengnv (approved):
Looks great!

nvflare/private/fed/server/server_commands.py
IsaacYangSLA: Yan is right.  I put that TODO when searching for places to update as reminders.

PR #2053, opened 2023-10-03
Fix AV FL Issues

yanchengnv:
Fixes # .

### Description

- Changed CP's heartbeat to use a short timeout (maint_msg_timeout) to fix the HB bug.
- Added retry logic for submitting task results
- 
### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/private/fed/client/client_runner.py
+++ b/nvflare/private/fed/client/client_runner.py
@@ -111,7 +111,7 @@ class ClientRunner(FLComponent):
         self.task_lock = threading.Lock()
         self.end_run_fired = False
         self.end_run_lock = threading.Lock()
-        self.task_check_timeout = 2.0
+        self.task_check_timeout = 5.0
         self.task_check_interval = 5.0
         self._register_aux_message_handler(engine)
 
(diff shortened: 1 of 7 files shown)


IsaacYangSLA (approved):
Thanks.  The only part that may need change is the value of main_msg_timeout in communicator.  Its argument, `timeout`, is configured via communication_timeout.

nvflare/private/fed/client/client_runner.py
```diff
@@ -473,19 +478,108 @@ def fetch_and_run_one_task(self, fl_ctx) -> (float, bool):
...
 
-        reply_sent = self.engine.send_task_result(task_reply, fl_ctx)
-        if reply_sent:
-            self.log_info(fl_ctx, "result sent to server for task: name={}, id={}".format(task.name, task.task_id))
-        else:
-            self.log_error(
-                fl_ctx,
-                "failed to send result to server for task: name={}, id={}".format(task.name, task.task_id),
-            )
+        self._send_task_result(task_reply, task.task_id, fl_ctx)
```
SYangster, line 481: Do we need to add a check here for the return value to see if result was sent or not? As it could return false from ask_to_stop or TASK_GONE

yanchengnv: We could but not necessary since the method already logged detailed message. In fact, it doesn't need to return anything.

PR #2058, opened 2023-10-04
Restore non-aio GRPC and a few improvements

yanchengnv:
Fixes # .

### Description

- Restore non-asyncio GRPC driver.  A site can choose which driver to use by setting "use_aio_grpc" in comm_config.json, or via system environment variable NVFLARE_USE_AIO_GRPC. The default is True to be backward compatible
- Consolidated common functions used by grpc driver and aio-grpc driver.
- Removed unnecessary use of multiprocess since it could cause semaphore leak at the end of process.
- Changed to directly send messages to server instead of using pool map.
- Fixed CP/SP heartbeat bug and use small timeout value.
- Added retry logic for sending task results from CJ to SJ.
- Fixed download_job bug in flare_api

### Types of changes

- [ ] Documentation updated.


IsaacYangSLA (dismissed):
Thanks.  LGTM.

nvflare/fuel/f3/drivers/net_utils.py
```diff
@@ -79,7 +79,8 @@ def get_ssl_context(params: dict, ssl_server: bool) -> Optional[SSLContext]:
 def get_address(params: dict) -> str:
     host = params.get(DriverParams.HOST.value, "0.0.0.0")
     port = params.get(DriverParams.PORT.value, 0)
-
+    if not host:
```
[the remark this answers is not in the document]
yanchengnv: Initially I though so too, but it actually is necessary because the value of host is empty string after executing line 80! Somehow the params dict contains the host value as empty str.

nvidianz (approved):

tests/unit_test/fuel/f3/drivers/driver_manager_test.py
```diff
@@ -37,6 +38,8 @@ def manager(self):
             ("stcp", TcpDriver),
             ("grpc", AioGrpcDriver),
             ("grpcs", AioGrpcDriver),
+            ("nagrpc", GrpcDriver),
+            ("nagrpcs", GrpcDriver),
```
YuanTingHsieh, line 42: minor thing, we have tcp vs atcp,
maybe for grpc we want grpc vs agrpc.
but this does not affect things.

nvflare/private/fed/server/server_runner.py
```diff
@@ -475,6 +483,31 @@ def process_submission(self, client: Client, task_name: str, task_id: str, resul
...
+        self.log_info(fl_ctx, "received client job_heartbeat aux request")
+        return make_reply(ReturnCode.OK)
+
+    def _handle_task_check(self, topic: str, request: Shareable, fl_ctx: FLContext) -> Shareable:
+        task_id = request.get_header(ReservedHeaderKey.TASK_ID)
+        if not task_id:
+            self.log_error(fl_ctx, f"missing {ReservedHeaderKey.TASK_ID} in task_check request")
+            return make_reply(ReturnCode.BAD_REQUEST_DATA)
+
+        self.log_info(fl_ctx, f"received task_check on task {task_id}")
```
YuanTingHsieh, line 496: Do we need this on "INFO" level?

Another thing, should the message be `f"received task_check on ClientTask {task_id}"`?
Since we have Task VS ClientTask?

Or that is too much for the regular users?

nvflare/private/fed/server/server_runner.py
```diff
@@ -475,6 +483,31 @@ def process_submission(self, client: Client, task_name: str, task_id: str, resul
...
+                self.log_info(fl_ctx, "no current workflow - dropped task_check.")
+                return make_reply(ReturnCode.TASK_UNKNOWN)
+
+            task = self.current_wf.responder.process_task_check(task_id=task_id, fl_ctx=fl_ctx)
+            if task:
+                self.log_info(fl_ctx, f"task {task_id} is still good")
+                return make_reply(ReturnCode.OK)
+            else:
+                self.log_info(fl_ctx, f"task {task_id} is not found")
+                return make_reply(ReturnCode.TASK_UNKNOWN)
```
YuanTingHsieh, line 509: Do we need this on "INFO" level?

Should the message be `f"ClientTask {task_id} is still good / is not found"`?
Since we have Task VS ClientTask?

Or that is too much for the regular users?

nvflare/private/fed/client/fed_client_base.py
```diff
@@ -336,63 +326,38 @@ def quit_remote(self, project_name, fl_ctx: FLContext):
...
     def push_results(self, shareable: Shareable, fl_ctx: FLContext):
         """Push the local model to multiple servers."""
-        pool = None
-        try:
-            pool = ThreadPool(len(self.servers))
-            return pool.map(partial(self.push_execute_result, shareable=shareable, fl_ctx=fl_ctx), tuple(self.servers))
-        finally:
-            if pool:
-                pool.terminate()
+        return self.push_execute_result(self._get_project_name(), shareable, fl_ctx)
```
YuanTingHsieh, line 353: The original way "theoretically" can map this to multiple servers.

Although in reality we never support multiple servers.
I like the change to make things cleaner.
If later we need to support multiple servers/services we need to look at the whole system and do it.

nvflare/private/fed/client/client_runner.py
```diff
@@ -439,19 +468,105 @@ def fetch_and_run_one_task(self, fl_ctx) -> (float, bool):
...
+            try_count += 1
+            rc = self._try_send_result_once(result, task_id, fl_ctx)
+
+            if rc == _TASK_CHECK_RESULT_OK:
+                return True
+            elif rc == _TASK_CHECK_RESULT_TASK_GONE:
+                return False
+            else:
+                # retry
+                time.sleep(self.task_check_interval)
```
YuanTingHsieh, line 495: If we are not using the return value of this method, should we just replace all "return XXX" to return?

nvflare/private/fed/client/client_runner.py
```diff
@@ -439,19 +468,105 @@ def fetch_and_run_one_task(self, fl_ctx) -> (float, bool):
...
+            if not isinstance(reply, Shareable):
+                self.log_error(fl_ctx, f"bad task_check reply from server: expect Shareable but got {type(reply)}")
+                return _TASK_CHECK_RESULT_TRY_AGAIN
+
+            rc = reply.get_return_code()
+            if rc == ReturnCode.OK:
+                return _TASK_CHECK_RESULT_OK
+            elif rc == ReturnCode.COMMUNICATION_ERROR:
+                self.log_error(fl_ctx, f"failed task_check: {rc}")
+                return _TASK_CHECK_RESULT_TRY_AGAIN
```
YuanTingHsieh, line 555: If the communication is dead (we get `ReturnCode.COMMUNICATION_ERROR`), will ClientRunner keep trying to send result, will it hang?

nvflare/private/fed/client/client_run_manager.py
```diff
@@ -128,10 +128,10 @@ def new_context(self) -> FLContext:
 
     def send_task_result(self, result: Shareable, fl_ctx: FLContext) -> bool:
         push_result = self.client.push_results(result, fl_ctx)  # push task execution results
-        if push_result[0] == CellReturnCode.OK:
+        if push_result == CellReturnCode.OK:
             return True
         else:
-            return False
+            self.logger.error(f"failed to send task result: {push_result}")
```
YuanTingHsieh, line 134: Do we need to `return False` here?

nvflare/fuel/f3/drivers/grpc/qq.py
```diff
@@ -0,0 +1,52 @@
...
+
+import logging
+import queue
+
+
+class QueueClosed(Exception):
+    pass
+
+
+class QQ:
```
YuanTingHsieh, line 23: why do we need this class?
is the reason because we want to use a "for" loop to iterate through it?

nvflare/private/fed/client/client_runner.py
```diff
@@ -398,11 +406,32 @@ def _do_process_task(self, task: TaskAssignment, fl_ctx: FLContext, abort_signal
...
+        heartbeat_thread = threading.Thread(target=self._send_job_heartbeat, args=[], daemon=True)
+        heartbeat_thread.start()
+
         while not self.run_abort_signal.triggered:
             with self.engine.new_context() as fl_ctx:
                 task_fetch_interval, _ = self.fetch_and_run_one_task(fl_ctx)
             time.sleep(task_fetch_interval)
 
+    def _send_job_heartbeat(self, interval=30.0):
+        wait_times = int(interval / 2)
```
YuanTingHsieh, line 418: maybe we can make this 2 becomes a variable? it corresponds to the sleeping time right?
`poll_time=2`?

nvflare/fuel/f3/drivers/net_utils.py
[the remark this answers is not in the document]
YuanTingHsieh: this is good for this PR, maybe we should fix that how can DriverParams.HOST.value be an empty string `""`

nvflare/private/fed/client/fed_client_base.py
yanchengnv: Yes, but that original way was not well thought. Even if we need to support multiple servers at the same time, the simple pool map won't work. And I doubt we'll ever need that.

nvflare/private/fed/client/client_runner.py
yanchengnv: We don't use it now, but we may need it later. That's why I decided to keep this.

nvflare/private/fed/client/client_run_manager.py
yanchengnv: Yes we should, even though Python would return None for us (which is almost equivalent to False). Will fix.

nvflare/private/fed/server/server_runner.py
yanchengnv: See my previous comment.  INFO level is okay since this is not a repeated message for each task under normal situation. Under abnormal situation, we'll need this info to analyze errors. Task ID is already included in the log context.

nvflare/private/fed/client/client_runner.py
yanchengnv: It won't hang, but it will keep trying forever until either the message is sent or the RUN is ended/aborted.

nvflare/fuel/f3/drivers/grpc/qq.py
yanchengnv: Grpc requires iterator to send things out. We use this class to meet this requirement.

nvflare/private/fed/client/client_runner.py
yanchengnv: This was how it was implemented in 2.3. Basically the idea here is that you want to send a HB every "interval" seconds, but you don't want to wait for that long between HBs (since this will block the thread from doing anything while sleeping). We want to be able to sleep short period of time so we can wake up and check other things (e.g. check abort signal triggered). In this impl, the sleep time is 2 secs (which I think may still be too long).

nvflare/private/fed/client/client_run_manager.py
yanchengnv: Fixed.

nvflare/private/fed/client/client_runner.py
yanchengnv: Changed to use 1.0 sec for sleep time and added explanation.

YuanTingHsieh (approved):
Thanks for the explanation, LGTM!

PR #2060, opened 2023-10-05
Fix job cli boolean replacement issue

YuanTingHsieh:
### Issue

I encounter a situation where in config:
```
launch_once = true
```

And in CLI I use `-f config_fed_client.conf app_script=train_diff.py launch_once=false`

In the old logic, we will do this type conversion: bool("false") which equal to True.

Instead, we should be utilizing the same parsing rule as pyhocon and treat this value as boolean "False".

### Description

- Use pyhocon to do type conversion
- Remove unused argument

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/tool/job/config/configer.py
+++ b/nvflare/tool/job/config/configer.py
@@ -122,11 +122,12 @@ def merge_configs(
     """Merges configurations from indices_configs and cli_file_configs.
 
     Args:
-        app_indices_configs (Dict[str,Dict[str, tuple]]): A dictionary containing indices and configurations.
+        app_indices_configs (Dict[str, Dict[str, tuple]]): A dictionary containing indices and configurations.
         app_cli_file_configs (Dict[str, Dict[str, Dict]]): A dictionary containing CLI configurations.
 
     Returns:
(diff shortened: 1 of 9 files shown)


nvflare/tool/job/config/configer.py
```diff
@@ -103,11 +104,22 @@ def filter_config_name_and_values(
...
 
+def _cast_type(key_index, cli_value):
+    """Casts cli_value to correct type.
+
+    Since build_reverse_order_index is using pyhocon, we need to do the same here.
+    """
+    if key_index.value is None:
+        return cli_value
+
+    new_value = ConfigFactory.parse_string(f"{key_index.key}={cli_value}")[key_index.key]
```
SYangster, line 115: Just to double check, if cli_value is "false" rather than "False" does this still convert correctly to False?

SYangster (approved):
Good fix, LGTM

nvflare/tool/job/config/configer.py
YuanTingHsieh: thanks, let me try adding some unit test cases here as well.

SYangster (approved):
the unit tests look good, thanks!

PR #2063, opened 2023-10-06
fix job config section

yanchengnv:
Fixes # .

### Description

This PR fixes a bug that could cause the whole system to become unworkable when a job's config files contain errors. At the end of job config processing, the config service is initialized with the section file to be the job's server/client config. However the config path is set to workspace! Since a workspace could contain many jobs, if any job happens to have a bad config file in it, and the config service happens to use it when searching the section file, the current job will crash.

The fix simply removes this section file since it is not used in our system at all.

### Types of changes

- [ ] Documentation updated.


IsaacYangSLA (approved):
Thanks.  Will build 2.3.6 after this is merged.

PR #2064, opened 2023-10-06
Update swarm learning example

SYangster:
- Change swarm template to use .conf files with comments
- Use script-dir option to get custom code rather than setting PYTHONPATH
- Add section on runtime performance comparison

### Types of changes

- [ ] Documentation updated.


examples/advanced/swarm_learning/README.md
```diff
@@ -120,3 +117,16 @@ Since swarm learning and federated averaging are algorithmically the same, we ca
 Here is the graph of the validation accuracies for the global models of swarm learning and FedAvg compared with the centralized baseline, and we can see the achieved performances are very similar:
 
 <img src="./figs/swarm_graph.png" alt="swarm learning, fedavg, centralized graph" width="500"/>
+
+### Runtime Performance
+
+Swarm learning and FedAvg perform the same amount of training, as seen in the graph above with the same number of steps taken to achieve the same accuracy.
+Therefore, the only difference in runtime is the communication time.
+
+Since swarm learning peer-to-peer messaging actually passes through the server, the communication time is roughly double that of FedAvg. (Swarm learning: client <-> server <-> client vs. FedAvg: server <-> client)
```
YuanTingHsieh, line 126: Is this true?

If the ad-hoc connection is enabled, I thought client will pass the weights directly to the aggregating client?

Only the control message need to pass through the server ?

YuanTingHsieh (commented):
Mostly good, only one question

examples/advanced/swarm_learning/README.md
SYangster: How exactly is the ad-hoc connection enabled? If it is enabled I agree, the communication time with passing the weights would not be doubled.

YuanTingHsieh: @SYangster ad-hoc connection will be established IF it is enabled.

First message pass will be server <-> starting client <-> server <-> working client
Follow up message pass will be starting client <- ad hoc connection <-> working client.

It is enabled by specifying "allow_adhoc_conns" in "comm_config.json"

If ad-hoc is disabled, then the statement of "roughly double" is correct.

@yanchengnv @IsaacYangSLA  do we need to explain this ad-hoc ON/OFF stuff to users?
Or just say it will be double is OK for now?

YuanTingHsieh (approved):
Thanks, LGTM

PR #2065, opened 2023-10-08
Add non-DXO shareable support to FLModelUtil

YuanTingHsieh:
### Description 

- Support raw Shareable for FLModelUtils

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/apis/dxo.py
+++ b/nvflare/apis/dxo.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 import copy
-from typing import List, Union
+from typing import List, Optional, Union
 
 from nvflare.apis.fl_constant import FLMetaKey
 from nvflare.apis.shareable import ReservedHeaderKey, Shareable
@@ -30,6 +30,7 @@ class DataKind(object):
     COLLECTION = "COLLECTION"  # Dict or List of DXO objects
     STATISTICS = "STATISTICS"
     PSI = "PSI"
+    META = "META"  # only contains meta
 
 
 class MetaKey(FLMetaKey):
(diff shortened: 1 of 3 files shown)


nvflare/apis/dxo.py
```diff
@@ -30,6 +30,7 @@ class DataKind(object):
     COLLECTION = "COLLECTION"  # Dict or List of DXO objects
     STATISTICS = "STATISTICS"
     PSI = "PSI"
+    META = "META"  # only contains meta
```
yanchengnv, line 33: In this case, I'd think it should be "NONE" meaning that there is no data.

nvflare/app_common/ccwf/cse_client_ctl.py
```diff
@@ -216,8 +217,10 @@ def _do_process_get_model_request(self, request: Shareable, fl_ctx: FLContext) -
             return make_reply(ReturnCode.BAD_REQUEST_DATA)
 
         if not self.local_model:
+            dxo = DXO(data_kind=DataKind.META)
```
yanchengnv, line 220: I don't see why this is needed.

yanchengnv (changes requested):
I don't understand what this PR is for.

nvflare/app_common/utils/fl_model_utils.py
```diff
@@ -111,7 +111,7 @@ def from_shareable(
```
YuanTingHsieh, line 111: fix it here, need to accept "RAW" Shareable, but not just DXO

yanchengnv:
I don't think we need this PR.

YuanTingHsieh:
As discussed, we should only pass the shareable header in the case of "cross site evaluation"

PR #2070, opened 2023-10-12
Client api use exchange task

YuanTingHsieh:
### Description

- Use ExchangeTask instead of FLModel to communicate
- Rename ModelExchanger to DataExchanger as it does not need to be restricted to FLModel

### Types of changes

- [ ] Documentation updated.


nvflare/client/model_registry.py
```diff
@@ -73,75 +36,94 @@ class ModelRegistry:
...
 
-    def get_model(self):
-        if not self.cache_loaded:
-            self.receive()
-        return copy.deepcopy(self.cached_model)
+    def get_task(self, timeout: Optional[float] = None) -> Optional[ExchangeTask]:
+        try:
+            if not self.cache_loaded:
+                self.receive()
+            return copy.deepcopy(self.received_task)
```
yanchengnv, line 73: Why deepcopy here? Could be very expensive. Should leave it to app to determine whether deepcopy is needed

nvflare/client/model_registry.py
```diff
@@ -73,75 +36,94 @@ class ModelRegistry:
...
+        except:
+            return None
+
+    def get_model(self, timeout: Optional[float] = None) -> Optional[FLModel]:
```
yanchengnv, line 77: Shouldn't this method simply call get_task() and then return the task.data?

nvflare/client/model_registry.py
```diff
@@ -73,75 +36,94 @@ class ModelRegistry:
...
+
+        self.received_task = task
+        self.task_name = task_name
         self.cache_loaded = True
 
-    def get_model(self):
-        if not self.cache_loaded:
-            self.receive()
-        return copy.deepcopy(self.cached_model)
+    def get_task(self, timeout: Optional[float] = None) -> Optional[ExchangeTask]:
```
yanchengnv, line 69: How do you ensure that the received data is type ExchangeTask?

nvflare/app_common/data_exchange/data_exchanger.py
```diff
@@ -165,6 +169,6 @@ def _send_reply(self, data: Any, req_id: str, timeout: Optional[float] = None) -
         """
         if self.pipe_handler is None:
             raise RuntimeError("PipeMonitor is not initialized.")
-        msg = Message.new_reply(topic=self._topic, data=data, req_msg_id=req_id)
+        msg = Message.new_reply(topic=topic, data=data, req_msg_id=req_id)
         has_been_read = self.pipe_handler.send_to_peer(msg, timeout)
```
yanchengnv, line 173: Where is the retry logic in case the send_to_peer failed? Maybe it should be built into the pipe_handler.

nvflare/client/api.py
```diff
@@ -49,46 +51,65 @@ def init(config: Union[str, Dict] = f"config/{CONFIG_EXCHANGE}", rank: Optional[
...
+            data_exchange_path=client_config.get_exchange_path(),
+            pipe_name=client_config.get_pipe_name(),
+        )
 
-    PROCESS_MODEL_REGISTRY[pid] = ModelRegistry(client_config, rank, mdx)
+    PROCESS_MODEL_REGISTRY[pid] = ModelRegistry(client_config, rank, dx)
 
 
-def _get_model_registry() -> Optional[ModelRegistry]:
+def _get_model_registry() -> ModelRegistry:
```
yanchengnv, line 73: why do you need a dict based on pid? Why can't the model registry be just a global singleton?

nvflare/client/api.py
```diff
@@ -49,46 +51,65 @@ def init(config: Union[str, Dict] = f"config/{CONFIG_EXCHANGE}", rank: Optional[
...
         # TODO: make Pipe configurable
-        mdx = FilePipeModelExchanger(data_exchange_path=client_config.get_exchange_path())
+        dx = FilePipeDataExchanger(
+            supported_topics=client_config.get_supported_topics(),
+            data_exchange_path=client_config.get_exchange_path(),
+            pipe_name=client_config.get_pipe_name(),
+        )
 
-    PROCESS_MODEL_REGISTRY[pid] = ModelRegistry(client_config, rank, mdx)
+    PROCESS_MODEL_REGISTRY[pid] = ModelRegistry(client_config, rank, dx)
```
yanchengnv, line 70: why use a dict on pid?

nvflare/client/api.py
```diff
@@ -49,46 +51,65 @@ def init(config: Union[str, Dict] = f"config/{CONFIG_EXCHANGE}", rank: Optional[
...
-def _get_model_registry() -> Optional[ModelRegistry]:
+def _get_model_registry() -> ModelRegistry:
     pid = os.getpid()
     if pid not in PROCESS_MODEL_REGISTRY:
         raise RuntimeError("needs to call init method first")
     return PROCESS_MODEL_REGISTRY[pid]
 
 
-def receive() -> FLModel:
+def receive_task(timeout: Optional[float] = None) -> Tuple[str, Any, dict]:
```
yanchengnv, line 80: why not return ExchangeTask object?

nvflare/client/api.py
```diff
@@ -49,46 +51,65 @@ def init(config: Union[str, Dict] = f"config/{CONFIG_EXCHANGE}", rank: Optional[
...
+    """Receives a task from NVFlare side."""
+    model_registry = _get_model_registry()
+    task = model_registry.get_task(timeout)
+    if task is None:
+        return "", None, {}
+    return model_registry.task_name, task.data, task.meta
+
+
+def send_task_result(data: dict, meta: dict, return_code: str):
```
yanchengnv, line 89: why not task a ExchangeTask object?

nvflare/client/api.py
```diff
@@ -49,46 +51,65 @@ def init(config: Union[str, Dict] = f"config/{CONFIG_EXCHANGE}", rank: Optional[
...
+    model_registry.send_task(data, meta, return_code)
+
+
+def receive(timeout: Optional[float] = None) -> Optional[FLModel]:
     """Receives model from NVFlare side.
 
     Returns:
         An FLModel received.
     """
     model_registry = _get_model_registry()
```
yanchengnv, line 100: Shouldn't you call receive_task() and then convert task to FLModel?

nvflare/client/api.py
```diff
@@ -49,46 +51,65 @@ def init(config: Union[str, Dict] = f"config/{CONFIG_EXCHANGE}", rank: Optional[
...
 def send(fl_model: FLModel, clear_registry: bool = True) -> None:
     """Sends the model to NVFlare side.
 
     Args:
+        fl_model (FLModel): Sends a FLModel object.
         clear_registry (bool): To clear the registry or not.
     """
     model_registry = _get_model_registry()
-    model_registry.send(model=fl_model)
+    model_registry.send_model(model=fl_model)
```
yanchengnv, line 112: similarly shouldn't you convert FLModel to task and then send call send_task()?
The layering isn't very clear: who is responsible for conversion btw ExchangeTask and FLModel?

yanchengnv (changes requested):
It looks good in general, but the concept layering is a little fuzzy. Since now you introduced the ExchangeTask concept, the lower layer should be task based, and then build FLModel based data exchange on top of that. The ModelRegistry has both in it. Perhaps you could build base class "TaskRegistry" and then create ModelRegistry as a subclass to provide convenience method of get_model and receive_model (which use the get_task and receive_task of the base class). 

Another option is to build ModelRegistry with a TaskRegistry component.

The 3rd-option is to not create ModelRegistry at all. Instead just provide some convenience function like get_model(task_reg), which will call task_reg.get_task() and then convert the result to FLModel.

nvflare/client/model_registry.py
YuanTingHsieh: ok, will leave it to app.

nvflare/client/model_registry.py
YuanTingHsieh: change as suggested.

nvflare/client/model_registry.py
YuanTingHsieh: added checks

nvflare/app_common/data_exchange/data_exchanger.py
YuanTingHsieh: it should, let's fix that in another PR.

nvflare/client/api.py
YuanTingHsieh: As discussed, on the api level, let's not expose the ExchangeTask concept for this version.

nvflare/client/api.py
YuanTingHsieh: As discussed, on the api level, let's not expose the ExchangeTask concept for this version.

nvflare/client/api.py
YuanTingHsieh: Introduce ModelRegistry that do this convenient conversion

nvflare/client/api.py
YuanTingHsieh: Introduce ModelRegistry that do this convenient conversion

yanchengnv (changes requested):
Looks better. Only thing is that the name "send_task" and "get_task" are confusing. Let's use the following terms:
- get_task: this is to receive the task from FLARE site
- submit_task: submit task result to the FLARE site.

When applied to FLModel, use get_model, and submit_model.

I also used these terms for ipc_agent impl.

yanchengnv (approved):
LGTM

PR #2077, opened 2023-10-13
Restore non-aio grpc driver

yanchengnv:
Fixes # .

### Description

This PR adds non-asyncio version of GRPC driver.  The AIO version of GRPC library could crash when the underlying network is unstable (like SIGSEGV) on the client side. This PR adds a GRPC driver using non-aio version of the GRPC library.

- The user can control which version of GRPC driver to use by either configuring comm_config.json or set the NVFLARE_USE_AIO_GRPC env variable.
- AIO and non-AIO version GRPC drivers can work together. For example, the server could be using AIO version whereas clients can use non-AIO version.
- Changed MAX_WAIT for reconnect to 10 seconds from 60 secs to make reconnect more quickly.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/fuel/f3/drivers/aio_grpc_driver.py
+++ b/nvflare/fuel/f3/drivers/aio_grpc_driver.py
@@ -34,6 +34,7 @@ from nvflare.security.logging import secure_format_exception, secure_format_trac
 from .base_driver import BaseDriver
 from .driver_params import DriverCap, DriverParams
 from .grpc.streamer_pb2 import Frame
+from .grpc.utils import get_grpc_client_credentials, get_grpc_server_credentials, use_aio_grpc
 from .net_utils import MAX_FRAME_SIZE, get_address, get_tcp_urls, ssl_required
 
 GRPC_DEFAULT_OPTIONS = [
@@ -207,10 +208,12 @@ class Server:
 
             secure = ssl_required(params)
             if secure:
-                credentials = AioGrpcDriver.get_grpc_server_credentials(params)
+                credentials = get_grpc_server_credentials(params)
                 self.grpc_server.add_secure_port(addr, server_credentials=credentials)
+                self.logger.info(f"added secure port at {addr}")
             else:
                 self.grpc_server.add_insecure_port(addr)
+                self.logger.info(f"added insecure port at {addr}")
         except Exception as ex:
             conn_ctx.error = f"cannot listen on {addr}: {type(ex)}: {secure_format_exception(ex)}"
             self.logger.debug(conn_ctx.error)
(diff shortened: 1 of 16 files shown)


nvflare/fuel/f3/drivers/aio_grpc_driver.py
```diff
@@ -374,38 +382,9 @@ def shutdown(self):
     def get_urls(scheme: str, resources: dict) -> (str, str):
         secure = resources.get(DriverParams.SECURE)
         if secure:
-            scheme = "grpcs"
+            if use_aio_grpc():
+                scheme = "grpcs"
+            else:
+                scheme = "agrpcs"
```
YuanTingHsieh, line 388: Should it be "grpc" VS "nagrpc"?

nvflare/fuel/f3/drivers/aio_grpc_driver.py
```diff
@@ -251,7 +254,10 @@ def __init__(self):
 
     @staticmethod
     def supported_transports() -> List[str]:
-        return ["grpc", "grpcs"]
+        if use_aio_grpc():
+            return ["grpc", "grpcs"]
+        else:
+            return ["agrpc", "agrpcs"]
```
YuanTingHsieh, line 260: Should it be "grpc" VS "nagrpc"?
"grpcs" VS "nagrpcs"?

nvflare/fuel/f3/drivers/aio_grpc_driver.py
yanchengnv: The default scheme of FLARE is always "grpc", and we don't want to change that.  On the other hand, each driver must provide its schemes supported. Since we want to have the flexibility of choosing which driver to use for the scheme "grpc", we need to be able to choose it dynamically based on config of "use_aio_grpc".

When this is set to True, then the AioGrpcDriver will take the scheme "grpc", and the  GrpcDriver will take the scheme of "nagrpc" (two drivers cannot have the same scheme). If use_aio_grpc is False, then the GrpcDriver will take the scheme "grpc", and the AioGrpcDriver will take the scheme "agrpc".

When "grpc" is taken by one driver, you could let the other driver always take the same name like "xgrpc". But it is more clear to have their meaning name (agrpc = aio-grpc, nagrpc = non-aio-grpc).

nvflare/fuel/f3/drivers/aio_grpc_driver.py
yanchengnv: See my answers above.

nvflare/fuel/f3/drivers/aio_grpc_driver.py
YuanTingHsieh: Got it, thanks for the explanation!

It might be hard to get these points for regular users, but we could enhance that later if this gets more interests from them.

YuanTingHsieh (approved):

PR #2084, opened 2023-10-17
fix aio exception handling

yanchengnv:
Fixes # .

### Description

- Set exception handler for AIO event loop. This is necessary to gracefully stop AIO tasks that run into exceptions.
- Ensure GRPC channel is not closed multiple times to avoid potential exception

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/fuel/f3/drivers/aio_grpc_driver.py
+++ b/nvflare/fuel/f3/drivers/aio_grpc_driver.py
@@ -69,8 +69,10 @@ class AioStreamSession(Connection):
 
         conf = CommConfigurator()
         if conf.get_bool_var("simulate_unstable_network", default=False):
-            self.disconn = threading.Thread(target=self._disconnect, daemon=True)
-            self.disconn.start()
+            if context:
+                # only server side
+                self.disconn = threading.Thread(target=self._disconnect, daemon=True)
+                self.disconn.start()
 
     def _disconnect(self):
         t = random.randint(10, 60)
@@ -88,9 +90,11 @@ class AioStreamSession(Connection):
             if self.context:
                 self.aio_ctx.run_coro(self.context.abort(grpc.StatusCode.CANCELLED, "service closed"))
                 self.context = None
+                self.logger.info("Closed GRPC context")
             if self.channel:
                 self.aio_ctx.run_coro(self.channel.close())
                 self.channel = None
+                self.logger.info("Closed GRPC Channel")
 
     def send_frame(self, frame: BytesAlike):
         try:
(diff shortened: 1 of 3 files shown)


nvflare/fuel/f3/drivers/aio_grpc_driver.py
```diff
@@ -69,8 +69,10 @@ def __init__(self, aio_ctx: AioContext, connector: ConnectorInfo, conn_props: di
 
         conf = CommConfigurator()
         if conf.get_bool_var("simulate_unstable_network", default=False):
-            self.disconn = threading.Thread(target=self._disconnect, daemon=True)
-            self.disconn.start()
+            if context:
```
YuanTingHsieh, line 72: maybe we could remove/move this testing related code to other place in future PRs

nvflare/fuel/f3/drivers/aio_grpc_driver.py
```diff
@@ -298,57 +302,42 @@ async def _start_connect(self, connector: ConnectorInfo, aio_ctx: AioContext, co
...
-                    if self.closing:
-                        self.logger.debug(
-                            f"Connection {connection} closed by {type(ex)}: {secure_format_exception(ex)}"
-                        )
-                    else:
-                        self.logger.debug(
-                            f"Connection {connection} client read exception {type(ex)}: {secure_format_exception(ex)}"
-                        )
-                    self.logger.debug(secure_format_traceback())
+            connection = AioStreamSession(aio_ctx=aio_ctx, connector=connector, conn_props=conn_props, channel=channel)
```
YuanTingHsieh, line 324: One question, for grpc we call its connection "StreamConnection" but here we call "AioStreamSession",
Why Connection VS Session?
Should we make them consistent in 2.4?
Or they have semantic/logic differences?

YuanTingHsieh (commented):
Mostly good, some questions

nvflare/fuel/f3/drivers/aio_grpc_driver.py
```diff
@@ -88,9 +90,11 @@ def close(self):
             if self.context:
                 self.aio_ctx.run_coro(self.context.abort(grpc.StatusCode.CANCELLED, "service closed"))
```
SYangster, line 91: Do we want to catch exceptions here similar to the close() in grpc_driver.py:73,80?
`self.logger.debug(f"exception aborting GRPC context: {secure_format_exception(ex)}")` and
`self.logger.debug(f"exception closing GRPC channel: {secure_format_exception(ex)}")`

nvflare/fuel/f3/drivers/aio_grpc_driver.py
yanchengnv: Yes, this should become a general-purpose feature for all driver types.

nvflare/fuel/f3/drivers/aio_grpc_driver.py
yanchengnv: No difference at all. Regular grpc driver was actually written before aio version was written. I guess Zhihong changed the name of aio version. Not a big deal.

YuanTingHsieh (approved):
Thanks, LGTM

PR #2088, opened 2023-10-19
Improve grpc 24

yanchengnv:
Fixes # .

### Description

This PR applies 2.3 enhancements to 2.4:
- Improve connection handling of GRPC drivers
- Set event loop's exception handler
- Reduce max wait for connection reconnect
- Enhance credential info
- Remove application conf section when initializing ConfigService. 
- Fixed byte_streamer to not create CommConfigurator during class creation, since the ConfigServer.initialize may not have been called!

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/fuel/f3/drivers/aio_grpc_driver.py
+++ b/nvflare/fuel/f3/drivers/aio_grpc_driver.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import asyncio
+import random
 import threading
 import time
 from typing import Any, Dict, List
@@ -64,7 +65,22 @@ class AioStreamSession(Connection):
         self.conn_props = conn_props
         self.context = context  # for server side
         self.channel = channel  # for client side
+        self.read_task = None
         self.lock = threading.Lock()
+
+        conf = CommConfigurator()
+        if conf.get_bool_var("simulate_unstable_network", default=False):
+            if context:
+                # only server side
+                self.disconn = threading.Thread(target=self._disconnect, daemon=True)
+                self.disconn.start()
+
+    def _disconnect(self):
+        t = random.randint(10, 60)
+        self.logger.info(f"will close connection after {t} secs")
+        time.sleep(t)
+        self.logger.info(f"close connection now after {t} secs")
+        self.close()
 
     def get_conn_properties(self) -> dict:
         return self.conn_props
@@ -79,9 +95,18 @@ class AioStreamSession(Connection):
     def close(self):
         self.closing = True
         with self.lock:
+            if self.read_task:
+                try:
+                    self.logger.info("canceling read_output: connection is closed")
+                    self.read_task.cancel()
+                except Exception as ex:
+                    self.logger.debug(f"exception cancelling read task: {secure_format_exception(ex)}")
+                self.read_task = None
+
             if self.context:
                 self.aio_ctx.run_coro(self._abort())
                 self.context = None
+
             if self.channel:
                 self.aio_ctx.run_coro(self.channel.close())
                 self.channel = None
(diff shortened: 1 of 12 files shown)


nvflare/private/fed/server/server_json_config.py
```diff
@@ -163,7 +161,7 @@ def finalize_config(self, config_ctx: ConfigContext):
         )
 
         ConfigService.initialize(
-            section_files={SystemConfigs.APPLICATION_CONF: os.path.basename(self.config_files[0])},
+            section_files={},
```
YuanTingHsieh, line 164: Do we need this "section_files" arguments in "ConfigService" anymore?
Should we remove it and related get/set sections in follow PRs?

nvflare/private/fed/client/client_json_config.py
```diff
@@ -141,7 +139,7 @@ def finalize_config(self, config_ctx: ConfigContext):
         )
 
         ConfigService.initialize(
-            section_files={SystemConfigs.APPLICATION_CONF: os.path.basename(self.config_files[0])},
+            section_files={},
```
YuanTingHsieh, line 142: Do we need this "section_files" arguments in "ConfigService" anymore?
Should we remove it and related get/set sections in follow PRs?

nvflare/fuel/f3/drivers/aio_grpc_driver.py
```diff
@@ -127,16 +152,29 @@ async def generate_output(self):
...
+        # closing the connection. Once cancelled, "await task" will finish with asyncio.CancelledError exception.
+        with self.lock:
+            if self.closing:
+                raise asyncio.CancelledError("cancelled read_oq: connection closed")
 
-        self.logger.debug(f"{self}: done generate_output")
+            # wrap the queue read into a task
+            task = asyncio.create_task(self.oq.get())
+            self.read_task = task
+        return await task
```
YuanTingHsieh, line 177: We might need to apply this change to 2.3 as well?

nvflare/fuel/f3/drivers/grpc_driver.py
```diff
@@ -262,9 +258,9 @@ def connect(self, connector: ConnectorInfo):
         self.logger.debug("CLIENT: added connection")
         try:
             received = stub.Stream(connection.generate_output())
-            connection.read_loop(received, oq)
-        except BaseException as ex:
-            self.logger.info(f"CLIENT: connection done: {type(ex)}")
+            connection.read_loop(received)
+        except Exception as ex:
+            self.logger.info(f"CLIENT: connection done: {secure_format_exception(ex)}")
         connection.close()
```
YuanTingHsieh, line 264: Should we move 
"connection = StreamConnection(oq, connector, conn_props, "CLIENT", channel=channel)"
inside of the try block as in 2.3 branch?

nvflare/private/fed/server/server_json_config.py
yanchengnv: The idea of the "section_files" is to load commonly needed configs (e.g. server config - not job config) once and let all other places use them. But this idea wasn't used properly. We need to have a consistent config approach. Right now there are different ways. Let's discuss these later. For now let's keep section files.

nvflare/private/fed/client/client_json_config.py
yanchengnv: Before making further changes, let's discuss our overall approach.

nvflare/fuel/f3/drivers/aio_grpc_driver.py
yanchengnv: Yes. I'll leave this to you. :-)

nvflare/fuel/f3/drivers/grpc_driver.py
yanchengnv: Yes, we should, to keep consistent between versions. Will change.

yanchengnv: Changed.

YuanTingHsieh (approved):
Thanks, LGTM

PR #2093, opened 2023-10-24
Fix cell and core cell

YuanTingHsieh:
### Issue

1. There will be a racing condition if the "send_blob" return right away, so we need to setup the waiter first in cell.py
2. The "_msg_size_mbs" needs to be called before "self._send_direct_message" because "self._send_direct_message" will transform the bytes back to objects, then it will lead to "_msg_size_mbs" failed.

### Description

- Move setup waiter in the front of "send_blob" in cell.py
- Move _msg_size_mbs call in the front of "_send_direct_message" in core_cell.py

### Types of changes

- [ ] Documentation updated.


IsaacYangSLA (approved):
Thanks.  The fix on Cell is good.

nvidianz (approved):
LGTM

PR #2100, opened 2023-10-26
Fixes for QA scripts

IsaacYangSLA:
### Description

This PR addresses various issues related to QA scripts.  Mainly to update old codes and to arrange codes in correct order.

### Types of changes

- [ ] Documentation updated.


examples/hello-world/hello-pt/requirements.txt
```diff
@@ -1 +1,3 @@
 nvflare[PT]>=2.3.0
+torch
+torchvision
```
YuanTingHsieh, line 3: the first line already included torch and torchvision
Or should we NOT provide these options ([PT], [HE] etc)?

examples/hello-world/hw_pre_start.sh
```diff
@@ -15,18 +15,18 @@ export NVFLARE_POC_WORKSPACE
...
-python <<END
-from nvflare.tool.api_utils import wait_for_system_start
-wait_for_system_start(num_clients = 2,  prod_dir = "${NVFLARE_POC_WORKSPACE}")
+# # Check if the FL system is ready until 2 clients are active
+# python <<END
+# from nvflare.tool.api_utils import wait_for_system_start
+# wait_for_system_start(username = "admin@nvidia.com", num_clients = 2,  prod_dir = "${NVFLARE_POC_WORKSPACE}/example_project/prod_00")
 
-END
+# END
```
YuanTingHsieh, line 31: maybe we just remove these commented out lines?

examples/advanced/random_forest/utils/prepare_job_config.py
```diff
@@ -110,14 +110,14 @@ def _get_lr_scale_from_split_json(data_split: dict):
...
     # update client config
-    config["executors"][0]["executor"]["args"]["data_split_filename"] = data_split_name
     config["executors"][0]["executor"]["args"]["num_client_bagging"] = args.site_num
     config["executors"][0]["executor"]["args"]["num_local_parallel_tree"] = args.num_local_parallel_tree
     config["executors"][0]["executor"]["args"]["local_subsample"] = args.local_subsample
     config["executors"][0]["executor"]["args"]["lr_scale"] = lr_scale
     config["executors"][0]["executor"]["args"]["lr_mode"] = args.lr_mode
     config["executors"][0]["executor"]["args"]["nthread"] = args.nthread
     config["executors"][0]["executor"]["args"]["tree_method"] = args.tree_method
+    config["components"][0]["args"]["data_split_filename"] = data_split_name
```
YuanTingHsieh, line 120: @ZiyueXu77 @SYangster please help review the random forest example, thanks!

examples/hello-world/hello-pt/requirements.txt
ZiyueXu77: either way works, personally I would do without [PT], but others may prefer the other way

examples/hello-world/hw_pre_start.sh
ZiyueXu77: don't quite get the purpose of this wait_for ..

ZiyueXu77 (commented):
Looks good overall, some comments to your points

examples/hello-world/hello-pt/requirements.txt
IsaacYangSLA: QA scripts also require installing a specific nvflare version.  It could not install RC packages with >=.  Therefore, that nvflare[PT]>=2.3.0 always installs 2.3.8, which is not what qa scripts requires.  We have to remove that line explicitly from requirements.txt and install torch torchvision separately.
I pushed another commit to remove torch torchvision and to install them directly from scripts.

examples/hello-world/hw_pre_start.sh
IsaacYangSLA: This is to ensure the system is up and running.  However, the logic does not work all the time.  So it should be removed.

ZiyueXu77 (approved):

PR #2103, opened 2023-10-27
Make client api pipe configurable

YuanTingHsieh:
### Description

- Make Pipe configurable
- ClientAPILauncherExecutor will determine its own Pipe and choose appropriate Pipe class and arguments for external scripts
- Update all the existing PT example configs that use Client API

### Types of changes

- [ ] Documentation updated.


nvflare/fuel/utils/pipe/pipe_handler.py
```diff
@@ -144,6 +146,14 @@ def stop(self, close_pipe=True):
     def _make_event_message(topic: str, data):
         return Message.new_request(topic, data)
 
+    def wait_for_other_end(self, timeout: Optional[float] = None):
+        start_time = time.time()
+        while True:
+            if timeout and time.time() - start_time > timeout:
+                raise TimeoutError(f"The other end is not up within {timeout} seconds")
+            if self._other_end_is_up:
+                break
```
yhwen, line 155: When self._other_end_is_up is False, this will cause extremely high CPU usage. If you use this approach, you must add a time.sleep() call.

Or it's better to use threading.Event() to wait for other side to end.

yhwen (changes requested):

nvflare/fuel/utils/pipe/pipe_handler.py
YuanTingHsieh: thanks let me change this

nvflare/app_common/executors/client_api_launcher_executor.py
```diff
@@ -24,22 +24,28 @@
...
 from nvflare.fuel.utils.constants import Mode
 from nvflare.fuel.utils.pipe.file_pipe import FilePipe
+from nvflare.fuel.utils.pipe.pipe import Pipe
 from nvflare.fuel.utils.validation_utils import check_object_type
 
 
+def get_pipe_class(pipe: Pipe) -> str:
+    if isinstance(pipe, FilePipe):
+        return "FilePipe"
+    return ""
```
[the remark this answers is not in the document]
YuanTingHsieh: thanks let me change this

nvflare/client/api.py
```diff
@@ -39,36 +39,46 @@ def init(config: Union[str, Dict] = f"config/{CONFIG_EXCHANGE}", rank: Optional[
...
-        dx = FilePipeDataExchanger(
-            supported_topics=client_config.get_supported_topics(),
-            data_exchange_path=client_config.get_exchange_path(),
-            pipe_name=client_config.get_pipe_name(),
-        )
+            dx = DataExchanger(
+                supported_topics=client_config.get_supported_topics(),
+                pipe=pipe,
+                pipe_name=client_config.get_pipe_name(),
+            )
```
yhwen, line 77: Does it miss an "else" condition here?

YuanTingHsieh: @yhwen no, only rank 0 is going to communicate with NVFlare side.

yhwen (approved):
LGTM

PR #2111, opened 2023-10-31
Add None to is_primitive to support null in conf/json files

YuanTingHsieh:
If the json or conf file contains a value of null, for example:

```
{
  "a": 100,
  "b": null
}
```
We should be able to parse / substitute it easily.

Bugs found and suggested fix by @holgerroth 

### Description

- Add value is None to is_primitive method

### Types of changes

- [ ] Documentation updated.


SYangster (approved):

holgerroth (approved):
LGTM

PR #2114, opened 2023-11-01
Update gnn example

ZiyueXu77:
Fixes # .

### Description

For env preparation steps, torch geometric needs special treatment

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/examples/advanced/gnn/README.md
+++ b/examples/advanced/gnn/README.md
@@ -29,6 +29,11 @@ Install additional requirements:
 ```
 python3 -m pip install -r requirements.txt
 ```
+To further support PyTorch Geometric, we need to install in the following manner, please check the torch version and perform the installation accordingly:
+```
+pip install torch-scatter torch-sparse torch-cluster torch-spline-conv torch-geometric -f https://data.pyg.org/whl/torch-2.1.0+cpu.html
+```
+
 #### Protein Classification
 The PPI dataset is directly available via torch_geometric library, we randomly split the dataset to 2 subsets, one for each client.
 First, we run the local training on each client, as well as the whole dataset.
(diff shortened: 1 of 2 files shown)


examples/advanced/gnn/README.md
```diff
@@ -29,6 +29,11 @@ Install additional requirements:
 ```
 python3 -m pip install -r requirements.txt
 ```
+To further support PyTorch Geometric, we need to install in the following manner, please check the torch version and perform the installation accordingly:
+```
+pip install torch-scatter torch-sparse torch-cluster torch-spline-conv torch-geometric -f https://data.pyg.org/whl/torch-2.1.0+cpu.html
```
YuanTingHsieh, line 34: suggest change this to the official documentation about installation:
https://pytorch-geometric.readthedocs.io/en/latest/install/installation.html

examples/advanced/gnn/requirements.txt
```diff
@@ -1,3 +1,5 @@
+nvflare
+torch
+tensorboard
 scikit_learn
```
IsaacYangSLA, line 4: Thanks.  I agree that we might need to quote the steps in the official installation document if it's available.

examples/advanced/gnn/README.md
ZiyueXu77: good point, thanks! updated

examples/advanced/gnn/requirements.txt
ZiyueXu77: sounds good! moved main torch_geometric here

YuanTingHsieh (approved):
Thanks, LGTM

PR #2120, opened 2023-11-04
Print logs and write to logfile at the same time

YuanTingHsieh:
### Description

- Support both printing logs and write to a logfile for the SubprocessLauncher launched processes

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/app_common/launchers/subprocess_launcher.py
+++ b/nvflare/app_common/launchers/subprocess_launcher.py
@@ -26,7 +26,7 @@ from nvflare.app_common.abstract.launcher import Launcher, LauncherCompleteStatu
 
 
 def log_subprocess_output(process, log_file):
-    with open(log_file, "wb") as f:
+    with open(log_file, "ab") as f:
         for c in iter(process.stdout.readline, b""):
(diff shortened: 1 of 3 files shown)


nvflare/app_common/launchers/subprocess_launcher.py
```diff
@@ -47,15 +58,24 @@ def launch_task(self, task_name: str, shareable: Shareable, fl_ctx: FLContext, a
             command = self._script
             env = os.environ.copy()
             command_seq = shlex.split(command)
+            log_file = os.path.join(self._app_dir, f"{task_name}_logfile")
```
SYangster, line 61: If not using `launch_once`, and multiple tasks of the same name are done in succession, then the log file will be overwritten each time. Should we append/have a different log file for each task, or is this ok for now?

YuanTingHsieh: @SYangster good catch!
I remember this in the beginning but at some point I forgot.

I'll do the append mode.

SYangster (approved):
Thanks for the fix, LGTM

PR #2128, opened 2023-11-07
Job CLI Upgrade to support path notation

chesterxgchen:
### Description
1. expand the function to allow key in path format ( dot notation) 
2. add new section on JOB CLI tutorial

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/tool/job/config/configer.py
+++ b/nvflare/tool/job/config/configer.py
@@ -117,6 +117,21 @@ def _cast_type(key_index, cli_value):
     return new_value
 
 
+def split_array_key(key: str) -> Tuple:
+    tokens = key.split("[")
+    if len(tokens) == 1:
+        return None, None, key
+    else:
+        # len(tokens) > 1
+        parent = tokens[0]
+        tokens = tokens[1].split("]")
+        if len(tokens) == 1:
+            raise ValueError(f"invalid key '{key}'")
+        index = int(tokens[0])
+        key = tokens[1].strip(".")
+        return parent, index, key
+
+
 def merge_configs(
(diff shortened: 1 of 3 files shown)


nvflare/tool/job/config/configer.py
```diff
@@ -146,7 +161,26 @@ def merge_configs(
...
+                                    key_value = index_config.get(key)
+                                else:
+                                    config.put(key, cli_value)
+                                    key_value = config.get(key)
+
+                                tokens = key.split(".")
+                                if len(tokens) > 1:
+                                    key = tokens[-1]
+                                key_index = KeyIndex(key, key_value)
+                                key_indices[key] = [key_index]
```
YuanTingHsieh, line 183: For situations like:
`-f xxx.conf args.model.length=100 args.length=50`
The first key is "args.model.length" and the key "length" will be put inside the key_indices dict.

The second key is "args.length" and the key "length" will be put inside the key_indices dict, replacing the value of the first one?

tests/unit_test/tool/job/config/configer_test.py
```diff
@@ -104,3 +104,16 @@ def test_merge_configs(self, origin_job, origin_config, expect_job, expect_confi
         expected_merged = _get_merged_configs(args)
 
         assert result_merged == expected_merged
+
+    def test_split_key(self):
```
YuanTingHsieh, line 108: need to fix this file's format for unit test to pass

nvflare/tool/job/config/configer.py
YuanTingHsieh: thanks for the example, then we can enhance the display later.

YuanTingHsieh (approved):
Thanks, very good fix, we can simplify a lot of job templates.

PR #2138, opened 2023-11-09
Change to get_identity

YuanTingHsieh:
### Description

- Change `get_site_name` to `get_identity` to be consistent with `fl_ctx.get_identity_name()`

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/apis/fl_constant.py
+++ b/nvflare/apis/fl_constant.py
@@ -412,8 +412,8 @@ class FLMetaKey:
     CURRENT_ROUND = "current_round"
     TOTAL_ROUNDS = "total_rounds"
     JOB_ID = "job_id"
-    SITE_NAME = "site_name"
     PROCESS_RC_FILE = "_process_rc.txt"
+    IDENTITY = "identity"
     SUBMIT_MODEL_NAME = "submit_model_name"
 
 
--- a/nvflare/app_common/executors/client_api_launcher_executor.py
+++ b/nvflare/app_common/executors/client_api_launcher_executor.py
@@ -147,7 +147,7 @@ class ClientAPILauncherExecutor(LauncherExecutor):
         config_dict[ConfigKey.PIPE_CLASS] = self._pipe.__class__.__name__
         config_dict[ConfigKey.PIPE_ARGS] = self._get_external_pipe_args()
         config_dict[ConfigKey.TOTAL_ROUNDS] = total_rounds
-        config_dict[ConfigKey.SITE_NAME] = fl_ctx.get_identity_name()
+        config_dict[ConfigKey.IDENTITY] = fl_ctx.get_identity_name()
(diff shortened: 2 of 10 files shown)


nvflare/apis/fl_constant.py
```diff
@@ -412,7 +412,7 @@ class FLMetaKey:
     CURRENT_ROUND = "current_round"
     TOTAL_ROUNDS = "total_rounds"
     JOB_ID = "job_id"
-    SITE_NAME = "site_name"
+    IDENTITY = "identity"
```
yhwen, line 415: There are several places we defined this "SITE_NAME". And since these "key names" have been published in the APIs, and there's quite possibilities that customer's application are using them directly. Even if try our best to change those in our codes, we may still potentially break customer's applications.

I don't see too much values of making this name change. Even if we make a change, we may think just by adding the "IDENTITY" key, while still keeping the "SITE_NAME" to make them referencing the same.

YuanTingHsieh: @yhwen thanks for the review!

Yes, we have other places that use the word SITE_NAME, I did not change those.

This whole FLMetaKey class is added by me in 2.4 just for these FLModel related operations.

So no user is using this, we can safely remove this one.

But it comes to another question, which method name you think is more appropriate "get_site_name" or "get_identity"? Or they are the same? 

@holgerroth @ZiyueXu77 you can also share your thoughts

holgerroth: I like `get_identity`

ZiyueXu77: yes, get_identity sounds better

nvflare/app_common/executors/client_api_launcher_executor.py
```diff
@@ -147,7 +147,7 @@ def prepare_config_for_launch(self, shareable: Shareable, fl_ctx: FLContext):
         config_dict[ConfigKey.PIPE_CLASS] = self._pipe.__class__.__name__
         config_dict[ConfigKey.PIPE_ARGS] = self._get_external_pipe_args()
         config_dict[ConfigKey.TOTAL_ROUNDS] = total_rounds
-        config_dict[ConfigKey.SITE_NAME] = fl_ctx.get_identity_name()
+        config_dict[ConfigKey.IDENTITY] = fl_ctx.get_identity_name()
```
ZiyueXu77, line 150: is it "get_identity()" (without _name)? (according to the import code below `from nvflare.client import get_identity as get_identity`)

YuanTingHsieh: @ZiyueXu77 thanks for the review.

The "get_identity_name" method is from "fl_ctx".

So two different methods.

Do you think it is better to have both methods calling the same name ?

PR #2140, opened 2023-11-09
Fixed the communicator get_task size issue.

yhwen:
Fixes # .

To fix the  communicator not showing the correct get_task and submit_result object size issue.

### Description

Make use of the pympler to calculate the proper object size, including numpy.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/private/fed/client/communicator.py
+++ b/nvflare/private/fed/client/communicator.py
@@ -17,6 +17,8 @@ import socket
 import time
 from typing import List, Optional
 
+from pympler import asizeof
+
 from nvflare.apis.filter import Filter
 from nvflare.apis.fl_constant import FLContextKey
 from nvflare.apis.fl_constant import ReturnCode as ShareableRC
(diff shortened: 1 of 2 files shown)


nvflare/private/fed/client/communicator.py
```diff
@@ -194,8 +196,8 @@ def pull_task(self, servers, project_name, token, ssid, fl_ctx: FLContext):
         return_code = task.get_header(MessageHeaderKey.RETURN_CODE)
 
         if return_code == ReturnCode.OK:
-            size = len(task.payload)
-            task.payload = task.payload
+            size = asizeof.asizeof(task.payload)
```
yanchengnv, line 199: Maybe this is okay here since task.payload is already serialized into bytes or list of bytes...

setup.cfg
```diff
@@ -33,6 +33,7 @@ install_requires =
     docker>=6.0
     websockets>=10.4
     pyhocon
+    pympler==1.0.1
```
yanchengnv, line 37: We may need to include pympler in SWIPAT request.

yanchengnv (commented):
Please add some unit test cases to make sure the sizeof func returns correct values for different types of payload structure (bytes, list of bytes, memory view, etc.).

nvflare/private/fed/client/communicator.py
yanchengnv: Now think of it, I'm not sure this is the right func to use. According to its documentation, this func computes the bytes used for the python object. But we are interested in the size of the message sent. Maybe this func can do it, but if the payload is just bytes or list of bytes, we should be able compute the size easily.

yhwen:
Will use the message header from Cell to solve this issue.

PR #2145, opened 2023-11-14
Support sys vars for job config and support parameterized template in job config

yanchengnv:
Fixes # .

### Description

This PR adds the following enhancements:
- Make system variables (workspace, job id, client name, server root url, secure mode, etc.) available for job config definition. These vars will make it easy to configure CellPipe (separate PR).
- Make os env vars available for job config definition. This can make job args controlled by env vars.
- Added report_env admin command and a func in flare_api to report client site's system variable values. This command is useful for 3rd-party system to accurately configure its CellPipe.
- Fixed status handling in flare_api.
- Fixed missing meta data in some command processing.

### Types of changes

- [ ] Documentation updated.


yanchengnv:
Reopen

IsaacYangSLA:
The main change is in json_config.  @chesterxgchen , does this affect your config with other syntax/tools?

nvflare/fuel/f3/cellnet/core_cell.py
```diff
@@ -460,7 +460,7 @@ def __init__(
         )
 
         self.sent_msg_size_pool = StatsPoolManager.add_msg_size_pool(
-            "Sent_Msg_Sizes", "Sizes of messages sent (MBs)", scope=self.my_info.fqcn
+            "Sent_Msg_sizes", "Sizes of messages sent (MBs)", scope=self.my_info.fqcn
```
YuanTingHsieh, line 463: need to keep the original "Sent_Msg_Sizes"

nvflare/fuel/f3/cellnet/core_cell.py
```diff
@@ -1124,15 +1124,16 @@ def _send_to_endpoint(self, to_endpoint: Endpoint, message: Message) -> str:
...
                     # create a thread and fire the cell's process_message!
                     # self.DIRECT_MSG_EXECUTOR.submit(self._send_direct_message, direct_cell, message)
                     self._send_direct_message(direct_cell, message)
 
                 else:
                     self.communicator.send(to_endpoint, CoreCell.APP_ID, message)
-                self.sent_msg_size_pool.record_value(category=self._stats_category(message), value=msg_size_mbs)
+                self.sent_msg_size_pool.record_value(
+                    category=self._stats_category(message), value=self._msg_size_mbs(message)
+                )
```
YuanTingHsieh, line 1136: need to keep the original change in PR https://github.com/NVIDIA/NVFlare/pull/2093
which means calculate the msg size BEFORE "_send_direct_message"

nvflare/fuel/hci/server/authz.py
```diff
@@ -62,9 +62,6 @@ def pre_command(self, conn: Connection, args: List[str]):
             return True
 
         if return_code == PreAuthzReturnCode.ERROR:
-            conn.append_error(
-                "Authorization error", meta=make_meta(MetaStatusValue.NOT_AUTHORIZED, "Authorization error")
-            )
             return False
```
YuanTingHsieh, line 65: need to keep the original change

nvflare/private/fed/client/client_json_config.py
```diff
@@ -49,20 +50,35 @@ def __init__(self, config_file_name: str, args, app_root: str, kv_list=None, exc
...
+        if kv_list:
+            assert isinstance(kv_list, list), "cmd_vars must be list, but got {}".format(type(kv_list))
+            self.cmd_vars = parse_vars(kv_list)
+        else:
+            self.cmd_vars = {}
+
+        # determine the values of variables that can be used in job config.
+        sp_scheme = args.sp_scheme
+        sp_target = args.sp_target
+        sp_url = f"{sp_scheme}://{sp_target}"
```
YuanTingHsieh, line 62: should SP thing come from Overseer? @IsaacYangSLA

nvflare/private/fed/server/sys_cmd.py
```diff
@@ -154,3 +163,15 @@ def report_resources(self, conn: Connection, args: List[str]):
         table = conn.append_table(["Sites", "Resources"])
         for k, v in site_resources.items():
             table.add_row([str(k), str(v)])
+
+    def report_env(self, conn: Connection, args: List[str]):
+        message = new_message(conn, topic=SysCommandTopic.REPORT_ENV, body="", require_authz=True)
```
YuanTingHsieh, line 168: are we using the arg "client-name" if specified?

nvflare/private/fed/client/client_json_config.py
YuanTingHsieh, line 62: Should we get this from Overseer? @IsaacYangSLA

nvflare/fuel/utils/wfconf.py
```diff
@@ -58,14 +60,96 @@ def process_element(self, node: Node):
...
+            if params:
+                # the var_name must reference a dict
+                local_vars = parse_vars(params)
+                item = self.vars.get(var_name)
+                if item:
+                    if isinstance(item, dict):
+                        # scan the item to resolve var refs
+                        new_item = copy.deepcopy(item)
+                        scanner = JsonScanner(new_item)
+                        new_vars = copy.copy(self.vars)
```
YuanTingHsieh, line 103: Do we need deepcopy for this as well?

YuanTingHsieh (changes requested):
Mostly LGTM.

Some previous fixes were accidentally reverted.

nvflare/fuel/f3/cellnet/core_cell.py
yanchengnv: What do you mean? This "change" was done by the "black" command and I don't see any difference. Do you?

nvflare/fuel/hci/server/authz.py
yanchengnv: No, this one must be removed, otherwise you will get two errors in the console output.

nvflare/private/fed/client/client_json_config.py
yanchengnv: I thought about that. But this is the CJ that was just started by CP, which has the CP from Overseer. Though it is possible, in theory, that when CJ is started, the SP is just changed, but I don't think we need to worry about it.

Besides it's impossible to get SP from Overseer in this case because the job config parsing happens at the very beginning of the process before SP CB happens.

yanchengnv:
> > The main change is in json_config. @chesterxgchen , does this affect your config with other syntax/tools?
> 
> The quick answer is no. Most of the configuration proposed is with double quote "", for example "{pipe}". For pyhocon, within double quote is consider a string.
> 
> But the PR proposed something else seems to be redundant to the existing configuration tools (Besides JSON) such as "{@pipe_def:pipe_id=task_pipe}", "{@pipe_def:pipe_id=metric_pipe}", if you refer to the documentation (https://docs.google.com/document/d/15a4QMQ2dSdyvhEHYG-jTh7Zg100Ge-13lgcRJUtj5YI/edit)
> 
> The effort is try to reduce duplication by introduce configuration merge or configuration replacement. For Pyhocon and Omegacon they are already mechanism to do this. Of course the syntax is different. I am not sure introduce these new syntax make it easier or difficult for user to grasp.

Can you show how Pyhocon can do the CellPipe example in the doc? 
This feature is useful for cases that the same object config is used multiple times. Besides, since most users (if not all) use JSON, I think we need this feature. Of course, users can still choose to do it old-fashioned way - just repeating multiple times.

On the other hand, if we are successful in revolutionizing user's programming experience, the user may no longer need to deal with job config files. We'll see.

nvflare/fuel/f3/cellnet/core_cell.py
yanchengnv: I see. Will fix.

nvflare/private/fed/server/sys_cmd.py
yanchengnv: Not needed here. The self.send_request_to_clients(conn, message) call knows what clients to send the message to, based on client-name arg(s). BTW, this is a common util that is used in many commands.

nvflare/fuel/utils/wfconf.py
yanchengnv: No, because "vars" are not updated. The reason that we need "new_vars" (and cannot use self.vars directly here) is because we update it with "local_vars".

nvflare/fuel/f3/cellnet/core_cell.py
yanchengnv: Fixed

nvflare/fuel/hci/server/authz.py
YuanTingHsieh: Got it, thanks for the explanation.

YuanTingHsieh (dismissed):
Logic-wise LGTM.

As for user experiences, if possible, we could try to make .json and .conf(pyhocon) experience similar.

But I am not familiar with Pyhocon / Omegaconf way of defining things and replacement.

yanchengnv:
The pyhocon's variable substitution mechanism seems very similar to this PR. Very nice. Unfortunately JSON doesn't have it and most of our users still use JSON.
Let's keep this feature here for now and see how things will evolve.

YuanTingHsieh (approved):
Thanks, LGTM

PR #2150, opened 2023-11-16
Added serialized size in the message header

nvidianz:
### Description
1. Added serialized size in message header as PAYLOAD_LEN so caller can get the size.
2. Added format_size function to format size into human-readable format like (MB, MiB etc)
3. Reformatted the getTask message to display correct size.

### Types of changes

- [ ] Documentation updated.


yhwen (dismissed):
LGTM

nvflare/fuel/f3/cellnet/utils.py
```diff
@@ -93,3 +99,38 @@ def decode_payload(message: Message, encoding_key=MessageHeaderKey.PAYLOAD_ENCOD
...
+    else:
+        kilo = 1000.0
+        suffix = "B"
+
+    unit_found = False
+    for unit in ("", "K", "M", "G", "T", "P", "E", "Z"):
+        if abs(num) < kilo:
+            unit_found = True
+            break
+        num /= 1000.0
```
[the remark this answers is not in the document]
YuanTingHsieh: suggest adding unit test cases specifically to this method

nvflare/fuel/f3/cellnet/utils.py
```diff
@@ -79,8 +79,14 @@ def encode_payload(message: Message, encoding_key=MessageHeaderKey.PAYLOAD_ENCOD
             message.payload = fobs.dumps(message.payload, buffer_list=True)
         message.set_header(encoding_key, encoding)
 
+    size = len(message.payload) if message.payload else 0
+    message.set_header(MessageHeaderKey.PAYLOAD_LEN, size)
+
 
 def decode_payload(message: Message, encoding_key=MessageHeaderKey.PAYLOAD_ENCODING):
+    size = len(message.payload) if message.payload else 0
```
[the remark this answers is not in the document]
nvidianz: The header should be there in most cases but we have to deal with the special case that 2 different versions of server/client. This is very cheap so we just do it twice.

nvflare/fuel/f3/cellnet/utils.py
```diff
@@ -93,3 +99,38 @@ def decode_payload(message: Message, encoding_key=MessageHeaderKey.PAYLOAD_ENCOD
...
+    Args:
+        num: Size in bytes
+        binary: If binary, one K is 1024 bytes, otherwise 1000 bytes.
+
+    Returns: Size in human-readable format (like 10MB, 100.2GiB etc)
+
+    """
+
+    if binary:
+        kilo = 1024.0
```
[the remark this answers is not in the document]
nvidianz: So we can get fraction of kilo in the output. For example, 1.5KB

nvflare/fuel/f3/cellnet/utils.py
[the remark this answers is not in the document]
nvidianz: Good catch! Added test cases.

nvflare/fuel/f3/cellnet/utils.py
```diff
@@ -93,3 +99,38 @@ def decode_payload(message: Message, encoding_key=MessageHeaderKey.PAYLOAD_ENCOD
...
+
+    unit_found = False
+    for unit in ("", "K", "M", "G", "T"):
+        if abs(num) < kilo:
+            unit_found = True
+            break
+        num /= kilo
+
+    if not unit_found:
+        unit = "P"
```
YuanTingHsieh, line 130: A question regarding implementation, why "P" needs to be outside of the for loop?

nvflare/fuel/f3/cellnet/utils.py
```diff
@@ -93,3 +99,38 @@ def decode_payload(message: Message, encoding_key=MessageHeaderKey.PAYLOAD_ENCOD
         # assume to be bytes
         pass
     message.remove_header(encoding_key)
+
+
+def format_size(num, binary=False):
```
YuanTingHsieh, line 104: What if the "num" passed in is NOT a integer?
Should we check for that?

tests/unit_test/fuel/f3/cellnet/utils_test.py
```diff
@@ -0,0 +1,41 @@
...
+
+        assert format_size(1000000000, False) == "1GB"
+        assert format_size(1000000000, True) == "953.7MiB"
+        assert format_size(1073741824, False) == "1.1GB"
+        assert format_size(1073741824, True) == "1GiB"
+
+        assert format_size(1000000000000, False) == "1TB"
+        assert format_size(1000000000000, True) == "931.3GiB"
+        assert format_size(1099511627776, False) == "1.1TB"
+        assert format_size(1099511627776, True) == "1TiB"
```
YuanTingHsieh, line 41: Suggest add one case of "PB" and negative cases.

nvflare/fuel/f3/cellnet/utils.py
nvidianz: The loop can only go so far, we choose 'T' as last one in the loop. P is used for any larger numbers. So there must be a line to handle this. I added a test case to demonstrate this.

tests/unit_test/fuel/f3/cellnet/utils_test.py
nvidianz: Added PB, negative, string test cases.

nvflare/fuel/f3/cellnet/utils.py
nvidianz: Added support for numbers in string.

nvflare/private/fed/client/communicator.py
```diff
@@ -266,8 +267,9 @@ def submit_update(
         )
         end_time = time.time()
         return_code = result.get_header(MessageHeaderKey.RETURN_CODE)
+        size = task_message.get_header(MessageHeaderKey.PAYLOAD_LEN)
```
yhwen, line 270: Here the size calculation is not correct. It always got 2. It looks like the header not been set correctly. Could you please check?

yhwen (changes requested):

nvflare/fuel/f3/cellnet/utils.py
```diff
@@ -93,3 +99,39 @@ def decode_payload(message: Message, encoding_key=MessageHeaderKey.PAYLOAD_ENCOD
         # assume to be bytes
         pass
     message.remove_header(encoding_key)
+
+
+def format_size(size, binary=False):
```
yanchengnv, line 104: I thought "binary" should be True by default. 
But I wonder whether we need this option. What's the use case?

nvflare/private/fed/client/communicator.py
```diff
@@ -194,14 +195,14 @@ def pull_task(self, servers, project_name, token, ssid, fl_ctx: FLContext):
         return_code = task.get_header(MessageHeaderKey.RETURN_CODE)
 
         if return_code == ReturnCode.OK:
-            size = len(task.payload)
+            size = task.get_header(MessageHeaderKey.PAYLOAD_LEN)
             task.payload = task.payload
```
yanchengnv, line 199: Please remove this line.

nvflare/private/fed/client/communicator.py
yanchengnv: I think there is an error in the calculation - see my comment in core_cell.py.

yanchengnv (changes requested):
The msg size calculation seems incorrect.

nvflare/fuel/f3/cellnet/utils.py
```diff
@@ -93,3 +99,39 @@ def decode_payload(message: Message, encoding_key=MessageHeaderKey.PAYLOAD_ENCOD
...
+        size: Size in bytes
+        binary: If binary, one K is 1024 bytes, otherwise 1000 bytes.
+
+    Returns: Size in human-readable format (like 10MB, 100.2GiB etc)
+
+    """
+
+    if binary:
+        kilo = 1024.0
+        suffix = "iB"
```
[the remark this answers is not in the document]
nvidianz: For binary Kilo (1024), the standard suffix is MiB, GiB etc.

nvflare/fuel/f3/cellnet/utils.py
nvidianz: To conform with convention and standard, when we say MB, GB, it's non binary mode. That's what we have been using so left it as default.

nvflare/private/fed/client/communicator.py
nvidianz: Fixed.

nvflare/private/fed/client/communicator.py
nvidianz: Done

nvflare/private/fed/client/communicator.py
```diff
@@ -194,14 +195,13 @@ def pull_task(self, servers, project_name, token, ssid, fl_ctx: FLContext):
...
-            size = len(task.payload)
-            task.payload = task.payload
+            size = task.get_header(MessageHeaderKey.PAYLOAD_LEN)
             task_name = task.payload.get_header(ServerCommandKey.TASK_NAME)
             fl_ctx.set_prop(FLContextKey.SSID, ssid, sticky=False)
             if task_name not in [SpecialTaskName.END_RUN, SpecialTaskName.TRY_AGAIN]:
                 self.logger.info(
-                    f"Received from {project_name} server "
-                    f" ({size} Bytes). getTask: {task_name} time: {end_time - start_time} seconds"
+                    f"Received from {project_name} server. getTask: {task_name} size:{format_size(size)} "
```
yanchengnv, line 203: Please add a space btw "size:" and "{format_size(size)".

yanchengnv (changes requested):
LGTM, except for a small change needed in the log msg - see my comment.

nvflare/private/fed/client/communicator.py
nvidianz: Done.

yhwen (approved):

yanchengnv (approved):
LGTM

PR #2156, opened 2023-11-18
Fix simulator

YuanTingHsieh:
### Description

Fix issues introduced by PR #2145 that prevents simulator from running.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/private/fed/app/simulator/simulator_runner.py
+++ b/nvflare/private/fed/app/simulator/simulator_runner.py
@@ -26,6 +26,7 @@ from argparse import Namespace
 from concurrent.futures import ThreadPoolExecutor
 from multiprocessing import Manager, Process
 from multiprocessing.connection import Client
+from urllib.parse import urlparse
 
 from nvflare.apis.fl_component import FLComponent
 from nvflare.apis.fl_constant import JobConstants, MachineStatus, RunProcessKey, WorkspaceConstants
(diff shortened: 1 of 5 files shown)


nvflare/private/fed/app/simulator/simulator_runner.py
```diff
@@ -221,6 +221,11 @@ def setup(self):
             simulator_server, self.server = self.deployer.create_fl_server(self.args)
             # self.services.deploy(self.args, grpc_args=simulator_server)
 
+            url = self.server.cell.get_root_url_for_child()
+            target, scheme = url.split("://")
+            self.args.sp_target = target
+            self.args.sp_scheme = scheme
```
YuanTingHsieh, line 227: not sure if this is the best way to parse URL, @nvidianz what do you think?

nvflare/private/fed/server/server_json_config.py
```diff
@@ -66,7 +66,6 @@ def __init__(self, config_file_name: str, args, app_root: str, kv_list=None, exc
             SystemVarName.JOB_ID: args.job_id,
             SystemVarName.SITE_NAME: "server",
             SystemVarName.WORKSPACE: args.workspace,
-            SystemVarName.ROOT_URL: args.root_url,
             SystemVarName.SECURE_MODE: self.cmd_vars.get("secure_train", True),
```
YuanTingHsieh, line 69: Since we DO NOT pass the root_url as arg when starting server, I remove it here.

Or do we WANT to pass the root_url when server is starting?

yanchengnv: I guess this is fine for now since we don't have a case of using CellPipe on the server side.

yanchengnv (approved):
LGTM

PR #2158, opened 2023-11-20
Support graceful end_run processing

yanchengnv:
Fixes # .

### Description

Currently when the server_runner reaches the end of the work (all workflows are done), it returns and causes the job processing to end immediately. This will cause all communications to stop. The client_runner has the same behavior.

This could be problematic if some components still have unfinished work (e.g. the ServerFedEventRunner may still have unprocessed incoming log records).

This PR tries to fix this potential hole:

- Before ending the process, the runner fires a new event EventType.CHECK_END_RUN_READINESS to "ask" all components whether they are ready to end the run. 
- A component listens to this event. In its event handler, if the component has unfinished work, it sets the FLContextKey.NOT_READY_TO_END_RUN property into the fl_context.
- The runner looks for the FLContextKey.NOT_READY_TO_END_RUN prop from fl_context after the event is done. If no component sets this prop, then everyone is ready to end run; otherwise the runner will wait for a while and then ask again.
- This process keeps repeating until all components are ready to end, or the configured "end_run_readiness_timeout" is reached.
- The ServerFedEventRunner is modified to handle EventType.CHECK_END_RUN_READINESS to ensure that the all incoming events are processed before ending the run.
- Changed to pass the job config data to the runner such that variables defined in the config could be used in the processing logic (e.g. the end_run_readiness_timeout could be defined in the config file).

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/fuel/utils/config_service.py
+++ b/nvflare/fuel/utils/config_service.py
@@ -229,8 +229,27 @@ class ConfigService:
         if env_var_name in os.environ:
             return os.environ.get(env_var_name)
 
-        if isinstance(conf, dict):
-            return conf.get(name)
+        if conf is None:
+            return None
+
+        # conf could be a single config source (a section name or a dict)
+        # conf could also be a list of config sources
+        if not isinstance(conf, list):
+            conf = [conf]
+
+        # check each conf source until the var is found
+        for src in conf:
+            if isinstance(src, str):
+                # this is a section name
+                src = cls.get_section(src)
+
+            if isinstance(src, dict):
+                v = src.get(name)
+                if v is not None:
+                    return v
+
+        # not found from any source
+        return None
 
     @classmethod
     def _int_var(cls, name: str, conf=None, default=None):
--- a/nvflare/fuel/utils/config_service.py
+++ b/nvflare/fuel/utils/config_service.py
@@ -208,18 +208,8 @@ class ConfigService:
             raise TypeError(f"file_basename must be str but got {type(file_basename)}")
         return search_file(file_basename, cls._config_path)
 
-    @classmethod
-    def _get_var(cls, name: str, conf):
-        if not isinstance(name, str):
-            raise ValueError(f"var name must be str but got {type(name)}")
-
-        # see whether command args have it
-        if cls._cmd_args and name in cls._cmd_args:
-            return cls._cmd_args.get(name)
-        if cls._var_dict and name in cls._var_dict:
-            return cls._var_dict.get(name)
-
-        # check OS env vars
+    @staticmethod
+    def _get_var_from_os_env(name: str):
         if not name.startswith(ENV_VAR_PREFIX):
             env_var_name = ENV_VAR_PREFIX + name
         else:
@@ -228,12 +218,17 @@ class ConfigService:
         env_var_name = env_var_name.upper()
         if env_var_name in os.environ:
             return os.environ.get(env_var_name)
+        else:
+            return None
 
+    @classmethod
+    def _get_var_from_config_sources(cls, name: str, conf):
         if conf is None:
             return None
 
-        # conf could be a single config source (a section name or a dict)
-        # conf could also be a list of config sources
+        # conf could be:
+        #   a single config source (a section name or a dict)
+        #   a list of config sources
         if not isinstance(conf, list):
             conf = [conf]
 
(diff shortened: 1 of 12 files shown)


nvflare/private/fed/client/client_runner.py
```diff
@@ -652,6 +655,32 @@ def end_run_events_sequence(self):
...
+                        # we have waited too long
+                        self.log_warning(
+                            fl_ctx, f"quit waiting for component ready-to-end-run after {max_wait} seconds"
+                        )
+                        break
+                    else:
+                        time.sleep(0.5)
+                else:
+                    # all components are ready to end
+                    break
```
yhwen, line 682: This approach of checking if all the components are READY_TO_END_RUN could potentially cause  multiple fire_event(EventType.CHECK_END_RUN_READINESS, fl_ctx) calls. If there's any one component has a long processing task, the client_runner will fire this event to check every 0.5 seconds, which is a lot and maybe high CPU usage.

We can consider to change this to use the threading.Event() for waiting the component to finish tasking run. When firing the event to each component to handle, it creates a threading.Event(). Each component only sets the event if it finishes run. The firing event will wait for this event to be set with a timeout. With this approach, we don't have to call the fire_event(EventType.CHECK_END_RUN_READINESS, fl_ctx) multiple times.

yanchengnv: I'm not sure. First of all, the handler of this event should only check readiness but not do anything heavy. For example, the FedEventRunner (the real use case here) only checks to see whether it has any pending records. The processing of the pending records (the heavy processing) is not done in the event handler. 

Secondly, there can be any number of components/event handlers, managing threading.Event objects for them could be complex (and most of them are useless anyway since they are ready to end anyway).

Firing the events repeatedly shouldn't be a big deal here since checking for readiness should be a short process. Of course anyone could misuse this, but then this is true for all event types.

yanchengnv: Another thing is that the threading.Event based programming is non-trivial.  It may be harder to expect the application code (event handlers and their processing logic) to be able to handle it properly.

nvflare/private/fed/tbi.py
```diff
@@ -0,0 +1,100 @@
...
+        """Check with all components for their readiness to end run
+
+        Args:
+            fl_ctx: the FL context
+
+        Returns:
+
+        """
+        default_max_wait = 5.0
+        default_check_interval = 0.5
```
YuanTingHsieh, line 62: minor thing: maybe we can make this configurable

nvflare/private/fed/tbi.py
```diff
@@ -0,0 +1,100 @@
...
+class TBI(FLComponent):
+
+    """(TBI) Task Based Interaction is the base class for ServerRunner and ClientRunner that implement details of
+    task based interactions.
+
+    TBI implements common behavior of ServerRunner and ClientRunner.
+    """
+
+    def __init__(self):
+        super().__init__()
```
YuanTingHsieh, line 34: minor thing, since we are not adding more stuff in constructor, we can ignore these lines.

YuanTingHsieh (commented):
minor comments

nvflare/private/fed/client/client_runner.py
```diff
@@ -130,11 +131,12 @@ def __init__(
             engine: ClientEngine object
         """
 
-        FLComponent.__init__(self)
+        TBI.__init__(self)
         self.task_router = config.task_router
         self.task_data_filters = config.task_data_filters
         self.task_result_filters = config.task_result_filters
         self.default_task_fetch_interval = config.default_task_fetch_interval
+        self.config = config
```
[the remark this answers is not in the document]
yanchengnv: Yes. Will fix.

nvflare/private/fed/tbi.py
```diff
@@ -0,0 +1,100 @@
...
+
+    def _any_component_is_not_ready(self, fl_ctx: FLContext) -> bool:
+        any_component_not_ready = fl_ctx.get_prop(FLContextKey.NOT_READY_TO_END_RUN, False)
```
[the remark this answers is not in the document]
yanchengnv: True. The component should only set the prop to True when it is not ready to end. 
It's easier for the component to simply raise the NotReadyToEndRun exception.
We support both methods.

nvflare/fuel/utils/config_service.py
```diff
@@ -229,8 +229,27 @@ def _get_var(cls, name: str, conf):
...
+        # check each conf source until the var is found
+        for src in conf:
+            if isinstance(src, str):
+                # this is a section name
+                src = cls.get_section(src)
+
+            if isinstance(src, dict):
+                v = src.get(name)
+                if v is not None:
+                    return v
```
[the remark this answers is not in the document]
yanchengnv: No, it is not.  In your code, the function returns regardless (None or not). In my code, we only return if v is not None.

nvflare/private/fed/tbi.py
yanchengnv: We already make the two variables configurable. The default is the value when there is no configured value for the variable. It is pointless to make the default also configurable. What if it is not configured? You will have to make a default default hard-coded...

nvflare/private/fed/tbi.py
yanchengnv: We may need to do something in the future.

nvflare/private/fed/client/client_runner.py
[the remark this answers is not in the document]
yanchengnv: Fixed

nvflare/fuel/utils/config_service.py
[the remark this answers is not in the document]
YuanTingHsieh: I agree, just this line of code is not equivalent.

But combine with the logic below it is the same.

But this one is minor so we can ignore it

YuanTingHsieh (approved):
Thanks LGTM

YuanTingHsieh (approved):
LGTM,
Breaking large methods into smaller once make it more readable.

The return of "source" could become useful in the future

PR #2161, opened 2023-11-21
Add TaskExchanger and FlareAgent

YuanTingHsieh:
### Description

Rebase Original Changes in https://github.com/yanchengnv/NVFlare/blob/improve_grpc_24 on top of main branch:

- Add TaskExchanger
- Add FlareAgent
- Re-factor Hub related code based on TaskExchanger
- Enhance PipeHandler for status_cb and resend mechanism

New changes:
- Change "FLModelUtils.from_shareable": DON'T need fl_ctx as input argument, shareable should be sufficient
- Add FlareAgentWithFLModel

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/app_common/executors/task_exchanger.py
+++ b/nvflare/app_common/executors/task_exchanger.py
@@ -1,0 +1,202 @@
+# Copyright (c) 2023, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import time
+from typing import Optional
+
+from nvflare.apis.event_type import EventType
+from nvflare.apis.executor import Executor
+from nvflare.apis.fl_constant import FLContextKey
+from nvflare.apis.fl_context import FLContext
+from nvflare.apis.shareable import ReturnCode, Shareable, make_reply
+from nvflare.apis.signal import Signal
+from nvflare.app_common.app_constant import AppConstants
+from nvflare.fuel.utils.constants import PipeChannelName
+from nvflare.fuel.utils.pipe.pipe import Message, Pipe
+from nvflare.fuel.utils.pipe.pipe_handler import PipeHandler
+from nvflare.fuel.utils.validation_utils import check_non_negative_int, check_positive_number, check_str
+from nvflare.security.logging import secure_format_exception
+
+
+class TaskExchanger(Executor):
+    def __init__(
+        self,
+        pipe_id: str,
+        read_interval: float = 0.1,
+        heartbeat_interval: float = 5.0,
+        heartbeat_timeout: Optional[float] = 30.0,
+        resend_interval: float = 2.0,
+        max_resends: Optional[int] = None,
+        peer_read_timeout: Optional[float] = 5.0,
+        task_wait_time: Optional[float] = None,
+        result_poll_interval: float = 0.5,
+        pipe_channel_name=PipeChannelName.TASK,
+    ):
+        """Constructor of TaskExchanger.
+
+        Args:
+            pipe_id: component id of pipe
+            read_interval: how often to read from pipe
+            heartbeat_interval: how often to send heartbeat to peer
+            heartbeat_timeout: max amount of time to allow missing heartbeats before treating peer as dead
+            resend_interval: how often to resend a message when failing to send
+            max_resends: max number of resends. None means no limit
+            peer_read_timeout: time to wait for peer to accept sent message
+            task_wait_time: how long to wait for a task to complete. None means waiting forever
+            result_poll_interval: how often to poll task result
+            pipe_channel_name: the channel name for sending task requests
+        """
+        Executor.__init__(self)
+        check_str("pipe_id", pipe_id)
+        check_positive_number("read_interval", read_interval)
+        check_positive_number("heartbeat_interval", heartbeat_interval)
+        if heartbeat_timeout is not None:
+            check_positive_number("heartbeat_timeout", heartbeat_timeout)
+        check_positive_number("resend_interval", resend_interval)
+        if max_resends is not None:
+            check_non_negative_int("max_resends", max_resends)
+        if peer_read_timeout is not None:
+            check_positive_number("peer_read_timeout", peer_read_timeout)
+        if task_wait_time is not None:
+            check_positive_number("task_wait_time", task_wait_time)
+        check_positive_number("result_poll_interval", result_poll_interval)
+        check_str("pipe_channel_name", pipe_channel_name)
+
+        self.pipe_id = pipe_id
+        self.read_interval = read_interval
+        self.heartbeat_interval = heartbeat_interval
+        self.heartbeat_timeout = heartbeat_timeout
+        self.resend_interval = resend_interval
+        self.max_resends = max_resends
+        self.peer_read_timeout = peer_read_timeout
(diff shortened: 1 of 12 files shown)


nvflare/app_common/utils/fl_model_utils.py
```diff
@@ -93,9 +92,7 @@ def to_shareable(fl_model: FLModel, params_converter: Optional[ParamsConverter]
         return shareable
 
     @staticmethod
-    def from_shareable(
-        shareable: Shareable, params_converter: Optional[ParamsConverter] = None, fl_ctx: Optional[FLContext] = None
-    ) -> FLModel:
+    def from_shareable(shareable: Shareable, params_converter: Optional[ParamsConverter] = None) -> FLModel:
```
yanchengnv, line 95: I'm not sure about this. There is a lot of info in fl_ctx that could be needed in FLModel in the future. We may not need much from fl_ctx right now, but that doesn't mean we'll never need it. I think we should keep fi_ctx here.

nvflare/client/flare_agent.py
```diff
@@ -0,0 +1,356 @@
...
+        pipe_handler.stop()
+
+    def stop(self):
+        """Stop the agent. After this is called, there will be no more communications between CJ and agent.
+
+        Returns: None
+
+        """
+        self.asked_to_stop = True
+        self.pipe_handler.stop()
```
yanchengnv, line 148: To be consistent, we should also stop self.metric_pipe_handler (if it is not None).

nvflare/client/flare_agent.py
```diff
@@ -0,0 +1,356 @@
...
+
+    def get_task(self, timeout: Optional[float] = None):
+        """Get a task from FLARE. This is a blocking call.
+
+        Args:
+            timeout (float, optional): If specified, this call is blocked only for the specified amount of time.
+                If not specified, this call is blocked forever until a task has been received or agent has been closed.
+
+        Returns:
+            None if no task is available during before timeout; or a Task object if task is available.
```
yanchengnv, line 181: Remove "during".

nvflare/client/flare_agent.py
```diff
@@ -0,0 +1,356 @@
...
+        start_time = time.time()
+        while True:
+            if self.asked_to_stop:
+                raise AgentClosed("agent closed")
+
+            if self.current_task:
+                raise CallStateError("application called get_task while the current task is not processed")
+
+            if timeout is not None and time.time() - start_time >= timeout:
+                self.pipe_handler.notify_abort()
```
yanchengnv, line 202: Why should we notify_abort here? This is the case that the caller only wants to wait for some time, it doesn't mean that the caller is aborting.

nvflare/fuel/utils/pipe/pipe_handler.py
```diff
@@ -149,53 +223,70 @@ def _make_event_message(topic: str, data):
...
     def _try_read(self):
         last_heartbeat_received_time = time.time()
         last_heartbeat_sent_time = 0.0
         while not self.asked_to_stop:
             now = time.time()
-            msg = self._receive_from_pipe()
+            msg = self.pipe.receive()
             if msg:
                 last_heartbeat_received_time = now
                 self._other_end_is_up.set()
```
yanchengnv, line 292: This event is used to indicate that the other end (the peer) is up. You need to check the msg's topic and decide whether this is truly a message from the other end.  Only if the topic is not one of (PEER_GONE, ABORT, END), you can say the peer is up.

yanchengnv (changes requested):
Looks good in general, but some changes are needed - see my comments.

nvflare/fuel/utils/pipe/pipe_handler.py
```diff
@@ -149,53 +223,70 @@ def _make_event_message(topic: str, data):
     def wait_for_other_end(self, timeout: Optional[float] = None):
```
yanchengnv, line 223: What is this method for? Are you imagining that the caller can wait for the peer to come up?

nvflare/app_common/utils/fl_model_utils.py
YuanTingHsieh: I added this "FlareAgentWithFLModel", in this class I use these methods (from_shareable, to_shareable) to convert the shareable to FLModel and vice versa.

Since the FlareAgentWithFLModel is running on 3rd party, it does not have FLContext though.

I do have a case that need something from fl_ctx in the case of LauncherExecutor, what I did is that I set the shareable header first before sending it via pipe to the 3rd party.

nvflare/client/flare_agent.py
YuanTingHsieh: thanks, will do

nvflare/client/flare_agent.py
YuanTingHsieh: thanks, will do

nvflare/client/flare_agent.py
YuanTingHsieh: Make sense, let me fix it.

nvflare/app_common/utils/fl_model_utils.py
yanchengnv: You can pass None for fl_ctx in the case that you don't have it. Just make sure that when you use prop from fl_ctx, check for None first.

nvflare/fuel/utils/pipe/pipe_handler.py
YuanTingHsieh: In the LauncherExecutor case,

There is a situation where the 3rd party is launched by subprocess.Popen,
then it takes a long time for it to get to the PipeHandler.start (data loading, model setup takes a long time)

In that scenario, if the LauncherExecutor sends the task right away and wait for "peer_read_timeout", it can easily times out. (since the other end is still setting up data, model)

I add this so that the LauncherExecutor will only sends the task AFTER the other side's PipeHandler is up and running.

nvflare/app_common/utils/fl_model_utils.py
YuanTingHsieh: OK, let me keep the signature so if we want more stuff from fl_ctx we can do that.

nvflare/fuel/utils/pipe/pipe_handler.py
YuanTingHsieh: good point, fixed.

nvflare/fuel/utils/pipe/pipe_handler.py
yanchengnv: I see. Make sense.

yanchengnv (approved):
LGTM

nvflare/app_common/executors/task_exchanger.py
```diff
@@ -0,0 +1,202 @@
...
+from nvflare.apis.signal import Signal
+from nvflare.app_common.app_constant import AppConstants
+from nvflare.fuel.utils.constants import PipeChannelName
+from nvflare.fuel.utils.pipe.pipe import Message, Pipe
+from nvflare.fuel.utils.pipe.pipe_handler import PipeHandler
+from nvflare.fuel.utils.validation_utils import check_non_negative_int, check_positive_number, check_str
+from nvflare.security.logging import secure_format_exception
+
+
+class TaskExchanger(Executor):
```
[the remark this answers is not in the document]
YuanTingHsieh: It will be like this:

LauncherExecutor extends TaskExchanger , so they are on the nvflare client side.

Our client api will use FlareAgent, they are on the 3rd party side (user custom code side)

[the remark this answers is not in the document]
YuanTingHsieh: Yes, so this is the first PR.

That refactoring will come the next PR along with changes to api

It need to change both left side and right side for things to work.

nvflare/client/flare_agent.py
```diff
@@ -0,0 +1,358 @@
...
+                resend_interval=resend_interval,
+                max_resends=max_resends,
+            )
+
+        self.current_task = None
+        self.task_lock = threading.Lock()
+        self.asked_to_stop = False
+
+    def start(self):
+        """Start the agent. This method must be called to enable CJ/Agent communication.
```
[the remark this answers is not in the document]
YuanTingHsieh: Will fix this docstring in follow up PR

PR #2164, opened 2023-11-23
Refactor LauncherExecutor and client api

YuanTingHsieh:
### Description

Added:
- Added FlareAgentWithFLModel
- Refactored LauncherExecutor and client api based on TaskExchanger and FlareAgentWithFLModel
- Refactored Launcher and move the "launch_once" argument to SubprocessLauncher
- Added a new job template "sag_np_cell_pipe"
- Updated "sag_np" / "sag_pt" / "sag_pt_deploy_map" / "sag_tf" job templates

Removed:
- Removed DataExchanger and its tests
- Removed ExchangeTask

### Types of changes

- [ ] Documentation updated.


nvflare/app_common/data_exchange/piper.py
```diff
@@ -0,0 +1,85 @@
...
+            args["site_name"] = fl_ctx.get_identity_name()
+            args["token"] = self._pipe.token
+            args["root_url"] = self._pipe.cell.core_cell.root_url
+            args["secure_mode"] = self._pipe.cell.core_cell.secure
+            args["workspace_dir"] = fl_ctx.get_engine().get_workspace().get_root_dir()
+        else:
+            raise RuntimeError(f"Pipe ({self._pipe_id}) of type ({type(self._pipe)}) is not supported")
+        return args
+
+    def init_pipe(self, fl_ctx: FLContext) -> None:
```
yanchengnv, line 63: Not sure what this is for. The pipe object should have been created already based on the config args in the config file. Why do you want to init again?

nvflare/app_common/data_exchange/piper.py
```diff
@@ -0,0 +1,85 @@
...
+
+        # gets Pipe using _pipe_id
+        pipe = engine.get_component(self._pipe_id)
+        check_object_type(self._pipe_id, pipe, Pipe)
+
+        if isinstance(pipe, FilePipe):
+            if pipe.root_path is None or pipe.root_path == "":
+                app_dir = engine.get_workspace().get_app_dir(fl_ctx.get_job_id())
+                pipe.root_path = os.path.abspath(app_dir)
+        elif isinstance(pipe, CellPipe):
```
yanchengnv, line 74: You are setting the object's args directly here. In theory, this is not the same as init the object - the init method of the object could have processing logic in it.

nvflare/app_common/executors/launcher_executor.py
```diff
@@ -95,51 +101,38 @@ def __init__(
...
             self._job_end = True
+            self._error_happened.clear()
             self.launcher.finalize(fl_ctx)
             self.log_info(fl_ctx, "END_RUN event received - telling external to stop")
-            if self.pipe_handler is not None:
-                self.pipe_handler.notify_end("END_RUN event received")
-                self.pipe_handler.stop(close_pipe=True)
+            super().handle_event(event_type, fl_ctx)
 
     def execute(self, task_name: str, shareable: Shareable, fl_ctx: FLContext, abort_signal: Signal) -> Shareable:
```
yanchengnv, line 137: This code should be re-organized for better readability. This executor has the following 3 major steps:
- Launch the 3rd-party program
- Execute the task
- End the 3rd-party program

nvflare/app_common/executors/launcher_executor.py
```diff
@@ -161,67 +154,81 @@ def execute(self, task_name: str, shareable: Shareable, fl_ctx: FLContext, abort
...
+                target=self._check_last_result_transfer,
+                args=(
+                    fl_ctx,
+                    abort_signal,
+                ),
+            )
+            self._check_last_result_transfer_thread.start()
 
         # pipe handler starts checking for heartbeats only after the 3rd party code has been launched
         self.pipe_handler.start()
```
yanchengnv, line 174: You shouldn't directly access the "pipe_handler", which is part of the TaskExchanger. 
This pipe_handler should be started by the TaskExchanger already, no?

nvflare/app_common/executors/launcher_executor.py
```diff
@@ -161,67 +154,81 @@ def execute(self, task_name: str, shareable: Shareable, fl_ctx: FLContext, abort
...
-        result = self._exchange(task_name, shareable, fl_ctx, abort_signal)
-        self._result_fl_model = None
-        self._result_metrics = None
+        # wait for external process to setup their pipe_handler
+        setup_success = self._wait_external_setup(task_name, fl_ctx, abort_signal)
+        if not setup_success:
+            self.log_error(fl_ctx, "External process set up failed.")
+            return make_reply(ReturnCode.EXECUTION_EXCEPTION)
+
+        shareable.set_header(FLMetaKey.JOB_ID, fl_ctx.get_job_id())
```
yanchengnv, line 182: You should let the TaskExchanger handle common things like these (setting shareable headers).

nvflare/app_common/executors/launcher_executor.py
```diff
@@ -161,67 +154,81 @@ def execute(self, task_name: str, shareable: Shareable, fl_ctx: FLContext, abort
...
+        if check_result != "":
+            self.log_error(fl_ctx, check_result)
+            return make_reply(ReturnCode.EXECUTION_EXCEPTION)
+        self._received_result = True
 
         if not self._launch_once or self._job_end:
             req = Message.new_request(topic=Topic.END, data="END")
-            has_been_read = self.pipe_handler.send_to_peer(req, timeout=self._peer_read_timeout)
-            if self._peer_read_timeout and not has_been_read:
+            has_been_read = self.pipe_handler.send_to_peer(req, timeout=self.peer_read_timeout)
```
yanchengnv, line 194: Shouldn't access pipe_handler of TaskExchanger. You can enhance TaskExchanger to do this by adding a method like ask_peer_to_end(). This way, you let TaskExchanger deal with all interaction needs with the peer. You only need to focus on managing the launcher and task exchanger.

nvflare/app_common/executors/launcher_executor.py
```diff
@@ -161,67 +154,81 @@ def execute(self, task_name: str, shareable: Shareable, fl_ctx: FLContext, abort
...
                 self.log_warning(
                     fl_ctx,
-                    f"3rd party does not get END msg in {self._peer_read_timeout} secs!",
+                    f"3rd party does not get END msg in {self.peer_read_timeout} secs!",
                 )
                 return make_reply(ReturnCode.EXECUTION_EXCEPTION)
             launch_finish = self._wait_launch_finish(task_name, shareable, fl_ctx, abort_signal)
             if not launch_finish:
                 return make_reply(ReturnCode.EXECUTION_EXCEPTION)
+            self._check_last_result_transfer_thread.join()
```
yanchengnv, line 204: Why do we need this? After super().execute(), the task is done. What are we waiting for here?

nvflare/app_common/launchers/subprocess_launcher.py
```diff
@@ -58,28 +46,23 @@ def launch_task(self, task_name: str, shareable: Shareable, fl_ctx: FLContext, a
...
                 cwd=self._app_dir,
                 env=env,
             )
-            self._log_thread = Thread(target=log_subprocess_output, args=(self._process, log_file))
-            self._log_thread.start()
             return True
         return False
 
-    def wait_task(self, task_name: str, fl_ctx: FLContext, timeout: Optional[float] = None) -> LauncherCompleteStatus:
+    def wait_task(self, task_name: str, fl_ctx: FLContext, timeout: Optional[float] = None) -> LauncherRunStatus:
```
yanchengnv, line 59: Since LauncherRunStatus is no longer enum, the return type here should be str.

nvflare/app_common/launchers/subprocess_launcher.py
```diff
@@ -90,3 +73,13 @@ def stop_task(self, task_name: str, fl_ctx: FLContext) -> None:
                 process = subprocess.Popen(command_seq, cwd=self._app_dir)
                 process.wait()
             self._process = None
+
+    def check_run_status(self, task_name: str, fl_ctx: FLContext) -> LauncherRunStatus:
```
yanchengnv, line 77: Return type should be str.

nvflare/fuel/utils/pipe/pipe_handler.py
```diff
@@ -289,8 +287,9 @@ def _try_read(self):
             msg = self.pipe.receive()
             if msg:
                 last_heartbeat_received_time = now
-                if msg.topic not in [Topic.END, Topic.ABORT, Topic.PEER_GONE]:
-                    self._other_end_is_up.set()
+                # if receive any messages even if Topic is END or ABORT or PEER_GONE
+                #    we still set other_end_is_up_or_dead, as we no longer needs to wait
```
yanchengnv, line 291: Typo in comment:  "needs" => "need"

yanchengnv (changes requested):
The LauncherExecutor needs to be better organized.
Enhance TaskExchanger with functions needed by LauncherExecutor and avoid accessing internal components of TaskExchanger.

job_templates/sag_np_cell_pipe/config_fed_client.conf
```diff
@@ -0,0 +1,88 @@
...
+    },
+    {
+      id = "pipe"
+      path = "nvflare.fuel.utils.pipe.cell_pipe.CellPipe"
+      args {
+        mode = "PASSIVE"
+        site_name = "{SITE_NAME}"
+        token = "pipe"
+        root_url = "{ROOT_URL}"
+        secure_mode = False
```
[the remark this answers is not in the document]
YuanTingHsieh: If I use secure_mode = "{SECURE_MODE}",
The simulator will make this be True, we might need to fix that first.
Other replacements work good with simulator.

nvflare/app_common/abstract/launcher.py
```diff
@@ -58,7 +59,7 @@ def launch_task(self, task_name: str, shareable: Shareable, fl_ctx: FLContext, a
         pass
 
     @abstractmethod
-    def wait_task(self, task_name: str, fl_ctx: FLContext, timeout: Optional[float] = None) -> LauncherCompleteStatus:
+    def wait_task(self, task_name: str, fl_ctx: FLContext, timeout: Optional[float] = None) -> LauncherRunStatus:
```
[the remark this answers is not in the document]
YuanTingHsieh: fixed

nvflare/app_common/abstract/launcher.py
```diff
@@ -80,3 +81,8 @@ def stop_task(self, task_name: str, fl_ctx: FLContext) -> None:
             fl_ctx (FLContext): fl context.
         """
         pass
+
+    @abstractmethod
+    def check_run_status(self, task_name: str, fl_ctx: FLContext) -> LauncherRunStatus:
```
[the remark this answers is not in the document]
YuanTingHsieh: fixed

nvflare/app_common/data_exchange/data_exchanger.py
```diff
@@ -1,179 +0,0 @@
-# Copyright (c) 2023, NVIDIA CORPORATION.  All rights reserved.
```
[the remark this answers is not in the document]
YuanTingHsieh: yes

nvflare/app_common/data_exchange/piper.py
YuanTingHsieh: correct, before the variable substitution you added, I am using some hacks here to replace empty root_path, removed this.

nvflare/app_common/data_exchange/piper.py
YuanTingHsieh: correct.
Before the variable substitution you added, I am using some hacks here to replace empty root_path.
I removed this.

nvflare/app_common/executors/launcher_executor.py
YuanTingHsieh: re-organized as suggested.

nvflare/app_common/executors/launcher_executor.py
YuanTingHsieh: removed this line.

nvflare/app_common/executors/launcher_executor.py
YuanTingHsieh: change as suggested.

nvflare/app_common/launchers/subprocess_launcher.py
YuanTingHsieh: thanks, fixed.

nvflare/app_common/launchers/subprocess_launcher.py
YuanTingHsieh: thanks, fixed.

nvflare/fuel/utils/pipe/pipe_handler.py
YuanTingHsieh: thanks, fixed.

nvflare/client/api.py
```diff
@@ -39,46 +41,49 @@ def init(config: Union[str, Dict] = f"config/{CONFIG_EXCHANGE}", rank: Optional[
...
+            if ok:
+                fobs.register(tensor_decomposer)
             else:
-                raise RuntimeError(f"Pipe class {client_config.get_pipe_class()} is not supported.")
+                raise RuntimeError(f"Can't import TensorDecomposer for format: {ExchangeFormat.PYTORCH}")
+            from_nvflare_converter = NumpyToPTParamsConverter()
+            to_nvflare_converter = PTToNumpyParamsConverter()
+
+        pipe_args = client_config.get_pipe_args()
+        if client_config.get_pipe_class() == "FilePipe":
```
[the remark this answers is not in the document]
YuanTingHsieh: thanks, re-factored to use dict as you suggested.

nvflare/app_common/data_exchange/piper.py
```diff
@@ -0,0 +1,85 @@
...
+        if isinstance(self._pipe, FilePipe):
+            return "FilePipe"
+        elif isinstance(self._pipe, CellPipe):
+            return "CellPipe"
+        else:
+            raise RuntimeError(f"Pipe ({self._pipe_id}) of type ({type(self._pipe)}) is not supported")
+
+    def get_external_pipe_args(self, fl_ctx: FLContext) -> dict:
+        args = {}
+        if isinstance(self._pipe, FilePipe):
```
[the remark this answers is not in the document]
YuanTingHsieh: good suggestion, re-factored.

nvflare/app_common/data_exchange/piper.py
```diff
@@ -0,0 +1,85 @@
...
+        Args:
+            piper_id (str): id of the Piper
+            pipe_id (str): the id to be used to get the Pipe component from engine.
+        """
+        self._pipe = None
+        self._pipe_id = pipe_id
+
+    def get_external_pipe_class(self) -> str:
+        if isinstance(self._pipe, FilePipe):
+            return "FilePipe"
```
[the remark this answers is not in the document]
YuanTingHsieh: No, it only needs to match the key of the class mapping in: nvflare/client/api.py

nvflare/app_common/executors/launcher_executor.py
YuanTingHsieh: thanks, fixed as suggested.

job_templates/sag_np_cell_pipe/config_fed_client.conf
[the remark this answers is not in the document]
YuanTingHsieh: @yhwen we need to fix the simulator secure mode , I think it should be False by default

examples/hello-world/ml-to-fl/np/README.md
```diff
@@ -108,3 +108,20 @@ Then we can run it using the NVFlare Simulator:
 nvflare simulator -n 2 -t 2 ./jobs/np_loop -w np_loop_workspace
 ```
 
+## Using CellPipe instead of FilePipe
+
```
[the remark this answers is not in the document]
YuanTingHsieh: Thanks, added descriptions as suggested.

nvflare/app_common/executors/launcher_executor.py
YuanTingHsieh: Thanks, changed name to "_monitor_launcher"

nvflare/app_common/executors/launcher_executor.py
```diff
@@ -78,149 +77,128 @@ def __init__(
...
-            launch_finish = self._wait_launch_finish(task_name, shareable, fl_ctx, abort_signal)
-            if not launch_finish:
-                return make_reply(ReturnCode.EXECUTION_EXCEPTION)
-            self._clear_state()
-            self.log_info(fl_ctx, f"Launched external process for task ({task_name}) is finished.")
+                return False
+        return True
 
-        return result
+    def check_output_shareable(self, task_name: str, shareable: Shareable, fl_ctx: FLContext) -> bool:
```
yanchengnv, line 185: Where is this called? It should be called in the "execute" method, right? But it's not called there.

nvflare/app_common/executors/launcher_executor.py
```diff
@@ -78,149 +77,128 @@ def __init__(
...
+        if not self._end_external_process(task_name, shareable, fl_ctx, abort_signal):
+            return make_reply(ReturnCode.EXECUTION_EXCEPTION)
+
+        return result
+
+    def stop(self, task_name: str, fl_ctx: FLContext):
+        """Stops the LauncherExecutor."""
+        self._stop_launcher(task_name, fl_ctx)
+
+    def check_input_shareable(self, task_name: str, shareable: Shareable, fl_ctx: FLContext) -> bool:
```
yanchengnv, line 167: This should be called in "execute", but don't see it.

nvflare/app_common/executors/launcher_executor.py
```diff
@@ -243,6 +221,56 @@ def _init_converter(self, fl_ctx: FLContext):
...
+            if self.peer_is_up_or_dead():
+                return True
+
+            if self.launcher.check_run_status(task_name, fl_ctx) != LauncherRunStatus.RUNNING:
+                return False
+
+            time.sleep(0.1)
+
     def _launch(self, task_name: str, shareable: Shareable, fl_ctx: FLContext, abort_signal: Signal) -> bool:
         future = self._thread_pool_executor.submit(self._launch_task, task_name, shareable, fl_ctx, abort_signal)
```
yanchengnv, line 275: Why do you need another thread to do this? You are not doing anything in parallel.

nvflare/app_common/executors/launcher_executor.py
```diff
@@ -266,7 +294,7 @@ def _wait_launch_finish(
         future = self._thread_pool_executor.submit(self._wait_launcher, task_name, fl_ctx, self._wait_timeout)
```
yanchengnv, line 294: Why another thread?

nvflare/app_common/executors/launcher_executor.py
```diff
@@ -78,149 +77,128 @@ def __init__(
...
-        self._launched = False
+        self._launched = Event()
         self._launcher_finish = Event()
         self._launcher_finish_status = None
         self._launcher_finish_time = None
         self._last_result_transfer_timeout = last_result_transfer_timeout
+        self._received_result = False
         self._job_end = False
 
         self._thread_pool_executor = ThreadPoolExecutor(max_workers=workers, thread_name_prefix=self.__class__.__name__)
```
yanchengnv, line 108: Don't think you need this thread_pool_executor.

nvflare/app_common/executors/task_exchanger.py
```diff
@@ -190,13 +194,55 @@ def execute(self, task_name: str, shareable: Shareable, fl_ctx: FLContext, abort
...
+        """
+        return True
+
+    def ask_peer_to_end(self, fl_ctx: FLContext) -> bool:
+        req = Message.new_request(topic=Topic.END, data="END")
+        has_been_read = self.pipe_handler.send_to_peer(req, timeout=self.peer_read_timeout)
+        if self.peer_read_timeout and not has_been_read:
+            self.log_warning(
+                fl_ctx,
+                f"3rd party does not get END msg in {self.peer_read_timeout} secs!",
```
yanchengnv, line 238: Change to "... does not read END msg ..."

nvflare/client/flare_agent.py
```diff
@@ -343,16 +346,11 @@ def __init__(
 
 
 class FlareAgentWithFLModel(FlareAgent):
```
yanchengnv, line 348: Please move this class to a separate file so that the flare_agent.py will be more generic (and not importing those FLModel stuff).

nvflare/fuel/utils/pipe/pipe_handler.py
```diff
@@ -109,7 +110,7 @@ def __init__(
         self.msg_cb = None
         self.msg_cb_args = None
         self.msg_cb_kwargs = None
-        self._other_end_is_up = threading.Event()
+        self.other_end_is_up_or_dead = threading.Event()
```
yanchengnv, line 113: Rename to "peer_is_up_or_dead" since we have been using "peer" to mean "the other end".

yanchengnv (changes requested):
See my comments. The LauncherExecutor still needs some work.

nvflare/app_common/data_exchange/piper.py
```diff
@@ -0,0 +1,63 @@
...
+                "root_url": lambda pipe, fl_ctx: pipe.cell.get_root_url_for_child(),
+                "secure_mode": lambda pipe, fl_ctx: pipe.cell.core_cell.secure,
+                "workspace_dir": lambda pipe, fl_ctx: fl_ctx.get_engine().get_workspace().get_root_dir(),
+            },
+        },
+    }
+
+    @classmethod
+    def get_external_pipe_class(cls, pipe_id: str, fl_ctx: FLContext) -> str:
+        pipe = fl_ctx.get_engine().get_component(pipe_id)
```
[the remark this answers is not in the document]
YuanTingHsieh: let me add a check here as well.

nvflare/app_common/executors/launcher_executor.py
YuanTingHsieh: it is called in base class TaskExchanger's execute method.

nvflare/app_common/executors/launcher_executor.py
YuanTingHsieh: I was using this because I want to have the pattern below: `future.result(timeout=xxx)`

I can change this to a while loop and avoid the `ThreadPoolExecutor`

YuanTingHsieh: After discussion, I remember the original reason I did this is because I assume when users write "Launcher", they could make mistake saying a infinite loop in "launch_task" implementation, that's why we use a different thread to end it.

Let me simplify the code and add more comments.

YuanTingHsieh: wraps into "_execute_launcher_method_in_thread_executor" for clarity.

nvflare/fuel/utils/pipe/pipe_handler.py
```diff
@@ -109,7 +110,7 @@ def __init__(
         self.msg_cb = None
         self.msg_cb_args = None
         self.msg_cb_kwargs = None
-        self._other_end_is_up = threading.Event()
+        self.peer_is_up_or_dead = threading.Event()
```
yanchengnv, line 114: Need pause/resume methods for heartbeat mechanism.

YuanTingHsieh: Added pause/resume function and tested with CyclicController

nvflare/app_common/executors/task_exchanger.py
```diff
@@ -190,13 +193,51 @@ def execute(self, task_name: str, shareable: Shareable, fl_ctx: FLContext, abort
...
+            self.log_warning(
+                fl_ctx,
+                f"3rd party does not read END msg in {self.peer_read_timeout} secs!",
+            )
+            return False
+        return True
+
+    def peer_is_up_or_dead(self) -> bool:
+        return self.pipe_handler.peer_is_up_or_dead.is_set()
+
```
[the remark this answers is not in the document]
YuanTingHsieh: Added pause/resume function and tested with CyclicController

nvflare/app_common/executors/launcher_executor.py
```diff
@@ -55,181 +52,154 @@ def __init__(
...
+
+        if not self._initialize_external_execution(task_name, shareable, fl_ctx, abort_signal):
+            return make_reply(ReturnCode.EXECUTION_EXCEPTION)
+
+        if self._from_nvflare_converter is not None:
+            shareable = self._from_nvflare_converter.process(shareable, fl_ctx)
+
+        self.resume_pipe_handler()
+        result = super().execute(task_name, shareable, fl_ctx, abort_signal)
+        self.pause_pipe_handler()
```
yanchengnv, line 152: I don't think you should pause pipe handler unconditionally. You should do so only if the launch mode is per-task.
If the mode is launch once, the FlareAgent in 3rd-party will be running and expecting heartbeat from TaskExchanger. But if HB is stopped since pipe handler is paused here, the FlareAgent won't be happy.

yanchengnv (changes requested):
The pause/resume logic isn't quite right. See my comments.

nvflare/fuel/utils/pipe/pipe_handler.py
```diff
@@ -286,21 +285,32 @@ def _try_read(self):
...
+                    ):
+                        self._add_message(
+                            self._make_event_message(
+                                Topic.PEER_GONE, f"missing heartbeat after {self.heartbeat_timeout} secs"
+                            )
+                        )
+                        break
 
             # send heartbeat to the peer
             if now - last_heartbeat_sent_time > self.heartbeat_interval:
```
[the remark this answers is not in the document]
YuanTingHsieh: Got it, good point, I was just disabling the checking of HB.
Changed to stop/resume both checking and sending HB.

nvflare/fuel/utils/pipe/pipe_handler.py
```diff
@@ -325,3 +335,11 @@ def get_next(self) -> Optional[Message]:
...
             else:
                 return None
+
+    def pause(self):
+        """Stops checking for heartbeats from peer."""
+        self._pause = True
+
+    def resume(self):
+        """Resumes checking for heartbeats from peer."""
+        self._pause = False
```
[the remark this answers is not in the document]
YuanTingHsieh: good point, fixed

nvflare/app_common/executors/launcher_executor.py
YuanTingHsieh: Got it, I was just disabling the checking of HB.

Changed to stop/resume both checking and sending HB.
And change call to these "resume_pipe_handler" and "pause_pipe_handler" accordingly.

Note that I move the "launch_once" option into the SubprocessLauncher because I feel that whether the external system should launch once or multiple times it depends on the nature of the Launcher.

The LauncherExecutor now will ALWAYS invoke the Launcher's launch_task.

The Launcher implementation should decide whether it wants to start a new process for each task, or start a new process in the very beginning, and each task come it just ignore.
Or it can say every time a task come, it submits a new "task" to it's 3rd party system.

yanchengnv:
> > > > LGTM. Please test thoroughly to make sure FlareAgent and the executor work together in all cases:
> > > > 
> > > > * Launch once with agent disconnected and then reconnected
> > > > * Launch per-task with agent come and go.
> > > > * Make sure the pause/resume logic works as intended.
> > > > * Should Launcher provide a "kill" function so that you can kill the 3rd-party program if it fails to end for some period of time?
> > > 
> > > 
> > > I am not sure about "kill" function to a 3rd party system job we don't control. But send message to indicating the 3rd party job should stop, it's up to the 3rd party job to decide to ignore or handle the message.
> > 
> > 
> > The Launcher object (which is supposed to be implemented by the 3rd-party) could implement the "kill" method if it makes sense. For example, the SubprocessLauncher knows how to do it. Other types of Launchers may have different ways to do it; or it could be a "no op" if it cannot be done. The Executor simply calls "launcher.kill()" in it detects that the 3rd-party program is not ended after some period of time.
> 
> Launcher (on the same side of client executor, or the left side of the pipe), I thought, is not implemented by 3rd party, rather by us. The 3rd party only implement the right-side of the pipe, the interaction with IPCAgent, is that correct understanding ?

The Launcher is on the left-side with the executor. But this doesn't mean that it will be implemented by FLARE team. For example, the ORNL team knows how to launch (and kill) their stuff on their OS. Of course we could help them develop the launcher, but the launcher is very specific to the compute environment of the 3rd-party system.

examples/hello-world/ml-to-fl/np/README.md
```diff
@@ -108,3 +108,30 @@ Then we can run it using the NVFlare Simulator:
 nvflare simulator -n 2 -t 2 ./jobs/np_loop -w np_loop_workspace
 ```
 
+## Using CellPipe instead of FilePipe
```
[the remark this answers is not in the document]
YuanTingHsieh: good suggestion, let me enhance the text in the follow up PR.

PR #2170, opened 2023-11-27
Remove gen2 restriction from ad-hoc policy

yanchengnv:
Fixes # .

### Description

Currently ad-hoc connections are only allowed between gen2 cells. With this restriction, the two endpoints of a CellPipe won't be able to communicate directly.

This PR removes this restriction.

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh (approved):
LGTM

PR #2171, opened 2023-11-27
Add experimental decorator and test

holgerroth:
Fixes # .

### Description

Add experimental decorator and test.

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh (approved):
LGTM as a first step, later we can think how to re-organize these decorators and see if we can do some refactoring to reduce duplicate codes.

PR #2172, opened 2023-11-28
Fix Swarm learning README [skip ci]

YuanTingHsieh:
### Description

The job create "-sd" option will copy the whole thing inside the job, we don't want to copy all cifar 10 example scripts/data/jobs into swarm learning job.

### Types of changes

- [ ] Documentation updated.


SYangster (approved):
Thanks, good change

PR #2173, opened 2023-11-28
support getTask and submitResult timeout in job config

yanchengnv:
Fixes # .

### Description

The timeout values for getTask and submitUpdate (task result) should/could be related to the task sizes of the job.

However, currently these timeout values can only be configured in "local/resources.json".  Since this file is for site admin to maintain and not accessible to end users, users won't be able to configure these values for their jobs. 

This PR makes it possible for the user to set the timeout values in config_fed_client.json of the job. If the values are not configured in config_fed_client.json, values from local/resources.json will be used.

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh (approved):
LGTM.

PR #2175, opened 2023-11-29
Enable client side end event sequence

YuanTingHsieh:
Some Client Side Executors or Components have specific Clean up or actions for event "END_RUN" and other stuffs.

This PR proposes to add those in simulator to better simulate the full application scenario.

### Description

- Add the client side `end_run_events_sequence` in simulator

### Types of changes

- [ ] Documentation updated.


nvflare/private/fed/app/simulator/simulator_worker.py
```diff
@@ -100,6 +100,7 @@ def do_one_task(self, client):
                     if client_runner.run_abort_signal.triggered:
                         stop_run = True
                         self.logger.info("End the Simulator run.")
+                        client_runner.end_run_events_sequence()
```
yhwen, line 103: Actually, this is not totally correct. Call the client_runner.end_run_events_sequence() here only means the current client finishes its running turn. The simulator running thread will pass the "running turn" to the next client to run. The end_run_events_sequence() will fire the END_RUN event, which the client may come back to run again.

Instead, the START_RUN and END_RUN event needs to be handled in simulator_runner's run_client_thread().

yhwen (changes requested):

PR #2186, opened 2023-12-01
Fix meta file processing in storage and improve schedule job retrieval

yanchengnv:
Fixes # .

### Description

This PR address the following issues:

- Currently the meta file stored in the job store could cause memory exhaustion since it uses ast.literal_eval to convert the content in meta file. This PR fixed this issue by using simple JSON files.
- The job scheduler retrieves job candidates for scheduling every second. This requires the scanning of the whole job store and reading each job's meta file. This is a lot of wasted processing since job's status never goes back to SUBMITTED. Added an enhancement such that if a job's status is not SUBMITTED, it will be skipped in future retrievals. 
- Improved job_runner to avoid wasted scheduling attempts when no clients are online. Not only the scheduling attempts are wasted, but it could also delay the scheduling of pending submitted jobs since it will consider the schedule failed and delay its scheduling.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/apis/impl/job_def_manager.py
+++ b/nvflare/apis/impl/job_def_manager.py
@@ -33,17 +33,24 @@ from nvflare.fuel.utils.zip_utils import unzip_all_from_bytes, zip_directory_to_
 
 class _JobFilter(ABC):
     @abstractmethod
-    def filter_job(self, meta: dict) -> bool:
+    def filter_job(self, meta: dict, job_id: str) -> bool:
         pass
+
+    def is_job_skipped(self, job_id: str) -> bool:
+        return False
 
 
 class _StatusFilter(_JobFilter):
     def __init__(self, status_to_check):
         self.result = []
+        if not isinstance(status_to_check, list):
+            # turning to list
+            status_to_check = [status_to_check]
         self.status_to_check = status_to_check
 
-    def filter_job(self, meta: dict):
-        if meta[JobMetaKey.STATUS] == self.status_to_check:
+    def filter_job(self, meta: dict, job_id: str):
+        status = meta.get(JobMetaKey.STATUS.value)
+        if status in self.status_to_check:
             self.result.append(job_from_meta(meta))
         return True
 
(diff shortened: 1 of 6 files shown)


nvidianz (approved):

nvflare/app_common/storages/filesystem_storage.py
```diff
@@ -62,6 +62,21 @@ def _object_exists(uri: str):
...
 
+def _encode_meta(meta: dict) -> bytes:
+    return json.dumps(meta).encode("utf-8")
+
+
+def _decode_meta(data: bytes) -> dict:
+    s = data.decode("utf-8")
+    if s.startswith('"'):
+        # this is in old format
+        result = ast.literal_eval(json.loads(s))
```
YuanTingHsieh, line 73: not about the logic, 
I am just thinking, if there is any possibility some people can put weird stuff in the meta?
But I guess that means they already compromise the file system on server side, 
we have nothing we can do by that time then

nvflare/apis/job_def_manager_spec.py
```diff
@@ -162,7 +173,7 @@ def get_jobs_by_status(self, run_status: RunStatus, fl_ctx: FLContext) -> List[J
         """Gets Jobs of a specified status.
 
         Args:
-            run_status (RunStatus): status to filter for
+            run_status (RunStatus): status to filter for: a single or a list of status values
```
YuanTingHsieh, line 176: it will be more clear if we also update the type hint here to be:
`Union[RunStatus, List[RunStatus]]`

nvflare/apis/impl/job_def_manager.py
```diff
@@ -52,25 +59,53 @@ class _AllJobsFilter(_JobFilter):
...
-# TODO:: use try block around storage calls
+class _ScheduleJobFilter(_JobFilter):
+
+    """
+    This filter is optimized for selecting jobs to schedule since it is used so frequently (every 1 sec).
+    """
+
+    def __init__(self):
+        self.result = []
+        self.skipped_jobs = {}  # job id => bool for quick lookup.
```
YuanTingHsieh, line 88: One question here, if the server is going to run forever
Will this self.skipped_jobs keeps increasing?
Do we ever reset it?

YuanTingHsieh (commented):
Logic LGTM, have some questions

nvflare/app_common/storages/filesystem_storage.py
yanchengnv: No I don't think we need to worry about that. The problem is that our system may put some strings in the data that literal_eval doesn't like. Remember that we keep a lot of info about the job in the meta, and we may put even more there. A lot of info is free strings.

nvflare/apis/impl/job_def_manager.py
yanchengnv: Right. I'm working on a better version that uses timestamp instead.

nvflare/apis/job_def_manager_spec.py
yanchengnv: I'm scared of these type hints after running into Python's "TypeError: Subscripted generics cannot be used with class and instance checks".  This error happens in a totally unrelated place if you use List[Str] as type hint of an arg of an abstract class. It took me hours to nail this down.

YuanTingHsieh (commented):
Logic LGTM

nvflare/apis/job_def_manager_spec.py
```diff
@@ -145,6 +145,17 @@ def set_status(self, jid: str, status: RunStatus, fl_ctx: FLContext):
         """
         pass
 
+    def get_jobs_to_schedule(self, fl_ctx: FLContext) -> List[Job]:
```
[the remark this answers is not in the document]
yanchengnv: Changed

nvflare/app_common/storages/filesystem_storage.py
```diff
@@ -307,11 +336,18 @@ def delete_object(self, uri: str):
...
+        full_uri = self._object_path(uri)
 
         if not _object_exists(full_uri):
             raise StorageException("object {} does not exist".format(uri))
 
         shutil.rmtree(full_uri)
 
         return full_uri
+
+    def tag_object(self, uri: str, tag: str, data=None):
```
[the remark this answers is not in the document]
yanchengnv: That's right. We'll add it when needed. For now, we'll keep changes to minimal.

YuanTingHsieh (approved):

PR #2199, opened 2023-12-08
change Fed Stats output format

chesterxgchen:
1. change fed Stats output format. 
The prev. out format is created to fits the needs of visualization by concatenate "site" and "dataset". 
  The output is dict of 
     ``` result = { feature: {statistic : {site-dataset : value} } } ```
 This PR changed that. 
   there is no need to change the format to fit the visualization needs at this point.  The site and dataset should have their own keys in dictionary for other purpose such as query result. 

   ``` result = { feature: {statistic : {site: {dataset : value} } } } ```

4. change the visualization to reformat the dictionary to fit the visualization needs. 
  move the site-dataset concatenation here for the visualization 
  before visualization
  we convert 
   ``` result = { feature: {statistic : {site: {dataset : value} } } } ```
to 
    ``` result = { feature: {statistic : {site-dataset : value} } } ```

Fixes # .

### Description

### Types of changes

- [ ] Documentation updated.


nvflare/app_common/workflows/statistics_controller.py
```diff
@@ -384,34 +384,44 @@ def _combine_all_statistics(self):
...
+
                     if statistic == StC.STATS_HISTOGRAM:
                         hist: Histogram = self.global_statistics[statistic][ds][feature_name]
                         buckets = StatisticsController._apply_histogram_precision(hist.bins, self.precision)
-                        result[feature_name][statistic][global_dataset] = buckets
+                        result[feature_name][statistic][StC.GLOBAL][ds] = buckets
                     else:
                         result[feature_name][statistic].update(
-                            {global_dataset: round(self.global_statistics[statistic][ds][feature_name], precision)}
+                            {StC.GLOBAL: {ds: round(self.global_statistics[statistic][ds][feature_name], precision)}}
```
yanchengnv, line 424: Looks like you are manually putting together the result structure. Think about creating a "result collector" type of object to manage these with methods like add_feature, add_stats,  etc. This object could be useful in other cases.

We can do this later.

nvflare/app_opt/statistics/visualization/statistics_visualization.py
```diff
@@ -17,6 +17,17 @@
 from nvflare.fuel.utils.import_utils import optional_import
 
 
+def convert_data(feature_metrics) -> dict:
```
yanchengnv, line 20: If you had the "result collector" object, then this func could be a method of that object to export the data for different purposes.

PR #2212, opened 2023-12-13
POC bug fix

chesterxgchen:
Fixes #  FLARE-1661

### Description
Fix poc prepare bug 

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/examples/advanced/keycloak-site-authentication/README.md
+++ b/examples/advanced/keycloak-site-authentication/README.md
@@ -38,6 +38,8 @@ All the startup kits will be generated in this folder,
 
 Download and install the KeyCloak following the instruction guide from https://www.keycloak.org/getting-started/getting-started-zip. Key steps for the installation:
 
+<!-- markdown-link-check-disable -->
+
 * Start the KeyCLoak by running "bin/kc.sh start-dev"
 * Set up the realm called "myrealm"
 * Setup the user "myuser@example.com"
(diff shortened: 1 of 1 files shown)


SYangster (dismissed):

yhwen (dismissed):
LGTM

nvflare/tool/poc/poc_commands.py
```diff
@@ -473,26 +471,20 @@ def _prepare_poc(
                 )
 
             shutil.rmtree(workspace, ignore_errors=True)
-            project_config = prepare_poc_provision(
-                clients, number_of_clients, workspace, docker_image, use_he, project_conf_path, examples_dir
-            )
-            result = True
         else:
-            result = False
-    else:
```
SYangster, line 482: I think removing this else case most likely is causing the error, as now if the workspace path doesn't exist, then `prepare_poc_provision` isn't called

SYangster: For the else, I was referring to the one that matches `if os.path.exists(workspace):`. The Y/N user check is inside the "if" confirming whether to overrwrite the previous workspace directory in the case that a previous one exists. 

For when the `prepare_poc_provision()` is called, I agree it will only create a `resource.json.default`. However in the test code if the `resources.json` and `resources.json.default` does not exist, then the error will be:

`raise RuntimeError(f"Missing {RESOURCE_CONFIG} at: {resource_json_path}")` ->
 RuntimeError: Missing resources.json at: /tmp/tmpu28mcac9/example_project/prod_00/server/local/resources.json

which matches what we see

YuanTingHsieh (approved):
Thanks, LGTM

PR #2213, opened 2023-12-13
Fixed a few issues with streaming error logging

nvidianz:
### Description
Fixed a few issues with error handling of streaming,

1. Including original headers in all chunks in streaming. It was only included in first chunk before.
2. Error message is optional so it doesn't generate an error log.
3. Including original headers in error messages.
4. Added the sequence of streaming to the log message.

### Types of changes

- [ ] Documentation updated.


nvflare/fuel/f3/streaming/byte_streamer.py
```diff
@@ -177,23 +177,13 @@ def _transmit(self, task: TxTask, final=False):
...
-                message.add_headers(task.headers)
-
-            message.add_headers(
-                {
-                    StreamHeaderKey.CHANNEL: task.channel,
-                    StreamHeaderKey.TOPIC: task.topic,
-                    StreamHeaderKey.SIZE: task.stream.get_size(),
-                }
-            )
+        message = Message(task.headers, payload)
```
yanchengnv, line 180: No, you should keep the original code: message = Message(None, payload), or make a copy of task.headers.
Otherwise you are reusing the headers from task.headers and have the potential of using the headers from previous message.

nvflare/fuel/f3/streaming/byte_streamer.py
```diff
@@ -230,7 +220,7 @@ def _stop_task(self, task: TxTask, error: StreamError = None, notify=True):
                 task.stream_future.set_exception(error)
 
             if notify:
-                message = Message(None, None)
+                message = Message(task.headers)
```
yanchengnv, line 223: Should make a copy of these headers, instead of reusing task.headers, to be safe.

yanchengnv (changes requested):
See my comments.

nvflare/fuel/f3/streaming/byte_streamer.py
nvidianz: Good catch. Changed.

nvflare/fuel/f3/streaming/byte_streamer.py
nvidianz: Changed.

yanchengnv (changes requested):
LGTM. Minor improvement needed about the string formatting - see my comment.

nvflare/fuel/f3/cellnet/utils.py
```diff
@@ -77,7 +78,7 @@ def format_log_message(fqcn: str, message: Message, log: str) -> str:
         string = f"{k}=" + shorten_fqcn(get_msg_header_value(message, v))
         context.append(string)
     for k, v in msg_mapping.items():
-        string = f"{k}=" + get_msg_header_value(message, v)
+        string = f"{k}=" + str(get_msg_header_value(message, v))
```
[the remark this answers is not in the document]
nvidianz: Good suggestion. Changed.

yanchengnv (approved):
LGTM

PR #2214, opened 2023-12-14
az vm run-command invoke does not take multi-line scripts

IsaacYangSLA:
### Description

This PR updates the scripts in az vm run-command invoke to single-line format.

### Types of changes

- [ ] Documentation updated.


nvflare/lighter/impl/master_template.yml
```diff
@@ -1334,25 +1324,24 @@ azure_start_cln_sh: |
...
-    )
+  )
     az vm run-command invoke \
       --output json \
       --resource-group $RESOURCE_GROUP \
       --command-id RunShellScript \
       --name $VM_NAME \
       --scripts \
       "$scripts" > /tmp/docker_engine.json
-      sudo usermod -aG docker $ADMIN_USERNAME" > /tmp/docker_engine.json
```
chenming1986, line 1355: It seems I only had to remove this line to make the script work.

IsaacYangSLA: This line 1355 was a bug when we fixed docker installation script into heredoc.  Very likely it's a copy-and-paste error.  Anyway, the single-line format is still what Azure document says.  So we will follow that in this PR.

PR #2221, opened 2023-12-15
Update xgboost example requirements [skip ci]

SYangster:
Update requirements for xgboost example (https://nvbugspro.nvidia.com/bug/4418021)

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/examples/advanced/xgboost/histogram-based/requirements.txt
+++ b/examples/advanced/xgboost/histogram-based/requirements.txt
@@ -2,3 +2,5 @@ nvflare>=2.3.0
 pandas
 xgboost>=2.0.0
 scikit-learn
+torch
+tensorboard


examples/advanced/xgboost/histogram-based/requirements.txt
```diff
@@ -2,3 +2,5 @@ nvflare>=2.3.0
 pandas
 xgboost>=2.0.0
 scikit-learn
+torch
+tensorboard
```
YuanTingHsieh, line 6: What are we visualizing here?
I did not find where we write out metrics, can you point that out?
Do we need torch?

SYangster: torch is used in utils/baseline_centralized.py

PR #2225, opened 2023-12-17
[2.3 Branch] Add setup scripts for flare light version 

chesterxgchen:
# Description
We need to package a light version of NVFLARE for 3rd party integration usage, where grpc and protobuf dependencies are removed. This PR is will package the only selected files into a new wheel: nvflare-light

## steps to build nvflight (nvflare-light)

```
python nvflight/build_wheel.py 

```
If everything works correctly, the dist directory will contains a new dev version nvflare-light-nightly wheel
To build release wheel, use the following command

```
NVFLIGHT_RELEASE=1  python nvflight/build_wheel.py 
```

 

### Types of changes

- [ ] Documentation updated.


nvflight/setup.py
```diff
@@ -0,0 +1,54 @@
...
+)
+
+package_data = {"": ["*.yml", "*.config"], }
+
+release = os.environ.get("NVFLIGHT_RELEASE")
+version = os.environ.get("NVFLIGHT_VERSION")
+
+if release == "1":
+    package_dir = {"nvflare": "nvflare"}
+    package_name = "nvflare-light"
```
nvidianz, line 42: Should we call the package nflight?

nvflight/setup.py
```diff
@@ -0,0 +1,54 @@
...
+    include=[
+        "*",
+    ],
+    exclude=["tests", "tests.*"],
+)
+
+package_data = {"": ["*.yml", "*.config"], }
+
+release = os.environ.get("NVFLIGHT_RELEASE")
+version = os.environ.get("NVFLIGHT_VERSION")
```
nvidianz, line 38: Should we use the same version variable? Since it's the same source so the version should be always the same.

nvidianz (approved):

PR #2250, opened 2023-12-29
Workflow communication APIs and Simplified ML Algorithms

chesterxgchen:
### Description
We like to make writing a new Federated DL/ML Algorithm workflow simpler without introducing the NVFLARE specific constructs. Similar to the Client API that makes data scientist easy to transition from DL/ML to FL by just modify a few lines of the code. We like to let data scientists write different Fed Algorithms by just a making few lines of changes needed for communication, the rest of workflow logic should be independent of NVFLARE communications. 
Instead of writing a NVFLARE controller for different algorithms, we can simply write the workflow logic, separate the controller and workflow APIs. 

### Proposal 
Introduce an Workflow Communication API that responsible only for communication such as broadcast_and_wait(). User doesn't need to deal with controller APIs. 
For example: 

```
from nvflare.app_common.workflows import wf_comm as flare

class FedAvg(Controller):
    def __init__(
            self,
            min_clients: int,
            num_rounds: int,
            output_path: str,
            start_round: int = 1,
            stop_cond: str = None,
            model_selection_rule: str = None,
    ):
        super(FedAvg, self).__init__()
        
        <skip init code>

 
    def run(self):
        start = self.start_round
        end = self.start_round + self.num_rounds

        model = self.init_model()
        for current_round in range(start, end):
            self.current_round = current_round
            sag_results = self.scatter_and_gather(model, current_round)
            aggr_result = self.aggr_fn(sag_results)
            model = update_model(model, aggr_result)
            self.select_best_model(model)

        self.save_model(self.best_model, self.output_path)
        self.logger.info("end Fed Avg Workflow\n \n")

```
Where ```scatter_and_gather``` is merely call broadcast_and_wait()

```
scatter_and_gather(self, model: FLModel, current_round):
 
       msg_payload = {
            MIN_RESPONSES: self.min_clients,
            CURRENT_ROUND: current_round,
            NUM_ROUNDS: self.num_rounds,
            START_ROUND: self.start_round,
            DATA: model,
        }

        # (2) broadcast and wait
        results = self.broadcast_and_wait(msg_payload)
        return results
``` 
We also created a new Controller ( old Controller will be renamed to MPCommunicator (multi-part communicator)) , which communicate to clients
We demonstrate that user can easily write different Workflows (FedAvg and Kaplan-Meier) using the same Controller class

### Types of changes

- [ ] Documentation updated.


nvflare/app_common/workflows/wf_controller.py
```diff
@@ -0,0 +1,296 @@
...
+
+        self.log_info(fl_ctx, "workflow controller started")
+
+    def setup_wf_queue(self):
+        wf_comm_api = self.find_wf_comm_in_wf()
+        wf_comm_api.set_queue(self.wf_queue)
+        wf_comm_api.set_result_pull_interval(self.comm_msg_pull_interval)
+        wf_comm_api.meta.update({SITE_NAMES: self.get_site_names()})
+
+        print(f"\n \n  {wf_comm_api.meta=}")
```
IsaacYangSLA, line 94: If this is for debugging, self.log_debug should be better.

nvflare/app_common/workflows/wf_controller.py
```diff
@@ -0,0 +1,296 @@
...
+            return wf_comm_attrs[0]
+        else:
+            raise RuntimeError(f"missing required attribute with type of 'WFCommAPI' in {self.wf.__class__.__name__}")
+
+    def control_flow(self, abort_signal: Signal, fl_ctx: FLContext):
+        try:
+            future = self._thread_pool_executor.submit(self.ctrl_msg_loop, fl_ctx=fl_ctx, abort_signal=abort_signal)
+            self.wf.run()
+            self.stop_msg_queue("job completed", fl_ctx)
+            future.result()
```
IsaacYangSLA, line 109: This future.result() seems not necessary since ctrl_msg_loop does not return any value and the return of results() is not stored.

nvflare/app_common/workflows/wf_controller.py
```diff
@@ -0,0 +1,296 @@
...
+        except Exception as e:
+            error_msg = secure_format_traceback()
+            self.log_error(fl_ctx, error_msg)
+            self.system_panic(error_msg, fl_ctx=fl_ctx)
+        finally:
+            wait_time = self.comm_msg_pull_interval + 0.05
+            self.stop_msg_queue("job finished", fl_ctx, wait_time)
```
IsaacYangSLA, line 116: stop_msg_queue could be called twice (a normal 'job completed' and here 'job finished').  The logic of this function supports re-entrace, but it could confuse users.

nvflare/app_common/workflows/wf_controller.py
```diff
@@ -0,0 +1,296 @@
...
+            raise ValueError("WFQueue must provided")
+
+        try:
+            while True:
+                if abort_signal.triggered:
+                    break
+                if not self.wf_queue.has_ctrl_msg():
+                    time.sleep(self.comm_msg_pull_interval)
+                else:
+                    item = self.wf_queue.get_ctrl_msg()
```
IsaacYangSLA, line 148: From https://docs.python.org/3/library/queue.html#queue.Queue.task_done, we might need to consider when to call this.  Without calling it, it might result in memory leak (even the element is just a one-key dictionary).

examples/hello-world/hello-km/jobs/kaplan-meier/app/custom/km_train.py
```diff
@@ -0,0 +1,71 @@
...
+
+
+def load_data():
+    data = {
+        "site-1": {
+            "duration": [5, 10, 15, 25, 30, 35, 40, 45, 50, 55, 60, 65],
+            "event": [1, 0, 1, 1, 1, 0, 0, 1, 1, 1, 2, 4],
+        },
+        "site-2": {"duration": [10, 25, 30, 40, 50, 60, 70], "event": [1, 1, 0, 1, 0, 3, 4]},
+    }
```
ZiyueXu77, line 32: KM by lifelines currently do not support "event histogram", the input event is binary - True if the the death was observed, False if the event was lost (right-censored), so numbers > 1 does not make sense here, and event cannot be directly added (https://lifelines.readthedocs.io/en/latest/fitters/univariate/KaplanMeierFitter.html).
Let's replace this toy data with a public dataset (from sksurv.datasets import load_veterans_lung_cancer)

ZiyueXu77: I implemented a HE version based on this one (https://github.com/NVIDIA/NVFlare/pull/2259), the data / aggregation part can transfer here.

ZiyueXu77: sounds good

SYangster:
cleanup, fix server config json parsing with original controllers

todo in follow up PRs: 
- add client-side controller
- refactor to rename original Controller to MPCommunicator (MultiParty)
- convert Holger’s FedAvg class to use new controller and add examples

nvflare/app_common/wf_comm/wf_comm_api.py
```diff
@@ -0,0 +1,233 @@
...
+        self.logger = logging.getLogger(self.__class__.__name__)
+
+        self.task_results = {}
+        self.task_result_lock = threading.Lock()
+
+        data_bus = DataBus()
+        data_bus.subscribe(topics=[CommConstants.TASK_RESULT], callback=self.result_callback)
+
+        self.event_manager = EventManager(data_bus)
+        self.ctrl = data_bus.get_data(CommConstants.COMMUNICATOR)
```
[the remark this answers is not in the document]
YuanTingHsieh: what if there are multiple workflows with communicator? what will happen?

nvflare/private/fed/server/server_json_config.py
```diff
@@ -143,10 +171,49 @@ def process_config_element(self, config_ctx: ConfigContext, node: Node):
...
+            communicator = wf_config.get(CommConstants.COMMUNICATOR)
+            if communicator is None:
+                communicator = WFCommunicator()
+
+            if isinstance(communicator, WFCommunicatorSpec):
+                controller_config = wf_config.get(CommConstants.CONTROLLER)
+                controller_config["lazy_instantiate"] = False
+                communicator.set_controller_config(controller_config)
+            data_bus = DataBus()
+            data_bus.put_data(CommConstants.COMMUNICATOR, communicator)
```
YuanTingHsieh, line 191: what if there are multiple workflows with communicator? what will happen?

YuanTingHsieh: the later communicator will just overwrite the former communicator which seems incorrect? maybe we need to use a different key here?

nvflare/app_common/wf_comm/wf_comm_api.py
```diff
@@ -0,0 +1,233 @@
...
+    TARGET_SITES,
+    TASK_NAME,
+    WFCommAPISpec,
+)
+from nvflare.fuel.data_event.data_bus import DataBus
+from nvflare.fuel.data_event.event_manager import EventManager
+from nvflare.private.defs import CommConstants
+
+
+class WFCommAPI(WFCommAPISpec):
```
YuanTingHsieh, line 40: do we even need this class?
if we ask the users to just sub-class the WFController, that should be very similar to Holger's implementation of ModelController?

nvflare/app_common/wf_comm/__init__.py
```diff
@@ -0,0 +1,22 @@
+# Copyright (c) 2023, NVIDIA CORPORATION.  All rights reserved.
```
YuanTingHsieh, line 1: 2024

nvflare/app_common/wf_comm/__init__.py
```diff
@@ -0,0 +1,22 @@
...
+# limitations under the License.
+
+from nvflare.app_common.wf_comm.wf_comm_api_spec import WFCommAPISpec
+from nvflare.fuel.data_event.data_bus import DataBus
+
+data_bus = DataBus()
+
+
+def get_wf_comm_api() -> WFCommAPISpec:
+    return data_bus.get_data("wf_comm_api")
```
YuanTingHsieh, line 22: How do we expect users to use this?
In a main.py just call these the same as Client API without writing any type of Classes (Controller / Executor)?

YuanTingHsieh (changes requested):
Let's clarify on what each class should do.

examples/hello-world/hello-fedavg/README.md
```diff
@@ -0,0 +1,172 @@
+# FedAvg: simplified
+
+This example illustrates  How to use the new Workflow Communication API to contract a workflow: no need to write a controller.
```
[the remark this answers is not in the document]
SYangster: removing hello-fedavg here, will re-add rewritten version in follow up PR

nvflare/app_common/wf_comm/wf_communicator.py
```diff
@@ -0,0 +1,27 @@
...
+from nvflare.app_common.wf_comm.base_wf_communicator import BaseWFCommunicator
+
+
+class WFCommunicator(BaseWFCommunicator, Controller):
+    def __init__(self):
+        super().__init__()
+        self.register_serializers()
+
+    def control_flow(self, abort_signal: Signal, fl_ctx: FLContext):
+        self.start_workflow(abort_signal, fl_ctx)
```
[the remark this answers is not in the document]
SYangster: needed to differentiate from server-side communicator and client-side communicator (in the future)

nvflare/app_common/wf_comm/wf_comm_api.py
SYangster: this provides a better separation from nvflare specifics

nvflare/app_common/wf_comm/__init__.py
SYangster: Currently has to be used in a class with a `wf_controller_fn_name()` entry function (default `run()`) due to the component builder. Users can either subclass `WFController` which automatically calls this for the user, or they can call it themselves.

nvflare/app_common/wf_comm/wf_comm_api.py
```diff
@@ -0,0 +1,233 @@
...
+
+        meta = {} if meta is None else meta
+        msg_payload = self._prepare_input_payload(task_name, data, meta, min_responses, targets)
+        self.register_callback(callback)
+        self.ctrl.broadcast_to_peers_and_wait(msg_payload)
+
+        if callback is None:
+            return self._get_results(task_name)
+
+        return 0
```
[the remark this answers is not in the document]
SYangster: removed

nvflare/fuel/utils/component_builder.py
```diff
@@ -67,23 +67,46 @@ def build_component(self, config_dict):
...
-            if isinstance(v, dict) and self.is_class_config(v):
-                # try to replace the arg with a component
-                try:
-                    t = self.build_component(v)
-                    class_args[k] = t
-                except Exception as e:
-                    raise ValueError(f"failed to instantiate class: {secure_format_exception(e)} ")
-
-        class_path = self.get_class_path(config_dict)
+        lazy_instantiate = config_dict.get("lazy_instantiate", False)
```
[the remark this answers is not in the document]
SYangster: decided to keep in this pr

nvflare/app_common/workflows/fed_avg.py
```diff
@@ -0,0 +1,245 @@
...
+        if self.callback:
+            return self.get_aggr_result(self.aggr_params_helper, self.aggr_metrics_helper)
+        else:
+            raise ValueError("callback function needs to be defined")
+
+    def aggr_fn(self, sag_result: Optional[Dict[str, Dict[str, FLModel]]] = None) -> FLModel:
+        self.logger.info("fed avg aggregate \n")
+
+        if not sag_result:
+            raise RuntimeError("input is None or empty")
```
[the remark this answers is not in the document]
SYangster: remove temp example, will submit rewritten version of Holger's FedAvg class instead

nvflare/private/fed/server/server_json_config.py
SYangster: Good catch, appended workflow communicator "id" key as different workflows can use different communicators.

yanchengnv:
I'm not sure about many of these changes. It seems to be more complex than I imagined.

SYangster:
> I'm not sure about many of these changes. It seems to be more complex than I imagined. Does it require the user to do any config about the communicator? It's also not clear how a workflow can have its own communicator object. I need more time to review this PR.

Yes the communicator object can be configured, but is not necessary as it will default to the WFCommunicator. Also currently clarifying with the team whether we want each workflow to share a single Communicator or use its own.

config_fed_server.conf:
```
workflows = [
     {
        id = "fed_avg"
        communicator {
           # default communicator
           path = "nvflare.app_common.wf_comm.wf_communicator.WFCommunicator"
           args {}
        }
        controller {
           # subclass of WFController
           path = "nvflare.app_opt.pt.fed_avg_pt.PTFedAvg"
           args {
              min_clients = 2
              num_rounds = 2
              output_path = "/tmp/nvflare/fedavg/mode.pth"
           }
        }
     }
     
# is the same as

workflows = [
     {
        id = "fed_avg"
        # subclass of WFController, default WFCommunicator automatically used
        path = "nvflare.app_opt.pt.fed_avg_pt.PTFedAvg"
        args {
           min_clients = 2
           num_rounds = 2
           output_path = "/tmp/nvflare/fedavg/mode.pth"
        }
     }
```

SYangster:
changed to alternative design https://github.com/NVIDIA/NVFlare/pull/2390

PR #2279, opened 2024-01-12
Fix Auto login "communication error" 

chesterxgchen:
Fixes # .
increase max_login_tries from 5 to 15 and increase the sleep interval from 1 to 1.5 sec

### Description
auto login try to login with 5 login attempts and 1 second sleep in between.  But recent startup seems take longer time. 
the login will take a longer time especially in a new env. and python cache is not created. 
The start time increased to around 15 seconds instead of 5 seconds. 

Change the number of max login retries to 15 and the sleep interval to 1.5 second

### Types of changes

- [ ] Documentation updated.


SYangster (approved):
LGTM

PR #2285, opened 2024-01-15
DataBus

chesterxgchen:
### Description
As part of the series PRs related to In-Process Client API Executor and WFCommAPI. This is the 1st PR for common utiliies
MessageBus is an singleton message store and publish (callbacks) structure that other can subscribe (callbacks) and receive message in callback when message is published. The event Manager can be built on top of messagebus. 
The message can also send to/receive from message store for given topic. 

The messagebus allows two components from strong coupling to loose coupling. For example, instead comp1 set a proprty directly on comp2. We can use messagebus, where comp1 send a message to messagebus, the comp2 receive the message from messagebus. comp2 is unaware who send the message and comp1 is unware who received the message. 

### Types of changes

- [ ] Documentation updated.


nvflare/fuel/message/message_bus.py
```diff
@@ -0,0 +1,49 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import threading
+
+
+class MessageBus:
```
yanchengnv, line 18: Not sure about this class name. We used the word "message" too many times already.
In my mind, this is just a simple store of properties.

nvflare/fuel/message/message_bus.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def subscribe(self, topic, callback):
+        if topic not in self.subscribers:
+            self.subscribers[topic] = []
+        self.subscribers[topic].append(callback)
+
+    def publish(self, topic, message):
+        if topic in self.subscribers:
+            for callback in self.subscribers[topic]:
+                callback(message)
```
yanchengnv, line 38: The callback should also take "topic" as an input arg. This allows the same cb to be used for multiple topics.

nvflare/fuel/message/message_bus.py
```diff
@@ -0,0 +1,49 @@
...
+
+    def __new__(cls):
+        with cls._lock:
+            if not cls._instance:
+                cls._instance = super(MessageBus, cls).__new__(cls)
+                cls._instance.subscribers = {}
+                cls._instance.message_store = {}
+        return cls._instance
+
+    def subscribe(self, topic, callback):
```
yanchengnv, line 30: Maybe change "topic" to "topics" to allow the same cb to be registered for multiple topics?

nvflare/fuel/message/message_bus.py
```diff
@@ -0,0 +1,49 @@
...
+        if topic not in self.subscribers:
+            self.subscribers[topic] = []
+        self.subscribers[topic].append(callback)
+
+    def publish(self, topic, message):
```
yanchengnv, line 35: Should "publish" be an internal method "_publish"? No one should call this directly from outside, right?

nvflare/fuel/message/message_bus.py
```diff
@@ -0,0 +1,49 @@
...
+        if topic in self.subscribers:
+            for callback in self.subscribers[topic]:
+                callback(message)
+
+    def send_message(self, key, message, topic: str = "default"):
```
yanchengnv, line 40: What's the difference between "topic" and "key"? On line 46, you are using the key as topic. 
I don't think we need both - keep the data model very simple: each piece of data has a unique topic, though the data itself could be of any complexity.

nvflare/fuel/message/event_manger.py
```diff
@@ -0,0 +1,21 @@
+# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
```
yanchengnv, line 1: There is a typo in the file name: event_manger.py => event_manager.py

nvflare/fuel/message/message_bus.py
yanchengnv, line 38: Should we also support cases like: data changed, data deleted?

yanchengnv (changes requested):
We need to discuss the general design of the data model and how the data is intended to be used.

nvflare/fuel/message/data_bus.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def __new__(cls):
+        with cls._lock:
+            if not cls._instance:
+                cls._instance = super(DataBus, cls).__new__(cls)
+                cls._instance.subscribers = {}
+                cls._instance.message_store = {}
+        return cls._instance
+
+    def subscribe(self, topics: List[str], callback: Callable):
```
YuanTingHsieh, line 31: It would be better if we can add some docstrings here to describe the method

nvflare/fuel/message/data_bus.py
```diff
@@ -0,0 +1,52 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import threading
+from typing import Callable, List
+
+
+class DataBus:
```
YuanTingHsieh, line 19: It would be better if we can add some docstrings here to describe the class

nvflare/fuel/message/data_bus.py
```diff
@@ -0,0 +1,52 @@
...
+        return cls._instance
+
+    def subscribe(self, topics: List[str], callback: Callable):
+        if topics:
+            for topic in topics:
+                if topic not in self.subscribers:
+                    self.subscribers[topic] = []
+                self.subscribers[topic].append(callback)
+
+    def publish(self, topics: List[str], message: any):
```
YuanTingHsieh, line 38: It would be better if we can add some docstrings here to describe the method

nvflare/fuel/message/event_manager.py
```diff
@@ -0,0 +1,22 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from nvflare.fuel.message.data_bus import DataBus
+
+
+class EventManager:
```
YuanTingHsieh, line 17: It would be better if we can add some docstrings here to describe the class

nvflare/fuel/utils/function_utils.py
```diff
@@ -0,0 +1,36 @@
...
+    # Split the text by the last dot
+    tokens = task_fn_path.rsplit(".", 1)
+    module_name = tokens[0]
+    fn_name = tokens[1] if len(tokens) > 1 else ""
+    module = importlib.import_module(module_name)
+    fn = getattr(module, fn_name)
+    return fn
+
+
+def require_arguments(func):
```
YuanTingHsieh, line 29: It would be better if we can add some docstrings here to describe the method

nvflare/fuel/message/data_bus.py
```diff
@@ -0,0 +1,52 @@
...
+                    for callback in self.subscribers[topic]:
+                        callback(message, topic)
+
+    def send_message(self, key, message, topic: str = "default"):
+        if topic not in self.message_store:
+            self.message_store[topic] = {}
+
+        self.message_store[topic][key] = message
+
+    def receive_messages(self, key, topic: str = "default"):
```
YuanTingHsieh, line 51: suggest to be consistent with "send_message"
so send_message <-> receive_message

nvflare/fuel/message/data_bus.py
```diff
@@ -0,0 +1,52 @@
...
+                self.subscribers[topic].append(callback)
+
+    def publish(self, topics: List[str], message: any):
+        if topics:
+            for topic in topics:
+                if topic in self.subscribers:
+                    for callback in self.subscribers[topic]:
+                        callback(message, topic)
+
+    def send_message(self, key, message, topic: str = "default"):
```
YuanTingHsieh, line 45: It would be better if we can add some docstrings here to describe these methods

YuanTingHsieh (commented):
Suggest add docstrings, one method naming.

nvflare/fuel/utils/function_utils.py
```diff
@@ -0,0 +1,57 @@
...
+def find_task_fn(task_fn_path: str) -> Callable:
+    """
+    Find and return a callable task function based on its module path.
+
+    Parameters:
+    - task_fn_path (str): The path to the task function in the format "module_path.function_name".
+
+    Returns:
+    - Callable: The callable task function.
+    """
```
YuanTingHsieh, line 28: Thanks for all the docstrings, it helps a lot.

Our other docstrings are following google python docstring, I think we should be consistent throughout the code base. (google python doc: https://google.github.io/styleguide/pyguide.html)

For example this should be:

```
"""Finds and returns a callable task function based on its module path.
    Args:
        task_fn_path (str): The path to the task function in the format "module_path.function_name".
    Returns:
        Callable: The callable task function.
"""
```

nvflare/fuel/utils/function_utils.py
```diff
@@ -0,0 +1,57 @@
...
+    module_name = tokens[0]
+    fn_name = tokens[1] if len(tokens) > 1 else ""
+    module = importlib.import_module(module_name)
+    fn = getattr(module, fn_name)
+    return fn
+
+
+def require_arguments(func: Callable) -> Tuple[bool, int, int]:
+    """
+    Check if a function requires arguments and provide information about its signature.
```
YuanTingHsieh, line 40: All string and code looks good.

If we want to be 100% compatible to google style: https://google.github.io/styleguide/pyguide.html#381-docstrings

The summary begins after `"""`, so `"""Checks if a function requires arguments and provide information about its signature.`

But we have other places that were not followed strictly, so we can fix this later as well.

nvflare/fuel/message/event_manager.py
```diff
@@ -0,0 +1,45 @@
...
+        """
+        Initialize the EventManager with a DataBus instance.
+
+        Args:
+            data_bus (DataBus): An instance of the DataBus class used for event communication.
+        """
+        self.data_bus = data_bus
+
+    def fire_event(self, event_name: str, event_data: Optional[Any] = None) -> None:
+        """
```
yanchengnv, line 38: I guess I know what you are trying to do now: the DataBus is a storage mechanism for data, and at the same time you want to be able to fire events. The "data" here is not the same as the data saved into the bus.

These two needs are independent with each other. DataBus should first be a data store/sharing mechanism for keyed data. Secondly, you could define a Pub/Sub interface and make DataBus also implement this interface (by the same token, we could make FLContext also implement the Pub/Sub interface).

The callback signature for pub/sub should include: topic, the event data, and the databus itself since the CB will need to access the data bus for stored data.

nvflare/fuel/utils/function_utils.py
```diff
@@ -0,0 +1,57 @@
+# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
```
yanchengnv, line 1: What is this for? It doesn't seem related to data bus. Should be in a separate PR.

yanchengnv (changes requested):
See my comments. Mainly about the separation of the data storage and pub/sub interface.

nvflare/fuel/data_event/data_bus.py
```diff
@@ -0,0 +1,106 @@
...
+        if topics:
+            for topic in topics:
+                if topic in self.subscribers:
+                    with self._lock:
+                        executor = ThreadPoolExecutor(max_workers=len(self.subscribers[topic]))
+                        for callback in self.subscribers[topic]:
+                            executor.submit(callback, topic, datum, self)
+                    executor.shutdown()
+
+    def send_data(self, key: Any, datum: Any) -> None:
```
yanchengnv, line 85: Suggest change the name of this method to set_data/put_data/store_data.  "send" implies a move from source to destination.

nvflare/fuel/data_event/data_bus.py
```diff
@@ -0,0 +1,106 @@
...
+        Store a data associated with a key and topic.
+
+        Args:
+            key (Any): The key to associate with the stored message.
+            datum (Any): The message to be stored.
+        """
+        with self._lock:
+            self.data_store[key] = datum
+
+    def receive_data(self, key: Any) -> Any:
```
yanchengnv, line 96: Suggest change the name of this method to get_data/retrieve_data. "receive" implies a move from source to destination.

yanchengnv (changes requested):
Looks very good in general. Suggest to change some method names.

PR #2301, opened 2024-01-22
Added debug headers for all message route in CoreCell

nvidianz:
### Description
Many error logs are missing debug headers in CoreCell. Added the header information for all cases.

This feature doesn't affect any functions. It just makes debugging error easier.

### Types of changes

- [ ] Documentation updated.


nvflare/fuel/f3/cellnet/core_cell.py
```diff
@@ -1506,9 +1517,10 @@ def send_reply(self, reply: Message, to_cell: str, for_req_ids: List[str], secur
         reply.add_headers(
             {
                 MessageHeaderKey.FROM_CELL: self.my_info.fqcn,
+                MessageHeaderKey.TO_CELL: to_cell,
                 MessageHeaderKey.ORIGIN: self.my_info.fqcn,
-                MessageHeaderKey.ROUTE: [(self.my_info.fqcn, time.time())],
                 MessageHeaderKey.DESTINATION: to_cell,
+                MessageHeaderKey.ROUTE: [(self.my_info.fqcn, time.time())],
```
YuanTingHsieh, line 1523: should this ROUTE be after `MessageHeaderKey.MSG_TYPE` to be consistent with other places?

nvflare/fuel/f3/cellnet/core_cell.py
```diff
@@ -1159,7 +1159,18 @@ def _send_target_messages(
...
+                msg.add_headers(
+                    {
+                        MessageHeaderKey.CHANNEL: tm.channel,
+                        MessageHeaderKey.TOPIC: tm.topic,
+                        MessageHeaderKey.FROM_CELL: self.my_info.fqcn,
+                        MessageHeaderKey.TO_CELL: t,
+                        MessageHeaderKey.ORIGIN: self.my_info.fqcn,
+                        MessageHeaderKey.DESTINATION: t,
+                    }
+                )
```
yhwen, line 1172: You are adding these headers later for the reachable_targets as well. Should you change to add these headers to all the target_msgs regardless? That will also improve the readability.

nvflare/fuel/f3/cellnet/core_cell.py
nvidianz: Yeah. Changed.

nvflare/fuel/f3/cellnet/core_cell.py
nvidianz: As a good practice, we should not change the input argument. So we always create a new instance of Message and copy the headers to keep the target_msgs intact.

YuanTingHsieh (approved):
LGTM

nvflare/fuel/f3/cellnet/core_cell.py
yhwen: I mean this is the same logic as the target_msgs to create a Message and add headers to it, not to change the input argument.

PR #2308, opened 2024-01-23
[2.4] Move workspace setup inside constructor

YuanTingHsieh:
### Description

Line 354 `run_status = self._get_return_code(return_dict, process, self.workspace)` is running in the main process while 
the workspace was being setup in the forked process.

Then this workspace setup will not be reflected back into the main process.

We move the workspace setup code inside the class constructor to solve this problem.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/private/fed/app/simulator/simulator_runner.py
+++ b/nvflare/private/fed/app/simulator/simulator_runner.py
@@ -94,6 +94,15 @@ class SimulatorRunner(FLComponent):
 
         self.clients_created = 0
 
+        running_dir = os.getcwd()
+        if self.workspace is None:
+            self.workspace = "simulator_workspace"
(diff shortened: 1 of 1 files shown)


nvflare/private/fed/app/simulator/simulator_runner.py
```diff
@@ -380,7 +380,7 @@ def _get_return_code(self, return_dict, process, workspace):
                 self.logger.info(f"return_code from process.exitcode: {return_code}")
         return return_code
 
-    def run_processs(self, return_dict):
+    def run_process(self, return_dict):
```
IsaacYangSLA, line 383: This is a good catch.

IsaacYangSLA (approved):
LGTM.

PR #2310, opened 2024-01-23
[2.4] Fix ParamsConverter signature

YuanTingHsieh:
### Description

PR https://github.com/NVIDIA/NVFlare/pull/2260 add required functionalities but it breaks the "Filter" base class signature.
This PR removes the "Filter" base class inheritance

### Types of changes

- [ ] Documentation updated.


SYangster (commented):
Should we keep as Filter and move task_name argument to be passed through fl_ctx instead, or we rather remove Filter base class?

PR #2321, opened 2024-01-26
[2.4] Removed the no need client custom path

yhwen:
Fixes # .

### Description

Removed the no need client custom path when creating the simulator client. Which caused appending long system PATH.

### Types of changes

- [ ] Documentation updated.


nvflare/private/fed/app/simulator/simulator_runner.py
```diff
@@ -331,9 +331,9 @@ def create_client(self, client_name):
             client_name, self.args
         )
         self.federated_clients.append(client)
-        app_root = os.path.join(self.simulator_root, "app_" + client_name)
-        app_custom_folder = os.path.join(app_root, "custom")
-        sys.path.append(app_custom_folder)
+        # app_root = os.path.join(self.simulator_root, "app_" + client_name)
+        # app_custom_folder = os.path.join(app_root, "custom")
+        # sys.path.append(app_custom_folder)
```
YuanTingHsieh, line 336: Where is the other place that we set this custom path for the clients?

yhwen: It's here:

https://github.com/NVIDIA/NVFlare/blob/main/nvflare/private/fed/app/simulator/simulator_worker.py#L149-L151

YuanTingHsieh (approved):
LGTM

PR #2324, opened 2024-01-29
[2.4] Exclude hidden Json file when parsing config 

chesterxgchen:
### Description
Exclude hidden Json file when parsing config  ".__nvfl_xxx.json"

### Types of changes

- [ ] Documentation updated.


nvflare/lighter/tool_consts.py
```diff
@@ -0,0 +1,17 @@
+# Copyright (c) 2023, NVIDIA CORPORATION.  All rights reserved.
```
YuanTingHsieh, line 1: this should be 2024.

PR #2350, opened 2024-02-02
[2.4] Add flush=True to print in subprocess

IsaacYangSLA:
### Description

Capture stdout and stderr from subprocess in a thread and write the captured result to both logfile and stdout.

Note that since the output from print inside child process is usually buffered and may not appear on PIPE soon enough, the example python code, train_loop.py, always flushes its print.

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh (approved):
LGTM

PR #2355, opened 2024-02-06
[2.4] Fix Pipe and PipeHandler

### LINKED ISSUE #2326: client runner keeps restarting and closing the connection until my model is saved in the mlflow instance
I use Nvflare version 2.4.0rc6+11.gfdcf5785.dirty and a client api to convert my custom code to the Nvflare version. My code uses Tensorflow 2.10 and I use `model_persistor.TFModelPersistor` as the model persistor. I have changed the model persistor save_model so that it saves the global_model on my mlflow instance. as follows
```python
    def save_model_file(self, save_path: str, result, run_ids):
        """Saves model.

        Args:
            model_learnable: ModelLearnable object
            fl_ctx: FLContext
        """
        # result = unflat_layer_weights_dict(model_learnable[ModelLearnableKey.WEIGHTS])
        for k in result:
            layer = self.model.get_layer(name=k)
            layer.set_weights(result[k])
        self.model.save_weights(save_path)

        input_arr = tf.random.uniform((1, SIZE_H, SIZE_W, 3))

 
        custom_objects = {
            "Addons>SigmoidFocalCrossEntropy": tfa.losses.SigmoidFocalCrossEntropy(
                alpha=alpha, gamma=gamma
            ),
        }
        for metric_name, metric_instance in config["eval"][
            "SM_METRICS"
        ].items():
            custom_objects[metric_name] = eval(metric_instance)
        custom_objects["f1-score"] = sm.metrics.FScore()
        server_key = next(
            (
                key
                for key in run_ids.keys()
                if key.startswith("server")
            ),
            None,
        )
        run_id = run_ids[server_key]
        print("run_id of the server: %s" % run_id)

        with mlflow.start_run(run_id=run_id):
            _ = self.model(input_arr)
            mlflow.keras.log_model(
                self.model,
                custom_objects=custom_objects,
                artifact_path="tf-models",
            )
```
Since I use unet, it takes some time to save the model in the mlflow instance. But during this time, the client runner keeps restarting and closing the connection until my model is saved in my mlflow instance. How can I prevent this restarting and closing of the connection until my model is saved? In the attachment, you can see my log 
[log.txt](https://github.com/NVIDIA/NVFlare/files/14087509/log.txt)

YuanTingHsieh:
### Description

Fix PipeHandler receive blocks "send_heartbeat"

### Issue

When sending heartbeats in PipeHandler (Line 323), we use timeout = None, which will be translated to use the default_request_timeout set to 5 seconds. This causes a problem where this send_to_peer(msg) can block the entire _try_read thread for up to 5 seconds. It is supposed to be a fast-checking loop.

When the loop is blocked, it also slows down the follow-up checking/sending heartbeat logic, making the system more prone to timeouts.

### Solution
- Separate the sending of heartbeats outside of reading thread.
- To ensure the PipeHandler code re-usability, we make the "send" assumption of FilePipe to match CellPipe: If timeout is None, then wait forever.

### Types of changes

- [x] Documentation updated.

### DIFF
--- a/nvflare/app_common/executors/client_api_launcher_executor.py
+++ b/nvflare/app_common/executors/client_api_launcher_executor.py
@@ -33,7 +33,7 @@ class ClientAPILauncherExecutor(LauncherExecutor):
         external_execution_wait: float = 5.0,
         peer_read_timeout: Optional[float] = None,
         monitor_interval: float = 0.01,
-        read_interval: float = 0.001,
+        read_interval: float = 0.5,
         heartbeat_interval: float = 5.0,
         heartbeat_timeout: float = 30.0,
         workers: int = 4,
--- a/nvflare/app_common/executors/launcher_executor.py
+++ b/nvflare/app_common/executors/launcher_executor.py
@@ -45,7 +45,7 @@ class LauncherExecutor(TaskExchanger):
         external_execution_wait: float = 5.0,
         peer_read_timeout: Optional[float] = None,
         monitor_interval: float = 1.0,
-        read_interval: float = 0.1,
+        read_interval: float = 0.5,
         heartbeat_interval: float = 5.0,
         heartbeat_timeout: float = 30.0,
         workers: int = 1,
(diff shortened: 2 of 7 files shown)


yanchengnv (approved):
LGTM

PR #2363, opened 2024-02-15
Fix header parameter handling in sklearn's data_loader, update README with new AUC

mikez3:
Fixes # .

### Description

Resolved an issue in `load_data_for_range` function where the `header` parameter was not correctly handled. Previously, in some cases, the first line of data was skipped even when there was no header. The function now sets header=None only if the reader's  `header` attribute and `require_header` are not equal. After implementing this fix, I reran the example from sklearn-svm and updated the AUC score.

### Types of changes

- [ ] Documentation updated.


nvflare/app_opt/sklearn/data_loader.py
```diff
@@ -55,14 +55,15 @@ def load_data(data_path: str, require_header: bool = False):
...
     if hasattr(reader, "skiprows"):
         data_size = end - start
-        if hasattr(reader, "header") and not require_header:
-            data = reader(data_path, header=None, skiprows=start, nrows=data_size)
-        else:
-            data = reader(data_path, skiprows=start, nrows=data_size)
+        data = reader(data_path, header=header, skiprows=start, nrows=data_size)
     else:
-        data = reader(data_path).iloc[start:end]
+        data = reader(data_path, header=header).iloc[start:end]
```
ZiyueXu77, line 67: Thanks for noticing this! I think we can just change previous line 66 to 
data = reader(data_path, header=None).iloc[start:end]
Meaning if reader does not find header, then it should read with header=None.
(the above line 51 also need to add "header=None")
Keeping all other lines untouched.

ZiyueXu77 (commented):
can have less change

nvflare/app_opt/sklearn/data_loader.py
mikez3: In this case, on line 66, wouldn't it set header=None for every case where the reader doesn't find a "skiprows" attribute? Even if there is a header?

ZiyueXu77: ah you are right, we need another "if hasattr(reader, "header") and not require_header: else:" for none "skiprow" cases. My point is mainly for not introducing an "arbitrary" header of "infer".

ZiyueXu77: and it should be 
if hasattr(reader, "header") and require_header:
            data = reader(data_path, skiprows=start, nrows=data_size)
else:
            data = reader(data_path, header=None, skiprows=start, nrows=data_size)

mikez3: I've implemented the changes. Changed the logic from "not require_header" to "require_header". I also kept the `.iloc` for cases where "skiprows" is not supported (specifically for read_json). I think it should be fine now

ZiyueXu77: Thanks! could you also update line 45-49 to combine with line 51 removing the redundancy? using our updated condition logic.

ZiyueXu77 (approved):
Looks good, thanks!

ZiyueXu77 (dismissed):

nvflare/app_opt/sklearn/data_loader.py
```diff
@@ -42,13 +42,10 @@ def get_pandas_reader(data_path: str):
...
-    if hasattr(reader, "header"):
-        if require_header:
-            data = reader(data_path)
-        else:
-            data = reader(data_path, header=None)
-    else:
+    if hasattr(reader, "header") and require_header:
         data = reader(data_path)
+    else:
+        data = reader(data_path, header=None)
```
YuanTingHsieh, line 48: One of the "reader" that we get from "get_pandas_reader" is "read_json",
That method does not take in argument "header", I think we can't pass the "header" argument to it.

nvflare/app_opt/sklearn/data_loader.py
```diff
@@ -58,11 +55,14 @@ def load_data_for_range(data_path: str, start: int, end: int, require_header: bo
 
     if hasattr(reader, "skiprows"):
         data_size = end - start
-        if hasattr(reader, "header") and not require_header:
-            data = reader(data_path, header=None, skiprows=start, nrows=data_size)
-        else:
+        if hasattr(reader, "header") and require_header:
             data = reader(data_path, skiprows=start, nrows=data_size)
+        else:
+            data = reader(data_path, header=None, skiprows=start, nrows=data_size)
```
YuanTingHsieh, line 61: One of the "reader" that we get from "get_pandas_reader" is "read_json",
That method does not take in argument "header", I think we can't pass the "header" argument to it.

YuanTingHsieh (changes requested):
Not 100% correct against all cases.

YuanTingHsieh:
@mikez3 thanks for your effort on this.

@ZiyueXu77 there is a fundamental thing we need to decide, do we really want to wrap "read_excel", "read_json" and "read_csv" into this data_load? what would that benefit us? are we really using all these methods in our examples?

ZiyueXu77:
> @mikez3 thanks for your effort on this.
> 
> @ZiyueXu77 there is a fundamental thing we need to decide, do we really want to wrap "read_excel", "read_json" and "read_csv" into this data_load? what would that benefit us? are we really using all these methods in our examples?

Yeah I think we could just do csv

nvflare/app_opt/sklearn/data_loader.py
YuanTingHsieh: @mikez3 thanks for the PR, can you remove the "read_json" in line 21 in this PR?

nvflare/app_opt/sklearn/data_loader.py
YuanTingHsieh: @mikez3 thanks for the PR, can you remove the "read_json" in line 21 in this PR?

nvflare/app_opt/sklearn/data_loader.py
mikez3: Okay, it's done

YuanTingHsieh: Thanks!

YuanTingHsieh (approved):
Thanks for enhancing it

PR #2380, opened 2024-03-04
Reliable aux msg

nvidianz:
Added support for reliable message to be used by XGBoost native implementation.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/apis/utils/reliable_message.py
+++ b/nvflare/apis/utils/reliable_message.py
@@ -17,44 +17,48 @@ import time
 import uuid
 
 from nvflare.apis.fl_context import FLContext
-from nvflare.apis.shareable import ReturnCode, Shareable, make_reply
+from nvflare.apis.shareable import ReservedHeaderKey, ReturnCode, Shareable, make_reply
 from nvflare.apis.signal import Signal
 
+# Operation Types
 OP_REQUEST = "req"
 OP_QUERY = "query"
 OP_REPLY = "reply"
 
-HEADER_OP = "aux.op"
-HEADER_TOPIC = "aux.topic"
-HEADER_TX = "aux.tx"
-HEADER_TIMEOUT = "aux.timeout"
-HEADER_STATUS = "aux.status"
+# Reliable Message headers
+HEADER_OP = "rm.op"
+HEADER_TOPIC = "rm.topic"
+HEADER_TX = "rm.tx"
+HEADER_TIMEOUT = "rm.timeout"
+HEADER_STATUS = "rm.status"
 
+# Status
 STATUS_IN_PROCESS = "in_process"
 STATUS_IN_REPLY = "in_reply"
 STATUS_NOT_RECEIVED = "not_received"
 STATUS_REPLIED = "replied"
 STATUS_ABORTED = "aborted"
 
-TOPIC_RELIABLE_AUX_REQUEST = "AUX.RELIABLE_REQUEST"
-TOPIC_RELIABLE_AUX_REPLY = "AUX.RELIABLE_REPLY"
+# Topics for Reliable Message
+TOPIC_RELIABLE_REQUEST = "RM.RELIABLE_REQUEST"
+TOPIC_RELIABLE_REPLY = "RM.RELIABLE_REPLY"
 
 
 def _extract_result(reply: Shareable, target: str):
-    if not reply:
-        return None, None
     if not isinstance(reply, dict):
         return None, None
     result = reply.get(target)
     if not result:
-        return None, None
-    if not isinstance(result, Shareable):
         return None, None
     return result, result.get_return_code()
 
 
 def _status_reply(status: str):
     return make_reply(rc=ReturnCode.OK, headers={HEADER_STATUS: status})
+
+
+def _error_reply(rc: str, error: str):
+    return make_reply(rc, headers={ReservedHeaderKey.ERROR: error})
 
 
 class _RequestReceiver:
@@ -114,7 +118,7 @@ class _RequestReceiver:
         self.replying = True
         ack = engine.send_aux_request(
             targets=[self.source],
-            topic=TOPIC_RELIABLE_AUX_REPLY,
+            topic=TOPIC_RELIABLE_REPLY,
             request=self.result,
             timeout=self.timeout,
             fl_ctx=fl_ctx,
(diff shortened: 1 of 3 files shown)


yanchengnv (approved):
LGTM

nvflare/apis/utils/reliable_message.py
```diff
@@ -0,0 +1,369 @@
...
+
+# Topics for Reliable Message
+TOPIC_RELIABLE_REQUEST = "RM.RELIABLE_REQUEST"
+TOPIC_RELIABLE_REPLY = "RM.RELIABLE_REPLY"
+
+
+def _extract_result(reply: Shareable, target: str):
+    if not isinstance(reply, dict):
+        return None, None
+    result = reply.get(target)
```
YuanTingHsieh, line 50: If we hinted that reply is a Shareable, should we use shareable getter methods?
Someday we might change it to not inherit from dict?

OR we should just type hint it to dict?

nvflare/apis/utils/reliable_message.py
```diff
@@ -0,0 +1,369 @@
...
+def _status_reply(status: str):
+    return make_reply(rc=ReturnCode.OK, headers={HEADER_STATUS: status})
+
+
+def _error_reply(rc: str, error: str):
+    return make_reply(rc, headers={ReservedHeaderKey.ERROR: error})
+
+
+class _RequestReceiver:
+    def __init__(self, topic, request_handler_f, executor):
```
YuanTingHsieh, line 65: what is the type of executor?

what is the signature of request_handler_f?

nvflare/apis/utils/reliable_message.py
```diff
@@ -0,0 +1,369 @@
...
+                return receiver.process(request, fl_ctx)
+        else:
+            return make_reply(rc=ReturnCode.BAD_REQUEST_DATA)
+
+    @classmethod
+    def _receive_reply(cls, topic: str, request: Shareable, fl_ctx: FLContext):
+        tx_id = request.get_header(HEADER_TX)
+        receiver = cls._reply_receivers.get(tx_id)
+        if not receiver:
+            return make_reply(ReturnCode.OK)
```
YuanTingHsieh, line 208: Why if not reply receiver is OK?

nvflare/apis/utils/reliable_message.py
```diff
@@ -0,0 +1,369 @@
...
+                    # status should never be STATUS_NOT_RECEIVED, unless there is a bug in the receiving logic
+                    # STATUS_NOT_RECEIVED is only possible during "query" phase.
+                    break
+
+            # we didn't get a positive ack - wait a short time and re-send the request.
+            num_tries += 1
+            if num_tries > cls._max_retries:
+                # enough tries
+                return _error_reply(ReturnCode.COMMUNICATION_ERROR, f"Max send retries ({cls._max_retries}) reached")
+            time.sleep(cls._query_interval)
```
YuanTingHsieh, line 322: should this be "resend_interval" or a fast checking period, then we use other "if" statement to determine if we need to resend?

nvflare/apis/utils/reliable_message.py
```diff
@@ -0,0 +1,369 @@
...
+                topic=TOPIC_RELIABLE_REQUEST,
+                request=query,
+                timeout=timeout,
+                fl_ctx=fl_ctx,
+            )
+            ack, rc = _extract_result(ack, target)
+            if ack and rc != ReturnCode.COMMUNICATION_ERROR:
+                op = ack.get_header(HEADER_OP)
+                if op == OP_REPLY:
+                    # the ack is result itself!
```
YuanTingHsieh, line 353: Do you think it make sense to have separate methods 
- one call "_send_request" which contains the previous while loop
- and one method call "_query_result" which contains this second while loop

nvflare/apis/utils/reliable_message.py
```diff
@@ -0,0 +1,369 @@
...
+    _executor = None
+    _query_interval = 1.0
+    _max_retries = 5
+    _max_tx_time = 300.0  # 5 minutes
+    _reply_receivers = {}  # tx id => receiver
+    _tx_lock = threading.Lock()
+    _shutdown_asked = False
+
+    @classmethod
+    def register_request_handler(cls, topic: str, handler_f):
```
YuanTingHsieh, line 172: might be good to add docstring about what "handler_f" needs to be

YuanTingHsieh (commented):
Some feedbacks

nvflare/apis/utils/reliable_message.py
nvidianz: The reply is a dict, not shareable. Changed the hint.

nvflare/apis/utils/reliable_message.py
nvidianz: Added DocString.

nvflare/apis/utils/reliable_message.py
nvidianz: Added DocString.

nvflare/apis/utils/reliable_message.py
nvidianz: If no receiver found, it means the request is already processed so it returns OK. This happens when the query is sent before the result is received and the 2nd reply is received after the request is already finished.

nvflare/apis/utils/reliable_message.py
yanchengnv: Not sure what you  mean, but this is to try to make it more efficient by reducing the number of messages. So the query reply could be the result, or could also be an ack.

YuanTingHsieh: I am suggesting we break this method into 2 smaller methods, each has their own for loop.

But that is just personal preference for better readability, logic wise this part LGTM

nvflare/apis/utils/reliable_message.py
YuanTingHsieh: For example:

'''
While True:

    If current time - previous time > resend interval:
         Do resend 

    time.sleep(0.1)
'''

nvidianz: Changed to a fast loop.

nvflare/apis/utils/reliable_message.py
nvidianz: Good idea. Changed.

YuanTingHsieh (approved):
Thanks LGTM

PR #2390, opened 2024-03-07
Controller Refactor Part 1: separate communication

SYangster:
- separate out communication from controller with composition
- WFCommServer (closely coupled with Responder pattern)
- WFCommClient (uses aux channel -> will require new executor in follow up PR to support converting existing server-controlled workflows into client-controlled workflows)

Update: `Responder: handle_dead_job()`  and `ControllerSpec: process_result_of_unknown_task()` are implemented by some Controller subclasses, but are needed by the WFCommServer. For now I passed the Controller to the WFCommServer, but maybe need a cleaner solution.

### Types of changes

- [ ] Documentation updated.


nvflare/apis/impl/controller.py
```diff
@@ -525,42 +68,12 @@ def broadcast_and_wait(
...
-
-        Args:
-            task (Task): the task to be scheduled
-            fl_ctx (FLContext): FLContext associated with this task
-            targets (Union[List[Client], List[str], None], optional): the list of eligible clients or client names or None (all clients). Defaults to None.
-        """
-        _check_inputs(task=task, fl_ctx=fl_ctx, targets=targets)
-        manager = BcastForeverTaskManager()
-        self._schedule_task(task=task, fl_ctx=fl_ctx, manager=manager, targets=targets)
+        self.communicator.broadcast_forever(task, fl_ctx, targets)
```
YuanTingHsieh, line 76: we can try to remove this in next PR if no one is using this.

nvflare/apis/event_type.py
```diff
@@ -37,6 +37,7 @@ class EventType(object):
 
     BEFORE_PULL_TASK = "_before_pull_task"
     AFTER_PULL_TASK = "_after_pull_task"
+    BEFORE_PROCESS_TASK = "_before_process_task_request"
```
YuanTingHsieh, line 40: just for symmetric, do we also want "AFTER_PROCESS_TASK_REQUEST"?

nvflare/apis/impl/wf_comm_server.py
```diff
@@ -0,0 +1,1012 @@
+# Copyright (c) 2021, NVIDIA CORPORATION.  All rights reserved.
```
YuanTingHsieh, line 1: let's change this to 2024

nvflare/apis/impl/wf_comm_server.py
```diff
@@ -0,0 +1,1012 @@
...
+
+
+def _get_client_task(target, task: Task):
+    for ct in task.client_tasks:
+        if target == ct.client.name:
+            return ct
+    return None
+
+
+class WFCommServer(Responder, WFCommSpec):
```
YuanTingHsieh, line 81: I know this is taken from previous.
Do we still need a Responder class/abstract?

nvflare/private/fed/server/server_json_config.py
```diff
@@ -30,15 +30,15 @@
...
+    def __init__(self, id, controller: Controller):
         """Workflow is a responder with ID.
 
         Args:
             id: identification
             responder (Responder): A responder
         """
         self.id = id
-        self.responder = responder
+        self.controller = controller
```
YuanTingHsieh, line 41: Please update the docstring accordingly

nvflare/private/fed/server/server_runner.py
```diff
@@ -122,12 +123,15 @@ def _execute_run(self):
...
             try:
                 with self.engine.new_context() as fl_ctx:
-                    self.log_info(fl_ctx, "starting workflow {} ({}) ...".format(wf.id, type(wf.responder)))
+                    self.log_info(fl_ctx, "starting workflow {} ({}) ...".format(wf.id, type(wf.controller)))
 
                     fl_ctx.set_prop(FLContextKey.WORKFLOW, wf.id, sticky=True)
-                    wf.responder.initialize_run(fl_ctx)
 
-                    self.log_info(fl_ctx, "Workflow {} ({}) started".format(wf.id, type(wf.responder)))
+                    wf.controller.set_communicator(WFCommServer(), fl_ctx)
```
YuanTingHsieh, line 130: instead of creating `WFCommServer()` here, I think it is better to do set_communicator part in the "Workflow" object construction? Since that object is only on the server side?

nvflare/private/fed/server/server_json_config.py
YuanTingHsieh, line 41: we can do:
`controller.set_communicator(WFCommServer())` here.

nvflare/apis/impl/controller.py
```diff
@@ -11,474 +11,43 @@
...
-        """
+        self.communicator = communicator
+        self.communicator.controller = self
+        self.communicator.task_check_period = self._task_check_period
         engine = fl_ctx.get_engine()
         if not engine:
             self.system_panic(f"Engine not found. {self.__class__.__name__} exiting.", fl_ctx)
             return
 
         self._engine = engine
```
YuanTingHsieh, line 50: I think we should separate the `set_communicator` logic VS `set_engine/initialize` logic, as I argued below.

nvflare/private/fed/server/server_runner.py
YuanTingHsieh: we can do the `set_engine/initialize` here with fl_ctx.

YuanTingHsieh (commented):
Mostly LGTM, added some comments.

yanchengnv (changes requested):
Very impressive! I like it that all existing controllers don't need changes (other than change some imports).
We need to make the WFCommSpec more complete (see my comments). I understand that most of these are only specific to the server side (for now), but the concept is good and may still be applicable to other implementations. The default impl of these methods can raise exception NOT_IMPLEMENTED.

yanchengnv (approved):
LGTM

PR #2397, opened 2024-03-12
Support Responder functions

SYangster:
- fixes for integration pipeline issues
- add events for `handle_dead_job()` and `process_result_of_unknown_task`
- change `multicast_and_wait` to use standing tasks
- combine `initialize` and `start_controller`

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/apis/controller_spec.py
+++ b/nvflare/apis/controller_spec.py
@@ -543,3 +543,20 @@ class ControllerSpec(ABC):
             fl_ctx: the FL context
         """
         pass
+
+    def wait_for_task(self, task: Task, abort_signal: Signal):
+        """Wait for task to complete or abort.
+
+        Args:
+            task: the task to wait for
+            abort_signal: the abort signal. If triggered, this method stops waiting and returns to the caller.
+        """
+        pass
+
+    def handle_dead_job(self, client_name: str, fl_ctx: FLContext):
+        """Called by the Engine to handle the case that the job on the client is dead.
+        Args:
+            client_name: name of the client on which the job is dead
+            fl_ctx: the FLContext
+        """
+        pass
--- a/nvflare/apis/controller_spec.py
+++ b/nvflare/apis/controller_spec.py
@@ -293,7 +293,6 @@ class ControllerSpec(ABC):
         """
         pass
 
-    @abstractmethod
     def process_result_of_unknown_task(
         self, client: Client, task_name: str, client_task_id: str, result: Shareable, fl_ctx: FLContext
     ):
@@ -541,22 +540,5 @@ class ControllerSpec(ABC):
         Args:
             completion_status: the TaskCompletionStatus of the task
             fl_ctx: the FL context
-        """
-        pass
-
-    def wait_for_task(self, task: Task, abort_signal: Signal):
-        """Wait for task to complete or abort.
-
-        Args:
-            task: the task to wait for
-            abort_signal: the abort signal. If triggered, this method stops waiting and returns to the caller.
-        """
-        pass
-
-    def handle_dead_job(self, client_name: str, fl_ctx: FLContext):
-        """Called by the Engine to handle the case that the job on the client is dead.
-        Args:
-            client_name: name of the client on which the job is dead
-            fl_ctx: the FLContext
         """
         pass
--- a/nvflare/apis/impl/controller.py
+++ b/nvflare/apis/impl/controller.py
@@ -35,7 +35,7 @@ class Controller(FLComponent, ControllerSpec, ABC):
         self._task_check_period = task_check_period
         self.communicator = None
 
-    def initialize_controller(self, fl_ctx: FLContext):
+    def initialize(self, fl_ctx: FLContext):
         engine = fl_ctx.get_engine()
         if not engine:
             self.system_panic(f"Engine not found. {self.__class__.__name__} exiting.", fl_ctx)
--- a/nvflare/apis/impl/controller.py
+++ b/nvflare/apis/impl/controller.py
@@ -35,13 +35,14 @@ class Controller(FLComponent, ControllerSpec, ABC):
         self._task_check_period = task_check_period
         self.communicator = None
 
-    def initialize(self, fl_ctx: FLContext):
+    def initialize_controller(self, fl_ctx: FLContext):
         engine = fl_ctx.get_engine()
         if not engine:
             self.system_panic(f"Engine not found. {self.__class__.__name__} exiting.", fl_ctx)
             return
 
         self._engine = engine
+        self.start_controller(fl_ctx)
 
     def set_communicator(self, communicator: WFCommSpec):
         if not isinstance(communicator, WFCommSpec):
@@ -144,3 +145,12 @@ class Controller(FLComponent, ControllerSpec, ABC):
 
     def cancel_all_tasks(self, completion_status=TaskCompletionStatus.CANCELLED, fl_ctx: Optional[FLContext] = None):
         self.communicator.cancel_all_tasks(completion_status, fl_ctx)
+
+    def wait_for_task(self, task: Task, abort_signal: Signal):
+        self.communicator.wait_for_task(task, abort_signal)
+
+    def handle_event(self, event_type: str, fl_ctx: FLContext):
+        self.communicator.handle_event(event_type, fl_ctx)
+
+    def handle_dead_job(self, client_name: str, fl_ctx: FLContext):
+        self.communicator.handle_dead_job(client_name, fl_ctx)
--- a/nvflare/apis/impl/controller.py
+++ b/nvflare/apis/impl/controller.py
@@ -145,6 +145,3 @@ class Controller(FLComponent, ControllerSpec, ABC):
 
     def cancel_all_tasks(self, completion_status=TaskCompletionStatus.CANCELLED, fl_ctx: Optional[FLContext] = None):
         self.communicator.cancel_all_tasks(completion_status, fl_ctx)
-
-    def handle_event(self, event_type: str, fl_ctx: FLContext):
-        self.communicator.handle_event(event_type, fl_ctx)
--- a/nvflare/apis/impl/controller.py
+++ b/nvflare/apis/impl/controller.py
@@ -146,11 +146,5 @@ class Controller(FLComponent, ControllerSpec, ABC):
     def cancel_all_tasks(self, completion_status=TaskCompletionStatus.CANCELLED, fl_ctx: Optional[FLContext] = None):
         self.communicator.cancel_all_tasks(completion_status, fl_ctx)
 
-    def wait_for_task(self, task: Task, abort_signal: Signal):
-        self.communicator.wait_for_task(task, abort_signal)
-
     def handle_event(self, event_type: str, fl_ctx: FLContext):
         self.communicator.handle_event(event_type, fl_ctx)
-
-    def handle_dead_job(self, client_name: str, fl_ctx: FLContext):
-        self.communicator.handle_dead_job(client_name, fl_ctx)
--- a/nvflare/apis/impl/wf_comm_server.py
+++ b/nvflare/apis/impl/wf_comm_server.py
@@ -148,8 +148,9 @@ class WFCommServer(FLComponent, WFCommSpec):
             event_type (str): all event types, including AppEventType and EventType
             fl_ctx (FLContext): FLContext information with current event type
         """
-        if event_type == InfoCollector.EVENT_TYPE_GET_STATS:
+        if event_type == InfoCollector.EVENT_TYPE_SET_STATS:
             self._set_stats(fl_ctx)
+            self.fire_event(InfoCollector.EVENT_TYPE_GET_STATS, fl_ctx)
         elif event_type == EventType.JOB_DEAD:
             client_name = fl_ctx.get_prop(FLContextKey.DEAD_JOB_CLIENT_NAME)
             with self._dead_clients_lock:
--- a/nvflare/apis/impl/wf_comm_server.py
+++ b/nvflare/apis/impl/wf_comm_server.py
@@ -134,7 +134,7 @@ class WFCommServer(FLComponent, WFCommSpec):
                     "collector must be an instance of GroupInfoCollector, but got {}".format(type(collector))
                 )
             collector.set_info(
-                group_name=self._name,
+                group_name=self.controller._name,
                 info={
                     "tasks": {t.name: [ct.client.name for ct in t.client_tasks] for t in self._tasks},
                 },
@@ -343,8 +343,6 @@ class WFCommServer(FLComponent, WFCommSpec):
             self.log_info(fl_ctx, f"received dead job report from client {client_name}")
             if not self._dead_client_reports.get(client_name):
                 self._dead_client_reports[client_name] = time.time()
-
-        self.controller.handle_dead_job(client_name, fl_ctx)
 
     def process_task_check(self, task_id: str, fl_ctx: FLContext):
         with self._task_lock:
--- a/nvflare/apis/impl/wf_comm_server.py
+++ b/nvflare/apis/impl/wf_comm_server.py
@@ -134,7 +134,7 @@ class WFCommServer(FLComponent, WFCommSpec):
                 raise TypeError(
                     "collector must be an instance of GroupInfoCollector, but got {}".format(type(collector))
                 )
-            collector.set_info(
+            collector.add_info(
                 group_name=self.controller._name,
                 info={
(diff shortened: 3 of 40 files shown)


nvflare/apis/controller_spec.py
```diff
@@ -543,3 +543,20 @@ def cancel_all_tasks(self, completion_status=TaskCompletionStatus.CANCELLED, fl_
...
+        """
+        pass
+
+    def handle_dead_job(self, client_name: str, fl_ctx: FLContext):
+        """Called by the Engine to handle the case that the job on the client is dead.
+        Args:
+            client_name: name of the client on which the job is dead
+            fl_ctx: the FLContext
+        """
+        pass
```
YuanTingHsieh, line 562: should these be inside the comm spec?

nvflare/apis/impl/wf_comm_server.py
```diff
@@ -134,7 +134,7 @@ def _set_stats(self, fl_ctx: FLContext):
                     "collector must be an instance of GroupInfoCollector, but got {}".format(type(collector))
                 )
             collector.set_info(
-                group_name=self._name,
+                group_name=self.controller._name,
```
YuanTingHsieh, line 137: Does it make sense to pass in controller here though?

YuanTingHsieh (commented):
Some questions

nvflare/apis/controller_spec.py
SYangster: `wait_for_task()` is only used in broadcast_operator.py as you mentioned above, maybe we no longer need this? If we can find an alternate approach for `handle_dead_job()` (such as with an event?), then I agree we can take this out of the controller_spec.

YuanTingHsieh:
@yanchengnv @chesterxgchen @holgerroth 

After this effort of re-factoring controller,
I think we might need to re-evaluate the following classes, and try to simplify them or merge them:

- BroadcastAndWait (https://github.com/NVIDIA/NVFlare/blob/main/nvflare/app_common/workflows/broadcast_operator.py)
- BroadcastAndProcess (https://github.com/NVIDIA/NVFlare/blob/main/nvflare/app_common/workflows/broadcast_and_process.py)
- HubController (https://github.com/NVIDIA/NVFlare/blob/main/nvflare/app_common/hub/hub_controller.py)

nvflare/apis/impl/wf_comm_server.py
SYangster: Sure, we can look to change/improve `process_result_of_unknown_task()` which is implemented in several controllers, but used in the WFCommServer, so then we would no longer need to pass in controller here.

nvflare/apis/controller_spec.py
yanchengnv: We should revisit broadcast_operator.py. I'm not sure why it needs to implement multicasts_and_wait - the regular broadcast_and_wait should be able to do it.

nvflare/apis/wf_comm_spec.py
```diff
@@ -251,6 +251,15 @@ def cancel_all_tasks(self, completion_status=TaskCompletionStatus.CANCELLED, fl_
         """
         raise NotImplementedError
 
+    def wait_for_task(self, task: Task, abort_signal: Signal):
```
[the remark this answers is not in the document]
yanchengnv: We shouldn't need wait_for_task here. Let's revisit broadcast_operator.

nvflare/apis/controller_spec.py
```diff
@@ -543,3 +543,20 @@ def cancel_all_tasks(self, completion_status=TaskCompletionStatus.CANCELLED, fl_
             fl_ctx: the FL context
         """
         pass
+
+    def wait_for_task(self, task: Task, abort_signal: Signal):
```
yanchengnv, line 547: Shouldn't need this.

tests/unit_test/apis/impl/controller_test.py
```diff
@@ -126,7 +126,7 @@ def _setup_system(num_clients=1):
     fl_ctx = mock_server_engine.new_context()
     communicator = WFCommServer()
     controller.set_communicator(communicator)
-    controller.initialize(fl_ctx)
+    controller.initialize_controller(fl_ctx)
```
yanchengnv, line 129: Why change the method name?

yanchengnv (changes requested):
We shouldn't need wait_for_task. Let's discuss.

nvflare/apis/wf_comm_spec.py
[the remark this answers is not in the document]
SYangster: changed to use `send()` and `get_num_standing_tasks()`

nvflare/apis/controller_spec.py
SYangster: removed

nvflare/apis/controller_spec.py
SYangster: Removed `wait_for_task()` from controller. Also removed `handle_dead_job()` by adding an event.

nvflare/apis/impl/wf_comm_server.py
SYangster: We decided to keep `process_result_of_unknown_task()` for backwards compatibility, but added an event and removed the `@abstractmethod` decorator. For now `controller` will still need to be passed to `communicator` unless we remove `process_result_of_unknown_task()` in the future.

tests/unit_test/apis/impl/controller_test.py
SYangster: `initialize()` wasn't originally a method of the controller. This I added in the refactor to replace the responder `initialize_run()`, and think `initialize_controller()` name would be better to differentiate from other `initialize()` methods such as in FLComponentWrapper for example.

nvflare/app_common/workflows/cyclic_ctl.py
```diff
@@ -264,11 +265,11 @@ def restore(self, state_data: dict, fl_ctx: FLContext):
...
-                new_client_list.append(client)
-        self._participating_clients = new_client_list
+    def handle_event(self, event_type, fl_ctx):
+        if event_type == EventType.JOB_DEAD:
+            client_name = fl_ctx.get_prop(FLContextKey.DEAD_JOB_CLIENT_NAME)
+            new_client_list = []
+            for client in self._participating_clients:
+                if client_name != client.name:
+                    new_client_list.append(client)
+            self._participating_clients = new_client_list
```
YuanTingHsieh, line 275: We might also need to add "super().handle_event(event_type, fl_ctx)" in the end so whatever logic the parent component has will be invoked.

tests/unit_test/apis/impl/controller_test.py
YuanTingHsieh: So there will be name conflicts if we use `initialize()`?

Because I also find ``controller.initialize_controller`` a bit redundant right?

nvflare/apis/impl/controller.py
```diff
@@ -35,13 +35,14 @@ def __init__(self, task_check_period=0.2):
         self._task_check_period = task_check_period
         self.communicator = None
 
-    def initialize(self, fl_ctx: FLContext):
+    def initialize_controller(self, fl_ctx: FLContext):
```
YuanTingHsieh, line 38: Do we have to change this name? new name seems a bit redundant?

YuanTingHsieh (commented):
Mostly LGTM

tests/unit_test/apis/impl/controller_test.py
SYangster: Right that does seem redundant, I can change it back to `initialize()` and resolve one minor conflict instead

nvflare/apis/impl/controller.py
SYangster: reverted to initialize(), fixed minor conflict

nvflare/app_common/workflows/cyclic_ctl.py
```diff
@@ -264,11 +265,12 @@ def restore(self, state_data: dict, fl_ctx: FLContext):
...
-    def handle_dead_job(self, client_name: str, fl_ctx: FLContext):
-        super().handle_dead_job(client_name, fl_ctx)
-
-        new_client_list = []
-        for client in self._participating_clients:
-            if client_name != client.name:
-                new_client_list.append(client)
-        self._participating_clients = new_client_list
+    def handle_event(self, event_type, fl_ctx):
+        super().handle_event(event_type, fl_ctx)
```
YuanTingHsieh, line 269: This is good, just a note, if we have this, for the same event, if parent has some logic to handle as well, we will be handling the same event both in parent and in child.

nvflare/app_common/workflows/model_controller.py
```diff
@@ -144,7 +144,7 @@ def start_controller(self, fl_ctx: FLContext) -> None:
         self.event(AppEventType.INITIAL_MODEL_LOADED)
```
YuanTingHsieh, line 144: we might change this class as a whole later, but good for this PR.

YuanTingHsieh (dismissed):
LGTM

nvflare/app_common/workflows/cyclic_ctl.py
SYangster: Right we could also call this in an `else` if we want

YuanTingHsieh: right, it depends on the situations, right now it is good as it is .

nvflare/apis/impl/controller.py
```diff
@@ -144,3 +145,6 @@ def cancel_task(
 
     def cancel_all_tasks(self, completion_status=TaskCompletionStatus.CANCELLED, fl_ctx: Optional[FLContext] = None):
         self.communicator.cancel_all_tasks(completion_status, fl_ctx)
+
+    def handle_event(self, event_type: str, fl_ctx: FLContext):
```
yanchengnv, line 149: No, this would make all subclasses of Controller have to call super().handle_event.
A better approach is to add the communicator to the handler_list. See build_component() in server_json_config.py in nvflare/private/fed/server how to do that. Basically once you create the "communicator" object, simply call self.handlers.append(communicator). Of course the communicator must be a FLComponent.

nvflare/app_common/workflows/broadcast_operator.py
```diff
@@ -55,15 +56,19 @@ def multicasts_and_wait(
         task_name: str,
         task_inputs: Dict[str, Shareable],
         abort_signal: Signal = None,
+        task_check_period: int = 0.5,
     ) -> Dict[str, DXO]:
 
         tasks: Dict[str, Task] = self.get_tasks(task_name, task_inputs)
         for client_name in tasks:
-            self.controller.broadcast(task=tasks[client_name], fl_ctx=self.fl_ctx, targets=[client_name])
+            self.controller.send(task=tasks[client_name], fl_ctx=self.fl_ctx, targets=[client_name])
```
yanchengnv, line 64: It's a bad idea to keep fl_ctx in self. Instead, the method multicasts_and_wait should have a fl_ctx passed to it. The caller needs to provide fl_ctx.

nvflare/app_common/workflows/cyclic_ctl.py
yanchengnv: Once we remove the handle_event from controller.py, then we won't have this issue.

nvflare/app_common/workflows/cyclic_ctl.py
yanchengnv: We should try to avoid this as much as possible because workflow writers won't remember to do so.

yanchengnv (changes requested):
Please see my comments.

nvflare/app_common/workflows/cyclic_ctl.py
SYangster: removed the handle_event from controller.py, added additional `InfoCollector.EVENT_TYPE_SET_STATS` since `collector.set_info()` must occur before `collector.add_info()`

nvflare/apis/impl/wf_comm_server.py
```diff
@@ -147,8 +148,15 @@ def handle_event(self, event_type: str, fl_ctx: FLContext):
             event_type (str): all event types, including AppEventType and EventType
             fl_ctx (FLContext): FLContext information with current event type
         """
-        if event_type == InfoCollector.EVENT_TYPE_GET_STATS:
+        if event_type == InfoCollector.EVENT_TYPE_SET_STATS:
```
yanchengnv, line 151: I'm not sure why you need two event types here.  Please explain.

nvflare/app_common/psi/dh_psi/dh_psi_workflow.py
```diff
@@ -163,7 +163,7 @@ def pairwise_setup(self, ordered_sites: List[SiteSize]):
 
         bop = BroadcastAndWait(self.fl_ctx, self.controller)
         results = bop.multicasts_and_wait(
-            task_name=self.task_name, task_inputs=task_inputs, abort_signal=self.abort_signal
+            task_name=self.task_name, task_inputs=task_inputs, fl_ctx=self.fl_ctx, abort_signal=self.abort_signal
```
yanchengnv, line 166: This isn't perfect. No object should keep self.fl_ctx. Caller should provide fl_ctx. You can keep "engine", and then create new fl_ctx with self.engine.new_context().

We can change this in another PR.

nvflare/private/fed/client/client_runner.py
```diff
@@ -672,7 +672,7 @@ def abort(self, msg: str = ""):
         self.run_abort_signal.trigger(True)
 
     def handle_event(self, event_type: str, fl_ctx: FLContext):
-        if event_type == InfoCollector.EVENT_TYPE_GET_STATS:
+        if event_type == InfoCollector.EVENT_TYPE_SET_STATS:
```
yanchengnv, line 675: Why change GET to SET?

nvflare/widgets/info_collector.py
```diff
@@ -137,7 +138,7 @@ def get_run_stats(self) -> dict:
             coll = GroupInfoCollector()
             fl_ctx.set_prop(key=self.CTX_KEY_STATS_COLLECTOR, value=coll, sticky=False, private=True)
 
-            engine.fire_event(event_type=self.EVENT_TYPE_GET_STATS, fl_ctx=fl_ctx)
+            engine.fire_event(event_type=self.EVENT_TYPE_SET_STATS, fl_ctx=fl_ctx)
```
yanchengnv, line 141: Why change GET to SET?

yanchengnv (changes requested):
Don't see why you need to change GET_STATS to SET_STATS.
You changed on place to remove self.fl_ctx. In general, self.fl_ctx should never be used.

nvflare/app_common/psi/dh_psi/dh_psi_workflow.py
SYangster: Agreed, can change in another PR since this involves more refactoring

nvflare/apis/impl/wf_comm_server.py
SYangster: Originally many classes (ScatterAndGather, CrossSiteModelEval, ModelController etc.) handle the `InfoCollector.EVENT_TYPE_GET_STATS` which calls `GroupInfoCollector:add_info()`. However beforehand they all call the controller's `super().handle_event(event_type, fl_ctx)` which does `GroupInfoCollector:set_info()`.

Now that we are removing `handle_event()` from the controller (and keeping it in the communicator), the communicator's `set_info` still must be called first before `add_info` so it doesn't override the info, so handling the same event will no longer work. I added a `InfoCollector.EVENT_TYPE_SET_STATS` event first, and then the communicator fires `InfoCollector.EVENT_TYPE_GET_STATS` afterwards.

yanchengnv: The purpose of the GET_STATS event is simply to collect info from components that are willing to contribute (by handling this event). You don't need to keep the behavior verbatim. The simplest way is to change the call "set_info" to "add_info" in the communicator. This way, even if the communicator and controllers use the same time, info will be additive. But they don't need to use the same time. You could use whatever name for the communicator when adding/setting stats info.

yanchengnv (changes requested):
Please see my comment about GET_STATS. No need for SET_STATS.

yanchengnv (approved):
LGTM

nvflare/apis/impl/wf_comm_server.py
SYangster: Reverted back to use only GET_STATS event, and changed `set_info()` to `add_info()` in the communicator so as to not overwrite. Keeping the same name for now, as this is used in other places.

PR #2400, opened 2024-03-14
Exclude non-training clients from cse

yanchengnv:
Fixes #4562056

### Description

The CCWF's cross-site-eval workflow (CSE) assumes all evaluatees have local models. However, depending on the config of the Swarm Learning workflow (usually before CSE) of the job, some clients may only be aggregators and don't train. In this case, the client that is not a trainer won't be able to submit local model, which causes the CSE to fail.

This PR fixes this issue by checking the client against the training clients list of Swarm Learning workflow. Only training clients will be treated as evaluatees.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/app_common/ccwf/cse_server_ctl.py
+++ b/nvflare/app_common/ccwf/cse_server_ctl.py
@@ -204,7 +204,13 @@ class CrossSiteEvalServerController(ServerSideController):
             self.current_round += 1
 
         # ask everyone to eval everyone else's local model
+        train_clients = fl_ctx.get_prop(Constant.PROP_KEY_TRAIN_CLIENTS)
         for c in self.evaluatees:
+            if train_clients and c not in train_clients:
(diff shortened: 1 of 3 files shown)


nvflare/app_common/ccwf/cse_server_ctl.py
```diff
@@ -204,7 +204,13 @@ def sub_flow(self, abort_signal: Signal, fl_ctx: FLContext):
             self.current_round += 1
 
         # ask everyone to eval everyone else's local model
+        train_clients = fl_ctx.get_prop(Constant.PROP_KEY_TRAIN_CLIENTS)
```
SYangster, line 207: Good fix, LGTM

SYangster (approved):

SYangster (approved):
Nice fix, LGTM

PR #2401, opened 2024-03-15
[2.4] Add support to large object in LauncherExecutor using CellPipe

YuanTingHsieh:
### Issue

We need https://github.com/NVIDIA/NVFlare/pull/2406 to enable streaming capabilities to CellPipe as well

1. When the object/model is large, the pull_task from client side will takes a lot of time, so we don't want to start the PipeHandler heartbeat at "START_RUN" event

2. Metric relay component does not need separate timeout itself, so adding an option in PipeHandler to disable the heartbeat

3. Since we always will have task pipe for flare_agent, we just need it to do heartbeats so even if metric pipe is provided, it does not need to send heartbeats 

4. When peer_read_timeout is None, it is not waiting forever, instead the underlying PipeHandler will just use a default request timeout of 5 seconds which is not enough for the model weights/weight diff, so change to a larger default value 

5. PipeHandler send Abort or End signal does not need to wait forever, so provide a timeout there

6. Same as https://github.com/NVIDIA/NVFlare/pull/2370 we need to invoke TaskExchanger's handle_event in LauncherExecutor 

### Description

Add support to large object to LauncherExecutor/ClientAPI utilizing CellPipe

- Update timeout values for larger objects
- Should start the pipe_handler's heartbeat / checking for heartbeat mechanism after the task is pulled and ready to be executed
- MetricRelay component does not need its own heartbeat mechanism

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/app_common/executors/client_api_launcher_executor.py
+++ b/nvflare/app_common/executors/client_api_launcher_executor.py
@@ -31,11 +31,11 @@ class ClientAPILauncherExecutor(LauncherExecutor):
         task_wait_timeout: Optional[float] = None,
         last_result_transfer_timeout: float = 300.0,
         external_execution_wait: float = 5.0,
-        peer_read_timeout: Optional[float] = None,
+        peer_read_timeout: Optional[float] = 60.0,
         monitor_interval: float = 0.01,
         read_interval: float = 0.5,
         heartbeat_interval: float = 5.0,
-        heartbeat_timeout: float = 30.0,
+        heartbeat_timeout: float = 60.0,
         workers: int = 4,
         train_with_evaluation: bool = True,
         train_task_name: str = "train",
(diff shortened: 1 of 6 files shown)


nvflare/app_common/executors/client_api_launcher_executor.py
```diff
@@ -31,11 +31,11 @@ def __init__(
         task_wait_timeout: Optional[float] = None,
         last_result_transfer_timeout: float = 300.0,
         external_execution_wait: float = 5.0,
-        peer_read_timeout: Optional[float] = None,
+        peer_read_timeout: Optional[float] = 60.0,
         monitor_interval: float = 0.01,
         read_interval: float = 0.5,
         heartbeat_interval: float = 5.0,
-        heartbeat_timeout: float = 30.0,
+        heartbeat_timeout: float = 60.0,
```
SYangster, line 38: update default value in docstring

nvflare/app_common/executors/launcher_executor.py
```diff
@@ -43,12 +43,12 @@ def __init__(
...
         external_execution_wait: float = 5.0,
-        peer_read_timeout: Optional[float] = None,
+        peer_read_timeout: Optional[float] = 60.0,
         monitor_interval: float = 1.0,
         read_interval: float = 0.5,
         heartbeat_interval: float = 5.0,
-        heartbeat_timeout: float = 30.0,
-        workers: int = 1,
+        heartbeat_timeout: float = 60.0,
+        workers: int = 4,
```
SYangster, line 51: update default values in docstring

yanchengnv (changes requested):
Why so many changes? The only change needed is cell.py. 

If you want to improve other things (seem to be non-trivial), please do so in another PR and describe the reason.

YuanTingHsieh:
Sounds good, the cell changing is moved to https://github.com/NVIDIA/NVFlare/pull/2406 


This PR handles all other issues I encountered in testing large models + LauncherExecutor/Client API

PR #2435, opened 2024-03-21
Add IPC agent and exchanger

yanchengnv:
### Description

This PR applies following changes made for AV in 2.3 to 2.5. 

- IPC exchanger and IPC agent. Those similar changes were made in 2.4 with cell pipe and task exchanger, these IPC exchanger/agent have more robust logic for handling AV job corner conditions.
- Support app-defined training message payload without requiring payload to be dict or numpy.

### Types of changes

- [ ] Documentation updated.


nvflare/app_common/app_defined/shareable_generator.py
```diff
@@ -0,0 +1,94 @@
...
+
+        Args:
+            model_obj: base model object that weights will be applied to
+            weights: trained weights
+            meta: trained meta
+
+        Returns: the updated model object
+
+        """
+        pass
```
YuanTingHsieh, line 60: We can discuss this with AV / users about what interface they want?

I personally think apply_weights_to_model sounds like an aggregator job?
Generator should be "model_to_trainable" and "trainable_to_model"?
What is even better if user don't need to write this class?

nvflare/app_common/app_defined/__init__.py
```diff
@@ -0,0 +1,13 @@
+# Copyright (c) 2023, NVIDIA CORPORATION.  All rights reserved.
```
YuanTingHsieh, line 1: 2024

YuanTingHsieh (commented):
IPC agent and exchanger LGTM.

We could ask AV about app defined components interfaces for feedbacks

nvflare/app_common/app_defined/shareable_generator.py
yanchengnv: I debated with myself on this. I think "apply_weights_to_model" is a better name here: the model could have a lot more info than just the weights. What this method does is to apply trained result to the model (and leave other things of the model intact).

Maybe we should change the name to "apply_training_result_to_model", instead of calling the training_result "weights".

nvflare/app_common/app_defined/aggregator.py
```diff
@@ -0,0 +1,75 @@
...
+from nvflare.apis.shareable import Shareable
+from nvflare.app_common.abstract.aggregator import Aggregator
+from nvflare.app_common.abstract.model import ModelLearnableKey
+from nvflare.app_common.app_constant import AppConstants
+from nvflare.app_common.app_event_type import AppEventType
+
+from .component_base import ComponentBase
+
+
+class AppDefinedAggregator(Aggregator, ComponentBase, ABC):
```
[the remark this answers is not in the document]
yanchengnv: The whole point here is to support users (like AV) that do NOT want to use what we defined (FLModel, Numpy, DXO, etc). App Defined means that it is whatever the user wants to define as the model.

nvflare/app_common/app_defined/shareable_generator.py
```diff
@@ -0,0 +1,94 @@
+# Copyright (c) 2023, NVIDIA CORPORATION.  All rights reserved.
```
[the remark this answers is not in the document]
yanchengnv: Fixed

nvflare/app_common/app_defined/__init__.py
yanchengnv: changed

YuanTingHsieh (approved):
Thanks, LGTM

PR #2441, opened 2024-03-23
[2.4] Improve cell pipe timeout handling

yanchengnv:
Fixes # 4572538.

### Description

Currently pipe timeout replies on heartbeat messages. Even though heartbeats are sent in an independent thread, when the pipe is busy with large task data (e.g. 8G), the heartbeat messages will be queued after some chunks of task data. It will take time for the heartbeat messages to arrive, which could be too late (> configured timeout value).

This PR tries to improve it by intercepting all messages from the peer and remember the last time that any message is received from the peer. This time is used in the timeout logic: if it is > the last heartbeat time, it is treated as the last HB time.

### Types of changes

- [ ] Documentation updated.


nvflare/fuel/utils/pipe/cell_pipe.py
```diff
@@ -249,6 +292,9 @@ def send(self, msg: Message, timeout=None) -> bool:
                         self.logger.error(err)
                     return False
             else:
+                self.logger.error(
```
SYangster, line 295: add f-string for format

nvflare/fuel/utils/pipe/cell_pipe.py
```diff
@@ -46,8 +49,10 @@ def _cell_fqcn(mode, site_name, token):
     return f"{site_name}_{token}_{mode}"
 
 
-def _to_cell_message(msg: Message) -> CellMessage:
-    headers = {_HEADER_MSG_TYPE: msg.msg_type, _HEADER_MSG_ID: msg.msg_id}
+def _to_cell_message(msg: Message, extra=None) -> CellMessage:
+    headers = {_HEADER_MSG_TYPE: msg.msg_type, _HEADER_MSG_ID: msg.msg_id, _HEADER_START_TIME: time.time()}
```
SYangster, line 53: `_HEADER_START_TIME` doesn't seem to be used anywhere?

YuanTingHsieh: If `_HEADER_START_TIME` is for debugging purpose, should we pass it in as `extra_headers = {_HEADER_HB_SEQ: self.hb_seq}`?

YuanTingHsieh (approved):
Thanks, LGTM, 

This PR change control signals (ABORT/END/HB) handling, instead of waiting for reply (which might takes a long time):
- This PR use fire_and_forget for HB
- This PR just check if sending is done for ABORT/END

This PR also enhances the timeout/keep_alive mechanism, if one side has received anything (reply/request) from the other side, the active time will be updated.

nvflare/fuel/utils/pipe/cell_pipe.py
yanchengnv: It is

PR #2451, opened 2024-03-27
Add controller interface

SYangster:
- Add WFControllerSpec API for simpler definition for users
- Add non-blocking option to send_model()

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/apis/wf_controller_spec.py
+++ b/nvflare/apis/wf_controller_spec.py
@@ -1,0 +1,50 @@
+# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from abc import ABC, abstractmethod
+from typing import Callable, List, Union
+
+
+class WFControllerSpec(ABC):
+    @abstractmethod
+    def run(self):
+        """Main `run` routine for the controller workflow."""
+        raise NotImplementedError
+
+    def send_model(
+        self,
+        task_name: str,
+        data: any,
+        targets: Union[List[str], None],
+        timeout: int,
+        wait_time_after_min_received: int,
+        blocking: bool,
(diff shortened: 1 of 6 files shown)


nvflare/apis/wf_controller_spec.py
```diff
@@ -0,0 +1,50 @@
...
+class WFControllerSpec(ABC):
+    @abstractmethod
+    def run(self):
+        """Main `run` routine for the controller workflow."""
+        raise NotImplementedError
+
+    def send_model(
+        self,
+        task_name: str,
+        data: any,
```
holgerroth, line 28: Should this be `any` or `FLModel`?

YuanTingHsieh: We can have this base method call "send" so it is not tied to `FLModel`
and use `any`.

And then in the ModelController, we implement both `send` and `send_model` method?

So other classes remain the same change as @SYangster  already have?

nvflare/app_common/workflows/base_fedavg.py
```diff
@@ -27,7 +28,7 @@
 from .model_controller import ModelController
 
 
-class BaseFedAvg(ModelController):
+class BaseFedAvg(ModelController, WFControllerSpec):
```
YuanTingHsieh, line 31: Since `ModelController` already inherit `WFControllerSpec` we don't need to duplicate it here?

nvflare/apis/wf_controller_spec.py
SYangster: Right `any` is used since ModelController implements the spec with `FLModel`, but another implementation could use something different. I agree with changing the base method name of `send_model` (however `send` is already a method of Controller which could be confusing)

nvflare/app_common/workflows/base_fedavg.py
SYangster: Agree this is redundant, however one of the goals was to expose the interface to users so they are not required to only look into the internal `ModelController` implementation. @chesterxgchen

nvflare/apis/wf_controller_spec.py
YuanTingHsieh: got it, if there's conflict this current name is good.

nvflare/app_common/workflows/model_controller.py
```diff
@@ -36,7 +37,7 @@
 
 
 @experimental
-class ModelController(Controller, FLComponentWrapper):
+class ModelController(Controller, FLComponentWrapper, WFControllerSpec, ABC):
```
YuanTingHsieh, line 40: maybe we move this "WFControllerSpec, ABC" outside of here?

SYangster: decided to keep since `ModelController` uses `run()`

PR #2469, opened 2024-04-03
[2.4] Add warning when the same admin in project.yml has different role

IsaacYangSLA:
### Description

By design, provisioners can add clients or admins in project.yml and do provision to get additional client/admin startup kits.  The certs and private keys of existing participants are kept the same so that there is no need to re-distribute those existing startup kits.

When the role of existing admin changes in project.yml, its cert/private key and thus the role in subject's unstructured_name, do not change.  This causes confusion as provisioners may think a new set of cert/private key is generated.

This PR adds a warning during provisioning time so that provisioners know what to do to generate the new startup kits.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/lighter/impl/cert.py
+++ b/nvflare/lighter/impl/cert.py
@@ -87,6 +87,17 @@ class CertBuilder(Builder):
             pri_key = serialization.load_pem_private_key(
                 self.persistent_state[subject]["pri_key"].encode("ascii"), password=None, backend=default_backend()
             )
+            if participant.type == "admin":
+                cn_list = cert.subject.get_attributes_for_oid(NameOID.UNSTRUCTURED_NAME)
+                for cn in cn_list:
+                    role = cn.value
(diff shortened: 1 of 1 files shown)


nvflare/lighter/impl/cert.py
```diff
@@ -87,6 +87,17 @@ def _build_write_cert_pair(self, participant, base_name, ctx):
             pri_key = serialization.load_pem_private_key(
                 self.persistent_state[subject]["pri_key"].encode("ascii"), password=None, backend=default_backend()
             )
+            if participant.type == "admin":
```
yhwen, line 90: What's the reason we only do this for the "admin" type?

IsaacYangSLA: Only admin certs have unstructured name, which contains the 'role' information, such as project_admin.

yhwen (approved):
LGTM

PR #2481, opened 2024-04-10
[2.4] Added more logging for the job status changing

yhwen:
Fixes # .

### Description

Added more logging for the job status changing.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/private/fed/server/job_runner.py
+++ b/nvflare/private/fed/server/job_runner.py
@@ -443,7 +443,7 @@ class JobRunner(FLComponent):
                                     },
                                     fl_ctx,
                                 )
-                                self.log_info(f"Updated the schedule history of Job: {job_id}")
+                                self.log_info(fl_ctx, f"Updated the schedule history of Job: {job_id}")
 
                             if failed_clients:
                                 deployable_clients = {k: v for k, v in client_sites.items() if k not in failed_clients}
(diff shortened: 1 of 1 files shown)


nvflare/private/fed/server/job_runner.py
```diff
@@ -464,6 +465,7 @@ def run(self, fl_ctx: FLContext):
                             with self.lock:
                                 self.running_jobs[job_id] = ready_job
                             job_manager.set_status(ready_job.job_id, RunStatus.RUNNING, fl_ctx)
+                            self.log_info(fl_ctx, f"Job: {job_id} started to run, status changed to RUNNING.")
```
yanchengnv, line 468: Can you add more messages:
- right before self._start_run()
- after self._start_run() and before "with self.lock"
- before job_manager.set_status(ready_job.job_id, RunStatus.RUNNING, fl_ctx)

This can help better understand where it's stuck.

yanchengnv (changes requested):
Please add more log messages. See my comments.

nvflare/private/fed/server/job_runner.py
yhwen: Added more logging.

yanchengnv (approved):
Looks good.

PR #2486, opened 2024-04-10
Fix xgboost integration tests

SYangster:
use path instead of name

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh:
@SYangster you might also need to do this on the config of the job template that is used by finance and vertical xgboost

PR #2490, opened 2024-04-10
[2.3] Fix CyclicController bug

YuanTingHsieh:
Follows PR https://github.com/NVIDIA/NVFlare/pull/2484,
we need to change the handle_dead_job method, 
the current implementation is flawed as it does not wait for the grace period.

Controllers should not handle client dropping, we will enhance this properly after 2.5

### Description

- Fix client drop handling
- Allow users to pass custom "order" for relay
- Print out more logs 

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/app_common/workflows/cyclic_ctl.py
+++ b/nvflare/app_common/workflows/cyclic_ctl.py
@@ -14,6 +14,7 @@
 
 import gc
 import random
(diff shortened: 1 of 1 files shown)


yanchengnv (approved):
LGTM

PR #2493, opened 2024-04-12
Fix issues on hello-world TF2 notebook

IsaacYangSLA:
### Description

This PR fixes two issues on hello-tensorflow section of hello-world jupyter notebook.

The first one is the env variable set inside notebook take no effects in the POC clients because all POC sites and server are up and running before that cell runs.  To set that env var properly, this PR added two lines of codes at the very beginning of trainer.py of hello tf2 example.  This will ensure that env var been set in all clients before starting any TF2 related codes. 

The second issue is the path of server workspace in POC changed and needs to be updated.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/examples/hello-world/hello-tf2/jobs/hello-tf2/app/custom/trainer.py
+++ b/examples/hello-world/hello-tf2/jobs/hello-tf2/app/custom/trainer.py
@@ -12,6 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import os
+
+os.enviro["TF_FORCE_GPU_ALLOW_GROWTH"] = "true"
+
 import numpy as np
 import tensorflow as tf
 from tf2_net import Net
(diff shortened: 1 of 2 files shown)


examples/hello-world/hello-tf2/jobs/hello-tf2/app/custom/trainer.py
```diff
@@ -12,6 +12,10 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import os
+
+os.enviro["TF_FORCE_GPU_ALLOW_GROWTH"] = "true"
```
YuanTingHsieh, line 17: typo, @IsaacYangSLA

IsaacYangSLA: Will fix it now.

PR #2495, opened 2024-04-12
[2.4] Fix pipe handler timeout in task exchanger and launcher executor

YuanTingHsieh:
Fix 4593298

### Description

The pipe handler HB should start right away in START_RUN, then the monitor_launcher can check for external process's status to pause/resume.

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh:
I miss a thing here, if launcher is launch multiple times, and the task is very large then the pull task going to take a lot of time, so START_RUN event starts right away but the pull task will leads to timeout, we should pause right after and resume whenever a task is going to be executed, let me change and add test cases for all these conditions

PR #2502, opened 2024-04-15
[2.4] Rely on launcher check_run_status to pause/resume heartbeat

YuanTingHsieh:
### Issue

To address both scenarios where launch_once is set to False or True, we remove the skip condition ``if self._current_task is None: continue``. This skip condition is inaccurate.
The LauncherExecutor should not use this skip; instead, it should verify whether the external process is running to determine whether to pause or resume the PipeHandler heartbeat.

For instance, consider a scenario where the TASK is very large (4GB or more).
When launch_once is set to False, the PipeHandler heartbeat begins at the "START_RUN" event, while the NVFlare client process attempts to pull the task from the server. This process can take a significant amount of time (up to 300 seconds or more). The external process is only started after the task is pulled (in the execute method). Since the PipeHandler heartbeat checking/sending begins at the start, it may timeout (default 60 seconds) before the external process has a chance to launch (because 60 < 300).

To prevent users from needing to adjust the heartbeat timeout in proportion to the time taken to pull the task, we propose the following change in this PR.

### Description
The revised logic is as follows:

1. When launch_once is True, the `TaskExchanger` calls the `pipe_handler.start` at the "START_RUN" event, and the SubprocessLauncher launches the external process at the same event. Since the external process continues running throughout, it will not be paused or resumed.

2. When launch_once is False, the `TaskExchanger` still calls the `pipe_handler.start` at the "START_RUN" event. In `_monitor_launcher`, it detects that the SubprocessLauncher has not launched the external process, so it calls `pause_pipe_handler`. After the task is pulled, the SubprocessLauncher launches the external process. Subsequently, in `_monitor_launcher`, it detects that the SubprocessLauncher has launched the external process, so it calls `resume_pipe_handler`.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/app_common/executors/launcher_executor.py
+++ b/nvflare/app_common/executors/launcher_executor.py
@@ -373,10 +373,6 @@ class LauncherExecutor(TaskExchanger):
                 if self.launcher is None:
                     break
 
-                # no task is running
-                if self._current_task is None:
-                    continue
-
                 task_name = self._current_task
                 run_status = self._execute_launcher_method_in_thread_executor(
                     method_name="check_run_status",
--- a/nvflare/app_common/executors/task_exchanger.py
+++ b/nvflare/app_common/executors/task_exchanger.py
@@ -142,7 +142,7 @@ class TaskExchanger(Executor):
         task_id = shareable.get_header(key=FLContextKey.TASK_ID)
 
         # send to peer
-        self.log_debug(fl_ctx, f"sending task to peer {self.peer_read_timeout=}")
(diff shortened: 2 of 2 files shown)


YuanTingHsieh:
For launch_once=False, even if we start the PipeHandler heartbeat thread after the first pulled task (of nvflare client), when the first task is done the external process will be stopped.

For the consecutive tasks, we still need to pause/resume PipeHandler heartbeat, it will not make things simpler.


The pull_task behavior of nvflare client is not controlled by the executor class.

We do have timeout when the TaskExchanger sends the task to external process, it is called "peer_read_timeout"

nvidianz (approved):

PR #2505, opened 2024-04-15
Enhance WFController

SYangster:
- make round-based information optional for controller, launcher, and in-process client api sides
- fix in-process client api bug (previously we were creating a new InProcessClientAPI for each task which would require calling flare.init() on the client-side each time)
- rename ModelController method to broadcast_model()
- add default persistor_id with warning if persistor does not exist

### Types of changes

- [ ] Documentation updated.


holgerroth (approved):
These are good enhancements.

PR #2501, opened 2024-04-14
[2.3] Improve dead client handling

yanchengnv:
Fixes # .

### Description

This PR added following improvement to v2.3:

- Improved dead client determination logic
- Added log messages for dead client event tracking
- Added a debug admin command for analyzing dead client events
- Improved Cyclic Controller

### Types of changes

- [x] Non-breaking change (fix or new feature that would not break existing functionality).
- [ ] Breaking change (fix or new feature that would cause existing functionality to change).
- [ ] New tests added to cover the changes.
- [ ] Quick tests passed locally by running `./runtest.sh`.
- [ ] In-line docstrings updated.
- [ ] Documentation updated.

PR #2506, opened 2024-04-15
Improve dead client handling

yanchengnv:
Fixes # .

### Description

This PR applies changes in https://github.com/NVIDIA/NVFlare/pull/2501 to v2.5:
- Better and more reliable handling of dead client/job reports
- More detailed log messages to help analyze dead client events when occurred
- Improved client revive logic
- Updated CyclicController to use the new client dead function.

### Types of changes

- [ ] Documentation updated.


nvflare/apis/impl/wf_comm_server.py
```diff
@@ -135,7 +139,7 @@ def _set_stats(self, fl_ctx: FLContext):
                     "collector must be an instance of GroupInfoCollector, but got {}".format(type(collector))
                 )
             collector.add_info(
-                group_name=self.controller._name,
+                group_name=self.name,
```
SYangster, line 142: Here `self.name='WFCommServer'` while `self.controller._name='ScatterAndGather'`. This change is fine, but I do remember one of the integration tests uses the stats collector and expects the Controller name (eg 'ScatterAndGather'), so that would have to be changed accordingly.

nvflare/apis/impl/controller.py
```diff
@@ -145,3 +145,16 @@ def cancel_task(
 
     def cancel_all_tasks(self, completion_status=TaskCompletionStatus.CANCELLED, fl_ctx: Optional[FLContext] = None):
         self.communicator.cancel_all_tasks(completion_status, fl_ctx)
+
+    def get_client_disconnect_time(self, client_name):
+        """Get the time that the client is deemed dead.
```
SYangster, line 150: Suggest change docstring to use same term (either "disconnected" or "dead") for consistency.

Just to confirm, it seems that they roughly mean the same thing, except disconnected occurs after a client has been reported dead for > dead_client_grace_period correct? Maybe using `_DeadClientStatus.disconnect_time` instead of `_DeadClientStatus.death_time` makes sense?

SYangster (commented):
LGTM, minor comments

nvflare/apis/impl/wf_comm_server.py
yanchengnv: The reason I changed this is to avoid accessing private "_name" of self.controller. Besides, the group_name can be anything. If you know where the test is, please modify it - it shouldn't be checking the name.

nvflare/apis/impl/controller.py
yanchengnv: Good catch. The reasons I wanted to change to "disconnected" are:
- There is really no way to know for sure that the client is dead. All we could tell is that the client is currently disconnected. Hence disconnected is more accurate.
- secondly, I don't like the term "death" in this public method that users may need to call. It is too dramatic and may even offend some people. I don't want to change other places that use the term "dead" since they are all internal code and won't be called by other developers.

nvflare/apis/impl/wf_comm_server.py
SYangster: Sure, I will look into which tests have to be changed

SYangster (approved):
Thanks, LGTM

nvflare/private/fed/server/sys_cmd.py
```diff
@@ -76,6 +76,14 @@ def get_spec(self):
                     authz_func=self.authorize_client_operation,
                     visible=True,
                 ),
+                CommandSpec(
+                    name="dead",
+                    description="send dead client msg to SJ",
+                    usage="dead <client-name>",
```
yhwen, line 82: What's the reason to build an admin command "dead client"?

if so, should we make it consistent for the command format to be "dead client XXX" and "dead client XXX YYY" ?

yanchengnv: As I said in the description, this command is used to simulate dead client report for the SP, since it's very hard to reproduce it thru connection disrupt. It's only for debugging/testing of dead client handling logic; and is not visible to regular users.

This is only for SP, hence it doesn't make sense to send to clients XXX YYY.

PR #2507, opened 2024-04-15
[2.3] Fix 2.3 unit tests

SYangster:
AioGrpcDriver exception handling bug: in python 3.8, `asyncio.CancelledError` was changed to subclass `BaseException` instead of `Exception`, so it will not be caught with `except Exception as ex:`

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/fuel/f3/drivers/aio_grpc_driver.py
+++ b/nvflare/fuel/f3/drivers/aio_grpc_driver.py
@@ -328,6 +328,9 @@ class AioGrpcDriver(BaseDriver):
             msg_iter = stub.Stream(connection.generate_output())
             conn_ctx.conn = connection
             await connection.read_loop(msg_iter)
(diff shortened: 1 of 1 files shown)


nvflare/fuel/f3/drivers/aio_grpc_driver.py
```diff
@@ -328,6 +328,9 @@ async def _start_connect(self, connector: ConnectorInfo, aio_ctx: AioContext, co
             msg_iter = stub.Stream(connection.generate_output())
             conn_ctx.conn = connection
             await connection.read_loop(msg_iter)
+        except asyncio.CancelledError:
```
nvidianz, line 331: We have a few other `except Exception` without this. Please add this catch clause to all of them.

nvidianz (approved):

PR #2508, opened 2024-04-15
Fix update client status

SYangster:
Check workflow id when handling event before calling `_update_client_status()` for the case of multiple workflows

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/app_common/ccwf/server_ctl.py
+++ b/nvflare/app_common/ccwf/server_ctl.py
@@ -345,7 +345,9 @@ class ServerSideController(Controller):
 
     def handle_event(self, event_type: str, fl_ctx: FLContext):
         if event_type == EventType.BEFORE_PROCESS_TASK_REQUEST:
-            self._update_client_status(fl_ctx)
+            peer_ctx = fl_ctx.get_peer_context()
+            if peer_ctx and self.workflow_id in peer_ctx.get_prop(Constant.STATUS_REPORTS, {}):
+                self._update_client_status(fl_ctx)
 
     def process_config_reply(self, client_name: str, reply: Shareable, fl_ctx: FLContext) -> bool:
         return True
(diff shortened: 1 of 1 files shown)


holgerroth (dismissed):
Looks good to me.

nvflare/app_common/ccwf/server_ctl.py
```diff
@@ -345,7 +345,9 @@ def control_flow(self, abort_signal: Signal, fl_ctx: FLContext):
 
     def handle_event(self, event_type: str, fl_ctx: FLContext):
         if event_type == EventType.BEFORE_PROCESS_TASK_REQUEST:
-            self._update_client_status(fl_ctx)
+            peer_ctx = fl_ctx.get_peer_context()
```
yanchengnv, line 348: Why is this change necessary? self._update_client_status already checked these conditions.

yanchengnv (changes requested):
Don't see why this change is needed. Did I miss something?

nvflare/app_common/ccwf/server_ctl.py
SYangster: When there are multiple workflows (e.g swarm and cse), they will both handle this event. 

During the swarm workflow, the cse workflow will also call `_update_client_status()` and at `if client_name not in self.client_statuses:` since self.client_statuses is not set yet, it will continuously log an error `self.log_error(fl_ctx, f"received result from unknown client {client_name}!")`.

Some options are to change it from log_error or check here to see if self.client_statuses is set, but I was thinking to check beforehand to see if the current workflow matches to avoid calling `_update_client_status()` unnecessarily.

yanchengnv: I see. To fix this, you could modify the _update_client_status() method, instead of here, since we want one method to make the right the decision (not spreading decision making to multiple places).
Instead of changing log level to debug, we should change the order of condition checking - check workflow_id first.

yanchengnv (approved):
LGTM

PR #2514, opened 2024-04-17
Simulator end run for all clients

yhwen:
Fixes # 

Address the simulator run large number of clients performance issue.

### Description

Simulator only runs the END_RUN events for those active running clients when the workflow finishes the run. All those swapped out clients need to be swapped in and re-created to run the END_RUN event handling. Most of the applications the END_RUN event handling are not needed when the client processes are not active. Change the simulator to provide an "end_run_for_all" option. The default is false. Only set it to run END_RUN event handling for all clients when explicitly set.

Also change to use multi-threads to run the END_RUN event handling if needed.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/private/fed/app/simulator/simulator.py
+++ b/nvflare/private/fed/app/simulator/simulator.py
@@ -30,6 +30,7 @@ def define_simulator_parser(simulator_parser):
     simulator_parser.add_argument("-t", "--threads", type=int, help="number of parallel running clients")
     simulator_parser.add_argument("-gpu", "--gpu", type=str, help="list of GPU Device Ids, comma separated")
     simulator_parser.add_argument("-m", "--max_clients", type=int, default=100, help="max number of clients")
+    parser.add_argument('--end_run_all', default=False, action=argparse.BooleanOptionalAction)
 
 
 def run_simulator(simulator_args):
@@ -41,6 +42,7 @@ def run_simulator(simulator_args):
         threads=simulator_args.threads,
         gpu=simulator_args.gpu,
         max_clients=simulator_args.max_clients,
+        end_run_all=simulator_args.end_run_all
     )
     run_status = simulator.run()
 
--- a/nvflare/private/fed/app/simulator/simulator.py
+++ b/nvflare/private/fed/app/simulator/simulator.py
@@ -30,7 +30,12 @@ def define_simulator_parser(simulator_parser):
     simulator_parser.add_argument("-t", "--threads", type=int, help="number of parallel running clients")
     simulator_parser.add_argument("-gpu", "--gpu", type=str, help="list of GPU Device Ids, comma separated")
     simulator_parser.add_argument("-m", "--max_clients", type=int, default=100, help="max number of clients")
-    parser.add_argument("--end_run_all", default=False, action=argparse.BooleanOptionalAction)
+    simulator_parser.add_argument(
+        "--end_run_for_all",
+        default=False,
+        action=argparse.BooleanOptionalAction,
+        help="flag to indicate if running END_RUN event for all clients",
+    )
 
 
 def run_simulator(simulator_args):
@@ -42,7 +47,7 @@ def run_simulator(simulator_args):
         threads=simulator_args.threads,
         gpu=simulator_args.gpu,
         max_clients=simulator_args.max_clients,
-        end_run_all=simulator_args.end_run_all,
+        end_run_for_all=simulator_args.end_run_for_all,
     )
     run_status = simulator.run()
 
--- a/nvflare/private/fed/app/simulator/simulator.py
+++ b/nvflare/private/fed/app/simulator/simulator.py
@@ -33,7 +33,7 @@ def define_simulator_parser(simulator_parser):
     simulator_parser.add_argument(
         "--end_run_for_all",
         default=False,
-        action='store_true',
+        action="store_true",
         help="flag to indicate if running END_RUN event for all clients",
     )
 
--- a/nvflare/private/fed/app/simulator/simulator.py
+++ b/nvflare/private/fed/app/simulator/simulator.py
@@ -33,7 +33,7 @@ def define_simulator_parser(simulator_parser):
     simulator_parser.add_argument(
         "--end_run_for_all",
         default=False,
-        action=argparse.BooleanOptionalAction,
+        action='store_true',
         help="flag to indicate if running END_RUN event for all clients",
     )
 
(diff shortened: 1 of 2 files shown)


nvflare/private/fed/app/simulator/simulator.py
```diff
@@ -41,6 +42,7 @@ def run_simulator(simulator_args):
         threads=simulator_args.threads,
```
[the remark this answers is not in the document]
yhwen: changed it to "end_run_for_all". Also added a help message to explain the flag meaning.

nvflare/private/fed/app/simulator/simulator_runner.py
```diff
@@ -571,9 +577,26 @@ def run_client_thread(self, num_of_threads, gpu, lock, rank, timeout=60):
                         self.end_run_clients.append(end_run_client)
 
                 client.simulate_running = False
+
+            if end_run_all:
+                self._end_run_clients(client, gpu, lock, num_of_threads, timeout)
         except Exception as e:
             self.logger.error(f"run_client_thread error: {secure_format_exception(e)}")
 
+    def _end_run_clients(self, client, gpu, lock, num_of_threads, timeout):
```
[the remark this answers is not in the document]
yhwen: Added the "end_run_for_all" meaning in the add_argument() to explain.

nvflare/private/fed/app/simulator/simulator_runner.py
```diff
@@ -519,17 +529,13 @@ def run(self, gpu):
...
             for i in range(self.args.threads):
-                executor.submit(lambda p: self.run_client_thread(*p), [self.args.threads, gpu, lock, i, timeout])
+                executor.submit(
+                    lambda p: self.run_client_thread(*p), [self.args.threads, gpu, lock, self.args.end_run_all, timeout]
+                )
 
             # wait for the server and client running thread to finish.
             executor.shutdown()
 
-            for client in self.federated_clients:
```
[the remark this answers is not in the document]
yhwen: Yes, the slow down is caused by large number of clients need to be swapped in the running process again to run the END_RUN event. All the clients need to be recreated and run the swapped in process.

nvflare/private/fed/app/simulator/simulator_runner.py
```diff
@@ -571,9 +578,40 @@ def run_client_thread(self, num_of_threads, gpu, lock, rank, timeout=60):
...
+        These client END_RUN event handler only execute when "end_run_for_all" has been set.
+
+        Multiple client running threads will try to pick up the client from the same clients pool.
+
+        """
+        # Each thread only stop picking up the NOT-DONE client until all clients have run the END_RUN event.
+        while len(self.end_run_clients) != len(self.federated_clients):
+            with lock:
+                end_run_client = self._pick_next_client()
+                self.end_run_clients.append(end_run_client.client_name)
```
YuanTingHsieh, line 599: end_run_client could be None

nvflare/private/fed/app/simulator/simulator_runner.py
```diff
@@ -571,9 +578,40 @@ def run_client_thread(self, num_of_threads, gpu, lock, rank, timeout=60):
...
+            if end_run_client:
+                self.do_one_task(
+                    end_run_client, num_of_threads, gpu, lock, timeout=timeout, task_name=RunnerTask.END_RUN
+                )
+                with lock:
```
YuanTingHsieh, line 604: can't acquire lock inside with lock , will deadlock

YuanTingHsieh (changes requested):

yanchengnv:
Do we really need these changes?
@YuanTingHsieh what would happen if we don't do?

nvflare/private/fed/app/simulator/simulator_runner.py
```diff
@@ -571,9 +578,40 @@ def run_client_thread(self, num_of_threads, gpu, lock, rank, timeout=60):
...
+                self._end_run_clients(gpu, lock, num_of_threads, timeout)
         except Exception as e:
             self.logger.error(f"run_client_thread error: {secure_format_exception(e)}")
 
+    def _end_run_clients(self, gpu, lock, num_of_threads, timeout):
+        """After the WF reaches the END_RUN, each running thread will try to pick up one of the remaining client
+        which has not run the END_RUN yet, then execute the END_RUN handler, until all the clients have done so.
+        These client END_RUN event handler only execute when "end_run_for_all" has been set.
+
+        Multiple client running threads will try to pick up the client from the same clients pool.
```
yanchengnv, line 592: Are they threads or processes?

YuanTingHsieh:
If we are not firing END_RUN events in the simulator, based on our current components that utilizes this event, we will have following issues when running in simulator:

1. Metrics streaming results could be missing when running in simulator, but since https://github.com/NVIDIA/NVFlare/discussions/2477 report the current one also has this issue so I guess the current END_RUN mechanism not helps
2. Some components will not be gracefully shutdown, for example the external file pipe will just timeout and print "PEER_GONE", this will not affect correctness of job. There will be potential warning / error messages print out that does not affect the correctness.

We could warn the users saying that simulator will not fire END_RUN / ABOUT_TO_END_RUN / CHECK_FOR_END_RUN_READINESS events.

nvflare/private/fed/app/simulator/simulator_runner.py
yhwen: This is not acquire a lock inside another lock. They are separate.

nvflare/private/fed/app/simulator/simulator_runner.py
yhwen: When running the simulator worker, it's in an individual sub-process. But they are each managed and created by the multi-threads.

yhwen:
> If we are not firing END_RUN events in the simulator, based on our current components that utilizes this event, we will have following issues when running in simulator:
> 
> 1. Metrics streaming results could be missing when running in simulator, but since [How to avoids lost tracking messages and lost events #2477](https://github.com/NVIDIA/NVFlare/discussions/2477) report the current one also has this issue so I guess the current END_RUN mechanism not helps
> 2. Some components will not be gracefully shutdown, for example the external file pipe will just timeout and print "PEER_GONE", this will not affect correctness of job. There will be potential warning / error messages print out that does not affect the correctness.
> 
> We could warn the users saying that simulator will not fire END_RUN / ABOUT_TO_END_RUN / CHECK_FOR_END_RUN_READINESS events.

The "--end_run_for_all" option is to allow the user to run the END_RUN event handling for all the clients. Users can choose to run in this mode if needed for the Application. By default it's False.

yanchengnv:
Then this PR is not worth it.

nvflare/private/fed/app/simulator/simulator_runner.py
yanchengnv: @YuanTingHsieh you mis-read the code. The first "with lock" only applies to line 598.

YuanTingHsieh (approved):
LGTM

PR #2523, opened 2024-04-19
[2.5] TIE (Technology for Integrating Everything) and Flower Inegration

yanchengnv:
Fixes # .

### Description

This PR adds the following features:

- TIE - Technology for Integrating Everything. Refactored from XGBoost integration work to become general-purpose, TIE allows external applications to be executed independently while using FLARE as a reliable communication mechanism. 

- Flower integration using TIE. On the server side, two Flower processes are launched: superlink and server app. On the client side, one Flower process is launched: the client app. Communications among Flower processes use gRPC protocol, but relayed through Flare's ReliableMessage mechanism. 

### Types of changes

- [ ] Documentation updated.


nvflare/app_common/tie/process_mgr.py
```diff
@@ -0,0 +1,148 @@
...
+            lf = open(log_file, "a")
+            self.log_file = lf
+
+        if lf and log_stdout:
+            stdout = subprocess.PIPE
+        elif lf and not log_stdout:
+            stdout = lf
+        else:
+            stdout = None
+
```
YuanTingHsieh, line 69: If users don't specify log_file, and set log_stdout = True.
We will set stdout=None, and nothing gets printed out?

nvflare/app_common/tie/process_mgr.py
```diff
@@ -0,0 +1,148 @@
...
+            sys.stdout.flush()
+
+    def poll(self):
+        """Perform a poll request on the process.
+
+        Returns: None if the process is still running; an exit code (int) if process is not running.
+
+        """
+        if not self.process:
+            return 0
```
YuanTingHsieh, line 104: if not process, we want to return 0 which means the same thing as "the command or program executed successfully without any errors." ?

nvflare/app_common/tie/connector.py
```diff
@@ -0,0 +1,394 @@
...
+
+        Returns: a tuple of (whether the applet is stopped, exit code)
+
+        """
+        if self.applet_env == Constant.APPLET_ENV_SELF:
+            if self.applet:
+                return self.applet.is_stopped()
+            else:
+                self.logger.warning("applet is not set with the connector")
+                return True, 0
```
YuanTingHsieh, line 318: should we have a different exit code other than 0?
maybe None or some other NVFlare-defined number?

nvflare/app_opt/flower/connectors/grpc_client_connector.py
```diff
@@ -0,0 +1,128 @@
...
+        # abort the FL client
+        with self.engine.new_context() as fl_ctx:
+            self.system_panic(reason, fl_ctx)
+
+    def SendReceive(self, request: pb2.MessageContainer, context):
+        """Process request received from a Flower client.
+        This implements the SendReceive method required by Flower gRPC server (LGS on FLARE Client).
+        1. convert the request to a Shareable object.
+        2. send the Shareable request to FLARE server.
+        3. convert the
```
YuanTingHsieh, line 109: unfinished sentence

nvflare/app_opt/flower/grpc_server.py
```diff
@@ -0,0 +1,82 @@
+# Copyright (c) 2023, NVIDIA CORPORATION.  All rights reserved.
```
YuanTingHsieh, line 1: 2024

nvflare/app_opt/flower/utils.py
YuanTingHsieh, line 1: if this is general enough, in later re-factoring of xgboost we can move this file outside of flower

nvflare/app_opt/flower/mock/applet.py
YuanTingHsieh, line 1: suggest move mock into testing codes/folder

nvflare/app_common/tie/controller.py
```diff
@@ -0,0 +1,565 @@
...
+        """
+        Constructor
+
+        Args:
+            configure_task_name - name of the config task
+            configure_task_timeout - time to wait for clients’ responses to the config task before timeout.
+            start_task_name - name of the start task
+            start_task_timeout - time to wait for clients’ responses to the start task before timeout.
+            job_status_check_interval - how often to check client statuses of the job
+            max_client_op_interval - max amount of time allowed between XGB ops from a client
```
YuanTingHsieh, line 77: XGB should be changed

nvflare/app_common/tie/controller.py
```diff
@@ -0,0 +1,565 @@
...
+    def _is_stopped(self):
+        # check whether the abort signal is triggered
+        return self.abort_signal and self.abort_signal.triggered
+
+    def _update_client_status(self, fl_ctx: FLContext, op=None, client_done=False):
+        """Update the status of the requesting client.
+
+        Args:
+            fl_ctx: FL context
+            op: the XGB operation requested
```
YuanTingHsieh, line 236: XGB should be changed

nvflare/app_common/tie/controller.py
```diff
@@ -0,0 +1,565 @@
...
+
+        """
+        self.log_debug(fl_ctx, f"_handle_app_request {topic}")
+        op = request.get_header(Constant.MSG_KEY_OP)
+        if self._is_stopped():
+            self.log_warning(fl_ctx, f"dropped app request ({op=}) since server is already stopped")
+            return make_reply(ReturnCode.SERVICE_UNAVAILABLE)
+
+        # since XGB protocol is very strict, we'll stop the control flow when any error occurs
+        process_error = "XGB request process error"
```
YuanTingHsieh, line 311: XGB should be changed

nvflare/app_common/tie/controller.py
```diff
@@ -0,0 +1,565 @@
...
+
+        if exit_code == 0:
+            self.log_info(fl_ctx, f"app client is done with exit code {exit_code}")
+        elif exit_code == Constant.EXIT_CODE_CANT_START:
+            self.log_error(fl_ctx, f"app client failed to start (exit code {exit_code})")
+            self.system_panic("app client failed to start", fl_ctx)
+        else:
+            # Should we stop here?
+            # Problem is that even if the exit_code is not 0, we can't say the job failed.
+            self.log_warning(fl_ctx, f"app client is done with exit code {exit_code}")
```
YuanTingHsieh, line 286: I think we should take a more conservative approach and stop when the exit code is not 0?

What are the situations where we want it to continue?

nvflare/app_common/tie/controller.py
```diff
@@ -0,0 +1,565 @@
...
+        """
+        self.log_debug(fl_ctx, f"_handle_app_request {topic}")
+        op = request.get_header(Constant.MSG_KEY_OP)
+        if self._is_stopped():
+            self.log_warning(fl_ctx, f"dropped app request ({op=}) since server is already stopped")
+            return make_reply(ReturnCode.SERVICE_UNAVAILABLE)
+
+        # since XGB protocol is very strict, we'll stop the control flow when any error occurs
+        process_error = "XGB request process error"
+        self._update_client_status(fl_ctx, op=op)
```
YuanTingHsieh, line 312: A general thought, should we be monitoring and handling the client status in each controller?

Could some of these status be general that it is worth doing at a lower level or wrap as an common method/API of the base class Controller?

YuanTingHsieh (commented):
Mostly good, added some comments

nvflare/app_common/tie/applet.py
```diff
@@ -0,0 +1,67 @@
...
+
+        """
+        pass
+
+    @abstractmethod
+    def is_stopped(self) -> (bool, int):
+        """Called to check whether the applet is already stopped.
+
+        Returns: whether the applet is stopped. If stopped, the exit code.
+
```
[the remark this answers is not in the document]
yanchengnv: Changed

examples/hello-world/hello-flower/requirements.txt
```diff
@@ -0,0 +1,4 @@
+nvflare~=2.5.0rc
+flwr[simulation]>=1.8.0
+torch==2.2.1
+torchvision==0.17.1
```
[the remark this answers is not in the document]
yanchengnv: Leave for Holger later.

nvflare/app_opt/flower/proto/grpcadapter.proto
```diff
@@ -0,0 +1,13 @@
+syntax = "proto3";
+
+package flower.proto;
```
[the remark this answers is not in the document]
yanchengnv: Yes, but we need to change the package name from flwr.proto to flower.proto, otherwise protobuf will complain about namespace conflict when both Flare and Flower are installed.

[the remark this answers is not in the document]
yanchengnv: Added comments to proto definition.

nvflare/app_common/tie/controller.py
yanchengnv: Good catch! Changed to replace XGB with "app".

nvflare/app_common/tie/cli_applet.py
```diff
@@ -0,0 +1,126 @@
...
+        if not mgr:
+            return
+
+        # wait for the applet to stop by itself
+        start = time.time()
+        while time.time() - start < timeout:
+            rc = mgr.poll()
+            if rc is not None:
+                # already stopped
+                self.logger.info(f"applet stopped ({rc=}) gracefully after {time.time()-start} seconds")
```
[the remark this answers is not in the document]
yanchengnv: removed "gracefully".

nvflare/app_common/tie/cli_applet.py
```diff
@@ -0,0 +1,126 @@
...
+        rc = mgr.stop()
+        if rc is None:
+            self.logger.warn(f"killed the applet process after waiting {timeout} seconds")
+
+    def is_stopped(self) -> (bool, int):
+        if self._start_error:
+            return True, Constant.EXIT_CODE_CANT_START
+
+        if self._proc_mgr:
+            return_code = self._proc_mgr.poll()
```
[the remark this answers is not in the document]
yanchengnv: Yes, good catch! Changed to use a light-weight solution to assign self._proc_mgr to mgr, and then only work with mgr. Even if self._proc_mgr is set to None in the middle, it won't cause trouble.

nvflare/app_common/tie/connector.py
```diff
@@ -0,0 +1,394 @@
...
+        """
+        try:
+            if not self.in_process:
+                # enable logging
+                run_dir = self.workspace.get_run_dir(self.job_id)
+                log_file_name = os.path.join(run_dir, "applet_log.txt")
+                configure_logging(self.workspace)
+                add_log_file_handler(log_file_name)
+            self.applet.start(ctx)
+            self.stopped = True
```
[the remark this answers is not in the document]
yanchengnv: No. self.applet.start() does not return until the applet is done. This code is from refactored from XGB, where the applet (called runner in XGB) is started in a separate thread or subprocess.

nvflare/app_common/tie/connector.py
```diff
@@ -0,0 +1,394 @@
...
+        if applet_env not in VALID_APPLET_ENV:
+            raise ValueError(f"invalid applet_env {applet_env}: must be in {VALID_APPLET_ENV}")
+
+        self.in_process = self.applet_env == Constant.APPLET_ENV_THREAD
+
+    def set_applet(self, applet: Applet):
+        """Set the applet that will be used to run app processing logic.
+        Note that the connector is only responsible for starting the applet appropriately (in a thread or in a
+        separate process).
+
```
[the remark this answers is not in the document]
yanchengnv: Improved. But the applet will never be started within the same thread.

nvflare/app_common/tie/connector.py
```diff
@@ -0,0 +1,394 @@
...
+            raise RuntimeError(f"connector_stopped_cb must be callable but got {type(connector_stopped_cb)}")
+
+        # start the monitor in a separate daemon thread!
+        t = threading.Thread(target=self._monitor, args=(fl_ctx, connector_stopped_cb), daemon=True)
+        t.start()
+
+    def start_applet(self, app_ctx: dict, fl_ctx: FLContext):
+        """Start the applet set to the connector.
+        If self.in_process is True, then the applet will be started in a separate thread.
+        If self.in_process is False, then the applet will be started in a separate process.
```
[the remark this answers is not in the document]
yanchengnv: The APPLET_ENV_SELF means that the applet itself is responsible for starting the external app (e.g. cli_applet), whereas the in_process flag only applies when the connector is responsible for starting the applet (the XGB model). In this model, the applet runs python directly. In case of APPLET_ENV_SELF, the applet may or may not be python function at all.

nvflare/app_common/tie/connector.py
```diff
@@ -0,0 +1,394 @@
...
+            operation result
+        """
+        request.set_header(Constant.MSG_KEY_OP, op)
+        if not target:
+            target = FQCN.ROOT_SERVER
+
+        if not fl_ctx:
+            fl_ctx = self.engine.new_context()
+
+        self.logger.debug(f"sending request with RM: {request=}; {type(request['flower.headers'])}")
```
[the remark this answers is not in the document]
yanchengnv: Good catch! Changed.

nvflare/app_common/tie/executor.py
```diff
@@ -0,0 +1,197 @@
...
+        Note that this method is always called after the "configure" method, hence it's possible to dynamically
+        determine the connector's config based on the config params in the "configure" step.
+
+        Args:
+            fl_ctx: the FL context
+
+        Returns: a dict of config params
+
+        """
+        pass
```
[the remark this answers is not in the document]
yanchengnv: Good idea.

nvflare/app_common/tie/connector.py
```diff
@@ -0,0 +1,394 @@
...
+
+        """
+        try:
+            if not self.in_process:
+                # enable logging
+                run_dir = self.workspace.get_run_dir(self.job_id)
+                log_file_name = os.path.join(run_dir, "applet_log.txt")
+                configure_logging(self.workspace)
+                add_log_file_handler(log_file_name)
+            self.applet.start(ctx)
```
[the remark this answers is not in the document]
yanchengnv: Correct. Note that this piece of code runs in a separate thread or subprocess.

nvflare/app_common/tie/process_mgr.py
yanchengnv: Actually I'm not sure. None is the default value of stdout for popen and it seems to print it to sys.stdout by default.
Maybe we should manage it through our code. I tried to avoid the use of another thread.
We can wait a bit before doing anything because in reality both log file and log_stdout will be used.

nvflare/app_common/tie/connector.py
yanchengnv: This happens when is_applet_stopped is called before the applet is started (a logic error) or after the applet is stopped, which could happen but harmless.  Don't think we need a special code for this because the warning log message can tell us.

The use of the applet-related methods can not be random - they are called in specific order by other components (controller or executor) to manage its life cycle.

nvflare/app_opt/flower/utils.py
yanchengnv: True. I have the same feeling. This file used to have some other flower-specific functions, but not only contains this generic function.

It probably should belong to somewhere in fuel.

nvflare/app_opt/flower/mock/applet.py
yanchengnv: Let's keep it here for now. It's not exactly for the testing in regular sense.

nvflare/app_common/tie/controller.py
yanchengnv: Changed to "app".

nvflare/app_common/tie/controller.py
yanchengnv: Changed.

nvflare/app_common/tie/controller.py
yanchengnv: This should be determined by the nature of the external system design. In case of XGB, we know it's MPI based and hence fragile. But Flower may not be that fragile. 

In general, Flare cannot decide whether the job should be aborted if one client app failed.

nvflare/app_common/tie/controller.py
yanchengnv: Let's think about that. I can see that some of the client lifecycle management logic (e.g. no activity for an extended period of time) may apply to most/all controllers. In fact, similar logic already exists for xgb, swarm learning (or CWF), and TIE.

We should think about how to consolidate them into a base controller.

nvflare/apis/utils/reliable_message.py
```diff
@@ -345,7 +376,7 @@ def _monitor_req_receivers(cls):
                 now = time.time()
                 for tx_id, receiver in cls._req_receivers.items():
                     assert isinstance(receiver, _RequestReceiver)
-                    if receiver.rcv_time and now - receiver.rcv_time > 4 * receiver.tx_timeout:
+                    if receiver.rcv_time and now - receiver.rcv_time > receiver.tx_timeout:
```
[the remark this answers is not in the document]
yanchengnv: Actually the changes from 2.4 are already incorporated - I resolved some conflicts. Please review.

nvflare/app_common/tie/process_mgr.py
yanchengnv: BTW, I changed to manage all of these thru our code so will leave no ambiguity.

nvflare/app_common/tie/process_mgr.py
yanchengnv: This case happens only when you call the poll() method before the process is even started. This use is invalid. 
I will change to raise exception.

nvflare/app_opt/flower/connectors/grpc_client_connector.py
yanchengnv: Fixed

nvflare/app_common/tie/connector.py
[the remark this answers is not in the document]
YuanTingHsieh: Got it, I think this "Applet" class is always in Python in our code, but "external app" can be any language as you said.

We can also have APPLET_ENV_PROCESS combine with CLIApplet.
This would mean that Connector will spawn a process to call "self.applet.start()".
Then CLIApplet start will spawn another process to run external script (external app).

For the applet.start() method, whether users want to use spawn process to start external script (CLIApplet) or they just signaling other process to do anything depends on the users' implementation.

Suggest enhance the docstring and code logic to the following:

```
if self.applet_env is Constant.APPLET_ENV_SELF, then "self.applet.start" will be called in the same thread as the connector.
If self.applet_env is Constant.APPLET_ENV_THREAD, then "self.applet.start" will be called in a separate thread.
If self.applet_env is Constant.APPLET_ENV_PROCESS, then "self.applet.start" will be called in a separate process.
```

Another note, I think the Applet class abstract is similar to the Launcher abstract. We might be able to consolidate them later.

YuanTingHsieh (commented):
One question regarding the relationship about Applet, external app, APPLET_ENV.

Other changes LGTM.

yanchengnv:
> One question regarding the relationship about Applet, external app, APPLET_ENV.
> 
> Other changes LGTM.

You have a good point. Applet **represents** external app to Flare. It could be the external app itself, or a CLI to spawn the external app, or any other way.

I think a cleaner way is to simply let the applet decide how to manage external app, instead of building some of the logic into Connector (inherited from XGB integration).

I will refactor for the clean design.

nvflare/app_opt/flower/connectors/grpc_server_connector.py
```diff
@@ -0,0 +1,109 @@
...
+
+    def _is_stopped(self) -> (bool, int):
+        runner_stopped, ec = self.is_applet_stopped()
+        if runner_stopped:
+            self.logger.info("applet is stopped!")
+            return runner_stopped, ec
+
+        if self._server_stopped:
+            self.logger.info("Flower grpc server is stopped!")
+            return True, self._exit_code
```
YuanTingHsieh, line 57: Where do we set this "self._exit_code"?

nvflare/app_opt/flower/connectors/flower_connector.py
```diff
@@ -0,0 +1,144 @@
...
+        """
+        num_rounds = config.get(Constant.CONF_KEY_NUM_ROUNDS)
+        if num_rounds is None:
+            raise RuntimeError("num_rounds is not configured")
+
+        check_positive_int(Constant.CONF_KEY_NUM_ROUNDS, num_rounds)
+        self.num_rounds = num_rounds
+
+    def _send_flower_request(self, request: Shareable) -> Shareable:
+        """Send Flower request to the FL server via FLARE message.
```
YuanTingHsieh, line 125: minor, personally prefer "send_request_to_flare" for this method to match the server side method "send_request_to_flower"

nvflare/app_common/tie/py_applet.py
```diff
@@ -0,0 +1,242 @@
...
+        else:
+            if self.process:
+                assert isinstance(self.process, multiprocessing.Process)
+                ec = self.process.exitcode
+                if ec is None:
+                    return False, 0
+                else:
+                    return True, ec
+            else:
+                return True, 0
```
YuanTingHsieh, line 242: I am also thinking is that, is it better to have one common class with a flag "self.in_process" to control its behavior
or we could have two classes one is ThreadedPyApplet, the other ProcessPyApplet or something.

nvflare/app_common/tie/py_applet.py
```diff
@@ -0,0 +1,242 @@
...
+            # must remove the fl_context from ctx since it's not pickleable!
+            ctx.pop(Constant.APP_CTX_FL_CONTEXT)
+            self.logger.info("starting applet in another process")
+            self.process = multiprocessing.Process(
+                target=starter.start,
+                args=(ctx,),
+                daemon=True,
+                name="applet",
+            )
+            self.process.start()
```
YuanTingHsieh, line 187: we could enhance this to be more clear:

```
        if self.in_process:
            self._start_in_thread(ctx)
        else:
            self._start_in_process(ctx)

    def _start_in_thread(self, ctx: dict):
        """Start the applet in a separate thread."""
        self.logger.info("Starting applet in another thread")
        thread = threading.Thread(
            target=self.starter.start,
            args=(ctx,),
            daemon=True,
            name="applet"
        )
        thread.start()
        if not self.starter.started:
            self.logger.error(f"Cannot start applet: {self.starter.error}")
            raise RuntimeError(self.starter.error)

    def _start_in_process(self, ctx: dict):
        """Start the applet in a separate process."""
        ctx.pop(Constant.APP_CTX_FL_CONTEXT, None)
        self.logger.info("Starting applet in another process")
        self.process = multiprocessing.Process(
            target=self.starter.start,
            args=(ctx,),
            daemon=True,
            name="applet"
        )
        self.process.start()
```

YuanTingHsieh (dismissed):
Thanks!
Mostly LGTM, minor notes.

nvflare/app_common/tie/py_applet.py
yanchengnv: Not really. The reason is that the "runner" is the same for both cases.

nvflare/app_opt/flower/connectors/grpc_server_connector.py
yanchengnv: good catch. fixed

nvflare/app_opt/flower/connectors/flower_connector.py
yanchengnv: This is to send a request containing Flower content to the FL Server.

nvflare/app_common/tie/py_applet.py
yanchengnv: Good idea!

YuanTingHsieh (approved):
thanks, LGTM

nvflare/app_common/tie/connector.py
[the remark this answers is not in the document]
yanchengnv: Changed to use PyApplet for external python code that can be run in_process or not.

YuanTingHsieh (approved):

PR #2530, opened 2024-04-25
Add client controller executor

SYangster:
- follow up to https://github.com/NVIDIA/NVFlare/pull/2390
- added ClientControllerExecutor to run controllers on client side using WFCommClient aux channel communication, and distribute final results to clients
- example job template to showcase using pre-existing ScatterAndGather and CrossSiteModelEval controllers on the client side with no additional changes needed

### Types of changes

- [ ] Documentation updated.


nvflare/apis/impl/wf_comm_client.py
```diff
@@ -92,27 +96,42 @@ def broadcast_and_wait(
...
+            if task_cb_error:
+                return self._make_error_reply(ReturnCode.ERROR, targets)
+
+        if task.timeout < 0:
+            raise ValueError(f"The task timeout must >= 0. But got {task.timeout}")
 
-        if task.timeout <= 0:
-            raise ValueError(f"The task timeout must > 0. But got {task.timeout}")
+        if task.timeout == 0:
+            task.timeout = MAX_TASK_TIMEOUT
```
YuanTingHsieh, line 107: This might not be clear to the users, we need to document this somewhere saying set to 0 will be MAX_TASK_TIMEOUT

nvflare/apis/impl/wf_comm_client.py
```diff
@@ -57,7 +59,6 @@ def broadcast_and_wait(
         abort_signal: Signal = None,
     ):
         engine = fl_ctx.get_engine()
-        request = task.data
         # apply task filters
```
YuanTingHsieh, line 62: if we do the old way of `request = task.data` what is the problem of it?

nvflare/apis/impl/wf_comm_client.py
```diff
@@ -220,21 +239,49 @@ def send(
...
     ):
         engine = fl_ctx.get_engine()
 
         self._validate_target(engine, targets)
 
-        return self.send_and_wait(task, fl_ctx, targets, send_order, task_assignment_timeout)
+        for target in targets:
+            reply = self.broadcast_and_wait(task, fl_ctx, [target], abort_signal=abort_signal)
+            if reply.get_return_code() == ReturnCode.OK:
+                return reply
```
YuanTingHsieh, line 261: So our `send_and_wait` is expected to be 
"send 1st -> 1st execution -> recv 1st -> send 2nd -> 2nd execution -> recv 2nd"?

Or we can do these in parallel:

"send 1st -> 1st execution -> recv 1st"
"send 2nd -> 2nd execution -> recv 2nd"
"send 3rd -> 3rd execution -> recv 3rd"

nvflare/app_common/ccwf/client_controller_executor.py
```diff
@@ -0,0 +1,356 @@
...
+            if self.workflow_done:
+                return
+            report = self._get_status_report()
+            if not report:
+                self.log_debug(fl_ctx, "nothing to report this time")
+                return
+            self._add_status_report(report, fl_ctx)
+            self.last_status_report_time = report.timestamp
+
+        elif event_type in [EventType.ABORT_TASK, EventType.END_RUN]:
```
YuanTingHsieh, line 143: Do we want `ABORT_TASK` here to shutdown the executor?
What if after `ABORT_TASK` this client executor receive a new task from server, what will happen?

YuanTingHsieh (commented):
mostly good, add more comments

job_templates/sag_cse_cc_pt/meta.conf
```diff
@@ -0,0 +1,8 @@
+name = "sag_cse_cc_pt"
```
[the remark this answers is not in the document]
SYangster: good point, maybe sag_cse_ccwf_pt or sag_cse_client_pt?

nvflare/apis/impl/wf_comm_client.py
```diff
@@ -28,6 +28,8 @@
 from nvflare.private.privacy_manager import Scope
 from nvflare.security.logging import secure_format_exception
 
+MAX_TASK_TIMEOUT = 3600
```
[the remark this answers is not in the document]
SYangster: I can add this as a argument to WFCommClient, so it can be configurable in ClientControllerExecutor

nvflare/app_common/aggregators/intime_accumulate_model_aggregator.py
```diff
@@ -196,6 +196,11 @@ def accept(self, shareable: Shareable, fl_ctx: FLContext) -> bool:
             return False
 
         contributor_name = shareable.get_peer_prop(key=ReservedKey.IDENTITY_NAME, default="?")
+
+        peer_ctx = fl_ctx.get_peer_context()
+        if contributor_name == "?" and peer_ctx:
+            contributor_name = peer_ctx.get_identity_name("?")
+
```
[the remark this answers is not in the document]
SYangster: the shareable IDENTITY_NAME prop is not set in the WFCommClient case, however the peer identity name can also be found in the peer context in this case

nvflare/private/fed/utils/fed_utils.py
```diff
@@ -314,7 +314,7 @@ def get_target_names(targets):
             continue
 
         if name not in target_names:
-            target_names.append(t)
+            target_names.append(name)
     return target_names
```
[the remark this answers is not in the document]
SYangster: yes, I can submit PRs for the other branches too

nvflare/app_common/workflows/model_controller.py
```diff
@@ -63,7 +63,7 @@ def __init__(
             raise ValueError("task_check_period must be greater than 0.")
         self._task_check_period = task_check_period
         self._persistor_id = persistor_id
-        self._persistor = None
+        self.persistor = None
```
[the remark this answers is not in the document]
SYangster: I made this change to standardize since all of our existing controllers used `self.persistor` rather than `self._persistor`. This will help in the optional case of distributing the final results, as we can check for a controller’s `self.persistor` attribute.

nvflare/app_common/ccwf/server_ctl.py
```diff
@@ -182,13 +184,16 @@ def start_controller(self, fl_ctx: FLContext):
             allow_none=False,
         )
 
+        random.shuffle(self.participating_clients)
+        self.log_info(fl_ctx, f"Using participating clients: {self.participating_clients}")
```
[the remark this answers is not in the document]
SYangster: right good idea, I can remove this shuffle, I had this here before I added the “random” `starting_client_policy`

nvflare/fuel/utils/validation_utils.py
```diff
@@ -263,6 +266,8 @@ def validate_candidate(var_name: str, candidate, base: list, default_policy: str
             return ""
         elif default_policy == DefaultValuePolicy.ANY:
             return base[0]
+        elif default_policy == DefaultValuePolicy.RANDOM:
+            return random.choice(base)
```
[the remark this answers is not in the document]
SYangster: this would replace the shuffle

nvflare/apis/impl/wf_comm_client.py
SYangster: I can add this to the `class Task` docstring https://github.com/NVIDIA/NVFlare/blob/main/nvflare/apis/controller_spec.py#L91

nvflare/app_common/aggregators/intime_accumulate_model_aggregator.py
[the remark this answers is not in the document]
YuanTingHsieh: instead of doing it here, maybe we want to somehow set it in upstream (WFCommClient) so any other aggregators that try to get this "IDENTITY_NAME" can get it?

job_templates/sag_cse_cc_pt/meta.conf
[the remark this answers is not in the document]
YuanTingHsieh: even now is ok, this is subjective though, you can go either
sag_cse_ccwf_pt or sag_cse_cc_pt

nvflare/apis/impl/wf_comm_client.py
SYangster: This is because the `before_task_sent_cb` of CrossSiteModelEval sets the `task.data` to new data rather than modifying the original data, so `request` will still point to the old data. To fix this I moved `request = task.data` to right before `send_aux_request()` so it will point to the correct new `task.data`.

https://github.com/NVIDIA/NVFlare/blob/main/nvflare/app_common/workflows/cross_site_model_eval.py#L262

nvflare/apis/impl/wf_comm_client.py
SYangster: `send_and_wait()` was originally defined in our controller as sending a task to a single target, wherein the list of targets is just a list of priority of backup targets in case sending fails. The parallel version is essentially `broadcast()`, and the sequential version would be `relay()`.

nvflare/fuel/utils/validation_utils.py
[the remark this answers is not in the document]
YuanTingHsieh: ok, can your update the docstrings to reflect the changes thanks

nvflare/apis/impl/wf_comm_client.py
YuanTingHsieh: got it, let's add a comment saying "we need to get this task.data since the user callback might change it"

nvflare/apis/impl/wf_comm_client.py
YuanTingHsieh: Does this mean that our `send_and_wait` now for multiple targets is almost equivalent to `relay_and_wait` with task_assignment_timeout set to 0?

nvflare/app_common/ccwf/client_controller_executor.py
SYangster: Right if `ABORT_TASK`, then the executor should be shutdown. If after `ABORT_TASK` the client executor receives a new task from the server, I can add a check at the beginning of `execute()` to see if `self.workflow_done`, which would be set in `finalize()`.

nvflare/apis/impl/wf_comm_client.py
SYangster: The difference is that `send_and_wait` will finish after sending the task to one successful target, while `relay_and_wait` will send the task to all of the targets in sequential order.

YuanTingHsieh: got it, sounds good, this is just following the original behavior (although that is a bit tricky to get around).
But for the scope of this PR, this is good.

nvflare/app_common/ccwf/client_controller_executor.py
```diff
@@ -0,0 +1,382 @@
...
+                status.last_round = last_round
+            elif last_round is not None and last_round > status.last_round:
+                status.last_round = last_round
+
+            status_dict = status.to_dict()
+            self.logger.info(f"updated my last status: {status_dict}")
+
+    def _process_final_result(self, request: Shareable, fl_ctx: FLContext) -> Shareable:
+        peer_ctx = fl_ctx.get_peer_context()
+        assert isinstance(peer_ctx, FLContext)
```
YuanTingHsieh, line 291: assert usually just used in test code

nvflare/app_common/ccwf/client_controller_executor.py
YuanTingHsieh: ok, so after `ABORT_TASK`, this executor should be shutdown.

If any chances in the future where we could "re-start" a task, then we should create a new instance of client controller executor?

nvflare/app_common/ccwf/client_controller_executor.py
```diff
@@ -0,0 +1,382 @@
...
+            return
+
+        self.me = fl_ctx.get_identity_name()
+
+        self.persistor = self.engine.get_component(self.persistor_id)
+        if not isinstance(self.persistor, LearnablePersistor):
+            self.log_warning(
+                fl_ctx, f"Persistor {self.persistor_id} must be a Persistor instance, but got {type(self.persistor)}"
+            )
+            self.persistor = None
```
YuanTingHsieh, line 113: Does this persistor have to be a `LearnablePersistor`?
If this is what it is as before then its fine for this PR, we can think in the future how to improve this class.

nvflare/app_common/ccwf/client_controller_executor.py
```diff
@@ -0,0 +1,382 @@
...
+            return make_reply(ReturnCode.BAD_REQUEST_DATA)
+
+        self.log_info(fl_ctx, f"Got final result from client {client_name}")
+
+        fl_ctx.set_prop(AppConstants.GLOBAL_MODEL, result, private=True, sticky=True)
+
+        if self.persistor:
+            self.persistor.save(result, fl_ctx)
+        else:
+            self.log_error(fl_ctx, "persistor not configured, model will not be saved")
```
YuanTingHsieh, line 310: Maybe we can add some additional enhancement saying if users really want to persist things. (like persist_per_round or something if set to 0 means no persist)
If they did not specify that then no persistor is not an error?

nvflare/app_common/ccwf/client_controller_executor.py
```diff
@@ -0,0 +1,382 @@
...
+        )
+
+        resp = self.controller.broadcast_and_wait(
+            task=task,
+            targets=targets,
+            min_responses=len(targets),
+            fl_ctx=fl_ctx,
+        )
+
+        assert isinstance(resp, dict)
```
YuanTingHsieh, line 363: usually should just be used in test code

YuanTingHsieh (commented):
If the original code has these logics, then we can leave it in this PR.

We can add enhancements later.

nvflare/app_common/aggregators/intime_accumulate_model_aggregator.py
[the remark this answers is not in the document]
YuanTingHsieh: @SYangster where did we do this now?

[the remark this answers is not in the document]
SYangster: https://github.com/NVIDIA/NVFlare/pull/2530/files#diff-7207f57b6f393aefda5f21823d25b52ab07902f0257b80dd73f7ea86c0104f4fR163

nvflare/app_common/ccwf/client_controller_executor.py
SYangster: I think since the ServerSideController and ClientControllerExecutor are paired, then we shouldn’t need to re-create an instance of the ClientControllerExecutor. A different ServerSideController could have a different corresponding ClientControllerExecutor however.

nvflare/app_common/ccwf/client_controller_executor.py
SYangster: Good point, will remove in this PR, and can eventually remove from ClientSideController

nvflare/app_common/ccwf/client_controller_executor.py
SYangster: This is how we checked persistors previously, will leave in this PR but agree that this can be improve in the future

job_templates/sag_cse_ccwf_pt/config_fed_client.conf
```diff
@@ -0,0 +1,183 @@
...
+
+          train_task_name = "train"
+          evaluate_task_name = "validate"
+          submit_model_task_name = "submit_model"
+        }
+      }
+    }
+    {
+      # All tasks prefixed with wf1_ are routed to this ClientControllerExecutor
+      tasks = ["wf1_*"]
```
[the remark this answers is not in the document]
SYangster: I was previously testing with multiple ClientControllerExecutors, I can remove the “1” here

nvflare/app_common/ccwf/client_controller_executor.py
SYangster: I can change to a warning, as this persistor is only used when distributing the final results

YuanTingHsieh: maybe change this to warning is better

nvflare/app_common/ccwf/client_controller_executor.py
YuanTingHsieh: Just hypothetical discussion so don't need to resolve in this PR.

For the same pair of ServerSideController and ClientControllerExecutor, if we can re-start a task in the future, we should think of how that is possible with these two classes.

YuanTingHsieh (approved):
thanks LGTM

PR #2536, opened 2024-04-29
Improve InProcessClientAPIExecutor 

chesterxgchen:
### Description
Prob 1: 
Current InProcessClientAPIExecutor requires user must have a main() function defined. The executor call main() to execute the training function.  Although this is not major restrictions, but does require user to define the training script to a function, and function must be called "main()" 

To remove this limitation, we re-implement the executor by using runpy.run_path(<script_path>, "```__main__```"). This approach doesn't require user to define main() function or doesn't require function at all. 

Prob 2
User use print() function the training script, we wish the print will go to logger so it captured into the logging files. Instead of asking user to change print() to logger. We does this for user automatically. all print() will goes to the logger without user to make any changes

Changes. 
1. rename ExeTaskFnWrapper class to TaskScriptRunner
2. Replace implementation InProcessClientAPI from calling a main() function to user runpy.run_path() to remove limitatation
3. redirect print() to logger.info()

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/private/fed/app/simulator/simulator_runner.py
+++ b/nvflare/private/fed/app/simulator/simulator_runner.py
@@ -468,7 +468,7 @@ class SimulatorRunner(FLComponent):
 
     def start_server_app(self, args):
         app_server_root = os.path.join(self.simulator_root, "server", SimulatorConstants.JOB_NAME, "app_server")
-        args.workspace = app_server_root
+        args.workspace = os.path.join(self.simulator_root, "server")
         os.chdir(args.workspace)
 
         log_file = os.path.join(self.simulator_root, "server", WorkspaceConstants.LOG_FILE_NAME)
--- a/tests/unit_test/private/fed/app/simulator/simulator_runner_test.py
+++ b/tests/unit_test/private/fed/app/simulator/simulator_runner_test.py
@@ -14,18 +14,24 @@
 
 import os
 import shutil
+import threading
+import time
 import uuid
-from unittest.mock import patch
+from tempfile import TemporaryDirectory
+from unittest.mock import Mock, patch
 
 import pytest
 
-from nvflare.apis.fl_constant import WorkspaceConstants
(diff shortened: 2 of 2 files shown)


nvflare/app_common/executors/in_process_client_api_executor.py
```diff
@@ -104,18 +105,17 @@ def handle_event(self, event_type: str, fl_ctx: FLContext):
...
             )
 
             self._task_fn_thread = threading.Thread(target=self._task_fn_wrapper.run)
             self._task_fn_thread.start()
 
             meta = self._prepare_task_meta(fl_ctx, None)
-            self.client_api = InProcessClientAPI(task_metadata=meta, result_check_interval=0.5)
-            self.client_api.init()
-            self._data_bus.put_data(CLIENT_API_KEY, self.client_api)
+            self._client_api = InProcessClientAPI(task_metadata=meta, result_check_interval=0.5)
```
IsaacYangSLA, line 116: Can this result_check_interval=0.5 be configurable?

nvflare/app_common/executors/exec_task_fn_wrapper.py
```diff
@@ -11,69 +11,67 @@
...
 import traceback
-from typing import Dict
 
-from nvflare.fuel.utils.function_utils import find_task_fn, require_arguments
+print_fn = builtins.print
 
 
-class ExecTaskFuncWrapper:
-    def __init__(self, task_fn_path: str, task_main_args: str = None, task_fn_args: Dict = None):
+class TaskScriptRunner:
```
SYangster, line 23: we should rename the filenames (this and the test file) to reflect new class name

SYangster (commented):
Good changes, will improve user experience. Minor comment

IsaacYangSLA (approved):
Good PR with test cases.

SYangster (approved):

PR #2543, opened 2024-04-30
handle cases where the script with relative path in Script Runner

chesterxgchen:
### Description
handle cases where the script with relative path in Script Runner

### Types of changes

- [ ] Documentation updated.


nvflare/app_common/executors/task_script_runner.py
```diff
@@ -62,8 +62,13 @@ def get_sys_argv(self):
 
     def get_script_full_path(self, script_path) -> str:
         target_files = None
+        script_filename = os.path.basename(script_path)
+        script_dir = os.path.dirname(script_path)
         for r, dirs, files in os.walk(os.getcwd()):
-            target_files = [os.path.join(r, f) for f in files if f == script_path]
+            path_components = r.split(os.sep)
+            if script_dir == "" or script_dir in path_components:
+                target_files = [os.path.join(r, f) for f in files if f == script_filename]
```
yhwen, line 70: This if condition may not handle the path like this "src/sub/cli.py", and search for "src/cli.py" so well. Probably we can simply do this:

            target_files = [os.path.join(r, f) for f in files if os.path.join(r, f).endswith(script_path)]

yhwen: what I mean is if there are two files in the script, one is "src/cli.py" another one is "src/sub/cli.py", then the solution may find the wrong script to execute. Of course if there only one in "src/sub/cli.py", you can not find it by "src/cli.py".

yhwen: if the script is in "src/sub/cli.py", search for "src/cli.py" should result with not found, instead of True.

nvflare/app_common/executors/task_script_runner.py
```diff
@@ -61,14 +61,27 @@ def get_sys_argv(self):
...
+        cwd = os.getcwd()
+        script_filename = os.path.basename(script_path)
+        script_dirs = os.path.dirname(script_path)
         for r, dirs, files in os.walk(os.getcwd()):
-            target_files = [os.path.join(r, f) for f in files if f == script_path]
-            if target_files:
+            for f in files:
+                absolute_path = os.path.join(r, f)
+                relative_path = os.path.relpath(absolute_path, cwd)
+                if relative_path == script_path:
```
yhwen, line 72: This is not going to work. The "absolute_path" is pointing to '/private/tmp/nvflare/jobs/workdir/site-1/simulate_job/app_site-1/custom/src/cifar10_fl.py', and the "cwd" is pointing to '/private/tmp/nvflare/jobs/workdir/site-1'. So this makes the "relative_path" to 'simulate_job/app_site-1/custom/src/cifar10_fl.py'.

The "cwd" needs to point to '/private/tmp/nvflare/jobs/workdir/site-1/simulate_job/app_site-1/custom'.

yhwen (changes requested):

nvflare/app_common/executors/task_script_runner.py
```diff
@@ -61,14 +61,29 @@ def get_sys_argv(self):
...
+                absolute_path = os.path.join(r, f)
+                if absolute_path.endswith(script_path):
+                    parent_dir = absolute_path[: absolute_path.find(script_path)].rstrip(os.sep)
+                    if os.path.isdir(parent_dir):
+                        target_file = absolute_path
+                        break
+
+                if not script_dirs and f == script_filename:
+                    target_file = absolute_path
+                    break
```
yhwen, line 79: I am not quite sure what's the purpose of doing these additional checks. But this unit test will fail. Do you want to support the absolute path script search?

    def test_app_scripts_and_args(self):
        curr_dir = os.getcwd()
        script_path = os.path.join(curr_dir, "nvflare/cli.py")
        script_args = "--batch_size 4"
        wrapper = TaskScriptRunner(script_path=script_path, script_args=script_args)

        self.assertTrue(wrapper.script_path.endswith(script_path))
        self.assertEqual(wrapper.get_sys_argv(), [os.path.join(curr_dir, "nvflare", "cli.py"), "--batch_size", "4"])

yhwen: Actually, we don't support the absolute path script in the add_ext_script(). so no need to handle this case.

yhwen (approved):

PR #2547, opened 2024-05-01
Improve AWS cloud launch script

### LINKED ISSUE #2540: [BUG] aws default vpc should not be required.
**Describe the bug**

nvflare commands bomb out if no default VPC is defined in AWS account, many AWS accounts (per AWS recommendation) have no default VPC assigned, however many enterprise accounts have exactly one VPC assigned so if there is only one VPC (check `aws ec2 describe-vpcs`) you do not have to ask the user. If there is none you can tell the user to create a default vpc if they have the permissions to do so:  `aws ec2 create-default-vpc`

**To Reproduce**

```
 dp@grammy:~/NVFlare$ nvflare dashboard --cloud aws
Dashboard launch script for cloud is written at /home/dp/NVFlare/aws_start_dsb.sh.  Now running the script.
One initial user will be created when starting dashboard.
Please enter the email address for this user.
xxxxx@domain.edu
Generating key pair for VM

An error occurred (VPCIdNotSpecified) when calling the CreateSecurityGroup operation: No default VPC for this user
Security group id:

usage: aws [options] <command> <subcommand> [<subcommand> ...] [parameters]
To see help text, you can run:

  aws help
  aws <command> help
  aws <command> <subcommand> help

aws: error: argument --group-id: expected one argument


usage: aws [options] <command> <subcommand> [<subcommand> ...] [parameters]
To see help text, you can run:

  aws help
  aws <command> help
  aws <command> <subcommand> help

aws: error: argument --group-id: expected one argument



 - OS: [e.g. wsl  windows 11]
 - Python Version [e.g. 3.9]
 - NVFlare Version [e.g. 2.4]

** Other comments 

Ask ChatGPT/Claude: 

Does aws recommend to not create a default vpc for enterprise aws accounts? 

Yes, AWS generally recommends not creating a default VPC for enterprise AWS accounts. Instead, they recommend creating custom VPCs tailored to your specific network requirements.
Here are a few reasons why:

Security: Default VPCs have predetermined settings, which might not align with your organization's security policies. Creating a custom VPC allows you to configure security groups, network ACLs, and other settings as per your needs.
IP Address Range: Default VPCs come with a predetermined IP address range. This might lead to IP address conflicts if you have multiple AWS accounts or if you need to connect your AWS resources with your on-premises network. With a custom VPC, you can choose your IP address range.
VPC Peering: If you need to set up VPC peering between multiple VPCs, using default VPCs can make the process more complex due to potential overlapping IP address ranges.
Compliance: Depending on your industry, you might have to adhere to certain compliance requirements. Default VPCs might not be set up to meet these requirements out of the box.
Customization: Custom VPCs give you complete control over your network setup, including subnets, route tables, gateways, and more. This level of customization is essential for many enterprise setups.

For these reasons, AWS recommends creating custom VPCs for enterprise accounts, as it provides more flexibility, security, and control over your network infrastructure. Cop

### PR DESCRIPTION
**Pull request by IsaacYangSLA**
Fixes #2540

### Description

This PR improves the user experiences on cloud launch scripts for AWS.  It checks if default-vpc exists.  If not, the script shows the command to create one and exits.

### Types of changes

- [ ] Documentation updated.

### CODE REVIEW

**Comment by YuanTingHsieh** [comment]
@IsaacYangSLA users have additional feedback in the issue, can you check? otherwise LGTM.

**Review by YuanTingHsieh — COMMENTED** [review]
Have you check the users latest feedback? otherwise this LGTM.

**Comment by YuanTingHsieh** [comment]
quote:

I think your initial fix is great, it certainly helped me, but a user prompt that would print both the command to create the default VPC and then may be nvflare dashboard --cloud aws --vpc xxxxxxxx if the user is not allowed to run 'aws ec2 create-default-vpc' would be much preferred. This does not increase the complexity for the users for whom it worked until now and you don't even have to document it, but those who run into problems, would get past a work stoppage issue.

**Comment by IsaacYangSLA** [comment]
I think this PR is the first step to address no default VPC issue.  I will comment on that issue.

**Comment by IsaacYangSLA** [comment]
The issue reporter agreed with my last comment on that issue.  So the issue was addressed by this PR.

File: nvflare/lighter/impl/aws_template.yml [hunk 1]
```diff
@@ -12,14 +12,21 @@ aws_start_sh: |
...
+
   if [ -z ${image_name+x} ]
   then
       container=false
   else
       container=true
   fi
 
-  if [ $container = true ]
+  if [ $container == "true" ]
```
YuanTingHsieh, line 29: why is the changing from "=" to "=="?

nvflare/lighter/impl/master_template.yml
```diff
@@ -819,6 +819,14 @@ cloud_script_header: |
         image_name=$2
         shift
       ;;
+      --vpc-id)
+        vpc_id=$2
+        shift
+      ;;
```
YuanTingHsieh, line 825: Does VPC only exist on AWS?

YuanTingHsieh (commented):
some questions, otherwise LGTM

nvflare/lighter/impl/master_template.yml
IsaacYangSLA: No.  So that will be ignored.

nvflare/lighter/impl/aws_template.yml
IsaacYangSLA: I tried to make all comparison to ``==`` as bash complains unary operator.

YuanTingHsieh (approved):

nvflare/lighter/impl/aws_template.yml
```diff
@@ -74,7 +100,12 @@ aws_start_sh: |
 
   # Generate Security Group
   # Try not reusing existing security group because we have to modify it for our own need.
-  sg_id=$(aws ec2 create-security-group --group-name $SECURITY_GROUP --description "NVFlare security group" | jq -r .GroupId)
+  if [ $using_default_vpc == true ]
```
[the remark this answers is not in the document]
IsaacYangSLA: Agreed.

PR #2564, opened 2024-05-07
[2.4] Fix get_target_names

SYangster:
correctly add Client.name to target_names

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/private/fed/utils/fed_utils.py
+++ b/nvflare/private/fed/utils/fed_utils.py
@@ -313,7 +313,7 @@ def get_target_names(targets):
             continue
 
         if name not in target_names:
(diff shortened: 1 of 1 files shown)


YuanTingHsieh (approved):
thanks, LGTM

nvflare/private/fed/utils/fed_utils.py
```diff
@@ -313,7 +313,7 @@ def get_target_names(targets):
             continue
 
         if name not in target_names:
-            target_names.append(t)
+            target_names.append(name)
     return target_names
```
YuanTingHsieh, line 317: it would be good if we just add a small test to this method.

PR #2565, opened 2024-05-07
allow absolute path for the external script

yhwen:
Fixes # .

### Description

- Allow to use the absolute path for the external script.
- If the external script exists in the job API running machine, the external script will be packaged into the job config folder.
- If the external script does not exists in the job API running machine, it will be treated as a pre-installed script on the job running machine. The job creation API will skip the packaging of this script. The job execution will load it from the job execution runtime.

### Types of changes

- [ ] Documentation updated.


nvflare/job_config/base_app_config.py
```diff
@@ -57,6 +57,9 @@ def add_ext_script(self, ext_script: str):
         if not os.path.exists(ext_script):
             raise RuntimeError(f"Could not locate external script: {ext_script}")
 
+        if os.path.isabs(ext_script):
```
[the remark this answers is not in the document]
yhwen: Changed the behavior to allow absolution path ext_scripts, but not copy the script during the job_config export. We won't check if the script exist or not.

nvflare/job_config/fed_job_config.py
```diff
@@ -154,9 +154,10 @@ def _get_server_app(self, config_dir, custom_dir, fed_app):
 
     def _copy_ext_scripts(self, custom_dir, ext_scripts):
         for script in ext_scripts:
-            dest_file = os.path.join(custom_dir, script)
-            module = "".join(script.rsplit(".py", 1)).replace(os.sep, ".")
-            self._copy_source_file(custom_dir, module, script, dest_file)
+            if not os.path.isabs(script):
```
holgerroth, line 157: What if I want to copy my scripts from an absolute path, e.g. /my_custom_scripts/...? I think this is the wrong place to handle any security concerns. If there's a concern, we should just not allow any custom folder when submitting the job.

holgerroth (changes requested):
This doesn't look like the right approach to me.

nvflare/job_config/fed_job_config.py
holgerroth: but we also want to use this job config API to generate a job folder that I can submit. So, in the case there's no security concern, why can I not submit the script using an absolute path?

yhwen: You can still generate the job folder using the job creation API. Only thing is that the absolute path script won't be generated within the job folder. It is assumed to be pre-installed in the particular location. The "submit_job" command can only submit the package within the job folder structure.

yhwen: Changed to the following behavior: (Added the unit tests)

1. When adding an absolute path external script, user also needs to provide the package path to that external script to generate the proper module codes.
2. If the external script exists in the job API running machine, the external script will be packaged into the job config folder.
3. If the external script does not exists in the job API running machine, it will be treated as a pre-installed script on the job running machine. The job creation API will skip the packaging of this script. The job execution will load it from the job execution runtime.

yhwen:
> This doesn't look like the right approach to me.

Changed to the following behavior: (Added the unit tests)

1. When adding an absolute path external script, user also needs to provide the package path to that external script to generate the proper module codes.
2. If the external script exists in the job API running machine, the external script will be packaged into the job config folder.
3. If the external script does not exists in the job API running machine, it will be treated as a pre-installed script on the job running machine. The job creation API will skip the packaging of this script. The job execution will load it from the job execution runtime.

nvflare/job_config/base_app_config.py
```diff
@@ -50,17 +50,23 @@ def add_task_data_filter(self, tasks: List[str], filter: Filter):
     def add_task_result_filter(self, tasks: List[str], filter: Filter):
         self._add_task_filter(tasks, filter, self.task_result_filters)
 
-    def add_ext_script(self, ext_script: str):
+    def add_ext_script(self, ext_script: str, package_path: str = None):
```
holgerroth, line 53: Can this be done using PYTHONPATH directly rather than user having to provide a package_path? cc @yanchengnv

yhwen: PYTHONPATH could have multiple matches, which may cause confusion, and cause the wrong codes generated.

yanchengnv: If ext_script is an absolute dir, then it can be two cases:
- The file contains main(), and the file does not belong to any package
- The file is part of a package.

In the 2nd case, I think we should use PYTHONPATH to determine what the file's relative dir should be.
For example, if the file name is /a/b/c/xyz.py, and the PYTHONPATH contains /a/b, then the file's relative dir should be c/xyz.py.

If, as Yuhong said, the PYTHONPATH also contains /a/b/c, then in theory, you could reference the xyz with "c.xyz" or "xyz".  But not sure this is a common case though.

If we really want to make it identical to user's env, perhaps, we could copy everything verbatim and simply replace "/" with "custom" in the job folder, and then add user's PYTHONPATH to our PYTHONPATH (with / substitued with custom).

In this example, the file /a/b/c/xyz.py will be copied into job folder:  custom/a/b/c/xyz.py.
If the user's PYTHONPATH is /a/b:/a/b/c, then when running on FL server/client, our PYTHONPATH will have: 
<job_dir>/custom/a/b:<job_dir>/custom/a/b/c.

This way, if the user's code has:

import xyz
or 
import c.xyz

It will continue to work.

yhwen: In the case this is a simple script, yes, the package_path may not make too much difference. However, if the script has another import from another source code file, then it's much difference case. The import could event use relative path as well.

When job creation API generating the codes, it needs to parse all the source codes to find all the dependent codes to be included as well.

yanchengnv: Then this sounds like very deep water, if we have to parse source code. How is it done? What if there are conditional imports? This is what the python interpreter does.

yhwen: Here's the codes to parse the source codes, and find all the dependent codes from the import. Conditional imports will be included as well.

https://github.com/NVIDIA/NVFlare/blob/2a71c61ed0774433431e8c920799aca39ac5a874/nvflare/job_config/fed_job_config.py#L189-L206

yanchengnv: This doesn't handle all cases. For example:

from \
    nvflare.app_common.tie.defs \
    import \
    Constant as TieConstant

It's almost like that we are writing a small lex parser here.

yhwen: yeah, the import are required to be in the same line.

yhwen: Added handle for the line break of continuous lines support, added unit tests.

Also removed the requirement of external absolute path script to provide the package path. Use the longest matching path in the PYTHONPATH instead.

holgerroth (approved):
Looks good. Nice work.

yhwen:
> The PR title ("not allow ..." ) and description ("allow ...") are opposite

updated.

PR #2579, opened 2024-05-14
Multiprocessing start method option

yhwen:
Fixes # .

### Description

Provide the option to choose which start method to use for multiprocessing.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/private/fed/app/server/server_train.py
+++ b/nvflare/private/fed/app/server/server_train.py
@@ -16,6 +16,7 @@
 
 import argparse
 import logging
+import multiprocessing
 import os
 import sys
 import time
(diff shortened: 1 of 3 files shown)


nvflare/private/fed/app/server/server_train.py
```diff
@@ -158,5 +160,7 @@ def parse_arguments():
 
     version_check()
     args = parse_arguments()
+    if args.start_method and args.start_method in ["spawn", "fork", "forkserver"]:
+        multiprocessing.set_start_method(args.start_method)
```
nvidianz, line 164: This may not work. The GPU subprocess is started from child process and this needs to set in the child process.

PR #2580, opened 2024-05-14
FOBS Enhancement

nvidianz:
Added 2 features to FOBS,
1. On serializing, DataClassDecomposer will be registered for an object if no decomposer is registered for this class. This feature can be turned off by calling fobs.auto_register_data_classes(False)
2. Decomposer registration is not needed on deserializing. The decomposer is stored in the serialized data and is automatically registered.

### Types of changes

- [ ] Documentation updated.


nvflare/fuel/utils/fobs/fobs.py
```diff
@@ -81,14 +143,28 @@ def unpack(self, obj: Any) -> Any:
...
-            if _enum_auto_register:
-                cls = self._load_class(type_name)
-                if issubclass(cls, Enum):
-                    register_enum_types(cls)
-                    error = False
-            if error:
-                raise TypeError(f"Unknown type {type_name}, caused by mismatching decomposers")
+            registered = False
+            decomposer_name = obj.get(FOBS_DECOMPOSER)
+            cls = _load_class(type_name)
```
yanchengnv, line 148: What happens if this call raises exception? Should it be in try block?

yanchengnv (changes requested):
Looks good in general. Please see my comments for improvement.

nvflare/fuel/utils/fobs/fobs.py
nvidianz: Added exception handling in _load_class() so only TypeError is raised.

nvflare/fuel/utils/fobs/fobs.py
```diff
@@ -143,6 +104,80 @@ def register(decomposer: Union[Decomposer, Type[Decomposer]]) -> None:
...
+        if type_name not in _decomposers:
+            registered = False
+            decomposer_name = obj.get(FOBS_DECOMPOSER)
+            cls = _load_class(type_name)
+            if not decomposer_name:
+                # Maintaining backward compatibility with auto enum registration
+                if _enum_auto_registration:
+                    if issubclass(cls, Enum):
+                        register_enum_types(cls)
+                        registered = True
```
YuanTingHsieh, line 158: Do we need this part?
Users are writing their own decomposers, but we control both ends of pack <-> unpack right?
So this new Packer without this code, should already compatible with all existing decomposers?

YuanTingHsieh (commented):
Mostly good, have some questions

nvflare/fuel/utils/fobs/fobs.py
nvidianz: This is needed to unpack buffer generated by old version.

yanchengnv (approved):
LGTM

PR #2583, opened 2024-05-16
Missing sj heartbeat

yhwen:
Fixes # .

### Description

Add the missing SJ heartbeat calls to the parent. Add the logic to abort the job if it's not supposed to be running.

### Types of changes

- [ ] Documentation updated.


nvflare/private/fed/server/server_engine.py
```diff
@@ -317,8 +317,6 @@ def abort_app_on_clients(self, clients: List[str]) -> str:
         return ""
 
     def abort_app_on_server(self, job_id: str, turn_to_cold: bool = False) -> str:
-        if job_id not in self.run_processes.keys():
```
yanchengnv, line 320: can you add a log message here to help understanding the sequence of event. This is a very unusual case.

yanchengnv (commented):

nvflare/private/fed/server/fed_server.py
```diff
@@ -687,12 +700,22 @@ def start_run(self, job_id, run_root, conf, args, snapshot):
...
+                self._send_parent_heartbeat(job_id)
                 time.sleep(self.check_engine_frequency)
 
         finally:
             self.engine.engine_info.status = MachineStatus.STOPPED
             self.run_manager = None
 
+    def _send_parent_heartbeat(self, job_id):
+        request = new_cell_message({CellMessageHeaderKeys.JOB_ID: job_id}, {})
+        self.cell.fire_and_forget(
```
yanchengnv, line 712: The cell may have not been created. Check to make sure self.cell is created before sending.

yanchengnv (changes requested):
Please check cell before sending request.

nvflare/private/fed/server/server_engine.py
yhwen: Added log message to show the abort SJ job which not supposed to be running.

nvflare/private/fed/server/fed_server.py
yhwen: The cell should have been created earlier before this. But added the check any way.

nvflare/private/fed/server/server_engine.py
YuanTingHsieh: So we don't need this check anymore here?

yhwen: We use the same abort_app_on_server() to abort the job which is not in the parent's running job list. In this case, within abort_app_on_server() we don't need to check this any more. The checking logic is at where calls this method.

nvflare/private/fed/server/fed_server.py
```diff
@@ -374,10 +376,23 @@ def _listen_command(self, request: Message) -> Message:
                 reply = make_cellnet_reply(F3ReturnCode.OK, "", None)
                 return reply
         elif command == ServerCommandNames.HEARTBEAT:
+            if job_id not in self.engine.run_processes:
```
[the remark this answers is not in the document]
yhwen: No, this means the job is not in the parent's job running list, however, the job is still sending heartbeat calls to the parent. In this case, we need to abort the child job process. The network interrupt does not affect this.

For example, the job supposed to be aborted at the server, or the admin action. However, the job child process didn't get the abort command in time, or didn't get it due to network interruption, this will ensure to send the abort command to the child process again.

yanchengnv (approved):
LGTM

PR #2603, opened 2024-05-29
Enhance the client sync_runner process to timeout use max_timeout, in…

yhwen:
Fixes # .

### Description

Enhance the client sync_runner process to timeout use MAX_RUNNER_SYNC_TIMEOUT, instead of using the MAX_RUNNER_SYNC_TRIES, because the server OK response will make the sync_runner call the return before the timeout.

### Types of changes

- [ ] Documentation updated.


yanchengnv (approved):
LGTM

PR #2606, opened 2024-05-31
Added target validation for job API

yhwen:
Fixes # .

### Description

- Added target validation for job API, also added the unit test.

1. must provide a target name
2. target name can not use special character

 

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/job_config/fed_job.py
+++ b/nvflare/job_config/fed_job.py
@@ -270,6 +270,9 @@ class FedJob:
         )
 
     def _validate_target(self, target):
+        if not target:
+            raise ValueError("Must provide a valid target name")
+
         if any(c in SPECIAL_CHARACTERS for c in target):
             raise ValueError(f"target {target} name contains invalid character")
         pass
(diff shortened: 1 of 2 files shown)


tests/unit_test/job_config/fed_job_test.py
```diff
@@ -0,0 +1,30 @@
...
+
+class TestFedJob:
+    def test_validate_targets(self):
+        job = FedJob()
+        component = FedAvg()
+
+        job.to(component, "server")
+        job.to(component, f"site-1")
+
+        with pytest.raises(Exception):
```
holgerroth, line 29: Should we test the target=None as well?

yhwen: Added the unit test for this case.

holgerroth (approved):
Looks good.

nvflare/job_config/fed_job.py
```diff
@@ -265,6 +269,14 @@ def simulator_run(self, workspace, threads: int = None):
             gpu=",".join([self._gpus[client] for client in self._gpus.keys()]),
         )
 
+    def _validate_target(self, target):
```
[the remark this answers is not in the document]
yhwen: We do allow the admin to use "@", however, we don't allow "@" as a site-name. The "@" character may cause several potential site job execution issue.

PR #2613, opened 2024-06-05
Added empty results check for aggregate_fn

yhwen:
Fixes # .

### Description

Added empty results check for aggregate_fn in base_fedavg. in the aggregate() will catch the raised exception and handle the aggregation error.

### Types of changes

- [ ] Documentation updated.


holgerroth (approved):
Makes sense!

PR #2616, opened 2024-06-06
XGBoost User Experience Improvment

nvidianz:
### Description
Made following changes to XGBoost,
1. Consolidate all XGBoost parameters to controller so all clients share the same parameters.

### Types of changes

- [ ] Documentation updated.


nvflare/app_opt/xgboost/histogram_based_v2/fed_controller.py
```diff
@@ -50,8 +60,14 @@ def __init__(
         )
         self.int_client_grpc_options = int_client_grpc_options
         self.in_process = in_process
+        self.server_handler_id = server_handler_id
 
     def get_adaptor(self, fl_ctx: FLContext):
+        if not self.server_handler_id:
```
yanchengnv, line 66: What is this for? Should we always create the handler so the user will never need to create it?

nvflare/app_opt/xgboost/histogram_based_v2/fed_executor.py
```diff
@@ -50,12 +51,15 @@ def __init__(
         self.in_process = in_process
 
     def get_adaptor(self, fl_ctx: FLContext):
+
+        if not self.client_handler_id:
```
yanchengnv, line 55: See my comment about this in Controller.

nvflare/app_opt/xgboost/histogram_based_v2/runners/xgb_client_runner.py
```diff
@@ -109,40 +115,60 @@ def run(self, ctx: dict):
...
-            early_stopping_rounds=self.early_stopping_rounds,
-            verbose_eval=self.verbose_eval,
-        )
+            self._xgb_params["device"] = f"cuda:{self._rank}"
 
+        self.logger.info(f"XGB trainging_mode: {self._training_mode} "
+                         f"params: {self._xgb_params} XGB options: {self._xgb_options}")
         self.logger.info(f"server address is {self._server_addr}")
+
+        xgb_plugin_name = ConfigService.get_str_var(name="xgb_plugin_name",
```
yanchengnv: Is plugin always needed, even in non-secure modes? What happens if XGBoost is started without plugin info?

nvflare/app_opt/xgboost/histogram_based_v2/runners/xgb_client_runner.py
```diff
@@ -109,40 +115,60 @@ def run(self, ctx: dict):
...
+                                                    conf=SystemConfigs.RESOURCES_CONF,
+                                                    default="nvflare")
+
+        xgb_loader_params = ConfigService.get_dict_var(name="xgb_loader_params", conf=SystemConfigs.RESOURCES_CONF)
```
yanchengnv, line 141: You can specify default={} in this call.

nvflare/private/fed/client/client_app_runner.py
```diff
@@ -106,6 +106,7 @@ def create_client_runner(self, app_root, args, config_folder, federated_client,
 
         run_manager = self.create_run_manager(args, conf, federated_client, workspace)
         federated_client.run_manager = run_manager
+        federated_client.runner_config = runner_config
```
yanchengnv, line 109: Why is this changed?

yanchengnv (changes requested):

nvflare/app_opt/xgboost/histogram_based_v2/fed_controller.py
nvidianz: This is for future use. Removed.

nvflare/app_opt/xgboost/histogram_based_v2/fed_executor.py
nvidianz: Removed.

nvflare/app_opt/xgboost/histogram_based_v2/runners/xgb_client_runner.py
nvidianz: Changed.

nvflare/private/fed/client/client_app_runner.py
nvidianz: To expose the add_component() call.

nvflare/app_opt/xgboost/histogram_based_v2/runners/xgb_client_runner.py
nvidianz: plugin is not used in non-secure mode. Changed to only set plugin info for secure mode.

yanchengnv (approved):
LGTM

examples/hello-world/hello-tf2/README.md
```diff
@@ -18,6 +18,23 @@ Install additional requirements:
...
+In scenarios where multiple clients are involved, you have to prevent TensorFlow from allocating all GPU memory 
+by setting the following flags.
+```bash
+TF_FORCE_GPU_ALLOW_GROWTH=true TF_GPU_ALLOCATOR=cuda_malloc_async
+```
+
+If you possess more GPUs than clients, a good strategy is to run one client on each GPU.
+This can be achieved using the `-gpu` argument during simulation, e.g., `nvflare simulator -n 2 -gpu 0,1 [job]`.
+
+
```
YuanTingHsieh, line 37: just a note some of the changes are already in the main branch so we will need to rebase this branch so we don't change 278 files in this PR

PR #2618, opened 2024-06-06
aws cloud deployment improvements 

dirkpetersen:
the following changes for aws cloud deployment have been tested with 6 configurations, including 3 with T4 GPU (g4dn.xlarge) and 3 versions of Ubuntu 

t2.small:
    20.04: ami-04bad3c587fe60d89
    22.04: ami-03c983f9003cb9cd1
    24.04: ami-0406d1fdd021121cd

g4dn.xlarge:
    20.04: ami-04bad3c587fe60d89
    22.04: ami-03c983f9003cb9cd1
    24.04: ami-0406d1fdd021121cd

Note: nvflare installs on 24.04 but is currently not supported 

- changed default image to ami-03c983f9003cb9cd1 (22.04 / Python 3.10)
- added g4dn.xlarge as an option to prompt EC2_TYPE
- fixed typo in prompt REGION
- increase the default block device size by 8GB
- install apt package nvidia-driver-535-server if GPU found
- run modprobe nvidia to avoid reboot if GPU found
- adding ~/.local/bin to PATH
- add --break-system-packages to pip install (required by Python 3.12)
- add --no-cache-dir to pip install to avoid disk space issues
- add @reboot cronjob to ensure nvflare is restarted after a server (re)start

Fixes # .

### Description

The critical change is here the disk size increase by 8GB. This results in all tested configurations having between 30% and 60% free disk space. The default install can run out of disk space if nvidia drivers, pytorch and a few other packages need to be installed 

### Types of changes

- [x] Documentation updated.


IsaacYangSLA:
@dirkpetersen , thanks for the PR.  It improves the template quite a lot.  However, we may need to have our internal discussion on the deployment scenario.  We know different jobs require different dependencies.  Sometimes, pytorch is one of the dependencies and thus the deployment will need larger volumes.  We also see numpy-only jobs.
I will approve this PR.  Later, we might have additional PR's that change this PR after our internal discussions reach conclusions.

IsaacYangSLA (approved):
Good PR.  It covers some scenarios we did not consider previously.

dirkpetersen:
Thanks @IsaacYangSLA, this helps us a lot as we no longer have run a patched version. Another thing to consider is that this will not run on systems with GPU that run RHEL based images such as amazon linux, only ubuntu.  The AL/ML community, that does not use Ubuntu, seems to be small  though.

dirkpetersen:
@IsaacYangSLA , as i was working with other AWS regions, I noticed that some of our users would be challenged to look up appropriate AMI id without help from a cloud engineer as these ids are different in each region, also extensive testing triggered a number of fine tuning improvements that I would like to add to a pull request. I assume it might be better to have this current pull request go through and then create a new one after that or is it better to add to this one, looking for your guidance 

I have made the following changes: 

- get exact GPU requirements from ../local/resources.json.[default]
- add function to propose best suitable instance type (find_ec2_gpu_instance_type)
- check environment to get AWS region (default still us-west-2) 
- propose newest image id per region based on pattern, e.g "ubuntu-*-22.04-amd64-pro-server"
- add AWS_PROFILE hint to allow for different AWS profile
- call "aws sts get-caller-identity" to see if SSO token is still active
- change 'prompt' command to read -e -i <default> -p for better prompt editing 
- changed ssh commands to single quote for readability and avoid escaping chars
- clarify requirements.txt instruction
- set --region for all AWS commands to allow multi-region support
- adding suffixes to KEY_FILE, nvflare.log etc to allow parallel deployments 
- install additional os packages (python3-dev & gcc) to packages without wheels (e.g. arm64)
- use driver-550-server os package for >= 22.04 and driver-535-server for <= 20.04
- separated os and user packages to allow for failed install of os packages 
- starting os and user package install in parallel but wait twice for install of os packages
- support waiting for build of nvidia.ko as well as nvidia.ko.zst (for newer os)
- improve final message to allow for copy/paste of ssh and tail command
- switch default instance type to t4g.small image name pattern has arm64 in it.
- change control structures to "if [ something ]; then" for readability.  

the last one is of course optional based on your team's coding preferences 

I tested the following combinations over the last couple of days but only the 24.04 ARM option seems to fail with nvidia drivers 535 and 550

image

The new UI would look like this: 

```
startup/start.sh --cloud aws
Note: run this command first for a different AWS profile:
  export AWS_PROFILE=your-profile-name.

Checking AWS identity ...

* Cloud EC2 region, press ENTER to accept default: us-east-2
* Cloud AMI image name, press ENTER to accept default (use amd64 or arm64): ubuntu-*-22.04-amd64-pro-server
    retrieving AMI ID for ubuntu-*-22.04-amd64-pro-server ... ami-0a7e9bed072bb379b found
    finding smallest instance type with 1 GPUs and 15360 MiB VRAM ... g6.xlarge found
* Cloud EC2 type, press ENTER to accept default: g6.xlarge
* Cloud AMI image id, press ENTER to accept default: ami-0a7e9bed072bb379b
region = us-east-2, EC2 type = g6.xlarge, ami image = ami-0a7e9bed072bb379b , OK? (Y/n)
If the client requires additional Python packages, please add them to:
    /home/dp/NVFlare/dirk/Test/AWS-T4.X/startup/requirements.txt
Press ENTER when it's done or no additional dependencies.

Checking if default VPC exists
Default VPC found
Generating key pair for VM
Creating VM at region us-east-2, this may take a few minutes ...
VM created with IP address: 52.14.44.113
Copying files to nvflare_client
Destination folder is ubuntu@52.14.44.113:/var/tmp/cloud
Installing os packages as root in the background, this may take a few minutes ...
Installing user space packages in the background, this may take a few minutes ...
System was provisioned, packages may continue to install in the background.
To terminate the EC2 instance, run the following command:
  aws ec2 terminate-instances --region us-east-2 --instance-ids i-0837e105f2661a4e3
Other resources provisioned
security group: nvflare_client_sg_2036
key pair: NVFlareClientKeyPair
review install progress:
  tail -f /tmp/nvflare-aws-YGR.log
login to instance:
  ssh -i /home/dirk/AWS-T4.X/NVFlareClientKeyPair_i-0837e105f2661a4e3.pem ubuntu@52.14.44.113
```

here is the current aws_start.sh script i am using: 
https://raw.githubusercontent.com/dirkpetersen/nvflare-cancer/main/aws_start.sh

PR #2624, opened 2024-06-10
[2.5] Support log file rotation in admin commands

yanchengnv:
Fixes # .

### Description

This PR fixes https://nvbugspro.nvidia.com/bug/4686492

When a log file (log.txt) is rotated, the previous file name becomes "log.txt.1". and file extension becomes "1". Admin commands like "cat" won't work from this file.

This PR fixes this problem.

### Types of changes

- [ ] Documentation updated.


SYangster (commented):
Changes look good. Maybe we also need to apply a similar fix to the `_validate_file_string()` for the fl admin api, and the `validate_file_string()` in the cmd_arg_utils.py for the flare api?

yanchengnv:
> Changes look good. Maybe we also need to apply a similar fix to the `_validate_file_string()` for the fl admin api, and the `validate_file_string()` in the cmd_arg_utils.py for the flare api?

Good catch! Applied the changes as suggested.

SYangster (approved):
Thanks LGTM! Good idea to reuse get_file_extension

PR #2637, opened 2024-06-12
Changed the ModuleScanner to ignore any Exception -- [2.4]

yhwen:
Fixes # .

### Description

Changed the ModuleScanner to ignore any Exception.

### Types of changes

- [ ] Documentation updated.


yanchengnv:
I do not think we should swallow exceptions. We need to at least discuss the pros and cons before committing to this solution.

PR #2654, opened 2024-06-18
Add feedback when users launch dashboard to cloud with editable version

IsaacYangSLA:
### Description

When users run nvflare dashboard --cloud azure or --cloud aws from non-released NVFlare (like pip install -e . with source codes), the cloud scripts fail install nvflare in the cloud VM instances because user's NVFlare version is not publicly available to those VM instances.

This PR detect if '+' is inside the NVFlare version string.  If it is, the version users use to launch dashboard on cloud is definitely not an official release and the dashboard launch command will display a message asking users installing official NVFlare version from PyPi.

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh (approved):
Thanks, LGTM

PR #2662, opened 2024-06-24
Fix DockerBuilder to add `build` section

apatole:
Fixes # .

The `compose.yml` generated by the `DockerBuilder` class adds a build section only for the overseer. As a result, when we have non-HA mode `project.yml`, generated `compose.yml` won't have the build section and so `docker compose build` won't build the required NVFlare docker image.

This change adds a build section in the generated `compose.yml` even for clients and servers. It fixes the above issue and also with this change, same `compose.yml` can be used to build/run individual servers and clients using the `docker compose` method.

### Types of changes


apatole:
Hi @chesterxgchen , uploaded a small fix for the DockerBuilder. I was using DockerBuilder last week and was wondering why `docker compose build` is not building the required `nvflare-service` image. I then later realized that the generated `compose.yml` was missing the `build` sections. thanks.

IsaacYangSLA (approved):
Actually, we currently do not support docker compose in non-HA mode.  Users are free to create their own docker compose folder and yaml file when running in non-HA mode.  We may change the entire provision process to fit future updates.
For now, this PR is good.

PR #2665, opened 2024-06-25
Use npm for pages deployment

SYangster:
Change github pages site deploy workflow to use npm to avoid using unverified actions

### Types of changes

- [ ] Documentation updated.


.github/workflows/deploy.yml
```diff
@@ -4,10 +4,13 @@ on:
   # Trigger the workflow every time you push to the `main` branch
   # Using a different branch name? Replace `main` with your branch’s name
   push:
-    branches: [ web ]
```
YuanTingHsieh, line 7: Do we need web branch anymore if we are doing this?

If there is inconsistency between main and stable release , which branch should we use?

PR #2667, opened 2024-06-26
[2.4] Release RM request receiver after finished

yanchengnv:
Fixes # .

### Description

This PR fixes a potential memory issue in the receiving side of the ReliableMessage.

When a request is received, a RequestReceiver object is created and kept in a table. Currently, after the request is finished, we still keep the RequestReceiver object in the table. RM has a monitoring process that will eventually release the object from the table. However this could take many minutes (depending on how tx_timeout is configured) before this happens. The problem is that all reply messages (could be very big) will remain in the table until the monitoring process remove them. This could potentially cause OOM if there are many requests in quick succession.

This PR solve this problem by releasing the RequestReceiver object immediately after the request is done and replied successfully. For requests that cannot be replied successfully, the RequestReceiver object will still remain in the table such that the requester can query it later. But the object will only stay in the table for tx_timeout seconds.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/apis/utils/reliable_message.py
+++ b/nvflare/apis/utils/reliable_message.py
@@ -165,6 +165,9 @@ class _RequestReceiver:
             # reply sent successfully!
             self.reply_time = time.time()
             ReliableMessage.debug(fl_ctx, f"sent reply successfully in {time_spent} secs")
+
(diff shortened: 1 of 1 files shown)


YuanTingHsieh (approved):
LGTM

PR #2672, opened 2024-06-27
[2.4] Remove ipcl requirement for other training mode except vertical secure

nvidianz:
Fixes FLARE-2053

### Description

Made changes so ipcl-python package is only required for vertical secure training using nvflare plugin.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/apis/event_type.py
+++ b/nvflare/apis/event_type.py
@@ -34,7 +34,6 @@ class EventType(object):
     JOB_COMPLETED = "_job_completed"
     JOB_ABORTED = "_job_aborted"
     JOB_CANCELLED = "_job_cancelled"
-    JOB_CONFIGURED = "_job_configured"
 
     BEFORE_PULL_TASK = "_before_pull_task"
     AFTER_PULL_TASK = "_after_pull_task"
--- a/nvflare/apis/event_type.py
+++ b/nvflare/apis/event_type.py
@@ -34,6 +34,7 @@ class EventType(object):
     JOB_COMPLETED = "_job_completed"
     JOB_ABORTED = "_job_aborted"
     JOB_CANCELLED = "_job_cancelled"
+    JOB_CONFIGURED = "_job_configured"
 
     BEFORE_PULL_TASK = "_before_pull_task"
     AFTER_PULL_TASK = "_after_pull_task"
--- a/nvflare/app_opt/xgboost/histogram_based_v2/defs.py
+++ b/nvflare/app_opt/xgboost/histogram_based_v2/defs.py
@@ -107,6 +107,7 @@ class Constant:
     EVENT_BEFORE_ALL_GATHER_V = "xgb.before_all_gather_v"
     EVENT_AFTER_ALL_GATHER_V = "xgb.after_all_gather_v"
     EVENT_XGB_JOB_CONFIGURED = "xgb.job_configured"
+    EVENT_XGB_ABORTED = "xgb.aborted"
 
     HEADER_KEY_ENCRYPTED_DATA = "xgb.encrypted_data"
     HEADER_KEY_HORIZONTAL = "xgb.horizontal"
--- a/nvflare/app_opt/xgboost/histogram_based_v2/defs.py
+++ b/nvflare/app_opt/xgboost/histogram_based_v2/defs.py
@@ -73,6 +73,7 @@ class Constant:
     ERR_TARGET_ERROR = -4
 
     EXIT_CODE_CANT_START = 101
+    EXIT_CODE_JOB_ABORT = 102
 
     # XGB operation parameter keys
     PARAM_KEY_RANK = "xgb.rank"
@@ -105,6 +106,7 @@ class Constant:
     EVENT_AFTER_BROADCAST = "xgb.after_broadcast"
     EVENT_BEFORE_ALL_GATHER_V = "xgb.before_all_gather_v"
     EVENT_AFTER_ALL_GATHER_V = "xgb.after_all_gather_v"
+    EVENT_XGB_JOB_CONFIGURED = "xgb.job_configured"
 
     HEADER_KEY_ENCRYPTED_DATA = "xgb.encrypted_data"
     HEADER_KEY_HORIZONTAL = "xgb.horizontal"
--- a/nvflare/app_opt/xgboost/histogram_based_v2/defs.py
+++ b/nvflare/app_opt/xgboost/histogram_based_v2/defs.py
@@ -86,6 +86,7 @@ class Constant:
     PARAM_KEY_REPLY = "xgb.reply"
     PARAM_KEY_REQUEST = "xgb.request"
     PARAM_KEY_EVENT = "xgb.event"
+    PARAM_KEY_TRAINING_MODE = "xgb.training_mode"
 
     RUNNER_CTX_SERVER_ADDR = "server_addr"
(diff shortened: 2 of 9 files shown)


nvflare/app_opt/xgboost/histogram_based_v2/sec/client_handler.py
```diff
@@ -91,6 +92,9 @@ def _process_before_broadcast(self, fl_ctx: FLContext):
             self.info(fl_ctx, "no clear gh pairs - ignore")
             return
 
+        if not ipcl_imported:
+            return self._abort("Required package ipcl-python is missing", fl_ctx)
+
```
YuanTingHsieh, line 97: Do we want this check here?
Could there be plugins that does not need ipcl?
Should we check this flag elsewhere?

YuanTingHsieh: when we use cuda plugin, will we enter this "_process_before_broadcast" ?

nvflare/app_opt/xgboost/histogram_based_v2/defs.py
```diff
@@ -86,6 +86,7 @@ class Constant:
     PARAM_KEY_REPLY = "xgb.reply"
     PARAM_KEY_REQUEST = "xgb.request"
     PARAM_KEY_EVENT = "xgb.event"
+    PARAM_KEY_TRAINING_MODE = "xgb.training_mode"
```
YuanTingHsieh, line 89: make sure we document different options for training mode

nvflare/app_opt/xgboost/histogram_based_v2/sec/client_handler.py
yanchengnv: Yes, this is always entered. Instead of checking ipcl_imported, you could check the existence of self.adder because the adder will be needed. This way, it will also work even if the adder is based on cuda.

nvidianz: We need to check here because we don't know which plugin goes through this flow. 

Modified to check encryptor instead of ipcl is imported.

nvflare/app_opt/xgboost/histogram_based_v2/defs.py
nvidianz: The modes are described in detail in documentation.

nvflare/apis/event_type.py
```diff
@@ -34,6 +34,7 @@ class EventType(object):
     JOB_COMPLETED = "_job_completed"
     JOB_ABORTED = "_job_aborted"
     JOB_CANCELLED = "_job_cancelled"
+    JOB_CONFIGURED = "_job_configured"
```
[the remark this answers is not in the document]
nvidianz: Yeah. Moved to defs.py

nvflare/app_opt/xgboost/histogram_based_v2/sec/client_handler.py
```diff
@@ -365,11 +369,15 @@ def _process_after_all_gather_v_horizontal(self, fl_ctx: FLContext):
...
     def handle_event(self, event_type: str, fl_ctx: FLContext):
-        if event_type == EventType.START_RUN:
-            self.public_key, self.private_key = generate_keys(self.key_length)
-            self.encryptor = Encryptor(self.public_key, self.num_workers)
-            self.decrypter = Decrypter(self.private_key, self.num_workers)
-            self.adder = Adder(self.num_workers)
+        if event_type == EventType.JOB_CONFIGURED:
+            training_mode = fl_ctx.get_prop(Constant.PARAM_KEY_TRAINING_MODE)
+            if training_mode in {"vertical_secure", "vs"}:
+                if ipcl_imported:
```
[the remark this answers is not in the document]
nvidianz: Yes. This needs to change when we support multiple adder etc.

nvflare/app_opt/xgboost/histogram_based_v2/sec/sec_handler.py
```diff
@@ -51,6 +52,7 @@ def error(self, fl_ctx: FLContext, msg: str):
     def _abort(self, error: str, fl_ctx: FLContext):
         self.error(fl_ctx, error)
         self.system_panic(reason=error, fl_ctx=fl_ctx)
+        XGBExecutor.notify_client_done(Constant.EXIT_CODE_JOB_ABORT, fl_ctx)
```
yanchengnv, line 55: See my previous comment: change to fire the XGB_ABORTED event and do not call system_panic here - let the executor's event handler call it.

yanchengnv (changes requested):
See my comments about the aborting process.

nvflare/app_opt/xgboost/histogram_based_v2/executor.py
```diff
@@ -85,10 +85,16 @@ def handle_event(self, event_type: str, fl_ctx: FLContext):
             adaptor.set_abort_signal(self.abort_signal)
             adaptor.initialize(fl_ctx)
             self.adaptor = adaptor
+        elif event_type == Constant.EVENT_XGB_ABORTED:
+            error = fl_ctx.get_prop(FLContextKey.FATAL_SYSTEM_ERROR)
+            self.log_error(fl_ctx, error)
+            self.system_panic(reason=error, fl_ctx=fl_ctx)
```
yanchengnv, line 91: Please check with YT about PR https://github.com/NVIDIA/NVFlare/pull/2674. There he changed _notify_client_done() to call system_panic() if rc != 0. I think we need to keep that change, then you don't need to call system_panic() here since the rc Constant.EXIT_CODE_JOB_ABORT is not 0.

yanchengnv (changes requested):
Very nice! 

See my comment about Yuan-Ting's PR https://github.com/NVIDIA/NVFlare/pull/2674, especially the changes to notify_client_done.  Make sure that system_panic() is not called multiple times.

If you put all his changes to your PR, then his PR won't be needed.

Please coordinate with YT.

YuanTingHsieh (dismissed):
LGTM

nvflare/app_opt/xgboost/histogram_based_v2/executor.py
```diff
@@ -85,10 +85,13 @@ def handle_event(self, event_type: str, fl_ctx: FLContext):
             adaptor.set_abort_signal(self.abort_signal)
             adaptor.initialize(fl_ctx)
             self.adaptor = adaptor
+        elif event_type == Constant.EVENT_XGB_ABORTED:
```
yanchengnv, line 88: Since we changed SecurityHandler's _abort() method to simply fire this event (and not call system_panic()), and SecurityHandler is also inherited by ServerHandler (and it also calls self._abort), we need to update the XGBController to also catch this event and call system_panic(). 

 Make sure you call super().handle_event() from XGBController's event handler since its parent Controller also has logic in its handle_event() method!

yanchengnv (changes requested):
Need to change XGBController accordingly since we changed the behavior of self._abort() of SecurityHandler. See my comments for detail.

nvflare/app_opt/xgboost/histogram_based_v2/executor.py
YuanTingHsieh: good catch, @nvidianz ^ Yan suggest we add this logic to XGBController as well.

PR #2674, opened 2024-06-27
[2.4] Enhance error handling

YuanTingHsieh:
### Description

1. When XGB Client stopped with RC other than 0, the whole job should stop.
2. enhance log msgs       

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/app_opt/xgboost/histogram_based_v2/executor.py
+++ b/nvflare/app_opt/xgboost/histogram_based_v2/executor.py
@@ -149,6 +149,7 @@ class XGBExecutor(Executor):
         """
         if rc != 0:
             self.log_error(fl_ctx, f"XGB Client stopped with RC {rc}")
+            self.system_panic(f"XGB Client stopped with non zero RC {rc}", fl_ctx)
         else:
             self.log_info(fl_ctx, "XGB Client Stopped")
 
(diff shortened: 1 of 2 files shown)


nvidianz (approved):
LGTM

nvflare/app_opt/xgboost/histogram_based_v2/executor.py
```diff
@@ -149,6 +149,7 @@ def _notify_client_done(self, rc, fl_ctx: FLContext):
         """
         if rc != 0:
             self.log_error(fl_ctx, f"XGB Client stopped with RC {rc}")
+            self.system_panic(f"XGB Client stopped with non zero RC {rc}", fl_ctx)
```
yanchengnv, line 152: Do we need to call it here? If rc is not 0, is the system already aborted?

yanchengnv (changes requested):
Please see my comments and check whether the system_panic call is necessary when rc is not 0.

nvflare/app_opt/xgboost/histogram_based_v2/executor.py
YuanTingHsieh: When I am doing testing, I found there are a few situations where the RC return things other than 0, but the job continues.

It keep happening if we use "in_process=False" and there is trouble loading the plugin ".so" (due to either environment issue or typos)

PR #2675, opened 2024-06-27
Add min_responses to send_model_and_wait()

SYangster:
#[4717625](https://nvbugspro.nvidia.com/bug/4717625)

- expose `min_responses` argument in ModelController's `send_model_and_wait()`
- hide `wait_time_after_min_received` argument, set default to 0

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/app_common/workflows/base_model_controller.py
+++ b/nvflare/app_common/workflows/base_model_controller.py
@@ -102,7 +102,7 @@ class BaseModelController(Controller, FLComponentWrapper, ABC):
         task_name: str = AppConstants.TASK_TRAIN,
         data: FLModel = None,
         targets: Union[List[Client], List[str], None] = None,
-        min_responses: int = 0,
+        min_responses: int = None,
         timeout: int = 0,
         wait_time_after_min_received: int = 0,
         blocking: bool = True,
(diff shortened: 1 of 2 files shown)


nvflare/app_common/workflows/base_model_controller.py
```diff
@@ -113,9 +114,11 @@ def broadcast_model(
             task_name (str, optional): name of the task. Defaults to "train".
             data (FLModel, optional): FLModel to be sent to clients. If no data is given, send empty FLModel.
             targets (List[str], optional): the list of target client names or None (all clients). Defaults to None.
+            min_responses (int, optional): the minimum number of responses expected. If == 0, must receive responses from
+              all clients that the task has been sent to. Defaults to 0.
             timeout (int, optional): time to wait for clients to perform task. Defaults to 0, i.e., never time out.
             wait_time_after_min_received (int, optional): time to wait after
-                minimum number of clients responses has been received. Defaults to 10.
+                minimum number of clients responses has been received. Defaults to 0.
```
YuanTingHsieh, line 121: Our timeout when set to 0 means never timeout VS  wait_time_after_min_received set to 0 means don't wait, is there any conflicts here?
This has nothing to do with this PR though.

YuanTingHsieh (dismissed):
PR is good.

nvflare/app_common/workflows/base_model_controller.py
```diff
@@ -113,9 +114,11 @@ def broadcast_model(
             task_name (str, optional): name of the task. Defaults to "train".
             data (FLModel, optional): FLModel to be sent to clients. If no data is given, send empty FLModel.
             targets (List[str], optional): the list of target client names or None (all clients). Defaults to None.
+            min_responses (int, optional): the minimum number of responses expected. If == 0, must receive responses from
```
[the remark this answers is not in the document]
SYangster: This was how the original internal controller uses min_responses https://github.com/NVIDIA/NVFlare/blob/main/nvflare/apis/impl/bcast_manager.py#L69

I can change ModelController to default to None to be more clear, but we should probably keep it internally to still use 0 for backwards compatibility for old controllers. Also min_response=0 wouldn't make sense anyways, as it would automatically end the task.

nvflare/app_common/workflows/base_model_controller.py
SYangster: Good question, there's no conflict here as timeout=0 (never timeout) just means that the task will not end early after 'timeout' seconds regardless if completed.

wait_time_after_min_received means that after min_responses are received, the task will wait x more seconds and then end.

PR #2699, opened 2024-07-15
Add CrossSiteEval with ModelController

SYangster:
Updated CSE workflow for recent ModelController changes.

### Types of changes

- [ ] Documentation updated.


nvflare/app_common/workflows/cross_site_eval.py
```diff
@@ -0,0 +1,187 @@
...
+        if self.persistor:
+            for server_model_name in self._server_models:
+                server_model_path = os.path.join(self.get_app_dir(), server_model_name)
+                server_model_learnable = self.persistor.get_model_from_location(server_model_path, self.fl_ctx)
+                server_model = FLModelUtils.from_model_learnable(server_model_learnable)
+                self._send_validation_task(server_model_name, server_model)
+        else:
+            for server_model_name in self._server_models:
+                try:
+                    server_model = fobs.loadf(server_model_name)
```
YuanTingHsieh, line 107: Do we want to add this default model persistor ?

Or we add it in follow up PRs

nvflare/app_common/utils/fl_model_utils.py
```diff
@@ -19,6 +19,7 @@
 from nvflare.apis.fl_context import FLContext
 from nvflare.apis.shareable import Shareable
 from nvflare.app_common.abstract.fl_model import FLModel, FLModelConst, MetaKey, ParamsType
+from nvflare.app_common.abstract.model import ModelLearnable, ModelLearnableKey, make_model_learnable
```
YuanTingHsieh, line 22: Not sure where should we put these conversion methods though.

nvflare/app_common/workflows/cross_site_eval.py
SYangster: We can add it in follow up PR. One thing to consider is how to specify different save/load locations, as the ModelPersistor doesn't have this parameter for the save/load.

nvflare/app_common/utils/fl_model_utils.py
SYangster: I think fl_model_utils makes sense since we also have the other conversion methods here as well. Do you have any other places in mind?

YuanTingHsieh (approved):
LGTM

PR #2707, opened 2024-07-16
Add ModelController documentation

SYangster:
### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/docs/programming_guide/controllers/model_controller.rst
+++ b/docs/programming_guide/controllers/model_controller.rst
@@ -112,6 +112,16 @@ Below is a comprehensive table overview of the :class:`ModelController<nvflare.a
 Communication
 =============
 
+The ModelController uses a task based communication where tasks are sent to targets, and targets execute the tasks and return results.
+The :ref:`fl_model` is standadized data structure object that can be sent along with each task, and :ref:`fl_model` responses are received for the results.
+
+.. note::
+
+    The :ref:`fl_model` object can be any type of data.
+    For example, in the "train" and "validate" tasks we send the model parameters along with the task so the target clients can train and validate the model.
+    However in many other tasks that do not involve sending the model (e.g. "submit_model"), the :ref:`fl_model` can contain any type of data (e.g. metadata, metrics etc.) or may be not be needed at all.
+
+
 send_model_and_wait
 -------------------
 :func:`send_model_and_wait<nvflare.app_common.workflows.model_controller.ModelController.send_model_and_wait>` is the core communication function which enables users to send tasks to targets, and wait for responses.
@@ -215,7 +225,8 @@ For the JobAPI, define the controller and send it to the server.
   )
   job.to(controller, "server")
 
-The controller can also be configured in ``config_fed_server.json`` in the workflows section.
+The above JobAPI code will automatically generate the server configuration for the controller.
+The controller can also be configured manually in ``config_fed_server.json`` in the workflows section.
 
 Examples
 ========
--- a/docs/programming_guide/controllers/model_controller.rst
+++ b/docs/programming_guide/controllers/model_controller.rst
@@ -1,0 +1,236 @@
+.. _model_controller:
+
+###################
+ModelController API
+###################
+
+The FLARE :mod:`ModelController<nvflare.app_common.workflows.model_controller>` API provides an easy way for users to write and customize FLModel-based controller workflows.
+
+* Highly flexible with a simple API (run routine and basic communication and utility functions)
+* :ref:`fl_model`for the communication data structure, everything else is pure Python
+* Option to support pre-existing components and FLARE-specific functionalities
+
+.. note::
+
+    The ModelController API is a high-level API meant to simplify writing workflows.
+    If users prefer or need the full flexibility of the Controller with all the capabilites of FLARE functions, refer to the :ref:`controllers`.
+
+
+Core Concepts
+=============
+
+As an example, we can take a look at the popular federated learning workflow, "FedAvg" which has the following steps:
+
+#. FL server initializes an initial model
+#. For each round (global iteration):
+
+   #. FL server sends the global model to clients
+   #. Each FL client starts with this global model and trains on their own data
+   #. Each FL client sends back their trained model
+   #. FL server aggregates all the models and produces a new global model
+
+
+To implement this workflow using the ModelController there are a few essential parts:
+
+* Import and subclass the :class:`nvflare.app_common.workflows.model_controller.ModelController`.
+* Implement the ``run()`` routine for the workflow logic.
+* Utilize ``send_model()`` / ``send_model_and_wait()`` for communication to send tasks with FLModel to target clients, and receive FLModel results.
+* Customize workflow using predefined utility functions and components, or implement your own logics.
+
+
+Here is an example of the FedAvg workflow using the :class:`BaseFedAvg<nvflare.app_common.workflows.base_fedavg.BaseFedAvg>` base class:
+
+.. code-block:: python
+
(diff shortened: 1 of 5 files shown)


docs/programming_guide/controllers/model_controller.rst
```diff
@@ -0,0 +1,230 @@
+.. _model_controller:
+
+###################
+ModelController API
```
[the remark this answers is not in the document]
SYangster: added note

docs/programming_guide/controllers/model_controller.rst
```diff
@@ -0,0 +1,236 @@
...
+-------------------
+:func:`send_model_and_wait<nvflare.app_common.workflows.model_controller.ModelController.send_model_and_wait>` is the core communication function which enables users to send tasks to targets, and wait for responses.
+
+The ``data`` is an :ref:`fl_model` object, and the ``task_name`` is the task for the target executors to execute (Client API executors by default support "train", "validate", and "submit_model", however executors can be written for any arbitrary task name).
+
+``targets`` can be chosen from client names obtained with ``sample_clients()``.
+
+Returns the :ref:`fl_model` responses from the target clients once the task is completed (``min_responses`` have been received, or ``timeout`` time has passed).
+
+send_model
```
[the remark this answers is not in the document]
SYangster: Right this is a good point, @YuanTingHsieh and I were also discussing about these options too. 

For (1), rather than introducing a separate FLTask object, maybe we can instead rename `send_model` to either `send` or `send_task` to be more clear that we are sending tasks? (as we are always sending tasks, not just in KM and CSE) 

For (2) I agree it is good to emphasize that the FLModel `data` is specific to the task being sent and may or may not be an actual ML model, but is rather an optional data structure paired with the task.

nvflare/app_common/workflows/base_model_controller.py
```diff
@@ -351,7 +351,7 @@ def save_model(self, model):
             self.error("persistor not configured, model will not be saved")
 
     def sample_clients(self, num_clients=None):
-        clients = self.engine.get_clients()
+        clients = [client.name for client in self.engine.get_clients()]
```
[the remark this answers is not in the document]
SYangster: Returns client names to avoid exposing Client object to users. I added the return type hint to ModelController, will also add to this class

docs/programming_guide/controllers/model_controller.rst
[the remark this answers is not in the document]
YuanTingHsieh: I like `send_task`.

@holgerroth @ZiyueXu77 do you have any preferences?

nvflare/app_common/workflows/base_model_controller.py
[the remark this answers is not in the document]
YuanTingHsieh: if want to be precise we can do:

```
num_clients: Optional[int] = None
```

docs/programming_guide/controllers/model_controller.rst
```diff
@@ -0,0 +1,247 @@
...
+
+Saving & Loading
+================
+
+persistor
+---------
+The :func:`save_model<nvflare.app_common.workflows.model_controller.ModelController.save_model>` and :func:`load_model<nvflare.app_common.workflows.model_controller.ModelController.load_model>`
+functions utilize the configured persistor set in the ModelController ``persistor_id: str = "persistor"`` argument.
+
+For the JobAPI, the persistor with ``id = "persistor"`` will automatically be configured based on the type of the model sent to the server.
```
YuanTingHsieh, line 154: Do we want to explain the JobAPI's saving and loading in ModelController's doc?

docs/programming_guide/controllers/model_controller.rst
```diff
@@ -0,0 +1,247 @@
...
+
+  job.to(Net(), "server")
+
+The persistor can also be configured in ``config_fed_server.json`` in the components section.
+
+custom save & load
+------------------
+Users can also choose to instead create their own save and load functions rather than use a persistor.
+
+For example we can use PyTorch's save and load functions for the model parameters, and save the FLModel metadata with :mod:`FOBS<nvflare.fuel.utils.fobs>` separately:
```
YuanTingHsieh, line 166: This is a bit confusing I think.

So to customize save & load.
Users can do either:
1. write a custom Persistor and then put that in the config file.
2. write a custom ModelController and overwrite the "save_model" "load_model" methods.

Do we really want to introduce all these different options?
I feel it should be better to keep it simple, so maybe provide option 1 is enough?

We should leave these "saving/loading/serialization" explanation to other section.

docs/programming_guide/controllers/model_controller.rst
```diff
@@ -0,0 +1,247 @@
...
+
+For the JobAPI, define the controller and send it to the server.
+
+.. code-block:: python
+
+  controller = FedAvg(
+      num_clients=n_clients,
+      num_rounds=num_rounds,
+  )
+  job.to(controller, "server")
```
YuanTingHsieh, line 226: Not sure if we want this section in "ModelController" though.
This should be a generic thing for all controllers?
So maybe we should put this in some workflow configuration doc that is in "user_guide" or something?

YuanTingHsieh (commented):
mostly good, some comments

docs/programming_guide/controllers/model_controller.rst
SYangster: Good point, I can remove how to configure the persistor here

docs/programming_guide/controllers/model_controller.rst
SYangster: Option 1 is good, however even a custom ModelPersistor is limiting as it is centered around ModelLearnable, and `save_model()`/`load_model()` don't have arguments for save location. I think Option 2 could be useful to showcase the flexibility?

YuanTingHsieh: I see, but when you add additional argument to "save_model" and "load_model" that actually just overwrites them, they become new methods essentially.

Users can just add new methods call "save_model_with_location"?
Then in this case, maybe we don't even include these two methods in the class interface (model_controller.py)?
So remove "save_model" and "load_model" in "ModelController" class.

If we want to emphasis the saving / loading using persistor, maybe we need to move them up from BaseModelController to ModelController

docs/programming_guide/controllers/model_controller.rst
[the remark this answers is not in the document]
ZiyueXu77: Since we need FLModel defined to be "sent", I think send_model makes sense for current pipeline. I would prefer @chesterxgchen's option 2 - explain that the action of send_model is in fact dispatching a task with the "core content" as the model. Otherwise if renaming to "send_task", we may need to further define FLTask(model= , ...) to pair with it.

docs/programming_guide/controllers/model_controller.rst
```diff
@@ -0,0 +1,223 @@
...
+Users can also choose to instead create their own custom save and load functions rather than use a persistor.
+
+For example we can use PyTorch's save and load functions for the model parameters, and save the FLModel metadata with :mod:`FOBS<nvflare.fuel.utils.fobs>` separately to different filepaths.
+
+.. code-block:: python
+
+  import torch
+  from nvflare.fuel.utils import fobs
+
+  def save_model(self, model, filepath=""):
```
YuanTingHsieh, line 165: minor thing though, indentation is different save_model vs load_model.

another thing, do you want to add something like the below to clarify these are in this class?:

```
class ModelController:
    ...
    def save_model(self, xxx

YuanTingHsieh (approved):
Thanks, LGTM

PR #2722, opened 2024-07-22
[DRAFT] Improve metrics streaming

YuanTingHsieh:
### Issue

Some metrics receiver support client side, some don't
If users mis-configure metrics streaming, its error handling should NOT stop the training job.

### Description

- Remove unused file
- Add supporting type of metrics receiver (either on server side only or both sides)
- Enhance error handling 

### Types of changes

- [ ] Documentation updated.


nvflare/app_opt/tracking/wandb/wandb_receiver.py
```diff
@@ -35,9 +34,9 @@ class WandBTask(NamedTuple):
 
 
 class WandBReceiver(AnalyticsReceiver):
-    def __init__(self, kwargs: dict, mode: str = "offline", events=None, process_timeout=10):
-        if events is None:
-            events = ["fed.analytix_log_stats"]
```
nvkevlu, line 40: are we intentionally removing setting events to ["fed.analytix_log_stats"] by default?

nvkevlu (approved):
Thanks for reorganizing and cleaning this up, LGTM

PR #2730, opened 2024-07-25
keep the local resources for simulator

yhwen:
Fixes # .

### Description

keep the local resources for simulator. Users can add any local settings in the "local" folder within the simulator workspace.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/private/fed/app/simulator/simulator_runner.py
+++ b/nvflare/private/fed/app/simulator/simulator_runner.py
@@ -248,7 +248,7 @@ class SimulatorRunner(FLComponent):
             self.args.sp_scheme = parsed_url.scheme
 
             self.logger.info("Deploy the Apps.")
-            self._deploy_apps(job_name, data_bytes, meta)
+            self._deploy_apps(job_name, data_bytes, meta, log_config_file_path)
 
             server_workspace = os.path.join(self.args.workspace, "server")
             workspace = Workspace(root_dir=server_workspace, site_name="server")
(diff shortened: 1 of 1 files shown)


nvflare/private/fed/app/simulator/simulator_runner.py
```diff
@@ -271,18 +271,29 @@ def _cleanup_workspace(self):
...
+                shutil.move(temp_local_dir, local_dir)
 
     def _setup_local_startup(self, log_config_file_path, workspace):
         local_dir = os.path.join(workspace, "local")
         startup = os.path.join(workspace, "startup")
         os.makedirs(local_dir, exist_ok=True)
         shutil.copyfile(log_config_file_path, os.path.join(local_dir, WorkspaceConstants.LOGGING_CONFIG))
+        workspace_local = os.path.join(self.simulator_root, "local")
+        if os.path.exists(workspace_local):
+            shutil.copytree(workspace_local, local_dir, dirs_exist_ok=True)
```
YuanTingHsieh, line 296: So we want the content from "simulator_root"/"local" to overwrite "simulator_root" / p / "local"?
For simulator, are all sites share the same "simulator_root"/"local" folder hence same config in "simulator_root"/"local"?

YuanTingHsieh (commented):

nvflare/private/fed/app/simulator/simulator_runner.py
yhwen: Yes. For simulator there's no fix number, or fix name of the clients. Each run will have its own number of clients.

PR #2732, opened 2024-07-26
Fix cell encrypt

YuanTingHsieh:
### Issue

After 7/20/2024, our CI has been failing with the following error:

```
TypeError: unsupported operand type(s) for +: 'memoryview' and 'bytes'
```

### Cause

The old cryptography package (version 42.0.8) when it does the following code:

```
def _sym_enc(k: bytes, n: bytes, m: bytes) -> bytes:
    cipher = Cipher(algorithms.AES(k), modes.CBC(n), backend=default_backend())
    encryptor = cipher.encryptor()
    padder = padding.PKCS7(PADDING_LENGTH).padder()
    padded_data = padder.update(m) + padder.finalize()
    return encryptor.update(padded_data) + encryptor.finalize()
```

If the "m" is an object of type `memoryview`, the `padder.update(m)` will return an object of `bytes` type.

With the new cryptography package (version 43.0.0) released in 7/20/2024 when it does `padder.update(m)` with a `memoryview` type it will return an object of `memoryview` type.

So if we want to use this same method, we need to ensure the passed in "m" is indeed of type `bytes`.

### Description

- Convert message payload to bytes before feed into the encrypt method
- Remove unused CellCipher class

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/fuel/f3/cellnet/cell_cipher.py
+++ b/nvflare/fuel/f3/cellnet/cell_cipher.py
@@ -89,7 +89,7 @@ def _verify(k, m, s):
     )
 
 
-def _sym_enc(k, n, m):
+def _sym_enc(k: bytes, n: bytes, m: bytes):
     cipher = ciphers.Cipher(ciphers.algorithms.AES(k), ciphers.modes.CBC(n))
     encryptor = cipher.encryptor()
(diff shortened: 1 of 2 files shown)


nvidianz (approved):
Looks good! This doesn't support Buffer List but it's a known issue.

PR #2735, opened 2024-07-26
Moved the hello-pt example initialization to START_RUN

yhwen:
… downloaded dataset to each individual site.

Fixes # .

### Description

Moved the hello-pt example initialization to START_RUN event. Also store the downloaded dataset to separate individual site without conflict.

### Types of changes

- [ ] Documentation updated.


nvkevlu:
Thank you for pointing this out @yhwen, I will take this into account in the updated hello-pt example using the Job API.

examples/hello-world/hello-pt/jobs/hello-pt/app/custom/cifar10validator.py
```diff
@@ -48,6 +53,11 @@ def __init__(self, data_path="~/data", validate_task_name=AppConstants.TASK_VALI
         test_data = CIFAR10(root=data_path, train=False, transform=transforms)
         self._test_loader = DataLoader(test_data, batch_size=4, shuffle=False)
 
+    def handle_event(self, event_type: str, fl_ctx: FLContext):
+        if event_type == EventType.START_RUN:
```
YuanTingHsieh, line 57: why do we move into "START_RUN" event?

examples/hello-world/hello-pt/jobs/hello-pt/app/custom/cifar10validator.py
```diff
@@ -48,6 +53,11 @@ def __init__(self, data_path="~/data", validate_task_name=AppConstants.TASK_VALI
...
+            data_path = os.path.join(self.data_path, fl_ctx.get_identity_name())
```
YuanTingHsieh, line 58: There is good and bad about adding this additional "fl_ctx.get_identity_name()" to it.
Since this data_path is user's passed in, IMO would be better to use it directly instead of adding a sub-folder to it.
Another thing, for the new code, if the data_path is not writable we can not run this code.

On the other hand you could argue that by adding this sub-folder, we remove the burden to ask user to run prepare_data.sh in the hello example.

examples/hello-world/hello-pt/jobs/hello-pt/app/custom/cifar10validator.py
yhwen: Because we want each client to store the download data to its unique location, which will be under the client_name (fl_ctx.get_identity_name()). Otherwise, concurrent clients will overwrite each other in the same location. This client_name is getting from the fl_ctx, only available when START_RUN.

examples/hello-world/hello-pt/jobs/hello-pt/app/custom/cifar10validator.py
yhwen: fl_ctx.get_identity_name() is to use the unique client_name. This is to avoid the concurrent clients overwrite the data in the same place.

YuanTingHsieh (approved):
Holger prefer this user experience

PR #2737, opened 2024-07-29
Update getting_started.rst

hwpang:
### Description

Fixing the path in getting started examples

### Types of changes

- [x] Documentation updated.

### DIFF
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -295,6 +295,7 @@ to a working directory:
 
 .. code-block:: shell
 
+  rm -rf simulator-example
   cp -rf NVFlare/examples/hello-world/hello-pt simulator-example/
 
 The hello-pt application requires a few dependencies to be installed.  As in the installation section,
--- a/docs/getting_started.rst
+++ b/docs/getting_started.rst
@@ -309,13 +309,13 @@ we can install these in the Python virtual environment by running:
 If using the Dockerfile above to run in a container, these dependencies have already been installed.
 
 Next, we can create a workspace for the Simulator to use for outputs of the application run, and launch
-the simulator using ``simulator-example/hello-pt/jobs/hello-pt`` as the input job directory.  In this example, we'll
+the simulator using ``simulator-example/jobs/hello-pt`` as the input job directory.  In this example, we'll
 run on two clients using two threads:
 
 .. code-block:: shell
 
   mkdir simulator-example/workspace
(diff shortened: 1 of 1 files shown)


YuanTingHsieh:
hi @hwpang , thanks for the interest, could you provide the motivation as to why we don't want to "mkdir simulator-example"?

hwpang:
Hi @YuanTingHsieh , with `mkdir simulator-example`, the copied folder has structure `simulator-example/hello-pt/requirements.txt`, while without it, the copied folder has structure `simulator-example/requirements.txt`. I made these changes to make the folder structure under `simulator-example` consistent throughout the getting started example.

docs/getting_started.rst
```diff
@@ -295,7 +295,6 @@ to a working directory:
 
 .. code-block:: shell
 
-  mkdir simulator-example
   cp -rf NVFlare/examples/hello-world/hello-pt simulator-example/
```
YuanTingHsieh, line 298: @hwpang to ensure the behavior is always consistent, we probably need to add a `rm -rf simulator-example` in front of this command to achieve what we want

YuanTingHsieh (commented):
suggest one change

docs/getting_started.rst
hwpang: @YuanTingHsieh That makes sense, thanks! Updated!

YuanTingHsieh (approved):
Thanks, LGTM

@SYangster and @nvkevlu 

other thing I notice we should remove this "python 3.8" requirement text in getting started doc, we can do that in other PR

YuanTingHsieh:
@hwpang 

Can you think of any benefit we change from "mkdir" to "rm -rf"?

hwpang:
@YuanTingHsieh Thanks. I realize that there is a simpler fix than what I implemented before. I have implemented the new fix to add the `hello-pt` in the `python3 -m pip install -r simulator-example/hello-pt/requirements.txt`. This way, we don't need to change any other part of the documentation. I followed through this new instruction locally, and it works for me.

YuanTingHsieh (approved):
Thanks, LGTM

SYangster:
Hi @hwpang, it seems that your most recent commit (https://github.com/NVIDIA/NVFlare/pull/2737/commits/4200f77b7b04741468d0082cfe10e36f35d2d71c) was not signed. Once it is signed, then it'll be ready to merge, thanks!

hwpang:
@SYangster Thanks! I have signed and rebased to main.

PR #2742, opened 2024-07-30
Job API added check for the duplicate executor

yhwen:
Fixes # .

### Description

Added checks for the duplicate executor in the Job API. Generate the unique "persiter_id" and "locator_id" if needed.

### Types of changes

- [ ] Documentation updated.


nvflare/job_config/fed_app_config.py
```diff
@@ -36,6 +36,10 @@ def add_executor(self, tasks: List[str], executor: Executor):
         if not isinstance(executor, Executor):
             raise RuntimeError(f"workflow must be type of Executor, but got {executor.__class__}")
 
+        for item in self.executors:
+            if item.tasks == tasks:
+                raise RuntimeError(f"executor for tasks: {tasks} already exist.")
```
SYangster, line 41: What if two executors share some tasks but not all? Or if they have the same tasks, but listed in different order.

yhwen: enhanced the common task checking for adding the executors.

SYangster:
This PR is no longer needed, we incorporated the fix in the new refactor https://github.com/NVIDIA/NVFlare/pull/2799

PR #2743, opened 2024-07-30
Fix overseer test timing

YuanTingHsieh:
10 second is not enough for the client agent to get the updated SP address.

### Description

Increase this test case to wait 20 seconds before checking.

### Types of changes

- [ ] Documentation updated.


SYangster (approved):
LGTM

PR #2748, opened 2024-07-31
Add job API to support additional external dir in the custom dir

yhwen:
Fixes # .

### Description

To support add contents of external dir into the job "custom" folder for Job API.

### Types of changes

- [ ] Documentation updated.


nvflare/job_config/base_app_config.py
```diff
@@ -62,6 +63,12 @@ def add_ext_script(self, ext_script: str):
 
         self.ext_scripts.append(ext_script)
 
+    def add_ext_dir(self, ext_dir: str):
```
yanchengnv, line 66: Please add docstring explaining what this method does and how it's supposed to be used.

nvflare/job_config/fed_job.py
```diff
@@ -104,6 +104,14 @@ def add_external_scripts(self, external_scripts: List):
         for _script in external_scripts:
             self.app.add_ext_script(_script)
 
+    def add_external_dir(self, ext_dir: str):
```
yanchengnv, line 107: Please name the methods consistently in all places, either add_external_dir or add_ext_dir.

nvflare/job_config/fed_job_config.py
```diff
@@ -174,6 +176,11 @@ def _copy_ext_scripts(self, custom_dir, ext_scripts):
                 module = "".join(relative_script.rsplit(".py", 1)).replace(os.sep, ".")
                 self._copy_source_file(custom_dir, module, script, dest_file)
 
+    def _copy_ext_dirs(self, custom_dir, app_config: BaseAppConfig):
+        for dir in app_config.ext_dirs:
+            dest = os.path.join(custom_dir, os.path.basename(dir))
```
yanchengnv, line 181: Not sure this would work for Flower's use case. Here a subdir is created within the custom_dir, but in the Flower case, the content of the dir should be copied to the custom_dir directly - no subdir. The custom_dir is where their toml file should be stored.

On the other hand, I can see that the user might call add_ext_dir multiple times.

Maybe we should add a flag in add_ext_dir to indicate whether the content of the dir should be placed into the custom_dir directly.

yanchengnv (changes requested):
Please add docstrings.
The behavior of copy_ext_dirs may need to be revisited. See my comments.

nvflare/job_config/fed_job_config.py
SYangster: Also related, `_copy_ext_scripts()` can place the script in a subdir as well depending on the script path. Perhaps we would want to add the same flag here for placing it at the custom_dir instead? (this would require modifying obj._task_script_path too for ScriptExecutors)

nvflare/job_config/base_app_config.py
yhwen: This is the internal implementation place. The docstring is added to the add_external_dir() infed_job.py

nvflare/job_config/fed_job.py
yhwen: add_ext_dir() is the internal implementation. add_external_dir() and add_external_scripts() are consistent in the fed_job.py

nvflare/job_config/fed_job_config.py
yhwen: Discussed with @yanchengnv, changed the behavior to copying flat to job "custom" folder.

SYangster (approved):

PR #2758, opened 2024-08-05
Removed the app_opt scan

yhwen:
Fixes # .

### Description

Removed the app_opt scan, to solve the slow server / client and simulator startup issue.

### Types of changes

- [ ] Documentation updated.


nvflare/private/fed/app/fl_conf.py
```diff
@@ -35,7 +35,7 @@
 from .fl_app_validator import FLAppValidator
 
 FL_PACKAGES = ["nvflare"]
-FL_MODULES = ["server", "client", "app_common", "private", "app_opt"]
```
YuanTingHsieh, line 38: we need to check all our examples / tutorials / notebooks to make sure we are not using "name" to refer to components in "app_opt"

@SYangster @nvkevlu @holgerroth

YuanTingHsieh (approved):

PR #2763, opened 2024-08-06
Add support of just doing metrics streaming with client api

YuanTingHsieh:
Add support when people just want to use metrics streaming (i.e. Metric Pipe) of client api.

### Description

- Make "task_pipe" non-required
- Make TASK_EXCHANGE section optional for client api config

### Types of changes

- [ ] Documentation updated.


nvflare/client/api.py
```diff
@@ -45,10 +46,11 @@ def init(rank: Optional[str] = None):
...
-        client_api = data_bus.get_data(CLIENT_API_KEY)
-    else:
-        client_api = ExProcessClientAPI()
+    if client_api is None:
+        if api_type == ClientAPIType.IN_PROCESS_API:
+            client_api = data_bus.get_data(CLIENT_API_KEY)
+        else:
+            client_api = ExProcessClientAPI()
 
     client_api.init(rank=rank)
```
yanchengnv, line 55: Shouldn't you move client_api.init() to the inside of the "if client_api is None", so it will be called only once?

nvflare/client/ex_process/api.py
```diff
@@ -81,20 +82,23 @@ def init(self, rank: Optional[str] = None):
             rank = os.environ.get("RANK", "0")
 
         if self.process_model_registry:
-            print("Warning: called init() more than once. The subsequence calls are ignored")
+            logging.warn("Warning: called init() more than once. The subsequence calls are ignored")
```
yanchengnv, line 85: Change to use self.logger.warning(). 

from nvflare.fuel.utils.obj_utils import get_logger
self.logger = get_logger(self)

nvflare/client/ex_process/api.py
```diff
@@ -106,12 +110,13 @@ def init(self, rank: Optional[str] = None):
...
                     metric_pipe=metric_pipe,
                     metric_channel_name=metric_channel_name,
+                    heartbeat_timeout=client_config.config.get(ConfigKey.HEARTBEAT_TIMEOUT, 60),
                 )
                 flare_agent.start()
 
             self.process_model_registry = ModelRegistry(client_config, rank, flare_agent)
         except Exception as e:
-            print(f"flare.init failed: {e}")
+            logging.error(f"flare.init failed: {e}")
```
yanchengnv, line 119: Change to use self.logger.

yanchengnv (changes requested):
See my comments.

nvflare/client/api.py
YuanTingHsieh: thanks, changed as suggested

nvflare/client/ex_process/api.py
YuanTingHsieh: thanks, changed as suggested

nvflare/client/ex_process/api.py
YuanTingHsieh: thanks, changed as suggested

PR #2764, opened 2024-08-06
Add flower metrics streaming example

YuanTingHsieh:
### Description

- Add flower metrics streaming example

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh:
require changes in #2763

examples/hello-world/hello-flower/README.md
```diff
@@ -14,15 +14,24 @@ $ tree jobs/hello-flwr-pt
...
 nvflare simulator jobs/hello-flwr-pt -n 2 -t 2 -w /tmp/nvflare/flwr
 ```
+
+## 2.2 Run a simulation with TensorBoard streaming
+
+Next, we run 2 Flower clients and Flower Server in parallel using NVFlare while streaming 
+the TensorBoard metrics to the server at each iteration using NVFlare's metric streaming.
+
+```bash
+CLIENT_API_TYPE="EX_PROCESS_API" nvflare simulator jobs/hello-flwr-pt_tb_streaming -n 2 -t 2 -w /tmp/nvflare/flwr_tb_streaming
```
holgerroth, line 36: Why do we need to set `CLIENT_API_TYPE` here? Couldn't there be a default option that is used if the environment variable is not specified?

holgerroth: If we keep the env variable here, we probably should add some explanation as to why this is needed. Is there documentation? cc @nvkevlu

YuanTingHsieh: Since flower process is launched outside, the default IN_PROCESS_API will not work

YuanTingHsieh:
let me resolve the conflicts

PR #2765, opened 2024-08-06
Enhance cyclic controller

YuanTingHsieh:
### Description

Port changes from #2337 and enhance the logic of cyclic controller to check RC first before trying to call shareable_generator.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/app_common/workflows/cyclic_ctl.py
+++ b/nvflare/app_common/workflows/cyclic_ctl.py
@@ -170,6 +170,19 @@ class CyclicController(Controller):
         self._is_done = True
 
     def _process_result(self, client_task: ClientTask, fl_ctx: FLContext):
+        result = client_task.result
+        rc = result.get_return_code()
+        client_name = client_task.client.name
+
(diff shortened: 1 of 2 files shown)


nvflare/app_common/workflows/cyclic_ctl.py
```diff
@@ -176,39 +176,47 @@ def _process_result(self, client_task: ClientTask, fl_ctx: FLContext):
         task = client_task.task
 
         result = client_task.result
+        client_name = client_task.client.name
         if isinstance(result, Shareable):
             # update the global learnable with the received result (shareable)
             # e.g. the received result could be weight_diffs, the learnable could be full weights.
             rc = result.get_return_code()
-            try:
```
yanchengnv, line 183: No, we should not change the order, because the last message may be early termination, but also contain the last training result, which should be converted to Learnable!

yanchengnv (changes requested):
The change is not totally correct, because even though the early termination, the message may still contain the last model data.

PR #2775, opened 2024-08-08
Fix JobAPI Persistor handling

SYangster:
[2153](https://jirasw.nvidia.com/browse/FLARE-2153) Fix for when both torch and tensorflow are installed.

### Types of changes

- [ ] Documentation updated.


nvflare/job_config/fed_job.py
```diff
@@ -260,19 +260,17 @@ def to(
...
-
-                        component = PTFileModelLocator(pt_persistor_id="persistor")
-                        self._deploy_map[target].app.add_component("model_locator", component)
-                        added_model = True
-                elif tf_ok:
-                    if isinstance(obj, tf.keras.Model):  # if model, create a TF persistor
-                        component = TFModelPersistor(model=obj)
-                        self._deploy_map[target].app.add_component("persistor", component)
-                        added_model = True
+                if torch_ok and isinstance(obj, nn.Module):  # if model, create a PT persistor
```
yanchengnv, line 263: What is this change for? The old and new code seem to be identical, other than different format.

yanchengnv (changes requested):
Don't see the meaning of the changes.
Besides, are you sure this is the root cause of the QA reported issue?

nvflare/job_config/fed_job.py
SYangster: In the old version, if torch is installed (`torch_ok=True`) but the model is a `tf.keras.Model`, it will not reach the second case, and won't add the TFModelPersistor.

yanchengnv: I see.

yanchengnv (approved):
LGTM.

PR #2779, opened 2024-08-08
Avoid user-defined device from being overwritten

YuanTingHsieh:
### Description

Avoid user-defined device from being overwritten

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/app_opt/xgboost/histogram_based/executor.py
+++ b/nvflare/app_opt/xgboost/histogram_based/executor.py
@@ -243,9 +243,10 @@ class FedXGBHistogramExecutor(Executor):
         self.world_size = world_size
 
         if self.use_gpus:
-            # mapping each rank to a GPU (can set to cuda:0 if simulating with only one gpu)
-            self.log_info(fl_ctx, f"Training with GPU {self.rank}")
-            self.xgb_params["device"] = f"cuda:{self.rank}"
+            # mapping each rank to a GPU if not set (can set to cuda:0 if simulating with only one gpu)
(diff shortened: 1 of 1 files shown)


nvflare/app_opt/xgboost/histogram_based/executor.py
```diff
@@ -243,9 +243,10 @@ def train(self, shareable: Shareable, fl_ctx: FLContext, abort_signal: Signal) -
         self.world_size = world_size
 
         if self.use_gpus:
-            # mapping each rank to a GPU (can set to cuda:0 if simulating with only one gpu)
-            self.log_info(fl_ctx, f"Training with GPU {self.rank}")
-            self.xgb_params["device"] = f"cuda:{self.rank}"
+            # mapping each rank to a GPU if not set (can set to cuda:0 if simulating with only one gpu)
+            device = self.xgb_params.get("device", f"cuda:{self.rank}")
```
yanchengnv, line 247: I'm not sure we should assume the default device id to be client rank. What if there are many clients?
Perhaps it's safer to assume device 0.

yanchengnv (commented):
Not sure it's a good idea to use client rank as default gpu device id.

nvflare/app_opt/xgboost/histogram_based/executor.py
YuanTingHsieh: good suggestion, make the suggested changes in https://github.com/NVIDIA/NVFlare/pull/2787
closing this pr

PR #2781, opened 2024-08-09
name to path changes in configuration

yhwen:
### Description
name to path changes due to app_opt scan changes
Update the config changes because we won't scan the app_opt package any more.

### Types of changes

- [ ] Documentation updated.


yanchengnv (approved):
This PR looks OK. But we are still scanning too much. Need another PR to address it.

YuanTingHsieh (approved):
LGTM @holgerroth @ZiyueXu77 please check that we have this changed because of performance issues, we might need to modify other places like MONAI toolkit as well

YuanTingHsieh:
I will update xgboost related examples

PR #2789, opened 2024-08-13
Improve the kill children processes

yhwen:
Fixes # .

### Description

Change to use process.kill() to kill the children processes. This force kill the example like flow child process.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/private/fed/app/utils.py
+++ b/nvflare/private/fed/app/utils.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 import os
-import signal
 import sys
 import threading
 import time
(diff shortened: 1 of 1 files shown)


nvflare/private/fed/app/utils.py
```diff
@@ -56,7 +56,7 @@ def kill_child_processes(parent_pid, sig=signal.SIGTERM):
         return
     children = parent.children(recursive=True)
     for process in children:
-        process.send_signal(sig)
```
YuanTingHsieh, line 59: Thanks for the fix, should we remove this "sig" argument since it is not used?

yhwen: removed.

YuanTingHsieh:
@yhwen can you fix the unit test, one import is not needed anymore, thanks

YuanTingHsieh (approved):
LGTM

PR #2790, opened 2024-08-13
Remove the module class scan 

yhwen:
Fixes # .

### Description

Remove the module class scan. Only include the necessary limit number of classes needed.

### Types of changes

- [ ] Documentation updated.


nvflare/utils/components_utils.py
```diff
@@ -0,0 +1,64 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+def create_classes_table_static():
+    from nvflare.app_common.abstract.aggregator import Aggregator
+    from nvflare.app_common.abstract.shareable_generator import ShareableGenerator
```
YuanTingHsieh, line 18: Can these abstract classes be used directly?

nvflare/fuel/utils/class_utils.py
```diff
@@ -85,10 +86,10 @@ def __init__(self, base_pkgs: List[str], module_names: List[str], exclude_libs=T
         self.exclude_libs = exclude_libs
 
         self._logger = logging.getLogger(self.__class__.__name__)
-        self._class_table: Dict[str, str] = {}
-        self._create_classes_table()
+        self._class_table = create_classes_table_static()
 
     def _create_classes_table(self):
```
YuanTingHsieh, line 91: this method is not used anymore should we remove it?

nvflare/utils/components_utils.py
```diff
@@ -0,0 +1,64 @@
...
+        DXOAggregator,
+        GlobalModelEval,
+        GlobalWeightsInitializer,
+        Gatherer,
+        ShareableGenerator,
+        SwarmClientController,
+        SwarmServerController,
+        FullModelShareableGenerator,
+        InTimeAccumulateWeightedAggregator,
+    }
```
YuanTingHsieh, line 59: are these the list of names where user can use "name" to refer to these components instead of the full path?

nvflare/utils/components_utils.py
yhwen: removed these and add the popular PTFileModelPersistor and PTFileModelLocator in the list.

nvflare/fuel/utils/class_utils.py
yhwen: Keep this implementation code for referencing purpose.

nvflare/utils/components_utils.py
yhwen: This is for the purpose of keeping some legacy examples and users application without broken.

nvflare/utils/components_utils.py
```diff
@@ -0,0 +1,63 @@
...
+        GlobalModelEval,
+        GlobalWeightsInitializer,
+        Gatherer,
+        SwarmClientController,
+        SwarmServerController,
+        FullModelShareableGenerator,
+        InTimeAccumulateWeightedAggregator,
+        PTFileModelPersistor,
+        PTFileModelLocator,
+    }
```
YuanTingHsieh, line 58: @holgerroth @ZiyueXu77 can you check if this list is good for you?
thanks

ZiyueXu77: LGTM

YuanTingHsieh (approved):
LGTM

PR #2801, opened 2024-08-15
Update autofedrl example

ZiyueXu77:
Fixes JIRA issue.

### Description

Update the research item so that it can run correctly with latest updates

### Types of changes

- [ ] Documentation updated.


research/auto-fed-rl/src/autofedrl/autofedrl_cifar10_learner.py
```diff
@@ -19,7 +19,6 @@
 import torch
 import torch.nn.functional as F
 import torch.optim as optim
-from pt.learners.cifar10_learner import CIFAR10Learner
```
ZiyueXu77, line 22: This is marked "deprecated" here, so I just made a copy into the local folder and removed the deprecated line

research/auto-fed-rl/src/autofedrl/cifar10_learner.py
```diff
@@ -0,0 +1,444 @@
...
+
+from nvflare.apis.dxo import DXO, DataKind, MetaKey, from_shareable
+from nvflare.apis.fl_constant import FLContextKey, ReturnCode
+from nvflare.apis.fl_context import FLContext
+from nvflare.apis.shareable import ReservedHeaderKey, Shareable, make_reply
+from nvflare.apis.signal import Signal
+from nvflare.app_common.abstract.learner_spec import Learner
+from nvflare.app_common.app_constant import AppConstants, ModelName, ValidateType
+from nvflare.app_opt.pt.fedproxloss import PTFedProxLoss
+from nvflare.fuel.utils.deprecated import deprecated
```
holgerroth, line 35: import is not used.

research/auto-fed-rl/src/autofedrl/autofedrl_cifar10_learner.py
holgerroth: You can still use the original file rather than copying it here.

ZiyueXu77: how to still import the deprecated? If I use the original file, it will error out saying class should be code rather than str

research/auto-fed-rl/src/autofedrl/cifar10_learner.py
ZiyueXu77: removed

research/auto-fed-rl/src/autofedrl/autofedrl_cifar10_learner.py
ZiyueXu77: cannot have the deprecated decorator for inheritance

holgerroth: Oh. I thought the deprecation decorator gives only a warning. Using it with class inheritance seems a limitation we cannot easily fix. So, copying the code here should be fine.

holgerroth (approved):
Looks good!

PR #2808, opened 2024-08-19
XGBoost user interface change and XGBoost version check

nvidianz:
Fixes #FLARE-2155

### Description

1. To be consistent with XGBoost and other parts of the NVFlare, split_mode is changed into data_split_mode.
2. Added XGBoost version check. For secure training, an error will be thrown if the XGBoost version is older than 2.2.0-dev

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh (approved):
LGTM

PR #2815, opened 2024-08-20
[2.4] Ported newest ReliableMessage to 2.4

nvidianz:
### Description

1. Ported newest ReliableMessage to 2.4 with locking fix.
2. Changed code to work with XGBoost current release version 2.1.1.

### Types of changes

- [ ] Documentation updated.


yanchengnv (approved):
LGTM

nvflare/app_opt/xgboost/histogram_based_v2/proto/federated.proto
```diff
@@ -1,26 +1,30 @@
 /*!
- * Copyright 2022 XGBoost contributors
- * needs to match file in https://github.com/dmlc/xgboost/blob/v2.0.3/plugin/federated/federated.proto
```
YuanTingHsieh, line 3: do we want to add this comment here for reference,

something like:

* needs to match file in https://github.com/dmlc/xgboost/blob/v2.11/plugin/federated/federated.proto ?

YuanTingHsieh (approved):

nvflare/app_opt/xgboost/histogram_based_v2/proto/federated.proto
nvidianz: This file is copied from XGBoost. Shouldn't make manual changes.

PR #2819, opened 2024-08-21
Revert "Remove the module class scan "

yhwen:
Reverts NVIDIA/NVFlare#2790

There are still too many places in the examples directly use "name" from the "app_common" module. The static class table is not a complete list. Need to roll back to avoid breaking those examples.


yanchengnv:
Should we change examples to use path instead of changing code? I should go with whatever is easier.

YuanTingHsieh (approved):
Agree , it's going to break users' old examples

YuanTingHsieh:
@yanchengnv Yuhong and me have gone through all our examples/integration/tests/research and change all "name" to "path".

It's a matter of if we want to support NOT breaking other people's examples developed outside NVFlare.

Or we can add the documentation and ask users to change to "path"

yhwen:
> Should we change examples to use path instead of changing code? I should go with whatever is easier.

We have too many examples use these.

PR #2820, opened 2024-08-21
ScriptExecutor improvements

SYangster:
- add `launch_external_process` flag to ScriptExecutor, configure InProcessClientAPIExecutor or ClientAPILauncherExecutor accordingly in `add_to_fed_job`
- check if any CellPipe args are SystemVarNames before building cell
- use lazy imports to with `framework` types to avoid unnecessary imports

### Types of changes

- [ ] Documentation updated.


SYangster:
Any suggestions for how to name ScriptExecutor / where to place it now?

nvflare/fuel/utils/pipe/cell_pipe.py
```diff
@@ -178,18 +179,22 @@ def __init__(
...
 
+        self.site_name = site_name
+        self.token = token
+        self.root_url = root_url
+        self.secure_mode = secure_mode
+        self.workspace_dir = workspace_dir
+
+        sysvarname_placeholders = ["{" + varname + "}" for varname in dir(SystemVarName)]
+        if any([arg in sysvarname_placeholders for arg in [site_name, token, root_url, secure_mode, workspace_dir]]):
+            return
```
YuanTingHsieh, line 190: what is this checking for?

Are we assuming if we see "{SITE_NAME}" or any system var then we are in JOB API mode so don't build the cell?
This seems a bit hacky to me though

nvflare/app_opt/script_executor.py
```diff
@@ -0,0 +1,181 @@
...
+                token="{JOB_ID}",
+                root_url="{ROOT_URL}",
+                secure_mode="{SECURE_MODE}",
+                workspace_dir="{WORKSPACE}",
+            )
+            pipe_id = job.add_component("pipe", component, ctx)
+
+            component = SubprocessLauncher(
+                script=self._command + " custom/" + os.path.basename(self._script) + " " + self._script_args,
+            )
```
YuanTingHsieh, line 137: We need "add_resources" for this script?
Only if we do that, then this "custom" needs to be added.

it should be the full script name without any directories.
not sure if we should do `os.path.basename`

People can also put absolute path (/tmp/hello/cool) here, then in that case this "custom" is not fitted into it.
Also we could argue in the case of ScriptExecutor we only support relative path.

YuanTingHsieh:
name: ExecutorSelector?
location: not sure, somewhere inside app_opt? since it has all these optional import

nvflare/fuel/utils/pipe/cell_pipe.py
SYangster: Yes this is needed to prevent the Cell from being built in the JobAPI with the SystemVarNames as arguments.

YuanTingHsieh: got it, we can think of enhancements later.

nvflare/app_opt/script_executor.py
SYangster: `self._script` is added to resources in both cases at the end. `custom` doesn't need to be added, as adding the resources is copying to the custom folder. 

Full script name is supported, as when we copy to the custom folder we take the basename.

YuanTingHsieh: got it, for script/file this behavior is clear. thanks!

nvflare/fuel/utils/pipe/cell_pipe.py
```diff
@@ -18,6 +18,7 @@
 import time
 from typing import Tuple, Union
 
+from nvflare.apis.fl_constant import SystemVarName
```
yanchengnv, line 21: Why is this file modified? It doesn't seem to be related to ScriptExecutor.

nvflare/job_config/base_app_config.py
```diff
@@ -58,9 +58,6 @@ def add_ext_script(self, ext_script: str):
         if not (os.path.isabs(ext_script) or os.path.exists(ext_script)):
             raise RuntimeError(f"Could not locate external script: {ext_script}")
 
-        if not ext_script.endswith(".py"):
```
yanchengnv, line 61: This change has nothing to do with ScriptExecutor. Please use a different PR.

nvflare/job_config/fed_job_config.py
```diff
@@ -172,7 +172,7 @@ def _copy_ext_scripts(self, custom_dir, ext_scripts):
                     relative_script = self._get_relative_script(script)
                 else:
                     relative_script = script
-                dest_file = os.path.join(custom_dir, relative_script)
+                dest_file = os.path.join(custom_dir, os.path.basename(relative_script))
```
yanchengnv, line 175: This change has nothing to do with ScriptExecutor. Please use a different PR.

nvflare/app_opt/pt/params_converter.py
```diff
@@ -12,15 +12,21 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-from typing import Dict
+from typing import Dict, List
```
yanchengnv, line 15: This change has nothing to do with ScriptExecutor. Please use a different PR.

yanchengnv (changes requested):
Some changes have nothing to do with ScriptExecutor. Please move them to different PR.

nvflare/fuel/utils/pipe/cell_pipe.py
SYangster: this is needed for the launch_external_process option of ScriptExecutor

nvflare/app_opt/pt/params_converter.py
SYangster: Since ScriptExecutor is no longer an Executor itself and cannot register the TensorDecomposer, we need this for the pytorch version of ScriptExecutor to work

yanchengnv:
I don't understand the motivation of this PR. Why is ScriptExecutor not an Executor any more? This doesn't seem right.

nvflare/job_config/fed_job_config.py
SYangster: This is needed for better experience with script and task_script_path args of SubprocessLauncher and InProcessClientAPIExecutor

nvflare/job_config/base_app_config.py
SYangster: This is minor, I can move to another PR, but this is to support adding more than just .py files to the custom dir (e.g. .toml files in flower's case)

PR #2833, opened 2024-08-23
improve the class_utils to handle the duplicate class name case

yhwen:
Fixes # .

### Description

Improve the class_utils to handle the duplicate class name case. Provide clear message to show the multiple module names with the same class name.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/fuel/utils/class_utils.py
+++ b/nvflare/fuel/utils/class_utils.py
@@ -18,6 +18,8 @@ import logging
 import pkgutil
 from typing import Dict, List, Optional
 
+from nvflare.apis.fl_component import FLComponent
+from nvflare.fuel.common.excepts import ConfigError
 from nvflare.fuel.utils.components_utils import create_classes_table_static
 from nvflare.security.logging import secure_format_exception
 
@@ -61,17 +63,6 @@ def instantiate_class(class_path, init_params):
     return instance
 
 
-class _ModuleScanResult:
-    """Data class for ModuleScanner."""
-
-    def __init__(self, class_name: str, module_name: str):
-        self.class_name = class_name
-        self.module_name = module_name
-
-    def __str__(self):
-        return f"{self.class_name}:{self.module_name}"
-
-
 class ModuleScanner:
     def __init__(self, base_pkgs: List[str], module_names: List[str], exclude_libs=True):
         """Loads specified modules from base packages and then constructs a class to module name mapping.
@@ -90,7 +81,6 @@ class ModuleScanner:
 
     def create_classes_table(self):
         class_table: Dict[str, str] = {}
-        scan_result_table = {}
         for base in self.base_pkgs:
             package = importlib.import_module(base)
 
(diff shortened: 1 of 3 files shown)


nvflare/fuel/utils/component_classes.json
```diff
@@ -1 +1,572 @@
...
+    "PTFedAvgEarlyStopping": [
+        "nvflare.app_opt.pt.fedavg_early_stopping"
+    ],
+    "PTFedOptModelShareableGenerator": [
+        "nvflare.app_opt.pt.fedopt"
+    ],
+    "FedOpt": [
+        "nvflare.app_opt.pt.fedopt_ctl",
+        "nvflare.app_opt.tf.fedopt_ctl"
+    ],
```
YuanTingHsieh, line 450: How do users determine between PT vs TF?

nvflare/fuel/utils/component_classes.json
```diff
@@ -1 +1,572 @@
...
+    "WandBReceiver": [
+        "nvflare.app_opt.tracking.wandb.wandb_receiver"
+    ],
+    "WandBWriter": [
+        "nvflare.app_opt.tracking.wandb.wandb_writer"
+    ],
+    "XGBFedController": [
+        "nvflare.app_opt.xgboost.histogram_based.controller",
+        "nvflare.app_opt.xgboost.histogram_based_v2.fed_controller"
+    ],
```
YuanTingHsieh, line 499: How do users determine between V1 vs V2?

Should we put v2 in front of v1?

nvflare/fuel/utils/component_classes.json
yhwen: for the same class name, users will get an error message when using "name". They have to change to use "path".

nvflare/fuel/utils/component_classes.json
yhwen: no need for versioning. The component_classes.json is checked in with the same version of codes.

nvflare/fuel/utils/class_utils.py
```diff
@@ -108,18 +98,12 @@ def create_classes_table(self):
                                         not name.startswith("_")
                                         and inspect.isclass(obj)
                                         and obj.__module__ == module_name
+                                        and issubclass(obj, FLComponent)
```
yanchengnv, line 101: Should we do this check for FLComponent? Any Python object could be added through "name". Maybe we should ...

nvflare/fuel/utils/class_utils.py
```diff
@@ -108,18 +98,12 @@ def create_classes_table(self):
...
                                     ):
-                                        # same class name exists in multiple modules
-                                        if name in scan_result_table:
-                                            scan_result = scan_result_table[name]
-                                            if name in class_table:
-                                                class_table.pop(name)
-                                                class_table[f"{scan_result.module_name}.{name}"] = module_name
-                                            class_table[f"{module_name}.{name}"] = module_name
+                                        if name in class_table:
+                                            class_table[name].append(module_name)
```
yanchengnv, line 104: Is it possible that the module_name already exists in the list?

nvflare/fuel/utils/class_utils.py
```diff
@@ -137,7 +121,14 @@ def get_module_name(self, class_name) -> Optional[str]:
         Returns:
             The module name if found.
         """
-        return self._class_table.get(class_name, None)
+        if class_name not in self._class_table:
+            return None
+
+        modules = self._class_table.get(class_name, None)
+        if modules and len(modules) > 1:
+            raise ConfigError(f"There are multiple modules with the class_name:{class_name}, modules are: {modules}")
```
yanchengnv, line 129: Improve the text to:
f"Multiple modules have the class '{class_name}': {modules}. Please specify the class using its full path."

nvflare/fuel/utils/class_utils.py
```diff
@@ -137,7 +121,14 @@ def get_module_name(self, class_name) -> Optional[str]:
         Returns:
             The module name if found.
         """
-        return self._class_table.get(class_name, None)
+        if class_name not in self._class_table:
+            return None
```
yanchengnv, line 125: Change to raise ConfigError with text:
f"Cannot find class '{class_name}'. Please check its spelling. If the spelling is correct, specify the class using its full path."

yanchengnv (changes requested):
Looks good. Please improve the exception text.

nvflare/fuel/utils/class_utils.py
yhwen: The same module_name should not be existed in the list. This is the same package and same file name. Should not happen.

nvflare/fuel/utils/class_utils.py
yhwen: updated.

nvflare/fuel/utils/class_utils.py
yhwen: We already have the proper error message when calling this when return None. But we can change to raise this ConfigError here as well.

PR #2853, opened 2024-08-27
add script_dir to ScriptRunner

yanchengnv:
Fixes # .

### Description

This PR allows a "script_dir" to be specified with ScriptRunner. If specified, the content of it will be added to the job's "custom" folder.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/job_config/script_runner.py
+++ b/nvflare/job_config/script_runner.py
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+import os.path
 from typing import Type
 
 from nvflare.app_common.executors.client_api_launcher_executor import ClientAPILauncherExecutor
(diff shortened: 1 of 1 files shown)


nvflare/job_config/script_runner.py
```diff
@@ -35,6 +35,7 @@ def __init__(
         launch_external_process: bool = False,
         command: str = "python3",
         framework: FrameworkType = FrameworkType.PYTORCH,
+        script_dir: str = None,
```
yhwen, line 38: Instead of adding a "script_dir" here in the ScriptRunner, I think it's better to use the:

     job.to_clients(script_dir)

And this way is more consistent how it is used for the server side.

yhwen: But the script_dir / content_dir is not only used for the ScriptRunner. It's a place for all the "contents" needed for the Application. Use the job.to_clients() once can included all the contents needed. The "script_dir" supposed only related to the script itself.

yhwen: My point is once we have the  job.to_clients(script_dir), we don't need the script_dir in the ScriptRunner any more.

yhwen: I am not sure about this. The  job.to_clients(script_dir) function already exist. And it covers all the things needed here for the script_dir. I don't feel necessary for the change any more, from the whole Job perspective.

yanchengnv:
Do not need this change!

PR #2858, opened 2024-08-27
Added check for duplicate RM request

nvidianz:
### Description
Added a check for duplicate ReliableMessage request so no duplicate message was received.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/apis/utils/reliable_message.py
+++ b/nvflare/apis/utils/reliable_message.py
@@ -341,8 +341,8 @@ class ReliableMessage:
 
         """
         with cls._tx_lock:
-            cls._req_receivers.pop(receiver.tx_id, None)
             cls._register_completed_req(receiver.tx_id, receiver.tx_timeout)
+            cls._req_receivers.pop(receiver.tx_id, None)
             cls.debug(fl_ctx, f"released request receiver of TX {receiver.tx_id}")
 
     @classmethod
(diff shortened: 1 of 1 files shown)


nvflare/apis/utils/reliable_message.py
```diff
@@ -337,6 +342,7 @@ def release_request_receiver(cls, receiver: _RequestReceiver, fl_ctx: FLContext)
         """
         with cls._tx_lock:
             cls._req_receivers.pop(receiver.tx_id, None)
+            cls._register_completed_req(receiver.tx_id, receiver.tx_timeout)
```
yanchengnv, line 345: We should do this before popping it from cls._req_receivers, in case another request is received right after the pop and before added to completed_reqs.

nvflare/apis/utils/reliable_message.py
```diff
@@ -580,6 +586,10 @@ def _send_request(
                 # the ack is a status report - check status
                 status = ack.get_header(HEADER_STATUS)
                 if status and status != STATUS_NOT_RECEIVED:
+                    if status == STATUS_DUP_REQUEST:
+                        cls.debug(fl_ctx, "Duplicate req status received")
+                        return ack
```
yanchengnv, line 591: No, should be "break" not "return ack" so that it will go to the cls._query_result().

yanchengnv (changes requested):
See my comments.

nvflare/apis/utils/reliable_message.py
nvidianz: In that case, this block of code is not needed. Removed.

nvflare/apis/utils/reliable_message.py
nvidianz: Good catch. Reversed.

yanchengnv (approved):
LGTM

PR #2859, opened 2024-08-27
Add FLModel parameter checks

SYangster:
- Add check for FLModel metrics to be a dict to avoid unclear error in [4778901](https://nvbugswb.nvidia.com/NVBugs5/redir.aspx?url=/4778901)
- Decorator was using float for metrics which was inconsistent with FLModel usage, however team decided decorator pattern will be deprecated

An alternative option is to instead check that the result.metrics are dicts before using WeightedAggregationHelper()

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh (approved):
LGTM

PR #2867, opened 2024-08-28
improve race condition handling

yanchengnv:
Fixes # .

### Description

This PR further improves the race condition handling in RM, especially in handling requests after reply was sent.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/apis/utils/reliable_message.py
+++ b/nvflare/apis/utils/reliable_message.py
@@ -295,15 +295,35 @@ class ReliableMessage:
                 # no handler registered for this topic!
                 cls.error(fl_ctx, f"no handler registered for request {rm_topic=}")
                 return make_reply(ReturnCode.TOPIC_UNKNOWN)
-            if cls._req_completed.get(tx_id):
-                cls.debug(fl_ctx, "Completed tx_id received")
-                return _status_reply(STATUS_DUP_REQUEST)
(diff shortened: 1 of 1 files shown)


nvidianz (approved):
Tested and no issues.

nvflare/apis/utils/reliable_message.py
```diff
@@ -295,15 +295,35 @@ def _receive_request(cls, topic: str, request: Shareable, fl_ctx: FLContext):
...
+                # no standing process for this request
+                # further check whether this request was already completed
+                if cls._req_completed.get(tx_id):
+                    # this request was already completed!
+                    cls.debug(fl_ctx, "Completed tx_id received")
+                    return _status_reply(STATUS_DUP_REQUEST)
+
+            if not receiver:
+                # this is a valid new request
+                receiver = cls._get_or_create_receiver(rm_topic, request, handler_f)
```
YuanTingHsieh, line 315: this part can be put inside previous block since they are under same condition "if not receiver"

YuanTingHsieh (commented):
LGTM, just one minor comment

nvflare/apis/utils/reliable_message.py
yanchengnv: I kept them separate because they are for different purposes.

PR #2870, opened 2024-08-28
support passing custom env vars for flower client

yanchengnv:
Fixes # .

### Description

This PR adds the support of passing custom env variables to Flower client. This is necessary to enable Flower client with Flare's client api to check the env var for config processing.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/app_opt/flower/flower_job.py
+++ b/nvflare/app_opt/flower/flower_job.py
@@ -47,7 +47,7 @@ class FlowerJob(FedJob):
         client_shutdown_timeout=5.0,
         stream_metrics=False,
         analytics_receiver=None,
-        client_api_type: str = "EX_PROCESS_API",
+        extra_env: dict = None,
     ):
         """
         Flower Job.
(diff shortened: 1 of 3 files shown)


nvflare/app_opt/flower/flower_job.py
```diff
@@ -69,17 +69,13 @@ def __init__(
...
+            extra_env (dict, optional): optional extra env variables to be passed to Flower client
         """
         if not os.path.isdir(flower_content):
             raise ValueError(f"{flower_content} is not a valid directory")
 
         super().__init__(name=name, min_clients=min_clients, mandatory_clients=mandatory_clients)
 
-        if client_api_type not in ["EX_PROCESS_API", "IN_PROCESS_API"]:
-            raise ValueError("Invalid client api type.")
-        os.environ["CLIENT_API_TYPE"] = client_api_type
```
SYangster, line 81: Just to clarify, why is this no longer needed? If CLIENT_API_TYPE is not set I believe it defaults to "IN_PROCESS_API" now

YuanTingHsieh: yes, we decided not to have this one hard-coded here.
Instead, users will pass all the extra env they need.
That way it also works with the situations outside of Job API (for example POC, production)

YuanTingHsieh (approved):
LGTM, I will make changes to example in another PR.

PR #2875, opened 2024-08-29
Make the Launcher extends FLComponent

yhwen:
Fixes # .

### Description

Make the Launcher extends FLComponent. Print out the class loading error for module scan.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/app_common/abstract/launcher.py
+++ b/nvflare/app_common/abstract/launcher.py
@@ -15,6 +15,7 @@
 import os
 from abc import ABC, abstractmethod
 
+from nvflare.apis.fl_component import FLComponent
 from nvflare.apis.fl_context import FLContext
 from nvflare.apis.shareable import Shareable
 from nvflare.apis.signal import Signal
@@ -28,7 +29,7 @@ class LauncherRunStatus:
     NOT_RUNNING = "not_running"
 
 
-class Launcher(ABC):
+class Launcher(FLComponent, ABC):
     def initialize(self, fl_ctx: FLContext) -> None:
         pass
 
--- a/nvflare/fuel/utils/class_utils.py
+++ b/nvflare/fuel/utils/class_utils.py
@@ -105,7 +105,7 @@ class ModuleScanner:
                                         else:
                                             class_table[name] = [module_name]
                             except (ModuleNotFoundError, RuntimeError, AttributeError) as e:
-                                self._logger.debug(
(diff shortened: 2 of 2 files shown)


nvflare/app_common/abstract/launcher.py
```diff
@@ -28,7 +29,7 @@ class LauncherRunStatus:
     NOT_RUNNING = "not_running"
 
 
-class Launcher(ABC):
+class Launcher(FLComponent, ABC):
```
YuanTingHsieh, line 32: What is the motivation of this change?

Previously I assume users might be writing their own launcher without knowing NVFlare specific concepts.

But maybe that assumption is not correct.

yhwen: This will enable the Launcher implementation class be able to handle the NVFlare events. Also be able to create the instance by "name".

yanchengnv: We could. On the other hand, I think the SubprocessLauncher should have been a FLComponent any way so that it will benefit from things like event handler.

Anyway, this is no big deal. I think we can call it done after this. Any other classes missing from the table will be added via "path".

yanchengnv: About YT's comment, I don't think making Launcher a FLComponent conflicts with the idea that "users might be writing their own launcher without knowing NVFlare specific concepts".  User still will be able to write whatever launcher they want, just now that their launcher can now also benefit from NVFLARE's event handling mechanism.

yanchengnv: Besides, YT's statement seems to indicate that knowing NVFLARE specific concepts is a bad thing. I have a serious problem with this! 

Knowing and understanding NVFLARE specific concepts is a wonderful thing, as long as we do not force the user to learn them before doing anything.

yanchengnv (approved):
LGTM

nvflare/app_common/abstract/launcher.py
YuanTingHsieh: Sounds good, I agree that as long as we do not force them, then it is good.

Apologize that my phrase was not accurate.

What I mean is we don't want users to be overwhelmed with our complicated (yet great) systems specifics to begin with any customization they might want to do.

This change LGTM.

In that argument, we could make some other components FLComponent as well in later release.

PR #2876, opened 2024-08-29
Added more handling for the source file import handling

yhwen:
Fixes # .

### Description

Added more handling for the source file import handling for the job API.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/job_config/fed_job_config.py
+++ b/nvflare/job_config/fed_job_config.py
@@ -230,7 +230,15 @@ class FedJobConfig:
 
             import_source_file = os.path.join(source_dir, import_source.replace(".", os.sep) + ".py")
             if os.path.exists(import_source_file):
+                # Handle the import from within the same module
                 self._get_custom_file(custom_dir, import_module, import_source_file)
+            else:
+                # Handle the import from outside the module
+                size = len(module.split(".")) - 1
+                source_root = os.sep.join(source_dir.split(os.sep)[0:-size])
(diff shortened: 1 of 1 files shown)


nvflare/job_config/fed_job_config.py
```diff
@@ -230,7 +230,15 @@ def _copy_source_file(self, custom_dir, module, source_file, dest_file):
...
             import_source_file = os.path.join(source_dir, import_source.replace(".", os.sep) + ".py")
             if os.path.exists(import_source_file):
+                # Handle the import from within the same module
                 self._get_custom_file(custom_dir, import_module, import_source_file)
+            else:
+                # Handle the import from outside the module
+                size = len(module.split(".")) - 1
+                source_root = os.sep.join(source_dir.split(os.sep)[0:-size])
+                import_source_file = os.path.join(source_root, import_source.replace(".", os.sep) + ".py")
+                if os.path.exists(import_source_file):
```
yanchengnv, line 240: What happens if import_source_file does not exist? Is this an error or a normal case that does not require the source file to be copied over?

yanchengnv (commented):
Looks good, but left questions.

nvflare/job_config/fed_job_config.py
yhwen: If import_source_file does not exist, it simply skip and not copy to the job custom folder. That applies to the standard package, like "torch", etc.

yhwen:
> Looks good, but left questions.

replied.

yanchengnv (approved):
LGTM

PR #2878, opened 2024-08-30
Fix heartbeat timeout config

SYangster:
Fixes [4820226](https://nvbugspro.nvidia.com/bug/4820226).

- Using the LauncherExecutor with Cyclic failed due to the FlareAgent timing out while the first site trains
- Add heartbeat_timeout to config for FlareAgent and set heartbeat_timeout=0 to not timeout in ScriptRunner ex-process mode
 

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/app_common/executors/client_api_launcher_executor.py
+++ b/nvflare/app_common/executors/client_api_launcher_executor.py
@@ -103,6 +103,7 @@ class ClientAPILauncherExecutor(LauncherExecutor):
         self._params_exchange_format = params_exchange_format
         self._params_transfer_type = params_transfer_type
         self._config_file_name = config_file_name
+        self._heartbeat_timeout = heartbeat_timeout
 
     def initialize(self, fl_ctx: FLContext) -> None:
         self.prepare_config_for_launch(fl_ctx)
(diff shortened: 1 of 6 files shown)


nvflare/app_common/executors/client_api_launcher_executor.py
```diff
@@ -103,6 +103,7 @@ def __init__(
         self._params_exchange_format = params_exchange_format
         self._params_transfer_type = params_transfer_type
         self._config_file_name = config_file_name
+        self._heartbeat_timeout = heartbeat_timeout
```
YuanTingHsieh, line 106: self.heartbeat_timeout is in the very base class TaskExchanger, should we reuse that instead?

nvflare/app_common/executors/client_api_launcher_executor.py
```diff
@@ -122,6 +123,7 @@ def prepare_config_for_launch(self, fl_ctx: FLContext):
                 ConfigKey.CLASS_NAME: pipe_export_class,
                 ConfigKey.ARG: pipe_export_args,
             },
+            ConfigKey.HEARTBEAT_TIMEOUT: self._heartbeat_timeout,
```
YuanTingHsieh, line 126: self.heartbeat_timeout is in the very base class TaskExchanger, should we reuse that instead?

YuanTingHsieh (commented):
Mostly good, thanks for finding the issue!

PR #2886, opened 2024-08-30
Update CrossSiteEval

YuanTingHsieh:
`get_model_from_location` is only available in PTModelFilePersistor.

### Description

- Adds a check to make sure what's passed in is `ModelPersistor`
- Change to a more generic `get_model` that is available for all `ModelPersistor`

### Types of changes

- [ ] Documentation updated.


SYangster (approved):
Good changes, using `get_model` is much better

PR #2889, opened 2024-08-30
Update getting_started cifar notebook

ZiyueXu77:
Add TB streaming components to notebook per VDR finding 

### Description

### Types of changes

- [ ] Documentation updated.


examples/getting_started/pt/nvflare_pt_getting_started.ipynb
```diff
@@ -377,7 +399,30 @@
...
+  {
+   "cell_type": "code",
+   "execution_count": null,
+   "id": "a8a733e0-c0a9-4c36-b49d-16b20c2df7f6",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "from nvflare.app_common.widgets.convert_to_fed_event import ConvertToFedEvent\n",
+    "\n",
+    "for i in range(n_clients):\n",
```
SYangster, line 423: Do you think for getting started maybe it's simpler to use job.to_clients() for this and the ScriptRunner?

Otherwise changes look good if we want this example to support metrics streaming. I'll update the other framework versions of the notebooks/site code blocks after this PR is merged.

ZiyueXu77: I think in this script we are using job.to for all places, so it would be better to avoid introducing variants like job.to_clients. Let's keep the current way.

YuanTingHsieh (approved):
LGTM

PR #2895, opened 2024-09-03
Update random forest and vertical xgb examples

ZiyueXu77:
Fixes findings regarding GPU utilization from QA.

### Description

Explicitly add use_gpus item to job so that it can be updated

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/examples/advanced/random_forest/jobs/random_forest_base/app/config/config_fed_client.json
+++ b/examples/advanced/random_forest/jobs/random_forest_base/app/config/config_fed_client.json
@@ -23,6 +23,7 @@
           "max_depth": 8,
           "eval_metric": "auc",
           "tree_method": "hist",
+          "use_gpus": false,
           "nthread": 16
         }
       }
--- a/examples/advanced/vertical_xgboost/README.md
+++ b/examples/advanced/vertical_xgboost/README.md
@@ -14,7 +14,7 @@ In this example we showcase a binary classification task based on the [HIGGS dat
 
 ### Download and Store Dataset
 First download the dataset from the HIGGS link above, which is a single zipped `.csv` file.
-By default, we assume the dataset is downloaded, uncompressed, and stored in `~/dataset/HIGGS.csv`.
+By default, we assume the dataset is downloaded, uncompressed, and stored in `DATASET_ROOT/HIGGS.csv`.
 
 ### Vertical Data Splits
 In vertical federated learning, sites share overlapping data samples (rows), but contain different features (columns).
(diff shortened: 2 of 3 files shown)


examples/advanced/random_forest/jobs/random_forest_base/app/config/config_fed_client.json
```diff
@@ -23,6 +23,7 @@
           "max_depth": 8,
           "eval_metric": "auc",
           "tree_method": "hist",
+          "use_gpus": false,
```
YuanTingHsieh, line 26: should we still use this "use_gpus" flag?
or we should have them specified "device": "cpu"? or "device": "cuda"?

ZiyueXu77: it will likely leads to several changes to any examples related to nvflare.app_opt.xgboost.tree_based.executor.FedXGBTreeExecutor
Shall we keep it for now?

SYangster: @YuanTingHsieh Also I noticed in a recent PR you changed the use_gpus logic from
`self.xgb_params["device"] = f"cuda:{self.rank}"`
to:
`device = self.xgb_params.get("device", "cuda:0")`
`self.log_info(fl_ctx, f"Training with GPU {device}")`
 `self.xgb_params["device"] = device`
 
 What was the reason behind this change, and do you think this could have possibly broken the vertical xgboost gpu case?

YuanTingHsieh: @SYangster good observation.

We need "device = self.xgb_params.get("device", "cuda:0")" so we don't overwrite what users specified.

As for the default value of "cuda:0" VS "cuda:{self.rank}", I get the motivation from the discussion here: https://github.com/NVIDIA/NVFlare/pull/2779

The old cuda:{self.rank} force rank 0 to run on GPU 0, rank 1 to GPU 1, etc.


If each client run on different machine, then this will be: rank 0 to run on machine 0 GPU 0, rank 1 to machine 1 GPU 1.
In reality, I think it makes more sense to default to "cuda:0", since each rank should run on different machines:
rank 0 to run on machine 0 GPU 0, rank 1 to machine 1 GPU 0.

BUT if QA is running multiple clients in the SAME machine, then only GPU 0 will be used. (V.S. before each client will be assigned to a different GPU)

YuanTingHsieh (approved):
LGTM, we also need one for "Same case for vertical XGB page:
https://github.com/NVIDIA/NVFlare/tree/main/examples/advanced/vertical_xgboost#gpu-support"

PR #2916, opened 2024-09-05
Added the current-round info the fl_ctx for BaseModelController

yhwen:
Fixes # .

### Description

Corrected the way how to add the current-round info the fl_ctx for BaseModelController. Fixed the missing current_round warning for cifar10 examples.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/app_common/workflows/base_model_controller.py
+++ b/nvflare/app_common/workflows/base_model_controller.py
@@ -224,9 +224,6 @@ class BaseModelController(Controller, FLComponentWrapper, ABC):
         result_model.meta["props"] = client_task.task.props[AppConstants.META_DATA]
         result_model.meta["client_name"] = client_name
 
-        if result_model.current_round is not None:
-            self.fl_ctx.set_prop(AppConstants.CURRENT_ROUND, result_model.current_round, private=True, sticky=True)
-
         self.event(AppEventType.BEFORE_CONTRIBUTION_ACCEPT)
         self._accept_train_result(client_name=client_name, result=result, fl_ctx=fl_ctx)
         self.event(AppEventType.AFTER_CONTRIBUTION_ACCEPT)
@@ -378,6 +375,15 @@ class BaseModelController(Controller, FLComponentWrapper, ABC):
         self.info(f"Sampled clients: {clients}")
 
         return clients
+
+    def set_fl_context(self, data: FLModel):
+        """Set up the fl_ctx information based on the passed in FLModel data.
+
+        """
+        if data and data.current_round is not None:
+            self.fl_ctx.set_prop(AppConstants.CURRENT_ROUND, data.current_round, private=True, sticky=True)
+        else:
+            self.warning(f"The FLModel data does not contain the current_round information.")
 
     def get_component(self, component_id: str):
         return self.engine.get_component(component_id)
--- a/nvflare/app_common/workflows/model_controller.py
+++ b/nvflare/app_common/workflows/model_controller.py
@@ -63,6 +63,7 @@ class ModelController(BaseModelController, ABC):
         Returns:
             List[FLModel]
         """
+        self.set_fl_context(data)
         return super().broadcast_model(
             task_name=task_name,
             data=data,
@@ -94,6 +95,7 @@ class ModelController(BaseModelController, ABC):
         Returns:
             None
         """
+        self.set_fl_context(data)
         super().broadcast_model(
             task_name=task_name,
             data=data,
--- a/research/fed-bpt/src/global_es.py
+++ b/research/fed-bpt/src/global_es.py
@@ -95,7 +95,7 @@ class GlobalES(FedAvg):
 
             clients = self.sample_clients(self.num_clients)
 
-            global_model = FLModel(params={"global_es": global_es})
+            global_model = FLModel(params={"global_es": global_es}, current_round=self.current_round)
             results = self.send_model_and_wait(targets=clients, data=global_model)
 
             # get solutions from clients


nvflare/app_common/workflows/model_controller.py
```diff
@@ -94,6 +95,7 @@ def send_model(
         Returns:
             None
         """
+        self.set_fl_context(data)
```
YuanTingHsieh, line 98: or should we add this in the "broadcast_model" of the "BaseModelController", what is the difference?

research/fed-bpt/src/global_es.py
```diff
@@ -95,7 +95,7 @@ def run(self) -> None:
 
             clients = self.sample_clients(self.num_clients)
 
-            global_model = FLModel(params={"global_es": global_es})
+            global_model = FLModel(params={"global_es": global_es}, current_round=self.current_round)
```
YuanTingHsieh, line 98: Is there any other places we need to change?

@holgerroth @SYangster @ZiyueXu77

nvflare/app_common/workflows/base_model_controller.py
```diff
@@ -379,6 +376,13 @@ def sample_clients(self, num_clients: int = None) -> List[str]:
 
         return clients
 
+    def set_fl_context(self, data: FLModel):
+        """Set up the fl_ctx information based on the passed in FLModel data."""
+        if data and data.current_round is not None:
+            self.fl_ctx.set_prop(AppConstants.CURRENT_ROUND, data.current_round, private=True, sticky=True)
+        else:
+            self.warning("The FLModel data does not contain the current_round information.")
+
```
YuanTingHsieh, line 385: any risk moving from `_process_result` to before @SYangster

nvflare/app_common/workflows/model_controller.py
yanchengnv: That's right. We should do everything in as low level as possible. BaseModelController is a better place. And we probably need to call set_fl_context() in broadcast_model as well.

nvflare/app_common/workflows/base_model_controller.py
yanchengnv: That was the kludge that we had to do in order to make the FLModel based approach still work with other FLContext/Event based components (such as best model selectors).

In FLModel based approach, both Data and Contextual info (such as current round) are in the FLModel object. But someone must place contextual info into the FLContext, otherwise those event based components will break.

We looked everywhere and just couldn't find a good place to do it. For example, the current round is supposed to be owned by the workflow, but the workflow writer could be a researcher that "we want to hide details from", and hence he won't know things like FLContext.

So what can we do? The next logical place that can put contextual info into FLContext is those "send" methods, which are under our control. There is a requirement that the user must do: set the current round in the FLModel object!

I don't really like this solution, but this is the best we can do right now.

I think we need to take another hard look at this, and come up with something better.

research/fed-bpt/src/global_es.py
yanchengnv: Now it is important the FLModel object MUST have the current round info!

nvflare/app_common/workflows/base_model_controller.py
```diff
@@ -379,6 +376,13 @@ def sample_clients(self, num_clients: int = None) -> List[str]:
 
         return clients
 
+    def set_fl_context(self, data: FLModel):
+        """Set up the fl_ctx information based on the passed in FLModel data."""
+        if data and data.current_round is not None:
+            self.fl_ctx.set_prop(AppConstants.CURRENT_ROUND, data.current_round, private=True, sticky=True)
+        else:
```
yanchengnv, line 383: Since this is now general purpose, I'm not sure this warning is needed. The workflow may not always be round-based ...

yanchengnv (changes requested):
Suggest to move the set_fl_context() calls to BaseModelController.

research/fed-bpt/src/global_es.py
SYangster: Right, passing in the round will be needed for all workflows that want to use the IntimeModelSelector for model selection

nvflare/app_common/workflows/model_controller.py
yhwen: Moved the self.set_fl_context(data) call to broadcast_model(). Also removed the no used logic to create the dummy FLModel object.

research/fed-bpt/src/global_es.py
yhwen: I have searched the usage. No other places need to add this current_round data when creating the FLModel object.

nvflare/app_common/workflows/base_model_controller.py
```diff
@@ -129,8 +129,10 @@ def broadcast_model(
 
         if not isinstance(task_name, str):
             raise TypeError("task_name must be a string but got {}".format(type(task_name)))
-        if data and not isinstance(data, FLModel):
-            raise TypeError("data must be a FLModel or None but got {}".format(type(data)))
+        if not data:
```
yanchengnv, line 132: Why are we adding this check here? Could this break existing code?

nvflare/app_common/workflows/base_model_controller.py
```diff
@@ -139,9 +141,7 @@ def broadcast_model(
         if not blocking and not isinstance(callback, Callable):
             raise TypeError("callback must be defined if blocking is False, but got {}".format(type(callback)))
 
-        if not data:
```
yanchengnv, line 142: Can this break existing cases?

nvflare/app_common/workflows/base_model_controller.py
```diff
@@ -139,9 +141,7 @@ def broadcast_model(
...
-            self.warning("data is None. Sending empty FLModel.")
-            data = FLModel(params_type=ParamsType.FULL, params={})
```
yanchengnv, line 144: We probably need to keep this code.

yanchengnv (changes requested):
I'm not sure we can disallow the "data" to be None in "broadcast_model", especially its default value is None.

If we are absolutely sure about this, then we need to remove the default None, and update the docstring accordingly.

If we are not sure, we'll need to keep original code.

nvflare/app_common/workflows/base_model_controller.py
yanchengnv: That's the thing - there will be a lot of warning messages. Maybe we simply don't log any warning messages because they could be misleading.  Or maybe just DEBUG level.

yhwen: Changed to DEBUG message.

nvflare/app_common/workflows/base_model_controller.py
yhwen: It does not make sense to broadcast_model() with an empty data. I checked all the places, we always create a FLModel object before calling this. This should not break existing codes.

nvflare/app_common/workflows/base_model_controller.py
yhwen: Changed the method signature to remove the default None value.

nvflare/app_common/workflows/base_model_controller.py
yhwen: Since we check the "not data" to raise exception, we don't need this code any more.

yanchengnv (approved):
LGTM

PR #2917, opened 2024-09-05
Fix data save path

YuanTingHsieh:
### Description

Fix data save path of test job

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/tests/integration_test/data/jobs/hello-pt-cse/app/custom/train.py
+++ b/tests/integration_test/data/jobs/hello-pt-cse/app/custom/train.py
@@ -27,9 +27,8 @@ import nvflare.client as flare
 from nvflare.app_common.app_constant import ModelName
 
 # (optional) set a fixed location so we don't need to download everytime
-CIFAR10_ROOT = "/tmp/nvflare/data"
-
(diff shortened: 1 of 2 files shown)


tests/integration_test/data/jobs/hello-pt-cse/app/custom/train.py
```diff
@@ -27,9 +27,8 @@
 from nvflare.app_common.app_constant import ModelName
 
 # (optional) set a fixed location so we don't need to download everytime
-CIFAR10_ROOT = "/tmp/nvflare/data"
-
-MODEL_SAVE_PATH_ROOT = "/tmp/nvflare/workdir/cifar10"
+CIFAR10_ROOT = "~/data"
+MODEL_SAVE_PATH_ROOT = "~/data"
```
SYangster, line 31: so we want to save the model to the same location as the data? looks good otherwise

SYangster (approved):

tests/integration_test/data/jobs/hello-pt-cse/app/custom/train.py
YuanTingHsieh: yes, this is testing, we create and remove this folder after each test case.

PR #2919, opened 2024-09-06
Update to use BaseFedJob

SYangster:
Change getting started examples to use BaseFedJob to allow users to explicitly define controller

### Types of changes

- [ ] Documentation updated.


holgerroth (approved):
This is a good change.

PR #2921, opened 2024-09-06
BugFix: Hierarchical Fed Stats, prepare data: replace os.rename() function

chesterxgchen:
Fixes # .replace os.rename() to shutil.move() to avoid os.error when destination and src are in different mnt or devices

### Types of changes

- [ ] Documentation updated.


IsaacYangSLA (dismissed):
os.rename doesn't work across filesystems while  shutl.move does.  However, why we need to write a csv file then move it.  Can we write it directly to dataset_path/university-N/university-N.csv?

IsaacYangSLA (approved):
This update is better.

PR #2953, opened 2024-09-19
Fix research links redirects

SYangster:
### Types of changes

- [ ] Documentation updated.


research/auto-fed-rl/index.html
YuanTingHsieh, line 1: so we don't need these index.html now?

SYangster: Right, now that we are using https://nvidia.github.io/NVFlare/ for our website, these index.html redirects aren't needed anymore

YuanTingHsieh: got it, sounds good.

YuanTingHsieh (approved):
LGTM

PR #2967, opened 2024-09-26
[2.5] Fix simulator result path

YuanTingHsieh:
### Description

Fix simulator result path

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/examples/advanced/xgboost/histogram-based/xgboost_histogram_higgs.ipynb
+++ b/examples/advanced/xgboost/histogram-based/xgboost_histogram_higgs.ipynb
@@ -138,7 +138,7 @@
    "outputs": [],
    "source": [
     "%load_ext tensorboard\n",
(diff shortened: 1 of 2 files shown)


YuanTingHsieh:
@ZiyueXu77 Do we need to correct the path for BRAT18 and prostate as well?

ZiyueXu77:
> @ZiyueXu77 Do we need to correct the path for BRAT18 and prostate as well?

For brats, I don't think so since we directly use the root folder "tensorboard --logdir='./workspace_brats'"
But for prostate, we may need to update the path in https://github.com/NVIDIA/NVFlare/blob/main/examples/advanced/prostate/result_stat/plot_tensorboard_events.py (although we used blob to search for the record)

ZiyueXu77 (approved):
looks good

PR #2972, opened 2024-10-01
update PSI to support python 3.11

chesterxgchen:
Update openmind-psi ==2.0.4 to support python 3.11

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh (approved):
We could see if we want to add it back to setup.cfg when 3.12 is supported

PR #2975, opened 2024-10-01
[Main] Support object reuse

yanchengnv:
Fixes # .

### Description

Currently when the same object is used in a job, duplicate config will be created, which will cause two object instances to be created during run time.

This PR makes it possible for the object to be reused multiple times without creating separate object instances.

This change is necessary in case that same object instance (e.g. model persistor) is used for different workflows in the same job.

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh (approved):
LGTM

SYangster (approved):

PR #2978, opened 2024-10-02
Replace the distutils with shutil

yhwen:
Fixes # .

### Description

The disutils package has been deprecated in Python 3.12. Replaces it with shutil package.

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh (approved):
Although these 2 methods are not 100% equivalent, in our use case of creating new directories via copying, these changes LGTM.

PR #2980, opened 2024-10-02
Allow multiple workflows in CCWF

yanchengnv:
Fixes # .

### Description

Allow multiple workflows in the same CCWF job (e.g. a Swarm followed by Cyclic).

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/app_common/ccwf/ccwf_job.py
+++ b/nvflare/app_common/ccwf/ccwf_job.py
@@ -30,6 +30,8 @@ from .cyclic_server_ctl import CyclicServerController
 from .swarm_client_ctl import SwarmClientController
 from .swarm_server_ctl import SwarmServerController
 
+_EXECUTOR_TASKS = ["train", "validate", "submit_model"]
+
 
 class SwarmServerConfig:
     def __init__(
@@ -202,6 +204,7 @@ class CCWFJob(FedJob):
             external_resources (str, optional): External resources directory or filename. Defaults to None.
         """
         super().__init__(name, min_clients, mandatory_clients)
+        self.executor = None
         if external_resources:
             self.to_server(external_resources)
             self.to_clients(external_resources)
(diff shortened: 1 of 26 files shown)


nvflare/app_common/ccwf/ccwf_job.py
```diff
@@ -249,7 +262,10 @@ def add_swarm(
             wait_time_after_min_resps_received=client_config.wait_time_after_min_resps_received,
         )
         self.to_clients(client_controller, tasks=["swarm_*"])
-        self.to_clients(client_config.executor, tasks=["train", "validate", "submit_model"])
+        if not self.executor:
+            # We add the executor only if it's not added yet.
```
SYangster, line 266: If the executor is already added, would it be clearer to raise an error? Or the current way of disregarding any additional client_config.executors is good enough

yanchengnv: The user could call add_swarm and then add_cyclic using the same executor in the client_config. Otherwise the user would have to set the "executor" to None in client_config of the 2nd call, which is not allowed, and would make it not reusable.

SYangster (approved):
LGTM

nvflare/app_common/ccwf/ccwf_job.py
```diff
@@ -287,7 +303,11 @@ def add_cyclic(
             final_result_ack_timeout=client_config.final_result_ack_timeout,
         )
         self.to_clients(client_controller, tasks=["cyclic_*"])
-        self.to_clients(client_config.executor, tasks=["train", "validate", "submit_model"])
+
+        if not self.executor:
+            # We add the executor only if it's not added yet.
```
YuanTingHsieh, line 308: it might be good if we have a way of warning the users that if one executor is added, then the following will be ignored.

YuanTingHsieh (approved):
LGTM

nvflare/app_common/ccwf/ccwf_job.py
yanchengnv: It might. But at this moment, let's go without since this is a very rarely used.

PR #2984, opened 2024-10-02
Fix tf model persistor and tf model

YuanTingHsieh:
Fix tf model persistor and tf model

### Description

- add missing argument
- allow passing persistor object

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/app_opt/tf/job_config/model.py
+++ b/nvflare/app_opt/tf/job_config/model.py
@@ -16,24 +16,27 @@ from typing import Optional
 
 import tensorflow as tf
 
+from nvflare.app_common.abstract.model_persistor import ModelPersistor
 from nvflare.app_opt.tf.model_persistor import TFModelPersistor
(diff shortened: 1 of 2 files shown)


nvflare/app_opt/tf/job_config/model.py
```diff
@@ -12,22 +12,30 @@
...
         If model is a tf.keras.Model, add a TFModelPersistor with the model.
 
         Args:
             model (any): model
+            persistor (Optional[TFModelPersistor]): A TFModelPersistor,
+            if provided will ignore argument `model`, defaults to None.
         """
         self.model = model
 
+        validate_object_for_job("persistor", persistor, TFModelPersistor)
```
SYangster, line 36: we should check `if persistor is not None:` before validating

YuanTingHsieh: good catch, let me fix it.

SYangster (approved):
Thanks LGTM

PR #2985, opened 2024-10-03
Allow customization of BaseFedJob 

YuanTingHsieh:
### Description
- Allow customization of BaseFedJob 
- Allow users to pass in customize ModelPersistor and ModelLocator for PTModel wrapper
- TF side need PR #2984

### Types of changes

- [ ] Documentation updated.


nvflare/app_opt/pt/job_config/base_fed_job.py
```diff
@@ -16,28 +16,35 @@
...
 from nvflare.app_common.widgets.intime_model_selector import IntimeModelSelector
+from nvflare.app_common.widgets.streaming import AnalyticsReceiver
 from nvflare.app_common.widgets.validation_json_generator import ValidationJsonGenerator
 from nvflare.app_opt.pt.job_config.model import PTModel
 from nvflare.app_opt.tracking.tb.tb_receiver import TBAnalyticsReceiver
+from nvflare.job_config.common_components_job import CommonComponentsJob
 
 
-class BaseFedJob(FedJob):
+class BaseFedJob(CommonComponentsJob):
```
yanchengnv, line 31: Can we simply combine CommonComponentsJob into BaseFedJob? BaseFedJob is already a "base" class, and it is not used directly by the users any way.

nvflare/job_config/common_components_job.py
```diff
@@ -0,0 +1,61 @@
...
+
+from typing import List, Optional
+
+from nvflare.app_common.widgets.convert_to_fed_event import ConvertToFedEvent
+from nvflare.app_common.widgets.intime_model_selector import IntimeModelSelector
+from nvflare.app_common.widgets.validation_json_generator import ValidationJsonGenerator
+from nvflare.job_config.api import FedJob, validate_object_for_job
+
+
+class CommonComponentsJob(FedJob):
```
yanchengnv, line 23: Do we have any other cases that will extend from this class? If not, it can be combined into BaseFedJob.

yanchengnv (changes requested):
See my comments. Not sure we need CommonComponentsJob.

nvflare/app_opt/pt/job_config/base_fed_job.py
YuanTingHsieh: thanks, merged them as suggested.

nvflare/job_config/common_components_job.py
YuanTingHsieh: thanks, merged them as suggested.

yanchengnv (approved):
LGTM

PR #2986, opened 2024-10-03
Refactoring F3 Streaming sender

nvidianz:
### Description

Refactored byte_streamer.py to change TxTask into an object, similar to the refactoring of RxTask.

## Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/fuel/f3/streaming/byte_streamer.py
+++ b/nvflare/fuel/f3/streaming/byte_streamer.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 import logging
 import threading
-from typing import Optional
+from typing import Callable, Optional
 
 from nvflare.fuel.f3.cellnet.core_cell import CoreCell
 from nvflare.fuel.f3.cellnet.defs import MessageHeaderKey
@@ -38,7 +38,7 @@ from nvflare.fuel.f3.streaming.stream_utils import (
 
 STREAM_CHUNK_SIZE = 1024 * 1024
 STREAM_WINDOW_SIZE = 16 * STREAM_CHUNK_SIZE
-STREAM_ACK_WAIT = 60
+STREAM_ACK_WAIT = 300
 
 STREAM_TYPE_BYTE = "byte"
 STREAM_TYPE_BLOB = "blob"
@@ -51,10 +51,21 @@ log = logging.getLogger(__name__)
 
 class TxTask:
     def __init__(
-        self, channel: str, topic: str, target: str, headers: dict, stream: Stream, secure: bool, optional: bool
+        self,
+        cell: CoreCell,
+        chunk_size: int,
+        channel: str,
+        topic: str,
+        target: str,
+        headers: dict,
+        stream: Stream,
+        secure: bool,
+        optional: bool,
     ):
+        self.cell = cell
+        self.chunk_size = chunk_size
         self.sid = gen_stream_id()
-        self.buffer = bytearray(ByteStreamer.get_chunk_size())
+        self.buffer = wrap_view(bytearray(chunk_size))
         # Optimization to send the original buffer without copying
         self.direct_buf: Optional[bytes] = None
         self.buffer_size = 0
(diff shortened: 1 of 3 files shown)


nvflare/fuel/f3/streaming/byte_streamer.py
```diff
@@ -71,17 +82,162 @@ def __init__(
...
 
 class ByteStreamer:
     def __init__(self, cell: CoreCell):
         self.cell = cell
         self.cell.register_request_cb(channel=STREAM_CHANNEL, topic=STREAM_ACK_TOPIC, cb=self._ack_handler)
         self.tx_task_map = {}
         self.map_lock = threading.Lock()
+        self.chunk_size = CommConfigurator().get_streaming_chunk_size(STREAM_CHUNK_SIZE)
 
         self.sent_stream_counter_pool = StatsPoolManager.add_counter_pool(
```
yanchengnv, line 242: Oh No! You are creating one pool for each ByteStreamer object. There will be many pool objects created and wasted. You should create one such pool for the whole process. Otherwise, the results in the pool won't be correct.

nvflare/fuel/f3/streaming/byte_streamer.py
```diff
@@ -109,13 +264,10 @@ def send(
...
     ) -> StreamFuture:
-        tx_task = TxTask(channel, topic, target, headers, stream, secure, optional)
+        tx_task = TxTask(self.cell, self.chunk_size, channel, topic, target, headers, stream, secure, optional)
         with self.map_lock:
             self.tx_task_map[tx_task.sid] = tx_task
 
-        future = StreamFuture(tx_task.sid)
-        future.set_size(stream.get_size())
-        tx_task.stream_future = future
         tx_task.task_future = stream_thread_pool.submit(self._transmit_task, tx_task)
```
yanchengnv, line 271: Is the task_future used anywhere else? If not, then no need to keep it with tx_task.
Besides, task_future belongs to tx_task, and shouldn't be set by ByteStreamer object.
This "submit" call could be put into a method of TxTask.

yanchengnv (changes requested):
See my comments. There are other places that have the pool creation issues. Please fix them all.

nvflare/fuel/f3/streaming/byte_streamer.py
```diff
@@ -71,17 +82,162 @@ def __init__(
         self.offset_ack = 0
         self.secure = secure
         self.optional = optional
+        self.stopped = False
+
+        self.stream_future = StreamFuture(self.sid)
```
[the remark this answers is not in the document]
nvidianz: It's for future use. Refactored to hide it.

nvflare/fuel/f3/streaming/byte_streamer.py
```diff
@@ -126,145 +278,33 @@ def send(
...
-            task = self.tx_task_map.get(sid, None)
+            tx_task = self.tx_task_map.get(sid, None)
 
-        if not task:
+        if not tx_task:
+            origin = message.get_header(MessageHeaderKey.ORIGIN)
+            offset = message.get_header(StreamHeaderKey.OFFSET, None)
             # Last few ACKs always arrive late so this is normal
-            log.debug(f"ACK for stream {sid} received late from {origin} with offset {offset}")
+            log.debug(f"{tx_task} ACK for stream {sid} received late from {origin} with offset {offset}")
```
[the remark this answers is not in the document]
nvidianz: Good catch. Fixed.

nvflare/fuel/f3/streaming/byte_streamer.py
nvidianz: Changed the pool to be static and moved cell name from scope to category.

nvflare/fuel/f3/streaming/byte_streamer.py
nvidianz: Refactored as suggested.

nvflare/fuel/f3/streaming/byte_streamer.py
```diff
@@ -71,200 +82,233 @@ def __init__(
...
         self.secure = secure
         self.optional = optional
+        self.stopped = False
 
-    def __str__(self):
-        return f"Tx[SID:{self.sid} to {self.target} for {self.channel}/{self.topic}]"
+        self.stream_future = StreamFuture(self.sid)
+        self.stream_future.set_size(stream.get_size())
 
+        self.window_size = CommConfigurator().get_streaming_window_size(STREAM_WINDOW_SIZE)
```
YuanTingHsieh, line 90: so stream size is the total size of the payload to be sent?
what about "stream_window_size" and "chunk_size"?

YuanTingHsieh: chunk_size is how big to send for each chunk (we break the payload to chunks)
and stream_window_size is how many chunks can be sent/handled at the same time?

yanchengnv (approved):
LGTM

PR #3012, opened 2024-10-08
Fix fobs doc

YuanTingHsieh:
Fix fobs doc

### Description

- Move from code into directly our documentation so users don't need to go back and forth
- Add more limitations as QA suggested
- Add dataclass explanation

### Types of changes

- [ ] Documentation updated.


docs/user_guide/security/serialization.rst
```diff
@@ -5,4 +5,226 @@ Message Serialization
...
+serialize objects.
+
+FOBS sacrifices convenience for security. With Pickle, most objects are supported
+automatically using introspection. To serialize an object using FOBS, a **Decomposer**
+must be registered for the class. A few decomposers for commonly used classes are
+pre-registered with the module.
+
+FOBS supports enum types by registering decomposers automatically for all classes that
+are subclasses of :code:`Enum`.
+
```
nvidianz, line 26: Should add a paragraph here,

FOBS treats all other classes as dataclass by registering a generic decomposer for dataclasses. Dataclass is a class whose constructor only changes the state of the object without side-effects. Side-effects include changing global variables, creating network connection, files etc.

YuanTingHsieh: thanks, good suggestion, added

nvidianz (approved):

PR #3018, opened 2024-10-09
Support multiple host names for FLARE server

yanchengnv:
Fixes # .

### Description

Currently, the FLARE Server's host name can only be specified with the "name" attribute of the server in project.yml. When the server cert is generated, this name is also used as the Common Name (CN) of the cert. There are several problems with this:

- The max length for CN is 63 chars. If the host name is longer than that, then a cert cannot be generated.
- There are cases that multiple host names may be desired for the FLARE server (e.g. internal clients and external clients may need to use different host names).
- IP addresses cannot be used as host name since CN does not allow that. But there are cases that IP addresses are desired.

This PR solves all these 3 issues:

- Multiple host names (host_names) can be specified in the "server" element in the project.yml. This is a list of host names or IP addresses. The CN will continue to be treated as a host name, to be backward compatible. When generating server cert, these host names will be included in the SubjectAlternativeNames extension.

When defined as "host_names", the max length for each value is 253 chars (much larger than 63 chars).

- For client and admin, you can specify the "connect_to" attribute to select which host to use to connect to the FLARE server. Of course, the value of "connect_to" must be either the server's "name" or one of values in the server's "host_names" list.

- When startup config files are generated for clients and admin users, the sp_end_point value (for dummy agent) will use the "connect_to" value if specified.

NOTE: this PR only implements the support of multiple host names with CLI provision. Dashboard needs to be updated later to make this available for dashboard-based provision.

### Types of changes

- [ ] Documentation updated.


nvidianz (approved):
LGTM

IsaacYangSLA (dismissed):
Thanks for the PR.  LGTM.

IsaacYangSLA (approved):
Approve this again.  LGTM.

PR #3028, opened 2024-10-10
Improve json serialization to accomodate numpy float32

YuanTingHsieh:
Since numpy 2.0, some type promoting rules are different: https://numpy.org/devdocs/numpy_2_0_migration_guide.html#changes-to-numpy-data-type-promotion

### Description

- Change to always cast np.float32 to np.float64 (for serialization) before calling json.dump
- Remove unused calls of dxo.validate, it is already invoked inside the DXO init method and checked

### Types of changes

- [ ] Documentation updated.


nvflare/app_common/widgets/validation_json_generator.py
```diff
@@ -58,7 +72,6 @@ def handle_event(self, event_type: str, fl_ctx: FLContext):
             if val_results:
                 try:
                     dxo = from_shareable(val_results)
-                    dxo.validate()
```
SYangster, line 61: why don't we need to validate the dxo here anymore?

YuanTingHsieh: good question, we already did that in creation: https://github.com/NVIDIA/NVFlare/blob/main/nvflare/apis/dxo.py#L65

nvidianz (approved):

PR #3035, opened 2024-10-14
[2.5] Add TF based TBAnalyticsReceiver

YuanTingHsieh:
Add TF based TBAnalyticsReceiver

### Description

- Add TF based TBAnalyticsReceiver

### Types of changes

- [ ] Documentation updated.


nvflare/app_opt/tf/job_config/base_fed_job.py
```diff
@@ -94,13 +93,10 @@ def __init__(
...
-            analytics_receiver = TBAnalyticsReceiver()
-
-        self.to_server(
-            id="receiver",
-            obj=analytics_receiver,
-        )
+            self.to_server(
+                id="receiver",
+                obj=analytics_receiver,
+            )
```
IsaacYangSLA, line 99: If init does not have analytics_receiver, self.to_server will not run.  Then what will happen when other methods are called?  self.to_server calls self.to which does some operations.

IsaacYangSLA (approved):
Thanks for implementing this new class.  LGTM.

PR #3039, opened 2024-10-14
Update the GNN encoding for XGB financial example

ZiyueXu77:
Fixes previous result where GNN encoding performs worse than raw feature

### Description

GNN encoding is supposed to be used as secondary complementary features, rather than by itself, changing the behavior

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh (approved):
LGTM, great changes, the GNN is for more feature engineering

PR #3043, opened 2024-10-15
Support file source in Job API

yanchengnv:
Fixes # .

### Description

Currently when a Python file is added, it is subject to special processing like imports scanning. Depending the result of the scanning and how the file path is specified, it may be placed into a subfolder within the job's "custom" folder. There is no way to guarantee that the source file is placed in the desired location within the "custom" folder.

This PR fixes this by introducing the FileSource object. The user can now place any file(s) into any desired location with the job's "custom" folder. For example:

job.to(obj=FileSource(src_path="/a/b/c/xyz.txt", dest_dir="src/docs"), target="server")

This will copy the file /a/b/c/xyz.txt to the <custom_dir>/src/docs.

The src_path can also be a directory. In this case, the content of the directory will be copied into the dest_dir within the "custom" folder.

If dest_dir is not specified (None), then the actual destination will be the "custom" dir.

### Types of changes

- [ ] Documentation updated.


nvflare/job_config/fed_job_config.py
```diff
@@ -186,6 +186,26 @@ def _get_server_app(self, config_dir, custom_dir, fed_app):
...
         self._copy_ext_dirs(custom_dir, fed_app.server_app)
+        self._copy_file_sources(custom_dir, fed_app.server_app.file_sources)
+
+    def _copy_file_sources(self, custom_dir, file_sources):
+        for s in file_sources:
+            # s is a tuple of (src_path, dest_dir)
+            src_path, dest_dir = s
+            if dest_dir:
+                dest_path = os.path.join(custom_dir, dest_dir)
+                if not os.path.exists(dest_path):
```
YuanTingHsieh, line 197: I think We don't need this if we already pass "exist_ok=True"?

nvflare/job_config/fed_job_config.py
```diff
@@ -186,6 +186,26 @@ def _get_server_app(self, config_dir, custom_dir, fed_app):
...
+            else:
+                dest_path = custom_dir
+
+            if os.path.isfile(src_path):
+                base_name = os.path.basename(src_path)
+                dest_file = os.path.join(dest_path, base_name)
+                shutil.copy(src_path, dest_file)
+            else:
+                # this is a dir
+                shutil.copytree(src_path, dest_path, dirs_exist_ok=True)
```
YuanTingHsieh, line 208: Do we need to worry about if users somehow overwriting their own files in custom?

YuanTingHsieh (approved):
Changes LGTM

nvflare/job_config/fed_job_config.py
yanchengnv: But we need this for copying file.

nvflare/job_config/fed_job_config.py
yanchengnv: No. We assume that the user knows what he is doing.

PR #3050, opened 2024-10-25
Modify log config to write errors to a separate log file

nvkevlu:
Modify log config to write errors to a separate log file in addition to log.txt.

### Description

Modify log config to write errors to a separate log file in addition to log.txt. I think this covers all the places we provide a default log config and this will enable future work for the error logs to be aggregated for reporting and downloading.

Updated nvflare/private/fed/utils/fed_utils.py to add an error log file to write errors to using the configuration of the error file handler based on the errorFileHandler in log.config.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/docs/resources/log.config
+++ b/docs/resources/log.config
@@ -1,27 +1,27 @@
 [loggers]
-keys=root,modelLogger
+keys=root
 
 [handlers]
-keys=consoleHandler
+keys=consoleHandler,errorFileHandler
 
 [formatters]
 keys=fullFormatter
 
 [logger_root]
 level=INFO
-handlers=consoleHandler
-
-[logger_modelLogger]
-level=DEBUG
-handlers=consoleHandler
-qualname=modelLogger
-propagate=0
+handlers=consoleHandler,errorFileHandler
 
 [handler_consoleHandler]
 class=StreamHandler
 level=DEBUG
 formatter=fullFormatter
 args=(sys.stdout,)
+
+[handler_errorFileHandler]
+class=FileHandler
+level=ERROR
+formatter=fullFormatter
+args=('error.log', 'a')
 
 [formatter_fullFormatter]
 format=%(asctime)s - %(name)s - %(levelname)s - %(message)s
--- a/nvflare/private/fed/app/simulator/log.config
+++ b/nvflare/private/fed/app/simulator/log.config
@@ -2,20 +2,26 @@
 keys=root
 
 [handlers]
-keys=consoleHandler
+keys=consoleHandler,errorFileHandler
 
 [formatters]
 keys=fullFormatter
 
 [logger_root]
 level=INFO
-handlers=consoleHandler
+handlers=consoleHandler,errorFileHandler
 
 [handler_consoleHandler]
 class=StreamHandler
 level=DEBUG
 formatter=fullFormatter
 args=(sys.stdout,)
+
+[handler_errorFileHandler]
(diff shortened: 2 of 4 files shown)


yanchengnv:
@IsaacYangSLA please review. Thanks!

yanchengnv (dismissed):
Very nice!

docs/resources/log.config
```diff
@@ -1,27 +1,27 @@
...
 class=StreamHandler
 level=DEBUG
 formatter=fullFormatter
 args=(sys.stdout,)
 
+[handler_errorFileHandler]
+class=FileHandler
+level=ERROR
+formatter=fullFormatter
+args=('error.log', 'a')
```
YuanTingHsieh, line 24: When we say "error.log" here, where exactly is the logging file located?

nvflare/private/fed/app/simulator/log.config
```diff
@@ -2,20 +2,26 @@
...
 class=StreamHandler
 level=DEBUG
 formatter=fullFormatter
 args=(sys.stdout,)
 
+[handler_errorFileHandler]
+class=FileHandler
+level=ERROR
+formatter=fullFormatter
+args=('error.log', 'a')
```
YuanTingHsieh, line 24: Are we using this arg?

YuanTingHsieh (commented):

docs/resources/log.config
```diff
@@ -1,27 +1,27 @@
...
-propagate=0
+handlers=consoleHandler,errorFileHandler
 
 [handler_consoleHandler]
 class=StreamHandler
 level=DEBUG
 formatter=fullFormatter
 args=(sys.stdout,)
 
+[handler_errorFileHandler]
```
SYangster, line 20: Just to clarify, the reason we are using the template approach with errorFileHandler rather than always adding it in add_file_log_handler() is because we want to have the option to disable the error log, correct?

nvkevlu: Right, also Yan mentioned that we wanted to have the ability to customize it, for example if he wanted it to be at INFO level or with a different formatter.

docs/resources/log.config
nvkevlu: The change in add_logfile_handler of nvflare/private/fed/utils/fed_utils.py makes it so an error.log file would be created in the same location that the log.txt is being created in, so for this would work for inside each job folder.

nvflare/private/fed/app/simulator/log.config
nvkevlu: Currently not

YuanTingHsieh: Got it.

I would suggest either removing it in this PR or submit a follow up PR that makes use of this arg.

Thanks!

nvkevlu: Actually, I tried removing it and there are errors, so it seems that it does go through some code that checks that there is at least one required positional argument so we may need to keep this for now.

yanchengnv: You may be able to use it as template: use the configured file name instead of hard-coding it to error.log in fed_utils.  But this may create other complications later when you try to send this log file to the server - you won't know the file name any more unless you keep it somewhere (e.g. data bus).

Let's deal with this in another PR.

yanchengnv (approved):
LGTM

PR #3053, opened 2024-10-30
Support aborting messages

yanchengnv:
Fixes # .

### Description

This PR adds the capability to abort/cancel the sending of a message. Without this, the system cannot be quickly aborted if the message is big.  This function is more important when streaming large messages or large number of objects, which can take long time.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/fuel/f3/cellnet/cell.py
+++ b/nvflare/fuel/f3/cellnet/cell.py
@@ -16,9 +16,11 @@ import concurrent.futures
 import copy
 import logging
 import threading
+import time
 import uuid
 from typing import Dict, List, Union
 
+from nvflare.apis.signal import Signal
 from nvflare.fuel.f3.cellnet.core_cell import CoreCell, TargetMessage
 from nvflare.fuel.f3.cellnet.defs import CellChannel, MessageHeaderKey, MessageType, ReturnCode
 from nvflare.fuel.f3.cellnet.utils import decode_payload, encode_payload, make_reply
@@ -26,6 +28,7 @@ from nvflare.fuel.f3.message import Message
 from nvflare.fuel.f3.stream_cell import StreamCell
 from nvflare.fuel.f3.streaming.stream_const import StreamHeaderKey
 from nvflare.fuel.f3.streaming.stream_types import StreamFuture
+from nvflare.fuel.utils.waiter_utils import WaiterRC, conditional_wait
 from nvflare.security.logging import secure_format_exception
 
 CHANNELS_TO_EXCLUDE = (
@@ -147,6 +150,7 @@ class Cell(StreamCell):
         timeout=None,
         secure=False,
         optional=False,
+        abort_signal: Signal = None,
     ) -> Dict[str, Message]:
         """
         Send a message over a channel to specified destination cell(s), and wait for reply
@@ -159,6 +163,7 @@ class Cell(StreamCell):
             timeout: how long to wait for replies
             secure: End-end encryption
             optional: whether the message is optional
+            abort_signal: signal to abort the message
 
         Returns: a dict of: cell_id => reply message
 
@@ -181,6 +186,7 @@ class Cell(StreamCell):
                 req = Message(copy.deepcopy(request.headers), request.payload)
                 target_argument["request"] = TargetMessage(t, channel, topic, req).message
                 target_argument["target"] = t
+                target_argument["abort_signal"] = abort_signal
                 target_argument.update(fixed_dict)
                 f = executor.submit(self._send_one_request, **target_argument)
                 future_to_target[f] = t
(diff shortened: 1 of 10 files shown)


nvflare/fuel/f3/cellnet/cell.py
```diff
@@ -232,26 +237,43 @@ def _get_result(self, req_id):
...
-                return False
+        while True:
+            rc = conditional_wait(future.waiter, timeout, abort_signal, condition_cb=self._check_error, future=future)
+            if rc == WaiterRC.IS_SET:
+                # waiter has been set!
+                break
+            elif rc == WaiterRC.TIMEOUT:
+                # timed out: check whether any progress has been made during this time
+                current_progress = future.get_progress()
+                if last_progress == current_progress:
```
nvidianz, line 258: This should use another timeout value. On a slow network, it's likely that there is no progress for a long time.

yanchengnv: Not sure. In this case, the timeout value simply needs to be set to a big number.
It's true that different steps of the processing could use a different timeout value, but that would make the API very hard to use. We could just use a timeout value that accommodates the slowest step.

nvidianz (approved):

nvflare/fuel/f3/cellnet/cell.py
```diff
@@ -321,15 +346,17 @@ def _send_one_request(
 
             # waiting for receiving first byte
             self.logger.debug(f"{req_id=}: entering remote process wait {timeout=}")
-            if not waiter.in_receiving.wait(timeout):
-                self.logger.debug(f"{req_id=}: remote processing timeout {timeout=}")
+
+            waiter_rc = conditional_wait(waiter.in_receiving, timeout, abort_signal)
+            if waiter_rc != WaiterRC.IS_SET:
+                self.logger.debug(f"{req_id=}: remote processing timeout {timeout=} {waiter_rc=}")
```
YuanTingHsieh, line 352: Do we want different debug log for WaiterRC.ABORTED?

nvflare/fuel/f3/cellnet/cell.py
```diff
@@ -232,26 +237,43 @@ def _get_result(self, req_id):
...
+                    return False
+                else:
+                    # good progress
+                    self.logger.debug(f"{current_progress=}")
+                    last_progress = current_progress
             else:
-                self.logger.debug(f"{current_progress=}")
-                last_progress = current_progress
+                # error condition: aborted or future error
+                return False
```
YuanTingHsieh, line 267: if future has no error but timeout, the return is "WaiterRC.OK" and we go to this else statement? So OK is returning False?

YuanTingHsieh (commented):
Thanks for fixing some places of typos, missing docstrings, make code more readable.

Mostly LGTM, have some questions.

nvflare/fuel/f3/cellnet/cell.py
yanchengnv: I think this is enough since waiter_rc is logged already.

nvflare/fuel/f3/cellnet/cell.py
yanchengnv: No, WaiterRC.OK is only used for condition_cb and used internally by conditional_wait, and is never returned back here!

If condition_cb (which is self._check_error) returned any thing bad (-1), it will get to this "else" block.
If the condition_cb  returns WaiterRC.OK, then we get to this "else" block only if the abort_signal is triggered.

We only check current_progress when timed out, and decide whether we should keep waiting. If there is any progress, then we'll keep waiting; otherwise returns False.

nvflare/fuel/utils/waiter_utils.py
```diff
@@ -0,0 +1,71 @@
...
+        if abort_signal and abort_signal.triggered:
+            return WaiterRC.ABORTED
+
+        if condition_cb:
+            try:
+                rc = condition_cb(**cb_kwargs)
+                if rc != WaiterRC.OK:
+                    # a bad condition is detected by the condition_cb
+                    # we return the rc from the condition_cb
+                    return rc
```
YuanTingHsieh, line 69: If the "condition_cb" is returning WaiterRC.OK, what will happen?
This "conditional_wait" method will return None?

yanchengnv: If condition_cb returns WaiterRC.OK, it will go back to the beginning of the "while" loop, no?

YuanTingHsieh: got it, thanks for the detail explanation.

YuanTingHsieh (approved):
LGTM

PR #3055, opened 2024-11-01
Job launcher server side 

yhwen:
Fixes # .

### Description

Added the server side job launcher API support. Also added the ServerK8sJobLauncher for K8s kubernete.

### Types of changes

- [ ] Documentation updated.


nvflare/private/fed/utils/fed_utils.py
```diff
@@ -501,3 +501,15 @@ def get_scope_prop(scope_name: str, key: str) -> Any:
     check_str("key", key)
     data_bus = DataBus()
     return data_bus.get_data(_scope_prop_key(scope_name, key))
+
+
+def get_job_launcher(job_meta: dict, fl_ctx: FLContext) -> dict:
```
yanchengnv, line 506: This fl_ctx could be a long-running one and not job-specific. The property FLContextKey.JOB_LAUNCHER may already exist in it. Before firing the event, you need to remove this property, otherwise you may get the wrong result. But this solution is not thread-safe: in case this function is called multiple times at the same time, then it won't work properly.

The cleanest way is to create a new fl_ctx from the engine and then use it for the event processing.

yanchengnv (changes requested):
Looks good in general. I like the refactoring of the launchers so server and client launchers could be written more easily.

The get_job_launcher() function needs to be improved to be fl-ctx safe and thread safe. See my comments there.

nvflare/private/fed/utils/fed_utils.py
yhwen: Changed to use a new fl_ctx for get_job_launcher to solve the thread safe issue.

nvflare/private/fed/utils/fed_utils.py
```diff
@@ -501,3 +501,19 @@ def get_scope_prop(scope_name: str, key: str) -> Any:
...
+    engine = fl_ctx.get_engine()
+
+    job_launcher_ctx = engine.new_context()
```
yanchengnv, line 509: Should use the pattern:
with engine.new_context() as job_launcher_ctx:
....

nvflare/private/fed/utils/fed_utils.py
```diff
@@ -501,3 +501,19 @@ def get_scope_prop(scope_name: str, key: str) -> Any:
...
+    # Remove the potential not cleaned up JOB_LAUNCHER
+    job_launcher_ctx.remove_prop(FLContextKey.JOB_LAUNCHER)
```
yanchengnv, line 511: This is unnecessary since job_launcher_ctx is a new context.

nvflare/private/fed/utils/fed_utils.py
yhwen: Changed.

nvflare/private/fed/utils/fed_utils.py
yhwen: I added this to handle the case if some components accidentally set the JOB_LAUNCHER as a sticky in the FLConext.

yanchengnv (approved):
LGTM

PR #3056, opened 2024-11-03
Support extra provision builder generated component files

yanchengnv:
Fixes # .

### Description

The Flare Provision system allows any number of Builders to support certain functions (e.g. Confidential Computing), and some extra resources may be needed in run time for the functions to work properly. This PR implements this requirement.

The design of the Provision system encourages builders to be self-contained and independent of each other. When a builder needs to add extra resources, it should create its own config file for these resources, instead of adding them to config files managed by other builders.

At the start of Flare process (SP, SJ, CP, CJ), appropriate builder-generated config files will be loaded. Note that not all such files are for all processes: some are for parent process only, some are for job processes only, and some are for both.

This PR defines the naming convention for builder-generated resource config files:

- <builder_name>__resources.json: for both parent and job processes;
- <builder_name>__p_resources.json: for parent process only;
- <builder_name>__j_resources.json: for job process only.

When generating a config file, the builder must decide which process(es) the resources are for and name the file properly.
The builder also must decide where to place the config file in the startup kit: in the "startup" folder or in the "local" folder. Files in the "startup" folder are signed and cannot be tampered with; whereas files in the "local" folder can be modified by the site admin.

The naming convention must be strictly followed; otherwise the builder-generated config files won't be loaded during run time.

The format of the builder-generated config file is defined by the builder - it can define any properties needed by any components. HOWEVER, if it needs to add components, then the config file must have a "components" section, and components must be added to this section! This also the way that existing resources files use.

### Types of changes

- [ ] Documentation updated.


nvflare/private/fed/utils/fed_utils.py
```diff
@@ -186,6 +198,33 @@ def security_init(secure_train: bool, site_org: str, workspace: Workspace, app_v
...
+    startup_dir = workspace.get_startup_kit_dir()
+    SecurityContentService.initialize(content_folder=startup_dir)
+
+    if secure_train:
+        insecure_list = _check_secure_content(site_type=site_type)
+        if len(insecure_list):
+            print("The following files are not secure content.")
+            for item in insecure_list:
+                print(item)
+            sys.exit(1)
```
YuanTingHsieh, line 220: logic is good, just a note for all these print statements we need to also check it when doing the logging re-factoring to see if they should be logged.

nvflare/private/fed/utils/fed_utils.py
```diff
@@ -186,6 +198,33 @@ def security_init(secure_train: bool, site_org: str, workspace: Workspace, app_v
...
+
+    # initialize the AuditService, which is used by command processing.
+    # The Audit Service can be used in other places as well.
+    audit_file_name = workspace.get_audit_file_path()
+    AuditService.initialize(audit_file_name)
+
+
 def security_close():
```
YuanTingHsieh, line 228: minor, does this need to be called, "security_close_for_job"?

nvflare/apis/workspace.py
```diff
@@ -178,3 +179,58 @@ def get_stats_pool_records_path(self, job_id: str, prefix=None) -> str:
         if prefix:
             file_name = f"{prefix}.{file_name}"
         return os.path.join(self.get_run_dir(job_id), file_name)
+
+    def get_config_files_for_startup(self, is_server: bool, for_job: bool) -> list:
+        """Get all config files to be used for startup of the process (SP, SJ, CP, CJ).
+
+        We first get required config files:
+            - the startup file (fed_server.json or fed_client.json) in "startup" folder
+            - resource file (resources.json.default or resources.json) in "local" folder
```
YuanTingHsieh, line 188: we might need some documentation to clarify this one VS the other ones generated by provision?

YuanTingHsieh (approved):
LGTM

nvflare/private/fed/utils/fed_utils.py
yanchengnv: I just followed what exists now. But I think it's a good idea to refactor later.

nvflare/private/fed/utils/fed_utils.py
yanchengnv: Not really. This is the same for both cases (for now).

nvflare/apis/workspace.py
yanchengnv: Right. I'll put together a doc to describe how the resource files are handled.

PR #3058, opened 2024-11-07
Adjust tf/fedopt_ctl to include updates for the model's non-trainable

falibabaei:
Hi,

tf/fedopt_ctl does not handle models with non-trainable variables, causing training to fail with an error about mismatched gradients and variables.


nvflare/app_opt/tf/fedopt_ctl.py
```diff
@@ -106,55 +106,69 @@ def _to_tf_params_list(self, params: Dict, negate: bool = False):
...
-            f"{type(self.lr_scheduler)} "
-            f"lr: {self.optimizer.learning_rate}, "
-            f"update: {secs} secs., detach: {secs_detach} secs.",
-        )
-
-        global_model.params = new_weights
-        global_model.meta = aggr_result.meta
-
-        return global_model
+        def update_model(self, global_model: FLModel, aggr_result: FLModel):
```
YuanTingHsieh, line 109: Is this indentation off?

falibabaei: Sorry, it happened during the copying process. It has been resolved.

nvflare/app_opt/tf/fedopt_ctl.py
```diff
@@ -118,39 +118,53 @@ def update_model(self, global_model: FLModel, aggr_result: FLModel):
...
+            if key in model_diff_params:
+                new_weights[key] = weights[w_idx]
+
+            else:
 
+                new_weights[key] = (
+                    aggr_result.params[key]
+                    if aggr_result.params_type == ParamsType.FULL
+                    else global_model.params[key] + aggr_result.params[key]
+                )
```
YuanTingHsieh, line 161: do you need the 
"if aggr_result.params[key].shape == global_model.params[key].shape:" ?

YuanTingHsieh (commented):
one question, otherwise LGTM, please verify this with one of the example you already have, thanks!

nvflare/app_opt/tf/fedopt_ctl.py
falibabaei: Yes, since in the keys in aggr_result differ from the model layers, I need to identify the trainable variables within aggr_result.params[key]

YuanTingHsieh: if you need that, could you add that condition to your PR? thanks

falibabaei: Since I added this condition on line 133 of my script to separate the trainable variables and store them in the model_diff_param dictionary, I don't need to apply that logic again here. At this point, I am certain that the model_diff_param dictionary contains only the trainable variables.

YuanTingHsieh: line 160, "global_model.params[key] + aggr_result.params[key]"

if the shape mismatch this statement will fail.

but if you are sure that global_model.params[key] will always have the same shape as  aggr_result.params[key]
then we are good.

YuanTingHsieh:
@falibabaei can you run the code refactor and make sure the runtest.sh is passing? thanks

falibabaei:
done

falibabaei:
done :)

holgerroth (dismissed):
Looks good!

holgerroth (approved):

PR #3061, opened 2024-11-11
Support large object streaming

yanchengnv:
Fixes # .

### Description

This PR implements the following features:

- A general-purpose object streamer. 
- A file streamer implemented with object streamer.

The object streaming solution works as follows:

- On the sender side, a generator object is provided that generates objects to be streamed to the targets.  
- On the receiver side, a processor factory object is registered on channel/topic. When a stream object is received on the specified channel/topic, the factory object is called to get a new processor, which is used to process objects on the same stream.
- Stream objects are defined as Shareable objects. This makes it possible to pass them through Filters before sending (in future PR).
- Stream objects are sent through aux messages.

For the application to use the object streamer:
- The application must first register processing information on the receiving side by calling engine.register_stream_processing() on certain channel and topic, and provide processor factory and optionally a stream_done callback to be called when the streaming is completed.
- On the sending side, the app calls the engine.stream_shareables() to stream objects. Generator object must be provided, which generates an object to be streamed when called. The generator object is also called to process replies from receivers for each object sent.

### Types of changes

- [ ] Documentation updated.


nvflare/app_common/streamers/streamer_base.py
```diff
@@ -0,0 +1,28 @@
...
+    def get_channel(meta: StreamMeta):
+        return meta.get(StreamMetaKey.CHANNEL)
+
+    @staticmethod
+    def get_topic(meta: StreamMeta):
+        return meta.get(StreamMetaKey.TOPIC)
+
+    @staticmethod
+    def get_rc(meta: StreamMeta):
+        return meta.get(StreamMetaKey.RC)
```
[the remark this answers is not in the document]
yanchengnv: Return Code

nvflare/apis/stream_shareable.py
```diff
@@ -0,0 +1,163 @@
...
+from abc import ABC, abstractmethod
+from builtins import dict as StreamMeta
+from typing import Any, Dict, Tuple
+
+from nvflare.apis.fl_context import FLContext
+from nvflare.apis.shareable import Shareable
+
+
+class StreamMetaKey:
+    CHANNEL = "__channel__"
```
nvidianz, line 23: channel/topic are required parameters for all message layers just like destination. They should not be treated as "meta" data. "meta" in my mind is for optional attributes.

nvflare/apis/stream_shareable.py
```diff
@@ -0,0 +1,163 @@
...
+    TOPIC = "__topic__"
+    RC = "__RC__"
+
+
+class StreamShareableGenerator(ABC):
+    @abstractmethod
+    def get_next(
```
nvidianz, line 30: This is the iterator pattern. We can use Iterator interface and pass the parameters in constructor.

nvflare/apis/stream_shareable.py
```diff
@@ -0,0 +1,163 @@
...
+
+        Returns: None
+
+        This method is guaranteed to be called at the end of streaming.
+
+        """
+        pass
+
+
+class StreamShareableProcessorFactory(ABC):
```
nvidianz, line 114: StreamShareableProcessorPool maybe a better name for this.

nvflare/private/fed/server/server_engine.py
```diff
@@ -570,17 +571,17 @@ def _get_aux_msg_target(self, name: str):
         else:
             return None
 
-    def send_aux_to_targets(self, targets, topic, request, timeout, fl_ctx, optional, secure):
+    def _to_aux_msg_targets(self, target_names: List[str]):
```
nvidianz, line 574: Shall we rename aux_msg to stream_msg?

nvidianz (commented):
Just a few name change suggestions.

nvflare/apis/stream_shareable.py
yanchengnv: It should be considered as "streaming context". It contains contextual information about the end-to-end streaming process. Channel and Topic are required.

I'll change to use the term "StreamContext", instead of StreamMeta.

nvflare/apis/stream_shareable.py
yanchengnv: It's not necessarily the iterator pattern since it could be called from multiple threads (when/if we implement the streaming with multiple threads) at the same time.

To avoid confusion, will change the class name to ObjectProducer, and the method name to "produce".

nvflare/apis/stream_shareable.py
yanchengnv: Will change to ConsumerFactory, which produces an ObjectConsumer object when called.

nvflare/private/fed/server/server_engine.py
yanchengnv: No, this is a utility method that turns target names to target FQCNs. This code used to be part of the aux messaging, and I refactored it so it can be used for both aux messaging and aux-msg based streaming.

nvidianz (approved):

PR #3073, opened 2024-11-25
Fix controller sequential relay test

YuanTingHsieh:
Adjust the timing to produce stable result and add comments to the tests

### Description

Adjust the timing to produce stable result and add comments to the tests

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/tests/unit_test/apis/impl/controller_test.py
+++ b/tests/unit_test/apis/impl/controller_test.py
@@ -1317,16 +1317,27 @@ def _process_task_request_test_cases():
     clients = [create_client(f"__test_client{i}") for i in range(3)]
     client_names = [c.name for c in clients]
 
+    # For each task_assignment_timeout pass by, the "window" grow by one
+    #   all the targets within the window can receive a task
     dynamic_targets_cases = [
+        # For dynamic window, the new target will be appended to the end of the list.
         (clients[1:], clients[0], True, 2, 0, False, [clients[1].name, clients[2].name, clients[0].name]),
+        # test_client1 is the first, is in target and is in window so it gets task
         (clients[1:], clients[1], True, 2, 0, True, client_names[1:]),
+        # test_client2 is the second, is in target BUT not in window so it does not get task
         (clients[1:], clients[2], True, 2, 0, False, client_names[1:]),
+        # test_client1 is not in target AND not in window so it does not get task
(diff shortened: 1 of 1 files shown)


tests/unit_test/apis/impl/controller_test.py
```diff
@@ -1317,16 +1317,27 @@ def _process_task_request_test_cases():
...
         (clients[1:], clients[0], True, 2, 0, False, [clients[1].name, clients[2].name, clients[0].name]),
+        # test_client1 is the first, is in target and is in window so it gets task
         (clients[1:], clients[1], True, 2, 0, True, client_names[1:]),
+        # test_client2 is the second, is in target BUT not in window so it does not get task
         (clients[1:], clients[2], True, 2, 0, False, client_names[1:]),
+        # test_client1 is not in target AND not in window so it does not get task
         ([clients[0]], clients[1], True, 2, 0, False, [clients[0].name, clients[1].name]),
-        ([clients[0]], clients[1], True, 1, 2, False, [clients[0].name]),
+        # the "time_before_first_request" takes too long the SequentialRelayTaskManager times out
+        ([clients[0]], clients[1], True, 1, 5, False, [clients[0].name]),
```
SYangster, line 1332: Good comments. So before `time_before_first_request` (2) would only occasionally timeout?

SYangster (approved):

tests/unit_test/apis/impl/controller_test.py
YuanTingHsieh: yes, it ties to how to determine whether a task has timed out.
The logic to determine the window: 
```
        if task_assignment_timeout:
            win_size = int((time.time() - win_start_time) / task_assignment_timeout) + 1
        else:
            win_size = 1

        self.logger.debug("win_size={}".format(win_size))
        win_end_idx = win_start_idx + win_size

        # Should exit if win extends past the entire target list + 1
        if task_assignment_timeout and win_end_idx > num_targets + 1:
            return -1, 0
```
At the time we make the request using "clients[1]" at 2 seconds (or approximate that time)
If win_size > 3 then we will timed out and get no task.
If win_size <=3 then we will get a task.
2 seconds is tricky as if too much then it is >3, if less than <=3, yields different result based on timing.
That's why sometimes unit test failed sometime passed.

We change this value to be farther away to allow some delay / additional time in this process to ensure reproducible result every time.

PR #3057, opened 2024-11-06
Add ability for clients to send error log to server

nvkevlu:
Renames the error log file to error_log.txt and adds a way to transmit the error logs from clients to the server.

### Description

Renames the error log file to error_log.txt and adds a way to transmit the error logs from clients to the server. Currently should_report_error_log is set to False. This can be set to True to test the feature for now, and in the future we will want to add a way to configure this for each site.

On the server, the error log is simply printed, but future PRs will modify this to save the error log in the proper place.

### Types of changes

- [x] Non-breaking change (fix or new feature that would not break existing functionality).
- [ ] Breaking change (fix or new feature that would cause existing functionality to change).
- [ ] New tests added to cover the changes.
- [ ] Quick tests passed locally by running `./runtest.sh`.
- [ ] In-line docstrings updated.
- [ ] Documentation updated.

PR #3075, opened 2024-11-26
Fix auth ls command response

YuanTingHsieh:
Follows the recent logging changes #3057 , the output of "ls" command changed.

### Description

Changed output accordingly

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/tests/integration_test/data/test_configs/authorization/shell_commands.yml
+++ b/tests/integration_test/data/test_configs/authorization/shell_commands.yml
@@ -14,7 +14,7 @@ tests:
         "admin_user_name": "admin@a.org"
         "result":
           "type": "admin_api_response"
-          "data": "audit.log daemon_pid.fl local log.txt pid.fl readme.txt startup transfer"
+          "data": "audit.log daemon_pid.fl error.log error_log.txt local log.txt pid.fl readme.txt startup transfer"
       - "trigger":
           "type": "server_log"
           "data": "Server started"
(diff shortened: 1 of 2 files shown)


tests/integration_test/data/test_configs/authorization/shell_commands.yml
```diff
@@ -14,7 +14,7 @@ tests:
         "admin_user_name": "admin@a.org"
         "result":
           "type": "admin_api_response"
-          "data": "audit.log daemon_pid.fl local log.txt pid.fl readme.txt startup transfer"
+          "data": "audit.log daemon_pid.fl error.log error_log.txt local log.txt pid.fl readme.txt startup transfer"
```
SYangster, line 17: Based on https://github.com/NVIDIA/NVFlare/pull/3057, error.log was renamed to error_log.txt. The master_template.yml still needs to be changed back right @nvkevlu?https://github.com/NVIDIA/NVFlare/blob/e55c523612fa9cab56e448d72520254305cd7573/nvflare/lighter/impl/master_template.yml#L329

Then we won't have both error.log and error_log.txt here.

nvkevlu: Correct, we should change master_template.yml to be error_log.txt instead of error.log so we won't need error.log here

nvkevlu:
Thank you for the fix YuanTing, do you think it would make sense to also make the change of error.log to error_log.txt in master_template.yml in this PR and remove error.log from the output?

nvkevlu (approved):
Thanks, LGTM

PR #3077, opened 2024-11-26
Add storage capability for client logs and allow for use with LogSender and LogReceiver

nvkevlu:
Add storage capability for client logs and allow for use with LogSender and LogReceiver.

### Description

To use this, add this component to the resources on the client site:
{
  "id": "log_sender",
  "path": "nvflare.app_common.widgets.log_streaming.LogSender",
  "args": {}
},

and with this configured on the server:
{
    "id": "log_receiver",
    "path": "nvflare.app_common.widgets.log_streaming.LogReceiver",
    "args": {}
},

### Types of changes

- [ ] Documentation updated.


nvflare/apis/dxo.py
```diff
@@ -29,7 +29,7 @@ class DataKind(object):
     COLLECTION = "COLLECTION"  # Dict or List of DXO objects
     STATISTICS = "STATISTICS"
     PSI = "PSI"
-    APP_DEFINED = "APP_DEFINED"  # data format is app defined
+    APP_DEFINED = "APP_DEFINED"  # data format is app defined=
```
YuanTingHsieh, line 32: "=" is a typo?

nvflare/apis/impl/job_def_manager.py
```diff
@@ -237,6 +236,34 @@ def get_content(self, meta: dict, fl_ctx: FLContext) -> Optional[bytes]:
...
+        self._save_log(jid, log, client_name, "ERRORLOG", fl_ctx)
+
+    def set_log(self, jid: str, log: str, client_name: str, fl_ctx: FLContext):
+        self._save_log(jid, log, client_name, "LOG", fl_ctx)
+
+    def get_error_log(self, jid: str, client_name: str, fl_ctx: FLContext) -> Optional[str]:
+        return self._get_log(jid, client_name, "ERRORLOG", fl_ctx)
+
+    def get_client_log(self, jid: str, client_name: str, fl_ctx: FLContext) -> Optional[str]:
+        return self._get_log(jid, client_name, "LOG", fl_ctx)
```
YuanTingHsieh, line 265: do we need to have separate methods? or we could also do set_log with additional argument LOG_TYPE or LOG_FILE?

nvflare/apis/storage.py
```diff
@@ -217,4 +217,9 @@ def tag_object(self, uri: str, tag: str, data=None):
 
     @staticmethod
     def is_valid_component(component_name):
-        return component_name in [DATA, META, WORKSPACE]
+        valid_components = {DATA, META, WORKSPACE, "ERRORLOG", "LOG"}
```
YuanTingHsieh, line 220: For "LOG" and "ERRORLOG" do we want to define some constant variable for these strings, and could there be more log types?

nvflare/app_common/widgets/log_streaming.py
```diff
@@ -0,0 +1,105 @@
...
+        file_location = FileStreamer.get_file_location(stream_ctx)
+        self.log_info(fl_ctx, f"File location: {file_location}")
+        with open(file_location, "r") as f:
+            log_contents = f.read()
+        client = stream_ctx.get(LogConst.CLIENT_NAME)
+        job_id = stream_ctx.get(LogConst.JOB_ID)
+        job_manager = fl_ctx.get_engine().get_component(SystemComponents.JOB_MANAGER)
+        self.log_info(fl_ctx, f"Saving ERRORLOG from {client} for {job_id}")
+        job_manager.set_error_log(job_id, log_contents, client, fl_ctx)
+
```
YuanTingHsieh, line 100: Does this have to be "ERROR_LOGS"?  could we get this type either from stream_ctx or peer_ctx?

nvflare/app_common/widgets/log_streaming.py
```diff
@@ -0,0 +1,105 @@
...
+    def handle_event(self, event_type: str, fl_ctx: FLContext):
+        if event_type == EventType.SYSTEM_START:
+            FileStreamer.register_stream_processing(
+                fl_ctx, channel="error_logs", topic=LOG_STREAM_EVENT_TYPE, stream_done_cb=self.process_log
+            )
```
YuanTingHsieh, line 105: can this be configurable? will there be another case where users want more than just ERROR_LOG

nvflare/apis/dxo.py
nvkevlu: yes, I will fix this

nvflare/apis/impl/job_def_manager.py
nvkevlu: I think we can refactor this to have an argument, I'll make the change

nvflare/apis/storage.py
nvkevlu: Yes, this is just to put something for now, and we can certainly add more, so I will create a variable for this.

nvkevlu:
Refactored to make it more configurable

nvflare/apis/impl/job_def_manager.py
```diff
@@ -237,6 +236,28 @@ def get_content(self, meta: dict, fl_ctx: FLContext) -> Optional[bytes]:
...
+        store = self._get_job_store(fl_ctx)
+        log_object_type = f"{log_type}_{client_name}"
+        try:
+            log_data = store.get_data(self.job_uri(jid), log_object_type)
+            return log_data.decode()
+        except StorageException:
+            return None
+
+    def set_log(self, jid: str, log: str, client_name: str, log_type: str, fl_ctx: FLContext):
+        self._save_log(jid, log, client_name, log_type, fl_ctx)
```
YuanTingHsieh, line 256: suggest "set_client_log", "_save_client_log", "_get_client_log" if "client_name" is a required argument

nvflare/app_common/widgets/log_streaming.py
```diff
@@ -0,0 +1,117 @@
+# Copyright (c) 2021, NVIDIA CORPORATION.  All rights reserved.
```
YuanTingHsieh, line 1: 2024

nvflare/app_common/widgets/log_streaming.py
```diff
@@ -0,0 +1,117 @@
...
+        By default, it receives error logs with the channel error_logs to save logs of type ERRORLOG. If adding
+        additional log types, make sure nvflare.apis.storage.ComponentPrefixes has the corresponding log type.
+
+        Args:
+            log_channels: dict of channel to log type mapping
+        """
+        super().__init__()
+        if log_channels is None:
+            log_channels = {"error_logs": "ERRORLOG"}
+        self.log_channels = log_channels
```
YuanTingHsieh, line 85: suggest adding a check here to see if log_channels is a dict

nvflare/app_common/widgets/log_streaming.py
```diff
@@ -0,0 +1,117 @@
...
+            )
+            return
+        file_location = FileStreamer.get_file_location(stream_ctx)
+        self.log_info(fl_ctx, f"File location: {file_location}")
+        with open(file_location, "r") as f:
+            log_contents = f.read()
+        client = stream_ctx.get(LogConst.CLIENT_NAME)
+        job_id = stream_ctx.get(LogConst.JOB_ID)
+        job_manager = fl_ctx.get_engine().get_component(SystemComponents.JOB_MANAGER)
+        log_type = self.log_channels.get(channel, "UNKNOWN")
```
YuanTingHsieh, line 108: I think if user specify something and this channel we get from peer is not matching, we should ignore it

nvflare/app_common/widgets/log_streaming.py
```diff
@@ -0,0 +1,117 @@
...
+        self.log_info(fl_ctx, f"Saving {log_type} from {client} for {job_id}")
+        job_manager.set_log(job_id, log_contents, client, log_type, fl_ctx)
+
+    def handle_event(self, event_type: str, fl_ctx: FLContext):
+        if event_type == EventType.SYSTEM_START:
+            for channel in self.log_channels.items():
```
YuanTingHsieh, line 114: items return a tuple, do we want tuple here? or we want keys or values?

nvflare/app_common/widgets/log_streaming.py
```diff
@@ -0,0 +1,117 @@
...
+        client_name = fl_ctx.get_prop(FLContextKey.CLIENT_NAME)
+        job_id = fl_ctx.get_prop(FLContextKey.CURRENT_JOB_ID)
+        workspace_object = Workspace(root_dir=workspace_root, site_name=client_name)
+        error_log_path = workspace_object.get_app_error_log_file_path(job_id=job_id)
+        if os.path.exists(error_log_path):
+            with open(error_log_path, "r") as f:
+                error_log_contents = f.read()
+        if error_log_contents:
+            FileStreamer.stream_file(
+                channel="error_logs",
```
YuanTingHsieh, line 54: could this be configurable? like an arg in init method?

nvflare/app_common/widgets/log_streaming.py
```diff
@@ -0,0 +1,117 @@
...
+        log_type = self.log_channels.get(channel, "UNKNOWN")
+        self.log_info(fl_ctx, f"Saving {log_type} from {client} for {job_id}")
+        job_manager.set_log(job_id, log_contents, client, log_type, fl_ctx)
+
+    def handle_event(self, event_type: str, fl_ctx: FLContext):
+        if event_type == EventType.SYSTEM_START:
+            for channel in self.log_channels.items():
+                FileStreamer.register_stream_processing(
+                    fl_ctx, channel=channel, topic=LOG_STREAM_EVENT_TYPE, stream_done_cb=self.process_log
+                )
```
YuanTingHsieh, line 117: We could also say this LogSender and LogReceiver is always using the same channel but for different types of logs they send/receive they use different "topic"?

Now is the same "topic" but different "channel" I guess we just need to explain what we prefer in the docstring

nvflare/app_common/widgets/log_streaming.py
```diff
@@ -0,0 +1,117 @@
...
+LOG_STREAM_EVENT_TYPE = "stream_log"
+
+
+class LogConst(object):
+    CLIENT_NAME = "client_name"
+    JOB_ID = "job_id"
+    LOG_DATA = "log_data"
+
+
+class LogSender(Widget):
```
YuanTingHsieh, line 35: since this component only works in client, what happen if users put this in server side?
Do we need some checking against that?

YuanTingHsieh (commented):
Mostly good, added some questions

nvflare/app_common/widgets/log_streaming.py
nvkevlu: I accidentally removed the variable thinking it was unused, but it seems it was needed for proper unpacking, I'll put it back

nvflare/apis/impl/job_def_manager.py
```diff
@@ -237,6 +236,20 @@ def get_content(self, meta: dict, fl_ctx: FLContext) -> Optional[bytes]:
         except StorageException:
             return None
 
+    def set_client_log(self, jid: str, log: str, client_name: str, log_type: str, fl_ctx: FLContext):
+        store = self._get_job_store(fl_ctx)
+        log_object_type = f"{log_type}_{client_name}"
+        store.update_object(self.job_uri(jid), log.encode(), log_object_type)
```
yanchengnv, line 242: Don't need log.encode(). The caller is responsible for converting to bytes.

nvflare/apis/impl/job_def_manager.py
```diff
@@ -237,6 +236,20 @@ def get_content(self, meta: dict, fl_ctx: FLContext) -> Optional[bytes]:
...
+
+    def get_client_log(self, jid: str, client_name: str, log_type: str, fl_ctx: FLContext) -> Optional[str]:
```
yanchengnv, line 244: Rename this method properly.

nvflare/apis/impl/job_def_manager.py
```diff
@@ -237,6 +236,20 @@ def get_content(self, meta: dict, fl_ctx: FLContext) -> Optional[bytes]:
...
+        store = self._get_job_store(fl_ctx)
+        log_object_type = f"{log_type}_{client_name}"
+        try:
+            log_data = store.get_data(self.job_uri(jid), log_object_type)
+            return log_data.decode()
```
yanchengnv, line 249: Should return bytes. Let caller take care of convertion.

nvflare/apis/job_def_manager_spec.py
```diff
@@ -122,6 +122,36 @@ def refresh_meta(self, job: Job, meta_keys: list, fl_ctx: FLContext):
         """
         pass
 
+    @abstractmethod
+    def set_client_log(self, jid: str, log: str, client_name: str, log_type: str, fl_ctx: FLContext):
```
yanchengnv, line 126: Rename method name to be more generic.

nvflare/apis/job_def_manager_spec.py
```diff
@@ -122,6 +122,36 @@ def refresh_meta(self, job: Job, meta_keys: list, fl_ctx: FLContext):
...
+            log (str): log content
+            client_name (str): client name
+            log_type (str): log type
+            fl_ctx (FLContext): FLContext information
+
+        """
+        pass
+
+    @abstractmethod
+    def get_client_log(self, jid: str, client_name: str, log_type: str, fl_ctx: FLContext) -> Optional[str]:
```
yanchengnv, line 140: Rename method name to be more generic.

nvflare/app_common/widgets/log_streaming.py
yanchengnv: You can do this now by checking the fl_ctx.get_process_type() to be ProcessType.CLIENT_PARENT.
Do this in handle_event.

nvflare/app_common/widgets/log_streaming.py
yanchengnv: I think it's okay to hard-code channel name. But use some prefix to make it less likely to conflict with others, something like log_streaming.error_logs

nvflare/app_common/widgets/log_streaming.py
```diff
@@ -0,0 +1,122 @@
...
+    def _stream_error_log_file(self, fl_ctx: FLContext):
+        error_log_contents = None
+        workspace_root = fl_ctx.get_prop(FLContextKey.WORKSPACE_ROOT)
+        client_name = fl_ctx.get_prop(FLContextKey.CLIENT_NAME)
+        job_id = fl_ctx.get_prop(FLContextKey.CURRENT_JOB_ID)
+        workspace_object = Workspace(root_dir=workspace_root, site_name=client_name)
+        error_log_path = workspace_object.get_app_error_log_file_path(job_id=job_id)
+        if os.path.exists(error_log_path):
+            with open(error_log_path, "r") as f:
+                error_log_contents = f.read()
```
yanchengnv, line 51: No, this will load the file content to memory, which defeats the purpose of streaming!
You can check the file size to know whether it has any content.

nvflare/app_common/widgets/log_streaming.py
```diff
@@ -0,0 +1,122 @@
...
+                stream_ctx={LogConst.CLIENT_NAME: client_name, LogConst.JOB_ID: job_id},
+                targets=["server"],
+                file_name=error_log_path,
+                fl_ctx=fl_ctx,
+            )
+
+    def handle_event(self, event_type: str, fl_ctx: FLContext):
+        if event_type == self.event_type:
+            if self.should_report_error_log:
+                t = threading.Thread(target=self._stream_error_log_file, args=(fl_ctx,), daemon=True)
```
yanchengnv, line 65: You should check the file existence and content before creating this thread: move the logic from _stream_error_log_file to here, so we can avoid unnecessary thread.

nvflare/app_common/widgets/log_streaming.py
```diff
@@ -0,0 +1,122 @@
...
+
+        By default, it receives error logs with the channel error_logs to save logs of type ERRORLOG. If adding
+        additional log types, make sure nvflare.apis.storage.ComponentPrefixes has the corresponding log type.
+
+        Args:
+            log_channels: dict of channel to log type mapping
+        """
+        super().__init__()
+        if log_channels is None:
+            log_channels = {"error_logs": "ERRORLOG"}
```
yanchengnv, line 84: Please define some symbols for the channel name and log type, instead of hard-coding.

nvflare/app_common/widgets/log_streaming.py
yanchengnv: You should use "*" to be the topic such that all logs will be handled by the same CB.
Let's use the same predefined channel for all log streaming. Topic is not important here, since any log specific info will be in the stream_ctx.

nvflare/app_common/widgets/log_streaming.py
```diff
@@ -0,0 +1,122 @@
...
+        rc = FileStreamer.get_rc(stream_ctx)
+        if rc != ReturnCode.OK:
+            self.log_error(
+                fl_ctx,
+                f"Error in streaming log file from {peer_name} on channel {channel} and topic {topic} with rc {rc}",
+            )
+            return
+        file_location = FileStreamer.get_file_location(stream_ctx)
+        self.log_info(fl_ctx, f"File location: {file_location}")
+        with open(file_location, "r") as f:
```
yanchengnv, line 105: No, please do not read the file into memory.

nvflare/app_common/widgets/log_streaming.py
```diff
@@ -0,0 +1,122 @@
...
+            log_contents = f.read()
+        client = stream_ctx.get(LogConst.CLIENT_NAME)
+        job_id = stream_ctx.get(LogConst.JOB_ID)
+        job_manager = fl_ctx.get_engine().get_component(SystemComponents.JOB_MANAGER)
+        log_type = self.log_channels.get(channel, "UNKNOWN")
+        if log_type == "UNKNOWN":
+            self.log_error(fl_ctx, f"Unknown log type for channel {channel}")
+        else:
+            self.log_info(fl_ctx, f"Saving {log_type} from {client} for {job_id}")
+            job_manager.set_client_log(job_id, log_contents, client, log_type, fl_ctx)
```
yanchengnv, line 115: Make this function take both bytes and a file name for log_contents. The storage's update_object already supports file.

yanchengnv (changes requested):
See my comments. The main thing here is to avoid reading file content into memory, on both the sending and receiving sides.

nvflare/app_common/widgets/log_streaming.py
```diff
@@ -0,0 +1,114 @@
+# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
```
yanchengnv, line 1: I thought we talked about splitting this into 3 separate files:
- A common file that contains the definition of channel and other constants, and potentially a base class for all senders.
- A file that implements error log sender;
- A file that implements the receiver.

Please follow this structure.

yanchengnv (changes requested):
Please re-organize file structure as we discussed before. See comments.

nvflare/app_common/logging/log_receiver.py
```diff
@@ -0,0 +1,58 @@
...
+class LogReceiver(Widget):
+    def __init__(self):
+        """Receives log data.
+
+        If adding additional log types, make sure nvflare.apis.storage.ComponentPrefixes has
+        the corresponding log type.
+        """
+        super().__init__()
+
+    def process_log(self, stream_ctx: dict, fl_ctx: FLContext):
```
yanchengnv, line 32: Change type hint here: stream_ctx: dict to. stream_ctx: StreamContext

You need to:
from nvflare.api.streaming import StreamContext

nvflare/app_common/logging/log_sender.py
```diff
@@ -0,0 +1,68 @@
...
+from nvflare.apis.fl_constant import FLContextKey, ProcessType, StreamCtxKey
+from nvflare.apis.fl_context import FLContext
+from nvflare.apis.storage import DataTypes
+from nvflare.apis.workspace import Workspace
+from nvflare.app_common.logging.constants import LOG_STREAM_EVENT_TYPE, Channels
+from nvflare.app_common.streamers.file_streamer import FileStreamer
+from nvflare.widgets.widget import Widget
+
+
+class LogSender(Widget):
```
yanchengnv, line 28: Please change to ErrorLogSender since it only works for error logs.
You could refactor this later to make some part of the code reusable for other type of logging.

yanchengnv (changes requested):
Looks good in general. See my comments for cosmetic fixes.

yanchengnv (approved):
Looks good.

PR #3085, opened 2024-11-28
Implement a new algorithm for the CUDA plugin

YuanTingHsieh:
The original algorithm looping through every bin and for each bin calls the reduce_sum on all the samples that belong to this bin, the number of total samples denoted as N.
The total number of bins can be calculated by F x B.
When the total number of bins  (FB) is huge, the original algorithm will launch F x B CUDA kernels, which is very slow.
The complexity is roughly FB log(N) + FB*C (C is the overhead to launch a CUDA kernel)
So we need to enhance it.

The new algorithm will first layout all the numbers to be added, and each kernel launch will reduce the number by half.
Thus we only need to launch kernel log(F x N) times.
The log(FxN) is much smaller then F x B.

### Description

- Implement a new algorithm for the CUDA plugin
- Enhance the LocalPlugin interface to reduce unnecessary conversion for the new CUDA plugin

### Types of changes

- [ ] Documentation updated.


integration/xgboost/encryption_plugins/v.py
```diff
@@ -0,0 +1,132 @@
+#!/usr/bin/python
```
ZiyueXu77, line 1: Add license header

integration/xgboost/encryption_plugins/cuda_plugin/src/cuda_utils.h
```diff
@@ -45,7 +46,6 @@ const static unsigned int key_len=1024;
 //const static unsigned int bits=6144;
 //const static unsigned int key_len=3072;
 
-
 const static int TPB=512;
```
ZiyueXu77, line 49: are these two needed (not sure their meaning)? maybe also remove the unused lines

ZiyueXu77: ah I see they are used later in paillier.h, maybe move to there if that's the only place using this?

integration/xgboost/encryption_plugins/v.py
ZiyueXu77: for this testing script, maybe we move it to the example folder?

ZiyueXu77 (commented):
some initial comments

integration/xgboost/encryption_plugins/v.py
YuanTingHsieh: yes, I should remove it

integration/xgboost/encryption_plugins/cuda_plugin/src/cuda_utils.h
YuanTingHsieh: it's also used in cuda_plugin.h, so keep it here for now.

integration/xgboost/encryption_plugins/cuda_plugin/src/cuda_plugin.h
```diff
@@ -56,16 +120,28 @@ class CUDAPlugin: public LocalPlugin {
...
 
     void setGHPairs() {
       if (debug_) std::cout << "setGHPairs is called" << std::endl;
       const std::uint8_t* pointer = encrypted_gh_.data();
 
       // Retrieve header
-      BufferHeader header;
-      std::memcpy(&header, pointer, sizeof(BufferHeader));
-      pointer += sizeof(BufferHeader);
+      BufferHeader2 header;
```
nvidianz, line 142: Should just use BufferHeader

YuanTingHsieh: good catch! will fix

yanchengnv:
@nvidianz and @YuanTingHsieh please get this PR merged before YT goes to TW.

nvidianz (approved):
The new names are much better.

PR #3087, opened 2024-12-02
Fix poc command

yanchengnv:
Fixes # .

### Description

Previous change to Provision broke the "poc start" command behavior: the user is prompted for username.  This PR fixes this.

The fix now allows a project to be provisioned in a specified "mode".  If the mode is not specified, then the project is provisioned in normal mode. If the mode is set to "poc", then when creating fed_admin.json (in StaticFileBuilder), the user's name will be added to the file, and the cred_type is set to "local_cert". 

The provision mode is set into the provision context (the "ctx"), which will allow any builder to do mode-specific processing.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/lighter/impl/static_file.py
+++ b/nvflare/lighter/impl/static_file.py
@@ -386,6 +386,13 @@ class StaticFileBuilder(Builder):
         agent_config = dict()
         self._prepare_overseer_agent(admin, agent_config, "admin", ctx)
         config["admin"].update(agent_config)
+
+        provision_mode = ctx.get("provision_mode")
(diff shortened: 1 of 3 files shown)


yanchengnv:
> overall looks good, still have the dummy_project.yml contains
> 
> ```
> sp_end_point: server1:8002:8003
> ```
> 
> should the server1 be changed "localhost" or we should have a function to change that ?

No, it's not necessary. This is changed dynamically to "localhost" since we set the "default_host" to "localhost".
This part is done in order to support multiple host names and IP addresses.

PR #3092, opened 2024-12-04
Refactor provision for general use - Part 1

yanchengnv:
Fixes # .

### Description

Today, CLI-based provision and Dashboard provision are two separate systems with similar code in both. This makes it very hard to add and improve provision features.

This PR refactors existing code into a common solution that can be used for both CLI and Dashboard based provision.

Here are the major changes:

- Improved code structure to make it more readable and maintainable
- Improved provision API for general-purpose use
- Move commonly used property values into formally defined symbols and replace hard-coded string values with symbol references. This helps avoid misspelling, avoid duplicate definitions, and provide a holistic view of the properties.
- Ensure only one server is allowed. Multi-server was for HA, but the current HA implementation is flawed and not used by any customers. 
- Dashboard and CLI currently use different ways to manage templates. This PR consolidates to one solution. 
- Refactored all builders to use the new API.
- Modified and added unit test cases for the new API.

NOTE: this PR only modified CLI-based provision to use this new solution. Dashboard provision will be changed in another PR.

### Types of changes

- [ ] Documentation updated.


nvflare/lighter/impl/signature.py
```diff
@@ -26,23 +27,31 @@ class SignatureBuilder(Builder):
     can be cryptographically verified to ensure any tampering is detected. This builder writes the signature.json file.
     """
 
-    def _do_sign(self, root_pri_key, dest_dir):
+    @staticmethod
+    def _do_sign(root_pri_key, dest_dir):
         signatures = sign_all(dest_dir, root_pri_key)
-        json.dump(signatures, open(os.path.join(dest_dir, "signature.json"), "wt"))
+        with open(os.path.join(dest_dir, ProvFileName.SIGNATURE_JSON), "wt") as f:
```
YuanTingHsieh, line 33: some of the constants only used within a specific builder, do we still want to move them into the ProvFileName?

nvflare/lighter/impl/workspace.py
```diff
@@ -43,50 +44,56 @@ def __init__(self, template_file):
...
+        temp_files_to_load = self.template_files
+        if not temp_files_to_load:
+            # load everything
+            temp_files_to_load = [f for f in os.listdir(temp_folder) if os.path.isfile(f)]
+        elif isinstance(temp_files_to_load, str):
+            temp_files_to_load = [temp_files_to_load]
+
+        template = dict()
+        for f in temp_files_to_load:
+            template.update(load_yaml(os.path.join(temp_folder, f)))
```
YuanTingHsieh, line 64: depends on the order of "temp_files_to_load", the latter file might overwrite the previous file if they have same keys

nvflare/lighter/impl/workspace.py
```diff
@@ -43,50 +44,56 @@ def __init__(self, template_file):
...
+        make_dirs(dirs)
+
+        dirs = [ctx.get_transfer_dir(p) for p in participants]
+        make_dirs(dirs)
+
+        dirs = [ctx.get_local_dir(p) for p in participants]
+        make_dirs(dirs)
+
+    def finalize(self, project: Project, ctx: ProvisionContext):
+        if ctx[CtxKey.LAST_PROD_STAGE] >= 99:
```
YuanTingHsieh, line 90: although we never exceed this number, why we need this 99 restriction?

nvflare/lighter/impl/static_file.py
```diff
@@ -220,93 +195,82 @@ def _build_client(self, client, ctx):
...
+    def _prepare_overseer_agent(self, participant, config, role, ctx: ProvisionContext):
+        project = ctx.get_project()
         server = project.get_server()
         if not server:
             raise ValueError(f"Missing server definition in project {project.name}")
 
-        fl_port = server.get_prop("fed_learn_port", 8002)
-        admin_port = server.get_prop("admin_port", 8003)
+        fl_port = server.get_prop(PropKey.FED_LEARN_PORT, 8002)
+        admin_port = server.get_prop(PropKey.ADMIN_PORT, 8003)
```
YuanTingHsieh, line 264: should we get the ports using context?

`ctx.get(CtxKey.ADMIN_PORT)`

nvflare/lighter/templates/aws_template.yml
```diff
@@ -0,0 +1,437 @@
+aws_start_sh: |
```
YuanTingHsieh, line 1: if we move these files into "templates" we should remove them in "impl"?

YuanTingHsieh (commented):
Changes are good, introducing "Entity", "Project" concept and separate "Builder" into "Builder" and "ProvisionContext" let user focus on Builder API without worry other stuffs.

tests/unit_test/lighter/project_test.py
```diff
@@ -17,28 +17,71 @@
...
 class TestProject:
-    def test_invalid_project(self):
-        p1 = create_participants("server", 3, "org", "server")
-        p2 = create_participants("server", 3, "org", "server")
-        p = p1 + p2
-        with pytest.raises(ValueError, match=r".* se0rver .*"):
+    def test_single_server(self):
+        p1 = Participant(name="server1", org="org", type="server")
+        p2 = Participant(name="server2", org="org", type="server")
+        with pytest.raises(ValueError, match=r".* already has server defined"):
```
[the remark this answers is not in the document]
yanchengnv: sure

tests/unit_test/lighter/project_test.py
```diff
@@ -17,28 +17,71 @@
...
+            _ = Project("name", "description", [p1, p2])
+
+    def test_single_overseer(self):
+        p1 = Participant(name="name1", org="org", type="overseer")
+        p2 = Participant(name="name2", org="org", type="overseer")
+        with pytest.raises(ValueError, match=r".* already has overseer defined"):
```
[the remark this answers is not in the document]
yanchengnv: yes

nvflare/lighter/spec.py
```diff
@@ -119,89 +138,239 @@ def get_default_host(self) -> str:
...
-        return found
+                    raise ValueError(f"invalid value for ParticipantType: {p.type}")
+
+    def _check_unique_name(self, name: str):
+        if name in self.all_names:
+            raise ValueError(f"the project {self.name} already has a participant with the name '{name}'")
+
+    def set_server(self, name: str, org: str, props: dict):
+        if self.server:
+            raise ValueError(f"project {self.name} already has server defined")
```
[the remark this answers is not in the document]
yanchengnv: yes

nvflare/lighter/spec.py
```diff
@@ -119,89 +138,239 @@ def get_default_host(self) -> str:
...
 
         Returns: server participant
 
         """
-        return self.get_participants_by_type("server", first_only=True)
+        return self.server
 
+    def set_overseer(self, name: str, org: str, props: dict):
+        if self.overseer:
+            raise ValueError(f"project {self.name} already has overseer defined")
```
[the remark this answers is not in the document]
yanchengnv: yes

nvflare/lighter/spec.py
```diff
@@ -119,89 +138,239 @@ def get_default_host(self) -> str:
...
+    def finalize(self, project: Project, ctx: ProvisionContext):
+        pass
+
+    @staticmethod
+    def build_from_template(
+        ctx: ProvisionContext, dest_dir: str, temp_section: str, file_name, replacement=None, mode="t", exe=False
+    ):
+        section = ctx.get_template_section(temp_section)
+        if replacement:
+            section = utils.sh_replace(section, replacement)
```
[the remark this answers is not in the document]
yanchengnv: This is from existing code that is out of the scope of this refactoring. It seems to be generic, but need to verify with Isaac before changing.

nvflare/lighter/impl/signature.py
yanchengnv: It depends. If the constant is only used within the builder, we usually don't need to move. BUT for provisioned file names are moved into a central place to give us a holistic view of what files are generated by the provision. This could make it a little easier for maintenance.

nvflare/lighter/impl/workspace.py
yanchengnv: That's why I at least move all section keys into the constants so dups can be detected.
Currently, the 3 temp files don't have conflicts. But in future, we should introduce naming convention. For example, all keys in a file should start with a unique prefix.

nvflare/lighter/impl/workspace.py
yanchengnv: Correct. We could improve more later.

nvflare/lighter/impl/static_file.py
yanchengnv: Yes, that's the intent, but I missed this one.

nvflare/lighter/templates/aws_template.yml
yanchengnv: Yes, will do this in Part 2 of the PR. This is because the Dashboard is using this file now.

nvflare/lighter/spec.py
```diff
@@ -230,51 +399,46 @@ def __init__(self, root_dir: str, builders: List[Builder]):
...
-        resources_dir = os.path.join(workspace, "resources")
-        ctx.update(dict(wip_dir=wip_dir, state_dir=state_dir, resources_dir=resources_dir))
-        dirs = [workspace, resources_dir, wip_dir, state_dir]
-        self._make_dir(dirs)
+        self.template = {}
+
+    def add_template(self, template: dict):
+        if not isinstance(template, dict):
+            raise ValueError(f"template must be a dict but got {type(template)}")
+        self.template.update(template)
```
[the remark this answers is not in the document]
yanchengnv: That's why I at least move all section keys into the constants so dups can be detected.
Currently, the 3 temp files don't have conflicts. But in future, we should introduce naming convention. For example, all keys in a file should start with a unique prefix.

We already have 3 templates, and may have more (each builder could introduce its templates), and any of them could be used together.

nvflare/lighter/impl/cert.py
```diff
@@ -35,47 +105,62 @@ def __init__(self):
...
+        self.subject = None
+        self.issuer = None
 
-    def initialize(self, ctx):
-        state_dir = self.get_state_dir(ctx)
-        cert_file = os.path.join(state_dir, "cert.json")
-        if os.path.exists(cert_file):
-            self.persistent_state = json.load(open(cert_file, "rt"))
-            self.serialized_cert = self.persistent_state["root_cert"].encode("ascii")
+    def initialize(self, project: Project, ctx: ProvisionContext):
```
yhwen, line 115: When creating the ProvisionContext(), the project is already placed into the ProvisionContext instance. No need to pass in the "project" parameter any more. Better get it from the ProvisionContext. Also the concept is clearer.

nvflare/lighter/impl/cert.py
```diff
@@ -84,58 +169,63 @@ def _build_write_cert_pair(self, participant, base_name, ctx):
...
                 f.write(serialize_cert(tmp_cert))
-            with open(os.path.join(dest_dir, "server.key"), "wb") as f:
+            with open(os.path.join(dest_dir, f"{bn}.key"), "wb") as f:
                 f.write(serialize_pri_key(tmp_pri_key))
 
         with open(os.path.join(dest_dir, "rootCA.pem"), "wb") as f:
             f.write(self.serialized_cert)
 
-    def build(self, project, ctx):
+    def build(self, project: Project, ctx: ProvisionContext):
```
yhwen, line 200: same, no need to have "project".

nvflare/lighter/spec.py
```diff
@@ -11,18 +11,96 @@
...
 # limitations under the License.
-
+import json
 import os
 import shutil
 import traceback
 from abc import ABC
 from typing import List
 
+import yaml
```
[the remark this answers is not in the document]
yanchengnv: This is to follow existing file structure to minimize scope of changes.
I actually first moved them into different files, but then found out that there are too many places to change.

nvflare/lighter/constants.py
```diff
@@ -0,0 +1,138 @@
...
+    DEFAULT_HOST = "default_host"
+    PROTOCOL = "protocol"
+    API_ROOT = "api_root"
+    PORT = "port"
+    OVERSEER_END_POINT = "overseer_end_point"
+    ADMIN_PORT = "admin_port"
+    FED_LEARN_PORT = "fed_learn_port"
+
+
+class CtxKey(WorkDir, PropKey):
```
yhwen, line 48: what's the benefit of having this class hierarchy in the constants definition? To me, separate them is less confusion.

nvflare/lighter/impl/cert.py
```diff
@@ -22,9 +22,79 @@
...
 from cryptography.x509.oid import NameOID
 
-from nvflare.lighter.spec import Builder, Participant
+from nvflare.lighter.constants import CtxKey
+from nvflare.lighter.spec import Builder, Participant, ParticipantType, Project, PropKey, ProvisionContext
 from nvflare.lighter.utils import serialize_cert, serialize_pri_key
 
+_CERT_BASE_NAME_CLIENT = "client"
+_CERT_BASE_NAME_SERVER = "server"
+_CERT_BASE_NAME_OVERSEER = "overseer"
```
yhwen, line 31: maybe move them into constants.py?

nvflare/lighter/impl/cert.py
```diff
@@ -22,9 +22,79 @@
...
+        subject_data = self.content.get(subject)
+        if not subject_data:
+            subject_data = {}
+            self.content[subject] = subject_data
+        subject_data[key] = value
+
+    def _get_subject_prop(self, subject: str, key: str):
+        subject_data = self.content.get(subject)
+        if not subject_data:
+            return None
```
yhwen, line 78: no sure if it's worthy to build a default value return here, just like the Python dict get().

nvflare/lighter/utils.py
```diff
@@ -273,6 +273,12 @@ def update_storage_locations(
         outfile.write(json_object)
 
 
+def make_dirs(dirs):
+    for d in dirs:
+        if not os.path.exists(d):
+            os.makedirs(d)
```
yhwen, line 279: the same as:
       os.makedirs(d, exist_ok=True)

yhwen (commented):
Looks good. some minor items.

nvflare/lighter/utils.py
yanchengnv: Yes, got this from original code and just moved here. Could improve later.

nvflare/lighter/impl/cert.py
yanchengnv: Could be, but right now there is no use case for that.

nvflare/lighter/impl/cert.py
yanchengnv: Yes

nvflare/lighter/constants.py
yanchengnv: Actually that's because all values in WorkDir and PropKey need to be the same when they are used in ctx.

yhwen (commented):
LGTM

yhwen (approved):

PR #3096, opened 2024-12-09
[2.5] Support explicit message authentication

yanchengnv:
Fixes # .

### Description

Currently Flare's message security comes from mutual TLS: server and client authenticate each other when making connections. This means that only clients that have the right startup kits can make a connection to the server. 

The requirement of one-way SSL between the server and clients breaks this assumption: the server could be exposed to the internet and any one could write a client to connect to the server. To ensure message security, explicit message authentication is required.

This PR implements message authentication: messages received by the server must have an auth token, and the token must be validated successfully to prove that it was issued by the server!

Here is how it works:
- The client first tries to login to the server. The server/client authenticate each other explicitly with the credentials in their startup kits. This step is independent of how client/server is connected.
- If the client credential is validated correctly, the server issues a token and a signature that binds the client name and token together.  The signature is generated with the server's private key to prove that the signature can only be issued by the server.
- When sending a message to the server, the client adds the its client name, token and the signature as headers to the message.
- When the message is received, the server validates the token and the signature. Messages that are missing these headers or fail to validate will be rejected.

Note that this mechanism is based the security of the startup kits. All sites must protect their startup kits securely.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/lighter/utils.py
+++ b/nvflare/lighter/utils.py
@@ -198,7 +198,19 @@ def sh_replace(src, mapping_dict):
 
 def update_project_server_name_config(project_config: dict, old_server_name, server_name) -> dict:
     update_participant_server_name(project_config, old_server_name, server_name)
+    update_overseer_server_name(project_config, old_server_name, server_name)
     return project_config
+
+
+def update_overseer_server_name(project_config, old_server_name, server_name):
+    # update overseer_agent builder
+    builders = project_config.get("builders", [])
+    for b in builders:
+        if "args" in b:
+            if "overseer_agent" in b["args"]:
+                end_point = b["args"]["overseer_agent"]["args"]["sp_end_point"]
+                new_end_point = end_point.replace(old_server_name, server_name)
+                b["args"]["overseer_agent"]["args"]["sp_end_point"] = new_end_point
 
 
 def update_participant_server_name(project_config, old_server_name, new_server_name):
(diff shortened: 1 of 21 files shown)


nvflare/lighter/utils.py
```diff
@@ -19,36 +19,93 @@
...
+
+def load_crt(path):
+    return load_crt_bytes(open(path, "rb").read())
+
+
+def load_crt_bytes(data: bytes):
+    return x509.load_pem_x509_certificate(data, default_backend())
+
+
 def generate_password(passlen=16):
```
nvidianz, line 50: Should use `secrets` module to generate password

yanchengnv: This was from existing code. Maybe address it in 2.6.

nvflare/lighter/utils.py
```diff
@@ -19,36 +19,93 @@
...
+        encryption_algorithm=serialization.NoEncryption(),
+    )
+
+
+def serialize_cert(cert):
+    return cert.public_bytes(serialization.Encoding.PEM)
+
+
+def load_crt(path):
+    return load_crt_bytes(open(path, "rb").read())
```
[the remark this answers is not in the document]
yanchengnv: Fixed all cases.

yhwen (approved):
Tests passed.

PR #3098, opened 2024-12-10
Add BaseFedJob layer

SYangster:
Add framework agnostic BaseFedJob layer, rename current classes to PTBaseFedJob and TFBaseFedJob

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/examples/getting_started/pt/nvflare_lightning_getting_started.ipynb
+++ b/examples/getting_started/pt/nvflare_lightning_getting_started.ipynb
@@ -321,8 +321,8 @@
     "#### 2. Define a FedJob\n",
     "The `FedJob` is used to define how controllers and executors are placed within a federated job using the `to(object, target)` routine.\n",
     "\n",
-    "Here we use a PyTorch `PTBaseFedJob`, where we can define the job name and the initial global model.\n",
-    "The `PTBaseFedJob` automatically configures components for model persistence, model selection, and TensorBoard streaming for convenience."
+    "Here we use a PyTorch `PTJob`, where we can define the job name and the initial global model.\n",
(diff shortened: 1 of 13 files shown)


examples/getting_started/pt/nvflare_lightning_getting_started.ipynb
```diff
@@ -335,10 +335,10 @@
     "from src.lit_net import LitNet\n",
     "\n",
     "from nvflare.app_common.workflows.fedavg import FedAvg\n",
-    "from nvflare.app_opt.pt.job_config.base_fed_job import BaseFedJob\n",
+    "from nvflare.app_opt.pt.job_config.base_fed_job import PTBaseFedJob\n",
```
holgerroth, line 338: So, are there `PTFedJob` and `TFFedJobs` as well? Maybe we can skip the "Base" in the name in these training framework specific job classes.

SYangster: We currently don't have PTFedJob/TFFedJobs. We could remove the "Base" in these names, would this class naming still make sense? 

FedJob -> BaseFedJob -> PTFedJob

YuanTingHsieh: Another proposal: Does this: PTJob -> CommonJob -> FedJob make sense?

Maybe we don’t need to make the names suggest the parent class they inherit from.

YuanTingHsieh (commented):
Logic is good, naming can discuss with team, from the name would BaseFedJob suggest it is more basic than FedJob?

SYangster:
needs more discussion

PR #3101, opened 2024-12-11
Remove the extra client app custom folder

yhwen:
Fixes # .

### Description

Removed the extra client app custom folder when running a job.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/private/fed/client/client_engine.py
+++ b/nvflare/private/fed/client/client_engine.py
@@ -347,14 +347,6 @@ class ClientEngine(ClientEngineInternalSpec, StreamableEngine):
         if not os.path.exists(app_root):
             return f"{ERROR_MSG_PREFIX}: Client app does not exist. Please deploy it before starting client."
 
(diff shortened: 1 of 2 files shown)


nvflare/private/fed/client/client_engine.py
```diff
@@ -347,22 +347,13 @@ def start_app(
         if not os.path.exists(app_root):
             return f"{ERROR_MSG_PREFIX}: Client app does not exist. Please deploy it before starting client."
 
-        app_custom_folder = os.path.join(app_root, "custom")
```
YuanTingHsieh, line 350: Thanks for the fix.

Where is our custom folder adding logic now?

nvidianz (approved):

nvflare/private/fed/client/client_engine.py
yhwen: It's in the ProcessJobLauncher.

PR #3103, opened 2024-12-11
[2.5] Add authentication to Client API messages

yanchengnv:
Fixes # .

### Description

Messages between CJ and Client API Agent could be relayed through the Server. Without proper authentication headers, these messages will be blocked. This PR adds auth headers to all such messages.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/private/fed/utils/fed_utils.py
+++ b/nvflare/private/fed/utils/fed_utils.py
@@ -33,14 +33,13 @@ from nvflare.apis.job_def import JobMetaKey
 from nvflare.apis.utils.decomposers import flare_decomposers
 from nvflare.apis.workspace import Workspace
 from nvflare.app_common.decomposers import common_decomposers
-from nvflare.fuel.data_event.data_bus import DataBus
+from nvflare.fuel.data_event.utils import get_scope_property, set_scope_property
 from nvflare.fuel.f3.stats_pool import CsvRecordHandler, StatsPoolManager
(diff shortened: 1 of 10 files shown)


nvflare/private/fed/utils/fed_utils.py
```diff
@@ -415,20 +414,15 @@ def set_scope_prop(scope_name: str, key: str, value: Any):
...
         scope_name: name of the scope
         key: key of the scope
+        default: value to return if the prop is not found
     Returns:
     """
-    check_str("scope_name", scope_name)
-    check_str("key", key)
-    data_bus = DataBus()
-    return data_bus.get_data(_scope_prop_key(scope_name, key))
+    return get_scope_property(scope_name, key, default)
```
SYangster, line 428: What is the reasoning for having these functions in addition to the ones added in fuel.data_event.utils?

SYangster (commented):
Good fix, minor comment

nvflare/private/fed/utils/fed_utils.py
yanchengnv: Just for backward compatibility - some places are already using fed_utils.

The real reason to move this function to nvflare.fuel.data_event.utils is for cleaner dependency. We don't want stuff in app_common (e.g. those client API stuff) to depend on fed_utils, which is in "private".

SYangster (approved):

PR #3114, opened 2024-12-17
Add commands to list and retrieve additional components

nvkevlu:
Add commands to list and retrieve additional components.

### Description

After adding additional components to store the error logs (and additional components if implemented in the future), in order to list and download the components, additional admin commands are added.

### Types of changes

- [ ] Documentation updated.


nvflare/apis/job_def_manager_spec.py
```diff
@@ -152,6 +152,20 @@ def get_client_data(self, jid: str, client_name: str, data_type: str, fl_ctx: FL
...
+        """Get list of all the client data for the specified job.
+
+        Args:
+            jid (str): Job ID
+            fl_ctx (FLContext): FLContext information
+
+        Returns:
+            data content
+
+        """
```
YuanTingHsieh, line 166: What do we mean by "all the client data for the specified job"
Do we mean "all the data coming from all the client sites that associates with this job"?
One example being all the client job logs?

Could we be more specific for the return, by data content do we expect bytes? or a list of names?

nvflare/private/fed/server/job_cmds.py
```diff
@@ -422,6 +435,34 @@ def get_job_meta(self, conn: Connection, args: List[str]):
...
+
+        job_id = conn.get_prop(self.JOB_ID)
+        engine = conn.app_ctx
+        job_def_manager = engine.job_def_manager
+        if not isinstance(job_def_manager, JobDefManagerSpec):
+            raise TypeError(
+                f"job_def_manager in engine is not of type JobDefManagerSpec, but got {type(job_def_manager)}"
+            )
+        with engine.new_context() as fl_ctx:
+            list_of_data = job_def_manager.list_client_data(jid=job_id, fl_ctx=fl_ctx)
```
YuanTingHsieh, line 451: since it is only used here, do we want to just call the method in job_def_manager  "list_job_components"?

nvflare/private/fed/server/job_cmds.py
```diff
@@ -422,6 +435,34 @@ def get_job_meta(self, conn: Connection, args: List[str]):
...
+            if list_of_data:
+                filtered_data = [
+                    item for item in list_of_data if item not in {"workspace", "meta", "scheduled", "data"}
+                ]
```
YuanTingHsieh, line 455: Or another way we move these filter mechanism into the "list_client_data" method, so that it does filter out the server-generated data?

One question here is that how do we generate these keys ("meta", "scheduled", "data"), do we have any constant value we can use?

nvflare/private/fed/server/job_cmds.py
```diff
@@ -422,6 +435,34 @@ def get_job_meta(self, conn: Connection, args: List[str]):
...
+                if filtered_data:
+                    data_str = ", ".join(filtered_data)
+                    conn.append_string(data_str)
+                else:
+                    conn.append_string("No client error logs found.")
```
YuanTingHsieh, line 460: Should we say "no additional job components from all clients found"? that is a bit unclear as well.
Or "No client data for this job"? But then this error message not matching the method "list_job_components"

I guess we need to define/add some documentation about what do we mean by "job components" and "client data for a job"

nvflare/private/fed/server/job_cmds.py
```diff
@@ -656,6 +697,51 @@ def download_job(self, conn: Connection, args: List[str]):
...
+                    conn,
+                    tx_id=tx_id,
+                    folder_name=job_id + "_components",
+                    download_file_cmd_name=AdminCommandNames.DOWNLOAD_JOB_FILE,
+                )
+            except Exception as e:
+                secure_log_traceback()
+                self.logger.error(f"exception downloading job {job_id}: {secure_format_exception(e)}")
+                self._clean_up_download(conn, tx_id)
+                conn.append_error("internal error", meta=make_meta(MetaStatusValue.INTERNAL_ERROR))
```
YuanTingHsieh, line 743: The "download_job" method also has some duplicate codes as this part, maybe we could separate out the common downloading/checking logic to another method and reuse it.

nvflare/private/fed/server/job_cmds.py
```diff
@@ -162,7 +169,13 @@ def get_spec(self):
                     handler_func=self.pull_file,
                     authz_func=self.authorize_job_file,
                     client_cmd=ftd.PULL_BINARY_FQN,
-                    visible=False,
+                ),
```
YuanTingHsieh, line 172: do we want download_job_file to be visible? is it still an internal command?

nvflare/private/fed/server/job_cmds.py
```diff
@@ -113,6 +113,13 @@ def get_spec(self):
                     handler_func=self.get_job_meta,
                     authz_func=self.authorize_job,
                 ),
+                CommandSpec(
+                    name=AdminCommandNames.LIST_JOB_COMPONENTS,
```
YuanTingHsieh, line 117: could this name be just "list_job"?

nvflare/private/fed/server/job_cmds.py
```diff
@@ -162,7 +169,13 @@ def get_spec(self):
...
                     authz_func=self.authorize_job_file,
                     client_cmd=ftd.PULL_BINARY_FQN,
-                    visible=False,
+                ),
+                CommandSpec(
+                    name=AdminCommandNames.DOWNLOAD_JOB_COMPONENTS,
+                    description="download additional components for a specified job",
+                    usage=f"{AdminCommandNames.DOWNLOAD_JOB_COMPONENTS} job_id",
+                    handler_func=self.download_job_components,
+                    client_cmd=ftd.PULL_FOLDER_FQN,
```
YuanTingHsieh, line 178: Instead of a new command, what if we enhance the current "download_job" command to also download all the "components" that belongs to the job?

YuanTingHsieh (commented):
Have some questions regarding user-facing concepts/methods

nvflare/apis/impl/job_def_manager.py
```diff
@@ -250,6 +250,13 @@ def get_client_data(self, jid: str, client_name: str, data_type: str, fl_ctx: FL
         except StorageException:
             return None
 
+    def list_client_data(self, jid: str, fl_ctx: FLContext):
+        store = self._get_job_store(fl_ctx)
+        self.log_debug(
+            fl_ctx, f"list_client_data called for {jid}: {store.list_components_of_object(self.job_uri(jid))}"
+        )
+        return store.list_components_of_object(self.job_uri(jid))
```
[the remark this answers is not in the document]
nvkevlu: the reason for this was to list the client logs, but then we made it more generic for data instead of just logs, but you are right, this it not even specific to clients or data, so I will rename it

nvflare/private/fed/server/job_cmds.py
nvkevlu: These keys are special since we use them in the system ourselves. For now, I put it into a variable system_components to make it clearer.

nvflare/apis/impl/job_def_manager.py
```diff
@@ -250,6 +250,13 @@ def get_client_data(self, jid: str, client_name: str, data_type: str, fl_ctx: FL
         except StorageException:
             return None
 
+    def list_components(self, jid: str, fl_ctx: FLContext):
```
yanchengnv, line 253: Please add type hint for returned type.

nvflare/private/fed/server/job_cmds.py
yanchengnv: It must stay invisible since it is used internally and cannot be issued by the user.

nvflare/private/fed/server/job_cmds.py
yanchengnv: No, this is because download_job_components may have different permission than download_job.

nvflare/private/fed/server/job_cmds.py
```diff
@@ -422,6 +435,33 @@ def get_job_meta(self, conn: Connection, args: List[str]):
...
+        with engine.new_context() as fl_ctx:
+            list_of_data = job_def_manager.list_components(jid=job_id, fl_ctx=fl_ctx)
+            if list_of_data:
+                system_components = {"workspace", "meta", "scheduled", "data"}
+                filtered_data = [item for item in list_of_data if item not in system_components]
+                if filtered_data:
+                    data_str = ", ".join(filtered_data)
+                    conn.append_string(data_str)
+                else:
+                    conn.append_string("No additional job components found.")
```
yanchengnv, line 459: Add meta.

yanchengnv (changes requested):

nvflare/private/fed/server/job_cmds.py
nvkevlu: we can clean this up in a future PR

nvflare/private/fed/server/job_cmds.py
YuanTingHsieh: got it, so we should not remove visible=False

nvflare/private/fed/server/job_cmds.py
```diff
@@ -422,6 +435,39 @@ def get_job_meta(self, conn: Connection, args: List[str]):
...
+        with engine.new_context() as fl_ctx:
+            list_of_data = job_def_manager.list_components(jid=job_id, fl_ctx=fl_ctx)
+            if list_of_data:
+                system_components = {"workspace", "meta", "scheduled", "data"}
+                filtered_data = [item for item in list_of_data if item not in system_components]
+                if filtered_data:
+                    data_str = ", ".join(filtered_data)
+                    conn.append_string(data_str)
+                    conn.append_success(
+                        "", meta=make_meta(MetaStatusValue.OK, extra={MetaKey.JOB_COMPONENTS: data_str})
```
yanchengnv, line 459: To make the meta content easier for programmers, don't use data_str here (otherwise the programmer will have to parse it). You can directly use filtered_data which is already a parsed list.

yanchengnv (changes requested):
See my comments to make the meta data easier for programmers to use.

nvflare/private/fed/server/job_cmds.py
```diff
@@ -162,7 +169,13 @@ def get_spec(self):
                     handler_func=self.pull_file,
                     authz_func=self.authorize_job_file,
                     client_cmd=ftd.PULL_BINARY_FQN,
-                    visible=False,
```
YuanTingHsieh, line 165: we should not remove this line

yanchengnv: Yes, must keep visible=False

yanchengnv (changes requested):
Must keep "visible=False" for download_job_file since this is an internal command and not to be exposed to users.

yanchengnv (approved):
LGTM

YuanTingHsieh (approved):

PR #3115, opened 2024-12-17
Handle param converter according to exchange format

ZiyueXu77:
Fixes # .

### Description

Torch can use both numpy and tensor converter, numpy/tf  use numpy, raw use none

### Types of changes

- [ ] Documentation updated.


nvflare/job_config/script_runner.py
```diff
@@ -98,25 +101,29 @@ def __init__(
         self._launch_external_process = launch_external_process
         self._framework = framework
         self._params_transfer_type = params_transfer_type
-        self._from_nvflare_converter_id = from_nvflare_converter_id
```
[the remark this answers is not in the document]
ZiyueXu77: yes indeed good catch! somehow the changes got lost in the last commit

examples/advanced/llm_hf/sft_job.py
```diff
@@ -92,22 +91,17 @@ def main():
...
         elif message_mode == "numpy":
-            runner = BaseScriptRunner(
-                script=train_script,
-                script_args=script_args,
-            )
+            params_exchange_format = "numpy"
         else:
             raise ValueError(f"Invalid message_mode: {message_mode}, only numpy and tensor are supported.")
+
+        runner = BaseScriptRunner(
```
holgerroth, line 100: Sould we just expose params_exchange_format in ScriptRunner as well and use here?

nvflare/job_config/script_runner.py
```diff
@@ -45,8 +45,7 @@ def __init__(
         framework: FrameworkType = FrameworkType.PYTORCH,
         params_transfer_type: str = TransferType.FULL,
         executor: Union[ClientAPILauncherExecutor, InProcessClientAPIExecutor, None] = None,
-        from_nvflare_converter_id: Optional[str] = None,
```
holgerroth, line 48: Should we still keep these? If the user provides converter_ids, we should use them. Otherwise, use the logic based on params_exchange_format.

ZiyueXu77: right, let's add them back to base runner

holgerroth (approved):
Looks good!

PR #3119, opened 2024-12-19
Fix external lib issue when used for pytorch with numpy exchange

ZiyueXu77:
Fixes # .

### Description

One line code update to fix the `module 'nvflare' has no attribute 'app_opt'` error when using external process with numpy exchange format for pytorch

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/client/ex_process/api.py
+++ b/nvflare/client/ex_process/api.py
@@ -52,14 +52,6 @@ def _create_pipe_using_config(client_config: ClientConfig, section: str) -> Tupl
     return pipe, pipe_channel_name
 
 
-def _register_tensor_decomposer():
-    tensor_decomposer, ok = optional_import(module="nvflare.app_opt.pt.decomposers", name="TensorDecomposer")
-    if ok:
-        fobs.register(tensor_decomposer)
-    else:
-        raise RuntimeError(f"Can't import TensorDecomposer for format: {ExchangeFormat.PYTORCH}")
-
-
 class ExProcessClientAPI(APISpec):
     def __init__(self):
         self.process_model_registry = None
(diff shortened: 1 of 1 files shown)


nvflare/client/ex_process/api.py
```diff
@@ -93,7 +93,8 @@ def init(self, rank: Optional[str] = None):
         flare_agent = None
         try:
             if rank == "0":
-                if client_config.get_exchange_format() == ExchangeFormat.PYTORCH:
+                if client_config.get_exchange_format() in [ExchangeFormat.PYTORCH, ExchangeFormat.NUMPY]:
+                    # both numpy and pytorch exchange format can need tensor decomposer
```
nvidianz, line 97: numpy exchange format shouldn't need TensorDecomposer. Something is off.

ZiyueXu77: makes sense, will figure out a better way to address the issue

ZiyueXu77: Tried to find alternatives, but seems it has to be loaded at this API level, otherwise the external process cannot find it. TensorDecomposer is needed by PT when exchange with numpy - PT can have two exchange modes: tensor or numpy, so numpy exchange form can still need TensorDecomposer (to convert the received numpy to tensor that can be loaded).

ZiyueXu77: To make it clearer, we may need to further decouple "framework" (TF, PT, Numpy, etc.) and "exchange format", one framework can use multiple exchange format, but that needs a more thorough design to make sure nothing's broken.

ZiyueXu77: I removed the fobs.register at API, perform import only. register will be performed by the particular framework code when needed (actually we previously "double register" for TensorDecomposer here https://github.com/NVIDIA/NVFlare/blob/main/nvflare/app_opt/pt/client_api_launcher_executor.py#L28 )

nvidianz (approved):

PR #3120, opened 2024-12-20
Kill subprocess on main thread shutting down

nvidianz:
### Description

flare API starts several thread pools but it doesn't provide a shutdown/cleanup call.
The user's code doesn't terminate if it errors out. A watch thread is added to kill the 
process on shutting down.

### Types of changes

- [ ] Documentation updated.


nvflare/client/api.py
```diff
@@ -34,6 +40,23 @@ class ClientAPIType(Enum):
...
 
 
+def death_watch():
+    """
+    Python's main thread doesn't die if there are running thread pools.
+    This function kills the process when the main thread is in the shutdown process
+    """
+    try:
+        while True:
+            if threading._SHUTTING_DOWN:
```
yanchengnv, line 50: I'm not sure this is a good practice. I googled about the use of SHUTTING_DOWN, and people seem to suggest against using it.

Maybe we should just make sure that all threads set daemon to True.

yanchengnv (commented):
Not sure about the death watcher

PR #3121, opened 2024-12-21
Fixed the FOBS lazy loading issue

nvidianz:
### Description

FOBS class loader didn't handle lazy-loaded class correctly. So some class can't be found.
It's fixed by using importlib.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/fuel/utils/fobs/fobs.py
+++ b/nvflare/fuel/utils/fobs/fobs.py
@@ -66,15 +66,10 @@ def _get_type_name(cls: Type) -> str:
 
 def _load_class(type_name: str):
     try:
-        parts = type_name.split(".")
-        if len(parts) == 1:
-            parts = ["builtins", type_name]
+        module_name, class_name = type_name.rsplit('.', 1)
+        module = importlib.import_module(module_name)
 
-        mod = __import__(parts[0])
-        for comp in parts[1:]:
-            mod = getattr(mod, comp)
-
-        return mod
+        return getattr(module, class_name)
     except Exception as ex:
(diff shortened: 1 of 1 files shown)


nvflare/fuel/utils/fobs/fobs.py
```diff
@@ -66,15 +66,10 @@ def _get_type_name(cls: Type) -> str:
 
 def _load_class(type_name: str):
     try:
-        parts = type_name.split(".")
-        if len(parts) == 1:
```
SYangster, line 70: Now that we are using `rsplit(".", 1)` we no longer cover the case where type_name could be a single component name with no "."

Do you know if this case is relevant?

IsaacYangSLA: I also encountered the same issue when trying this PR. 
```
$python test.py 
<class 'datetime.datetime'>
  File "/home/iscyang/wksp/test/test.py", line 9, in <module>
    print(test_class("dict"))
          ^^^^^^^^^^^^^^^^^^
  File "/home/iscyang/wksp/test/test.py", line 4, in test_class
    module_name, class_name = type_name.rsplit(".", 1)
    ^^^^^^^^^^^^^^^^^^^^^^^
ValueError: not enough values to unpack (expected 2, got 1)

$ cat test.py
import importlib

def test_class(type_name):
    module_name, class_name = type_name.rsplit(".", 1)
    module = importlib.import_module(module_name)
    return getattr(module, class_name)

print(test_class("datetime.datetime"))
print(test_class("dict"))
```

SYangster (approved):
LGTM

PR #3122, opened 2024-12-21
[2.5] Fixed the FOBS lazy loading issue

nvidianz:
### Description

FOBS class loader didn't handle lazy-loaded class correctly. So some class can't be found.
It's fixed by using importlib.

### Types of changes

- [ ] Documentation updated.


SYangster (approved):
LGTM

PR #3125, opened 2025-01-02
Add object retrieval thru streaming

yanchengnv:
Fixes # .

### Description

This PR implements a general capability to let application retrieve large data from another site, which will send the large data through streaming.

### Types of changes

- [ ] Documentation updated.


nvflare/app_common/streamers/object_retriever.py
```diff
@@ -0,0 +1,284 @@
...
+    def __init__(
+        self,
+        topic: str = None,
+    ):
+        FLComponent.__init__(self)
+        class_name = self.__class__.__name__
+        if not topic:
+            topic = class_name
+        self.topic = topic
+        self.stream_channel = class_name
```
nvidianz, line 59: Maybe it's better to use a fixed channel like "object_streaming" so we all know the message is for object streaming.

nvflare/app_common/streamers/object_retriever.py
```diff
@@ -0,0 +1,284 @@
...
+        **cb_kwargs,
+    ):
+        """Object requester side, which will receive data stream.
+        This is called to register the status_cb for received stream.
+
+        Args:
+            channel: stream channel
+            topic: stream topic
+            fl_ctx: FLContext object
+            stream_done_cb: the stream_done callback to be registered
```
nvidianz, line 78: The signature of stream_done_cb should be described here

nvflare/app_common/streamers/object_retriever.py
```diff
@@ -0,0 +1,284 @@
...
+        request = Shareable({RETRIEVER_TX_ID: tx_id})
+        if obj_attrs:
+            request.update(obj_attrs)
+
+        # ask the site to start streaming
+        reply = engine.send_aux_request(
+            targets=[from_site], request=request, topic=self.topic, fl_ctx=fl_ctx, timeout=timeout
+        )
+        # the reply is a dict keyed with site names!
+        reply = reply.get(from_site)
```
nvidianz, line 176: It's confusing to redefine the vars here. Should use 2 different vars here. Maybe replies and reply.

nvflare/app_common/streamers/object_retriever.py
```diff
@@ -0,0 +1,284 @@
...
+                    break
+
+                if abort_signal and abort_signal.triggered:
+                    rc = ReturnCode.TASK_ABORTED
+                    break
+            else:
+                # result available!
+                break
+
+        self.tx_table.pop(tx_id, None)
```
nvidianz, line 209: In case of an exception, there will be a leak. Should do this in final block.

nvflare/app_common/streamers/object_retriever.py
```diff
@@ -0,0 +1,284 @@
...
+
+    def _handle_stream_done(self, stream_ctx: StreamContext, fl_ctx: FLContext):
+        # On stream receiving side, which is also the requesting side
+        tx_id = stream_ctx.get(RETRIEVER_TX_ID)
+        waiter = self.tx_table.get(tx_id)
+        if not waiter:
+            self.log_error(fl_ctx, f"late stream completion {tx_id=} after timed out")
+            return
+
+        try:
```
nvidianz, line 227: It's probably more clear to treat result as a tuple, like

        try:
            result = self.get_result(stream_ctx)
        except Exception as ex:
            self.log_error(fl_ctx, f"Exception when get_result: {secure_format_exception(ex)}")
            result = ReturnCode.EXECUTION_EXCEPTION, None

        waiter.result = result

nvidianz (approved):
Some minor suggestions.

nvflare/app_common/streamers/object_retriever.py
yanchengnv: I thought about that, but decided that using the class name is more flexible - there will be multiple classes for object retrieval (e.g. FileRetriever, ModelRetriever, etc.). Each type of retrievers will use the same channel name. Since multiple instances of the same retriever class could be used (e.g. two FileRetriever could be used at the same time), the user could provide a topic.

nvflare/app_common/streamers/object_retriever.py
yanchengnv: It's the same stream_done_cb as documented in apis.streaming.

nvflare/app_common/streamers/object_retriever.py
yanchengnv: Could be.

nvflare/app_common/streamers/object_retriever.py
yanchengnv: Good catch. Fixed.

nvflare/app_common/streamers/object_retriever.py
yanchengnv: Good idea. Changed.

nvidianz (approved):

PR #3133, opened 2025-01-07
Add yaml include support

yhwen:
Fixes # .

### Description

Added the support for yaml to include another yaml configuration file.

YAML does not naturally support any kind of "import" or "include" statement to include another yaml file. Adding this to support the yaml config in the format like: (include could be single file, or a list of include files.)

....
include: 1.yml

or:
include:  [1.yml, 2.yml]

The "include" can be used at any level. Also support recursively include.

### Types of changes

- [ ] Documentation updated.


yanchengnv:
Please explain why this feature is needed.

nvflare/lighter/utils.py
```diff
@@ -180,11 +180,26 @@ def sign_all(content_folder, signing_pri_key):
...
+
+    def include(self, node):
+
+        filename = os.path.join(self._root, self.construct_scalar(node))
+        with open(filename, "r") as f:
+            return yaml.load(f, YamlLoader)
+
+
 def load_yaml(file):
+    YamlLoader.add_constructor("!include", YamlLoader.include)
```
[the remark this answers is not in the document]
yhwen: The disadvantage of using:

```
includes:
  - server_cc_config.yml
  - custom_config.yml
```
The "includes" has to be treated as a special key for including separate yaml file. But the "includes" is the same syntax as the other normal keys. With the "!include", we can clearly define the following is the included yaml file.

Added the unit tests for this.

yhwen:
> Please explain why this feature is needed.

Added the description of how to use the "include". Also include the example in the unit_test.

api_version: 3
name: example_project

include: 1.yml

participants:
  - name: server
    port: 123
    include: [1.yml]
  - name: client
    port: 234
    include: 2.yml

nvflare/lighter/utils.py
```diff
@@ -181,12 +181,40 @@ def sign_all(content_folder, signing_pri_key):
...
+            for item in includes:
+                new_data.update(load_yaml(os.path.join(root, item)))
+        elif isinstance(v, list):
+            new_list = []
+            for item in v:
+                if isinstance(item, dict):
+                    item = load_yaml_include(root, item)
+                new_list.append(item)
+            new_data[k] = new_list
+        else:
```
yanchengnv, line 214: v could be a dict that contains "include" too. For completeness, should do load_yaml_include(root, v).

yanchengnv (commented):
Should support dict too.

nvflare/lighter/utils.py
yhwen: Added dict value handling. Also added the unit test case.

nvflare/lighter/utils.py
```diff
@@ -181,12 +181,42 @@ def sign_all(content_folder, signing_pri_key):
 
 
 def load_yaml(file):
+
+    root = os.path.split(file)[0]
+    yaml_data = None
     if isinstance(file, str):
-        return yaml.safe_load(open(file, "r"))
+        yaml_data = yaml.safe_load(open(file, "r"))
```
yanchengnv, line 188: This will leave the file open forever. Please change to:
with open(file, "r") as f:
  yaml_data = yaml.safe_load(f).

There are probably other places like this.

yanchengnv (changes requested):
Looks good in general. Please fix the open file issue.

nvflare/lighter/utils.py
yhwen: Changed. Also changes all other places with the same open file pattern.

yanchengnv (approved):
Thanks for the fixes of open file issues!

PR #3149, opened 2025-01-17
[2.5] Fix API version mismatch

ZiyueXu77:
Fixes JIRA FLARE-2332 .

### Description

API updates from HF causing backward compatibility issue, added the version information in requirement.txt

### Types of changes

- [ ] Documentation updated.


examples/advanced/llm_hf/requirements.txt
```diff
@@ -1,8 +1,8 @@
...
-torch
+torch==2.5.1
 datasets
 tensorboard
-transformers
-peft
-trl
+transformers==4.44.2
+peft==0.13.2
+trl==0.11.4
```
YuanTingHsieh, line 7: what is the huggingface version?
are these restrictions come from huggingface lib?

ZiyueXu77: all three of them are from huggingface, the reported issue from QA is caused by HF's api updates on these libs, so it will be better to specify the version rather than updating the code

YuanTingHsieh (approved):

PR #3173, opened 2025-01-23
Container Streaming/Retriever

nvidianz:
### Description

1. ContainerStreamer to stream containers.
2. ContainerRetriever to retrieve containers from a remote site.
3. Added examples for file and container streaming.
4. Merged the class loading functions in class_utils and FOBS.
5. Fixed a F3 bug that wipes out original exception.

### Types of changes

- [ ] Documentation updated.


nvflare/app_common/streamers/container_streamer.py
```diff
@@ -0,0 +1,255 @@
...
+        stream_ctx: StreamContext,
+        fl_ctx: FLContext,
+    ) -> Tuple[Shareable, float]:
+
+        try:
+            entry = next(self.iterator)
+            self.count += 1
+            self.last = self.count >= self.size
+        except StopIteration:
+            self.logger.error(f"Producer called too many times {self.count}/{self.size}")
```
yanchengnv, line 119: This condition can only happen when the container is empty or programming error.
You should just return (None, 0) to ensure aborting, instead of returning a Shareable.

yanchengnv (changes requested):
Looks good in general. See my comments for enhancement.

ZiyueXu77:
I did some experiments but it seems the basic transmission without dict streaming has similar memory footprint as compared with streaming, not sure if it is expected (@nvidianz I created a PR on your branch, let me know if you find any mistakes in my code)

ZiyueXu77:
Mis-interpreted the results last Friday, the result actually makes sense:
 
Performed a test with a 6 GB model:
`Communicator - INFO - Received from simulator_server server. getTask: retrieve_dict size: 6GB (5994010153 Bytes)`
Had system memory usage print out every 0.5 s, and the peak usage below:
Simulator w/o dict streaming: 57451 MB
Simulator w/ dict streaming: 53147 MB
Poc w/o dict streaming: 40910 MB
Poc w/ dict streaming: 35422 MB

Considering the largest layer is ~1GB, the difference should be ~5GB, so the above results on diff are reasonable (but the overall memory usage is still pretty big - 10x the model size for simulator)

ZiyueXu77:
looks good and I can confirm the functionality part. This PR is ready for merge once Yan's comment is addressed, I will created a separate PR to refine the readme and example

ZiyueXu77 (commented):
looks good and I can confirm the functionality part. This PR is ready for merge once Yan's comment is addressed, I will created a separate PR to refine the example.

nvflare/app_common/streamers/container_streamer.py
nvidianz: Changed.

PR #3176, opened 2025-01-23
[2.5] Remove hardcode heartbeat_timeout 0

YuanTingHsieh:
### Description

The original hardcode 0 has a problem, if the external user code has an exception and the program will never return.
Our FL client job process (running LauncherExecutor) will never ends.
By using the default heartbeat_timeout value, if the FL client job process does not receive the heartbeat from the user process for heartbeat_timeout seconds, then we will consider it dead.

### Types of changes

- [ ] Documentation updated.


IsaacYangSLA (approved):
LGTM.

PR #3179, opened 2025-01-24
Fix log fl_ctx parsing

SYangster:
Add failure case when attempting to parse bracketed fl_ctx from message

### Types of changes

- [ ] Documentation updated.


nvflare/fuel/utils/log_utils.py
```diff
@@ -96,22 +96,29 @@ def format(self, record):
...
             fl_ctx_match = re.search(r"\[(.*?)\]: ", message)
             if fl_ctx_match:
-                fl_ctx_pairs = {
-                    pair.split("=", 1)[0]: pair.split("=", 1)[1] for pair in fl_ctx_match.group(1).split(", ")
-                }
-                record.fl_ctx = fl_ctx_match[0][:-2]
-                record.identity = fl_ctx_pairs["identity"]  # TODO add more values as attributes?
-                record.msg = message.replace(fl_ctx_match[0], "")
-                self._style._fmt = self.fmt
+                try:
```
yanchengnv, line 99: I guess you are trying to turn fl_ctx info in brackets into some JSON representation? Can you give some examples ?

SYangster: Right, I added a clarifying comment. 

We want to parse out the fl ctx part of the message into a separate log record field "[key0=value0, key1=value1,... ]: <rest of message>". This PR adds case when bracket match is not in fl ctx format.

yanchengnv: Please give some examples. For instance, if the original log msg is like:

[identity=blue, peer=red, time=12345] something happened

It will become what?

SYangster: original message = "[identity=blue, peer=red, time=12345]: something happened"
after parsing -->
record.msg = "something happened"
record.fl_ctx = "[identity=blue, peer=red, time=12345]"
record.identity = "blue" (for now identity is the only key extracted into additional field)

If original message does not contain the bracketed pattern, then record.fl_ctx and record.identity are "".

These attributes can now be configured as desired in the log_config.json format string:
"fmt": "%(asctime)s - %(identity)s - %(name)s - %(levelname)s - %(fl_ctx)s - %(message)s"

yanchengnv: I see, but what are the reasons to extract identity out and put it in the record?

SYangster: Just for convenience, as suggested by Chester. Identity is useful for cases when the log output is combined, such as in the simulator case. But it can be removed if we think otherwise.

holgerroth:
Need this also in 2.5 branch

SYangster:
> Need this also in 2.5 branch

Merged a fix for 2.5 last week: https://github.com/NVIDIA/NVFlare/pull/3178

holgerroth (approved):
Looks good to me.

PR #3182, opened 2025-01-27
[2.5] Fix subprocess read deadlock

nvidianz:
### Description

This PR reads the subprocess stdout as binary data. It addresses the subprocess launcher deadlock issue when the output is not line-oriented or binary.

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh (approved):
it might be better if we add some tests, LGTM

PR #3186, opened 2025-01-28
Separate DockerBuilder and DockerLauncherBuilder

yhwen:
Fixes # .

### Description

Separate DockerBuilder and DockerLauncherBuilder. Fix the **nvflare poc prepare -d 'nvflare/nvflare'**  by keeping the DockerBuilder as before.

### Types of changes

- [ ] Documentation updated.


nvflare/lighter/impl/docker_launcher.py
```diff
@@ -0,0 +1,205 @@
...
+# limitations under the License.
+
+import copy
+import json
+import os
+import shutil
+
+import yaml
+
+from nvflare.app_opt.job_launcher.docker_launcher import ClientDockerJobLauncher, ServerDockerJobLauncher
```
YuanTingHsieh, line 22: Since DockerLauncher depends on "import docker", if we do this import here, are we enforcing all the users must install the "docker" package? @yhwen @chesterxgchen @IsaacYangSLA

nvflare/lighter/impl/docker_launcher.py
```diff
@@ -0,0 +1,205 @@
...
+from nvflare.lighter import utils
+from nvflare.lighter.constants import CtxKey, PropKey, ProvFileName, TemplateSectionKey
+from nvflare.lighter.spec import Builder, Project, ProvisionContext
+
+
+class DockerLauncherBuilder(Builder):
```
YuanTingHsieh, line 28: Could you add some explanation to this? How is this different from DockerBuilder?

YuanTingHsieh (changes requested):

nvflare/lighter/impl/docker_launcher.py
yhwen: Yes, if user wants to build the DockerLauncher package, they need to have docker installed to use.

nvflare/lighter/impl/docker_launcher.py
yhwen: Added the docstring.

nvflare/lighter/impl/docker_launcher.py
YuanTingHsieh: Got it, if they don't want to use this builder they don't need to install docker to run NVFlare right?

yhwen: tested. This is not an issue.

YuanTingHsieh: @yhwen have already tested this, so no issue with this.
If user not use this in their project.yml then without docker installation, provision still work.

YuanTingHsieh (approved):

PR #3187, opened 2025-01-28
Add notebook content for DLI

nvkevlu:
Add notebook content for DLI

### Description

Add notebook content for DLI, experiment tracking in chapter 1 and then content for chapter 2.

### Types of changes

- [ ] Documentation updated.


nvkevlu:
The screenshot attached of the tensorboard chart shows the running_loss graphed, but it is an increasing value that then resets every 3000 steps, so I'm not sure this is the most meaningful thing to log. Is there a better value we should use?

Screenshot 2025-01-28 at 1 27 52 PM

PR #3190, opened 2025-01-29
Cleanup Bogus Errors

nvidianz:
Removed several bogus errors:
1. Error processing frame: RuntimeError: cannot schedule new futures after shutdown
   This can happen when the message arrives for a cell already being shutdown. Changed it to debug.
2. Logical Error: Endpoint is already removed
   This is actually not an error. The weak-ref set entry can be removed when the ref is gone. Changed it to debug. 
3. Added name to several threads to help debugging the dangling thread issue.

### Types of changes

- [ ] Documentation updated.


nvflare/fuel/f3/streaming/stream_utils.py
```diff
@@ -22,11 +23,33 @@
...
+class CheckedExecutor(ThreadPoolExecutor):
+    """This executor ignores task after shutting down"""
+
+    def __init__(self, max_workers=None, thread_name_prefix=""):
+        super().__init__(max_workers, thread_name_prefix)
+        self.stopped = False
+
+    def shutdown(self, wait=True):
+        super().shutdown(wait)
+        self.stopped = True
```
YuanTingHsieh, line 42: Should we move self.stopped before the super().shutdown(wait)?
I guess it comes down to if someone call "shutdown", do we block submit right away?
Or we wait for all things to come back then block.

nvidianz: Good idea. Changed.

YuanTingHsieh (approved):

PR #3215, opened 2025-02-10
Support relay provision

yanchengnv:
Fixes # .

### Description

This PR adds the following features and enhancements.

Relay Provision
Modified the Provisioner to support provisions of relays.  The main properties of a relay are "connect_to" and "listening_host". The "connect_to" property specify the parent site of the relay - it's either the server or another relay. A relay must allow other relays or clients to connect to, hence it must have a listener. The "listening_host" property specifies the host name(s) and port number that the relay will be listening to.

Enhanced "connect_to" implementation
The "connect_to" property was already introduced to support multiple host names for server. But at that time, it was only used for connecting to the server. This PR generalizes it such that it can also be used for connecting to relays as well. It can take two formats: string or dict. The str format is for backward compatibility (connecting to server); whereas the dict format lets you specify the parent name (which is name of either the server or a relay), host, port, and connection security.

Enhanced "listening_host" implementation
The "listening_host" was already used to specify the properties of a site's "internal connection" (which is used for connecting cell children). It was only used for connecting jobs (CJ or SJ). Now it is generalized such that it can also be used for connecting relays.  Since relays may be connected over internet (instead of connected locally for jobs), it must be able to support connections from external nets, hence could have multiple host names, default host, port number, scheme, and connection security.

Client Hierarchy
To support massive scale of mobil applications, CPs will need to be organized hierarchically so allow more efficient message processing and aggregation. Each client site can now specify a "parent" client.  Note that client hierarchy are not the same as relay hierarchy, though should be closely related. Relay hierarchy guarantees message paths between any nodes. Client hierarchy should be very similar to the relay hierarchy, though doesn't have to.

Other fixes
Docker Launch builder also requires the use of internal connection. This PR consolidates this use case and relay use case to the general-purpose "listening_host" mechanism. This PR also fixes some errors in dicker launcher implementation.

### Types of changes

- [ ] Documentation updated.


nvflare/lighter/entity.py
```diff
@@ -11,9 +11,42 @@
...
+
+    def __str__(self):
+        scheme, host_names, default_host, port, conn_sec = (
+            self.scheme,
+            self.host_names,
+            self.default_host,
+            self.port,
+            self.conn_sec,
+        )
+        return f"ConnectTo[{scheme=} {host_names=} {default_host=} {port=} {conn_sec=}]"
```
nvidianz, line 37: ConnectTo -> ListeningHost

nvidianz (approved):
LGTM

nvidianz (approved):

PR #3222, opened 2025-02-13
Remove meta, enable uint8 quantization, and update server filter behavior

ZiyueXu77:
Fixes # .

### Description

Bug in dequantization meta removal, also enabling uint8 quantization
Most importantly, update server behavior so that it will send correct message to multiple clients

### Types of changes

- [ ] Documentation updated.


nvflare/app_opt/pt/quantization/quantizor.py
```diff
@@ -142,9 +143,15 @@ def quantization(self, params: dict, fl_ctx: FLContext):
                         n_bytes_meta += quant_state[param_name]["code"].nbytes
                     else:
                         if source_data_format == "numpy":
+                            if source_data_bits == 8:
+                                # int8 cannot be directly quantized to 4bit, need to convert to float first
```
YuanTingHsieh, line 147: Why is this so?
This is because the constraint of the library we use?
Could you provide link to the doc or explain this is because that particular library

nvflare/app_opt/pt/quantization/quantizor.py
```diff
@@ -195,13 +202,34 @@ def process_dxo(self, dxo: DXO, shareable: Shareable, fl_ctx: FLContext) -> Unio
...
+
+        # for server job and already quantized message, skip quantization
+        # The reason is:
+        # server job in this case is 1-N communication with identical quantization operation
+        # the first communication to client will apply quantization and change the data on the server
+        # thus the subsequent communications to the rest of clients will no longer need to apply quantization
+        # This is not needed for client job, since the client job will be 1-1 and quantization applies to each client
+        # The behavior will also be different if each server-client filter is different, in which case
+        # a deep copy of the server data should be made by filter before applying the process
+        filter_flag = True
```
YuanTingHsieh, line 214: Can we simplified to "if this filter is applied then don't apply again"?

Do we need to check if it is server side?

In the future this could be a general feature

YuanTingHsieh (commented):
Added some questions

nvflare/app_opt/pt/quantization/quantizor.py
ZiyueXu77: Yes, see here https://github.com/bitsandbytes-foundation/bitsandbytes/blob/main/bitsandbytes/functional.py#L870

nvflare/app_opt/pt/quantization/quantizor.py
ZiyueXu77: I think we need to provide more details for users to understand. Discussed this with Yan and Zhihong, and our conclusion is that this will be something to be handled by individual filters, not general pipeline. So we should also include this in our guidelines for filter design.

YuanTingHsieh: got it, suggest rename to "should_apply_quantization"

nvflare/app_opt/pt/quantization/quantizor.py
YuanTingHsieh: thanks, could you add the comment here saying,
```
# bitsandbytes library supports only `float16`, `bfloat16`, or `float32
```

ZiyueXu77: I feel uint8 may not be a realistic use scenario (b&b supports these type is actually reasonable), so I removed the support for uint8 for now.

nvflare/app_opt/pt/quantization/quantizor.py
ZiyueXu77: changed to quantized_flag to represent the state of the message and indicate a bool

YuanTingHsieh: You already have "quantized_flag" what is that ?

ZiyueXu77: filter_flag only false when 1) server job 2) quantized_flag=True, so different var

YuanTingHsieh (approved):
thanks LGTM

YuanTingHsieh (approved):

Issue #3223, opened 2025-02-13
[BUG] False filter behavior from server to multiple clients


ZiyueXu77:
**Describe the bug**
When adding quantizaiton and dequantization filters from server to clients with multiple clients, only the first client gets the correct global model

**To Reproduce**
Following llm_hf example, but setup multiple clients (alternatively, using a simple network)
Examine the global model received by each clients

**Expected behavior**
Global model received by all clients should be the same

**Screenshots**
If using quant-dequant filters, using blockwise8 quantization, we can observe sth like:

model received by site-1:
weight:  tensor([[ 0.3073, -0.0675,  0.2144, -0.3073,  0.1798,  0.0329, -0.0718, -0.1107,
weight_type:  torch.FloatTensor

model received by site-2:
tensor([[255,  55, 233,   0, 225, 191,  54,  45, 201,  50],
weight_type:  torch.ByteTensor

It appears that the model received by site-1 is processed correctly as:
raw model - quantization - dequantization
However, the model received by site-2 appears to be starting from the already quantized data as:
quantized - further quantization - dequantization

**Desktop (please complete the following information):**
 - OS: ubuntu 22.04
 - Python Version 3.10
 - NVFlare Version main branch 

**Additional context**

ZiyueXu77:
Solution in this PR https://github.com/NVIDIA/NVFlare/pull/3227

ZiyueXu77:
Fixed with this PR: https://github.com/NVIDIA/NVFlare/pull/3222, rather than changing the overall behavior (as in https://github.com/NVIDIA/NVFlare/pull/3227), filter implementation should take care of this case, the following should be included in our filter design guideline to let users be aware:

for server job with filtering, special care will be needed:
- server job is a 1-N communication 
- the first communication to client will apply filter and can change the data on the server
- if so, the subsequent N-1 communications to the rest of clients may no longer need to apply filtering
- special treatments will be needed by filter implementation
- (this is not needed for client job, since the client job will be 1-N and filtering applies to each individual client)
- if each server-client filter is different, a deep copy of the server data should be made by filter before applying the filter

PR #3227, opened 2025-02-14
Fixed the filter bug that changes task data

nvidianz:
When server filters are applied, it changes the task data on server. This causes the client get a different data on every GET_TASK.

Fixed the problem in wf_comm_server.py with a deep copy of the data.

The same problem may exist in other controllers.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/apis/impl/wf_comm_server.py
+++ b/nvflare/apis/impl/wf_comm_server.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import copy
 import threading
 import time
 from threading import Lock
@@ -196,7 +197,10 @@ class WFCommServer(FLComponent, WFCommSpec):
         client_task_to_send = None
         with self._task_lock:
             self.logger.debug("self._tasks: {}".format(self._tasks))
-            for task in self._tasks:
+            for t in self._tasks:
+                task = copy.copy(t)
+                task.data = copy.deepcopy(t.data)
+
                 if task.completion_status is not None:
                     # this task is finished (and waiting for the monitor to exit it)
                     continue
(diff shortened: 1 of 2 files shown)


ZiyueXu77:
I think creating copies is more "general" - if potentially server->client filters can be client-specific, but if like in this case, the filters to all clients follow the same logic, maybe we can do it more efficiently (after the first filtering we flag it as filtered then all following just send it directly without applying filter)?

nvflare/apis/impl/wf_comm_server.py
```diff
@@ -196,7 +197,10 @@ def _do_process_task_request(self, client: Client, fl_ctx: FLContext) -> Tuple[s
         client_task_to_send = None
         with self._task_lock:
             self.logger.debug("self._tasks: {}".format(self._tasks))
-            for task in self._tasks:
+            for t in self._tasks:
+                task = copy.copy(t)
+                task.data = copy.deepcopy(t.data)
```
yanchengnv, line 202: Please do NOT do this. If the task data is huge, then deep copying it would be very inefficient.
Server filters are not used much, we shouldn't sacrifice everything for that small issue.
If a filter needs to change the data, the the filter needs to make deep copy there.

yanchengnv (commented):
Please do not do this. See my comment.

yanchengnv:
I don't think deepcopy should be done at the framework level, which will make the whole thing very inefficient. Whether to make a deep copy should be left for the filter to decide.

If the same filter is used on both client and server side, you should make deep copy only on the server side.
Now you can decide which side you are on by calling:

 process_type = fl_ctx.get_process_type()

The value of process_type is one of: ProcessType.CLIENT_JOB, ProcessType.SERVER_JOB.

nvflare/apis/impl/wf_comm_server.py
nvidianz: Moved the copy so it's only done when filters are applied.

ZiyueXu77:
we should not change framework's overall behavior, and this would be filter's job to decide how to properly handle the server-client message, we just need to let users aware. I updated my filter implementation and now it works
https://github.com/NVIDIA/NVFlare/pull/3222/files#diff-a6f6562e9d57991d859d5e89ee07de699220fcbf27d4cb17c98f93797f3358fbR206-R233

ZiyueXu77:
Also updated the issue ticket (we shall add the bullet points somewhere in our documentation)
https://github.com/NVIDIA/NVFlare/issues/3223

PR #3252, opened 2025-02-24
Use Provisioner for Dashboard

yanchengnv:
Fixes # .

### Description

This PR changes Dashboard to use the consolidated Provisioner:

- Consolidated all PKI related cert/key functions into a single utils module, instead of keeping two identical versions (one for dashboard, one for CLI).
- Changed Dashboard to use the same provision engine that is used by the CLI
- Changed NVFL_CREDENTIAL to require an org. Without org, generated server startup kit wouldn't work properly.
- Added AWSBuilder and AzureBuilder to generate Azure and AWS startup scripts for both CLI and Dashboard
- Fixed some other errors.

### Types of changes

- [ ] Documentation updated.


nvkevlu (dismissed):
The changes look good to me

nvkevlu (approved):

PR #3256, opened 2025-02-25
Update quantization to force gpu usage for blockwise8

ZiyueXu77:
Fixes # .

### Description

Account for QA finding and [bug from bitsandbytes ](https://github.com/bitsandbytes-foundation/bitsandbytes/issues/1540)
Also add info to supported precisions

### Types of changes

- [ ] Documentation updated.


nvflare/app_opt/pt/quantization/dequantizor.py
```diff
@@ -84,17 +84,18 @@ def dequantization(
                     params[param_name] = values
                 elif quantization_type in ["blockwise8", "float4", "normfloat4"]:
                     # use bitsandbytes to dequantize the values
+                    # need GPU for general support
```
YuanTingHsieh, line 87: We need to document that this only support GPU

nvflare/app_opt/pt/quantization/quantizor.py
YuanTingHsieh, line 1: Can we rename these files to quantizer/dequantizer? looks like those are the common used words.

nvflare/app_opt/pt/quantization/dequantizor.py
YuanTingHsieh, line 1: Can we rename these files to quantizer/dequantizer? looks like those are the common used words.

YuanTingHsieh (changes requested):

nvflare/app_opt/pt/quantization/quantizor.py
ZiyueXu77: updated

nvflare/app_opt/pt/quantization/dequantizor.py
ZiyueXu77: done

PR #3262, opened 2025-02-26
Support data root in workspace and add entity props for dashboard-based provision

yanchengnv:
Fixes # .

### Description

This PR adds features to support CC:

- Added support of extra data roots in Workspace. 

Three more extra roots can be specified with env vars: NVFL_RESULT_ROOT, NVFL_LOG_ROOT, NVFL_AUDIT_ROOT.

This allows different types of data to be placed in different file systems: startup kit content is under the root_dir (same as now), output generated by application is under RESULT_ROOT, log files will be under LOG_ROOT, and audit files will be under AUDIT_ROOT. Different file systems could be managed with different policies. For example, the content in RESULT_ROOT could be encrypted; LOG_ROOT or AUDIT_ROOT could be write-only, encrypted or not.

These extra roots are optional. When not specified, everything will be under the root_dir (like in current version). 

Content under other roots is organized as:  <root>/<site_name>/<job_id>.

- Added methods in Workspace for application to get result root. Changed np_trainer to use data root for model output.

- Make the job store root to be configurable using env var NVFL_JOB_STORE_ROOT. 

- Augmented dashboard SQL schema with extra properties to capture additional config parameters. This will be used for storing provision properties required by new features (CC, multiple hostnames, connection security, etc.).

These extra properties are dict but are stored as strings, which are generated by json.dumps(props). 

### Types of changes

- [ ] Documentation updated.


nvkevlu:
I agree that the user friendliness of the configuration with the paths separated by ":" is a bit challenging, it is not very intuitive so they would almost be forced to look at documentation to double check and make sure what they are doing is right.

Otherwise the changes look good to me

nvflare/apis/workspace.py
```diff
@@ -53,13 +53,67 @@ def __init__(self, root_dir: str, site_name: str = "", config_folder: str = "con
...
+        data_root = None
+        audit_root = None
+
+        num_parts = len(parts)
+
+        if num_parts == 1:
+            # check env var for other roots (Data, Log, Audit)
+            ws_root = os.environ.get("NVFL_WS_ROOT")
+            if ws_root:
+                root_dir = f"{root_dir}:{ws_root}"
```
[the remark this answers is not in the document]
yanchengnv: We could do that, but that would be many env vars. I don't know. Just preference.
data_root is INTENDED to be used by app to write whatever output too (e.g. models).
In case of CC, imagine that it's a system that will be mounted encrypted.

nvflare/apis/workspace.py
```diff
@@ -53,13 +53,67 @@ def __init__(self, root_dir: str, site_name: str = "", config_folder: str = "con
...
+                parts = root_dir.split(":")
+                num_parts = len(parts)
+
+        if num_parts > 1:
```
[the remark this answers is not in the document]
yanchengnv: This is to minimize code changes through out the code base. I first thought about using a separate dict to specify those roots, but it would take too many changes.

yanchengnv:
> I agree that the user friendliness of the configuration with the paths separated by ":" is a bit challenging, it is not very intuitive so they would almost be forced to look at documentation to double check and make sure what they are doing is right.
> 
> Otherwise the changes look good to me

No users will see or use these at all. All of these is to support multiple roots with minimum code changes throughout our code base.

Maybe we just need to support env var based solution, then we won't need the ":" notation.
We could use multiple env vars.

nvflare/apis/workspace.py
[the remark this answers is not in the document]
yanchengnv: Changed to use env vars to specify extra roots.

nvflare/apis/workspace.py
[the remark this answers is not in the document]
yanchengnv: Changed to use env vars only.

nvflare/dashboard/application/models.py
```diff
@@ -98,11 +118,13 @@ class User(CommonMixin, db.Model):
     role = db.relationship("Role", backref="users")
     organization_id = db.Column(db.Integer, db.ForeignKey("organization.id"), nullable=False)
     organization = db.relationship("Organization", backref="users")
+    props = db.Column(db.String(1048), default="")  # additional properties - JSON string
```
YuanTingHsieh, line 121: should we do 2048 as others?

yanchengnv: This is for admin users. We don't expect many props for users.

nvflare/apis/workspace.py
```diff
@@ -135,11 +168,33 @@ def get_site_custom_dir(self) -> str:
...
 
-    def get_log_file_path(self) -> str:
-        return os.path.join(self.root_dir, WorkspaceConstants.LOG_FILE_NAME)
+    def get_audit_file_path(self, job_id=None) -> str:
+        return os.path.join(self.get_audit_root(job_id), WorkspaceConstants.AUDIT_LOG)
+
+    def _get_site_root_dir(self, root, job_id=None):
+        if job_id:
+            site_root_dir = os.path.join(root, self.site_name, job_id)
+            if not os.path.exists(site_root_dir):
```
nvidianz, line 185: This pattern is used several times. Maybe we should create a convenience function.

nvflare/app_common/np/np_model_persistor.py
```diff
@@ -26,14 +25,11 @@
 
 
 def _get_run_dir(fl_ctx: FLContext):
-    engine = fl_ctx.get_engine()
-    if engine is None:
-        raise RuntimeError("engine is missing in fl_ctx.")
-    job_id = fl_ctx.get_prop(FLContextKey.CURRENT_RUN)
+    workspace = fl_ctx.get_workspace()
+    job_id = fl_ctx.get_job_id()
```
nvidianz, line 29: Should we change get_job_id to use CURRENT_JOB_ID, instead the old RUN_NUM?

nvflare/app_common/storages/filesystem_storage.py
```diff
@@ -72,6 +76,24 @@ def _write(path: str, content, mv_file=True):
         os.rename(tmp_path, path)
 
 
+def _write_multi(output_zip_file_name: str, content: List[str]):
+    with tempfile.TemporaryDirectory() as td:
+        for c in content:
+            if os.path.isfile(c):
+                basename = os.path.basename(c)
```
nvidianz, line 83: Should we keep the folder structure of the content? For example, if content has 'folder1/file', 'folder2/file', this would end up with one file.

nvidianz (approved):
LGTM

nvflare/apis/workspace.py
yanchengnv: This method _get_site_root_dir is the general pattern that is reused in multiple places.

nvflare/app_common/np/np_model_persistor.py
yanchengnv: This is from old days, and some examples already look for RUN_NUM directly. So we can't change the key now.

PR #3267, opened 2025-02-27
Add filter design notes to doc

ZiyueXu77:
Fixes # .

### Description

Add filter design regarding 1-N behavior

### Types of changes

- [ ] Documentation updated.


docs/programming_guide/filters.rst
```diff
@@ -71,6 +71,18 @@ Your subclass of DXOFilter benefits from the features of DXOFilter:
     - Filtering history recording. If a DXO node is processed by your filter, your filter's class name will be appended to the DXO's "filter_history"
     - Auditing. If your filter is applied, a job audit event will be created to record the fact that the filter is applied to data.
 
+DXO Filter Behavior in 1-N Communication
```
YuanTingHsieh, line 74: DXO Filter -> Filter

docs/programming_guide/filters.rst
```diff
@@ -71,6 +71,18 @@ Your subclass of DXOFilter benefits from the features of DXOFilter:
...
+==========
+Based on the design, when a DXO filter is applied to a DXO object, it modifies the DXO object in place. This is fine when the DXO object is expected to be sent to only one recipient, as in the case of 1-1 communication.
```
YuanTingHsieh, line 76: Reading from the DXOFilter, I think it is not mandatory "it modifies the DXO object in place."

So suggest remove this paragraph or reword it.

docs/programming_guide/filters.rst
```diff
@@ -71,6 +71,18 @@ Your subclass of DXOFilter benefits from the features of DXOFilter:
...
+But in the case of 1-N communication, the DXO object will be expected by multiple recipients.
+Assuming a common filter is being used, given the DXO object is modified in place, then the DXO object sent to the second and other recipients
```
YuanTingHsieh, line 78: In the case of 1-N communication, the DXO object will be expected by multiple recipients.
Assuming a common filter is being used, if the DXO object is modified in place, then the DXO object sent to the second and other recipients should not be filtered again, otherwise they might be different from the one sent to the first recipient.

docs/programming_guide/filters.rst
ZiyueXu77: updated

docs/programming_guide/filters.rst
ZiyueXu77: make sense, updated

docs/programming_guide/filters.rst
ZiyueXu77: updated

YuanTingHsieh (approved):

PR #3270, opened 2025-02-28
Fixed client sync retry logic to handle race condition

yanchengnv:
Fixes # .

### Description

At the start of a job, the CJ sends a message to the server to get information about the job's participating clients. It re-sends the message if the return code is not OK. However, there is a scenario (in simulator) that the return code is OK, but the server has not finished setting up the job, and the returned client list is empty.  This PR fixes this issue by improving the retry logic to also include this case.

This PR also changes the retry to be timeout based (max 15 seconds), rather than number of retries based.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/private/fed/client/client_run_manager.py
+++ b/nvflare/private/fed/client/client_run_manager.py
@@ -60,9 +60,9 @@ class ClientRunInfo(object):
 
 
 # TODO: make this configurable
-#   this is the number of retries for client side child/job process to get clients from server
+#   this is the max amount of time for client side child/job process to get clients from server
 #   we might need to think of removing the whole get clients from server logic from child process
-GET_CLIENTS_RETRY = 300
+GET_CLIENTS_TIMEOUT = 15
 
 
 class ClientRunManager(ClientEngineExecutorSpec, StreamableEngine):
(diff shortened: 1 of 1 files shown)


nvflare/apis/client.py
```diff
@@ -67,3 +73,52 @@ def set_is_leaf(self, value: bool):
...
+        d: the dict that contains Client data. This dict should be the result of to_dict() of a Client object.
+
+    Returns: a Client object
+
+    """
+    name = d.get(ClientDictKey.NAME)
+    c = Client(name=name, token="")
+
+    fqcn = d.get(ClientDictKey.FQCN, name)
+    c.set_fqcn(fqcn)
```
nvidianz, line 116: Should check for None here.

nvflare/private/fed/client/client_app_runner.py
```diff
@@ -145,23 +163,60 @@ def start_command_agent(self, args, federated_client, fl_ctx):
...
+                return
+
+            duration = time.time() - start_time
+            assert isinstance(reply, CellMessage)
+            rc = reply.get_header(MessageHeaderKey.RETURN_CODE)
+            if rc == CellReturnCode.OK:
+                self.logger.info(f"notified status {status} to {cp_fqcn} in {duration} seconds after {num_tries} tries")
+                return
+
+            if duration > retry_timeout:
```
nvidianz, line 214: Should handle None for retry_timeout

nvflare/private/fed/client/client_run_manager.py
```diff
@@ -263,7 +249,7 @@ def _to_aux_msg_targets(self, target_names: List[str]):
         if isinstance(target_names, str):
             if target_names == SiteType.ALL:
                 msg_targets = [AuxMsgTarget.server_target()]
-                for _, c in self.all_clients.items():
+                for c in self.all_clients:
```
nvidianz, line 252: Not sure about this loop. If all_clients is a dict, this is looping through the keys.

nvidianz: nm. Saw this is changed to a list now.

nvflare/apis/client.py
yanchengnv: fqcn is default to "name" if no FQCN in the dict.

nvidianz (approved):

PR #3280, opened 2025-03-05
Support custom participant types and custom roles in Provision

yanchengnv:
Fixes # .

### Description

This PR adds back support of custom participant types and custom roles to the Provision system. These features were in previous versions, but got removed accidentally during Provision refactoring.

Also added support of client capacity in StaticFileBuilder. This used to be only available for dashboard in previous versions. 

Also improve the handling of race condition that SJ is not ready when CJ tries to sync.

### Types of changes

- [ ] Documentation updated.


nvflare/dashboard/application/store.py
```diff
@@ -254,10 +251,9 @@ def patch_client_by_creator(cls, id, req):
             client.organization_id = org.id
 
         capacity = req.pop("capacity", None)
-        if capacity is not None:
-            capacity = json.dumps(capacity)
-            cap = get_or_create(db.session, Capacity, capacity=capacity)
-            client.capacity_id = cap.id
+        if capacity:
+            print(f"from creator: {capacity=}")
```
YuanTingHsieh, line 255: Do we need this print or this is for debugging purpose

tests/unit_test/lighter/participant_test.py
```diff
@@ -45,7 +45,7 @@ def test_invalid_org(self, invalid_org):
 
     @pytest.mark.parametrize(
         "invalid_type",
-        [("type1"), ("type@"), ("type!"), ("type~"), ("gateway"), ("firewall")],
+        [("type@"), ("type!"), ("type~")],
```
YuanTingHsieh, line 48: so we allow ("gateway"), ("firewall") as type right now?

nvflare/private/fed/server/server_command_agent.py
```diff
@@ -112,19 +112,17 @@ def aux_communicate(self, request: CellMessage) -> CellMessage:
...
-                return_message = new_cell_message({}, reply)
-                return_message.set_header(MessageHeaderKey.RETURN_CODE, ReturnCode.OK)
-            else:
-                return_message = new_cell_message({}, None)
-            return return_message
+            reply = engine.dispatch(topic=topic, request=data, fl_ctx=fl_ctx)
+            if reply:
+                shared_fl_ctx = gen_new_peer_ctx(fl_ctx)
+                reply.set_peer_context(shared_fl_ctx)
+            return new_cell_message({MessageHeaderKey.RETURN_CODE: ReturnCode.OK}, reply)
```
YuanTingHsieh, line 125: if reply is None, we will return a cell message with "ReturnCode.OK" is that what we want?

nvflare/lighter/entity.py
```diff
@@ -284,27 +315,50 @@ def get_listening_host(self) -> Optional[ListeningHost]:
...
+    Returns: the value of the property. If the property does not exist, ValueError exception is raised.
 
-    org = participant_def.pop("org", None)
-    if org:
-        raise ValueError("missing participant org")
+    """
+    v = d.pop(key, None)
+    if not v:
+        raise ValueError(f"missing participant {key}")
+    return v
```
YuanTingHsieh, line 343: this ```_must_get``` method is equivalent as just:

v = d.pop(key)

just that using pop will raise "KeyError"

nvflare/lighter/entity.py
```diff
@@ -284,27 +315,50 @@ def get_listening_host(self) -> Optional[ListeningHost]:
...
+    Args:
+        participant_def: the dict that contains participant definition
+
+    Returns: a Participant object
+
+    """
+    if not isinstance(participant_def, dict):
+        raise ValueError(f"participant_def must be dict but got {type(participant_def)}")
+
+    name = _must_get(participant_def, PropKey.NAME)
```
YuanTingHsieh, line 358: we can use ``participant_def.pop(PropKey.NAME)`` instead

nvflare/lighter/entity.py
```diff
@@ -355,114 +407,165 @@ def __init__(
...
+
+        Returns: all participants of the project of specified types.
+            If 'types' is not specified (None), it returns all participants of the project;
+            If 'types' is a str, it is treated as a single type and participants of this type is returned;
+            If 'types' is a list of types, participants of these types are returned;
+
+        """
+        if not types:
+            # get all types
+            return list(self._all_names.values())
```
YuanTingHsieh, line 556: if the ``self._all_names = a={"clients": [1,2,3], "admin": [4,5]}``

this ``list(self._all_names.values())``

will give a nested list:

``[[1, 2, 3], [4, 5]]``

Do we want a flat list?
``[1,2,3,4,5]``

nvflare/lighter/impl/static_file.py
```diff
@@ -860,14 +873,17 @@ def finalize(self, project: Project, ctx: ProvisionContext):
...
 
         # create start_all.sh
         self._create_start_all(project, ctx)
 
     @staticmethod
-    def _create_start_all(project: Project, ctx: ProvisionContext):
+    def _append(content: str, participant: Participant) -> str:
+        return content + f"./{participant.name}/startup/start.sh\n"
+
+    def _create_start_all(self, project: Project, ctx: ProvisionContext):
```
YuanTingHsieh, line 886: since _append is a staticmethod, this method can still remain static if we want to, we can call "StaticFileBuilder._append"

YuanTingHsieh (commented):
Mostly LGTM

nvflare/dashboard/application/store.py
yanchengnv: Oh, this is just for debug. Will remove.

tests/unit_test/lighter/participant_test.py
yanchengnv: Yes. This is a request from some customers so that they can use our Provision tool to provision their participant types. And this is one of the main purposes of this PR.

nvkevlu (approved):

yanchengnv (commented):
Reviewed

YuanTingHsieh (approved):
Thanks for the clarification, LGTM

PR #3283, opened 2025-03-07
Add Python Executorch TaskProcessor for edge emulator

YuanTingHsieh:
Add TaskProcessor that use the executorch python bindings directly.

Later we can remove those CPP binary delegation TaskProcessor if these python bindings are good enough.

### Description

- Rename ETTaskProcessor -> ETCppTaskProcessor to reflect it delegates training to cpp binary
- Add ETTaskProcessor
- Add CIFAR10ETTaskProcessor
- Add XORETTaskProcessor
- Add corresponding config files

### Types of changes

- [ ] Documentation updated.


nvflare/edge/emulator/cifar10_cpp.json
```diff
@@ -0,0 +1,4 @@
+{
+    "class_path": "et_cpp_task_processor.ETCppTaskProcessor",
+    "class_args": {"et_binary_path": "train_cifar10", "et_model_path": "cifar10.pte", "data_path": "/Users/yuantingh/Downloads/cifar-10-batches-bin/data_batch_1.bin"}
```
ZiyueXu77, line 3: same here

nvflare/edge/emulator/cifar10_et_task_processor.py
```diff
@@ -0,0 +1,9 @@
+from et_task_processor import ETTaskProcessor
+from torch.utils.data import Dataset
+from torchvision import datasets, transforms
+
+
+class CIFAR10ETTaskProcessor(ETTaskProcessor):
+    def get_dataset(self, data_path: str) -> Dataset:
+        transform = transforms.Compose([transforms.ToTensor(), transforms.Normalize((0.5, 0.5, 0.5), (0.5, 0.5, 0.5))])
```
ZiyueXu77, line 8: can the transform be performed on device (i.e. torchvision works on phone)?

ZiyueXu77 (commented):
few questions

nvflare/edge/emulator/cifar10_et_task_processor.py
YuanTingHsieh: torchvision lib does NOT work on phone, phone need to use other code to do the equivalent thing if they want to.

These code are just for our device emulator that runs in Mac/Linux

ZiyueXu77: For the demo, I think we can do pre-normalization, so that emulator devices and the actual training on your phone will have the same input (unless you already have the same normalization on your phone)

YuanTingHsieh: good point, let me be consistent across diff language

nvflare/edge/emulator/cifar10_cpp.json
nvidianz: Should change this to the new JSON format.

nvflare/edge/emulator/et_cpp_task_processor.py
```diff
@@ -0,0 +1,182 @@
...
+    with open(filename, "wb") as f:
+        f.write(binary_data)
+
+
+def run_training_with_timeout(
+    train_program: str, model_path: str, result_path: str, data_path: str = "", timeout_seconds: int = 300
+) -> int:
+    try:
+        command = [train_program, "--model_path", model_path, "--output_path", result_path, "--data_path", data_path]
+        process = subprocess.Popen(
```
nvidianz, line 50: subprocess.run() is a simpler.

nvflare/edge/emulator/et_task_processor.py
```diff
@@ -36,85 +41,78 @@
...
-            stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE,
-            text=True,
-        )
+def tensor_dict_to_json(d):
+    j = {}
+    for k, v in d.items():
+        entry = {}
+        # TODO: this needs to be compatible with the "Controller" side logic
+        entry["data"] = v.cpu().numpy().tolist()
```
nvidianz, line 49: base64(np.save()) is probably more compact..

nvidianz (approved):
Good enough for demo.

PR #3286, opened 2025-03-07
Improved hash algorithm

nvidianz:
### Description
Added UniformHash for a more evenly distributed hash due to the issue with modulus operation on a non-prime number.

### Types of changes

- [ ] Documentation updated.


nvflare/edge/web/routing_proxy.py
```diff
@@ -25,10 +25,46 @@
...
+    def get_num_buckets(self) -> int:
+        return self.num_buckets
+
+    def hash(self, key: str) -> int:
+        # The hash() function changes value every run so SHA256 is used
+        sha_bytes = hashlib.sha256(key.encode()).digest()
+        sha = int.from_bytes(sha_bytes[:8], 'big')
+        virtual_hash = sha % PRIME
+
+        start = 0
```
yanchengnv, line 57: I think this chunk of code can be simplified:
idx = virtual_hash // self.num
if idx < self.num_buckets:
     return idx
else:
    return virtual_hash % self.num_buckets

yanchengnv (changes requested):
See my comments for hash_utils.

nvflare/edge/web/routing_proxy.py
```diff
@@ -25,10 +25,46 @@
...
 
 
+class UniformHash:
+    """A hash algorithm with uniform distribution. It achieves this with following steps,
+        1. Get a hash value using SHA256
+        2. Map the hash value to a virtual hash table with a large prime number
+        3. Map the virtual bucket to real bucket by using an allocation table
+
+    """
+    def __init__(self, num_buckets: int):
```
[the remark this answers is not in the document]
nvidianz: The buckets are the servers and virtual_hash is the one with large prime number.

nvflare/edge/web/routing_proxy.py
nvidianz: Much simpler. Changed.

nvflare/edge/web/routing_proxy.py
```diff
@@ -25,10 +25,46 @@
...
 from nvflare.edge.web.web_server import FilteredJSONProvider
 
+# This is just a random large prime number used as number of buckets
+PRIME = 100003
+
 log = logging.getLogger(__name__)
 app = Flask(__name__)
 
 
+class UniformHash:
```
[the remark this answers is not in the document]
nvidianz: Moved.

nvflare/fuel/utils/hash_utils.py
```diff
@@ -0,0 +1,45 @@
...
+class UniformHash:
+    """A hash algorithm with uniform distribution. It achieves this with following steps,
+    1. Get a hash value using SHA256
+    2. Map the hash value to a virtual hash table
+    3. Map the virtual bucket to real bucket
+
+    """
+
+    def __init__(self, num_buckets: int):
+        self.num_buckets = num_buckets
```
yanchengnv, line 29: Add sanity check against num_buckets: should be a positive int less than a max.

nvflare/fuel/utils/hash_utils.py
```diff
@@ -0,0 +1,45 @@
...
+        self.virtual_hashes_per_bucket = PRIME // num_buckets
+
+    def get_num_buckets(self) -> int:
+        return self.num_buckets
+
+    def hash(self, key: str) -> int:
```
yanchengnv, line 35: Check key is actually a str, or raise exception.

yanchengnv (commented):
LGTM. Please add sanity check since hash_utils is a library now.

nvflare/fuel/utils/hash_utils.py
nvidianz: Added.

nvflare/fuel/utils/hash_utils.py
nvidianz: Done

nvflare/fuel/utils/hash_utils.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def __init__(self, num_buckets: int):
+        check_number_range("num_buckets", num_buckets, 1, MAX_NUM_BUCKETS)
+        self.num_buckets = num_buckets
+        self.virtual_hashes_per_bucket = PRIME // num_buckets
+
+    def get_num_buckets(self) -> int:
+        return self.num_buckets
+
+    def hash(self, key: str) -> int:
```
yanchengnv, line 39: Please add docstrings to explain every method.

yanchengnv (changes requested):
Looks good. Please add docstrings.

nvflare/fuel/utils/hash_utils.py
```diff
@@ -0,0 +1,52 @@
...
+        check_str("key", key)
+        # The hash() function changes value every run so SHA256 is used
+        sha_bytes = hashlib.sha256(key.encode()).digest()
+        sha = int.from_bytes(sha_bytes[:8], "big")
```
yanchengnv, line 43: I remember that in your previous version, you used crc to compute the hash key. Why did you change?
I also tried zlib.crc32(key.encode(), 0), and compared standard deviation between sha and crc-based methods for randomly generated list of keys. CRC's stdev is smaller/better most of the times (8 of 10 times) than sha-based.

yanchengnv: Actually their performance is about the same if trying enough times.
But CRC uses about half compute time.
I guess either is okay.
Another finding is that if you simply do int.from_bytes(key.encode[:8], "big"), it also works equally well, but takes less time. Of course the key must have enough 8 bytes.

nvidianz: Since we are making this part of util, I selected SHA256 for its robustness. CRC32 is not considered a real hash algorithm. We are not very sensitive to performance so this is ok.

nvflare/fuel/utils/hash_utils.py
nvidianz: Added.

yanchengnv (approved):
LGTM

PR #3302, opened 2025-03-16
Due to credential env var format change, the user prompt requires one

IsaacYangSLA:
more information during dashboard launch.  This PR fixes the issue with updated UI and logic.

### Description

Dashboard initial user (project admin) now requires organization name.  The env var `NVFL_CREDENTIAL` now has format as 'email:passowrd:org_name.'  All user facing operations requires to update to support this new format.

### Types of changes

- [ ] Documentation updated.


IsaacYangSLA:
I think this PR targets branch 2.6.  I will do the same on main.

PR #3307, opened 2025-03-18
Fix log exception dict, example and doc updates

SYangster:
Addresses [FLARE-2416](https://jirasw.nvidia.com/browse/FLARE-2416), [FLARE-2429](https://jirasw.nvidia.com/browse/FLARE-2429), and [FLARE-2428](https://jirasw.nvidia.com/browse/FLARE-2428)

- Fix issue for log exception with dictionaries for json formatter @nvidianz 
- Clarify LoggerNameFilter behavior for allowing all log records with level > INFO, added argument
- Update example and documentation simulator log arguments (for simulator examples, keeping the default LogMode.CONCISE)
- Will cherry-pick to 2.6 after merge

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -6,6 +6,7 @@ on:
     branches:
       - "main"
       - "2.5"
+      - "2.6"
     paths:
       - 'web/**'
   workflow_dispatch:
--- a/.github/workflows/premerge.yml
+++ b/.github/workflows/premerge.yml
@@ -26,7 +26,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-22.04, ubuntu-20.04 ]
+        os: [ ubuntu-22.04, ubuntu-24.04 ]
         python-version: [ "3.9", "3.10", "3.11", "3.12" ]
     steps:
       - uses: actions/checkout@v4
@@ -46,7 +46,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-22.04, ubuntu-20.04 ]
+        os: [ ubuntu-22.04, ubuntu-24.04 ]
         python-version: [ "3.9", "3.10", "3.11", "3.12" ]
     steps:
       - uses: actions/checkout@v4


docs/user_guide/configurations/logging_configuration.rst
```diff
@@ -193,9 +193,11 @@ LoggerNameFilter
 ----------------
 :class:`LoggerNameFilter<nvflare.fuel.utils.log_utils.LoggerNameFilter>` filters loggers based on a list of logger_names.
 Filters utilize the logger hierarchy, so any descendants of the specified names will also be allowed through the filter.
+By default, LoggerNameFilter is configured with allow_errors to allow all logs with level greater than INFO though even if they are not from a logger in logger_names.
 
 - **logger_names**: list of logger names to allow through filter
 - **exclude_logger_names**: list of logger names to disallow through filter (takes precedence over allowing from logger_names)
+- **allow_errors**: allow all log records with levelno greater logging.INFO through. Defaults to True.
```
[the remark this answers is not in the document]
SYangster: allow_errors is a parameter I added to LoggerNameFilter

nvflare/fuel/utils/log_utils.py
```diff
@@ -240,33 +240,34 @@ def format(self, record) -> str:
...
+        self.allow_errors = allow_errors
 
     def filter(self, record):
         name = getattr(record, "fullName", record.name)
 
         is_logger_included = self.matches_name(name, self.logger_names)
         is_logger_excluded = self.matches_name(name, self.exclude_logger_names)
 
-        return (record.levelno > logging.INFO) or (not is_logger_excluded and is_logger_included)
+        return (self.allow_errors and record.levelno > logging.INFO) or (is_logger_included and not is_logger_excluded)
```
YuanTingHsieh, line 270: just notice this right now,
why would we have both "included" and "excluded"?
Should one be enough?

docs/user_guide/configurations/logging_configuration.rst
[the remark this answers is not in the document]
YuanTingHsieh: then should we call it: allow_high_level_logs_to_bypass

docs/user_guide/configurations/logging_configuration.rst
```diff
@@ -193,9 +193,11 @@ LoggerNameFilter
 ----------------
 :class:`LoggerNameFilter<nvflare.fuel.utils.log_utils.LoggerNameFilter>` filters loggers based on a list of logger_names.
 Filters utilize the logger hierarchy, so any descendants of the specified names will also be allowed through the filter.
+By default, LoggerNameFilter is configured with allow_errors to allow all logs with level greater than INFO though even if they are not from a logger in logger_names.
```
YuanTingHsieh, line 196: Or maybe more descriptive here?
Not sure if this is overkill:

We have a mechanism in the `LoggerNameFilter` configuration that allows logs with a level greater than INFO, even if they do not come from loggers listed in logger_names. This behavior effectively bypasses the logger name filter for higher-level logs. By default, this mechanism is enabled (`allow_high_level_logs_to_bypass=True`), meaning logs above the INFO level are allowed regardless of the logger name.

docs/user_guide/configurations/logging_configuration.rst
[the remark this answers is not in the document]
SYangster: I see, would renaming such as "allow_high_level_logs_to_bypass" or "allow_all_error_logs" make more sense, or rather not having the parameter at all be clearer?

nvflare/fuel/utils/log_utils.py
SYangster: If we want, we can remove "excluded" for more simplicity. The idea was since logger names are hierarchical, you could include "a" but exclude "a.b.c" for example, which gives more fine-grained control. Otherwise with only include this would be much more difficult, but maybe it is not a common use case.

docs/user_guide/configurations/logging_configuration.rst
YuanTingHsieh: thanks for the name change, this line needs to be updated as well::

"allow_errors" -> `allow_all_error_logs`

YuanTingHsieh (approved):
Thanks. LGTM

nvidianz (approved):
LGTM

PR #3311, opened 2025-03-20
Fix job templates

YuanTingHsieh:
After https://github.com/NVIDIA/NVFlare/pull/3088 https://github.com/NVIDIA/NVFlare/pull/3115/
our params_exchange_format meaning changed a bit.
right now it means the exchange format between client script and NVFlare system.

### Description

Update the job_templates accordingly and add a migration doc

### Types of changes

- [ ] Documentation updated.


ZiyueXu77 (approved):
Good catch

PR #3312, opened 2025-03-20
Fix SubprocessLauncher backward compatibility

YuanTingHsieh:
Default should be the same as previous behavior.
If users want to wait the external process to exit, they need to explicitly specify number of seconds to wait.

### Description

Change the default behavior back to the same as 2.5

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh:
@holgerroth please verify the bionemo/nemo example is working, you might need to specify this "shutdown_timeout" to be a meaningful number

nvidianz (approved):

PR #3313, opened 2025-03-20
Remove start_app command

yanchengnv:
Fixes # .

### Description

The "start_app" admin command is removed for the following reasons:
- It does not work well with our job-oriented operation model. It is a feature from Clara Train time when the job had to be started manually for each client. Currently the job is automatically started by the SP. No manual operation is needed. 
- It could cause confusion as to when this command could be used.
-  No one actually uses this command any way. Our Flare API doesn't support it either.

### Types of changes

- [ ] Documentation updated.


nvkevlu (approved):
Thanks, LGTM

YuanTingHsieh (approved):
LGTM

PR #3318, opened 2025-03-21
Fix logic in AnalyticsReceiver to correctly check for server-side process

YuanTingHsieh:
The check added in #2722 is not 100% accurate. Since our identity sometimes will be project name for server job process.
We change the check to use `fl_ctx.get_process_type()`

### Description

- Change to use `fl_ctx.get_process_type()`

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh:
This fix will make sure the 2.6/main performs the same as before.
An enhancement PR is coming next.

PR #3330, opened 2025-03-26
Add dependency for tensorboard to fix example

nvkevlu:
Add dependency for tensorboard to fix example.

### Description

Add dependency for tensorboard to fix example to address bug 5178901.

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh:
Do we need an equivalent for other examples?
Do we need this PR in 2.6 branch?

nvkevlu:
I don't think other examples include any mention of tensorboard, but if they do and it is missing, we should include it too.

If the 2.6 branch won't be rebuilt from main, then yes, I think we should also make a PR for that branch.

PR #3346, opened 2025-04-01
Separate STDERR from STDOUT in tie

YuanTingHsieh:
### Issue

While testing the TIE + Flower job, I discovered that if the process outputs to STDERR, any attempt to use STDOUT for retrieving results fails when the two streams are mixed together.

For example: https://github.com/NVIDIA/NVFlare/blob/2.6/nvflare/app_opt/flower/applet.py#L225-L232

### Description

This PR separates STDERR from STDOUT to prevent mixing the two streams and ensure proper handling of the output.

### Types of changes

- [ ] Documentation updated.


yanchengnv (approved):
Good finding!

PR #3348, opened 2025-04-02
[2.6] Remove edge code and examples

YuanTingHsieh:
### Description

Remove edge code and examples in 2.6

### Types of changes

- [ ] Documentation updated.


nvflare/private/fed/client/client_runner.py
```diff
@@ -162,35 +157,6 @@ def __init__(
         self.submit_task_result_timeout = self.get_positive_float_var(ConfigVarName.SUBMIT_TASK_RESULT_TIMEOUT, None)
         self._register_aux_message_handlers(engine)
 
-    def set_cell(self, cell):
```
yanchengnv, line 165: No, you cannot remove the set_cell method! It is called from fed_client_base.py.
You should change its implementation to "pass".

yanchengnv (changes requested):
See my comment. Please fix the functional bug in client_runner.py.

PR #3349, opened 2025-04-02
[2.6] Fix azure start scripts generation in azure builder

yanchengnv:
Fixes # .

### Description

Added missing pattern replacement to azure builder.

### Types of changes

- [ ] Documentation updated.


IsaacYangSLA (approved):
Replacement dicts are critical when generating shell scripts or text files from templates as the pattern `{~~KEY~~}` would be replaced by the value of that key.

nvkevlu (approved):

PR #3358, opened 2025-04-03
[2.6] Fixed a race condition in testing tool

nvidianz:
### Description

The sender.py may fail if the connection is not established faster enough. Added an event to 
wait for receiver to go online.

### Types of changes

- [ ] Documentation updated.


yanchengnv (approved):
LGTM

PR #3362, opened 2025-04-03
[2.6] Fix TensorDecomposer

YuanTingHsieh:
msgpack can't handle a dict that contains big bytes that is bigger than 4GB.

### Description

In the decomposer, we use Internalizer and Externalizer to break down the large bytes object and assemble back.

### Types of changes

- [ ] Documentation updated.


yanchengnv (approved):
LGTM

YuanTingHsieh:
Use #3372 for a more general fix.

PR #3372, opened 2025-04-04
[2.6] Use externalizer in FOBS

nvidianz:
### Description

Changed FOBS to use externalizer to handle larger than 4GB items in dict.

### Types of changes

- [ ] Documentation updated.


tests/unit_test/apis/utils/decomposers/flare_decomposers_test.py
```diff
@@ -72,7 +72,7 @@ def test_byte_stream(self, size):
             "zb": b"123456789012345678901234567890123456789012345678",
             "zc": "中文字母测试两岸猿声啼不住轻舟已过万重山:;.'[]{}`~<>!@#$%^&*()-_+=",
         }
-        ds = fobs.dumps(d, max_value_size=10)
+        ds = fobs.dumps(d, max_value_size=1024)
```
YuanTingHsieh, line 75: why this max need to set to 1024? what does this mean?

YuanTingHsieh (approved):
One question regarding test, otherwise LGTM

PR #3385, opened 2025-04-09
[2.6] Enhance Params Exchange Mechanism

YuanTingHsieh:
(1) The params_exchange_format only defines client process to script communication format. We should have another format that describe the server process <-> client process communication.
(2) The ParamsConverter concept should be implementing using our Filter mechanism instead of baked in an Executor
(3) framework argument is vague in ScriptRunner, we should improve it
(4) When the “params_transfer_type” is DIFF and the user script sends an FLModel of FULL, AND when the “params_exchange_format” is PyTorch or NumPy, the system performs an implicit FULL to DIFF conversion. However, this behavior is limited to only these formats. For other formats, no conversion occurs. This causes more confusion.

### Description
- Add ParamsConverterFilter
- Add server_client_params_exchange_format and client_script_params_exchange_format to ScriptRunner
- Remove no need code after this refactor
- Update the `examples/hello-world/ml-to-fl/tf/tf_client_api_job.py` and `examples/hello-world/ml-to-fl/pt/pt_client_api_job.py` and `job_templates/sag_pt/config_fed_client.conf` to show the required changes after this
- Remove params_transfer_type

Note that we can't register a filter with tasks = ['*'] right now, it has to have some task names

### Types of changes

- [] Non-breaking change (fix or new feature that would not break existing functionality).
- [ ] Documentation updated.


examples/hello-world/ml-to-fl/tf/tf_client_api_job.py
```diff
@@ -51,7 +52,8 @@ def main():
     executor = ScriptRunner(
         script=script,
         launch_external_process=launch_process,
-        framework=FrameworkType.TENSORFLOW,
+        server_client_params_exchange_format=ExchangeFormat.NUMPY,
+        client_script_params_exchange_format=ExchangeFormat.KERAS_NUMPY,
```
nvidianz, line 56: Do we need 2 parameters? server/client should either use the same format or convert to the final format (script). Otherwise, we may have 3 formats involved.

nvflare/app_common/filters/params_converter_filter.py
```diff
@@ -0,0 +1,70 @@
...
+            dxo (DXO): DXO to be filtered.
+            shareable: that the dxo belongs to
+            fl_ctx (FLContext): only used for logging.
+
+        Returns:
+            Converted dxo.
+        """
+        self.log_info(fl_ctx, f"processing with {dxo.data_kind}")
+        self.log_info(fl_ctx, "Starts ParamsConverterFilter")
+        params_converters: List[ParamsConverter] = _get_params_converter(self.params_converter_ids, fl_ctx)
```
ZiyueXu77, line 61: don't quite get the point of this layer of abstraction, shall we directly write converter as filter?

ZiyueXu77 (commented):
Overall looks good! A few questions. 
One can be general: when multiple filters coexisting, their sequence can be important and we may need to perform some extra checks and loggings to give users useful info for debugging @yanchengnv

job_templates/sag_pt/config_fed_client.conf
```diff
@@ -49,10 +49,46 @@
...
+        "validate"
+      ],
+      "filters": [
+        {
+          "path": "nvflare.app_common.filters.params_converter_filter.ParamsConverterFilter",
+          "args": {
+            "params_converter_ids": [
+              "np_to_pt"
+            ]
+          }
```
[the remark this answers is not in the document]
ZiyueXu77: yes that's my guess too. I think it may be helpful to ask users to explicitly define it, otherwise it could be hard to debug, considering a mistakenly swapped filter, especially when the sending side and receiving side chould / sould have a reverse sequence.

examples/hello-world/ml-to-fl/tf/tf_client_api_job.py
yanchengnv: Before we have some auto/default mechanism, I think we do need two formats: 
- what the script expects the input to be
- what the server expects the response to be

Since the client (CJ) only plays as a middleman, we probably should rename these two formats:
server_client_params_exchange_format => server_expected_format
client_script_params_exchange_format => script_expected_format

job_templates/sag_pt/config_fed_client.conf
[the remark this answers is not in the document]
yanchengnv: The user has to define the list of filters in the right order.

nvflare/app_common/abstract/params_converter.py
```diff
@@ -13,26 +13,13 @@
...
-        self.logger = get_obj_logger(self)
-
-    def process(self, task_name: str, shareable: Shareable, fl_ctx: FLContext) -> Shareable:
-        if not self.supported_tasks or task_name in self.supported_tasks:
-            dxo = from_shareable(shareable)
-            dxo.data = self.convert(dxo.data, fl_ctx)
-            dxo.update_shareable(shareable)
-        return shareable
-
+class ParamsConverter(FLComponent, ABC):
```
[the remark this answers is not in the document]
yanchengnv: I think it's  good idea to define converters separately since they could be used independent of filters.

nvflare/app_common/filters/params_converter_filter.py
```diff
@@ -0,0 +1,70 @@
...
+from typing import List, Union
+
+from nvflare.apis.dxo_filter import DXO, DXOFilter
+from nvflare.apis.fl_context import FLContext
+from nvflare.apis.shareable import Shareable
+from nvflare.app_common.abstract.params_converter import ParamsConverter
+
+
+def _get_params_converter(params_converter_ids: List[str], fl_ctx: FLContext) -> List[ParamsConverter]:
+    filters = []
```
yanchengnv, line 24: These should be "converters", not "filters"

nvflare/app_common/filters/params_converter_filter.py
yanchengnv: I think it's okay to have one "converter" per filter, instead of a list of converters.
Then you can chain filters when necessary.

yanchengnv, line 61: You should do the get_params_converter logic in "START_RUN" event handler - this is the standard pattern how components should be resolved.

nvflare/job_config/script_runner.py
```diff
@@ -270,43 +251,85 @@ def add_to_fed_job(self, job: FedJob, ctx, **kwargs):
...
-
-            return PTClientAPILauncherExecutor
-        elif framework == FrameworkType.TENSORFLOW:
-            from nvflare.app_opt.tf.client_api_launcher_executor import TFClientAPILauncherExecutor
-
-            return TFClientAPILauncherExecutor
-        else:
-            return ClientAPILauncherExecutor
+    def _add_params_converter_filter(self, job, ctx):
+        if (
```
yanchengnv, line 268: Think about better re-organizing this block code.
You can use _SUPPORT_EXCHANGE_FORMAT to register a "add_filters_cb" for each supported combination. Then this chunk of code simply needs to call the CB of the combo.

yanchengnv (changes requested):
The filter-based idea streamlined model format processing and got rid of unnecessary components (those wrapper executors in opt). Suggestions for further enhancement:
- One converter per filter to keep things simple
- Use START_RUN event to resolve components
- Re-organize that big if-then-else to create filters for different combos

nvflare/app_common/abstract/params_converter.py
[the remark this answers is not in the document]
YuanTingHsieh: thanks for the input, keep it as separate for now.

nvflare/app_common/filters/params_converter_filter.py
YuanTingHsieh: thanks, changed to use one converter per filter.

examples/hello-world/ml-to-fl/tf/tf_client_api_job.py
YuanTingHsieh: Renamed as suggested.
Right now we already have the following combinations:

SJ <-NUMPY-> CJ <-PYTORCH-> SCRIPT
SJ <-PYTORCH-> CJ <-PYTORCH-> SCRIPT

So we need a way to express these as well.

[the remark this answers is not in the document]
YuanTingHsieh: changed the KERAS_NUMPY to KERAS_LAYER_WEIGHTS,
it means: {layer.name: layer.get_weights() for layer in model.layers}


@ZiyueXu77
Right now our "Aggregators" DOES NOT work directly with KERAS layer weights.
So we utilize flat and unflat methods here: https://github.com/NVIDIA/NVFlare/blob/main/nvflare/app_opt/tf/utils.py to make them compatible with our flat NumPy based aggregators.

If we have corresponding aggregators and model persistor for KERAS layer weights, we could remove this format.

job_templates/sag_pt/config_fed_client.conf
```diff
@@ -32,7 +32,7 @@
           heartbeat_timeout = 60
 
           # format of the exchange parameters
-          params_exchange_format =  "numpy"
+          params_exchange_format =  "pytorch"
```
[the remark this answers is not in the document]
YuanTingHsieh: changed to "script_expected_format"

nvflare/app_common/filters/params_converter_filter.py
YuanTingHsieh: fixed.

nvflare/app_common/filters/params_converter_filter.py
YuanTingHsieh: changed as suggested.

nvflare/job_config/script_runner.py
YuanTingHsieh: changed as suggested for these pre-built scenario.

nvflare/app_common/filters/params_converter_filter.py
```diff
@@ -0,0 +1,64 @@
...
+from nvflare.apis.event_type import EventType
+from nvflare.apis.fl_context import FLContext
+from nvflare.apis.shareable import Shareable
+from nvflare.app_common.abstract.params_converter import ParamsConverter
+
+
+def _get_params_converter(params_converter_id: str, fl_ctx: FLContext) -> ParamsConverter:
+    c = fl_ctx.get_engine().get_component(params_converter_id)
+
+    if c:
```
yanchengnv, line 27: What if "c" is None? I think we should simply check "if not isinstance(c, ParamsConverter)" and then raise exception.

yanchengnv (commented):
I saw some model registry changes. Are they related to model format conversion?

yanchengnv (commented):
See my comments.

nvflare/job_config/script_runner.py
```diff
@@ -270,44 +294,24 @@ def add_to_fed_job(self, job: FedJob, ctx, **kwargs):
...
+                    script_expected_format=self._script_expected_format,
                 )
             )
             job.add_executor(executor, tasks=tasks, ctx=ctx)
 
+        exchange_format_combination = (
+            self._server_expected_format,
+            self._script_expected_format,
+        )
+        if exchange_format_combination in AUTO_REGISTERED_EXCHANGE_FORMAT_COMBINATIONS:
```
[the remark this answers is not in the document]
YuanTingHsieh: I think we should not, if the combination is not in the AUTO_REGISTERED, that means we won't register this params converter automatically for them, they will have to do that on their own.

Or should we print out some warning to remind them?

yanchengnv (approved):
The current solution looks good enough for now. We need to improve it further to make model self-descriptive about its format. Then we may be able to eliminate the need of server_expected_format.

PR #3393, opened 2025-04-11
[2.6] Fix xgb sec issue

### LINKED ISSUE #3336: Prediction for xgboost VFL
Hi, 
I am currently working on a POC on xgboost based vertical federated learning and a perhaps silly question has arisen 
I got two models for two clients from the example: https://github.com/NVIDIA/NVFlare/tree/main/examples/advanced/xgboost/fedxgb without any problems.
but judging by the code only the training functionality of the model https://github.com/NVIDIA/NVFlare/tree/main/nvflare/app_opt/xgboost/histogram_based_v2 is implemented here. 
Is there an example how I could get predictions for the same sample I trained on still distributed between two clients?

YuanTingHsieh:
Fixes #3336.

I found out our vertical should be using 2 clients instead of 3.

### Description

- Update the instructions

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/examples/advanced/xgboost/fedxgb_secure/README.md
+++ b/examples/advanced/xgboost/fedxgb_secure/README.md
@@ -14,6 +14,9 @@ To be able to run all the examples, please install the requirements first.
 ```
 pip install -r requirements.txt
 ```
+
+If you want to run the secure vertical XGBoost with libnvflare.so plugin, you will need to install the [IPCL](https://github.com/intel/pailliercryptolib_python) library (by `git clone` and `pip install -e .`).
+
 ## Encryption Plugins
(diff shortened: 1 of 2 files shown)


ZiyueXu77:
why limiting 2? 3 works just fine

examples/advanced/xgboost/fedxgb_secure/README.md
```diff
@@ -14,6 +14,9 @@ To be able to run all the examples, please install the requirements first.
 ```
 pip install -r requirements.txt
 ```
+
+If you want to run the secure vertical XGBoost with libnvflare.so plugin, you will need to install the [IPCL](https://github.com/intel/pailliercryptolib_python) library (by `git clone` and `pip install -e .`).
+
```
ZiyueXu77, line 19: I think we are providing mixed and broken info for plugin, let's have a discussion

PR #3402, opened 2025-04-15
[2.6] Use server_expected_format to represent the server <-> client params format

YuanTingHsieh:
Make "params_exchange_format" has the same meaning as in previous release.
Introduce a new parameter "server_expected_format" to let user define the format between server <-> client.

"framework" in ScriptRunner and "params_exchange_format" in executors mean the same thing now.

We will enhance these in the next release.

### Description

- Make "params_exchange_format" has the same meaning as in previous release.
- Introduce a new parameter "server_expected_format" to let user define the format between server <-> client.
- Update examples/tutorials/release notes

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/examples/advanced/sklearn-kmeans/kmeans_job_clientapi.py
+++ b/examples/advanced/sklearn-kmeans/kmeans_job_clientapi.py
@@ -135,7 +135,7 @@ def main():
             script_args=f"--data_path {data_path} "
             f"--train_start {train_start} --train_end {train_end} "
             f"--valid_start {valid_start} --valid_end {valid_end}",
-            params_exchange_format="raw",
+            framework="raw",
         )
         job.to(executor, f"site-{i}", tasks=["train"])
 
--- a/job_templates/sag_gnn/app_2/config_fed_client.conf
+++ b/job_templates/sag_gnn/app_2/config_fed_client.conf
@@ -32,7 +32,7 @@
           heartbeat_timeout = 60
 
           # format of the exchange parameters
-          params_exchange_format =  "numpy"
+          params_exchange_format =  "pytorch"
 
           # if the transfer_type is FULL, then it will be sent directly
           # if the transfer_type is DIFF, then we will calculate the
--- a/job_templates/sag_nemo/app_1/config_fed_client.conf
+++ b/job_templates/sag_nemo/app_1/config_fed_client.conf
@@ -35,7 +35,7 @@
           heartbeat_timeout = 60
 
           # format of the exchange parameters
-          params_exchange_format =  "numpy"
+          params_exchange_format =  "pytorch"
 
           # if the transfer_type is FULL, then it will be sent directly
           # if the transfer_type is DIFF, then we will calculate the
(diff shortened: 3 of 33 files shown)


examples/advanced/sklearn-kmeans/kmeans_job_clientapi.py
```diff
@@ -135,7 +135,7 @@ def main():
             script_args=f"--data_path {data_path} "
             f"--train_start {train_start} --train_end {train_end} "
             f"--valid_start {valid_start} --valid_end {valid_end}",
-            params_exchange_format="raw",
+            framework="raw",
```
holgerroth, line 138: Is this example not using "numpy"?

job_templates/sag_gnn/app_2/config_fed_client.conf
```diff
@@ -32,7 +32,7 @@
           heartbeat_timeout = 60
 
           # format of the exchange parameters
-          params_exchange_format =  "numpy"
+          params_exchange_format =  "pytorch"
```
holgerroth, line 35: why change this sag_gnn example?

job_templates/sag_nemo/app_1/config_fed_client.conf
```diff
@@ -35,7 +35,7 @@
           heartbeat_timeout = 60
 
           # format of the exchange parameters
-          params_exchange_format =  "numpy"
+          params_exchange_format =  "pytorch"
```
holgerroth, line 38: Do we need to change these templates?

job_templates/sag_gnn/app_2/config_fed_client.conf
YuanTingHsieh: we need to change them back to 2.5.2 , revert https://github.com/NVIDIA/NVFlare/pull/3319

examples/advanced/sklearn-kmeans/kmeans_job_clientapi.py
YuanTingHsieh: based on this "exchange_format" seems it is not, @ZiyueXu77 can confirm

ZiyueXu77: numpy may also work, I just put raw to make sure no further process needed to modify the content for this

holgerroth:
I tested the bionemo 2 examples with this branch and everything works fine.

ZiyueXu77:
'Model Communication with Tensor' in https://github.com/YuanTingHsieh/NVFlare/blob/use_server_expected_format_26/examples/advanced/llm_hf/README.md
no longer work, because this case is tricky - first round server sends back numpy, while the rest will be tensor

nvflare/app_opt/pt/file_model_persistor.py
```diff
@@ -43,6 +43,7 @@ def __init__(
         source_ckpt_file_full_name=None,
         filter_id: str = None,
         load_weights_only: bool = False,
+        allow_numpy_conversion: bool = True,
```
holgerroth, line 46: need to add a docstring. Why not specify a return format? E.g., `return_format = "pytorch"`

YuanTingHsieh: what about `persist_format`?

PR #3431, opened 2025-04-18
[2.6] Handle None client in ClientRunManager

yanchengnv:
Fixes # .

### Description

The "client" arg of ClientRunManager is set to None when used in Simulator. This PR handles this case and determines the FQSN and is_leaf by using default values.

### Types of changes

- [ ] Documentation updated.


holgerroth (approved):
Works well on my side.

PR #3447, opened 2025-04-23
Async edge training

yanchengnv:
Fixes # .

### Description

This PR creates a framework for support async edge model training using the update-based mechanism.
It also introduces a simplified device simulation solution that does not need web nodes.

### Types of changes

- [ ] Documentation updated.


nvflare/edge/executors/hug.py
```diff
@@ -74,24 +74,36 @@ def __init__(
...
 
         self._msg_handler_registered = {}  # topic => bool
-        self.register_event_handler(EventType.START_RUN, self._handle_start_run)
+        self.register_event_handler(EventType.START_RUN, self._hug_handle_start_run)
         self.register_event_handler(EventType.POST_TASK_ASSIGNMENT_SENT, self._handle_task_sent)
         self.register_event_handler(EventType.POST_TASK_RESULT_RECEIVED, self._handle_result_received)
 
-    def _handle_start_run(self, event_type: str, fl_ctx: FLContext):
+    def get_updater(self, fl_ctx: FLContext) -> Optional[Updater]:
+        return None
```
ZiyueXu77, line 82: don't quite get this, is it a placeholder for future default updater?

nvflare/edge/widgets/device_runner.py
```diff
@@ -0,0 +1,212 @@
...
+
+    def _dr_start_run(self, event_type: str, fl_ctx: FLContext):
+        is_leaf = fl_ctx.get_prop(ReservedKey.IS_LEAF)
+        if not is_leaf:
+            # devices are only for leaf nodes
+            return
+
+        for _ in range(self.num_devices):
+            try:
+                device = self.device_factory.make_device()
```
ZiyueXu77, line 86: Good for initial experiments, thinking about millions device, if initiallized in one-shot (for-loop in 156), not sure whether it will become a burden (interval will be very small if num_devices is huge). We may further simulate more complicated mode for devices "dynamically added" and "join and leave" rather than a constant pool.

ZiyueXu77 (commented):
Good enough to start implementing cifar10 edge training simulation, two minor questions at this point

ZiyueXu77 (approved):
Good enough to get started

PR #3449, opened 2025-04-25
[2.6] Update CIFAR10 to use TBWriter

YuanTingHsieh:
The AnalyticsSender now is a lower level thing and does not have add_scalar.
The add_scalar functionality was moved already into TBWriter.
So we should use TBWriter instead

### Description

- Update CIFAR10 example to use TBWriter

### Types of changes

- [ ] Documentation updated.


holgerroth (approved):
Looks good!

PR #3455, opened 2025-04-28
Add async aggregator

yanchengnv:
Fixes # .

### Description

This PR adds a wrapper aggregator that can be used to perform another aggregator's "accept" method in a separate thread. This can drastically reduce the block time of the SAG workflow when processing client submission if the "accept" processing is time-consuming.

To use this aggregator, the user simply puts the original aggregator as a component in this wrapper aggregator (AsyncAggregator):

"components": [
    {
      "id": "aggregator",
      "path": "nvflare.app_common.aggregators.intime_accumulate_model_aggregator.InTimeAccumulateWeightedAggregator",
      "args": {
        "expected_data_kind": "WEIGHTS"
      }
    },
    {
      "id": "aggr_wrapper",
      "path": "nvflare.app_common.aggregators.async.AsyncAggregator",
      "args": {
        "aggregator_id": "aggregator"
      }
    },
    ...
  ],
  "workflows": [
    {
      "id": "scatter_and_gather",
      "path": "nvflare.app_common.workflows.scatter_and_gather.ScatterAndGather",
      "args": {
        "min_clients": "{min_clients}",
        "num_rounds": 2,
        "start_round": 0,
        "wait_time_after_min_received": "{wait_time}",
        "aggregator_id": "aggr_wrapper",
        "persistor_id": "persistor",
        "shareable_generator_id": "shareable_generator",
        "train_task_name": "train",
        "train_timeout": 6000,
        "ignore_result_error": true
      }
    }
  ]

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/apis/fl_context.py
+++ b/nvflare/apis/fl_context.py
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-
+import copy
 import threading
 from typing import Any, Dict, List
 
@@ -271,6 +271,19 @@ class FLContext(object):
 
         """
         self.props[key] = {V: value, M: make_mask(private, sticky)}
+
+    def clone(self):
+        """Make a copy from self.
+
+        Returns: a new FLContext object
+
+        """
+        with _update_lock:
+            new_ctx = FLContext()
+            new_ctx.model = self.model
+            new_ctx.logger = self.logger
+            new_ctx.props = copy.copy(self.props)  # shallow copy
+            return new_ctx
 
     # implement Context Manager protocol
     def __enter__(self):
--- a/nvflare/app_common/aggregators/async.py
+++ b/nvflare/app_common/aggregators/async.py
@@ -1,0 +1,176 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import queue
+import threading
+
+from nvflare.apis.event_type import EventType
+from nvflare.apis.fl_context import FLContext
+from nvflare.apis.shareable import Shareable
+from nvflare.app_common.abstract.aggregator import Aggregator
+from nvflare.fuel.utils.validation_utils import check_positive_number, check_str
+from nvflare.fuel.utils.waiter_utils import WaiterRC, conditional_wait
+from nvflare.security.logging import secure_format_exception
+
+_SHORT_WAIT = 0.1
+
+
+class _AcceptWaitRC(WaiterRC):
+    END_RUN = 10
+
+
+class _Contribution:
+
+    def __init__(self, data: Shareable, fl_ctx: FLContext):
+        self.data = data
+
+        # We need to make a copy of the fl_ctx since it could be used for multiple contributions and its
+        # content could be overwritten.
+        self.fl_ctx = fl_ctx.clone()
+
+
+class AsyncAggregator(Aggregator):
+
+    def __init__(self, aggregator_id: str, accept_timeout: float = 600.0):
+        """Constructor of AsyncAggregator.
+        AsyncAggregator is a wrapper for other aggregators to do accept processing in a separate thread.
+
+        During a typical SAG-based training, updates from clients are processed by the aggregator's "accept" method.
+        To ensure the integrity of training task, The SAG workflow processes client updates sequentially.
+        If the "accept" method is time-consuming and there are many clients, then the update processing will
+        become bottleneck.
+
+        Using the AsyncAggregator, the updates from clients are simply added to a queue quickly.
+        The actual "accept" processing is done in a separate thread that processes the queued updates sequentially.
+
+        Args:
+            aggregator_id: component ID of the real aggregator
+            accept_timeout: max amount of time to wait for accept to finish
+        """
+        Aggregator.__init__(self)
+        check_str("aggregator_id", aggregator_id)
+        check_positive_number("accept_timeout", accept_timeout)
+
+        self.aggregator_id = aggregator_id
+        self.accept_timeout = accept_timeout
+        self.aggregator = None
+        self.contributions = queue.Queue()
+        self.aggregating = False
+        self.run_ended = False
+        self.accept_done = threading.Event()
+        self.register_event_handler(EventType.START_RUN, self._async_aggr_start_run)
+        self.register_event_handler(EventType.END_RUN, self._async_aggr_end_run)
+
+    def _async_aggr_start_run(self, event_type: str, fl_ctx: FLContext):
+        engine = fl_ctx.get_engine()
+        aggr = engine.get_component(self.aggregator_id)
+        if not isinstance(aggr, Aggregator):
+            self.system_panic(f"component {self.aggregator_id} must be Aggregator but got {type(aggr)}", fl_ctx)
+            return
+
+        if isinstance(aggr, AsyncAggregator):
+            self.system_panic(f"component {self.aggregator_id} must not be AsyncAggregator", fl_ctx)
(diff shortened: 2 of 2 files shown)


nvflare/app_common/aggregators/async.py
```diff
@@ -0,0 +1,176 @@
...
+
+        if rc in [_AcceptWaitRC.ABORTED, _AcceptWaitRC.END_RUN]:
+            self.log_info(fl_ctx, "skipped aggregation since job is aborted")
+            return Shareable()
+
+        if rc != _AcceptWaitRC.IS_SET:
+            self.log_warning(fl_ctx, f"abnormal result {rc} waiting for accept thread")
+
+        # we then call the aggregator to perform actual aggregation
+        result = self.aggregator.aggregate(fl_ctx)
```
YuanTingHsieh, line 159: do we need try-except block here?

nvflare/app_common/aggregators/async.py
```diff
@@ -0,0 +1,176 @@
...
+    def _check_end_run(self):
+        if self.run_ended:
+            return _AcceptWaitRC.END_RUN
+        else:
+            return _AcceptWaitRC.OK
+
+    def reset(self, fl_ctx: FLContext):
+        self.aggregating = False
+        self.accept_done.clear()
+        self.aggregator.reset(fl_ctx)
```
YuanTingHsieh, line 176: do we need try-except block here?

nvflare/app_common/aggregators/async.py
```diff
@@ -0,0 +1,176 @@
...
+            abort_signal=fl_ctx.get_run_abort_signal(),
+            condition_cb=self._check_end_run,
+        )
+
+        if rc in [_AcceptWaitRC.ABORTED, _AcceptWaitRC.END_RUN]:
+            self.log_info(fl_ctx, "skipped aggregation since job is aborted")
+            return Shareable()
+
+        if rc != _AcceptWaitRC.IS_SET:
+            self.log_warning(fl_ctx, f"abnormal result {rc} waiting for accept thread")
```
YuanTingHsieh, line 156: if the accept thread has "abnormal" RC, should we just return here? or raise Exception?

YuanTingHsieh (commented):
Mostly LGTM

nvflare/app_common/aggregators/async.py
```diff
@@ -0,0 +1,176 @@
...
+        accept_thread = threading.Thread(target=self._do_accept, daemon=True)
+        accept_thread.start()
+
+    def _async_aggr_end_run(self, event_type: str, fl_ctx: FLContext):
+        self.run_ended = True
+
+    def accept(self, shareable: Shareable, fl_ctx: FLContext) -> bool:
+        if not self.aggregating:
+            self.log_debug(fl_ctx, "adding contribution to queue")
+            self.contributions.put(_Contribution(shareable, fl_ctx))
```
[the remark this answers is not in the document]
yanchengnv: The size of the queue is unset, so the queue has no size limit.

nvflare/app_common/aggregators/async.py
yanchengnv: I intentionally did not use try-except block to let the exception pop up and to let upper layer handle it.
In case of SAG, this would cause the workflow to end, which is fine.

nvflare/app_common/aggregators/async.py
yanchengnv: Same as above. The idea is that AsyncAggregator is just a wrapper that should not get involved too much in semantic handling.

nvflare/app_common/aggregators/async.py
yanchengnv: This could only happen when the "accept" thread is timed out (e.g. the original aggregator's accept method got stuck). I think you are right that we should raise exception, since there won't be a good way to recover.

nvflare/app_common/aggregators/async.py
```diff
@@ -0,0 +1,176 @@
...
+
+from nvflare.apis.event_type import EventType
+from nvflare.apis.fl_context import FLContext
+from nvflare.apis.shareable import Shareable
+from nvflare.app_common.abstract.aggregator import Aggregator
+from nvflare.fuel.utils.validation_utils import check_positive_number, check_str
+from nvflare.fuel.utils.waiter_utils import WaiterRC, conditional_wait
+from nvflare.security.logging import secure_format_exception
+
+_SHORT_WAIT = 0.1
```
holgerroth: should this be made configurable?

holgerroth:
Nice feature, but should we call it `LazyAggregator` to avoid confusion with asynchronous FL algorithms?

nvflare/app_common/aggregators/lazy.py
```diff
@@ -0,0 +1,176 @@
...
+            accept_timeout: max amount of time to wait for accept to finish
+        """
+        Aggregator.__init__(self)
+        check_str("aggregator_id", aggregator_id)
+        check_positive_number("accept_timeout", accept_timeout)
+
+        self.aggregator_id = aggregator_id
+        self.accept_timeout = accept_timeout
+        self.aggregator = None
+        self.contributions = queue.Queue()
```
YuanTingHsieh, line 67: the queue could grow indefinitely though

nvflare/app_common/aggregators/lazy.py
```diff
@@ -0,0 +1,176 @@
...
+                try:
+                    accepted = self.aggregator.accept(contrib.data, contrib.fl_ctx)
+                    self.log_debug(contrib.fl_ctx, f"{type(self.aggregator)} processed contribution: {accepted=}")
+                except Exception as ex:
+                    self.log_exception(
+                        contrib.fl_ctx,
+                        f"exception from {type(self.aggregator)} when accept: {secure_format_exception(ex)}",
+                    )
+
+            if self.aggregating and self.contributions.empty() and not self.accept_done.is_set():
```
YuanTingHsieh, line 146: the self.aggregating is not thread-safe, there could be race conditions

nvflare/app_common/aggregators/lazy.py
```diff
@@ -0,0 +1,176 @@
...
+            return True
+        else:
+            # when the aggregation is started, we no longer accept new contributions
+            self.log_warning(fl_ctx, "dropped contribution while aggregating")
+            return False
+
+    def _do_accept(self):
+        self.logger.debug("Started accept thread")
+        while True:
+            if self.run_ended:
```
YuanTingHsieh, line 123: the self.run_ended is not thread-safe, there could be race conditions

nvflare/app_common/aggregators/lazy.py
```diff
@@ -0,0 +1,176 @@
...
+    def _check_end_run(self):
+        if self.run_ended:
+            return _AcceptWaitRC.END_RUN
+        else:
+            return _AcceptWaitRC.OK
+
+    def reset(self, fl_ctx: FLContext):
+        self.aggregating = False
+        self.accept_done.clear()
+        self.aggregator.reset(fl_ctx)
```
YuanTingHsieh: should we clear the contributions queue in reset?

nvflare/apis/fl_context.py
```diff
@@ -272,6 +272,19 @@ def put(self, key: str, value, private, sticky):
...
 
+    def clone(self):
+        """Make a copy from self.
+
+        Returns: a new FLContext object
+
+        """
+        with _update_lock:
+            new_ctx = FLContext()
+            new_ctx.model = self.model
```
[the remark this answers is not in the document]
yanchengnv: Not sure. Just didn't want to miss anything.

nvflare/app_common/aggregators/async.py
yanchengnv: Not really. This is simply to make the event wake up soon enough to check other conditions.

nvflare/app_common/aggregators/lazy.py
yanchengnv: Not really, the contributions in the queue are taken out and processed by another thread. And the max size of the queue could only grow to the number of sites.

nvflare/app_common/aggregators/lazy.py
yanchengnv: This is just a simple bool value, and is only set under one condition - the END_RUN event. I don't see any thread safety issue here.

nvflare/app_common/aggregators/lazy.py
yanchengnv: Self.aggregating is only a simple bool value, and is set when the self.aggregate is called, which happens only =in the workflow thread. I don't see any thread safety issue here.

nvflare/app_common/aggregators/lazy.py
yanchengnv: Sure.

ZiyueXu77 (approved):

PR #3456, opened 2025-04-29
[2.6] add more info to address VDR questions

ZiyueXu77:
Fixes questions from VDR interim report .

### Description

More details for HF example (in fact, not related to 2.6)

### Types of changes

- [ ] Documentation updated.


examples/advanced/llm_hf/README.md
```diff
@@ -60,9 +79,10 @@ Note that the `trainer.train()` call is replaced by a `for` loop, and the three
 
 This setting (1 epoch per round) is for simplicity of this example. In practice, we can set the number of rounds and local epoch per round according to the needs: e.g. 2 rounds with 2 epochs per round will result in 4 training epochs in total.
```
YuanTingHsieh, line 80: In this example, we use one epoch per round for simplicity. In practice, the number of rounds and local epochs per round can be adjusted as needed. For example, two rounds with two epochs per round would result in a total of four training epochs.

examples/advanced/llm_hf/README.md
```diff
@@ -60,9 +79,10 @@ Note that the `trainer.train()` call is replaced by a `for` loop, and the three
...
 
-At the beginning of each round, we intentionally load a fixed model weights saved at the beginning, over-writing the previous round's saved model weights, then call `trainer.train(resume_from_checkpoint=True)` with `trainer.args.num_train_epochs` incremented by 1 so that previous logging results are not overwritten. 
+One important requirement from federated learning is that the model weights need to be set to the "global model" at the beginning of each round.
+Therefore, the iterative training needs to explicitly reflect this behavior - at the beginning of each round, we intentionally load a fixed model weights, over-writing the previous round's saved model weights, then call `trainer.train(resume_from_checkpoint=True)` with `trainer.args.num_train_epochs` incremented by 1 so that previous logging results are not overwritten.
```
YuanTingHsieh, line 83: A key requirement in federated learning is that model weights must be reset to the "global model" at the beginning of each round. To implement this, we modify the training process to explicitly load the fixed model weights at the start of each round, overwriting the previous round's model. Then, we call trainer.train(resume_from_checkpoint=True) with trainer.args.num_train_epochs incremented by 1 to ensure that logging results from previous rounds are preserved.

examples/advanced/llm_hf/README.md
```diff
@@ -60,9 +79,10 @@ Note that the `trainer.train()` call is replaced by a `for` loop, and the three
...
+One important requirement from federated learning is that the model weights need to be set to the "global model" at the beginning of each round.
+Therefore, the iterative training needs to explicitly reflect this behavior - at the beginning of each round, we intentionally load a fixed model weights, over-writing the previous round's saved model weights, then call `trainer.train(resume_from_checkpoint=True)` with `trainer.args.num_train_epochs` incremented by 1 so that previous logging results are not overwritten. 
 
-The purpose of doing so is to tell if the intended weights are succesfully loaded at each round. Without using a fixed starting model, even if the model weights are not properly loaded, the training loss curve will still follow the one-call result, which is not what we want to see. 
+The purpose of doing so is to tell if the intended weights are successfully loaded at each round. Without using a fixed starting model, even if the model weights are not properly loaded, the training loss curve will still follow the one-call result, which will not be able to tell us if we have loaded a new model, rather than reusing the existing trainer record.
```
YuanTingHsieh, line 85: This approach helps verify that the correct weights are being loaded at the start of each round. Without resetting to a fixed starting model, the training loss curve may continue to show improvement—even if the weights weren’t properly reloaded—making it hard to tell whether training resumed from the correct model or simply continued from the last round.

examples/advanced/llm_hf/README.md
```diff
@@ -60,9 +79,10 @@ Note that the `trainer.train()` call is replaced by a `for` loop, and the three
...
 
-The purpose of doing so is to tell if the intended weights are succesfully loaded at each round. Without using a fixed starting model, even if the model weights are not properly loaded, the training loss curve will still follow the one-call result, which is not what we want to see. 
+The purpose of doing so is to tell if the intended weights are successfully loaded at each round. Without using a fixed starting model, even if the model weights are not properly loaded, the training loss curve will still follow the one-call result, which will not be able to tell us if we have loaded a new model, rather than reusing the existing trainer record. 
 
 If the intended model weights (serving as the starting point for each round, the "global model" for FL use case) is properly loaded, then we shall observe a "zig-zag" pattern in the training loss curve. This is because the model weights are reset to the same starting point at the beginning of each round, in contrast to the one-shot centralized training, where the model weights are updated continuously, and the training loss curve should follow an overall decreasing trend.
```
YuanTingHsieh, line 87: If the correct "global model" is successfully reloaded each round, the training loss curve will show a characteristic "zig-zag" pattern. This occurs because the model is reset to the same initial weights at the start of each round, unlike in centralized training where weights are updated continuously and the loss typically follows a smooth, decreasing trend.

examples/advanced/llm_hf/sft_job.py
```diff
@@ -73,10 +73,16 @@ def main():
...
-    # First send the model to the server
-    job.to("src/hf_sft_model.py", "server")
-    # Then send the model persistor to the server
-    model_args = {"path": "src.hf_sft_model.CausalLMModel", "args": {"model_name_or_path": model_name_or_path}}
+    if train_mode.lower() == "sft":
+        # First send the model to the server
+        job.to("src/hf_sft_model.py", "server")
+        # Then send the model persistor to the server
+        model_args = {"path": "src.hf_sft_model.CausalLMModel", "args": {"model_name_or_path": model_name_or_path}}
+    elif train_mode.lower() == "peft":
```
YuanTingHsieh, line 81: minor thing, should this job call "sft_job.py"?

YuanTingHsieh (commented):
Mostly LGTM

examples/advanced/llm_hf/sft_job.py
ZiyueXu77: good point! updated to llm_hf_fl_job

YuanTingHsieh (approved):

PR #3473, opened 2025-05-06
Enhance edge simulation

yanchengnv:
Fixes # .

### Description

Consolidated simulation solution so that the same solution can be used as an independent simulator as well as a widget installed to CJ.

### Types of changes

- [ ] Documentation updated.


nvflare/edge/executors/edge_model_executor.py
```diff
@@ -124,7 +123,33 @@ def _make_cookie(model_version, device_selection_id):
             CookieKey.DEVICE_SELECTION_ID: device_selection_id,
         }
 
-    def handle_task_request(self, request: TaskRequest, current_task: TaskInfo, fl_ctx: FLContext) -> TaskResponse:
+    def _handle_selection_request(
```
ZiyueXu77, line 126: don't quite get the idea of selection_request on how it functions beyond task_request

ZiyueXu77 (commented):
good enough to start testing, only one question

nvflare/edge/constants.py
```diff
@@ -60,7 +60,15 @@ class EdgeApiStatus:
     ERROR = "ERROR"
     NO_JOB = "NO_JOB"
     NO_TASK = "NO_TASK"
+    INVALID_REQUEST = "invalid_request"
```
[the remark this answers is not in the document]
yanchengnv: Good question. We probably should establish a convention. Right now, it's all user preference.
Symbol names are usually all upper case. Personally, I like values to be all lower case.
This particular definition happens to get two mixed up. I will make consistent later.

nvflare/edge/executors/edge_model_executor.py
yanchengnv: Good question. This is to make simulator capable of supporting device reuse by the Assessor. When the Assessor reuses an old device, we need to make sure that the old device still has a chance to be selected by the Simulator.
But if the simulator only makes queries for the active devices (which won't include the old device any more), then the old device will never get a chance to be selected.

I solve this problem by always asking for the current selected devices at the beginning of each cycle, and then treat such devices with highest priority. This guarantees that the simulator will always send requests for selected devices.

ZiyueXu77 (approved):

PR #3497, opened 2025-05-20
[2.5] Fix 2.5 Jenkins CI

YuanTingHsieh:
### Description

Fix 2.5 Jenkins CI

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh:
there is some security implications by using load_weights_only = False,
perhaps we need to revisit it later.

we could do something like this: https://github.com/huggingface/transformers/pull/34632/files

PR #3506, opened 2025-05-22
Feature/add fedavg metric optrimization controller

rbagan:
Fixes # .

### Description

Hi all,

This pull request adds a new controller: **FedAvg Metric Optimizaiton**. This controller is the result of a joint effort between Roche and Universitätsspital Zurich (USZ) to train a model in a federated manner.

Purpose of the controller:
The goal is to obtain the best possible model by optimizing a specific metric (e.g., minimizing the loss or maximizing the F-Score) and to stop the training if the tracked metric does not improve after a certain number of FL rounds, as defined by the researcher. This approach saves computation time during FL training, especially when the model is large and requires a significant amount of data. This controller has been developed in a paper that is currently under peer review. 

Additionally, we wanted to provide the option to choose whether to optimize the metric during training or validation.

I would like to highlight that this contribution is thanks to Roche and the Universitätsspital Zurich (USZ)

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/examples/advanced/metric_optrimization/README.md
+++ b/examples/advanced/metric_optrimization/README.md
@@ -1,0 +1,37 @@
+Hello FedAvg Metric Optimization
+================================
+
+In this example we will introduce the FedAvg Metric Optimization. The
+FedAvgMetricOptimization class combines two techniques—early stopping and metric
+optimization (minimization or maximization)—to identify the best model during FL
+learning. It uses a patience parameter, which specifies how many FL rounds to wait
+without any metric improvement before stopping the training.
+
+1. Setup
+
+```cmd
+pip install nvflare torch torchvision tensorboard
+```
+
+2. PTFedAvgMetricOptimization using ModelController API
+
+The ModelController API enables the option to easily customize a workflow with Python code.
+
+* FedAvg: We subclass the BaseFedAvg class to leverage the predefined aggregation functions.
+* Metric Optimization: We have added a set of new arguments, most of which are optional, to provide greater flexibility in configuring the metric optimization:
+    - `target_metric_name`: specifies the name of the metric to optimize
+    - `optimization mode`: choose whether to minimize or maximize the metric
+    - `task_train_name`: specifies the name of the training task
+    - `task_validation_name`: specifies the name of the validation task
+    - `task_to_optimize`: indicates whether to apply metric optimization to the training or validation task
+    - `patience`: defines the number of FL rounds to wait without improvement before stopping the training
+* Model Selection: As and alternative to using a IntimeModelSelector componenet for model selection, we instead compare the metrics of the models in the workflow to select the best model each round.
+* Saving/Loading: Rather than configuring a persistor such as PTFileModelPersistor component, we choose to utilize PyTorch's save and load functions and save the metadata of the FLModel separately.
+
+3. Run the script
+
+Use the Job API to define and run the example with the simulator:
+
+```python
+python pt_fedavg_metric_optimization_script.py
+```
--- a/examples/advanced/metric_optrimization/pt_fedavg_metric_optimization_script.py
+++ b/examples/advanced/metric_optrimization/pt_fedavg_metric_optimization_script.py
@@ -1,0 +1,45 @@
+# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from src.net import Net
+
+from nvflare import FedJob
+from fedavg_metric_optimization import PTFedAvgMetricOptimization
+from nvflare.job_config.script_runner import ScriptRunner
+
+if __name__ == "__main__":
+    n_clients = 3
+    num_rounds = 15
+    train_script = "src/cifar_fl.py"
+
+    job = FedJob(name="FedAvg_Metric_Optimization")
+
+    # Define the controller workflow and send to server
+    controller = PTFedAvgMetricOptimization(
+        num_clients=n_clients,
+        num_rounds=num_rounds,
+        target_metric_name="accuracy",
+        optimization_mode="max",
+        patience=3,
+        initial_model=Net(),
+    )
+    job.to(controller, "server")
+
+    # Add clients
+    for i in range(n_clients):
+        executor = ScriptRunner(script=train_script, script_args="")
+        job.to(executor, f"site-{i}")
+
+    # job.export_job("/tmp/nvflare/jobs/job_config")
+    job.simulator_run("jobs/workdir", gpu="0")
--- a/examples/advanced/metric_optrimization/src/cifar_fl.py
+++ b/examples/advanced/metric_optrimization/src/cifar_fl.py
@@ -1,0 +1,208 @@
+# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+
+import torch
+import torch.nn as nn
+import torch.optim as optim
+import torchvision
+import torchvision.transforms as transforms
+from net import Net
+
+# (1) import nvflare client API
+import nvflare.client as flare
+from nvflare.app_common.app_constant import ModelName
+
+# (optional) set a fix place so we don't need to download everytime
+CIFAR10_ROOT = "data/cifar10"
+# (optional) We change to use GPU to speed things up.
+# if you want to use CPU, change DEVICE="cpu"
+DEVICE = "cuda:0"
+
+
+def define_parser():
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--dataset_path", type=str, default=CIFAR10_ROOT, nargs="?")
+    parser.add_argument("--batch_size", type=int, default=4, nargs="?")
+    parser.add_argument("--num_workers", type=int, default=1, nargs="?")
+    parser.add_argument("--local_epochs", type=int, default=2, nargs="?")
+    parser.add_argument(
+        "--model_path", type=str, default=f"{CIFAR10_ROOT}/cifar_net.pth", nargs="?"
+    )
+    return parser.parse_args()
+
+
+def main():
+    # define local parameters
+    args = define_parser()
+
+    dataset_path = args.dataset_path
+    batch_size = args.batch_size
+    num_workers = args.num_workers
+    local_epochs = args.local_epochs
+    model_path = args.model_path
+
+    transform = transforms.Compose(
+        [transforms.ToTensor(), transforms.Normalize((0.5, 0.5, 0.5), (0.5, 0.5, 0.5))]
+    )
+    trainset = torchvision.datasets.CIFAR10(
+        root=dataset_path, train=True, download=True, transform=transform
+    )
+    trainloader = torch.utils.data.DataLoader(
+        trainset, batch_size=batch_size, shuffle=True, num_workers=num_workers
+    )
+    testset = torchvision.datasets.CIFAR10(
+        root=dataset_path, train=False, download=True, transform=transform
+    )
+    testloader = torch.utils.data.DataLoader(
+        testset, batch_size=batch_size, shuffle=False, num_workers=num_workers
+    )
+
+    net = Net()
+    best_accuracy = 0.0
+
+    # wraps evaluation logic into a method to re-use for
+    #       evaluation on both trained and received model
+    def evaluate(input_weights):
+        net = Net()
+        net.load_state_dict(input_weights)
+        # (optional) use GPU to speed things up
+        net.to(DEVICE)
+
+        correct = 0
+        total = 0
+        # since we're not training, we don't need to calculate the gradients for our outputs
+        with torch.no_grad():
+            for data in testloader:
+                # (optional) use GPU to speed things up
+                images, labels = data[0].to(DEVICE), data[1].to(DEVICE)
+                # calculate outputs by running images through the network
+                outputs = net(images)
+                # the class with the highest energy is what we choose as prediction
+                _, predicted = torch.max(outputs.data, 1)
+                total += labels.size(0)
+                correct += (predicted == labels).sum().item()
+
+        return 100 * correct // total
+
+    # (2) initialize NVFlare client API
+    flare.init()
+
+    # (3) run continuously when launch_once=true
+    while flare.is_running():
+
+        # (4) receive FLModel from NVFlare
+        input_model = flare.receive()
+        client_id = flare.get_site_name()
+
+        # Based on different "task" we will do different things
+        # for "train" task (flare.is_train()) we use the received model to do training and/or evaluation
+        # and send back updated model and/or evaluation metrics, if the "train_with_evaluation" is specified as True
+        # in the config_fed_client we will need to do evaluation and include the evaluation metrics
+        # for "evaluate" task (flare.is_evaluate()) we use the received model to do evaluation
+        # and send back the evaluation metrics
+        # for "submit_model" task (flare.is_submit_model()) we just need to send back the local model
+        # (5) performing train task on received model
+        if flare.is_train():
+            print(
+                f"({client_id}) current_round={input_model.current_round}, total_rounds={input_model.total_rounds}"
+            )
+
+            # (5.1) loads model from NVFlare
+            net.load_state_dict(input_model.params)
+
+            criterion = nn.CrossEntropyLoss()
+            optimizer = optim.SGD(net.parameters(), lr=0.001, momentum=0.9)
+
+            # (optional) use GPU to speed things up
+            net.to(DEVICE)
+            # (optional) calculate total steps
+            steps = local_epochs * len(trainloader)
+            for epoch in range(local_epochs):  # loop over the dataset multiple times
+
+                running_loss = 0.0
+                for i, data in enumerate(trainloader, 0):
+                    # get the inputs; data is a list of [inputs, labels]
+                    # (optional) use GPU to speed things up
+                    inputs, labels = data[0].to(DEVICE), data[1].to(DEVICE)
+
+                    # zero the parameter gradients
+                    optimizer.zero_grad()
+
+                    # forward + backward + optimize
+                    outputs = net(inputs)
+                    loss = criterion(outputs, labels)
+                    loss.backward()
+                    optimizer.step()
+
+                    # print statistics
+                    running_loss += loss.item()
+                    if i % 2000 == 1999:  # print every 2000 mini-batches
+                        print(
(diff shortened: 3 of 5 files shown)


examples/advanced/metric_optrimization/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+* FedAvg: We subclass the BaseFedAvg class to leverage the predefined aggregation functions.
+* Metric Optimization: We have added a set of new arguments, most of which are optional, to provide greater flexibility in configuring the metric optimization:
+    - `target_metric_name`: specifies the name of the metric to optimize
+    - `optimization mode`: choose whether to minimize or maximize the metric
+    - `task_train_name`: specifies the name of the training task
+    - `task_validation_name`: specifies the name of the validation task
+    - `task_to_optimize`: indicates whether to apply metric optimization to the training or validation task
+    - `patience`: defines the number of FL rounds to wait without improvement before stopping the training
+* Model Selection: As and alternative to using a IntimeModelSelector componenet for model selection, we instead compare the metrics of the models in the workflow to select the best model each round.
```
holgerroth: typo: `componenet` -> `component`

nvflare/app_opt/pt/fedavg_metric_optimization.py
```diff
@@ -0,0 +1,284 @@
+# Copyright (c) 2024, NVIDIA CORPORATION.  All rights reserved.
```
holgerroth: Please use 2025 in the license files.

examples/advanced/metric_optrimization/pt_fedavg_metric_optimization_script.py
```diff
@@ -0,0 +1,45 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from src.net import Net
+
+from nvflare import FedJob
+from fedavg_metric_optimization import PTFedAvgMetricOptimization
```
holgerroth: This should be `from nvflare.app_opt.pt. fedavg_metric_optimization import PTFedAvgMetricOptimization`

examples/advanced/metric_optrimization/src/cifar_fl.py
```diff
@@ -0,0 +1,208 @@
...
+from net import Net
+
+# (1) import nvflare client API
+import nvflare.client as flare
+from nvflare.app_common.app_constant import ModelName
+
+# (optional) set a fix place so we don't need to download everytime
+CIFAR10_ROOT = "data/cifar10"
+# (optional) We change to use GPU to speed things up.
+# if you want to use CPU, change DEVICE="cpu"
```
holgerroth: We typically use this so the code automatically works with CPU or GPU
```
# If available, we use GPU to speed things up.
DEVICE = "cuda" if torch.cuda.is_available() else "CPU"
```

rbagan:
Dear @chesterxgchen, 

Thank you very much for your feedback.  

> 1. the files needed to formatted in certain way to pass the unit tests.

I will apply the script you mentioned and upload the changes. Thank you very much.

> 2. The proposed the PR is very similar to this controller. fedavg_early_stopping.py
>    Which based on the condition expressed such as: "accuracy > 0.8"
>    https://github.com/NVIDIA/NVFlare/blob/main/nvflare/app_opt/pt/fedavg_early_stopping.py
>    The client script for the this controller is in the example: https://github.com/NVIDIA/NVFlare/blob/main/examples/hello-world/hello-fedavg/pt_fedavg_early_stopping_script.py
> 
> Please see if your work has additional improvement beyond the fedavg_early_stopping controller.

I'm glad you asked this. The needs of our project cannot be met with the current `pt_fedavg_early_stopping_script` implementation.

The design of `pt_fedavg_early_stopping_script` limits the ability to apply more sophisticated early stopping criteria. For example, the PyTorch Lightning framework offers a more advanced version of this mechanism (see: https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.callbacks.EarlyStopping.html), providing several additional options to control when training should stop. Here are two examples of useful arguments:

* mode: one of 'min', 'max'. In 'min' mode, training will stop when the quantity monitored has stopped decreasing and in 'max' mode it will stop when the quantity monitored has stopped increasing.
* patience: number of checks with no improvement after which training will be stopped. Under the default configuration, one check happens after every training epoch. However, the frequency of validation can be modified by setting various parameters on the Trainer, for example check_val_every_n_epoch and val_check_interval.s.

We began by modifying the `pt_fedavg_early_stopping_script` to include these new arguments, but ultimately developed a completely new module. We could rename this module to something like `pt_fedavg_early_stopping_extended_script`, for example.

We believe these two modules can coexist: a basic version of early stopping (`pt_fedavg_early_stopping_script`) and a more "advanced" version (`pt_fedavg_early_stopping_extended_script`)

Hope this answers your question!

Dear @holgerroth,

Thank you very much for your feedback, I will address all the findings.

chesterxgchen:
> Dear @chesterxgchen,
> 
> Thank you very much for your feedback.
> 
> > 1. the files needed to formatted in certain way to pass the unit tests.
> 
> I will apply the script you mentioned and upload the changes. Thank you very much.
> 
> > 2. The proposed the PR is very similar to this controller. fedavg_early_stopping.py
> >    Which based on the condition expressed such as: "accuracy > 0.8"
> >    https://github.com/NVIDIA/NVFlare/blob/main/nvflare/app_opt/pt/fedavg_early_stopping.py
> >    The client script for the this controller is in the example: https://github.com/NVIDIA/NVFlare/blob/main/examples/hello-world/hello-fedavg/pt_fedavg_early_stopping_script.py
> > 
> > Please see if your work has additional improvement beyond the fedavg_early_stopping controller.
> 
> I'm glad you asked this. The needs of our project cannot be met with the current `pt_fedavg_early_stopping_script` implementation.
> 
> The design of `pt_fedavg_early_stopping_script` limits the ability to apply more sophisticated early stopping criteria. For example, the PyTorch Lightning framework offers a more advanced version of this mechanism (see: https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.callbacks.EarlyStopping.html), providing several additional options to control when training should stop. Here are two examples of useful arguments:
> 
> * mode: one of 'min', 'max'. In 'min' mode, training will stop when the quantity monitored has stopped decreasing and in 'max' mode it will stop when the quantity monitored has stopped increasing.
> * patience: number of checks with no improvement after which training will be stopped. Under the default configuration, one check happens after every training epoch. However, the frequency of validation can be modified by setting various parameters on the Trainer, for example check_val_every_n_epoch and val_check_interval.s.
> 
> We began by modifying the `pt_fedavg_early_stopping_script` to include these new arguments, but ultimately developed a completely new module. We could rename this module to something like `pt_fedavg_early_stopping_extended_script`, for example.
> 
> We believe these two modules can coexist: a basic version of early stopping (`pt_fedavg_early_stopping_script`) and a more "advanced" version (`pt_fedavg_early_stopping_extended_script`)
> 
> Hope this answers your question!
> 
> Dear @holgerroth,
> 
> Thank you very much for your feedback, I will address all the findings.

@rbagan  thank you for the detailed reply and the reason behind the changes. 

I don't mind an extra controller, but I am not completely agree with the design approach. 

Frankly, this was the approach we initially took and decided to abandoned as it was not general enough. 
Initially, we were thinking having a "mode" or "direction" to indicating how the metric is compared. 
For example, loss we have to use "greater or equal";  for accuracy we like to have "less or equal" as comparsion. 
This leads a lot of input parameters "metrics", "mode", and comparitor. 












> Dear @chesterxgchen,
> 
> Thank you very much for your feedback.
> 
> > 1. the files needed to formatted in certain way to pass the unit tests.
> 
> I will apply the script you mentioned and upload the changes. Thank you very much.
> 
> > 2. The proposed the PR is very similar to this controller. fedavg_early_stopping.py
> >    Which based on the condition expressed such as: "accuracy > 0.8"
> >    https://github.com/NVIDIA/NVFlare/blob/main/nvflare/app_opt/pt/fedavg_early_stopping.py
> >    The client script for the this controller is in the example: https://github.com/NVIDIA/NVFlare/blob/main/examples/hello-world/hello-fedavg/pt_fedavg_early_stopping_script.py
> > 
> > Please see if your work has additional improvement beyond the fedavg_early_stopping controller.
> 
> I'm glad you asked this. The needs of our project cannot be met with the current `pt_fedavg_early_stopping_script` implementation.
> 
> The design of `pt_fedavg_early_stopping_script` limits the ability to apply more sophisticated early stopping criteria. For example, the PyTorch Lightning framework offers a more advanced version of this mechanism (see: https://lightning.ai/docs/pytorch/stable/api/lightning.pytorch.callbacks.EarlyStopping.html), providing several additional options to control when training should stop. Here are two examples of useful arguments:
> 
> * mode: one of 'min', 'max'. In 'min' mode, training will stop when the quantity monitored has stopped decreasing and in 'max' mode it will stop when the quantity monitored has stopped increasing.
> * patience: number of checks with no improvement after which training will be stopped. Under the default configuration, one check happens after every training epoch. However, the frequency of validation can be modified by setting various parameters on the Trainer, for example check_val_every_n_epoch and val_check_interval.s.
> 
> We began by modifying the `pt_fedavg_early_stopping_script` to include these new arguments, but ultimately developed a completely new module. We could rename this module to something like `pt_fedavg_early_stopping_extended_script`, for example.
> 
> We believe these two modules can coexist: a basic version of early stopping (`pt_fedavg_early_stopping_script`) and a more "advanced" version (`pt_fedavg_early_stopping_extended_script`)
> 
> Hope this answers your question!
> 
> Dear @holgerroth,
> 
> Thank you very much for your feedback, I will address all the findings.

@rbagan  thank you for your detailed reply and reason behind of creating a new controller.  we welcome the contribution and don't mind having extra controller if user feel thats easier to understand. 

Unfortunately, I disagree with the design decision, as thats what we decided not to do in the early stopping controller.  Initially, we also considering how to express such conditions such as "loss > 90" or "accuracy < 0.3",  we would have to introduce at least two parameters: 1) metric (example: loss or accuracy) 2) direction (or mode) " use "<" or ">". 
All these are fine, but not general enough.  What if we decided to use more than one metrics in the comparison condition ? or other condition. For example 

loss > 80 or accuracy < 0.3  

accuracy < 0.5 and min_delta > 0.01 

none of these can be expressed in the proposed PR. Therefore, we need more general stop_condition function. Thats why we choose use a early stop function rather than simple "min"/"max" mode. Also use math simple ">=" or "<" are much easier to express than just max or min mode. 

I think the using more input parameters such as min_delta or patients are good for the users, and  it might be easier to map to the lightning etc. but the conditional expression is much more powerfully than the current "mode" option. 

I actually prefer you improve the early_stop controller by adding the missing arguments such as patients or others parameters than create an new one that is less powerful. 

But since I am not the end users. I might be biased on this one.

rbagan:
Dear @chesterxgchen ,

> Unfortunately, I disagree with the design decision, as thats what we decided not to do in the early stopping controller. Initially, we also considering how to express such conditions such as "loss > 90" or "accuracy < 0.3", we would have to introduce at least two parameters: 1) metric (example: loss or accuracy) 2) direction (or mode) " use "<" or ">". All these are fine, but not general enough. What if we decided to use more than one metrics in the comparison condition ? or other condition. For example
> 
> loss > 80 or accuracy < 0.3
> 
> accuracy < 0.5 and min_delta > 0.01
> 
> none of these can be expressed in the proposed PR. Therefore, we need more general stop_condition function. Thats why we choose use a early stop function rather than simple "min"/"max" mode. Also use math simple ">=" or "<" are much easier to express than just max or min mode.
> 
> I think the using more input parameters such as min_delta or patients are good for the users, and it might be easier to map to the lightning etc. but the conditional expression is much more powerfully than the current "mode" option.
> 
> I actually prefer you improve the early_stop controller by adding the missing arguments such as patients or others parameters than create an new one that is less powerful.
> 
> But since I am not the end users. I might be biased on this one.

I'm flexible and can adapt the contribution to best suit the NVFlare framework. I will rework the merge request and update it to reflect the changes. 

Thanks.

chesterxgchen:
@rbagan  thank you for the flexibility and contribution

rbagan:
Dear @chesterxgchen and @holgerroth,

I have updated the pull request with the reworked implementation. I believe these changes improve the Early Stopping module and add new features, including `patience` and `task_to_optimize`.

- patience (optional): Defines the number of checks with no improvement after which federated learning will stop. If set to None, this option has no effect.

- task_to_optimize (optional): Allows the user to specify where early stopping should be applied—training or validation. The task name is passed as an argument to the `send_model_and_wait` function. By default, it is set to "train."

During this rework, I also fixed a critical bug: previously, `best_model = curr_model` caused both variables to reference the same object in memory. As a result, when evaluating the stop condition, comparisons were always made against the same variable, making the condition always True. To address this, I minimized changes and avoided relying on a separate `best_model` object, thus maintaining efficiency without introducing a deep copy operation.

Breaking changes:

- Removed the class method `select_best_model`.

- Changed the signature of the `is_curr_model_better` method.

My intention was to avoid breaking changes, but introducing a deep copy for `best_model` would have added unnecessary computational overhead.

I have manually tested the code to ensure it works as expected.

I hope these changes meet your expectations. I’m happy to hear your feedback.

Thank you.

chesterxgchen:
@holgerroth this link is broken, not sure why causing this PR check to fail. 

## Errors per input

### Errors in /home/runner/work/NVFlare/NVFlare/examples/advanced/bionemo/README.md

* [404] [https://docs.nvidia.com/bionemo-framework/latest/user-guide/](https://docs.nvidia.com/bionemo-framework/latest/user-guide/) | Network error: Not Found

chesterxgchen:
@rbagan I approved the PR, 

but then I realized the following: 
1) I have showed the the early stopping FedAvg in the [tutorials](https://github.com/NVIDIA/NVFlare/blob/main/examples/tutorials/self-paced-training/part-1_federated_learning_introduction/chapter-1_running_federated_learning_applications/01.3_server_side_customization/customize_server_logics.ipynb) 

I assume that we don't need to update these examples ( as they are showing as custom code), but it worth double check

2) the stop condition logic maybe limited. 

although the stop_condition is general, but my original implementation is actually not allowed to have more than one metric such as 

``` loss > 80 or accuracy < 0.3```
```accuracy < 0.5 and min_delta > 0.01```
In above case, the stop_condition can be expressed by multiple conditions separate by "or" or "and"
cond1 or cond2 
cond1 and cond2
cond1 and cond2 and cond3 

as ```parse_compare_criteria``` only take one condition  ( A op B) 

to do this, we have to parse the expression into an AST, here is an example
```
import ast
from typing import Dict, Any

class ConditionEvaluator:
    def __init__(self, variables: Dict[str, Any] = None):
        self.variables = variables or {}
        self.comparison_ops = {
            ast.Eq: lambda a, b: a == b,
            ast.NotEq: lambda a, b: a != b,
            ast.Lt: lambda a, b: a < b,
            ast.LtE: lambda a, b: a <= b,
            ast.Gt: lambda a, b: a > b,
            ast.GtE: lambda a, b: a >= b,
            ast.In: lambda a, b: a in b,
            ast.NotIn: lambda a, b: a not in b
        }
        
    def evaluate_condition(self, condition: str) -> bool:
        """Evaluate a condition string with variables."""
        try:
            tree = ast.parse(condition, mode='eval')
            return self._evaluate_node(tree.body)
        except Exception as e:
            raise ValueError(f"Error evaluating condition '{condition}': {str(e)}")
    
    def _evaluate_node(self, node):
        """Recursively evaluate AST nodes."""
        if isinstance(node, ast.Compare):
            return self._evaluate_comparison(node)
        elif isinstance(node, ast.BoolOp):
            return self._evaluate_boolop(node)
        elif isinstance(node, ast.UnaryOp):
            return self._evaluate_unaryop(node)
        elif isinstance(node, ast.Name):
            return self._get_variable(node.id)
        elif isinstance(node, ast.Constant):
            return node.value
        else:
            raise ValueError(f"Unsupported node type: {type(node).__name__}")
    
    def _evaluate_comparison(self, node):
        """Evaluate comparison operations."""
        left = self._evaluate_node(node.left)
        
        # Handle chained comparisons (e.g., 1 < x < 10)
        if len(node.ops) > 1:
            values = [self._evaluate_node(comp) for comp in node.comparators]
            results = []
            current_left = left
            
            for op, right in zip(node.ops, values):
                if not isinstance(op, tuple(COMPARISON_OPS.keys())):
                    raise ValueError(f"Unsupported comparison operator: {op.__class__.__name__}")
                result = COMPARISON_OPS[type(op)](current_left, right)
                results.append(result)
                current_left = right
                
            return all(results)
        else:
            right = self._evaluate_node(node.comparators[0])
            op = node.ops[0]
            
            if isinstance(op, tuple(COMPARISON_OPS.keys())):
                return COMPARISON_OPS[type(op)](left, right)
            else:
                raise ValueError(f"Unsupported comparison operator: {op.__class__.__name__}")
    
    def _evaluate_boolop(self, node):
        """Evaluate boolean operations (and/or)."""
        values = [self._evaluate_node(value) for value in node.values]
        
        if isinstance(node.op, ast.And):
            return all(values)
        elif isinstance(node.op, ast.Or):
            return any(values)
        else:
            raise ValueError(f"Unsupported boolean operator: {node.op.__class__.__name__}")
    
    def _evaluate_unaryop(self, node):
        """Evaluate unary operations (not)."""
        if isinstance(node.op, ast.Not):
            return not self._evaluate_node(node.operand)
        else:
            raise ValueError(f"Unsupported unary operator: {node.op.__class__.__name__}")
    
    def _get_variable(self, name: str):
        """Get variable value, raising an error if not found."""
        if name not in self.variables:
            raise ValueError(f"Variable '{name}' not found in context")
        return self.variables[name]

# Example usage:
if __name__ == "__main__":
    # Define your metrics
    metrics = {
        'accuracy': 92.5,
        'loss': 85.3,
        'minor_delta': 0.3,
        'epoch': 10
    }
    
    # Create evaluator with metrics
    evaluator = ConditionEvaluator(metrics)
    
    # Test cases
    test_cases = [
        "accuracy > 90",
        "loss < 100 and minor_delta < 0.5",
        "(accuracy > 95 or loss < 80) and minor_delta < 0.4",
        "accuracy > 90 and loss < 100 and minor_delta < 0.5",
        "epoch > 5 and (accuracy > 92 or loss < 80)"
    ]
    
    for condition in test_cases:
        try:
            result = evaluator.evaluate_condition(condition)
            print(f"Condition: {condition:50} => {result}")
        except ValueError as e:
            print(f"Error in '{condition}': {str(e)}")

```
It would be ideal, we could incorporate something like this into this. But It would be ok we can do this in another PR.

**Comment by rbagan** [comment]
Dear @chesterxgchen ,

Thank you for your feedback. I have updated the ipynb to reflect the latest changes. Regarding the inclusion of new logic for more complex expressions, we can create a separate pull request for this. It would require additional effort and in-depth testing to ensure everything works as expected

File: examples/hello-world/hello-fedavg/hello-fedavg.ipynb [hunk 5]
```diff
@@ -29,7 +29,7 @@
    "metadata": {},
    "outputs": [],
    "source": [
-    "! pip install nvflare~=2.5.0rc torch torchvision tensorboard"
+    "! pip install nvflare==2.6 torch torchvision tensorboard"
```
chesterxgchen: ```nvflare==2.6```

probably should be 

```nvflare>=2.6```
rather than fix to the 2.6 version

chesterxgchen:
> Dear @chesterxgchen ,
> 
> Thank you for your feedback. I have updated the ipynb to reflect the latest changes. Regarding the inclusion of new logic for more complex expressions, we can create a separate pull request for this. It would require additional effort and in-depth testing to ensure everything works as expected

I did not see the ipynb file changes in the PR, did you forget to check-in ?  make sure you clean up the notebook output before check-in the notebook

examples/hello-world/hello-fedavg/hello-fedavg.ipynb
rbagan: Done, thanks!

rbagan:
Dear @chesterxgchen ,

You should see the updated notebook (.ipynb file) in the list of changes; the version has been fixed.

chesterxgchen:
@rbagan  sorry I did not mention earlier ( or maybe I did), there is another tutorial that use the same logics. 
https://github.com/NVIDIA/NVFlare/tree/main/examples/tutorials/self-paced-training/part-1_federated_learning_introduction/chapter-1_running_federated_learning_applications/01.3_server_side_customization/code/src

In this case, there are multiple versions (v1, v2).  do you want to make the same changes to fix anything thats incorrect to make them consistent ?

rbagan:
Dear @chesterxgchen ,

Updated, thanks!

PR #3509, opened 2025-05-23
Truncate the name of certificate subject longer than 64 characters before generating the certificate

### LINKED ISSUE #3508: [BUG] Using a long server name (>64) in Dashboard UI causes failure in generating server certificate and downloading startup kit
Similar to #3093 

**Describe the bug**
In NVFlare Dashboard, if using a server name longer than 64 characters, the certificate fails to generate as well as console and startup kit.  The zip with Flare Console and the zip with Server Startup Kit cannot be downloaded.

**Traceback**
```
  File "/home/stevo/.pyenv/versions/3.9.19/envs/ai4os-nvflare-2.6.0/lib/python3.9/site-packages/nvflare/lighter/provisioner.py", line 110, in provision
    b.build(project, ctx)
  File "/home/stevo/.pyenv/versions/3.9.19/envs/ai4os-nvflare-2.6.0/lib/python3.9/site-packages/nvflare/lighter/impl/cert.py", line 246, in build
    self._build_write_cert_pair(server, CertFileBasename.SERVER, ctx)
  File "/home/stevo/.pyenv/versions/3.9.19/envs/ai4os-nvflare-2.6.0/lib/python3.9/site-packages/nvflare/lighter/impl/cert.py", line 176, in _build_write_cert_pair
    pri_key, cert = self.get_pri_key_cert(participant)
  File "/home/stevo/.pyenv/versions/3.9.19/envs/ai4os-nvflare-2.6.0/lib/python3.9/site-packages/nvflare/lighter/impl/cert.py", line 267, in get_pri_key_cert
    cert = self._generate_cert(
  File "/home/stevo/.pyenv/versions/3.9.19/envs/ai4os-nvflare-2.6.0/lib/python3.9/site-packages/nvflare/lighter/impl/cert.py", line 299, in _generate_cert
    return generate_cert(
  File "/home/stevo/.pyenv/versions/3.9.19/envs/ai4os-nvflare-2.6.0/lib/python3.9/site-packages/nvflare/lighter/utils.py", line 52, in generate_cert
    x509_subject = x509_name(subject.name, subject.org, subject.role)
  File "/home/stevo/.pyenv/versions/3.9.19/envs/ai4os-nvflare-2.6.0/lib/python3.9/site-packages/nvflare/lighter/utils.py", line 117, in x509_name
    name = [x509.NameAttribute(NameOID.COMMON_NAME, cn_name)]
  File "/home/stevo/.pyenv/versions/3.9.19/envs/ai4os-nvflare-2.6.0/lib/python3.9/site-packages/cryptography/x509/name.py", line 162, in __init__
    raise ValueError(msg)
ValueError: Attribute's length must be >= 1 and <= 64, but it was 78
```

**To Reproduce**
1. Configure a project in NVFlare Dashboard UI.
2. In `Server Configuration` screen, set the `Server (DNS name)` longer than 64 characters.
3. Freeze project
4. click `Download FLARE Console`
5. click `Dowload Server Startup Kit (primary)`

**Expected behavior**
The requested zip file starts downloading.

 - OS: [ubuntu 22.04]
 - Python Version [3.9.19]
 - NVFlare Version [2.6]

Stifo:
Fixes #3508

### Description

There's a limit on the length of [COMMON_NAME](https://github.com/pyca/cryptography/blob/a7733878281ca261c4ada04022fc706ba5de9d8b/src/cryptography/x509/name.py#L65) in NameAttribute in [cryptoography v43.0.x](https://github.com/pyca/cryptography/blob/a7733878281ca261c4ada04022fc706ba5de9d8b/src/cryptography/x509/name.py#L141). If using name longer than 64 characters, the certificate fails to generate.

This fix truncates name longer than 64 characters before generating certificate.

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh:
@Stifo thanks for the issue and the PR!

Do you think we should either (1) raise the Exception or (2) as you did trim it down?

What do you think the pros and cons?

Stifo:
@YuanTingHsieh Good question! It made me look at the problem from another perspective. Well, it depends on how is the certificate and its common name attribute (CN) utilized.

In the context of HTTPS, truncated DNS name in CN can't be used to validate the subject identity. However, the identity can be still validated using the subject alternative name (SAN), which has no upper bound limit as CN.

In the context of NVFlare, CN has to be unique: ["Each identity is defined with a unique common name."](https://nvflare.readthedocs.io/en/2.6/user_guide/security/identity_security.html). In our usecase, we are using long DNS names for FL servers; e.g., `0f8d6ec6-2e90-11f0-b9c9-0242ac120004.iisas-deployments.cloud.ai4eosc.eu`. The subdomain part contains a unique ID, so if we truncate the CN, we're good. But this could be a problem when two different DNS names are truncated to the same string.

Perhaps a better solution would be to utilize the SAN attribute to hold original DNS names.

Anyway, for our use case, it is better to truncate CN than raise an exception and be stuck on not be able to use NVFlare. Currently, we're on custom NVFlare v2.5 with this fix and it's working fine. But we would like to use official releases in the future.

YuanTingHsieh:
@Stifo thanks for more information!
@yanchengnv thanks for points out similar issues.

So I guess the problem is that as of now (2.6) "host_names" can't be specified in the Dashboard?

@nvkevlu can you confirm?

Then before we add that support, should we merge this PR?

yanchengnv:
> @Stifo thanks for more information! @yanchengnv thanks for points out similar issues.
> 
> So I guess the problem is that as of now (2.6) "host_names" can't be specified in the Dashboard?
> 
> @nvkevlu can you confirm?
> 
> Then before we add that support, should we merge this PR?

True, currently Dashboard only provides very basic provisioning functions - none of the advanced features (including host_names) is supported.

YuanTingHsieh:
@Stifo thanks for your PR and issue, we fixed it in #3528

PR #3513, opened 2025-06-04
Improvements on edge Algo, Impl, and Exp

ZiyueXu77:
Fixes # .

### Description

Further improvements and bug fix on edge algorithm and implementation:
- abstraction of model_manager and device_manage to separate the two logic handled by assessor
- details of async algorithms with global_lr
- simulation of device computation heterogeneity 
- experiments and results 
- fix aggregation logic to clear an aggregator once it reports the aggr results to upper layer / applies the report to global model

### Types of changes

- [ ] Documentation updated.


examples/advanced/edge/jobs/configs/cifar10_device_config.json
```diff
@@ -9,9 +9,9 @@
     "path": "nvflare.edge.simulation.devices.pt_cifar10.PTCifar10Processor",
     "args": {
       "data_root": "/tmp/nvflare/datasets/cifar10",
-      "subset_size": 1000,
-      "min_train_time": 0.2,
-      "max_train_time": 1.0
+      "subset_size": 100,
+      "communication_delay": {"mean": 5.0, "std": 1.0},
+      "device_speed": {"mean": [10.0, 20.0, 40.0], "std": [1.0, 2.0, 4.0]}
```
YuanTingHsieh, line 14: why 3 mean and 3 std?

nvflare/edge/assessors/model_update.py
```diff
@@ -54,44 +37,65 @@ def __init__(
...
         min_hole_to_fill=1,
+        global_lr=1.0,
+        staleness_weight=False,
         device_reuse=True,
         const_selection=False,
     ):
-        """
-        Enable both asynchronous and synchronous model updates from clients.
-        For asynchronous updates, the staleness is calculated based on the current model version.
-        And the aggregation weights are calculated following FedBuff paper "Federated Learning with Buffered Asynchronous Aggregation".
```
YuanTingHsieh, line 63: are we still following FedBuff paper?

nvflare/edge/assessors/model_manager.py
```diff
@@ -0,0 +1,188 @@
...
+        self.devices = {}
+        self.last_update_time = None
+
+    def accept(self, model_update: ModelUpdate, fl_ctx: FLContext):
+        self.last_update_time = time.time()
+        self.devices.update(model_update.devices)
+        return self.aggregator.accept(model_update.update, fl_ctx)
+
+
+class ModelManager(FLComponent):
```
YuanTingHsieh: suggest add some docstring this is following the FedBuff paper?

nvflare/edge/assessors/model_update.py
```diff
@@ -54,44 +37,65 @@ def __init__(
...
+        """Initialize the ModelUpdateAssessor.
+
+        Args:
+            persistor_id (str): ID of the persistor component used to load and save models.
+            num_updates_for_model (int): Number of updates required before generating a new model version.
+            max_model_version (int): Maximum number of model versions allowed in the workflow.
+            max_model_history (int): Maximum number of historical model versions to keep in memory.
+            device_selection_size (int): Number of devices to select for each model update round.
```
YuanTingHsieh: Are some of these args only used in Model management?

Should we instead just pass like "device_manager_id", "model_manager_id" or something to this ModelUpdate Assessor?

YuanTingHsieh: hmm. but then those DeviceManager and ModelManager will be exposed to the users..

nvflare/edge/assessors/model_update.py
```diff
@@ -117,224 +122,76 @@ def start_task(self, fl_ctx: FLContext) -> Shareable:
...
+
+        # Update available devices
         if report.available_devices:
-            self.available_devices.update(report.available_devices)
+            self.device_manager.update_available_devices(report.available_devices)
             self.log_debug(
                 fl_ctx,
                 f"assessor got reported {len(report.available_devices)} available devices from child. "
-                f"total num available devices: {len(self.available_devices)}",
+                f"total num available devices: {len(self.device_manager.available_devices)}",
```
YuanTingHsieh: these 4 lines can all be inside the "update_available_devices" in DeviceManager

if you need to pass in fl_ctx then you can pass it along.

nvflare/edge/assessors/model_update.py
```diff
@@ -117,224 +122,76 @@ def start_task(self, fl_ctx: FLContext) -> Shareable:
...
+            # Handle device selection
+            num_holes = self.device_manager.device_selection_size - len(self.device_manager.current_selection)
+            if num_holes >= self.device_manager.min_hole_to_fill:
+                if not self.device_manager.const_selection:
+                    self.device_manager.fill_selection(self.model_manager.current_model_version, fl_ctx)
                 else:
-                    # print constant selection message
                     self.log_info(fl_ctx, "constant selection enabled, use the same original set.")
-                    self._repeat_selection(fl_ctx)
+                    self.device_manager.repeat_selection(self.model_manager.current_model_version, fl_ctx)
```
YuanTingHsieh: suggest move these lines into DeviceManager's method,
something like "refill_selection_if_needed"

nvflare/edge/assessors/model_update.py
```diff
@@ -117,224 +122,76 @@ def start_task(self, fl_ctx: FLContext) -> Shareable:
...
-            if len(self.available_devices) >= self.device_selection_size:
-                self.log_info(fl_ctx, f"got {len(self.available_devices)} devices - generate initial model")
-                self._generate_new_model(fl_ctx)
-                self._fill_selection(fl_ctx)
+        # Handle initial model generation
+        if self.model_manager.current_model_version == 0:
+            if len(self.device_manager.available_devices) >= self.device_manager.device_selection_size:
+                self.log_info(
+                    fl_ctx, f"got {len(self.device_manager.available_devices)} devices - generate initial model"
+                )
```
YuanTingHsieh: suggest move these two "if conditions" into a method call:
"self.need_to_generate_initial_model()" or something.

suggest move the ```"if len(self.device_manager.available_devices) >= self.device_manager.device_selection_size:"```
to a method called something like ```"self.device_manager.has_sufficient_devices()"```

nvflare/edge/assessors/model_update.py
```diff
@@ -117,224 +122,76 @@ def start_task(self, fl_ctx: FLContext) -> Shareable:
...
+                self.model_manager.generate_new_model(fl_ctx)
+                self.device_manager.fill_selection(self.model_manager.current_model_version, fl_ctx)
 
+        # Prepare reply
         model = None
-        if self.current_model_version != report.current_model_version:
-            model = self.current_model
+        if self.model_manager.current_model_version != report.current_model_version:
+            model = self.model_manager.get_current_model()
```
YuanTingHsieh: model could be None IF current_model_version is the same as report.current_model_version? the model will be cached somewhere?

And we could rewrite the ModelManager's get_current_model to handle this logic?

calling something like:
```
model = self.model_manager.get_current_model(report.current_model_version)```

File: nvflare/edge/assessors/device_manager.py [hunk 9]
```diff
@@ -0,0 +1,88 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import random
+from typing import Dict, Set
+
+from nvflare.apis.fl_component import FLComponent
+
+
+class DeviceManager(FLComponent):
```
YuanTingHsieh: can you add some docstrings to explain this DeviceManager's selection logic?

examples/advanced/edge/README.md
```diff
@@ -154,3 +155,10 @@ With the centralized training of 10 epochs, and the federated training of 10 rou
 
 Red curve is the centralized training, blue is the baseline federated training with regular single-layer setting, and green is the simulated cross-device federated training.
 The three learning will converge to similar accuracy, note that in this case each client holds partial data that is 1/16 of the whole training set sequentially split.
+
+Comparing sync v.s. async, we tested an async scheme with producing a new model upon receiving 4 model updates, so comparing with sync scheme requiring 16 model updates for 
+generating a new global model, we let the async scheme run for 40 model versions. The global accuracy curves are shown below:
+<img src="./figs/async.png" alt="Cifar10 Async Results" width="800" >
+
+Blue curve is sync training, and black curve is async training. As shown, the final model reached similar global accuracy, while the time cost of async scheme
+is around 2/3 of that of sync scheme under our simulated settings, even considering it performs 4x more global evaluations.
```
YuanTingHsieh: suggest make the text more explicit, something like:


"
Comparing synchronous (sync) vs. asynchronous (async) training, we tested an async scheme that produces a new global model after receiving 4 model updates, compared to the sync scheme which requires 16 model updates to generate a new global model.


In this comparison, the sync scheme generates 10 model versions (i.e., 10 global evaluations). To match the total number of model updates processed, we let the async scheme run for 40 model versions, because 
16×10/4=40

The global accuracy curves are shown below:


The blue curve represents sync training, and the black curve represents async training. Both approaches reach similar final global accuracy. However, the async scheme completes in approximately 2/3 of the time of the sync scheme under our simulated settings, despite performing 4 times more global evaluations.


"

examples/advanced/edge/jobs/configs/cifar10_device_config.json
ZiyueXu77: simulate hardware fast-med-slow three types of devices with three Gaussian - may not be able to directly sampled from a single distribution. Also, once selected, we need it fixed - one device may not "suddenly become fast" in a different round of training.

nvflare/edge/assessors/model_update.py
ZiyueXu77: yes, somehow this part did not get correctly recorded, thanks for noticing!

nvflare/edge/assessors/model_manager.py
ZiyueXu77: good point, added

nvflare/edge/assessors/model_update.py
ZiyueXu77: Yeah, that can be a better option, users can provide customized model manager and device manager for their own schemes

nvflare/edge/assessors/model_update.py
ZiyueXu77: moved

examples/advanced/edge/README.md
ZiyueXu77: updated

nvflare/edge/assessors/model_update.py
ZiyueXu77: everything is "update-based", so if none, then the model will not be updated across our hierarchical structure

nvflare/edge/assessors/device_manager.py
ZiyueXu77: added

nvflare/edge/assessors/model_update.py
ZiyueXu77: I think this is fine, it involves both manager's info, and current two conditions are clear enough

nvflare/edge/assessors/model_update.py
ZiyueXu77: good point! updated

nvflare/edge/assessors/device_manager.py
```diff
@@ -0,0 +1,93 @@
...
```
yanchengnv: DeviceManager should be an abstract class (defining only spec) to allow different kinds of implementations.

nvflare/edge/assessors/model_manager.py
```diff
@@ -0,0 +1,199 @@
...
```
yanchengnv: Similarly, ModelManager should be an abstract class that only defines the spec. This allows different implementations.

What you implemented here should be just one impl.

nvflare/edge/assessors/model_update.py
```diff
@@ -11,41 +11,24 @@
...
-    def accept(self, model_update: ModelUpdate, fl_ctx: FLContext):
-        self.last_update_time = time.time()
-        self.devices.update(model_update.devices)
-        return self.aggregator.accept(model_update.update, fl_ctx)
+from nvflare.edge.assessors.device_manager import DeviceManager
+from nvflare.edge.assessors.model_manager import ModelManager
+from nvflare.edge.mud import BaseState, StateUpdateReply, StateUpdateReport
 
 
 class ModelUpdateAssessor(Assessor):
```
yanchengnv, line 29: It should have args for "model_manager_component_id" and "device_manager_component_id" to allow different implementations to be used.

nvflare/edge/assessors/model_update.py
```diff
@@ -54,44 +37,68 @@ def __init__(
...
         self.min_hole_to_fill = min_hole_to_fill
         self.device_reuse = device_reuse
         self.const_selection = const_selection
+        self.global_lr = global_lr
+        self.staleness_weight = staleness_weight
+
         self.update_lock = threading.Lock()
         self.start_time = None
         self.register_event_handler(EventType.START_RUN, self._handle_start_run)
+        self.model_manager = ModelManager(
```
yanchengnv: Do not hard-code to a particular impl of ModelManager. Instead, you should get the model manager from the specified component id.

nvflare/edge/assessors/model_update.py
```diff
@@ -54,44 +37,68 @@ def __init__(
...
+            num_updates_for_model=num_updates_for_model,
+            max_model_version=max_model_version,
+            max_model_history=max_model_history,
+            global_lr=global_lr,
+            staleness_weight=staleness_weight,
+        )
+        self.device_manager = DeviceManager(
```
yanchengnv: Do not hard-code to a particular impl of DeviceManager. Instead, you should get the device manager from the specified component id.

nvflare/edge/assessors/model_update.py
```diff
@@ -54,44 +37,68 @@ def __init__(
...
-        For asynchronous updates, the staleness is calculated based on the current model version.
-        And the aggregation weights are calculated following FedBuff paper "Federated Learning with Buffered Asynchronous Aggregation".
+        For asynchronous updates, the staleness is calculated based on the starting and current model version.
+        And the aggregation scheme and weights are calculated following FedBuff paper "Federated Learning with Buffered Asynchronous Aggregation".
+
+        Args:
+            persistor_id (str): ID of the persistor component used to load and save models.
+            num_updates_for_model (int): Number of updates required before generating a new model version.
+            max_model_version (int): Maximum number of model versions allowed in the workflow.
+            max_model_history (int): Maximum number of historical model versions to keep in memory.
```
yanchengnv: These 3 args (num_updates_for_model, max_model_version, max_model_history) are related to model management, hence should belong to ModeManager, not here.

yanchengnv (commented):
Please revisit the abstraction of ModelManager and DeviceManager. What are their responsibilities? How are they supposed to be used by the Assessor?

nvflare/edge/assessors/model_update.py
ZiyueXu77: will update according to @yanchengnv 's comments

nvflare/edge/assessors/device_manager.py
ZiyueXu77: abstracted

nvflare/edge/assessors/model_update.py
ZiyueXu77: done

nvflare/edge/assessors/model_update.py
ZiyueXu77: updated to component rather than direct import

nvflare/edge/assessors/model_manager.py
ZiyueXu77: abstracted

nvflare/edge/assessors/model_update.py
ZiyueXu77: done

nvflare/edge/assessors/model_update.py
ZiyueXu77: updated

yanchengnv (approved):
Looks very good!

PR #3519, opened 2025-06-10
Consolidate admin port

yanchengnv:
Fixes # .

### Description

This PR consolidates admin port number with the FL port number.

- By default, the admin port is the same as the FL port. 
- User can still specify a separate admin_port in provision.
- Admin Client is changed to use cellnet to communicate with the admin server.
- Once logged in, the admin client does re-login automatically in case the admin server is restarted.
- Enforce file integrity check on fed_admin.json. If modified, admin client won't start. User can change other params in local/resources.json.default in his startup kit.
- Changed to use file streaming for submitting and downloading jobs.
- Removed the use of overseer agent for admin client.
- Changed FL Server to support multiple ports on the same cell.
- Updated flare_api and FLAdminAPI.
- Removed socket server testing, since we no longer use socket server for admin.
- Make default scheme AIO based for server and relay.

### Types of changes

- [ ] Documentation updated.


nvflare/fuel/f3/qat/two_head.py
```diff
@@ -0,0 +1,87 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+import threading
+
+from nvflare.fuel.f3.cellnet.core_cell import CoreCell
+from nvflare.fuel.f3.cellnet.net_agent import NetAgent
+from nvflare.fuel.f3.mpm import MainProcessMonitor
+
+
+class TwoHeader:
```
YuanTingHsieh, line 21: what is this class for?

nvflare/fuel/flare_api/api_spec.py
```diff
@@ -540,7 +510,7 @@ def get_connected_client_list(self) -> List[ClientInfo]:
     @abstractmethod
     def monitor_job(
         self, job_id: str, timeout: int = 0, poll_interval: float = 2.0, cb=None, *cb_args, **cb_kwargs
-    ) -> MonitorReturnCode:
+    ) -> (MonitorReturnCode, Optional[dict]):
```
YuanTingHsieh, line 513: for 3.10+, we could change the type hinting to:

`tuple[MonitorReturnCode, dict | None]`

but that affects multiple places we could change later.

nvflare/fuel/hci/client/config.py
```diff
@@ -72,10 +74,6 @@ def process_config_element(self, config_ctx: ConfigContext, node: Node):
         element = node.element
         path = node.path()
 
-        if path == "admin.overseer_agent":
-            self.overseer_agent = self.build_component(element)
-            return
-
```
YuanTingHsieh, line 78: should we remove all other places of overseer agent codes/configs?

nvflare/lighter/templates/master_template.yml
```diff
@@ -111,15 +111,26 @@ fed_client: |
...
+      "ssl_root_cert": "rootCA.pem",
+      "fqsn": "{~~fqsn~~}",
+      "is_leaf": {~~is_leaf~~},
+      "connection_security": "{~~conn_sec~~}"
+    },
+    "overseer_agent": {
+      "path": "nvflare.ha.dummy_overseer_agent.DummyOverseerAgent",
+      "args": {
+        "sp_end_point": "{~~sp_end_point~~}"
+      }
```
YuanTingHsieh, line 133: Do we still need this section of overseer agent?
Should we remove it from 2.7?

nvflare/fuel/hci/server/hci.py
```diff
@@ -13,14 +13,18 @@
...
+from nvflare.app_common.streamers.file_streamer import FileStreamer
+from nvflare.fuel.f3.cellnet.cell import Cell
+from nvflare.fuel.f3.cellnet.defs import CellChannel
+from nvflare.fuel.f3.message import Message as CellMessage
 from nvflare.fuel.hci.conn import Connection
-from nvflare.fuel.hci.proto import MetaKey, MetaStatusValue, ProtoKey, make_meta, validate_proto
-from nvflare.fuel.hci.security import IdentityKey, get_identity_info
+from nvflare.fuel.hci.proto import MetaKey, MetaStatusValue, ProtoKey, StreamChannel, make_meta, validate_proto
+from nvflare.fuel.utils.log_utils import get_obj_logger
+from nvflare.private.fed.server.cred_keeper import CredKeeper
```
YuanTingHsieh, line 27: later we could move these outside of private to fuel

nvflare/private/fed/server/fed_server.py
```diff
@@ -179,16 +180,16 @@ def deploy(self, args, grpc_args=None, secure_train=False):
...
-        if len(parts) > 1:
-            # "0" means all interfaces for all protocols (ipv4 and ipv6)
-            listen_target = "0:" + parts[1]
-        else:
-            listen_target = target
+        if len(parts) != 2:
+            raise RuntimeError(f"bad service target: {target}")
+
+        ports = parts[1].split(",")
+        root_url = [f"{scheme}://0:{p}" for p in ports]
```
YuanTingHsieh, line 187: maybe this could be a util function so it's easier to maintain

nvflare/private/fed/server/job_cmds.py
```diff
@@ -674,68 +646,65 @@ def download_job(self, conn: Connection, args: List[str]):
...
-                conn.append_error("internal error", meta=make_meta(MetaStatusValue.INTERNAL_ERROR))
+        self._download_job_comps(conn, args, self._get_extra_job_components)
+
+    def _get_extra_job_components(self, job_def_manager, job_id, fl_ctx):
+        all_components = job_def_manager.list_components(jid=job_id, fl_ctx=fl_ctx)
+        if all_components:
+            return [
+                (item, item)
+                for item in all_components
+                if item not in {WORKSPACE, META, "scheduled", DATA} and StorageSpec.is_valid_component(item)
```
YuanTingHsieh, line 704: in future we might need to enhance this "scheduled" to some Constant of job scheduling status.

nvflare/private/fed/server/job_cmds.py
```diff
@@ -674,68 +646,65 @@ def download_job(self, conn: Connection, args: List[str]):
...
-                job_def_manager.get_storage_for_download(job_id, job_download_dir, DATA, JOB_ZIP, fl_ctx)
-                job_def_manager.get_storage_for_download(job_id, job_download_dir, META, META_JSON, fl_ctx)
-                job_def_manager.get_storage_for_download(job_id, job_download_dir, WORKSPACE, WORKSPACE_ZIP, fl_ctx)
+                for ct, file_name in comps:
+                    job_def_manager.get_storage_for_download(job_id, job_download_dir, ct, file_name, fl_ctx)
 
                 self.download_folder(
                     conn,
                     tx_id=tx_id,
                     folder_name=job_id,
```
YuanTingHsieh, line 670: good change for reducing duplicate codes.
a minor difference for "download_job_components" previously it was downloaded to `job_id + "_components"`
functionality-wise, it is good.

nvflare/private/stream_runner.py
```diff
@@ -153,6 +161,7 @@ def register_stream_processing(
             channel: app channel
             topic: app topic
             factory: the factory to be registered
+            consumed_cb: the CB is called after a chunk is consumed
             stream_done_cb: the CB to be called when a stream is done
```
YuanTingHsieh, line 165: maybe we should add the signature of cb in the docstring

nvflare/fuel/f3/qat/two_head.py
yanchengnv: This is for testing the Cellnet's capability of having multiple cells within the same process.
This was very old code but wasn't checked in.

nvflare/fuel/flare_api/api_spec.py
yanchengnv: Sure, we'll change later if needed.

nvkevlu (dismissed):
The changes look pretty good to me, but I think we need to make sure to document all of the changes well and point them out in the what's new so it is easy for users to know the differences if they need to change anything.

nvflare/app_common/streamers/file_downloader.py
```diff
@@ -0,0 +1,230 @@
...
+            chunk_size = DEFAULT_CHUNK_SIZE
+
+        check_positive_int("chunk_size", chunk_size)
+        self.chunk_size = chunk_size
+
+    def produce(self, ref_id: str, obj, state: dict, requester: str) -> (str, Any, dict):
+        assert isinstance(obj, _File)
+        received_bytes = 0
+        if state:
+            received_bytes = state.get(_StateKey.RECEIVED_BYTES, 0)
```
nvidianz, line 62: Should this key be called SENT_BYTES?

nvflare/app_common/streamers/file_downloader.py
```diff
@@ -0,0 +1,230 @@
...
+            self.logger.error(f"bad {_StateKey.RECEIVED_BYTES} {received_bytes} from {requester}")
+            return ProduceRC.ERROR, None, None
+
+        if received_bytes >= obj.size:
+            # already done
+            return ProduceRC.EOF, None, None
+
+        num_bytes_to_send = min(self.chunk_size, obj.size - received_bytes)
+        with open(obj.name, "rb") as f:
+            f.seek(received_bytes)
```
nvidianz, line 74: The seek() here is not necessary. It should always point to the current position.

nvflare/app_common/streamers/file_streamer.py
```diff
@@ -178,7 +183,13 @@ def register_stream_processing(
         if not os.path.isdir(dest_dir):
             raise ValueError(f"dest_dir '{dest_dir}' is not a valid dir")
 
-        engine = fl_ctx.get_engine()
+        if isinstance(fl_ctx, StreamableEngine):
```
nvidianz: Should we allow this, passing engine as fl_ctx? I think it's better to throw an error here.

nvflare/dashboard/application/blob.py
```diff
@@ -91,7 +91,7 @@ def _gen_kit(download_key, prepare_target_cb=None, **cb_kwargs):
     u = Store.get_user(1)
     super_user = u.get("user")
     fl_port = 8002
-    admin_port = 8003
+    admin_port = fl_port
```
nvidianz, line 94: This parameter should be removed.

nvflare/fuel/flare_api/flare_api.py
```diff
@@ -62,94 +60,61 @@
...
         self.upload_dir = upload_dir
         self.download_dir = download_dir
-        self.overseer_agent = conf.overseer_agent
 
     def close(self):
         """Close the session."""
-        self.api.close()
+        self.api.logout()
 
     def try_connect(self, timeout):
```
nvidianz, line 112: This method can be removed.

nvflare/fuel/hci/base64_utils.py
```diff
@@ -66,10 +66,19 @@ def text_file_to_b64str(file_name) -> str:
 
     """
     data_string = open(file_name, "r").read()
-    data_bytes = data_string.encode("utf-8")
+    return str_to_b64str(data_string)
+
+
+def str_to_b64str(s: str) -> str:
```
nvidianz, line 72: Do we still need Base64?

nvflare/fuel/hci/client/api.py
```diff
@@ -360,46 +271,47 @@ def __init__(
...
+        self.user_identity = admin_config.get(AdminConfigKey.IDENTITY)
+        self.conn_sec = admin_config.get(AdminConfigKey.CONNECTION_SECURITY)
+        self.project_name = admin_config.get(AdminConfigKey.PROJECT_NAME)
+        self.server_identity = admin_config.get(AdminConfigKey.SERVER_IDENTITY, "server")
+        self.scheme = admin_config.get(AdminConfigKey.CONNECTION_SCHEME, "grpc")
+        self.ca_cert = admin_config.get(AdminConfigKey.CA_CERT)
+        self.client_cert = admin_config.get(AdminConfigKey.CLIENT_CERT)
+        self.client_key = admin_config.get(AdminConfigKey.CLIENT_KEY)
+        self.uid_source = admin_config.get(AdminConfigKey.UID_SOURCE, UidSource.USER_INPUT)
+        self.host = admin_config.get(AdminConfigKey.HOST, "localhost")
```
nvidianz, line 287: Should we remove admin host and port?

nvflare/lighter/prov_utils.py
```diff
@@ -71,4 +71,4 @@ def prepare_packager(project_dict: dict):
     if not packager_config:
         return None
     else:
-        return instantiate_from_config(packager_config[0])
+        return instantiate_from_config(packager_config)
```
YuanTingHsieh, line 74: packager_config is a list, and we only have one element of packager, so we use [0] here.

nvflare/lighter/impl/static_file.py
```diff
@@ -122,32 +87,52 @@ def _build_conn_properties(site: Participant, ctx: ProvisionContext, site_config
...
+            conn_security = ConnSecurity.MTLS
 
         custom_ca_cert = site.get_prop_fb(PropKey.CUSTOM_CA_CERT)
         if custom_ca_cert:
             shutil.copyfile(custom_ca_cert, os.path.join(ctx.get_kit_dir(site), ProvFileName.CUSTOM_CA_CERT_FILE_NAME))
+        return conn_security
+
+    def _determine_scheme(self, participant: Participant, scheme=None) -> str:
+        # use AIO for server by default
+        use_aio = participant.get_prop(PropKey.USE_AIO, participant.type == ParticipantType.SERVER)
```
YuanTingHsieh, line 100: need to add this to documentation to explain how to set scheme

nvflare/fuel/hci/client/api_spec.py
```diff
@@ -223,21 +209,36 @@ def check_command(self, command: str) -> CommandInfo:
...
+class AdminConfigKey:
+    PROJECT_NAME = "project_name"
+    IDENTITY = "identity"
+    SERVER_IDENTITY = "server_identity"
+    CONNECTION_SCHEME = "scheme"
+    CONNECTION_SECURITY = "connection_security"
+    CLIENT_KEY = "client_key"
+    CLIENT_CERT = "client_cert"
+    CA_CERT = "ca_cert"
+    HOST = "host"
```
nvidianz, line 221: We should remove host/port from admin config.

nvflare/fuel/hci/client/fl_admin_api_runner.py
```diff
@@ -52,23 +53,19 @@ def __init__(
...
+
         if debug:
             debug = True
 
         try:
             os.chdir(admin_dir)
             workspace = Workspace(root_dir=admin_dir)
-            conf = FLAdminClientStarterConfigurator(workspace)
+            conf = secure_load_admin_config(workspace)
             conf.configure()
```
YuanTingHsieh: Since we already configure() inside the "secure_load_admin_config", we don't need to configure here again?

nvflare/fuel/hci/server/login.py
```diff
@@ -120,100 +66,107 @@ def get_spec(self):
...
+        identity_verifier = hci.get_id_verifier()
+        id_asserter = hci.get_id_asserter()
+
+        cert = load_crt_bytes(cert_data)
+        try:
+            ok = identity_verifier.verify_common_name(
+                asserter_cert=cert,
+                asserted_cn=user_name,
+                signature=signature,
+                nonce="",
```
YuanTingHsieh, line 90: we might need to enhance this empty nonce in later PRs.

nvflare/lighter/ctx.py
```diff
@@ -40,9 +40,9 @@ def __init__(self, workspace_root_dir: str, project: Project):
         self[CtxKey.PROJECT] = project
 
         server = project.get_server()
-        admin_port = server.get_prop(PropKey.ADMIN_PORT, 8003)
-        self[CtxKey.ADMIN_PORT] = admin_port
         fed_learn_port = server.get_prop(PropKey.FED_LEARN_PORT, 8002)
+        admin_port = server.get_prop(PropKey.ADMIN_PORT, fed_learn_port)
```
nvidianz, line 44: Should remove admin_port

nvflare/private/fed/server/cred_keeper.py
```diff
@@ -0,0 +1,69 @@
...
+            if not self.id_asserter:
+                config = self._get_server_config(fl_ctx)
+                cert_file = config.get(SecureTrainConst.SSL_CERT)
+
+                if not cert_file:
+                    self.logger.error(f"missing {SecureTrainConst.SSL_CERT} in server config")
+                    return None
+
+                private_key_file = config.get(SecureTrainConst.PRIVATE_KEY)
+                self.id_asserter = IdentityAsserter(private_key_file=private_key_file, cert_file=cert_file)
```
YuanTingHsieh, line 68: should we rename this to IdentitySigner to reflect its functions?

nvflare/lighter/impl/static_file.py
```diff
@@ -122,32 +87,52 @@ def _build_conn_properties(site: Participant, ctx: ProvisionContext, site_config
...
-            site_config["connection_security"] = conn_security
+        else:
+            conn_security = ConnSecurity.MTLS
 
         custom_ca_cert = site.get_prop_fb(PropKey.CUSTOM_CA_CERT)
         if custom_ca_cert:
             shutil.copyfile(custom_ca_cert, os.path.join(ctx.get_kit_dir(site), ProvFileName.CUSTOM_CA_CERT_FILE_NAME))
+        return conn_security
+
+    def _determine_scheme(self, participant: Participant, scheme=None) -> str:
```
nvidianz, line 98: The logic here is the same as clients. Maybe we should consolidate the code.

nvflare/lighter/impl/static_file.py
```diff
@@ -482,80 +478,39 @@ def _validate_host_name_against_listener(
...
+            # Note: by doing this dynamically, we guarantee the sp_end_point to be correct, even if the
+            # project.yaml does not specify the default server host correctly!
+            conn_host = server.get_default_host()
+        return conn_host, port
 
     def _build_admin(self, admin: Participant, ctx: ProvisionContext):
+        self.prepare_admin_config(admin, ctx)
+
         dest_dir = ctx.get_kit_dir(admin)
         admin_port = ctx.get(CtxKey.ADMIN_PORT)
```
nvidianz, line 515: Should rename this to port or server_port, admin_port may cause confusion.

nvflare/fuel/hci/server/hci.py
```diff
@@ -29,176 +33,148 @@
...
+
+                data = req_json[ProtoKey.DATA]
+                for item in data:
+                    it = item[ProtoKey.TYPE]
+                    if it == ProtoKey.COMMAND:
+                        command = item[ProtoKey.DATA]
+                        break
+
+                if command is None:
+                    self.logger.error(f"no processor registered for command '{command}'")
```
YuanTingHsieh: where do we check the processor is registered?
we only check if req_json[ProtoKey.DATA] has an item that has [ProtoKey.DATA]?
Note that we reuse this ProtoKey.DATA at different level.

nvflare/fuel/hci/server/login.py
```diff
@@ -120,100 +66,107 @@ def get_spec(self):
...
+        self.logger.debug(f"got cert dict: {cert_dict}")
+        identity = get_identity_info(cert_dict)
+
+        request = conn.get_prop(ConnProps.REQUEST)
+        assert isinstance(request, CellMessage)
+        origin = request.get_header(MessageHeaderKey.ORIGIN)
+
         session = self.session_mgr.create_session(
-            user_name=identity[IdentityKey.NAME],
+            user_name=user_name,
```
YuanTingHsieh, line 111: what is the different user_name vs identity[IdentityKey.NAME] or at this point they are already matched

nvflare/fuel/hci/server/sess.py
```diff
@@ -11,37 +11,83 @@
...
+        parts = token.split(":")
+        if len(parts) != 2:
+            raise ValueError(f"invalid token {token}: expects 2 parts but got {len(parts)}")
+
+        bds = parts[0]
+        signature = parts[1]
+        ds = b64str_to_str(bds)
+        if id_asserter:
+            token_verifier = TokenVerifier(id_asserter.cert)
+            is_valid = token_verifier.verify("", ds, signature)
```
YuanTingHsieh, line 75: should we instead call the verify_content directly here?

nvflare/fuel/hci/server/sess.py
```diff
@@ -86,34 +133,47 @@ def monitor_sessions(self):
...
+            user_name=user_name,
+            org=user_org,
+            role=user_role,
+            origin_fqcn=origin_fqcn,
+        )
+        with self.sess_update_lock:
+            self.sessions[sess_id] = sess
+        return sess
+
+    def recreate_session(self, token: str, origin_fqcn, id_asserter: IdentityAsserter):
```
YuanTingHsieh, line 162: instead of passing IdentityAsserter, should we pass the "cert" directly?

nvflare/fuel/hci/client/fl_admin_api_runner.py
YuanTingHsieh: we have already called configure() inside the secure_load_admin_config

nvflare/fuel/hci/conn.py
```diff
@@ -16,178 +16,74 @@
...
-    def flush_bytes(self, data):
-        self.sock.sendall(data)
+        if props:
+            self.set_props(props)
 
     def append_table(self, headers: List[str], name=None) -> Table:
         return self.buffer.append_table(headers, name=name)
 
-    def append_string(self, data: str, flush=False, meta: dict = None):
+    def append_string(self, data: str, meta: dict = None):
```
YuanTingHsieh, line 39: change is good in this file, do we need to change: https://github.com/NVIDIA/NVFlare/blob/main/nvflare/fuel/f3/cellnet/net_manager.py#L301 and https://github.com/NVIDIA/NVFlare/blob/af09c2c5e6e2c473f6757761c01e49129b88626e/nvflare/fuel/f3/cellnet/net_manager.py#L333

nvflare/fuel/flare_api/flare_api.py
```diff
@@ -62,94 +60,61 @@
...
-        self.username = username
         assert isinstance(startup_path, str), "startup_path must be str"
-        self.secure_mode = secure_mode
-
         assert os.path.isdir(startup_path), f"startup kit does not exist at {startup_path}"
 
         workspace = Workspace(root_dir=startup_path)
-        conf = FLAdminClientStarterConfigurator(workspace)
+        conf = secure_load_admin_config(workspace)
         conf.configure()
```
YuanTingHsieh: Do we need to call configure() again?

nvflare/fuel/hci/tools/admin.py
```diff
@@ -42,10 +42,10 @@ def main():
     try:
         os.chdir(args.workspace)
         workspace = Workspace(root_dir=args.workspace)
-        conf = FLAdminClientStarterConfigurator(workspace=workspace)
+        conf = secure_load_admin_config(workspace)
         conf.configure()
```
YuanTingHsieh: here as well.

nvflare/fuel/hci/tools/admin.py
```diff
@@ -56,56 +56,25 @@ def main():
```
YuanTingHsieh, line 55: maybe this "admin_config" reading can also be wrapped into a util method

nvflare/lighter/prov_utils.py
YuanTingHsieh, line 74: this needs to added `[0]` otherwise this "packager_config" is a list.

nvflare/app_common/streamers/file_downloader.py
```diff
@@ -0,0 +1,230 @@
...
+        if app_downloaded_cb:
+            app_downloaded_cb(ref_id, to_site, status, obj.name, **cb_kwargs)
+
+
+class _ChunkConsumer(Consumer):
+
+    def __init__(self, location: str):
+        Consumer.__init__(self)
+        self.location = location
+        self.file_path = os.path.join(location, str(uuid.uuid4()))
```
YuanTingHsieh, line 164: if we want we could add some hash check (MD5) for file integrity in the future.

nvflare/app_common/streamers/file_downloader.py
```diff
@@ -0,0 +1,230 @@
...
+
+    """
+    if location is not None:
+        if not os.path.exists(location):
+            raise ValueError(f"location '{location}' does not exist")
+
+        if not os.path.isdir(location):
+            raise ValueError(f"location '{location}' is not a valid dir")
+    else:
+        location = tempfile.gettempdir()
```
YuanTingHsieh, line 216: if tempdir here, who is responsible for removing it?

nvflare/app_common/streamers/obj_downloader.py
```diff
@@ -0,0 +1,500 @@
...
+where tx_id is the ID of the transaction;
+objs is the list of large objects registered with the transaction;
+cb_kwargs are the kw args registered with the CB.
+
+You may need to use both mechanisms to fully take care of object life cycles. The object downloaded CB may never be 
+called since the site somehow didn't finish the downloading. In reality the timeout mechanism may be sufficient.
+
+Unlike with Object Streamer that the object owner pushes small objects to the recipients; with Object Downloader, 
+each recipient pulls the data from the object owner.
+"""
```
YuanTingHsieh, line 93: this is interesting, so previously we are pushing something to each client (after each of them "pull_task")? now should we change it to pull for other things as well? not sure what other places though

nvflare/app_common/streamers/file_downloader.py
YuanTingHsieh: We are downloading to a temp file path (uuid), so the caller are responsible for removing this file

Then does it matter if we should pass in "location" here?

nvflare/fuel/hci/client/api.py
```diff
@@ -421,22 +333,151 @@ def __init__(
...
+            per_request_timeout=self.file_download_progress_timeout,
+        )
+        if err:
+            print(f"failed to receive file {file_name}: {err}")
+            return None
+
+        file_stats = os.stat(file_path)
+        num_bytes_received = file_stats.st_size
+        Path(os.path.dirname(file_name)).mkdir(parents=True, exist_ok=True)
+        shutil.move(file_path, file_name)
```
YuanTingHsieh, line 470: some logic of renaming/removing tempfile/folder is handled here, should we move them into the downloader.download_file?

YuanTingHsieh (commented):
Good changes throughout, added some comments.

nvflare/app_common/streamers/file_downloader.py
yanchengnv: Depending on whose perspective. From downloading site, it's RECEIVED_BYTES.
Downloading perspective seems to be more important since downloading site is in control.

nvflare/app_common/streamers/file_downloader.py
yanchengnv: No. Each site is downloading on its own pace. There is no "current position" on the sending site.

nvflare/app_common/streamers/file_streamer.py
yanchengnv: Good catch! This was an early trick for admin server. Now fixed.

nvflare/dashboard/application/blob.py
yanchengnv: No. admin_port can still be separate from fl_port. It's just that by default it is the same as fl_port, but it can be explicitly defined a different value.

nvflare/fuel/hci/base64_utils.py
yanchengnv: Yes, it's still needed. Not for file transfer, but for session token encoding between admin client and server.

nvflare/fuel/hci/client/api.py
yanchengnv: No. Admin Port is still needed. Admin Host could be different than other FL host names.

nvflare/fuel/hci/client/api_spec.py
yanchengnv: No. How does the admin client know the host/port info without them?

nvflare/lighter/ctx.py
yanchengnv: No. As mentioned before, admin port can be different from fl port.

nvflare/lighter/impl/static_file.py
yanchengnv: What do you mean? Consolidate with what code?

nvflare/lighter/impl/static_file.py
yanchengnv: Don't think so.

nvflare/fuel/hci/client/config.py
yanchengnv: Save for later. Already too much change.

nvflare/lighter/templates/master_template.yml
yanchengnv: Yes we do for now. We can remove later. This will require changes to the both server and client logic for getting host and port info. Very messy.

nvflare/fuel/hci/server/hci.py
yanchengnv: yes.

nvflare/private/fed/server/fed_server.py
yanchengnv: Yes, for later.

nvflare/private/fed/server/job_cmds.py
yanchengnv: Yes.

nvflare/private/fed/server/job_cmds.py
yanchengnv: Good.

nvflare/private/stream_runner.py
yanchengnv: They are documented in nvflare.apis.streaming.

nvflare/lighter/prov_utils.py
yanchengnv: No, packager_config should be a single Packager object. Remember we discussed this? We don't support multiple packagers.

nvflare/lighter/impl/static_file.py
yanchengnv: Yes.

nvflare/fuel/hci/client/fl_admin_api_runner.py
yanchengnv: Good catch. Will fix.

nvflare/private/fed/server/cred_keeper.py
yanchengnv: That already existed. Asserter is to "assert my identity".

nvflare/fuel/hci/server/login.py
yanchengnv: Nonce is not used here since this is not a two-way challenge.

nvflare/fuel/hci/server/hci.py
yanchengnv: This log message is misleading. This error is that there is no command specified in the message from the client.
I'll change it.
Yes, the "data" element is used in multiple layers of the protocol.

nvflare/fuel/hci/server/login.py
yanchengnv: The user_name comes from the message directly, and it must match the CN in the cert for authentication.
In previous version, the comm layer populates the identity field based on CN in the connection. But now we use cell communication that may not be SSL based at all. So we need to do explicit cert-based authentication.

nvflare/fuel/hci/server/sess.py
yanchengnv: No, we should use the TokenVerifier since it's intended for token verification, even though it uses verify_content internally.

nvflare/fuel/hci/server/sess.py
yanchengnv: No, the id_asserter is designed to do token verification, and it has its own cert already.

nvflare/fuel/hci/client/fl_admin_api_runner.py
yanchengnv: Good catch! Changed all places.

nvflare/fuel/hci/conn.py
yanchengnv: Good catch! Changed.

nvflare/fuel/hci/tools/admin.py
yanchengnv: Changed.

nvflare/fuel/hci/tools/admin.py
yanchengnv: That's what secure_load_admin_config is for.

nvflare/lighter/prov_utils.py
yanchengnv: No, packager_config must be a single object, not a list, since the provisioner requires it.

nvflare/fuel/flare_api/flare_api.py
yanchengnv: Good catch. Removed.

nvflare/app_common/streamers/file_downloader.py
yanchengnv: It's up to the caller when to delete it or deleting it at all.

The  caller can specify a "location" so that the caller can manager its downloaded files in whatever way it wants. For example, it could just download everything into the same location, and remove the location completely at the end of the job.

nvflare/app_common/streamers/file_downloader.py
yanchengnv: The caller will get the path of the file after the download is done. The caller can delete the file after that.
In case of tempdir, it is managed by the OS (which periodically cleans it). If the caller remembers to remove the file, then it's okay.

nvflare/app_common/streamers/obj_downloader.py
yanchengnv: There are use cases for both.

In case of streaming/push, the data owner sends the data to other sites without being asked at all. Those sites must have already registered to receive the data. The streamer can push to multiple sites at the same time, but the logic is fairly complex since those sites may not respond with the same speed.

In case of downloading/pull, the data owner has to prepare the data to get refs first, and then send refs to one or more sites.  Each site then tries to pull the data on its own pace. The control logic is much simpler since we don't need to worry about those sites come back at different times.

Push is better in that no prep is needed before pushing, especially if just pushing to a single site (like submit job).
Download is better if prep is needed anyway and it's easier to manage.

If you compare obj_downloader with stream_runner, you will see that obj downloader is much simpler or more robust.

A side note: obj downloader only relies on Cell; whereas streamer relies on engine/fl_ctx.

We'll refactor the streamer to only reply on Cell. This will make the streamer available for 3rd-party app without requiring an engine (they have cells already).

nvflare/fuel/hci/client/api.py
yanchengnv: Not really. The lifecycle of the downloaded file should be managed by the caller, not the downloader.

nvflare/app_common/streamers/obj_downloader.py
YuanTingHsieh: got it, thanks for the explanation, that is clear to me.

nvflare/app_common/streamers/file_downloader.py
YuanTingHsieh: But here even if the caller pass in a "location", this FileDownloader is not downloading to there, but it's downloading to "location/[random uuid]"?

Oh got it, the "location" is a always a FOLDER, it's not an absolute path of a "File"

maybe just add a docstring:

`
"""
        Args:
            location (str): Folder path where the downloaded file will be saved.
                            A unique filename will be generated within this folder.
        """
`

YuanTingHsieh (approved):
Great PR! Dropping down to a single port definitely improves usability and simplifies deployment.

nvflare/app_common/streamers/file_downloader.py
yanchengnv: Right. :-)

PR #3528, opened 2025-06-25
Support long server name

yanchengnv:
Fixes # .

### Description

Server Name is used as CN of the cert. But the CN cannot exceed 63 chars. So we have to truncate it to make the cert.

Server Name also serves as the identity of the server for all clients to verify, and it must match the CN in the server's cert.

Server Name is also the default host name (unless default host is explicitly specified) for clients to connect to. Truncated name won't be a valid host name.

We have to accommodate all these factors:

- We truncate the server name and use it for both name and subject of the server. This will satisfy CN requirement of the cert, and will satisfy server identity validation by clients.

- We check whether the DEFAULT_HOST property is explicitly specified in the server. If not, we explicitly set it to the original name.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/lighter/impl/cert.py
+++ b/nvflare/lighter/impl/cert.py
@@ -26,6 +26,8 @@ from nvflare.lighter.entity import Participant, Project
 from nvflare.lighter.spec import Builder
 from nvflare.lighter.utils import Identity, generate_cert, generate_keys, serialize_cert, serialize_pri_key
 
+MAX_CN_LENGTH = 63
+
 
 class _CertState:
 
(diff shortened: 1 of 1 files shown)


nvflare/lighter/impl/cert.py
```diff
@@ -26,6 +26,8 @@
 from nvflare.lighter.spec import Builder
 from nvflare.lighter.utils import Identity, generate_cert, generate_keys, serialize_cert, serialize_pri_key
 
+MAX_CN_LENGTH = 63
```
YuanTingHsieh: based on https://github.com/pyca/cryptography/blob/main/src/cryptography/x509/name.py#L66 the max is 64, but this is minor

YuanTingHsieh (dismissed):

nvflare/lighter/impl/cert.py
yanchengnv: I somehow remembered that some article said that the limit is 63. But then I checked again, and everyone seems to say it's 64.

Anyhow, I changed to 64.

YuanTingHsieh (approved):
LGTM

PR #3530, opened 2025-06-27
Enhance admin functions

yanchengnv:
Fixes # .

### Description

This PR adds following enhancements to the admin functions:
- Use defined constants for admin config properties (instead of hard-coded strings)
- Make it possible for loading security content (the fed_admin.json) for multiple users in the same process. This is necessary when the admin API (flare API or FLAdminAPI) is used for multiple users, and in the integration tests (which requires multiple user sessions).

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh (approved):
LGTM

PR #3535, opened 2025-07-03
Fix admin issues

yanchengnv:
Fixes # .

### Description

This PR fixes a few issues caused by admin port consolidation PR:

- Keep the monitor_job behavior the same as before. Added a new monitor_job_and_return_job_meta method to api to return both monitor RC and job meta data.
- Fixed client check logic for admin commands.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/fuel/utils/admin_name_utils.py
+++ b/nvflare/fuel/utils/admin_name_utils.py
@@ -18,25 +18,55 @@ _NAME_PREFIX_LEN = len(_NAME_PREFIX)
 
 
 def new_admin_client_name():
-    return f"{_NAME_PREFIX}{uuid.uuid4()}"
+    """Create a unique name for a new admin client session.
+    Note that we cannot use the admin user's login name because the same user could have multiple
+    admin sessions at the same time. The admin client name is also the FQCN of the client when connecting
+    to the server.
+
+    The admin client name is created with this pattern:
+        <prefix><uid>
+
+    where the prefix is fixed, whereas uid is a unique ID.
+
+    Returns: a unique name
+
+    """
+    uid = str(uuid.uuid4())
+    return f"{_NAME_PREFIX}{uid}"
 
 
-def _is_valid_uuid(data: str) -> bool:
+def _is_valid_uid(data: str) -> bool:
+    """Check whether the specified data is a valid uid
+
+    Args:
+        data: the data to be checked
+
+    Returns: whether the data is a valid uid.
+
+    """
     try:
         val = uuid.UUID(data, version=4)
     except ValueError:
         return False
 
-    # If the jid string is a valid hex code, but an invalid uuid4,the UUID.__init__ will convert it to a
+    # If data is a valid hex code, but an invalid uuid4,the UUID.__init__ will convert it to a
     # valid uuid4. This is bad for validation purposes.
     return val.hex == data.replace("-", "")
 
 
 def is_valid_admin_client_name(name: str) -> bool:
+    """Check whether the specified name is a valid admin client name.
+
+    Args:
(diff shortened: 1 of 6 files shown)


nvflare/private/fed/server/cmd_utils.py
```diff
@@ -91,7 +92,17 @@ def validate_command_targets(self, conn: Connection, args: List[str]) -> str:
...
+            # check whether client names contain admin clients.
+            admin_clients = []
+            for c in client_names:
+                if is_valid_admin_client_name(c):
+                    admin_clients.append(c)
+
             clients, invalid_inputs = engine.validate_targets(client_names)
+            if admin_clients:
+                # admin clients are considered invalid since admin commands do not go to admin clients!
+                invalid_inputs.extend(admin_clients)
```
YuanTingHsieh, line 104: should we instead modify the "validate_targets" of "engine" to return admin_clients as "invalid_inputs"?

nvflare/fuel/utils/admin_name_utils.py
```diff
@@ -0,0 +1,72 @@
...
+
+    """
+    try:
+        val = uuid.UUID(data, version=4)
+    except ValueError:
+        return False
+
+    # If data is a valid hex code, but an invalid uuid4,the UUID.__init__ will convert it to a
+    # valid uuid4. This is bad for validation purposes.
+    return val.hex == data.replace("-", "")
```
YuanTingHsieh, line 54: should we do:

```
def _is_valid_uid(data: str) -> bool:
    """Check whether the specified data is a valid UUIDv4.

    Args:
        data: the UUID string to validate

    Returns:
        True if valid UUIDv4 format, False otherwise
    """
    try:
        val = uuid.UUID(data)
    except ValueError:
        return False

    return val.version == 4 and str(val) == data.lower()
```

instead?

nvflare/private/fed/server/cmd_utils.py
yanchengnv: Good question but no, we cannot do that because this would block the server sending messages to the admin client (e.g. sending/pushing files to the admin client, or sending session timeout, or any pushing in the future).

nvflare/fuel/utils/admin_name_utils.py
yanchengnv: Not sure your version will work in all cases.  The ChatGPT suggested: return str(val) == data, but not sure it works in all cases.

Remember that the "data" may or may not contain the hyphen sign.

YuanTingHsieh (approved):

nvkevlu:
LGTM, just wondering, what is the reasoning for why we need to have a check for valid admin client name?

yanchengnv:
> LGTM, just wondering, what is the reasoning for why we need to have a check for valid admin client name, is it just defensive programming?

In general we want to check a client is valid before sending a message to it. We used to only have FL clients in cellnet, but now we also have admin clients that are part of cellnet. We can check whether a FL client is valid by simply checking the client table that the server keeps. But server's client table does not include admin clients.  We could change the server code to also include admin clients, but that would cause a lot of additional changes, which we want to avoid.

Now the dilemma is how do we ensure that a given client name (e.g. entered by the user on admin console) is valid? First of all, unlike FL client names, admin client names are not created during provision time. This is because an admin user could have multiple admin client sessions at the same time, but all client names must be unique. Secondly we create admin client names following a pattern that can be validated. Now we can validate any string to see whether it's a valid client name (FL client or admin client).

In case of FL client, we treat it as valid if it's connected (in the server's client table);
In case of Admin client, we check whether it follows the naming pattern.

PR #3544, opened 2025-07-14
Safe tensor decomposer

yanchengnv:
Fixes # .

### Description

This PR address security issue with TensorDecompose. Instead of using jit.load to serialize BF16 type of tensors (which has security issues), this PR uses torch.save to save weights. 

Note that, in general, torch.save uses Pickle for saving whole model. BUT we only save model weights (dict of Tensors) and do not save whole model. This does not involve Pickling.

In addition, this PR makes file-based decomposition a general-purpose function that can be used to implement decomposers of any large objects. This is particularly useful when dealing with LLM.

The general ideas are here:

- Objects of target object types (e.g. Tensor, NP Array) contained in one message are collected into one single dict, and written to a file using the target object type specific file creation like savetensors.save_file, or np.save.
- Objects in the payload are replaced with simple references to the objects written to file. Hence the payload size is hugely reduced.
- The payload is sent to receivers.
- Each receiver downloads file from the sender, using FileDownloader.
- Once downloaded, the payload is recomposed and the object refs are replaced with received objects in the downloaded file.

With this approach, since file is downloaded with small chunks, the memory space needed is small (no need to hold and transfer the huge objects via memory).

This PR adds more enhancement to reduce memory space and increase efficiency when the same message needs to be sent to multiple receivers (e.g. in broadcasting task, or message retries in Reliable Message). Only the first message needs to go thru the file conversion process. All other messages will use cached result from the 1st message (the primary).

The core of this solution is the abstract class ViaFileDecomposer. NumpyArrayDecomposer and TensorDecomposer have been modified to be based on ViaFileDecomposer.

In order to be able to create file refs during decomposition and download files during recomposition, cell must be available to decomposers. To support this, the concept of "FOBS Context" is added, which is used to contain external information (e.g. cell or other config params like file download timeout) needed by decomposers.

One issue with this approach is the life cycle management of the generated temp files - they need to be removed as soon as they are no longer needed. Two ways are implemented to guarantee that those temp files are eventually removed.

- One, at the end of the job, always calls FileDownloader.shutdown(), which closes all pending transactions (which deletes temp files). This is the last resort that catches all files that somehow didn't get removed earlier.
- Two,  use the concept of "msg root" that lets the app layer manage the life cycle of the msg. The msg sender first assigns a unique msg root id to the msg. Then the msg is either sent (to one or multiple receivers), or waiting for receivers to pull. Eventually, when the msg sender knows that the msg is no longer needed, it calls delete_msg_root to remove all temp files associated with the msg root.

In most cases, the msg sender does know when to delete the msg root. The only exception is "fire and forget" since the msg sender does not wait for receivers. In this case, Solution One will be able to delete all the files (if any).

I have updated the following places to use msg root:
- Task-based interactions (wf_comm_server, wf_comm_client, task_contoller)
- Reliable Message, which may resend the message multiple times
- Task Exchanger, which sends message to client API
- Pipe Handler, which may resend the message multiple times

There might be other places to improve.

### Types of changes

- [ ] Documentation updated.


nvflare/apis/impl/task_controller.py
```diff
@@ -51,7 +53,7 @@ def start_controller(self, fl_ctx: FLContext):
         if not self.task_result_filters:
             self.task_result_filters = {}
 
-    def control_flow(self, fl_ctx: FLContext):
+    def control_flow(self, abort_signal: Signal, fl_ctx: FLContext):
```
chesterxgchen, line 55: Adding new argument, does this still implementing the controller Spec ?  the controller_spec seems not changed with new argument.  If spec is changed, we also have backward compatibility issue.

nvflare/apis/impl/task_controller.py
```diff
@@ -135,6 +141,8 @@ def broadcast_and_wait(
             secure=task.secure,
         )
 
+        delete_msg_root(msg_root_id)
```
chesterxgchen, line 142: at this point, is the msg_root_id no longer needed ?  is the aug_request is blocking ? if not blocking, is the next line delete msg_root_id safe to delete ?

nvflare/apis/impl/wf_comm_server.py
```diff
@@ -956,6 +959,8 @@ def _do_check_tasks(self):
 
                     if exit_task.task_done_cb is not None:
                         try:
+                            msg_root_id = exit_task.msg_root_id
+                            delete_msg_root(msg_root_id)
```
chesterxgchen, line 963: i guess I am not sure what delete_msg_root() intend to do then.

nvflare/fuel/data_event/data_bus.py
```diff
@@ -63,7 +68,47 @@ def subscribe(self, topics: List[str], callback: Callable[[str, Any, "DataBus"],
             with self._lock:
                 if topic not in self.subscribers:
                     self.subscribers[topic] = []
-                self.subscribers[topic].append(callback)
+                self.subscribers[topic].append((callback, cb_kwargs))
```
chesterxgchen, line 71: the callback becomes tuple of callback, cb_kwargs, we need to check if all places using the callbacks as it could be broken with this change

nvflare/apis/impl/task_controller.py
yanchengnv: No, this is not to add new arg - this is to fix the very old bug that missed this arg.

nvflare/apis/impl/task_controller.py
yanchengnv: In this case, the aux msg itself is the actual root. So when this msg is finished, we delete the msg root. And yes, the send_aux_request call is blocking. By the time it returns, the receivers must have finished processing the request, so it is safe to delete the msg root.

nvflare/apis/impl/wf_comm_server.py
yanchengnv: The msg root id is used manage all temporary resources (e.g. files) created during the serialization process. These files should be removed after all msg receivers finished request processing. But only applications know when to delete the msg root. In this case, the wf_comm_server knows to delete the msg root associated with the task when the task itself is finished.  For example, in case of SAG, the task will be standing there waiting for all clients to get this task. No matter how long this takes, the task eventually will end. When the task ends, all those temp files should be deleted as well.

nvflare/fuel/data_event/data_bus.py
yanchengnv: Only two places use this: subscribe and publish.

yanchengnv:
> add a few question. One part I probably missed is how the object is downloaded once the file ref_id is received.

On the sender side, all target objects have been changed to Item IDs. At the end of the serialization, a new special datum is added that specifies the file ref of the generated temp file.

On the receiver side, the deserialization process has been updated to:
- extract all sections (same as before), including all datums
- process all datums (new). When processing that special datum, the referenced file is downloaded and processed. This will produce a dict of Item ID => Target Object
- the recomposition process will look up the dict and replace the Item ID with the target object

nvflare/apis/impl/task_controller.py
```diff
@@ -126,6 +127,10 @@ def broadcast_and_wait(
             raise ValueError(f"The task timeout must > 0. But got {task.timeout}")
 
         request.set_header(ReservedKey.TASK_NAME, task.name)
+        msg_root_id = str(uuid.uuid4())
```
nvidianz: A msg_root_id is already created in controller_spec. Can we just use that?

nvflare/fuel/f3/cellnet/cell.py
```diff
@@ -118,6 +119,18 @@ def __init__(self, *args, **kwargs):
         self.requests_dict = dict()
         self.logger = get_obj_logger(self)
         self.register_blob_cb(CellChannel.RETURN_ONLY, "*", self._process_reply)  # this should be one-time registration
+        self.core_cell.update_fobs_context({FOBSContextKey.CELL: self})
+
+    def update_fobs_context(self, props: dict):
+        self.core_cell.update_fobs_context(props)
+
+    def get_fobs_context(self, props: dict = None):
```
nvidianz, line 127: The props is not used here. Should update_fobs_context() be called?

nvflare/apis/impl/task_controller.py
yanchengnv: Good catch. Changed.

nvflare/fuel/f3/cellnet/cell.py
yanchengnv: Good catch. Fixed.

PR #3545, opened 2025-07-15
Changed shell commands to call command directly

nvidianz:
### Description

Changed shell commands to call the OS command directly without using shell. This can 
prevent the shell injection like this,

   grep server ;bash${IFS}-c${IFS}who; test.txt

Calling ShellCommandValidator in client again in case the command is sent from 
malicious code.

All the commands works exactly the same.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/apis/controller_spec.py
+++ b/nvflare/apis/controller_spec.py
@@ -151,6 +151,7 @@ class Task(object):
         self.after_task_sent_cb = after_task_sent_cb
         self.result_received_cb = result_received_cb
         self.task_done_cb = task_done_cb
+        self.msg_root_id = str(uuid.uuid4())
 
         self.targets = None
         self.client_tasks = []  # list of ClientTasks sent
--- a/nvflare/apis/fl_constant.py
+++ b/nvflare/apis/fl_constant.py
@@ -214,6 +214,8 @@ class ReservedTopic(object):
     JOB_HEART_BEAT = "__job_heartbeat__"
     TASK_CHECK = "__task_check__"
     APP_METRICS = "__app_metrics__"
+    MSG_ROOT_DELETED = "__msg_root_deleted__"
+    STOP_CELLNET = "__stop_cellnet__"
 
 
 class AdminCommandNames(object):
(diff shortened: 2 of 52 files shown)


nvflare/private/fed/utils/fed_utils.py
```diff
@@ -446,3 +450,17 @@ def get_job_launcher(job_meta: dict, fl_ctx: FLContext) -> JobLauncherSpec:
...
 
     return job_launcher[0]
+
+
+def execute_command_directly(command: str) -> str:
+    """Execute a command directly, without using shell"""
+    args = shlex.split(command)
+
+    if args[0] not in ALLOWED_SHELL_COMMANDS:
+        return f"Error: {args[0]} is not allowed"
```
YuanTingHsieh: good idea here, should we also add some simple verification for other args for example:

```
def is_valid_arg(arg):
    forbidden = [';', '&', '|', '`', '$', '>', '<']
    return not any(c in arg for c in forbidden)

def execute_command_directly(command: str) -> str:
    args = shlex.split(command)
    cmd = args[0]

    if cmd not in ALLOWED_SHELL_COMMANDS:
        return f"Error: {cmd} is not allowed"

    if not all(is_valid_arg(arg) for arg in args[1:]):
        return "Error: invalid argument detected"
```

and could you add some unit tests for this new method to verify its effectiveness?

YuanTingHsieh (commented):
LGTM, suggest minor improvements and adding unit tests

nvflare/private/fed/utils/fed_utils.py
nvidianz: All those forbidden characters may be valid in grep pattern so this may not work. Adding validation to the code and that may address your concerns.

nvflare/private/fed/server/shell_cmd.py
```diff
@@ -85,7 +86,8 @@ def execute_command(self, conn: Connection, args: List[str]):
         shell_cmd = conn.get_prop("shell_cmd")
         if target == SiteType.SERVER:
             # run the shell command on server
-            output = subprocess.getoutput(shell_cmd)
+            cmd_args = shlex.split(shell_cmd)
+            output = execute_command_directly(cmd_args)
```
YuanTingHsieh, line 90: we might also need to call validators here for the server?
if that's the case for both, we could move common logic into a share place and reuse that code

nvidianz: Server validator is called inside the admin command framework. So it would be called twice on server if code is moved here.

YuanTingHsieh (approved):

PR #3546, opened 2025-07-15
Add dynamic nonce in both snp and gpu tokens.

IsaacYangSLA:
### Description

This PR adds one more check on the nonce value in the cross-site token verification stage.  The nonce is embedded into `reported data` for snpguest report and nonce parameter in the client of nvtrust.  At verification stage, the nonce is extracted from the received tokens, compared with previously received tokens.  If it matches any, then this token is considered an used one, which is likely a replay attack to the system.  Therefore, the verification will fail.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/app_opt/confidential_computing/gpu_authorizer.py
+++ b/nvflare/app_opt/confidential_computing/gpu_authorizer.py
@@ -22,6 +22,8 @@ from nv_attestation_sdk import attestation
 
 from nvflare.app_opt.confidential_computing.cc_authorizer import CCAuthorizer
 
+from .utils import NonceHistory
+
 GPU_NAMESPACE = "x-nv-gpu"
 default_policy = """{
   "version":"1.0",
@@ -57,12 +59,15 @@ default_policy = """{
 
 
 class GPUAuthorizer(CCAuthorizer):
-    def __init__(self, verifier_url="https://nras.attestation.nvidia.com/v1/attest/gpu", policy_file=None):
+    def __init__(
+        self, verifier_url="https://nras.attestation.nvidia.com/v1/attest/gpu", policy_file=None, max_nonce_history=1000
+    ):
         self._can_generate = True
         self.client = attestation.Attestation()
         self.client.set_name("nvflare_node")
-        nonce = uuid.uuid4().hex + uuid.uuid1().hex
-        self.client.set_nonce(nonce)
+        self.my_nonce_history = NonceHistory(max_nonce_history)
+        self.seen_nonce_history = NonceHistory(max_nonce_history)
+
         if policy_file is None:
             self.remote_att_result_policy = default_policy
         else:
@@ -72,6 +77,9 @@ class GPUAuthorizer(CCAuthorizer):
 
     def generate(self):
         try:
+            nonce = uuid.uuid4().hex + uuid.uuid1().hex
+            self.client.set_nonce(nonce)
+            self.my_nonce_history.add(nonce)
             self.client.attest()
             token = self.client.get_token()
         except BaseException:
@@ -85,6 +93,8 @@ class GPUAuthorizer(CCAuthorizer):
             claims = jwt.decode(jwt_token.get("REMOTE_GPU_CLAIMS"), options={"verify_signature": False})
             # With claims, we will retrieve the nonce
             nonce = claims.get("eat_nonce")
+            if not self.seen_nonce_history.add(nonce):
+                return False
             self.client.set_nonce(nonce)
             self.client.set_token(name="nvflare_node", eat_token=eat_token)
             result = self.client.validate_token(self.remote_att_result_policy)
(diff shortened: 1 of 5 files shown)


nvflare/app_opt/confidential_computing/gpu_authorizer.py
```diff
@@ -61,8 +63,11 @@ def __init__(self, verifier_url="https://nras.attestation.nvidia.com/v1/attest/g
         self._can_generate = True
         self.client = attestation.Attestation()
         self.client.set_name("nvflare_node")
-        nonce = uuid.uuid4().hex + uuid.uuid1().hex
-        self.client.set_nonce(nonce)
+        self.my_nonce_list = list()
```
chesterxgchen: this logics are almost identical for GPU or SNP authorizer, with small variation. 
Image we have ITA authorizer, MAA Authorizer or GCP Authorizer etc. 

We will duplicating the same logics N times. is possible to consolidate them into a function lib that can be used in all authorizers ?

IsaacYangSLA: That will be my next plan as we might need to do more investigation on the behavior of each attestation service.  After fully understanding their behaviors, we will abstract that portion to a better structure.

chesterxgchen: I was hoping you not duplicate among the two (SNP and GPU). and adapt more when more add to the system

IsaacYangSLA: I slightly abstract more into a util class that handles nonce history.

nvflare/app_opt/confidential_computing/utils.py
```diff
@@ -0,0 +1,31 @@
...
+    def __init__(self, size=100):
+        self.history_list = list()
+        self.history_size = size
+
+    def add(self, nonce):
+        if nonce in self.history_list:
+            return False
+        self.history_list.append(nonce)
+        self.normalize_history()
+        return True
```
YuanTingHsieh: suggest using deque:

```
from collections import deque

class NonceHistory:
    def __init__(self, size=100):
        self.history_list = deque(maxlen=2 * size)

    def add(self, nonce):
        if nonce in self.history_list:
            return False
        self.history_list.append(nonce)
        return True
```

nvflare/app_opt/confidential_computing/utils.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def normalize_history(self):
+        if len(self.history_list) > 2 * self.history_size:
+            self.history_list == self.history_list[-self.history_size :]
```
YuanTingHsieh: if sticking with current impl., should this be "=" instead of "=="?

nvflare/app_opt/confidential_computing/snp_authorizer.py
```diff
@@ -29,14 +32,18 @@ class SNPAuthorizer(CCAuthorizer):
     def __init__(self):
         super().__init__()
         self.logger = logging.getLogger(self.__class__.__name__)
+        self.my_nonce_history = NonceHistory(100)
+        self.seen_nonce_history = NonceHistory(5000)
```
YuanTingHsieh: could we either introduce constants for these default numbers or make them part of the init args of this class?

nvflare/app_opt/confidential_computing/gpu_authorizer.py
```diff
@@ -61,8 +63,9 @@ def __init__(self, verifier_url="https://nras.attestation.nvidia.com/v1/attest/g
         self._can_generate = True
         self.client = attestation.Attestation()
         self.client.set_name("nvflare_node")
-        nonce = uuid.uuid4().hex + uuid.uuid1().hex
-        self.client.set_nonce(nonce)
+        self.my_nonce_history = NonceHistory(100)
+        self.seen_nonce_history = NonceHistory(5000)
```
YuanTingHsieh: could we either introduce constants for these default numbers or make them part of the init args of this class?

nvflare/app_opt/confidential_computing/utils.py
IsaacYangSLA: Good suggestion.  It's a better option.

nvflare/app_opt/confidential_computing/utils.py
IsaacYangSLA: This function was removed after this class switched to deque.

nvflare/app_opt/confidential_computing/gpu_authorizer.py
IsaacYangSLA: Both will be specified by one argument in the `__init__` now.

nvflare/app_opt/confidential_computing/snp_authorizer.py
IsaacYangSLA: Both will be specified by one argument in the `__init__` now.

YuanTingHsieh (approved):
Thanks LGTM

PR #3552, opened 2025-07-18
Fix max client limit to exclude admin clients

yanchengnv:
Fixes # .

### Description

This PR fixes the bug 4 (reported in https://nvbugspro.nvidia.com/bug/5390831) that failed to allow admin clients once max number of FL clients are connected.  Admin clients will not be counted as FL clients.

This PR also attempts to fix bug 5 in the report. Though we couldn't reproduce the issue following QA instructions, we found and fixed a place that could potentially cause this.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/fuel/f3/cellnet/net_agent.py
+++ b/nvflare/fuel/f3/cellnet/net_agent.py
@@ -29,6 +29,7 @@ from nvflare.fuel.f3.cellnet.defs import MessageHeaderKey, ReturnCode
 from nvflare.fuel.f3.cellnet.fqcn import FQCN
 from nvflare.fuel.f3.cellnet.utils import make_reply
 from nvflare.fuel.f3.stats_pool import StatsPoolManager
+from nvflare.fuel.utils.admin_name_utils import is_valid_admin_client_name
 from nvflare.fuel.utils.config_service import ConfigService
 from nvflare.fuel.utils.log_utils import get_obj_logger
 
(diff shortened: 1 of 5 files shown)


nvflare/fuel/flare_api/flare_api.py
```diff
@@ -492,6 +495,7 @@ def shutdown(self, target_type: TargetType, client_names: Optional[List[str]] =
 
         command = " ".join(parts)
         self._do_command(command)
+        self.close()
```
GeorgeWang-nv: Do we need to check the TargetType before calling api.close() and execute api.close() only when TargetType is 'all' or 'server'? The connection should remain open if only the clients are being shut down. Additionally, attempting to shut down the server while clients are still connected will raise an error and skip the api.close() operation.
if target_type != TargetType.CLIENT:
    self.close()

yanchengnv: You are absolutely right! Thanks for catching this. Fixed.

nvflare/fuel/flare_api/flare_api.py
```diff
@@ -153,7 +153,10 @@ def _do_command(self, command: str, enforce_meta=True, props=None):
             elif cmd_status == MetaStatusValue.INTERNAL_ERROR:
                 raise InternalError(f"server internal error: {info}")
             elif cmd_status == MetaStatusValue.INVALID_TARGET:
-                raise InvalidTarget(info)
+                if info == "no clients available":
```
YuanTingHsieh: Use string comparison might be prone to errors when the source change text, where does this come from maybe we can add a comment here

YuanTingHsieh (approved):

nvflare/fuel/flare_api/flare_api.py
yanchengnv: This comes from various places historically. But perhaps we could make this str a constant and using it in all other places.

yanchengnv: Changed to use defined keywords

YuanTingHsieh: got it, good idea with ReplyKeyword

PR #3558, opened 2025-07-24
Fix preflight issues

yanchengnv:
Fixes # .

### Description

This PR fixes some of the preflight issues introduced by the admin port consolidation.

This PR modified provision system to keep the generated server startup format the same as previous version, so to male the preflight tool continue to work.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/lighter/impl/static_file.py
+++ b/nvflare/lighter/impl/static_file.py
@@ -113,11 +113,7 @@ class StaticFileBuilder(Builder):
 
         admin_port = ctx.get(CtxKey.ADMIN_PORT)
         fed_learn_port = ctx.get(CtxKey.FED_LEARN_PORT)
-        if admin_port != fed_learn_port:
-            ports = f"{fed_learn_port},{admin_port}"
-        else:
-            ports = f"{fed_learn_port}"
-        target = f"{server.name}:{ports}"
+        target = f"{server.name}:{fed_learn_port}"
         sp_end_point = f"{server.name}:{fed_learn_port}:{admin_port}"
         conn_sec = self._build_conn_properties(server, ctx)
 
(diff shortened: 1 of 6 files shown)


nvflare/lighter/impl/static_file.py
```diff
@@ -113,11 +113,7 @@ def _build_server(self, server: Participant, ctx: ProvisionContext):
 
         admin_port = ctx.get(CtxKey.ADMIN_PORT)
         fed_learn_port = ctx.get(CtxKey.FED_LEARN_PORT)
-        if admin_port != fed_learn_port:
-            ports = f"{fed_learn_port},{admin_port}"
-        else:
-            ports = f"{fed_learn_port}"
-        target = f"{server.name}:{ports}"
+        target = f"{server.name}:{fed_learn_port}"
```
chesterxgchen, line 116: is it possible, client still provide two ports ?

nvflare/lighter/templates/master_template.yml
```diff
@@ -275,6 +275,8 @@ fed_server: |
                 "target": "{~~target~~}",
                 "scheme": "{~~scheme~~}"
             },
+            "admin_server": "{~~admin_server~~}",
```
YuanTingHsieh, line 278: where are we using this "admin_server"?

nvflare/lighter/impl/static_file.py
```diff
@@ -128,6 +124,8 @@ def _build_server(self, server: Participant, ctx: ProvisionContext):
             replacement={
                 "name": project.name,
                 "target": target,
+                "admin_server": server.name,
```
YuanTingHsieh, line 127: where are we using this "admin_server"?

nvflare/lighter/impl/static_file.py
yanchengnv: Yes, it is. Customer can explicitly define a different admin port. This may be necessary for the customer to apply different network security policies on the two ports.

nvflare/lighter/templates/master_template.yml
yanchengnv: To be backward compatible with the preflight tool. Flare system does not need this.

nvflare/lighter/impl/static_file.py
yanchengnv: See my reply above.

PR #3561, opened 2025-07-28
Add an iOS ExampleApp and iOS NVFlareSDK

YuanTingHsieh:
Add an iOS ExampleApp to showcase how to use iOS NVFlareSDK

### Description

- Remove old NVFlareMobile which mix App code with SDK code
- Add a new ExampleProject which contains ExampleApp code
- We can think how to distribute this SDK code, right now this PR assume the SDK codes is copied inside the App project code

### Types of changes

- [ ] Documentation updated.


nvflare/edge/device/ios/ExampleProject/NVFlareSDK/Core/NVFlareConfig.swift
```diff
@@ -0,0 +1,198 @@
+//
+//  NVFlareConfig.swift
+//  NVFlare iOS SDK
```
yanchengnv: Why is this file in "ExampleProject"? This is the core SDK. It should not be in example.

nvflare/edge/device/ios/ExampleProject/NVFlareSDK/Core/NVFlareConnection.swift
```diff
@@ -1,33 +1,36 @@
 //
-//  Connection.swift
-//  NVFlareMobile
+//  NVFlareConnection.swift
+//  NVFlare iOS SDK
```
yanchengnv: Shouldn't be in ExampleProject.

PR #3562, opened 2025-07-28
Add functionality of adding individual args to JobAPI

ZiyueXu77:
### Description

Current JobAPI does not have the functionality to add individual configs (e.g. "submit_task_result_timeout = 300") to job config jsons at top level. But it is in some cases needed - e.g. llm model transmission. So adding this function by considering additional args as a dict input to ".to" function.

### Types of changes

- [ ] Documentation updated.


yanchengnv (approved):
This is a good feature.
We need to somehow let users know what params are available.
They are all defined in nvflare.apis.fl_constant.ConfigVarName.
We should expose them in our official document.

PR #3566, opened 2025-07-29
Fix msg root creation in task

yanchengnv:
Fixes # .

### Description

Previously, a msg root id is always created for a workflow task, assuming that the task data is never changed during the life of the task. This way, all messages originated from the same task can share the serialization result from the 1st message, hence saving time and memory usage.

Unfortunately, this assumption is not always true. For example, the cross-site validation controller reloads task data from disk every time it gets a request from a client for the same task).

This PR tries to fix this issue without losing the benefit of msg root sharing:

- Since task data modification always happens within the task's before_task_sent_cb, if this cb is not specified, then the task data is not mutable.
- If the cb is specified, then the app can further specify whether the task data is mutable explicitly when creating the task. For example, all variants of SAG specify the before_task_sent_cb, but they do not modify task data. 

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/app_common/workflows/scatter_and_gather.py
+++ b/nvflare/app_common/workflows/scatter_and_gather.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import gc
+import uuid
 from typing import Any
 
 from nvflare.apis.client import Client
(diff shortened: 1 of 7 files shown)


nvflare/app_common/workflows/scatter_and_gather.py
```diff
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import gc
+import uuid
```
YuanTingHsieh: need to remove this unused import

yanchengnv: Don't approve yet. I am thinking about a different cleaner way.

yanchengnv:
Close this PR. Need a better solution.

PR #3570, opened 2025-07-30
Add support for recipes

yanchengnv:
Fixes # .

### Description

This PR adds a simple framework for building recipes and a few example recipes.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/recipe/spec.py
+++ b/nvflare/recipe/spec.py
@@ -37,7 +37,7 @@ class Recipe(ABC):
         """
         self.job = job
 
-    def add_client_input_filter(self, filter: Filter, tasks=None):
+    def add_input_filter_to_clients(self, filter: Filter, tasks=None):
         """Add a filter to clients for incoming tasks from the server.
 
         Args:
@@ -49,7 +49,7 @@ class Recipe(ABC):
         """
         self.job.to_clients(filter, filter_type=FilterType.TASK_DATA, tasks=tasks)
 
-    def add_client_output_filter(self, filter: Filter, tasks=None):
+    def add_output_filter_to_clients(self, filter: Filter, tasks=None):
         """Add a filter to clients for outgoing result to server.
 
         Args:
@@ -61,7 +61,7 @@ class Recipe(ABC):
         """
         self.job.to_clients(filter, filter_type=FilterType.TASK_RESULT, tasks=tasks)
 
-    def add_server_output_filter(self, filter: Filter, tasks=None):
+    def add_output_filter_to_server(self, filter: Filter, tasks=None):
         """Add a filter to the server for outgoing tasks to clients.
 
         Args:
@@ -73,7 +73,7 @@ class Recipe(ABC):
         """
         self.job.to_server(filter, filter_type=FilterType.TASK_DATA, tasks=tasks)
 
-    def add_server_input_filter(self, filter: Filter, tasks=None):
+    def add_input_filter_to_server(self, filter: Filter, tasks=None):
         """Add a filter to server for incoming task result from clients. .
 
         Args:


YuanTingHsieh (approved):
LGTM

nvflare/recipe/spec.py
```diff
@@ -0,0 +1,50 @@
...
+class Recipe(ABC):
+
+    def __init__(self, job: FedJob):
+        self.job = job
+
+    def add_client_data_filter(self, filter: Filter, tasks=None):
+        self.job.to_clients(filter, filter_type=FilterType.TASK_DATA, tasks=tasks)
+
+    def add_client_result_filter(self, filter: Filter, tasks=None):
+        self.job.to_clients(filter, filter_type=FilterType.TASK_RESULT, tasks=tasks)
```
YuanTingHsieh: name of these methods, we might want input/output VS data/result, since data/result is a NVFlare specific concept

yanchengnv: good suggestion. Changed.

nvidianz (approved):

nvflare/recipe/spec.py
```diff
@@ -0,0 +1,124 @@
...
+        """
+        if server_exec_params:
+            self.job.to_server(server_exec_params)
+
+        if client_exec_params:
+            self.job.to_clients(client_exec_params)
+
+        self.job.export_job(job_dir)
+
+    def execute(self, env: ExecEnv, server_exec_params: dict = None, client_exec_params: dict = None) -> Any:
```
chesterxgchen, line 107: should add default arguments to make it easier for customer to write something like

    recipe.execute(clients=n_clients)


    def execute(
        self,
        env: Optional[ExecEnv] = None,
        clients: Union[int, List[str]] = None,
        gpus: Union[int, List[int]] = None,
        workspace_dir: Optional[str] = None,
    ):

        if env is None:
            env = SimEnv(clients=clients, gpus=gpus, workspace_dir=workspace_dir)

       ... rest of code ?

chesterxgchen: why server exec params be part of the execute () argument ? and client exec args is not ? this is not symetry. 

Seems server exec args should be arg of the server script args. Which is part of the recipe args. not execute args

nvflare/recipe/prod_env.py
```diff
@@ -0,0 +1,52 @@
...
+from .spec import ExecEnv
+
+
+class ProdExecEnv(ExecEnv):
+
+    def __init__(
+        self,
+        startup_kit_dir: str,
+        login_timeout: float = 5.0,
+        monitor_job_duration: int = None,
```
chesterxgchen, line 29: this should have settings (or config arguments) for the product for time out etc.

nvflare/recipe/simulation_env.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+class SimulationExecEnv(ExecEnv):
+
+    def __init__(
+        self,
+        workspace: str,
+        num_clients: int = None,
+        num_threads: int = None,
+        gpu_config: str = None,
```
chesterxgchen, line 26: gpus could be array or str

nvflare/recipe/poc_env.py
```diff
@@ -0,0 +1,27 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from nvflare.job_config.api import FedJob
+from nvflare.recipe.spec import ExecEnv
+
+
+class POCExecEnv(ExecEnv):
+
+    def __init__(self, num_clients: int = 2):
```
chesterxgchen, line 20: need workspace directory for POC as argument

nvflare/app_opt/pt/recipes/fed_avg.py
```diff
@@ -0,0 +1,42 @@
...
+class FedAvgRecipe(Recipe):
+
+    def __init__(
+        self,
+        name: str,
+        initial_model,
+        num_rounds: int,
+        min_clients: int,
+        train_script: str,
+        train_args: dict = None,
```
chesterxgchen, line 28: there are abunch of other arguments needed  early stop condition, patients, aggr_fn etc

yanchengnv: These can be handled by other recipes. This recipe is just an example for the simplest fedavg (from Holger). We could add more types of fedavg recipes, on as-needed basis.
As we discussed, recipes are cheap and we shouldn't make a big recipe that can do everything with many args.
Instead we should have many recipes, each focusing on a specific common use case.
Recipes should be simple to use - like the recipes in a restaurant, you simply "point" to it to order. Kungpao Chicken and Kungpao Shrimp are two different recipes, even though they are both part of the Kunpao series.  :-)

nvflare/recipe/poc_env.py
yanchengnv: YT will submit a PR to implement POC Env.

nvflare/recipe/prod_env.py
yanchengnv: Not sure what you mean.

nvflare/recipe/spec.py
yanchengnv: Disagree. I think we should keep it simple.
Besides not all these things apply to all envs.
Each type of ENV could provide some easy/simple methods for commonly used env config. But that belongs to env impl, not here.

nvflare/recipe/prod_env.py
chesterxgchen: I am saying there are potentially many configuration key value pairs, should we have an argument called "settings" or config which allows one to pass different configuration values, such as timeout for different connection etc.

yanchengnv: Yeah, I was debating with myself where these things should be placed. I added them to the execute() call as server_exec_params and client_exec_params, since they could be applicable to all env types, not just Prod Env.
We could add them as common args to the base class ExecEnv.
Bottom line is that these exec params are needed and supported.

PR #3572, opened 2025-07-31
detect and handle msg root mutation

yanchengnv:
Fixes # .

### Description

Secondary message's payload could be different from the primary message's payload under the same msg root. This could be caused, for example, by a task's before_task_sent callback or by any filters that change payload content.

In general, even if target object IDs are not changed, it is still risky to assume that the content of the object is not changed.

This PR removes reuse of primary message result. Instead, each message is treated independently. If msg root is specified, then all files generated from all messages under the same msg root id will be deleted when the msg root is deleted.

This PR also fixes the following issues:
- When submitting task result, set the task name to the name of the received name. Previously the name was taken from fl_ctx, which could be stuck with a different name during workflow transition.
- Change byte_streamer to log error msgs in DEBUG level if the txTask is optional. This is to suppress unnecessary error msgs.

### Types of changes

- [ ] Documentation updated.


nvflare/private/fed/client/client_runner.py
```diff
@@ -220,6 +220,12 @@ def _reply_and_audit(reply: Shareable, ref, msg, fl_ctx: FLContext) -> Shareable
         return reply
 
     def _process_task(self, task: TaskAssignment, fl_ctx: FLContext) -> Shareable:
+        reply = self._do_process_task(task, fl_ctx)
+        reply.set_header(ReservedHeaderKey.TASK_NAME, task.name)
+        reply.set_header(ReservedHeaderKey.TASK_ID, task.task_id)
+        return reply
+
+    def _do_process_task(self, task: TaskAssignment, fl_ctx: FLContext) -> Shareable:
```
YuanTingHsieh: why this another layer?
the end of `_do_process_task` already have 
```
reply.set_header(ReservedHeaderKey.TASK_NAME, task.name)
reply.set_header(ReservedHeaderKey.TASK_ID, task.task_id)
```

YuanTingHsieh (commented):
LGTM @nvidianz please check as you are also testing on this thanks

nvflare/private/fed/client/client_runner.py
yanchengnv: Updated the code.
The reason is because _do_process_task could return at any time without going to the end.

YuanTingHsieh (approved):
LGTM

nvflare/fuel/utils/fobs/decomposers/via_file.py
```diff
@@ -68,45 +57,26 @@ def __init__(self):
         self.target_to_item = {}  # target_id => item_id
         self.target_items = {}  # item_id => item value
         self.last_item_id = 0
```
nvidianz, line 59: Do we need a lock for last_item_id? If it's used in multiple threads, we might get duplicate item_id

nvflare/fuel/utils/fobs/decomposers/via_file.py
```diff
@@ -68,45 +57,26 @@ def __init__(self):
...
-            return item_id
-
-        item_id = f"T{self.last_item_id}"
-        self.last_item_id += 1
-        self.target_items[item_id] = item
-        self.target_to_item[target_id] = item_id
-        return item_id
+        if not item_id:
+            item_id = f"T{self.last_item_id}"
+            self.last_item_id += 1
```
nvidianz, line 76: This line is not thread-safe

nvidianz (approved):

nvflare/fuel/utils/fobs/decomposers/via_file.py
yanchengnv: No, this is for each encoding thread separately.

nvflare/fuel/utils/fobs/decomposers/via_file.py
yanchengnv: No thread issue here since FOBS processes all items sequentially for the same message.

PR #3573, opened 2025-08-02
Apply ruff fixes

cyyever:
### Description

Apply RUFF fixes to nvflare files.

### Types of changes


nvflare/app_common/logging/log_receiver.py
```diff
@@ -35,7 +35,6 @@ def process_log(self, stream_ctx: StreamContext, fl_ctx: FLContext):
         peer_ctx = fl_ctx.get_peer_context()
         assert isinstance(peer_ctx, FLContext)
         peer_name = peer_ctx.get_identity_name()
-        channel = FileStreamer.get_channel(stream_ctx)
         topic = FileStreamer.get_topic(stream_ctx)
```
chesterxgchen, line 38: channel is not used ? @yanchengnv

chesterxgchen:
@YuanTingHsieh  do we use RUFF linter and formatter ?

nvflare/app_common/logging/log_receiver.py
yanchengnv: Both topic and channel should be used in the error log msg, when error happens. Currently the log msg only records the topic. It should be changed to also record channel to something like:

self.log_error(fl_ctx, f"Error in streaming log file from {peer_name} and topic {topic}/{channel}: {rc}",)

YuanTingHsieh: ok, we can merge this PR, I will submit another PR to update the error msg^

YuanTingHsieh (dismissed):

YuanTingHsieh:
@cyyever thanks for the fix, can you rebase then resolve the conflicts in "nvflare/lighter/cc_provision/impl/onprem_packager.py"?

YuanTingHsieh:
> @YuanTingHsieh do we use RUFF linter and formatter ?

we are not running it now in ci, we could add it

cyyever:
@YuanTingHsieh Fixed it.

PR #3574, opened 2025-08-04
Added env-var to handle GRPC error

nvidianz:
### Description

Made 2 changes to grpc and aio grpc drivers to address the GRPC errors we see in test and CI,

1. Set default values for GRPC_ENABLE_FORK_SUPPORT and GRPC_POLL_STRATEGY to get around a GRPC bug.
2. Use setdefault so other values can be set in env-var.

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh:
changes included in #3569

PR #3582, opened 2025-08-07
Fix FedJob _get_args method

YuanTingHsieh:
### Description

This PR fixes a critical bug in `FedJob`'s component serialization logic where direct `self.__dict__` access fails for classes that don't store attributes in the standard dictionary, such as PyTorch `nn.Module` and other framework classes that override `__setattr__` behavior.

### Problem

The original code used `component.__dict__` to access component attributes during serialization. However, many frameworks (especially PyTorch) override `__setattr__` to store attributes in custom data structures rather than `self.__dict__`. This caused:

- `KeyError` when accessing attributes that exist but aren't in `__dict__`
- Failed serialization of PyTorch models and other framework components
- Inconsistent behavior across different class types

### Solution

- **Replace `__dict__` access with `getattr()`**: Uses Python's standard attribute resolution mechanism
- **Add proper exception handling**: Gracefully skips attributes that don't exist rather than crashing
- **Cleaner parameter filtering**: Explicitly exclude `"self"` along with `"args"` and `"kwargs"`
- **Universal compatibility**: Works with any class regardless of internal attribute storage

### Key Changes

```python
# Before: Direct dictionary access (fragile)
attr_key = param if param in attrs.keys() else "_" + param
if attr_key in attrs.keys() and parameters[param].default != attrs[attr_key]:
    # Process attrs[attr_key]

# After: Standard attribute access (robust)
try:
    attr_value = getattr(component, param)
except AttributeError:
    continue
if parameters[param].default != attr_value:
    # Process attr_value
```

### Benefits

- ✅ **Framework agnostic**: Works with PyTorch, TensorFlow, scikit-learn, and any other framework
- ✅ **Backward compatible**: Existing code continues to work unchanged
- ✅ **More robust**: Proper exception handling prevents crashes
- ✅ **Cleaner logic**: Simplified parameter handling without complex key mapping

### Impact

This change enables proper serialization of:
- PyTorch `nn.Module` components
- Any class that uses custom `__setattr__` implementations
- Standard Python classes (existing functionality preserved)

Without this fix, users cannot serialize FedJobs containing PyTorch models or other framework components, making the feature unusable for most deep learning workflows.
### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/job_config/fed_job_config.py
+++ b/nvflare/job_config/fed_job_config.py
@@ -378,24 +378,32 @@ class FedJobConfig:
         args = {}
         if hasattr(component, "__dict__"):
             parameters = get_component_init_parameters(component)
-            attrs = component.__dict__
-
             for param in parameters:
-                attr_key = param if param in attrs.keys() else "_" + param
-
-                if attr_key in ["args", "kwargs"]:
+                if param in ["self", "args", "kwargs"]:
                     continue
 
-                if attr_key in attrs.keys() and parameters[param].default != attrs[attr_key]:
-                    if type(attrs[attr_key]).__name__ in dir(builtins):
-                        args[param] = attrs[attr_key]
-                    elif issubclass(attrs[attr_key].__class__, Enum):
-                        args[param] = attrs[attr_key].value
+                # Use getattr() for universal attribute access - works with any framework
+                attr_value = None
+
+                try:
+                    attr_value = getattr(component, param)
+                except AttributeError:
+                    continue
+
+                if parameters[param].default != attr_value:
+                    if type(attr_value).__name__ in dir(builtins):
+                        args[param] = attr_value
+                    elif issubclass(attr_value.__class__, Enum):
+                        args[param] = attr_value.value
                     else:
                         args[param] = {
-                            "path": self._get_class_path(attrs[attr_key], custom_dir),
-                            "args": self._get_args(attrs[attr_key], custom_dir),
+                            "path": self._get_class_path(attr_value, custom_dir),
+                            "args": self._get_args(attr_value, custom_dir),
                         }
+                else:
+                    pass
+        else:
+            pass
 
         return args
 


nvflare/job_config/fed_job_config.py
```diff
@@ -378,23 +379,24 @@ def _get_args(self, component, custom_dir):
         args = {}
         if hasattr(component, "__dict__"):
             parameters = get_component_init_parameters(component)
-            attrs = component.__dict__
-
             for param in parameters:
-                attr_key = param if param in attrs.keys() else "_" + param
```
yanchengnv, line 402: We need to keep this to determine attr_key.
This trick is for the case that many components keep their init arg values in local var _xxx, where xxx is the original arg name.

nvflare/job_config/fed_job_config.py
```diff
@@ -378,23 +379,24 @@ def _get_args(self, component, custom_dir):
...
+                if param in ["self", "args", "kwargs"]:
```
yanchengnv: Change to use attr_key for param.

nvflare/job_config/fed_job_config.py
```diff
@@ -378,23 +379,24 @@ def _get_args(self, component, custom_dir):
...
+                    continue
 
-                if attr_key in ["args", "kwargs"]:
+                try:
+                    attr_value = getattr(component, param)
```
yanchengnv: attr_value = getattr(component, attr_key)

nvflare/job_config/fed_job_config.py
```diff
@@ -378,23 +379,24 @@ def _get_args(self, component, custom_dir):
...
+                except AttributeError:
```
yanchengnv: Should we continue or let it crash? 

When this happens, it means that we cannot determine the value of an init arg. Basically the value of the init arg is either ignored by the constructor or is used dynamically to compute something but is never kept. 

If the init arg is simply ignored by the constructor, then it's OK to continue here.
But if the arg is used but not kept (or kept under a very different name), then we shouldn't continue here since we won't be able to construct the object properly without this arg.

I think we should keep the original behavior - let it crash.

yanchengnv (commented):
See my comments. Don't think we should continue when init arg fails to resolve.

nvflare/job_config/fed_job_config.py
YuanTingHsieh: Changed as suggested.

nvflare/job_config/fed_job_config.py
YuanTingHsieh: Make sense, changed as suggested.

nvflare/job_config/fed_job_config.py
YuanTingHsieh: make sense, changed back to original behavior

nvflare/job_config/fed_job_config.py
```diff
@@ -396,23 +397,30 @@ def _get_args(self, component, custom_dir):
         args = {}
         if hasattr(component, "__dict__"):
```
yanchengnv: Is it necessary to check this hasattr?

nvflare/job_config/fed_job_config.py
```diff
@@ -396,23 +397,30 @@ def _get_args(self, component, custom_dir):
...
+                if param in ["self", "args", "kwargs"]:
                     continue
 
-                if attr_key in attrs.keys() and parameters[param].default != attrs[attr_key]:
-                    if type(attrs[attr_key]).__name__ in dir(builtins):
-                        args[param] = attrs[attr_key]
-                    elif issubclass(attrs[attr_key].__class__, Enum):
-                        args[param] = attrs[attr_key].value
+                # Determine the correct attribute key by checking if the attribute actually exists
+                if hasattr(component, param):
```
yanchengnv: This set of code of determining attr_key is used in _add_referenced_components as well. Consider making it a reusable func.

tests/unit_test/job_config/fed_job_config_test.py
```diff
@@ -40,3 +41,119 @@ def test_trim_whitespace(self):
...
+                return self._threshold * 2  # Different from stored value to test property access
+
+        job_config = FedJobConfig(job_name="test_job", min_clients=1)
+        component = ComponentWithProperty(0.8)
+
+        args = job_config._get_args(component, custom_dir=".")
+
+        # With old __dict__ approach: threshold property would be missed
+        # With new hasattr approach: should find and access the property correctly
+        assert args["threshold"] == 1.6  # 0.8 * 2 from property logic
```
yanchengnv: This is a problem! 

Your intent is to configure the component with arg threshold = 0.5. When you export the component to job, then the arg becomes 1.6. When this job is loaded and the component constructed, the object's init arg would be 1.6, and its property "threshold" would be 3.2!

Remember that the object you created using job API must be reconstructed precisely when the job is loaded later.

In this particular example, job API won't be able to support it properly.

yanchengnv (commented):
See my comments. Looks like it may not be a good idea to use getattr after all.
Let's not rush to merge this.

nvflare/job_config/fed_job_config.py
YuanTingHsieh: removed

nvflare/job_config/fed_job_config.py
YuanTingHsieh: changed as suggested, and added some more tests.

tests/unit_test/job_config/fed_job_config_test.py
YuanTingHsieh: right, this test case should be NOT supported.
Changed implementation to raise Exception

YuanTingHsieh:
yeah, we can wait after 2.7 branch out, then this can target at the main

chesterxgchen:
@YuanTingHsieh  is this PR still needed, it been here since early August

YuanTingHsieh:
close for inactive, we can revisit later

PR #3583, opened 2025-08-07
Update HUG timeout error msg

YuanTingHsieh:
### Description

This PR fixes an issue where `ETEdgeModelExecutor` (and other edge executors using `HierarchicalUpdateGatherer`) would log spurious timeout errors during normal job completion. The timeout occurs due to a race condition between task completion and final parent update attempts.

### Problem

When a federated learning job completes successfully, the following sequence can occur:
1. Parent sends `end_task` signal to child
2. Child attempts one final update to parent  
3. Parent has already finished and no longer accepts updates
4. Child receives `TIMEOUT` return code and logs it as an ERROR

This creates confusing error logs like:
```
ETEdgeModelExecutor - ERROR - error updating parent None: TIMEOUT, self.update_timeout=1000.0
```

Even though the job completed successfully, these timeout errors suggest something went wrong.

### Solution

Modified the error handling in `HierarchicalUpdateGatherer._do_task()` to treat `TIMEOUT` return codes gracefully:

- **Expected timeouts** (after `_task_done` is set): Log as INFO with explanatory message
- **Unexpected timeouts** (during active task): Log as WARNING suggesting parent may have finished
- **Other errors**: Continue logging as ERROR as before

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/edge/executors/hug.py
+++ b/nvflare/edge/executors/hug.py
@@ -315,8 +315,22 @@ class HierarchicalUpdateGatherer(Executor):
                 break
 
             if rc != ReturnCode.OK:
-                self.log_error(fl_ctx, f"error updating parent {self._parent_name}: {rc}")
-                break
+                # Handle TIMEOUT gracefully when parent is no longer accepting updates
+                # This commonly happens when the parent has finished the job or when there's a race condition
+                # between task completion and the final update attempt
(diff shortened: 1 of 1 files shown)


yanchengnv (approved):
LGTM

PR #3586, opened 2025-08-07
Support app packages in job API

yanchengnv:
Fixes # .

### Description

Currently, if a module is not in NVFLARE package, the Job API applies the magic of determining its class path and copying the source into the "custom" folder of generated job. When the module is in an app package that will be included in the final python env (e.g. docker container), then this magic is not desired. In fact, including the source code could cause confusion when resolving class path.

This PR implements a simple solution by letting the Job API user specify app packages to be excluded from this behavior.
If the package name of a module is in the specified app package list, then the magic won't be applied. 

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/job_config/fed_job_config.py
+++ b/nvflare/job_config/fed_job_config.py
@@ -21,7 +21,7 @@ import subprocess
 import sys
 from enum import Enum
 from tempfile import TemporaryDirectory
-from typing import Dict
+from typing import Dict, List
 
 from nvflare.fuel.utils.class_utils import get_component_init_parameters
 from nvflare.fuel.utils.log_utils import get_obj_logger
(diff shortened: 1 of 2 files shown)


nvflare/job_config/fed_job_config.py
```diff
@@ -268,7 +286,7 @@ def _get_class_path(self, obj, custom_dir):
     def _get_custom_file(self, custom_dir, module, source_file):
         package = module.split(".")[0]
         if os.path.exists(source_file):
-            if package not in FL_PACKAGES and module not in self.custom_modules:
+            if package not in FL_PACKAGES and package not in self.app_packages and module not in self.custom_modules:
                 module_path = module.replace(".", os.sep)
                 if module_path in source_file:
```
nvidianz, line 291: It might be safer to do 

    if source_file.startswith(module_path):

nvidianz (approved):
Just a minor comment.

PR #3588, opened 2025-08-08
Decomposer Registration Redesign

nvidianz:
### Description

The decomposer registration is revamped to achieve following goals,

1. Fix the security vulnerability, only registered decomposers can be used.
2. No registration is needed for all built-in decomposers. This is achieve through a type mapping table defined in builtin_decomposers.py. The decomposers are instantiated on demand so all decomposers can be added to the table without cause inter-dependencies between packages.
3. Added 'decomposers' in job meta so decomposers can be registered for the job without custom code.
4. Added warning when generic decomposers are used. They don't always work.
5. Keep the backward compatibility so all the existing jobs still work.

### Types of changes

- [ ] Documentation updated.


yanchengnv:
My concern is that the scope of changes is massive. We don't have time to throughly test it for Q3. We need a simpler and safer version.

Maybe we only need white-listing for unknown decomposers. We can have auto registration in next release.

nvidianz:
The changes are too massive and not enough time to test for this release. Will make another PR after this release.

PR #3590, opened 2025-08-08
Update pt_cifar10 for edge example

ZiyueXu77:
Fixes # .

### Description

Remove old code and migrate to edge job recipe, update Readme

### Types of changes

- [ ] Documentation updated.


ZiyueXu77:
@YuanTingHsieh I am still running exp to find out a good parameter set for an example with 40000 devices, but the rest can be reviewed. BTW, I further reorganized the folder structure a bit to make it cleaner at root

examples/advanced/edge/jobs/pt_job.py
```diff
@@ -0,0 +1,192 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+
+from processors.cifar10_pt_task_processor import Cifar10PTTaskProcessor
+from processors.models.cifar10_model import Cifar10ConvNet
```
YuanTingHsieh, line 18: it feels a bit weird since this model is orthogonal to processor.
Even without any processor, we need to define this model?
Users could already have a model running somewhere without NVFlare at the first hand?
Do we want them to put the model inside processor?

ZiyueXu77: Agree, but since PT processor needs to import the model, it is a bit tricky to properly do it. ET does not have this issue as ET buffer is self-contained

YuanTingHsieh: that's true, OK let's just use this for now, can't think of a better way that works with our job.
the other way is the flat but then it will have a lot of duplicate codes for example as well.

PR #3596, opened 2025-08-12
Add POCEnv impl

YuanTingHsieh:
### Description

- Add POCEnv impl
- Change Recipe API to "add_client_input_filter", "add_client_output_filter", "add_server_input_filter", "add_server_output_filter" as @holgerroth suggested.

### Types of changes

- [ ] Documentation updated.


nvflare/recipe/poc_env.py
```diff
@@ -11,17 +11,307 @@
...
 
-    def __init__(self, num_clients: int = 2):
+    This environment sets up a POC deployment on a single machine with multiple
+    processes representing the server, clients, and admin console.
+    """
+
+    def __init__(
+        self,
+        poc_workspace: Optional[str] = None,
+        num_clients: int = 2,
```
chesterxgchen, line 83: need new argument 
```
clients: Optional[List[str]] = None
```
similar to SimEnv

nvflare/recipe/poc_env.py
```diff
@@ -11,17 +11,307 @@
...
+            # Check if POC is running, prepare and start if needed
+            if not self.is_running():
+                print("POC services not running")
+                # Only prepare workspace if POC is not running
+                self._ensure_poc_workspace()
+                print("Starting POC services...")
+                self._start_poc_services()
+                self._poc_started_by_us = True
+            else:
+                print("POC services already running")
```
chesterxgchen: if POC is ready runing, then we need to make sure it has the same number of clients as the new Job requests 

Also, need to check if the client_name is the same or not.  for example,

current running one is site-1, site-2, 
but the newly request one is bank-1, bank-2

nvflare/recipe/poc_env.py
```diff
@@ -11,17 +11,307 @@
...
+
+        if not workspace_exists or not workspace_valid:
+            print(f"POC workspace not ready, preparing it at: {self.poc_workspace}")
+            self._prepare_poc_workspace()
+
+    def _prepare_poc_workspace(self):
+        """Prepare POC workspace by running poc prepare (non-interactive)."""
+        try:
+            # Generate client names if not provided
+            clients = [f"site-{i + 1}" for i in range(self.num_clients)]
```
chesterxgchen: poc does this automatically if the cllent list is not provided. 

see 
```
  nvflare poc prepare -n 2 
```

nvflare/recipe/poc_env.py
chesterxgchen: the simplest and cleanest way is do always stop and re-start the POC. 

the drawback of this is slow.

nvflare/recipe/poc_env.py
```diff
@@ -11,17 +11,307 @@
...
+                # Submit the job
+                job_id = sess.submit_job(job_path)
+                return job_id
+            finally:
+                sess.close()
+
+        except Exception as e:
+            raise RuntimeError(f"Failed to submit job via Flare API: {e}")
+
+    def _monitor_job(self, job_id: str, duration: int):
```
yanchengnv: You can use the same sess created when submitting the job. Don't create another sess.

nvflare/recipe/poc_env.py
```diff
@@ -11,17 +11,307 @@
...
+                username="admin@nvidia.com",  # Default POC admin user
+                startup_kit_location=admin_dir,
+            )
+
+            try:
+                # Submit the job
+                job_id = sess.submit_job(job_path)
+                return job_id
+            finally:
+                sess.close()
```
yanchengnv: Don't close the sess: return it for monitor_job to use.
Close the sess after monitor_job is done.

nvflare/recipe/spec.py
```diff
@@ -37,31 +38,33 @@ def __init__(self, job: FedJob):
...
 
         Args:
             filter: the filter to be added
             tasks: tasks that the filter applies to
+            clients: client names to add
 
         Returns: None
 
         """
         self.job.to_clients(filter, filter_type=FilterType.TASK_DATA, tasks=tasks)
```
yanchengnv: If clients is explicitly specified, you cannot call job.to_clients() any more - it only works to add to ALL clients.
Instead, you need to call job.to_client() for each client in the "clients" list.

nvflare/recipe/spec.py
```diff
@@ -37,31 +38,33 @@ def __init__(self, job: FedJob):
...
 
         Args:
             filter: the filter to be added
             tasks: tasks that the filter applies to
+            clients: client names to add
 
         Returns: None
 
         """
         self.job.to_clients(filter, filter_type=FilterType.TASK_RESULT, tasks=tasks)
```
yanchengnv: If clients is explicitly specified, you cannot call job.to_clients() any more - it only works to add to ALL clients.
Instead, you need to call job.to_client() for each client in the "clients" list.

yanchengnv (changes requested):
See my comments.

nvflare/recipe/poc_env.py
YuanTingHsieh: add as suggested.

nvflare/recipe/poc_env.py
YuanTingHsieh: make sense, keep it simple for now, will always stop and then restart now.

nvflare/recipe/poc_env.py
YuanTingHsieh: changed as suggested.

nvflare/recipe/poc_env.py
YuanTingHsieh: thanks, changed as suggested

nvflare/recipe/spec.py
YuanTingHsieh: thanks, changed as suggested

nvflare/recipe/poc_env.py
```diff
@@ -11,17 +11,298 @@
...
+        """
+        # Validate client configuration
+        if clients is not None:
+            if len(clients) == 0:
+                raise ValueError("clients list cannot be empty")
+            if num_clients > 0 and len(clients) != num_clients:
+                raise ValueError(f"Inconsistent: num_clients={num_clients} but clients list has {len(clients)} entries")
+        else:
+            if num_clients <= 0:
+                raise ValueError("num_clients must be greater than 0")
```
chesterxgchen: probably should also leverage pydantic to validate all the input fields

nvflare/recipe/poc_env.py
```diff
@@ -11,17 +11,298 @@
...
+
+        except Exception as e:
+            print(f"Error deploying job to POC environment: {e}")
+            raise
+        finally:
+            # Stop and clean if auto_stop is enabled (we always start our own POC)
+            if self.auto_stop:
+                self._stop_and_clean_poc()
+
+    def _ensure_poc_workspace(self):
```
chesterxgchen: since you always start with clean, when not just call 

            shutil.rmtree(workspace, ignore_errors=True)
            
           and prepare_poc after that ?

YuanTingHsieh:
OK, I will refactor this and poc_commands.py to reduce duplicate code

nvflare/recipe/poc_env.py
YuanTingHsieh: updated

PR #3599, opened 2025-08-12
Update edge with 40k device experiment, add docstring to recipe

ZiyueXu77:
Fixes # .

### Description

- update edge with 40k device experiment under reasonable parameter settings
- add docstring to recipe class

### Types of changes

- [ ] Documentation updated.


nvflare/edge/tools/edge_recipe.py
```diff
@@ -91,6 +151,49 @@ def __init__(
...
+            job_name="my_edge_job",
+            model=MyModel(),
+            model_manager_config=model_manager_config,
+            device_manager_config=device_manager_config
+        )
+        ```
+
+    Attributes:
+        job_name: Name of the federated learning job
+        method_name: Method type, always set to "edge"
```
YuanTingHsieh: maybe we should instead explain the init method's argument.
this "method_name" let's not expose it, this might be removed.

nvflare/edge/tools/edge_recipe.py
```diff
@@ -49,6 +69,21 @@ def __init__(
...
+
+    This class configures how the device manager selects and manages devices
+    for participation in federated learning workflow.
+
+    Attributes:
+        device_selection_size: Number of devices to select for each training round.
+            Default: 100
+        min_hole_to_fill: Minimum number of model updates to wait for before
+            sampling the next batch of devices. Default: 1 (immediately dispatch the current global model)
+        device_reuse: Whether to allow devices to participate in multiple rounds.
```
YuanTingHsieh, line 92: if devide_reuse is False, then this device will only participate ONCE in a round for the whole training process?

nvflare/edge/tools/edge_recipe.py
```diff
@@ -49,6 +69,21 @@ def __init__(
...
+    """Configuration class for the device manager in federated learning.
+
+    This class configures how the device manager selects and manages devices
+    for participation in federated learning workflow.
+
+    Attributes:
+        device_selection_size: Number of devices to select for each training round.
+            Default: 100
+        min_hole_to_fill: Minimum number of model updates to wait for before
+            sampling the next batch of devices. Default: 1 (immediately dispatch the current global model)
```
YuanTingHsieh: what is this for? if I set min_hole_to_fill to be the same since as device_selection_size, then that means I am synchronous training?

suggest make this more verbose to explain, something like:

```
min_hole_to_fill: int, optional (default=1)
    The minimum number of client model updates the server waits to receive 
    before aggregating and dispatching the next global model. 

    - If set to 1, the server immediately proceeds after receiving the first update.
    - Higher values cause the server to wait for more updates, potentially improving 
      aggregation quality but increasing waiting time.

    This parameter controls the trade-off between update latency and the amount of 
    client contribution used in each federated learning round.
```

nvflare/edge/tools/edge_recipe.py
```diff
@@ -28,6 +28,26 @@
...
 class ModelManagerConfig:
+    """Configuration class for the model manager in federated learning.
+
+    This class configures how the model manager handles model updates, versioning,
+    and aggregation strategies for federated learning workflow.
+
+    Attributes:
+        max_num_active_model_versions: Maximum number of active model versions
+            that can be processed for the current model version. Default: 3
+        max_model_version: Maximum model version number before stopping training.
```
YuanTingHsieh, line 39: maybe need to add explanation somewhere saying this can not be 1 or add validation that this needs to be >=2

YuanTingHsieh (commented):
LGTM

nvflare/edge/tools/edge_recipe.py
ZiyueXu77: removed

nvflare/edge/tools/edge_recipe.py
ZiyueXu77: yes, added a sentence to describe

nvflare/edge/tools/edge_recipe.py
ZiyueXu77: this is not related to aggregation (which is controlled by model manager), added more info to describe

nvflare/edge/tools/edge_recipe.py
ZiyueXu77: added

nvflare/edge/tools/edge_recipe.py
```diff
@@ -59,11 +106,27 @@ def __init__(
     ):
         self.device_selection_size = device_selection_size
         self.min_hole_to_fill = min_hole_to_fill
+        # check if min_hole_to_fill is smaller than device_selection_size
+        if min_hole_to_fill > device_selection_size:
+            raise ValueError("min_hole_to_fill needs to be smaller than device_selection_size")
```
YuanTingHsieh: good check suggest msg:

"min_hole_to_fill needs to be smaller than or equal to device_selection_size"

YuanTingHsieh (commented):
thanks, just one suggestion

YuanTingHsieh (approved):

nvflare/edge/tools/edge_recipe.py
ZiyueXu77: right, updated

examples/advanced/edge/jobs/pt_job_adv.py
```diff
@@ -76,7 +85,10 @@ def main():
             torchvision_dataset={"name": "CIFAR10", "path": dataset_root}, eval_frequency=eval_frequency
         ),
         simulation_config=SimulationConfig(
-            task_processor=task_processor, job_timeout=20.0, num_workers=10, num_devices=devices_per_leaf
+            task_processor=task_processor,
+            job_timeout=20.0,
+            num_workers=device_selection_size // num_leaf_nodes + 1,
```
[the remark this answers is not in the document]
ZiyueXu77: this is correct setting

YuanTingHsieh (approved):

PR #3600, opened 2025-08-13
Dashboard extra prop

yanchengnv:
Fixes # .

### Description

Currently, with the Dashboard, the user can only specify a small list of basic parameters, which may not be enough for projects that require more parameters to be specified (e.g. BYOConn parameters).  Adding these extra parameters to the Dashboard system will take time.

This PR adds a temporary solution by using a property file to specify those required params.  When generating a startup kit, the entity's properties are augmented with the properties in the prop file.

The prop file must be either "properties.yml" or "properties.json" and placed under the WEB ROOT.

This PR also change env var name strings to use defined constants.

### Types of changes

- [ ] Documentation updated.


IsaacYangSLA (approved):
LGTM.  We should be able to expand this in the future to support preset configuration file for dashboard.

PR #3603, opened 2025-08-14
Update device selection logic to align with new simulated device pattern

ZiyueXu77:
Fixes QA issue 5457195.

### Description

We could receive update from a client not in device selection (device removed from the selection pool and never put back under reuse), this is caused by an outdated handling of device id generation scheme we used previously. I realized that with our new device generation scheme (any number in one-shot), we no longer need the "const_selection" flag, and updated the code accordingly

### Types of changes

- [ ] Documentation updated.


nvflare/edge/assessors/async_num.py
```diff
@@ -201,15 +207,8 @@ def _fill_selection(self, fl_ctx: FLContext):
...
-                # remove all used devices from available devices
-                usable_devices = set(self.available_devices.keys()) - set(self.used_devices.keys())
-            else:
-                # remove only the devices that are associated with the current model version
-                usable_devices = set(self.available_devices.keys()) - set(
-                    k for k, v in self.used_devices.items() if v == self.current_model_version
-                )
-
+            # remove all used devices from available devices
+            usable_devices = set(self.available_devices.keys()) - set(self.used_devices.keys())
```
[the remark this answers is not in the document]
ZiyueXu77: This is handled in a different place now

nvflare/edge/assessors/model_update.py
```diff
@@ -128,6 +128,10 @@ def _do_child_update(self, update: Shareable, fl_ctx: FLContext) -> (bool, Optio
             for model_update in report.model_updates.values():
                 if model_update:
                     self.device_manager.remove_devices_from_selection(set(model_update.devices.keys()), fl_ctx)
+                    # if device_reuse, remove devices from used_devices
+                    # indicating that the reported devices becomes available again for reuse
+                    if self.device_manager.device_reuse:
+                        self.device_manager.remove_devices_from_used(set(model_update.devices.keys()), fl_ctx)
```
YuanTingHsieh, line 134: should we just call:

self.device_manager.remove_devices_from_selection?

since remove_devices_from_used is not device_manager API

nvflare/edge/assessors/buff_device_manager.py
```diff
@@ -87,13 +72,17 @@ def fill_selection(self, current_model_version: int, fl_ctx) -> None:
...
         )
 
     def remove_devices_from_selection(self, devices: Set[str], fl_ctx) -> None:
         for device_id in devices:
             self.current_selection.pop(device_id, None)
 
+    def remove_devices_from_used(self, devices: Set[str], fl_ctx) -> None:
+        for device_id in devices:
+            self.used_devices.pop(device_id, None)
+
```
YuanTingHsieh, line 90: we could merge this logic into "remove_devices_from_selection"

```
def remove_devices_from_selection(self, devices: Set[str], fl_ctx) -> None:
        for device_id in devices:
            self.current_selection.pop(device_id, None)
            if self.device_reuse:
                self.used_devices.pop(device_id, None)
```

nvflare/edge/assessors/async_num.py
```diff
@@ -218,7 +217,10 @@ def _fill_selection(self, fl_ctx: FLContext):
                     self.used_devices[device_id] = self.current_model_version
                     if not usable_devices:
                         break
```
YuanTingHsieh, line 219: if not enough usable_devices, and the num_holes is greater, what happens? should that consider an error or should we print out a warning?

nvflare/edge/assessors/model_update.py
ZiyueXu77: this operation is related (but not quite) to selection, in fact it is "opposite, but not entirely": remove from used make the device "useable" again, but it may not be finally selected (randomly), so it may or may not be end up in the selection list.
Better to add this remove_devices_from_used to device_manager API, added

nvflare/edge/assessors/buff_device_manager.py
ZiyueXu77: I think it is better to define a separate one, users may define their own logic to coordinate the updates of the three sets (available / used / selection)

nvflare/edge/assessors/async_num.py
ZiyueXu77: good point, added a warning

YuanTingHsieh (approved):

PR #3604, opened 2025-08-14
Add support of msg traffic control for swarm learning

yanchengnv:
Fixes # .

### Description

With Swarm Learning, training clients submit results back to the aggregation client when they finish training. In most cases, this happens at roughly the same time. This could cause "traffic jam" on the aggregation client that gets multiple messages at the same time, which could cause resource exhaust on the aggregation client, especially when the training result is huge (e.g. 32GB).

To mitigate this problem, this PR adds "traffic control" to limit the number of concurrent result submissions. It works as follows:

- Before submitting result, the training client sends "submission request" to the aggregation client for permission to submit. It does this repeatedly until a permission is received; or exceeds max wait time (if specified).
- Once a "submission request" is received, the aggregation client checks how many concurrent submissions are already in flight. If it's still below the configured limit, then permission is granted; otherwise permission is denied and the training client will have to ask again.

Foure config args are added to the SwarmClientController:

**request_to_submit_result_msg_timeout** - the timeout for "submission request" message. Since submission req is a tiny message, this timeout value should be small. Default to 5.0.

**request_to_submit_result_max_wait** - max amount of time to wait for the permission from the aggregation client. If the permission is not received within this period of time, the training result will not be submitted. If this value is not specified (None), then the training client will keep trying forever.  Default is None. 

**max_concurrent_submissions** - max number of concurrent submissions allowed on the aggregation client. Default is 1.

**request_to_submit_result_interval** - the wait time between result submission requests 

Note that if the max_concurrent_submissions is set to 1, it practically means that all training results will be submitted to the aggregation client sequentially.  This lowers the resource pressure on the aggr client, but makes the overall training process longer. The value of request_to_submit_result_max_wait, if specified, should be long enough to allow the aggr client sufficient time to process training results.

### Types of changes

- [ ] Documentation updated.


nvflare/app_common/ccwf/swarm_client_ctl.py
```diff
@@ -615,6 +713,73 @@ def do_learn_task(self, name: str, task_data: Shareable, fl_ctx: FLContext, abor
...
+                    return
+
+                resp = engine.send_aux_request(
+                    targets=[aggr],
+                    topic=self.request_to_submit_learn_result_task_name,
+                    request=submission_req,
+                    timeout=self.request_to_submit_result_msg_timeout,
+                    fl_ctx=fl_ctx,
+                    secure=False,
+                )
```
YuanTingHsieh, line 781: could this resp be a shareable with ReturnCode.TIMEOUT? since it's send_aux_request

nvflare/app_common/ccwf/swarm_client_ctl.py
```diff
@@ -66,6 +69,7 @@ def __init__(
         self.trainer_statuses = {}
         self.start_time = time.time()
         self.timeout = timeout
+        self.max_concurrent_submissions = max_concurrent_submissions
```
YuanTingHsieh, line 72: suggest adding some docstrings

nvflare/app_common/ccwf/swarm_client_ctl.py
yanchengnv: Yes, it could, and we will print a warning and retry.

nvflare/app_common/ccwf/swarm_client_ctl.py
yanchengnv: added

YuanTingHsieh (approved):
LGTM

PR #3609, opened 2025-08-15
Provisioning for Edge

nvidianz:
### Description

Provisioning for edge and a few minor changes,

1. Added -g and -e option to generate sample project.yml for regular and edge provisioning.
2. Removed interactive prompt when no arg is given to provision to be consistent with all other commands.
3. Removed overseer from project.yml.

Sample edge project.yml,

```
api_version: 3
name: example_project
description: NVFlare sample project yaml file for edge

# For edge project, the participants are created automatically
# and organized in a tree hierarchy.
edge:
  analyze: false
  lcp_only: false
  root_dir: .
  depth: 2
  width: 3
  max_sites: 100
  clients: 2
  rp: 4321

# The participants are ignored for edge provisioning except admins
participants:
  - name: admin@nvidia.com
    type: admin
    org: nvidia
    role: project_admin
```

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/lighter/edge_project.yml
+++ b/nvflare/lighter/edge_project.yml
@@ -1,0 +1,36 @@
+api_version: 3
+name: example_project
+description: NVFlare sample project yaml file for edge
+
+# For edge project, the participants are created automatically
+# and organized in a tree hierarchy.
+edge:
+  analyze: false
+  lcp_only: false
+  root_dir: .
+  depth: 2
+  width: 3
+  max_sites: 100
+  clients: 2
+  rp: 4321
+
+# The participants are ignored for edge provisioning except admins
+participants:
+  - name: admin@nvidia.com
+    type: admin
+    org: nvidia
+    role: project_admin
+
+builders:
+  - path: nvflare.lighter.impl.workspace.WorkspaceBuilder
+    args:
+      template_file:
+        - master_template.yml
+        - aws_template.yml
+        - azure_template.yml
+  - path: nvflare.lighter.impl.static_file.StaticFileBuilder
+    args:
+      # config_folder can be set to inform NVIDIA FLARE where to get configuration
+      config_folder: config
+  - path: nvflare.lighter.impl.cert.CertBuilder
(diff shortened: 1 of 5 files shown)


nvflare/lighter/impl/static_file.py
```diff
@@ -869,7 +869,7 @@ def finalize(self, project: Project, ctx: ProvisionContext):
 
     @staticmethod
     def _append(content: str, participant: Participant) -> str:
-        return content + f"./{participant.name}/startup/start.sh\n"
+        return content + f"$BASE_DIR/{participant.name}/startup/start.sh\n"
```
yanchengnv, line 872: What is BASE_DIR? Where is it defined?

nvflare/lighter/provision.py
```diff
@@ -21,6 +21,8 @@
 import sys
 from typing import Optional
 
+from nvflare.cli_unknown_cmd_exception import CLIUnknownCmdException
+from nvflare.edge.tools.tree_prov import edge_provision
```
yanchengnv: The dependency is not good.
Provision is independent of Flare code. 
Move tree_prov into "lighter".

nvflare/lighter/impl/static_file.py
nvidianz: It's defined in line 885 below

nvflare/lighter/provision.py
nvidianz: Good catch! Moved.

nvflare/lighter/edge_project.yml
```diff
@@ -0,0 +1,36 @@
...
+# and organized in a tree hierarchy.
+edge:
+  analyze: false
+  lcp_only: false
+  root_dir: .
+  depth: 2
+  width: 3
+  max_sites: 100
+  clients: 2
+  rp: 4321
```
YuanTingHsieh, line 15: might need to add comments/docstrings to explain what each field mean

PR #3616, opened 2025-08-19
Convert ET model once for each task

yanchengnv:
Fixes # .

### Description

The PT model needs to be converted to ET format before sending to a device. There are two issues with current implementation:

- The model is converted every time a new request is received from the device, even though the conversion is the same for all devices.
- The conversion is not done within a lock. When multiple devices ask for the model at the same time, the multiple conversions happen concurrently, which could break the integrity of the converted result.

This PR fixes these issues:
- The model is converted only once. Converted result is cached for later reuse.
- The check-and-convert process is placed within a lock. 

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh (approved):
LGTM

PR #3619, opened 2025-08-20
Make global eval widget async

ZiyueXu77:
Fixes QA finding 5454648 .

### Description

Previous global eval widget is sync, so it will block server from responding to client, and thus QA need to adjust the timeout to make it work. But since global eval only tracks training progress without interfering fl process, it should be made async.

Also updated readme to migrate from tree_prov to provision

### Types of changes

- [ ] Documentation updated.


nvflare/edge/widgets/evaluator.py
```diff
@@ -160,16 +224,102 @@ def _initialize(self, _event_type: str, fl_ctx: FLContext):
...
+                    break
+
+            # Wait a bit before checking again
+            time.sleep(0.1)
+
+        # Start evaluation in a separate thread
+        eval_thread = threading.Thread(
+            target=self._evaluate_async,
+            args=(global_weights, current_round, evaluation_id),
+            daemon=False,
```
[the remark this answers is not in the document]
ZiyueXu77: expected

examples/advanced/edge/setup_nvflare.sh
```diff
@@ -1,11 +1,26 @@
...
+if [ -f "$WORKSPACE_DIR/project.yml" ]; then
+    echo "Updating depth and width in project.yml..."
+    sed -i 's/depth: 2/depth: 1/' "$WORKSPACE_DIR/project.yml"
+    sed -i 's/width: 3/width: 2/' "$WORKSPACE_DIR/project.yml"
+    echo "Updated project.yml with depth: 1, width: 2"
+    sed -i 's/name: example_project/name: edge_example/' "$WORKSPACE_DIR/project.yml"
+    echo "Updated project.yml with name: edge_example"
+else
+    echo "Warning: project.yml not found at $WORKSPACE_DIR/$PROJECT_NAME/project.yml"
+fi
```
YuanTingHsieh: instead of calling provision twice, maybe we just check in a "project.yml" here for this example?
then we can ask user to just run "nvflare provision -p project.yml"

examples/advanced/edge/README.md
```diff
@@ -24,21 +24,21 @@ To run the ExecuTorch simulated devices, you need to install the executorch pybi
 
 ### Provision the NVFlare System
 
-We are using `nvflare/edge/tools/tree_prov.py` to provision a hierarchical NVFlare system:
+We are using `nvflare provision -e` to provision a hierarchical NVFlare system for edge:
```
YuanTingHsieh: see comments in setup_nvflare.sh, maybe we don't need that file anymore

nvflare/edge/widgets/evaluator.py
```diff
@@ -111,6 +127,9 @@ def _create_data_loader(self):
             self.data_loader = torch.utils.data.DataLoader(
                 test_set, batch_size=self.batch_size, shuffle=False, num_workers=2
             )
+            self.logger.info(
```
YuanTingHsieh, line 127: not related to this PR's changes, but a question regarding torchvision dataset.
DO they ALL have this interface "dataset_class(root=dataset_path, train=False, download=True, transform=transform)"? or what are the lists of dataset?

nvflare/edge/widgets/evaluator.py
```diff
@@ -160,16 +224,106 @@ def _initialize(self, _event_type: str, fl_ctx: FLContext):
...
+            True if all evaluations completed, False if timeout occurred
+        """
+        if not self._active_evaluations:
+            return True
+
+        start_time = time.time()
+        self.logger.info(f"Waiting for {len(self._active_evaluations)} active evaluations to complete...")
+
+        while self._active_evaluations and (time.time() - start_time) < self.timeout:
+            time.sleep(POLLING_INTERVAL)
```
YuanTingHsieh: there could be race condition when checking "self._action_evaluations",

suggest doing something like:

```
        with self._evaluation_lock:
            active_count = len(self._active_evaluations)
```

examples/advanced/edge/setup_nvflare.sh
ZiyueXu77: sure, let's do that instead

ZiyueXu77: sure, let's do that instead

nvflare/edge/widgets/evaluator.py
ZiyueXu77: good point! the bool "train" is not always there, added some words

examples/advanced/edge/project.yml
```diff
@@ -0,0 +1,36 @@
+api_version: 3
+name: edge_example
+description: NVFlare edge project yaml file for example
+
+# For edge project, the participants are created automatically
+# and organized in a tree hierarchy.
+edge:
+  analyze: false
+  lcp_only: false
+  root_dir: .
```
YuanTingHsieh: we can set this to be: ""/tmp/nvflare/workspaces", then we don't need setup_nvflare.sh?

nvflare/edge/widgets/evaluator.py
```diff
@@ -160,16 +225,103 @@ def _initialize(self, _event_type: str, fl_ctx: FLContext):
...
+
+        if self._active_evaluations:
+            with self._evaluation_lock:
+                active_count = len(self._active_evaluations)
+            self.logger.warning(f"Warning: {active_count} evaluations did not complete within timeout")
+            return False
+        else:
+            self.logger.info("All evaluations completed successfully")
+            return True
```
YuanTingHsieh: suggest something like:

```
def wait_for_evaluations(self) -> bool:
    """Wait for all active evaluations to complete.

    Returns:
        True if all evaluations completed, False if timeout occurred
    """
    with self._evaluation_lock:
        active_count = len(self._active_evaluations)
    if active_count == 0:
        return True

    start_time = time.time()
    self.logger.info(f"Waiting for {active_count} active evaluations to complete...")

    while (time.time() - start_time) < self.timeout:
        time.sleep(POLLING_INTERVAL)
        with self._evaluation_lock:
            active_ids = list(self._active_evaluations)
            active_count = len(self._active_evaluations)
        if active_count == 0:
            self.logger.info("All evaluations completed successfully")
            return True

    self.logger.warning(f"Evaluations {active_ids} did not complete within timeout {self.timeout}")
    return False

```

examples/advanced/edge/project.yml
ZiyueXu77: ah right, missed that, updated

nvflare/edge/widgets/evaluator.py
```diff
@@ -160,16 +225,103 @@ def _initialize(self, _event_type: str, fl_ctx: FLContext):
...
-        # Evaluate the model according to the evaluation frequency
-        if current_round % self.eval_frequency == 0:
-            metrics = self._eval_model()
-            for key, value in metrics.items():
-                self.tb_writer.add_scalar(key, value, current_round)
+
+        # Create unique evaluation ID
+        evaluation_id = f"eval_round_{current_round}"
+
+        # Wait for available slot if at capacity
```
yanchengnv: All the code below this line is unnecessarily complex. You could use ThreadPoolExecutor to manage easily.
First, you create a ThreadPoolExecutor in the "__init__" method:

from concurrent.futures import ThreadPoolExecutor
 
self.pool = ThreadPoolExecutor(max_workers=1)

Then, after this line, simply call:
self.pool.submit(self._evaluate_async, global_weights, current_round, evaluation_id)

Finally, in the "shutdown" method, call:
self.pool.shutdown(wait=False)

No need to manage your own thread and max_workers.

yanchengnv (dismissed):
See my comments about using ThreadPoolExecutor to manage async eval.

nvflare/edge/widgets/evaluator.py
ZiyueXu77: yes exactly, realized that and working on it, will get it done soon

ZiyueXu77: updated

nvflare/edge/widgets/evaluator.py
```diff
@@ -65,9 +69,13 @@ def __init__(
         self.model = None
         self.data_loader = None
         self.tb_writer = None
+        self.max_workers = max_workers
```
yanchengnv: Should validate to make sure max_workers is a positive int.
Similarly eval_frequency should be validated too.
You can use nvflare.fuel.utils.validation_utils to do these. It has many useful funcs like check_positive_int, check_positive_number, and many others.

nvflare/edge/widgets/evaluator.py
```diff
@@ -65,9 +69,13 @@ def __init__(
...
+
+        self._thread_pool = None
```
yanchengnv: You can create the thread_pool here - no need to do it later since it doesn't depend on other things.

nvflare/edge/widgets/evaluator.py
```diff
@@ -160,16 +213,51 @@ def _initialize(self, _event_type: str, fl_ctx: FLContext):
         app_root = fl_ctx.get_prop(FLContextKey.APP_ROOT)
         self.tb_writer = SummaryWriter(log_dir=app_root)
 
+        # Initialize thread pool
+        self._thread_pool = ThreadPoolExecutor(max_workers=self.max_workers, thread_name_prefix="GlobalEvaluator")
```
yanchengnv: Can be done in the constructor.

yanchengnv (commented):
Looks good in general. See comments for minor improvement.

nvflare/edge/widgets/evaluator.py
ZiyueXu77: added

nvflare/edge/widgets/evaluator.py
ZiyueXu77: moved here

ZiyueXu77:
further updated baselines and readme

examples/advanced/edge/figs/cifar10_acc.png
YuanTingHsieh, line 1: just curious why the line changed

nvflare/edge/widgets/evaluator.py
```diff
@@ -141,14 +165,49 @@ def _eval_model(self) -> Dict[str, float]:
...
+            # Safety check for model
+            if self.model is None:
+                self.logger.warning(f"Model not available for evaluation round {current_round}")
+                return
+
+            self.logger.info(f"Starting evaluation {evaluation_id} for round {current_round}")
+
+            # Load the model weights
+            global_weights_tensors = {k: torch.tensor(v) for k, v in global_weights.items()}
+            self.model.load_state_dict(global_weights_tensors)
```
YuanTingHsieh, line 181: if workers > 1, could there be a situation where multiple workers are loading different models into self.model?

nvflare/edge/widgets/evaluator.py
```diff
@@ -141,14 +165,49 @@ def _eval_model(self) -> Dict[str, float]:
...
+
+            # Evaluate the model
+            metrics = self._eval_model()
```
YuanTingHsieh, line 184: if more threads, would reusing the same self.dataloader has racing issue?

nvflare/edge/widgets/evaluator.py
```diff
@@ -141,14 +165,49 @@ def _eval_model(self) -> Dict[str, float]:
...
+
+            # Write metrics to tensorboard, starting from 0
+            if self.tb_writer:
+                for key, value in metrics.items():
+                    self.tb_writer.add_scalar(key, value, current_round - 1)
+
+            self.logger.info(f"Evaluation {evaluation_id} for round {current_round} completed with metrics: {metrics}")
```
YuanTingHsieh, line 191: could there be multiple worker threads try to log to same key? would that cause any trouble?

examples/advanced/edge/README.md
```diff
@@ -24,21 +24,23 @@ To run the ExecuTorch simulated devices, you need to install the executorch pybi
 
 ### Provision the NVFlare System
 
-We are using `nvflare/edge/tools/tree_prov.py` to provision a hierarchical NVFlare system:
+We are using `nvflare provision` to provision a hierarchical NVFlare system for edge:
```
nvidianz: Can you add 

nvflare provision -e

and explain the content of project.yml?

nvflare/edge/widgets/evaluator.py
ZiyueXu77: Good point! Let's keep max_workers=1 to make it sequential

examples/advanced/edge/figs/cifar10_acc.png
ZiyueXu77: I updated the centralized validation, previously it was after a training, so actually it starts in "version 1", now everything starts with the original model version 0

nvflare/edge/widgets/evaluator.py
ZiyueXu77: make max_worker=1 to avoid racing

nvflare/edge/widgets/evaluator.py
ZiyueXu77: make max_worker=1 to avoid

examples/advanced/edge/README.md
ZiyueXu77: added a note

YuanTingHsieh (approved):
LGTM

PR #3626, opened 2025-08-21
Fix poc shutdown

YuanTingHsieh:
Fix poc shutdown

### Description

Fix poc shutdown

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/examples/advanced/edge/README.md
+++ b/examples/advanced/edge/README.md
@@ -73,7 +73,15 @@ To start the system, run the following command:
 ```commandline
 cd /tmp/nvflare/workspaces/edge_example/prod_00/
 ./start_all.sh
-```    
+```
+
+To run with real device, we also need to start the proxy:
+```commandline
+cd /tmp/nvflare/workspaces/edge_example/prod_00/scripts/
+./start_rp.sh
+```
+
+By default, it will start listening on port 4321, feel free to adjust that.
 
 ## ExecuTorch-based FL
 ### ExecuTorch simulated devices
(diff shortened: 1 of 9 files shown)


nvflare/tool/poc/poc_commands.py
```diff
@@ -689,20 +689,20 @@ def _stop_poc(poc_workspace: str, excluded=None, services_list=None):
 
     p_size = len(services_list)
     if p_size == 0 or service_config[SC.FLARE_SERVER] in services_list:
-        print("start shutdown NVFLARE")
+        print("start shutdown NVFLARE using FLARE API")
         shutdown_system(prod_dir, username=service_config[SC.FLARE_PROJ_ADMIN])
```
yanchengnv, line 693: This is defined the api_utils. The problem is that it calls wait_for_system_shutdown after issuing shutdown command. The wait_for_system_shutdown function keeps sending "check status" command to the server, which could cause gRPC errors to be printed, as reported by QA. We didn't have this problem before consolidating admin port.

The simplest change is to remove the call of wait_for_system_shutdown from shutdown_system.

YuanTingHsieh: thanks, changed as suggested.

nvflare/tool/poc/poc_commands.py
```diff
@@ -689,20 +689,20 @@ def _stop_poc(poc_workspace: str, excluded=None, services_list=None):
 
     p_size = len(services_list)
     if p_size == 0 or service_config[SC.FLARE_SERVER] in services_list:
-        print("start shutdown NVFLARE")
+        print("start shutdown NVFLARE using FLARE API")
```
chesterxgchen: why doing add "using FLARE API", this can be triggered from CLI or API. This is not necessary

YuanTingHsieh: "shutdown_system" that method internally is using FLARE API, but that's implementation detail so remove from the log msg to not confuse users.

the below error msg added "stop_fl.sh" I think that is kind of important to show the users as that does not always guarantee a clean shutdown.

PR #3632, opened 2025-08-25
Add wait for device info and timeout

ZiyueXu77:
Fixes QA 5477637.

### Description

QA's question on waiting for sufficient device, I think it makes sense to add status printout to inform users and a timeout to limit the waiting time

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/edge/assessors/model_update.py
+++ b/nvflare/edge/assessors/model_update.py
@@ -33,6 +33,7 @@ class ModelUpdateAssessor(Assessor):
         model_manager_id,
         device_manager_id,
         max_model_version,
+        device_wait_timeout: float = 30.0,
     ):
         """Initialize the ModelUpdateAssessor.
         Enable both asynchronous and synchronous model updates from clients.
@@ -44,6 +45,7 @@ class ModelUpdateAssessor(Assessor):
             model_manager_id (str): ID of the model manager component.
             device_manager_id (str): ID of the device manager component.
             max_model_version (int): Maximum model version to stop the workflow.
+            device_wait_timeout (float): Timeout in seconds for waiting for sufficient devices. Default is 30 seconds.
         """
         Assessor.__init__(self)
         self.persistor_id = persistor_id
@@ -53,9 +55,40 @@ class ModelUpdateAssessor(Assessor):
         self.model_manager = None
         self.device_manager = None
         self.max_model_version = max_model_version
+        self.device_wait_timeout = device_wait_timeout
         self.update_lock = threading.Lock()
         self.start_time = None
+        self.device_wait_start_time = None
+        self.should_stop_job = False
         self.register_event_handler(EventType.START_RUN, self._handle_start_run)
+
+    def _check_device_timeout(self, fl_ctx: FLContext) -> bool:
+        """Check if device wait timeout has been exceeded.
+
+        Args:
+            fl_ctx: FL context
+
+        Returns:
+            bool: True if timeout exceeded, False otherwise
+        """
+        if self.device_wait_start_time is None:
+            return False
+
+        if time.time() - self.device_wait_start_time > self.device_wait_timeout:
+            return True
+        return False
+
+    def _get_remaining_timeout(self) -> float:
+        """Get remaining time before device wait timeout.
+
+        Returns:
+            float: Remaining time in seconds, or 0 if no timeout active
+        """
+        if self.device_wait_start_time is None:
+            return 0.0
+
+        remaining = self.device_wait_timeout - (time.time() - self.device_wait_start_time)
+        return max(0.0, remaining)
 
     def _handle_start_run(self, event_type: str, fl_ctx: FLContext):
         engine = fl_ctx.get_engine()
@@ -116,6 +149,43 @@ class ModelUpdateAssessor(Assessor):
         # Update available devices
         if report.available_devices:
             self.device_manager.update_available_devices(report.available_devices, fl_ctx)
+            # Reset wait timer if we now have enough devices
+            if self.device_wait_start_time is not None and self.device_manager.has_enough_devices(fl_ctx):
+                self.device_wait_start_time = None
+                self.should_stop_job = False  # Reset stop job flag
+                self.log_info(fl_ctx, "Sufficient devices now available, resetting wait timer and stop job flag")
+
+        # check if we have enough devices
+        if not self.device_manager.has_enough_devices(fl_ctx):
+            # Start timing if we haven't started yet
+            if self.device_wait_start_time is None:
+                self.device_wait_start_time = time.time()
+
+            # Check if we've exceeded the timeout
+            if self._check_device_timeout(fl_ctx):
+                self.log_error(
+                    fl_ctx,
+                    f"Device wait timeout ({self.device_wait_timeout}s) exceeded. "
+                    f"Available devices: {len(self.device_manager.available_devices)} out of {self.device_manager.device_selection_size}. "
+                    "Not enough devices joining, please adjust the server params. Stopping the job.",
+                )
+                # Stop the job due to insufficient devices
+                self.should_stop_job = True
+                # Prepare an empty reply
+                reply = StateUpdateReply(
+                    model_version=0,
+                    model=None,
+                    device_selection_version=self.device_manager.current_selection_version,
+                    device_selection=self.device_manager.get_selection(fl_ctx),
+                )
+                return False, reply.to_shareable()
+            else:
+                remaining_time = self._get_remaining_timeout()
+                self.log_info(
+                    fl_ctx,
+                    f"Not enough available devices, waiting for more: {len(self.device_manager.available_devices)} out of {self.device_manager.device_selection_size}. "
+                    f"Timeout in {remaining_time:.1f} seconds",
+                )
 
         accepted = True
         if report.model_updates:
@@ -148,6 +218,8 @@ class ModelUpdateAssessor(Assessor):
                 )
                 self.model_manager.generate_new_model(fl_ctx)
                 self.device_manager.fill_selection(self.model_manager.current_model_version, fl_ctx)
+                # Reset wait timer since we have enough devices
+                self.device_wait_start_time = None
 
         # Prepare reply
         model = None
@@ -163,7 +235,10 @@ class ModelUpdateAssessor(Assessor):
         return accepted, reply.to_shareable()
 
     def assess(self, fl_ctx: FLContext) -> Assessment:
-        if self.model_manager.current_model_version >= self.max_model_version:
+        if self.should_stop_job:
+            self.log_error(fl_ctx, "Job stopped due to insufficient devices joining within timeout period")
+            return Assessment.WORKFLOW_DONE
+        elif self.model_manager.current_model_version >= self.max_model_version:
             model_version = self.model_manager.current_model_version
             self.log_info(fl_ctx, f"Max model version {self.max_model_version} reached: {model_version=}")
             return Assessment.WORKFLOW_DONE
--- a/nvflare/edge/assessors/model_update.py
+++ b/nvflare/edge/assessors/model_update.py
@@ -143,6 +143,41 @@ class ModelUpdateAssessor(Assessor):
         with self.update_lock:
             return self._do_child_update(update, fl_ctx)
 
+    def _device_wait_countdown(self, fl_ctx: FLContext):
+        # Start timing if we haven't started yet
+        if self.device_wait_start_time is None:
+            self.device_wait_start_time = time.time()
+
+        # Check if we've exceeded the timeout
+        if self._check_device_timeout(fl_ctx):
+            usable_devices = set(self.device_manager.available_devices.keys()) - set(
+                self.device_manager.used_devices.keys()
+            )
+            self.log_error(
+                fl_ctx,
+                f"Device wait timeout ({self.device_wait_timeout}s) exceeded. "
+                f"Total devices: {len(self.device_manager.available_devices)}, usable: {len(usable_devices)}, expected: {self.device_manager.device_selection_size}. "
+                f"Device_reuse flag is set to: {self.device_manager.device_reuse}. "
+                "Not enough devices joining, please adjust the server params. Stopping the job.",
+            )
+            # Stop the job due to insufficient devices
+            self.should_stop_job = True
+            # Prepare an empty reply
+            reply = StateUpdateReply(
+                model_version=0,
+                model=None,
+                device_selection_version=self.device_manager.current_selection_version,
+                device_selection=self.device_manager.get_selection(fl_ctx),
+            )
+            return False, reply.to_shareable()
+        else:
+            remaining_time = self._get_remaining_timeout()
+
+            self.log_info(
+                fl_ctx,
+                f"Not enough available devices, waiting for more: " f"Timeout in {remaining_time:.1f} seconds",
+            )
+
     def _do_child_update(self, update: Shareable, fl_ctx: FLContext) -> (bool, Optional[Shareable]):
         report = StateUpdateReport.from_shareable(update)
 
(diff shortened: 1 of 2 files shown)


nvflare/edge/assessors/model_update.py
```diff
@@ -110,12 +143,52 @@ def process_child_update(self, update: Shareable, fl_ctx: FLContext) -> (bool, O
...
             self.device_manager.update_available_devices(report.available_devices, fl_ctx)
+            # Reset wait timer if we now have enough devices
+            if self.device_wait_start_time is not None and self.device_manager.has_enough_devices(fl_ctx):
+                self.device_wait_start_time = None
+                self.should_stop_job = False  # Reset stop job flag
+                self.log_info(fl_ctx, "Sufficient devices now available, resetting wait timer and stop job flag")
+
+        # Check for device wait timeout if we are waiting for devices
+        if self.device_wait_start_time is not None:
+            self._device_wait_countdown(fl_ctx)
```
YuanTingHsieh: this "wait_countdown" is only triggered when "_do_child_update" is called, if somehow "_do_child_update" is not called let's say 100 seconds, the device_wait_timeout of 30 seconds is not working as expected.

We might need another thread to handle this device wait timeout.

Make sure to ensure thread safety if state changed: for example: self.should_stop_job

nvflare/edge/assessors/model_update.py
```diff
@@ -110,12 +143,52 @@ def process_child_update(self, update: Shareable, fl_ctx: FLContext) -> (bool, O
...
+            # Stop the job due to insufficient devices
+            self.should_stop_job = True
+            # Prepare an empty reply
+            reply = StateUpdateReply(
+                model_version=0,
+                model=None,
+                device_selection_version=self.device_manager.current_selection_version,
+                device_selection=self.device_manager.get_selection(fl_ctx),
+            )
+            return False, reply.to_shareable()
```
YuanTingHsieh, line 214: the return value is not used for this method

nvflare/edge/assessors/model_update.py
```diff
@@ -110,12 +143,52 @@ def process_child_update(self, update: Shareable, fl_ctx: FLContext) -> (bool, O
...
+        else:
+            remaining_time = self._get_remaining_timeout()
+
+            self.log_info(
+                fl_ctx,
+                f"Not enough available devices, waiting for more: " f"Timeout in {remaining_time:.1f} seconds",
+            )
+
     def _do_child_update(self, update: Shareable, fl_ctx: FLContext) -> (bool, Optional[Shareable]):
```
YuanTingHsieh: for threaded changes, do something like:

```
# Check if we need to wait
if self.device_manager.should_fill_selection(fl_ctx):
    if self.device_manager.has_enough_devices(fl_ctx):
        if self.model_manager.current_model_version == 0:
            self.log_info(fl_ctx, "Generate initial model and fill selection")
            self.model_manager.generate_new_model(fl_ctx)
        self.device_manager.fill_selection(self.model_manager.current_model_version, fl_ctx)

        with self.state_lock:
            self.device_wait_start_time = None
    else:
        with self.state_lock:
            if self.device_wait_start_time is None:
                self.device_wait_start_time = time.time()
```

nvflare/edge/assessors/model_update.py
```diff
@@ -53,10 +55,41 @@ def __init__(
         self.model_manager = None
         self.device_manager = None
         self.max_model_version = max_model_version
+        self.device_wait_timeout = device_wait_timeout
         self.update_lock = threading.Lock()
         self.start_time = None
+        self.device_wait_start_time = None
+        self.should_stop_job = False
         self.register_event_handler(EventType.START_RUN, self._handle_start_run)
```
YuanTingHsieh, line 65: for threaded, do sometihng like:
```
    self.state_lock = threading.Lock()
    self.stop_event = threading.Event()
    self.device_wait_monitor_thread = None
    ```

File: nvflare/edge/assessors/model_update.py [hunk 5]
```diff
@@ -53,10 +55,41 @@ def __init__(
...
+        Returns:
+            float: Remaining time in seconds, or 0 if no timeout active
+        """
+        if self.device_wait_start_time is None:
+            return 0.0
+
+        remaining = self.device_wait_timeout - (time.time() - self.device_wait_start_time)
+        return max(0.0, remaining)
+
     def _handle_start_run(self, event_type: str, fl_ctx: FLContext):
```
YuanTingHsieh, line 133: can do something like:

```
self.device_wait_monitor_thread = threading.Thread(
    target=self._device_wait_monitor_loop,
    args=(fl_ctx.clone(),),  # Use a cloned FLContext
    daemon=True
)
self.device_wait_monitor_thread.start()

File: nvflare/edge/assessors/model_update.py [hunk 6]
```diff
@@ -53,10 +55,41 @@ def __init__(
...
+        Returns:
+            bool: True if timeout exceeded, False otherwise
+        """
+        if self.device_wait_start_time is None:
+            return False
+
+        if time.time() - self.device_wait_start_time > self.device_wait_timeout:
+            return True
+        return False
+
```
YuanTingHsieh, line 106: for threaded can do something like:

```
def _device_wait_monitor_loop(self, fl_ctx: FLContext):
    while not self.stop_event.is_set():
        with self.state_lock:
            if self.device_wait_start_time is not None:
                elapsed = time.time() - self.device_wait_start_time
                if elapsed >= self.device_wait_timeout:
                    usable_devices = set(self.device_manager.available_devices) - set(
                        self.device_manager.used_devices
                    )
                    self.log_error(
                        fl_ctx,
                        f"[Monitor] Device wait timeout after {elapsed:.1f}s. "
                        f"Available: {len(self.device_manager.available_devices)}, "
                        f"Usable: {len(usable_devices)}, "
                        f"Expected: {self.device_manager.device_selection_size}. "
                        f"Stopping job."
                    )
                    self.should_stop_job = True
                    self.device_wait_start_time = None
        time.sleep(1)

**Comment by ZiyueXu77** [comment]
> suggest changing to use a thread to monitor timeout

good point! will add an independent thread to trace the countdown

File: nvflare/edge/assessors/model_update.py (hunk 4 above)
Comment (ZiyueXu77) [reply nvflare/edge/assessors/model_update.py:65]: use threadpool for easier handling

File: nvflare/edge/assessors/model_update.py [hunk 7]
```diff
@@ -116,6 +240,38 @@ def _do_child_update(self, update: Shareable, fl_ctx: FLContext) -> (bool, Optio
         # Update available devices
         if report.available_devices:
             self.device_manager.update_available_devices(report.available_devices, fl_ctx)
+            # Reset wait timer if we now have enough devices
+            if self.device_wait_start_time is not None and self.device_manager.has_enough_devices(fl_ctx):
+                self.device_wait_start_time = None
+                self.should_stop_job = False  # Reset stop job flag
+                # Stop timeout tracking since we have enough devices
+                self._stop_timeout_tracking()
```
YuanTingHsieh: Suggest using a single long-lived thread with a wait flag, instead of repeatedly starting and stopping a thread.

nvflare/edge/assessors/model_update.py
```diff
@@ -116,6 +240,38 @@ def _do_child_update(self, update: Shareable, fl_ctx: FLContext) -> (bool, Optio
         # Update available devices
         if report.available_devices:
             self.device_manager.update_available_devices(report.available_devices, fl_ctx)
+            # Reset wait timer if we now have enough devices
+            if self.device_wait_start_time is not None and self.device_manager.has_enough_devices(fl_ctx):
+                self.device_wait_start_time = None
```
YuanTingHsieh, line 245: this is modified by the main thread

nvflare/edge/assessors/model_update.py
```diff
@@ -53,10 +56,132 @@ def __init__(
...
+                        f"Device wait timeout ({self.device_wait_timeout}s) exceeded. "
+                        f"Setting stop job flag to terminate workflow.",
+                    )
+
+        except Exception as e:
+            # Log error but don't crash the thread
+            self.log_error(fl_ctx, f"Error in timeout tracker: {e}")
+        finally:
+            # Clean up
+            self.timeout_future = None
```
YuanTingHsieh: this is changed both in main thread and this worker thread.

nvflare/edge/assessors/model_update.py
```diff
@@ -53,10 +56,132 @@ def __init__(
...
+                    self._last_status_log_time = current_time
+
+    def _start_timeout_tracking(self, fl_ctx: FLContext):
+        """Start independent timeout tracking using thread pool."""
+        if self.timeout_future is not None and not self.timeout_future.done():
+            # Cancel existing timeout if still running
+            self.timeout_future.cancel()
+
+        # Submit new timeout task to thread pool
+        self.timeout_future = self.thread_pool.submit(self._timeout_tracker, fl_ctx)
```
YuanTingHsieh: this is changed both in main thread and this worker thread.

nvflare/edge/assessors/model_update.py
ZiyueXu77: makes sense, updated

nvflare/edge/assessors/model_update.py
ZiyueXu77: updated

nvflare/edge/assessors/model_update.py
ZiyueXu77: updated

nvflare/edge/assessors/model_update.py
```diff
@@ -163,9 +347,25 @@ def _do_child_update(self, update: Shareable, fl_ctx: FLContext) -> (bool, Optio
...
+        elif self.model_manager.current_model_version >= self.max_model_version:
+            # Stop timeout tracking before ending the job
+            self._stop_timeout_tracking()
             model_version = self.model_manager.current_model_version
             self.log_info(fl_ctx, f"Max model version {self.max_model_version} reached: {model_version=}")
             return Assessment.WORKFLOW_DONE
         else:
             return Assessment.CONTINUE
+
+    def __del__(self):
```
YuanTingHsieh: putting things in "__del__" we can't guarantee when it will be called, we could use event handling mechanism to shutdown resources.

nvflare/edge/assessors/model_update.py
```diff
@@ -33,6 +34,7 @@ def __init__(
         model_manager_id,
         device_manager_id,
         max_model_version,
+        device_wait_timeout: float = 30.0,
```
YuanTingHsieh: can we use a default wait timeout of None that means wait forever?

nvflare/edge/assessors/model_update.py
```diff
@@ -116,6 +262,38 @@ def _do_child_update(self, update: Shareable, fl_ctx: FLContext) -> (bool, Optio
         # Update available devices
         if report.available_devices:
             self.device_manager.update_available_devices(report.available_devices, fl_ctx)
+            # Reset wait timer if we now have enough devices
+            if self.device_wait_start_time is not None and self.device_manager.has_enough_devices(fl_ctx):
+                self.device_wait_start_time = None
+                self.should_stop_job = False  # Reset stop job flag
+                # Stop timeout tracking since we have enough devices
+                self._stop_timeout_tracking()
```
YuanTingHsieh: do we need to call this here? maybe we just leave the clearing of event to `_timeout_tracker`?

nvflare/edge/assessors/model_update.py
```diff
@@ -133,21 +311,27 @@ def _do_child_update(self, update: Shareable, fl_ctx: FLContext) -> (bool, Optio
...
-                )
-                self.model_manager.generate_new_model(fl_ctx)
+                if self.model_manager.current_model_version == 0:
+                    self.log_info(fl_ctx, "Generate initial model and fill selection")
+                    self.model_manager.generate_new_model(fl_ctx)
                 self.device_manager.fill_selection(self.model_manager.current_model_version, fl_ctx)
+                # Reset wait timer since we have enough devices
+                self.device_wait_start_time = None
+                # Stop timeout tracking since we have enough devices
+                self._stop_timeout_tracking()
```
YuanTingHsieh: do we need to call this here? maybe we just leave the clearing of event to `_timeout_tracker`?

nvflare/edge/assessors/model_update.py
```diff
@@ -163,9 +347,25 @@ def _do_child_update(self, update: Shareable, fl_ctx: FLContext) -> (bool, Optio
         return accepted, reply.to_shareable()
 
     def assess(self, fl_ctx: FLContext) -> Assessment:
-        if self.model_manager.current_model_version >= self.max_model_version:
+        if self.should_stop_job:
+            # Stop timeout tracking before ending the job
+            self._stop_timeout_tracking()
```
YuanTingHsieh: do we need to call this here? maybe we just leave the clearing of event to `_timeout_tracker`?

nvflare/edge/assessors/model_update.py
```diff
@@ -163,9 +347,25 @@ def _do_child_update(self, update: Shareable, fl_ctx: FLContext) -> (bool, Optio
...
+            self.log_error(fl_ctx, "Job stopped due to insufficient devices joining within timeout period")
+            return Assessment.WORKFLOW_DONE
+        elif self.model_manager.current_model_version >= self.max_model_version:
+            # Stop timeout tracking before ending the job
+            self._stop_timeout_tracking()
```
YuanTingHsieh: do we need to call this here? maybe we just leave the clearing of event to `_timeout_tracker`?

nvflare/edge/assessors/model_update.py
```diff
@@ -53,10 +56,154 @@ def __init__(
...
+                        self.device_wait_start_time is None
+                        or self.should_stop_job
+                        or not self.timeout_wait_event.is_set()
+                    ):
+                        break
+
+                    # Log periodic status update
+                    if elapsed < self.device_wait_timeout:
+                        remaining = self.device_wait_timeout - elapsed
+                        self.log_info(fl_ctx, f"Device wait countdown: {remaining:.1f} seconds remaining")
```
YuanTingHsieh: instead of doing nested while loop, maybe we have time.sleep(1.0) for the outer while loop, and then we just use "if" statements here?

nvflare/edge/assessors/model_update.py
```diff
@@ -133,21 +311,27 @@ def _do_child_update(self, update: Shareable, fl_ctx: FLContext) -> (bool, Optio
...
+                # Reset wait timer since we have enough devices
+                self.device_wait_start_time = None
+                # Stop timeout tracking since we have enough devices
+                self._stop_timeout_tracking()
+            else:
+                # Start or continue wait timer since we don't have enough devices
+                if self.device_wait_start_time is None:
+                    self.device_wait_start_time = time.time()
+                    # Start independent timeout tracking
+                    self._start_timeout_tracking(fl_ctx)
```
YuanTingHsieh: instead of starting multiple times, maybe we should start the monitor thread only once in the START_RUN event.

nvflare/edge/assessors/model_update.py
```diff
@@ -53,10 +56,154 @@ def __init__(
...
+
+    def _log_device_wait_status(self, fl_ctx: FLContext, message_prefix: str = ""):
+        """Log current device wait status with countdown information."""
+        if self.device_wait_start_time is not None:
+            current_time = time.time()
+            elapsed = current_time - self.device_wait_start_time
+
+            # Only log if we haven't logged recently (rate limiting)
+            if not hasattr(self, "_last_status_log_time"):
+                self._last_status_log_time = 0
```
YuanTingHsieh: suggest move this instance variable to the `__init__` method

nvflare/edge/assessors/model_update.py
```diff
@@ -53,10 +56,154 @@ def __init__(
...
+            if remaining_time > 0:
+                # Determine logging interval based on remaining time
+                if remaining_time > 60:  # More than 1 minute: log every 30 seconds
+                    log_interval = 30.0
+                elif remaining_time > 30:  # 30 seconds to 1 minute: log every 15 seconds
+                    log_interval = 15.0
+                elif remaining_time > 10:  # 10 to 30 seconds: log every 5 seconds
+                    log_interval = 5.0
+                else:  # Final 10 seconds: log every 2 seconds
+                    log_interval = 2.0
```
YuanTingHsieh: while adaptive might be good, should we instead let user decide the logging interval? I am neutral here, either way is OK

YuanTingHsieh (approved):

PR #3634, opened 2025-08-25
1) PT FedAvg Recipe: Add missing arguments 2) LR FedAvg Recipe 

chesterxgchen:
### Description
* Update PT FedAvg Recipe for missing arguments
* add LR FedAvg Recipe 

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/app_opt/pt/recipes/fedavg.py
+++ b/nvflare/app_opt/pt/recipes/fedavg.py
@@ -40,10 +40,10 @@ class _FedAvgValidator(BaseModel):
     train_args: str
     aggregator: Optional[Aggregator]
     aggregator_data_kind: Optional[DataKind]
-    launch_external_process: bool = (False,)
-    command: str = ("python3 -u",)
-    server_expected_format: ExchangeFormat = (ExchangeFormat.NUMPY,)
-    params_transfer_type: TransferType = (TransferType.FULL,)
+    launch_external_process: bool = False
+    command: str = "python3 -u"
+    server_expected_format: ExchangeFormat = ExchangeFormat.NUMPY
+    params_transfer_type: TransferType = TransferType.FULL
 
     def model_post_init(self, __context):
         if self.clients and self.num_clients is None:


nvflare/app_opt/pt/job_config/fed_avg.py
```diff
@@ -21,14 +21,14 @@
...
-        mandatory_clients: Optional[List[str]] = None,
-        key_metric: str = "accuracy",
+            self,
+            initial_model: nn.Module,
+            n_clients: int,
+            num_rounds: int,
+            name: str = "fed_job",
+            min_clients: int = 1,
+            mandatory_clients: Optional[List[str]] = None,
+            key_metric: str = "accuracy",
```
YuanTingHsieh: suggest undo changes to this file

nvflare/app_common/np/recipes/lr/fedavg.py
```diff
@@ -0,0 +1,137 @@
...
+            runner = ScriptRunner(
+                script=self.train_script,
+                script_args=self.train_args,
+                launch_external_process=self.launch_external_process,
+                command=self.command,
+                framework=FrameworkType.RAW,
+                server_expected_format=ExchangeFormat.RAW,
+                params_transfer_type=TransferType.FULL
+            )
+            job.to(runner, client)
```
YuanTingHsieh: I think "job.to_clients" should work? if it's not working we can merge this and I can check later as well

nvflare/app_common/np/recipes/lr/fedavg.py
```diff
@@ -0,0 +1,137 @@
...
+
+
+class FedAvgLrRecipe(Recipe):
+    def __init__(
+            self,
+            *,
+            name: str = "lr_fedavg",
+            initial_model: Any = None,
+            clients: Optional[List[str]] = None,
+            num_clients: Optional[int] = None,
```
YuanTingHsieh: might need to add the docstring to explain these a bit, we can just copy from FedAvgRecipe

nvflare/app_common/np/recipes/lr/fedavg.py
chesterxgchen: no it did not work, I tried in prev. PR

nvflare/app_opt/pt/job_config/fed_avg.py
chesterxgchen: reverted

nvflare/app_common/np/recipes/lr/fedavg.py
```diff
@@ -0,0 +1,137 @@
...
+        for client in clients:
+            job.to(id="event_to_fed", obj=convert_to_fed_event, target=client)
+            runner = ScriptRunner(
+                script=self.train_script,
+                script_args=self.train_args,
+                launch_external_process=self.launch_external_process,
+                command=self.command,
+                framework=FrameworkType.RAW,
+                server_expected_format=ExchangeFormat.RAW,
+                params_transfer_type=TransferType.FULL,
```
YuanTingHsieh, line 174: these could be configurable as well, but we can leave the common refactoring in a later PR.

chesterxgchen: For LR, I think I don't think they will change. the model weight is almost always full, the data format is what RAW or NUMPY ( RAW and numpy is the same here). and Framework is RAW,.

nvflare/app_common/np/recipes/lr/fedavg.py
chesterxgchen: added

nvflare/app_opt/pt/recipes/fedavg.py
```diff
@@ -39,6 +40,10 @@ class _FedAvgValidator(BaseModel):
     train_args: str
     aggregator: Optional[Aggregator]
     aggregator_data_kind: Optional[DataKind]
+    launch_external_process: bool = (False,)
+    command: str = ("python3 -u",)
+    server_expected_format: ExchangeFormat = (ExchangeFormat.NUMPY,)
+    params_transfer_type: TransferType = (TransferType.FULL,)
```
YuanTingHsieh: this is interesting, does pydantic want tuples not the original type objects?

chesterxgchen: good catch, I think I copied from the generated code, which somehow decided to use Tuple.

PR #3642, opened 2025-08-27
Add auto version cleaning for aggr and global model version keeping

ZiyueXu77:
### Description

Currently we control the number of model/aggregator versions on server and on clients only with "max_model_history", as model versions scale, this can be both:
- limiting as it can be hard to set the history to a proper value
- wasting resource as some later versions can become obsolete before the earlier versions due to device latency variations

Therefore, updated the behavior of device selection dict to keep record of model versions to reflect the current active model versions, only those active will be kept, and others will be considered obsolete and removed since no device is working on them anymore. This saves resource while keep all updates' contribution without making hard-cut removal. The max_model_history can still be set, with the default updated to inf (apply all updates to global model).  

### Types of changes

- [ ] Documentation updated.


nvflare/edge/tools/edge_job.py
```diff
@@ -114,8 +114,15 @@ def configure_client(
         if self.client_config_added:
             raise RuntimeError("client config is already added")
 
+        # check the validity of max_model_versions and convert to int if it is a positive integer
+        if max_model_versions <= 0:
+            raise ValueError("max_model_versions must be a positive integer or float('inf')")
+        elif max_model_versions == float("inf"):
+            self.max_model_versions = float("inf")
+        else:
+            self.max_model_versions = int(max_model_versions)
```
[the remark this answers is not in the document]
ZiyueXu77: good catch

nvflare/edge/assessors/buff_model_manager.py
```diff
@@ -154,19 +173,16 @@ def process_updates(self, model_updates: Dict[int, ModelUpdate], fl_ctx: FLConte
...
+            if self.current_model_version - model_version >= self.max_model_history:
+                self.log_warning(
                     fl_ctx,
-                    f"dropped child update version {model_version}. Current version {self.current_model_version}",
+                    f"dropped child update version {model_version}. Current version {self.current_model_version}. Max history {self.max_model_history}",
                 )
                 continue
 
+            # Accept the update and aggregate it to the corresponding model version
             model_state = self.updates.get(model_version)
```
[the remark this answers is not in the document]
ZiyueXu77: added

nvflare/edge/assessors/buff_device_manager.py
```diff
@@ -66,8 +69,12 @@ def fill_selection(self, current_model_version: int, fl_ctx) -> None:
                 for _ in range(num_holes):
                     device_id = random.choice(list(usable_devices))
                     usable_devices.remove(device_id)
-                    self.current_selection[device_id] = self.current_selection_version
-                    self.used_devices[device_id] = current_model_version
+                    # current_selection keeps track of devices selected for a particular model version
+                    self.current_selection[device_id] = current_model_version
```
yanchengnv, line 73: Using current_model_version as the selection version will essentially make it impossible for the same device to train more than once on the same model version. I guess this is okay if you never want the same device to train multiple times on the same model version.

examples/advanced/edge/jobs/pt_job.py
```diff
@@ -92,16 +90,6 @@ def create_edge_recipe(fl_mode, devices_per_leaf, num_leaf_nodes, global_rounds,
             # sync - each global model covers total_devices data
             # async - each global model covers 1 device's data
             max_model_version=global_rounds * total_devices,
```
yanchengnv, line 92: This could be a huge number. Then does it make sense to have max_model_version any more here?

nvflare/edge/assessors/buff_model_manager.py
```diff
@@ -43,16 +43,22 @@ class BuffModelManager(ModelManager):
     def __init__(
         self,
         num_updates_for_model: int,
-        max_model_history: int,
+        max_model_history: float = float("inf"),  # float to support inf, default is inf
```
yanchengnv: I think we should use int instead of float. What if use enters 1.3? It would be meaningless.
You could use 0 as default to mean inf and then self.max_model_history = sys.maxsize.

nvflare/edge/assessors/model_update.py
```diff
@@ -53,10 +55,41 @@ def __init__(
...
         self.device_manager = None
         self.max_model_version = max_model_version
+        self.device_wait_timeout = device_wait_timeout
         self.update_lock = threading.Lock()
         self.start_time = None
+        self.device_wait_start_time = None
+        self.should_stop_job = False
         self.register_event_handler(EventType.START_RUN, self._handle_start_run)
 
+    def _check_device_timeout(self, fl_ctx: FLContext) -> bool:
```
yanchengnv: Not sure I follow your thinking here.
Each device is different. If you want to monitor device timeout, it should be done at per-device.

nvflare/edge/executors/edge_model_executor.py
```diff
@@ -34,10 +34,22 @@
 
 class EdgeModelExecutor(EdgeTaskExecutor):
 
-    def __init__(self, aggr_factory_id: str, max_model_versions: int, update_timeout=60):
+    def __init__(
+        self,
+        aggr_factory_id: str,
+        max_model_versions: float = float("inf"),  # float to support inf, default is inf
```
yanchengnv: Change to use int.

nvflare/edge/tools/edge_job.py
```diff
@@ -94,7 +94,7 @@ def configure_server(
     def configure_client(
         self,
         aggregator_factory: AggregatorFactory,
-        max_model_versions: int,
+        max_model_versions: float = float("inf"),
```
yanchengnv: Change to: max_model_versions: int = 0

nvflare/edge/tools/edge_recipe.py
```diff
@@ -52,23 +53,43 @@ class ModelManagerConfig:
 
     def __init__(
         self,
-        max_num_active_model_versions: int = 3,
+        max_num_active_model_versions: float = float("inf"),
         max_model_version: int = 20,
         update_timeout: int = 5.0,
         num_updates_for_model: int = 100,
-        max_model_history: int = 10,
+        max_model_history: float = float("inf"),
```
yanchengnv: Change to: max_model_history = 0.
BTW, what's the difference btw max_model_versions and max_model_history? If the same, can we use the same term?

examples/advanced/edge/jobs/pt_job.py
ZiyueXu77: this is only for this heavily controlled job to make sync and async comparible, not for extended usage

nvflare/edge/assessors/buff_device_manager.py
YuanTingHsieh: @ZiyueXu77 can you comment on this ^

nvflare/edge/assessors/buff_model_manager.py
```diff
@@ -43,16 +43,22 @@ class BuffModelManager(ModelManager):
...
+        max_model_history: int = None,
         global_lr: float = 1.0,
         staleness_weight: bool = False,
     ):
         """Initialize the ModelManager.
         The aggregation scheme and weights are calculated following FedBuff paper "Federated Learning with Buffered Asynchronous Aggregation".
         The staleness_weight can be enabled to apply staleness weighting to model updates.
+
+        Special cases for max_model_history:
+        - If inf: Keep every model versions, only remove a version when all devices processing it reports back (version no longer related with any device_id in the current_selection from device_manager).
```
YuanTingHsieh: If inf -> If None (default)

nvflare/edge/assessors/buff_model_manager.py
```diff
@@ -43,16 +43,22 @@ class BuffModelManager(ModelManager):
...
+
         Args:
             num_updates_for_model (int): Number of updates required before generating a new model version.
             max_model_history (int): Maximum number of historical model versions to keep in memory.
+                - None: keep every version until all devices processing a particular version report back.
```
YuanTingHsieh: None (default)

nvflare/edge/assessors/buff_model_manager.py
```diff
@@ -69,11 +75,29 @@ def initialize_model(self, model: DXO, fl_ctx: FLContext):
...
+        # go through all versions and remove the ones:
+        # - either not in versions_to_keep
+        # - or too old (current_model_version - v >= max_model_history)
+        versions_to_remove = []
+
+        for v in self.updates.keys():
+            if v not in versions_to_keep:
+                versions_to_remove.append(v)
+            if self.max_model_history and self.current_model_version - v >= self.max_model_history:
+                versions_to_remove.append(v)
```
YuanTingHsieh: could one version be added twice?

nvflare/edge/assessors/model_manager.py
```diff
@@ -62,6 +62,18 @@ def generate_new_model(self, fl_ctx: FLContext) -> None:
         """
         pass
 
+    @abstractmethod
+    def keep_model_versions(self, versions_to_keep: Set[int], fl_ctx: FLContext) -> None:
```
YuanTingHsieh: suggest "prune_model_versions" since we also remove some versions

nvflare/edge/executors/edge_model_executor.py
YuanTingHsieh: you can use Optional[int] = None

nvflare/edge/updaters/emd.py
```diff
@@ -155,15 +155,25 @@ def process_parent_update_reply(self, reply: Shareable, fl_ctx: FLContext):
...
-        # drop old model versions
+        # drop old aggr versions based on active_model_versions from parent
         with self._update_lock:
             old_versions = []
             for mv in self.aggr_states.keys():
-                if new_state.model_version - mv > self.max_model_versions:
+                # remove the model versions that are
+                # - either not in device_selection values
+                # make a set of current active model versions from unique value of device_selection dict
+                active_model_versions = set(new_state.device_selection.values())
```
YuanTingHsieh: maybe we move this line outside of "for" loop since it will not be changed?

nvflare/edge/assessors/buff_model_manager.py
```diff
@@ -69,11 +75,29 @@ def initialize_model(self, model: DXO, fl_ctx: FLContext):
         # updates is a dict of model version to _ModelState
         self.updates[self.current_model_version] = _ModelState(ModelUpdateDXOAggregator())
 
+    def keep_model_versions(self, versions_to_keep: Set[int], fl_ctx: FLContext) -> None:
```
YuanTingHsieh: where are we using this method?

nvflare/edge/assessors/buff_device_manager.py
ZiyueXu77: I think it is ok, it only makes sense for a same device to train multiple times on the same model version if somehow the data on the device is updated, but that's not quite realistic.

nvflare/edge/assessors/buff_model_manager.py
ZiyueXu77: ah right, forgot to update the docstring

nvflare/edge/assessors/buff_model_manager.py
ZiyueXu77: added

nvflare/edge/assessors/buff_model_manager.py
ZiyueXu77: right, update to set

nvflare/edge/assessors/model_manager.py
ZiyueXu77: good idea

nvflare/edge/executors/edge_model_executor.py
ZiyueXu77: updated

nvflare/edge/tools/edge_job.py
ZiyueXu77: updated

nvflare/edge/assessors/buff_model_manager.py
ZiyueXu77: good point! updated to int and use None as default

nvflare/edge/assessors/model_update.py
ZiyueXu77: unrealated to this PR (I shouldn't have based this PR on another one)

nvflare/edge/tools/edge_recipe.py
ZiyueXu77: max_model_versions is essentially the "num_of_rounds" if it is sync

nvflare/edge/assessors/buff_model_manager.py
ZiyueXu77: the use of this method got wrongly removed during sync with merge conflict, added back

nvflare/edge/assessors/buff_model_manager.py
```diff
@@ -43,16 +43,22 @@ class BuffModelManager(ModelManager):
     def __init__(
         self,
         num_updates_for_model: int,
-        max_model_history: int,
+        max_model_history: int = None,
```
YuanTingHsieh: could do Optional[int] here for more accurate hinting, but it's ok

nvflare/edge/assessors/model_update.py
```diff
@@ -240,6 +240,9 @@ def _do_child_update(self, update: Shareable, fl_ctx: FLContext) -> (bool, Optio
                     self.log_info(fl_ctx, "Generate initial model and fill selection")
                     self.model_manager.generate_new_model(fl_ctx)
                 self.device_manager.fill_selection(self.model_manager.current_model_version, fl_ctx)
+                # prune old model versions that are no longer active
+                active_model_versions = set(self.device_manager.current_selection.values())
```
YuanTingHsieh: by doing this, you are assuming ALL device_manager's implementation has:
(1)  this`current_selection` instance variable
(2) they somehow use the "values" of that variable means active_model_versions


Instead, you could introduce another base api / method something call get_active_model_versions? or some better name you have.

Then in here you just call that while you put these assumptions inside the implementation of "buff_device_manager.py"

nvflare/edge/executors/edge_model_executor.py
```diff
@@ -34,10 +34,16 @@
 
 class EdgeModelExecutor(EdgeTaskExecutor):
 
-    def __init__(self, aggr_factory_id: str, max_model_versions: int, update_timeout=60):
+    def __init__(
+        self,
+        aggr_factory_id: str,
+        max_model_versions: int = None,
```
YuanTingHsieh: suggest Optional[int]

nvflare/edge/tools/edge_job.py
```diff
@@ -94,7 +94,7 @@ def configure_server(
     def configure_client(
         self,
         aggregator_factory: AggregatorFactory,
-        max_model_versions: int,
+        max_model_versions: int = None,
```
YuanTingHsieh: Optional[int] = None

nvflare/edge/tools/edge_job.py
```diff
@@ -114,8 +114,11 @@ def configure_client(
         if self.client_config_added:
             raise RuntimeError("client config is already added")
 
+        # check the validity of max_model_versions if not None
+        if max_model_versions:
+            check_positive_number("max_model_versions", max_model_versions)
```
YuanTingHsieh: use check_positive_int is ok

nvflare/edge/updaters/emd.py
YuanTingHsieh: this `active_model_versions = set(new_state.device_selection.values())`

nvflare/edge/assessors/model_update.py
ZiyueXu77: good point! Although we specify in device_manager base class that all need to have current_selection, available_devices, and used_devices, we did not require the values of current_selection to be the active model versions. Add get_active_model_versions()

YuanTingHsieh (approved):

PR #3645, opened 2025-08-28
Add hello cyclic recipe

YuanTingHsieh:
Add cyclic recipe and update example 

### Description

Add cyclic recipe and update example

### Types of changes

- [ ] Documentation updated.


examples/hello-world/hello-cyclic/requirements.txt
```diff
@@ -1,2 +1,2 @@
-nvflare~=2.6.0rc
+nvflare~=2.6.0
```
holgerroth, line 1: Shouldn't this be 2.7?

holgerroth (approved):
Looks good to me.

examples/hello-world/hello-cyclic/requirements.txt
YuanTingHsieh: ok will change it later

PR #3651, opened 2025-09-02
Update Android edge app and documentation

nvkevlu:
Updates Android edge app and documentation.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/edge/device/android/app/src/main/java/com/nvidia/nvflare/app/controllers/FlareRunnerController.kt
+++ b/nvflare/edge/device/android/app/src/main/java/com/nvidia/nvflare/app/controllers/FlareRunnerController.kt
@@ -12,6 +12,7 @@ import com.nvidia.nvflare.sdk.core.Signal
 import com.nvidia.nvflare.sdk.core.Dataset
 import com.nvidia.nvflare.sdk.core.Connection
 
+
 import kotlinx.coroutines.CoroutineScope
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.Job
@@ -74,6 +75,17 @@ class FlareRunnerController(
             "methods" to listOf("cnn", "xor")
         )
     
+    /**
+     * Create a connection to the NVFlare server
+     */
+    private fun createConnection(): Connection {
+        val connection = Connection(context)
+        connection.hostname.postValue(serverHost)
+        connection.port.postValue(serverPort)
+        connection.setCapabilities(capabilities)
+        return connection
+    }
+    
     fun toggleJob(job: SupportedJob) {
         if (supportedJobs.contains(job)) {
             supportedJobs = supportedJobs - job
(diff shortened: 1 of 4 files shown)


nvflare/edge/device/android/app/src/main/java/com/nvidia/nvflare/app/controllers/FlareRunnerController.kt
```diff
@@ -74,6 +75,17 @@ class FlareRunnerController(
             "methods" to listOf("cnn", "xor")
         )
     
+    /**
+     * Create a connection to the NVFlare server
+     */
+    private fun createConnection(): Connection {
+        val connection = Connection(context)
+        connection.hostname.postValue(serverHost)
+        connection.port.postValue(serverPort)
```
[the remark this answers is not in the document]
nvkevlu: Fails with: Training failed with error: java.lang.IllegalStateException: Cannot invoke setValue on a background thread

nvflare/edge/device/android/app/src/main/java/com/nvidia/nvflare/app/controllers/FlareRunnerController.kt
```diff
@@ -74,6 +75,17 @@ class FlareRunnerController(
             "methods" to listOf("cnn", "xor")
         )
     
+    /**
+     * Create a connection to the NVFlare server
+     */
+    private fun createConnection(): Connection {
```
YuanTingHsieh, line 81: we need to support HTTPS as well, but that could be in a different PR, now I see android sdk we are hardcoded HTTP

nvflare/edge/device/android/app/src/main/java/com/nvidia/nvflare/app/controllers/FlareRunnerController.kt
```diff
@@ -105,66 +117,75 @@ class FlareRunnerController(
...
-                            context.contentResolver, 
+                            context.contentResolver,
                             android.provider.Settings.Secure.ANDROID_ID
                         ) ?: "unknown"),
                         "platform" to "android",
                         "app_version" to context.packageManager.getPackageInfo(context.packageName, 0).versionName
                     ),
                     userInfo = emptyMap(),
-                    jobTimeout = 86400.0f  // 24 hours
+                    jobTimeout = Float.MAX_VALUE  // No timeout - continue until manual stop
```
YuanTingHsieh, line 188: you can make this configurable if we want user to set, but this is also fine as this is reference implementation of app

YuanTingHsieh (approved):

nvflare/edge/device/android/app/src/main/java/com/nvidia/nvflare/app/controllers/FlareRunnerController.kt
nvkevlu: yes, that is in a separate one

nvflare/edge/device/android/sdk/training/ETTrainer.kt
```diff
@@ -185,7 +193,6 @@ class ETTrainer(
     suspend fun train(config: TrainingConfig, modelData: String?): Map<String, Any> {
         Log.d(TAG, "Starting ExecuTorch training with method: ${config.method}")
         
-        // Model data should be provided separately (like iOS)
         val actualModelData = modelData ?: throw RuntimeException("No model data provided for training")
         
         if (!isInitialized || tModule == null) {
```
YuanTingHsieh: Should we always use the received global model?

Why do we need this flag of isInitialized?

In our current impl with Executorch with device, we always received the model and weights from the server not initialize locally

nvkevlu: Thanks, this should be fixed now

PR #3654, opened 2025-09-04
handle cases where quantile config is un-defined for certain features 

chesterxgchen:
### Description
In some cases, the quantile is configured but only for selected features. We should be able to handle such cases

### Types of changes

- [ ] Documentation updated.


holgerroth (approved):
Works great on my side.

PR #3655, opened 2025-09-04
Enhance recipe experiment tracking

YuanTingHsieh:
For MLFlowReceiver, it will be good to include job name.
Additionally, previously if tracking_uri is not set, the metrics is NOT storing inside the server workspace, hence the admin users has no way to access if they download the job.

recipe.job.to_server(receiver, "receiver") this one is hardcoded for now to be consistent with BaseFedJob, in next release we need to fix this as well.

### Description
MLFlowReceiver:
- If no run name is provided, include the job name for the default run name.
- We set the default tracking uri to be inside the server's workspace folder.

### Types of changes

- [ ] Documentation updated.


nvflare/app_opt/tracking/mlflow/mlflow_receiver.py
```diff
@@ -89,8 +97,14 @@ def __init__(
         self.time_since_flush = 0
         self.buff_flush_time = buffer_flush_time
 
+    def _get_tracking_uri(self, fl_ctx: FLContext):
         if self.tracking_uri:
-            mlflow.set_tracking_uri(uri=self.tracking_uri)
+            return self.tracking_uri
+
```
chesterxgchen: not sure this make sense. if the tracking URI is specified, return file://<file_path> is not correct. 

the MLflow might use local db with localhost:5000. 

I don't think you should mess this function. This is not required.

YuanTingHsieh: Yes it is working good, I will move this enhancement into another PR and explain more there.

holgerroth:
Can we add a check somewhere that tracking_uri for MLFlowReceiver is a string? I think currently, there's no check.

PR #3657, opened 2025-09-04
Enhance MLflow receiver

YuanTingHsieh:
### Issue

Previously, if users did not explicitly specify a `tracking_uri`, MLflow would default to using `./mlruns`, which is a local directory on the FL server. This made it difficult for users to access logged metrics and artifacts, as the default path was not exposed or retrievable outside the server environment.

### Description

This PR introduces an enhancement to set a more accessible default tracking_uri when none is provided by the user. Specifically, the default is now set to:

```
file://[workspace]/[job_id]/mlflow
```

This change enables users to retrieve the logged metrics and artifacts using the FlareAPI, as they are stored in a job-specific, accessible path within the workspace.

### Notes
- This PR depends on #3655.
- Ensures MLflow logs are stored in a consistent, retrievable location tied to the job and workspace.

### Types of changes

- [ ] Documentation updated.


holgerroth (dismissed):
Makes sense!

holgerroth (approved):

PR #3659, opened 2025-09-04
handle None quantile returns.

chesterxgchen:
Fixes # .
Handle cases where the quantiles are not returned from client ( such as dependency is not installed correctly on client side)

### Types of changes

- [ ] Documentation updated.


holgerroth (approved):
works well!

PR #3661, opened 2025-09-05
Add functionality for a balanced device selection across clients

ZiyueXu77:
Fixes # .

### Description

Current behavior is random selection over available device list, treating devices from any clients the same, With this PR, we can select between totally random and balanced selection (default to balanced which will lead to less randomness for simulation).
Balanced selection try to select and dispatch global model to devices such that each leaf client will have relatively similar number of active devices at any time point.  

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/edge/assessors/buff_device_manager.py
+++ b/nvflare/edge/assessors/buff_device_manager.py
@@ -13,17 +13,21 @@
 # limitations under the License.
 
 import random
+from collections import Counter, defaultdict
 from typing import Dict, Set
 
 from nvflare.edge.assessors.device_manager import DeviceManager
+from nvflare.edge.mud import PropKey
 
 
 class BuffDeviceManager(DeviceManager):
     def __init__(
         self,
         device_selection_size: int,
+        initial_min_client_num: int = 1,
         min_hole_to_fill: int = 1,
         device_reuse: bool = True,
+        device_sampling_strategy: str = "balanced",
     ):
         """Initialize the BuffDeviceManager.
         BuffDeviceManager is responsible for managing the selection of devices for model training.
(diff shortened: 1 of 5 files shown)


nvflare/edge/assessors/buff_device_manager.py
```diff
@@ -95,11 +220,16 @@ def remove_devices_from_used(self, devices: Set[str], fl_ctx) -> None:
...
         num_holes = self.device_selection_size - len(self.current_selection)
         usable_devices = set(self.available_devices.keys()) - set(self.used_devices.keys())
         num_usable_devices = len(usable_devices)
-        return num_usable_devices >= num_holes
+        if num_usable_devices < num_holes:
+            return False
+
+        # Further check if we have enough clients
+        unique_clients = set(self.device_client_map.values())
+        return len(unique_clients) >= self.initial_min_client_num
```
YuanTingHsieh, line 235: why do we need "initial_min_client_num"?

nvflare/edge/tools/edge_fed_buff_recipe.py
```diff
@@ -95,20 +98,28 @@ class DeviceManagerConfig:
...
+        device_sampling_strategy: Strategy for sampling devices when filling selection.
+            - "balanced": try to balance the usage of devices across clients.
+            - "random": randomly select devices from the available pool.
+            Default: "balanced"
     """
 
     def __init__(
         self,
         device_selection_size: int = 100,
+        initial_min_client_num: int = 1,
```
YuanTingHsieh, line 110: why do we need this?
if we need, maybe we should put it in the end of existing args.

nvflare/edge/assessors/buff_device_manager.py
```diff
@@ -66,17 +178,30 @@ def fill_selection(self, current_model_version: int, fl_ctx) -> None:
...
+                        # current_selection keeps track of devices selected for a particular model version
+                        self.current_selection[device_id] = current_model_version
+                        self.used_devices[device_id] = {
+                            "model_version": current_model_version,
+                            "selection_version": self.current_selection_version,
+                        }
+                        if not usable_devices:
+                            break
+                else:
+                    raise ValueError(f"Invalid device sampling strategy: {self.device_sampling_strategy}")
```
YuanTingHsieh, line 207: can this sampling strategy be a class they can customize, and we just pass an object in?
Should we introduce another concept "SampleStrategy" or something for the edge stuff?

nvflare/edge/assessors/buff_device_manager.py
```diff
@@ -36,18 +40,121 @@ def __init__(
...
+        # Randomize client order for more balanced distribution
+        clients_list = list(client_device_counts.items())
+        random.shuffle(clients_list)
+
+        selected_devices = set()
+        remaining_holes = num_holes
+
+        # First pass: assign minimum possible to each client
+        min_per_client = remaining_holes // len(clients_list)
+        extra_holes = remaining_holes % len(clients_list)
```
YuanTingHsieh, line 106: Do we need to do this in 2 pass?

If we just want balance, we could just do in one for loop like round robin or something to assign (total number of devices/total number of clients) to each client?

nvflare/edge/assessors/buff_device_manager.py
ZiyueXu77: we need to let server know how many client it needs to wait for - otherwise let's say we need 100 device from 4 clients, without knowing how many client are out there, server will just assign all 100 devices to the first client reporting to it - then there won't be "balanced" selection

nvflare/edge/assessors/buff_device_manager.py
ZiyueXu77: total number of devices/total number of clients may not be an integer, so that's the second pass

nvflare/edge/assessors/buff_device_manager.py
ZiyueXu77: we can, but it will need to be designed carefully considering what information may need to be included, so I think we would need another PR for this

nvflare/edge/assessors/buff_device_manager.py
```diff
@@ -36,18 +40,121 @@ def __init__(
...
+        # Count devices per client efficiently using Counter
+        client_device_counts = Counter(
+            self.device_client_map[device_id] for device_id in usable_devices if device_id in self.device_client_map
+        )
+
+        # Group devices by client using defaultdict for efficiency
+        client_devices = defaultdict(list)
+        for device_id in usable_devices:
+            if device_id in self.device_client_map:
+                client_devices[self.device_client_map[device_id]].append(device_id)
```
YuanTingHsieh, line 91: we can do these in the same for loop?

nvflare/edge/assessors/buff_device_manager.py
```diff
@@ -56,6 +163,11 @@ def update_available_devices(self, devices: Dict, fl_ctx) -> None:
             f"assessor got reported {len(devices)} available devices from child. "
             f"total num available devices: {len(self.available_devices)}",
         )
+        # add new devices to device_client_map
+        for device_id, device in devices.items():
+            client_name = device.to_dict().get(PropKey.CLIENT_NAME)
```
YuanTingHsieh, line 171: we can instead store a client_to_device_map,

and then have another set/list total devices if we want to keep track?

nvflare/edge/assessors/buff_device_manager.py
YuanTingHsieh: if we store the other one, we dont need these for loops?

nvflare/edge/assessors/buff_device_manager.py
```diff
@@ -36,18 +40,121 @@ def __init__(
...
+
+        selected_devices = set()
+        remaining_holes = num_holes
+
+        # First pass: assign minimum possible to each client
+        min_per_client = remaining_holes // len(clients_list)
+        extra_holes = remaining_holes % len(clients_list)
+
+        for i, (client_name, device_count) in enumerate(clients_list):
+            # Calculate how many devices this client should get
```
YuanTingHsieh, line 109: so we mean "how many devices this client should use (regardless of how many devices connected to this client)"?

nvflare/edge/assessors/buff_device_manager.py
```diff
@@ -95,11 +223,16 @@ def remove_devices_from_used(self, devices: Set[str], fl_ctx) -> None:
...
+    def has_enough_devices_and_clients(self, fl_ctx) -> bool:
         num_holes = self.device_selection_size - len(self.current_selection)
         usable_devices = set(self.available_devices.keys()) - set(self.used_devices.keys())
         num_usable_devices = len(usable_devices)
-        return num_usable_devices >= num_holes
+        if num_usable_devices < num_holes:
+            return False
+
+        # Further check if we have enough clients
+        unique_clients = set(self.device_client_map.values())
```
YuanTingHsieh, line 234: can we get clients from fl_ctx?

nvflare/edge/assessors/buff_device_manager.py
YuanTingHsieh: got it, we are checking this every round, so this is not just "initial" it's "min_client_num"?

ZiyueXu77: right, but it matters more at initial phase since at that time leaf nodes reports one after another, while later phase they only update the available device list - not quite likely that we lose a leaf node entirely

nvflare/edge/assessors/buff_device_manager.py
ZiyueXu77: no it cannot exceed what a clint have

nvflare/edge/assessors/buff_device_manager.py
ZiyueXu77: I don't think we have this in ctx

ZiyueXu77:
> if this is tested and working we can merge it, we can do enhancements in another PR.

yeah this is tested, I actively used it in all my recent experiments with hierarchical simulations

PR #3663, opened 2025-09-07
Hello-LR Update

chesterxgchen:
# Description
## Fix data download and preparation issue. 

**Issues**:
1) Flamby download data doesn't work consistently, in many cases, it refuse to to download as the it has cached configure indicating download completed, and abort the operation, even through the data is already gone or not available. 
Flamby download is very convoluded requires config setup and etc.
As result, one can't not download and finish experiments

2) Even you manage to download, the Flamby Dataset class failed to load the data, as the datasets now contains many other data file which are in different formats. 

**Solution**
1) replace the Flamby download with directly download dataset from the source. Only extract the original processed the datasets. 
2) pass the data_path directly to Flamby avoid convoluted steps. 

## Update  doc

update README and documentation hello-lr/index.rst file

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/examples/hello-world/hello-lr/README.md
+++ b/examples/hello-world/hello-lr/README.md
@@ -100,52 +100,41 @@ The [Newton-Raphson optimization](https://en.wikipedia.org/wiki/Newton%27s_metho
 can be described as follows.
 
 In a binary classification task with logistic regression, the
-probability of a data sample \(x\) classified as positive is formulated
+probability of a data sample $x$ classified as positive is formulated
 as:
-
-\[ p(x) = \sigma(\beta \cdot x + \beta_{0}) \]
-
-where \(\sigma(.)\) denotes the sigmoid function. We can incorporate
-\(\beta_{0}\) and \(\beta\) into a single parameter vector \(\theta =
-( \beta_{0},  \beta)\). Let \(d\) be the number
-of features for each data sample \(x\) and let \(N\) be the number of data
+$$p(x) = \sigma(\beta \cdot x + \beta_{0})$$
+where $\sigma(.)$ denotes the sigmoid function. We can incorporate
+$\beta_{0}$ and $\beta$ into a single parameter vector $\theta =
+( \beta_{0},  \beta)$. Let $d$ be the number
+of features for each data sample $x$ and let $N$ be the number of data
 samples. We then have the matrix version of the above probability
 equation:
+$$p(X) = \sigma( X \theta )$$
+Here $X$ is the matrix of all samples, with shape $N \times (d+1)$,
+having it's first column filled with value 1 to account for the
+intercept $\theta_{0}$.
 
-\[ p(X) = \sigma( X \theta ) \]
-
-Here \(X\) is the matrix of all samples, with shape \(N \times (d+1)\),
-having its first column filled with value 1 to account for the
-intercept \(\theta_{0}\).
-
-The goal is to compute parameter vector \(\theta\) that maximizes the
+The goal is to compute parameter vector $\theta$ that maximizes the
 below likelihood function:
-
-\[ L_{\theta} = \prod_{i=1}^{N} p(x_i)^{y_i} (1 - p(x_i)^{1-y_i}) \]
+$$L_{\theta} = \prod_{i=1}^{N} p(x_i)^{y_i} (1 - p(x_i)^{1-y_i})$$
 
 The Newton-Raphson method optimizes the likelihood function via
 quadratic approximation. Omitting the maths, the theoretical update
-formula for parameter vector \(\theta\) is:
-
-\[ \theta^{n+1} = \theta^{n} - H_{\theta^{n}}^{-1} \nabla L_{\theta^{n}} \]
-
+formula for parameter vector $\theta$ is:
+$$\theta^{n+1} = \theta^{n} - H_{\theta^{n}}^{-1} \nabla L_{\theta^{n}}$$
 where
-
-\[ \nabla L_{\theta^{n}} = X^{T}(y - p(X)) \]
-
-is the gradient of the likelihood function, with \(y\) being the vector
-of ground truth for sample data matrix \(X\),  and
-
-\[ H_{\theta^{n}} = -X^{T} D X \]
-
-is the Hessian of the likelihood function, with \(D\) a diagonal matrix
-where diagonal value at \((i,i)\) is \(D(i,i) = p(x_i) (1 - p(x_i))\).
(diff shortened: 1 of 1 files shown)


examples/hello-world/hello-lr/download_data.py
```diff
@@ -0,0 +1,58 @@
...
+    return filepath
+
+
+def extract_files(zip_path: str, extract_to: str, pattern: str):
+    """Extract files from a zip archive that match a given pattern."""
+    with zipfile.ZipFile(zip_path, "r") as zip_ref:
+        for file_info in zip_ref.infolist():
+            if file_info.filename.startswith(pattern) and file_info.filename.endswith(".data"):
+                zip_ref.extract(file_info, extract_to)
+                print(f"Extracted: {file_info.filename}")
```
YuanTingHsieh, line 43: minor stuff, this might print out a lot depends on how big is this heart disease data, if a few lines then we are good

examples/hello-world/hello-lr/README.md
```diff
@@ -57,49 +75,77 @@ testing sets as described below:
...
+
+Here \(X\) is the matrix of all samples, with shape \(N \times (d+1)\),
+having its first column filled with value 1 to account for the
+intercept \(\theta_{0}\).
+
+The goal is to compute parameter vector \(\theta\) that maximizes the
 below likelihood function:
-$$L_{\theta} = \prod_{i=1}^{N} p(x_i)^{y_i} (1 - p(x_i)^{1-y_i})$$
+
+\[ L_{\theta} = \prod_{i=1}^{N} p(x_i)^{y_i} (1 - p(x_i)^{1-y_i}) \]
```
YuanTingHsieh: the formula format needs to stay original for .md to render correctly

examples/hello-world/hello-lr/README.md
```diff
@@ -57,49 +75,77 @@ testing sets as described below:
...
 
 The Newton-Raphson method optimizes the likelihood function via
 quadratic approximation. Omitting the maths, the theoretical update
-formula for parameter vector $\theta$ is:
-$$\theta^{n+1} = \theta^{n} - H_{\theta^{n}}^{-1} \nabla L_{\theta^{n}}$$
+formula for parameter vector \(\theta\) is:
+
+\[ \theta^{n+1} = \theta^{n} - H_{\theta^{n}}^{-1} \nabla L_{\theta^{n}} \]
```
YuanTingHsieh: the formula format needs to stay original for .md to render correctly

examples/hello-world/hello-lr/README.md
```diff
@@ -57,49 +75,77 @@ testing sets as described below:
...
+
 where
-$$\nabla L_{\theta^{n}} = X^{T}(y - p(X))$$
-is the gradient of the likelihood function, with $y$ being the vector
-of ground truth for sample data matrix $X$,  and
-$$H_{\theta^{n}} = -X^{T} D X$$
-is the Hessian of the likelihood function, with $D$ a diagonal matrix
-where diagonal value at $(i,i)$ is $D(i,i) = p(x_i) (1 - p(x_i))$.
+
+\[ \nabla L_{\theta^{n}} = X^{T}(y - p(X)) \]
```
YuanTingHsieh: the formula format needs to stay original for .md to render correctly

examples/hello-world/hello-lr/README.md
```diff
@@ -57,49 +75,77 @@ testing sets as described below:
...
+
+is the gradient of the likelihood function, with \(y\) being the vector
+of ground truth for sample data matrix \(X\),  and
+
+\[ H_{\theta^{n}} = -X^{T} D X \]
```
YuanTingHsieh: the formula format needs to stay original for .md to render correctly

examples/hello-world/hello-lr/prepare_data.py
```diff
@@ -12,42 +12,24 @@
...
-import subprocess
-import sys
-import sysconfig
 from pathlib import Path
 
 import numpy as np
-
-np.NaN = np.nan  # Patch for NumPy 2.0 compatibility
-
 from flamby.datasets.fed_heart_disease import FedHeartDisease
```
YuanTingHsieh, line 18: code-wise is good, I am just curious what is this "flamby" doing? could we find a way to load data without it?

docs/hello-world/hello-lr/index.rst
```diff
@@ -0,0 +1,259 @@
...
+can be described as follows.
+
+In a binary classification task with logistic regression, the
+probability of a data sample :math:`x` classified as positive is formulated
+as:
+
+.. math::
+
+    p(x) = \sigma(\beta \cdot x + \beta_{0})
+
```
YuanTingHsieh, line 135: a neutral question, not sure if we should duplicate READMD.md content and index.rst.

examples/hello-world/hello-lr/download_data.py
chesterxgchen: its 4 files, one for each data center.

examples/hello-world/hello-lr/README.md
chesterxgchen: I asked cursor to clean up, I guess it mess it up.

examples/hello-world/hello-lr/prepare_data.py
chesterxgchen: Its doing a bunch of normalization, I did not want to change too much from Zhijini's original code. so I kept it the code as long it was working. Before this, I can't even it run it.

docs/hello-world/hello-lr/index.rst
chesterxgchen: I think its ok.  Either we point them to RST to read more or point to README, nothing of them are good options. RST allows to show more code pieces but the rest are the same, so they can stay where they prefer

examples/hello-world/hello-lr/README.md
chesterxgchen: restored

examples/hello-world/hello-lr/prepare_data.py
zhijinl: flamby is a python lib from Owkin. It does a bunch of pre-processing of data on the Heart Disease dataset: https://github.com/owkin/FLamby/blob/main/flamby/datasets/fed_heart_disease/dataset.py, and convert it to multiple torch Datasets with split.

I used this lib when we were working with Owkin on this Logistic Regression implementation, to maintain the same preprocessing.

I guess theoretically we could get rid of flamby and implement things ourselves, but it could be a lot more work...

PR #3668, opened 2025-09-08
Add SSL support for Android Edge App

nvkevlu:
Add SSL support for Android Edge App

### Types of changes

- [ ] Documentation updated.


nvflare/edge/device/android/app/src/main/res/xml/network_security_config.xml
```diff
@@ -0,0 +1,17 @@
...
+            <certificates src="system"/>
+        </trust-anchors>
+    </base-config>
+    
+    <!-- Specific configuration for localhost -->
+    <domain-config cleartextTrafficPermitted="true">
+        <domain includeSubdomains="true">localhost</domain>
+        <domain includeSubdomains="true">127.0.0.1</domain>
+    </domain-config>
+</network-security-config>
```
YuanTingHsieh: why do we need these?
the certificates is only needed on the proxy side?

nvkevlu: Looks like:

If you're using a standard HTTPS setup with:

- Real domain name (not localhost/127.0.0.1)
- Valid SSL certificate from a trusted CA (like Let's Encrypt, DigiCert, etc.)
- Standard HTTPS port (443 or 8443)

Then you DON'T need the network security config. Android will handle it normally.

When You DO Need the Network Security Config
If you're using:

- Self-signed certificates
- Localhost/127.0.0.1 with HTTPS
- Custom CA certificates
- Development/testing environments

Then you DO need the network security config.

I tried without this and standard HTTPS port and it worked so I can remove this for now and people can configure this themselves if they want to

nvkevlu:
I have tested that this works with both https and http

nvflare/edge/device/android/app/src/main/res/xml/network_security_config.xml
YuanTingHsieh: Thanks, you could include your findings in readme so users know what to do if they want to

PR #3676, opened 2025-09-10
Fix use_aio_grpc processing

yanchengnv:
Fixes # .

### Description

When provisioning a project, the server is defaulted to use AIO GRPC (scheme agrpc), whereas clients are defaulted to use non-AIO GRPC (scheme grpc).  But an org admin can override the default by defining "use_aio_grpc" property explicitly in comm_config.json in the startup kit's "local" folder.

The use_aio_grpc property has 3 possible values:
None - this property is not explicitly defined. 
False - do not use AIO grpc, use non-AIO instead.
True - use AIO grpc.

Driver selection is based on each driver's supported schemes. It tries to find the driver that supports the configured scheme name in the startup kit.  For GRPC, the configured scheme name is either "grpc" or "agrpc".

There are two grpc drivers (one for AIO, and another for non-AIO). The two drivers must declare their supported schemes properly based on the setting of use_aio_grpc so that the driver selection will be able to find the right driver.

For AIO GRPC Driver, its supported schemes is determined as follows:
If use_aio_grpc is None (meaning default), return ["agrpc", "agrpcs"];
If use_aio_grpc is True, return ["grpc", "grpcs", "agrpc", "agrpcs"];
Otherwise, return empty.

For non-AIO GRPC Driver, its supported schemes is determined as follows:
If use_aio_grpc is None (meaning default), return ["grpc", "grpcs"];
If use_aio_grpc is True, return empty.
Otherwise, return ["grpc", "grpcs", "agrpc", "agrpcs"];.

End result is:
When "grpc" is configured in startup: 
if use_aio_grpc is None, choose non-AIO GRPC driver;
if use_aio_grpc is True, choose AIO GRPC driver;
if use_aio_grpc is False, choose non-AIO driver.

When "agrpc" is configured in startup: 
if use_aio_grpc is None, choose AIO GRPC driver;
if use_aio_grpc is True, choose AIO GRPC driver;
if use_aio_grpc is False, choose non-AIO driver.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/edge/assessors/device_manager.py
+++ b/nvflare/edge/assessors/device_manager.py
@@ -29,7 +29,7 @@ class DeviceManager(FLComponent, ABC):
     def __init__(self):
         FLComponent.__init__(self)
         """Initialize the DeviceManager.
-        DeviceManager keeps track of two dicts: 
+        DeviceManager keeps track of two dicts:
         - current_selection for devices of current task distribution
         - available_devices containing all devices that are available for selection
         """
--- a/nvflare/edge/assessors/model_manager.py
+++ b/nvflare/edge/assessors/model_manager.py
@@ -30,10 +30,10 @@ class ModelManager(FLComponent, ABC):
     def __init__(self):
         FLComponent.__init__(self)
         """Initialize the ModelManager.
-        ModelManager keeps track of three things: 
+        ModelManager keeps track of three things:
         - current_model holding the current global model
         - current_model_version holding the current global model version
-        - updates containing all received updates for updating the global model 
+        - updates containing all received updates for updating the global model
         """
         self.current_model = None
         self.current_model_version = 0
--- a/nvflare/edge/device/android/app/src/main/AndroidManifest.xml
+++ b/nvflare/edge/device/android/app/src/main/AndroidManifest.xml
@@ -11,7 +11,6 @@
         android:label="@string/app_name"
         android:roundIcon="@mipmap/ic_launcher_round"
         android:supportsRtl="true"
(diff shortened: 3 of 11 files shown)


chesterxgchen:
Good write up, we need to put this into the documentation. 

@yanchengnv 
Question, what is "agrpcs" -- notice the "s" at the end

nvidianz:
> Good write up, we need to put this into the documentation.
> 
> @yanchengnv Question, what is "agrpcs" -- notice the "s" at the end

a stands for async, s for secure

nvidianz (approved):

nvflare/fuel/f3/drivers/aio_grpc_driver.py
```diff
@@ -411,9 +417,5 @@ def shutdown(self):
     def get_urls(scheme: str, resources: dict) -> (str, str):
         secure = requires_secure_connection(resources)
         if secure:
-            if use_aio_grpc():
-                scheme = "grpcs"
-            else:
-                scheme = "agrpcs"
-
+            scheme = "grpcs"
```
YuanTingHsieh, line 420: Should this be agrpcs?

yanchengnv: No, should always be grpcs. This is the scheme to be exposed.

yanchengnv:
> Good write up, we need to put this into the documentation.
> 
> @yanchengnv Question, what is "agrpcs" -- notice the "s" at the end

"s" stands for "secure"

PR #3678, opened 2025-09-10
Add decomposer register

yanchengnv:
Fixes # .

### Description

Due to security concern raised by NVIDIA security team, we had to take out "auto decomposer registration" function from FOBS.  The application will have to explicitly register custom decomposers.

This PR adds a DecomposerRegister widget that can be used to configure and load any custom decomposers.
To use it, simply include it as a component in config_fed_server.json and/or config_fed_client.json.

"components": [
    {
      "id": "decomposer_reg",
      "path": "path.to.custom.decomposer",
      "args": {
        "classes": ["nvflare.app_opt.pt.decomposers.TensorDecomposer"]
      }
    }
  ]

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/app_common/widgets/decomposer_reg.py
+++ b/nvflare/app_common/widgets/decomposer_reg.py
@@ -1,0 +1,29 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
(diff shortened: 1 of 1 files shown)


yanchengnv:
> This is fine
> 
> For CC mode, does this also opened up security issue as it dynamic load python code ? The validation code now needs to check the configuration's class loading I guess

I think this is okay. This component itself does not add new code. The configured classes must already exist in the code base, otherwise exception will occur.

YuanTingHsieh (approved):

PR #3679, opened 2025-09-10
Sign local folders of startup kits.

IsaacYangSLA:
### Description

This PR changes the signing process of startup kits to cover local folders.  Originally, only startup folders are signed.

### Types of changes

- [ ] Documentation updated.


nvflare/lighter/impl/signature.py
```diff
@@ -41,3 +42,5 @@ def build(self, project: Project, ctx: ProvisionContext):
         for p in project.get_all_participants():
             dest_dir = ctx.get_kit_dir(p)
             self._do_sign(root_pri_key, dest_dir)
+            dest_dir = ctx.get_local_dir(p)
```
YuanTingHsieh, line 45: code is good, just wondering about the CC workflow,

IF this local is signed, then users at each site can not modify them to fit their needs anymore?
How can they edit these config files inside?
Or we are disallowing editing of these configs

chesterxgchen (approved):
LGTM
leave for Yan to review

nvflare/lighter/impl/signature.py
chesterxgchen: No, I think based on Issac, the validation code needs check the signature to see if the signature is the changed or not. In that validation code, even though the local is signed, we dont need to check it. I think here just prepare for the future needs ( as I see it)

chesterxgchen: but I did not check the validation code to see if he check the all files or just selected files

YuanTingHsieh (approved):

PR #3684, opened 2025-09-11
Make via-file decomposers configurable

yanchengnv:
Fixes # .

### Description

Apheris has started playing with 2.7 RC and noticed slowdown in model transfer.  Though this is to be expected (since it involves creation of a file and downloading of the file chunk by chunk), there is room for improvement, especially if the network bandwidth is big.

This PR makes it possible to configure decomposers for optimal model transfer performance.

The behavior of such decomposers (TensorDecomposer and NumpyArrayDecomposer) can be configured with two parameters:

**xxx_min_file_size_for_streaming** - specifies the min model size required for using file streaming. If the model size is less than this size, the model will be included in the message directly, and no file streaming is needed. If not specified, the default value is 2MB. For systems with large bandwidth and/or memory, this parameter can take a larger value (e.g. 1GB).

If the specified size is 0, then the object will use native decomposition - no file is used.

**xxx_streaming_per_request_timeout** - specifies the per-request timeout for streaming messages. If not specified, default to 10 secs.  The value of this parameter can be adjusted based on the network performance.

Where xxx is the prefix of the specific decomposer. For TensorDecomposer, the prefix is "tensor"; for NumpyArrayDecomposer, the prefix is "np".

Note that by default, NumpyArrayDecomposer uses native decomposition.  TensorDecomposer does not support native decomposition.

Both parameters can be set in job config files (config_fed_server.json and config_fed_client.json) as top-level elements.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/recipe/run.py
+++ b/nvflare/recipe/run.py
@@ -64,9 +64,6 @@ class Run:
     @contextmanager
     def _secure_session(self) -> Generator:
         """Context manager for secure session handling."""
-        if self._is_sim_env():
-            raise RuntimeError("Secure session not needed for simulation environment")
-
         sess = None
         try:
             sess = new_secure_session(**self._get_session_params())
@@ -77,17 +74,17 @@ class Run:
             if sess:
                 sess.close()
 
-    def get_status(self) -> str:
+    def get_status(self) -> Optional[str]:
         """Get the status of the run.
 
         Returns:
-            str: The status of the run.
+            Optional[str]: The status of the run, or None if called in a simulation environment.
         """
         if self._is_sim_env():
             print(
-                "get_status will always return completed in a simulation environment, please check the log inside the workspace returned by get_result()"
+                "get_status is not supported in a simulation environment, please check the log inside the workspace returned by get_result()"
             )
-            return "completed"
+            return None
 
         with self._secure_session() as sess:
             return sess.get_job_status(self.job_id)


nvflare/app_common/decomposers/numpy_decomposers.py
```diff
@@ -57,6 +58,12 @@ def supported_type(self):
 
 
 class NumpyArrayDecomposer(ViaFileDecomposer):
+
+    def __init__(self):
+        ViaFileDecomposer.__init__(self)
+        self.config_var_prefix = "np_"
+        self.min_size_for_file = 0  # by default do not use file downloading.
```
YuanTingHsieh: instead of overwriting, should we do:
```
ViaFileDecomposer.__init__(self, config_var_prefix="np_", min_size_for_file=0)
```

nvflare/app_opt/pt/decomposers.py
```diff
@@ -78,6 +79,13 @@ def _safe_save(state_dict, filename: str) -> Optional[dict]:
 
 class TensorDecomposer(ViaFileDecomposer):
 
+    def __init__(self):
+        ViaFileDecomposer.__init__(self)
+        self.config_var_prefix = "tensor_"
+
+        # if the file size for collected items is < 2MB, it will be attached to the message.
+        self.min_size_for_file = 2 * 1024 * 1024  # 2MB
```
YuanTingHsieh: Instead of overwrite, should we do

`ViaFileDecomposer.__init__(self, config_var_prefix="tensor_", min_size_for_file=2 * 1024 * 1024)`

nvflare/fuel/utils/fobs/decomposers/via_file.py
```diff
@@ -90,7 +99,8 @@ def __init__(self):
         self.decompose_ctx_key = f"{self.prefix}_dc"  # kept in fobs_ctx: each target type has its own DecomposeCtx
         self.items_key = f"{self.prefix}_items"  # in fobs_ctx: each target type has its own set of items
         self.file_downloader_class = FileDownloader
-        self.min_size_for_file = _MIN_SIZE_FOR_FILE
+        self.min_size_for_file = 0
+        self.config_var_prefix = ""
```
YuanTingHsieh: these two can get from init arguments

nvflare/fuel/utils/app_config_utils.py
```diff
@@ -0,0 +1,42 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from nvflare.apis.fl_constant import SystemConfigs
+from nvflare.fuel.utils.config_service import ConfigService
+
+
+def get_positive_float_var(var_name, default):
```
YuanTingHsieh, line 18: Note that in this method, people can pass a negative number to default.
Thus it does not guarantee to return a positive number.

nvflare/fuel/utils/app_config_utils.py
```diff
@@ -0,0 +1,42 @@
...
+    #   the job config could define variable var_name;
+    #   the user could define OS env var NVFLARE_VAR_NAME (the var_name turned to uppercase)
+    value = ConfigService.get_float_var(name=var_name, conf=SystemConfigs.APPLICATION_CONF, default=default)
+    if value is None:
+        return default
+    else:
+        return value if value > 0.0 else default
+
+
+def get_positive_int_var(var_name, default):
```
YuanTingHsieh, line 29: Note that in this method, people can pass a negative number to default.
Thus it does not guarantee to return a positive number.

nvflare/app_common/decomposers/numpy_decomposers.py
yanchengnv: Good idea! Changed.

nvflare/app_opt/pt/decomposers.py
yanchengnv: Fixed.

nvflare/fuel/utils/fobs/decomposers/via_file.py
yanchengnv: Fixed.

nvflare/fuel/utils/app_config_utils.py
yanchengnv: True. This was refactored from previous code. We could improve if necessary.

nvflare/fuel/utils/app_config_utils.py
yanchengnv: Correct, inherited from previous code.

YuanTingHsieh (approved):

PR #3689, opened 2025-09-15
Increase the grace period so error msg won't show up

YuanTingHsieh:
Fixes QA bug FLARE-2599

### Description

Increases the grace period so the shutdown error msg won't show up easily, as suggested by QA

### Types of changes

- [ ] Documentation updated.


IsaacYangSLA (approved):
LGTM.

PR #3690, opened 2025-09-15
Add Python utility code to verify startup kits

IsaacYangSLA:
### Description

Sign all folders in startup kits with rootCA at provision time and store signatures in signature.json inside each folder.
Add a utility code `nvflare/tool/verify_startup_kits.py` to verify all files inside a folder with signatures.
If any file is not signed or its signature is wrong, this utility exits with status code 1.

### Types of changes

- [ ] Documentation updated.


IsaacYangSLA:
Example usage of utility code:

> $ python verify_startup_kits.py -f workspace/example_project/prod_00/server1/ -c workspace/example_project/prod_00/server1/startup/rootCA.pem 
$ echo $?
0
$ touch workspace/example_project/prod_00/server1/local/hello_world.txt
$ python verify_startup_kits.py -f workspace/example_project/prod_00/server1/ -c workspace/example_project/prod_00/server1/startup/rootCA.pem 
$ echo $?
1
$ rm workspace/example_project/prod_00/server1/local/hello_world.txt
$ echo "1234" >> workspace/example_project/prod_00/server1/local/resources.json.default
$ python verify_startup_kits.py -f workspace/example_project/prod_00/server1/ -c workspace/example_project/prod_00/server1/startup/rootCA.pem 
$ echo $?
1

nvflare/tool/verify_startup_kits.py
```diff
@@ -0,0 +1,36 @@
...
+    parser = argparse.ArgumentParser()
+    parser.add_argument("-f", "--folder", type=str, help="folder to be verified")
+    parser.add_argument("-c", "--root_ca_path", type=str, help="root CA path")
+    args = parser.parse_args()
+    folder = args.folder
+    root_ca_path = args.root_ca_path
+    result = verify_folder_signature(
+        folder, root_ca_path, single_signer=True, signature_file=ProvFileName.SIGNATURE_JSON
+    )
+    if result is False:
```
nvidianz, line 32: Should print result here for quick check. You can add an option to disable/enable the printing, .e.g -q (quiet) or -v (verbose).

IsaacYangSLA: It's like the any / all functions in Python.  It only tells the True/False, not which one causes it.  Additionally, this function returns immediately whenever it detects any invalid signature.  Therefore, it will not tell all files that fail signature checks.  I think there is no need to return such lists.

nvflare/lighter/utils.py
```diff
@@ -224,29 +224,33 @@ def sign_folders(folder, signing_pri_key, crt_path, max_depth=9999):
...
             json.dump(signatures, f)
-        shutil.copyfile(crt_path, os.path.join(root, NVFLARE_SUBMITTER_CRT_FILE))
+        if crt_path is not None:
+            shutil.copyfile(crt_path, os.path.join(root, NVFLARE_SUBMITTER_CRT_FILE))
         if depth >= max_depth:
             break
 
 
-def verify_folder_signature(src_folder, root_ca_path):
+def verify_folder_signature(src_folder, root_ca_path, single_signer=False, signature_file=NVFLARE_SIG_FILE):
```
YuanTingHsieh, line 235: can you add a dostring to explain what does single_signer and other args mean?

nvflare/lighter/utils.py
```diff
@@ -224,29 +224,33 @@ def sign_folders(folder, signing_pri_key, crt_path, max_depth=9999):
                 signing_pri_key=signing_pri_key,
             )
 
-        with open(os.path.join(root, NVFLARE_SIG_FILE), "wt") as f:
+        with open(os.path.join(root, signature_file), "wt") as f:
             json.dump(signatures, f)
```
YuanTingHsieh, line 228: are we using this dump inside this method? I saw in other place we explicit save

nvflare/lighter/impl/signature.py
```diff
@@ -30,7 +30,7 @@ class SignatureBuilder(Builder):
     @staticmethod
     def _do_sign(root_pri_key, dest_dir):
         if dest_dir and os.path.isdir(dest_dir):
-            signatures = sign_all(dest_dir, root_pri_key)
+            signatures = sign_folders(dest_dir, root_pri_key)
             with open(os.path.join(dest_dir, ProvFileName.SIGNATURE_JSON), "wt") as f:
                 json.dump(signatures, f)
```
YuanTingHsieh: here we explicit save...

nvflare/lighter/impl/signature.py
```diff
@@ -41,6 +41,6 @@ def build(self, project: Project, ctx: ProvisionContext):
 
         for p in project.get_all_participants():
             dest_dir = ctx.get_kit_dir(p)
-            self._do_sign(root_pri_key, dest_dir)
+            sign_folders(dest_dir, root_pri_key, signature_file=ProvFileName.SIGNATURE_JSON)
```
YuanTingHsieh, line 34: here is saved inside the method, should we make it consistent

nvflare/tool/verify_startup_kits.py
```diff
@@ -0,0 +1,36 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import argparse
+
+from nvflare.lighter.constants import ProvFileName
+from nvflare.lighter.utils import verify_folder_signature
+
+
+def main():
```
YuanTingHsieh, line 22: do we want to modify the sub_start.sh to call this tool if we pass --verify or something

nvflare/tool/verify_startup_kits.py
nvidianz: I meant to print the verification result only (true/false). So when I verify the signature in interactive mode I can just run the program. Otherwise I have to do,

   verify_startup_kits.py 
   echo $?

nvflare/tool/verify_startup_kits.py
IsaacYangSLA: Yes.  If sub_start.sh is called with --verify, it will do python verify_startup_kits.py.

nvflare/lighter/impl/signature.py
IsaacYangSLA: the _do_sign method is not used any more.  Let me remove it.

nvflare/lighter/utils.py
IsaacYangSLA: Just pushed another commit with docstrings.

nvflare/lighter/utils.py
IsaacYangSLA: The other place is not used.  It's removed in the latest commit.

nvflare/lighter/impl/signature.py
IsaacYangSLA: _do_sign is removed.

YuanTingHsieh (approved):
This PR is good, I guess the sub_start.sh change is in another PR?

PR #3694, opened 2025-09-16
Update the sub_start.sh to support pre launch verification

IsaacYangSLA:
### Description

The sub_start.sh now has --verify option.  This option will check if the startup kit has been fully signed by the root CA.  It will not start the nvflare if this verification fails.

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh (dismissed):
some copilot suggestions might be good

PR #3696, opened 2025-09-17
Replace torch_dtype with dtype for latest transformers library

cyyever:
### Description

To avoid deprecation warnings when running on transformers>= 4.56.1.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/examples/advanced/llm_hf/src/hf_sft_peft_fl.py
+++ b/examples/advanced/llm_hf/src/hf_sft_peft_fl.py
@@ -158,7 +158,7 @@ def main():
         model_name_or_path,
         device_map=device_map,
         use_cache=False,
-        torch_dtype=torch.bfloat16,
+        dtype=torch.bfloat16,
     )
     torch.set_default_dtype(default_dtype)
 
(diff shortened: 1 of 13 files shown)


YuanTingHsieh:
If the transformer version is older, does this work?
If not we might need to update the documentation and example @ZiyueXu77 @holgerroth

cyyever:
@YuanTingHsieh What is the minimum supported `transformers`?

holgerroth:
Makes sense to me. In transformers version 4.56.1 and newer, the previously used argument torch_dtype has been replaced by dtype in various APIs, especially within pipelines. The dtype argument serves exactly the same function: it determines the PyTorch data type used to load the model and underlying tensors.

But you should update the requirement.txt of the examples to state the new versions this was tested with.

chesterxgchen:
should we update this after the 2.7.0 release ?

cyyever:
> should we update this after the 2.7.0 release ? 

Of course we can, let me add version check to requirements

YuanTingHsieh:
covered by PR #3765 
closing this one, thanks for the contribution!

PR #3702, opened 2025-09-19
Fix hello numpy

nvkevlu:
Fix issue with hello numpy.

### Types of changes

- [ ] Documentation updated.


nvflare/app_common/np/recipes/fedavg.py
```diff
@@ -157,21 +158,25 @@ def __init__(
         shareable_generator_id = job.to_server(shareable_generator, id="shareable_generator")
         aggregator_id = job.to_server(self.aggregator, id="aggregator")
 
+        # Handle initial model if provided
+        persistor_id = ""
```
chesterxgchen, line 162: should this init to None or empty string ?

nvkevlu: It seems the constructor expects string, so empty string should be good here

chesterxgchen (approved):

PR #3708, opened 2025-09-19
Increase default timeout value

YuanTingHsieh:
In the new release, we use file to stream, that help decreasing the memory consumption at a cost of slower speed.
So increase this timeout to accommodate that.

### Description
 Increase default timeout from 60 to 300

### Types of changes

- [ ] Documentation updated.


nvflare/app_common/executors/client_api_launcher_executor.py
```diff
@@ -32,12 +32,12 @@ def __init__(
         launch_timeout: Optional[float] = None,
         task_wait_timeout: Optional[float] = None,
         last_result_transfer_timeout: float = 300.0,
-        external_pre_init_timeout: float = 60.0,
-        peer_read_timeout: Optional[float] = 60.0,
+        external_pre_init_timeout: float = 300.0,
```
ZiyueXu77, line 35: this does not directly impact the error (60 is ok), but it is fine to set it to 300 too

ZiyueXu77 (approved):
quick fix for timeout error, we shall have a holistic solution for timeout settings on various parts of our pipeline

PR #3710, opened 2025-09-19
Add overall summary for XGB encryption and acceleration [skip ci]

ZiyueXu77:
Clarify the speedup expectation, security implementation under different modes

### Description

To both readme and doc

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/edge/device/android/README.md
+++ b/nvflare/edge/device/android/README.md
@@ -97,8 +97,9 @@ cp -r nvflare/edge/device/android/sdk \
 ### Step 5: Add Training Data
 
 ```bash
-# Copy CIFAR-10 dataset
-cp nvflare/edge/ios/NVFlareMobile/NVFlareMobile/Assets.xcassets/cifar10/data_batch_1.dataset/data_batch_1.bin \
+mkdir nvflare/edge/device/android/app/src/main/assets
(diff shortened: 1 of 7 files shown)


docs/user_guide/data_scientist_guide/federated_xgboost/secure_xgboost_user_guide.rst
```diff
@@ -23,6 +23,104 @@ The local GRPC ports are selected automatically by NVFlare.
 
 The encryption is handled in XGBoost by encryption plugins, which are external components that can be installed at runtime.
 
+Collaboration Modes and Data Split
```
chesterxgchen, line 25: for the prev. document
```
When running with NVFlare, all the GRPC connections in XGBoost are local and the messages > are forwarded to other clients through NVFlare's CellNet communication.
The local GRPC ports are selected automatically by NVFlare.
```

Why do we need to emphasize is GRPC ? we could change the driver to HTTPS, the same thing still works. 

Also, why do we need to bring CellNet, this doesn't mean anything to end user

ZiyueXu77: yeah not sure the initial motivation of this by the original writer, update it to higher level descriptions.

PR #3715, opened 2025-09-22
Update async_num assessor's selection list content

ZiyueXu77:
Fixes QA 5532344.

### Description

We have an update on model version control and auto aggregator keeping, which changes the content of selection list, the async_num will need to be updated accordingly for it to run

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/edge/assessors/async_num.py
+++ b/nvflare/edge/assessors/async_num.py
@@ -213,13 +213,16 @@ class AsyncNumAssessor(Assessor):
                 for _ in range(num_holes):
                     device_id = random.choice(list(usable_devices))
                     usable_devices.remove(device_id)
-                    self.current_selection[device_id] = self.current_selection_version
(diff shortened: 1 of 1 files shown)


nvflare/edge/assessors/async_num.py
```diff
@@ -213,13 +213,16 @@ def _fill_selection(self, fl_ctx: FLContext):
                 for _ in range(num_holes):
                     device_id = random.choice(list(usable_devices))
                     usable_devices.remove(device_id)
-                    self.current_selection[device_id] = self.current_selection_version
-                    self.used_devices[device_id] = self.current_model_version
+                    self.current_selection[device_id] = self.current_model_version
+                    self.used_devices[device_id] = {
+                        "model_version": self.current_model_version,
+                        "selection_version": self.current_selection_version,
```
YuanTingHsieh, line 219: later these strings could be replaced with some constant definition

PR #3718, opened 2025-09-24
make safetensors as required dependency

chesterxgchen:
Fixes # .

### Description

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -33,6 +33,7 @@ install_requires =
     aiohttp
     pyhocon
     pydantic>=2.0
+    safetensors
 
 [options.extras_require]
 HE =
@@ -61,7 +62,6 @@ app_opt =
     %(TRACKING)s
     %(MONITORING)s
     pytorch_lightning
-    safetensors
     xgboost
     bitsandbytes
 app_opt_mac =


setup.cfg
```diff
@@ -61,7 +62,6 @@ app_opt =
     %(TRACKING)s
     %(MONITORING)s
     pytorch_lightning
```
nvidianz, line 64: Those 2 should go to PT.

chesterxgchen: understand, I intentionally make this required, not optional (which is under PT). 

This is taking @holgerroth's point that avoid forcing everyone to do additional install when using Pytorch.

setup.cfg
```diff
@@ -33,6 +33,7 @@ install_requires =
     aiohttp
     pyhocon
     pydantic>=2.0
+    safetensors
```
YuanTingHsieh, line 36: If safetensors is only used in PT we should not make it general dependency

chesterxgchen: Holger argument that, which I tend to agree now, it is true this is pytorch only, but considering this: 

 install an extra  light weight lib is not going to hurt anything, it does have extra dependency not used in case we use no PT cases. 
 
 But by not install it by default, we force everyone using pytorch to do extra install, which they did not need to do before, user experience is much worse.

YuanTingHsieh: cool, no problem, I just did some search and realize this "safetensors" also supports TF and NumPy, 
maybe we do want to utilize it for general purpose later.

YuanTingHsieh: https://huggingface.co/docs/safetensors/index

YuanTingHsieh (approved):

setup.cfg
chesterxgchen: that make even more sense to put it as required package

PR #3725, opened 2025-09-29
Fix simulator hanging when client script raise OOM

YuanTingHsieh:
Fix a simulator hanging issue when a client thread throws Exception.

### Description

When a client thread throws Exception (for example OOM happens PyTorch or TensorFlow), our simulator has
``stop_run = conn.recv()``
which will blocks forever.

Adding mechanism to check if the process has died before calling ``conn.recv()``

### Types of changes

- [ ] Documentation updated.


nvflare/private/fed/app/simulator/simulator_runner.py
```diff
@@ -772,23 +773,66 @@ def do_one_task(self, client, num_of_threads, gpu, lock, timeout=60.0, task_name
...
+                if process.poll() is not None:
+                    self.logger.error(
+                        f"Subprocess died for client {client.client_name}, exit code: {process.returncode}"
+                    )
+                    return True, None, client.client_name
 
-            with lock:
-                if num_of_threads != len(self.federated_clients):
-                    next_client = self.get_next_run_client(gpu)
+                stop_run = conn.recv()
```
chesterxgchen, line 785: connection only receive stop_run instruction  ?

nvflare/private/fed/app/simulator/simulator_runner.py
```diff
@@ -772,23 +773,66 @@ def do_one_task(self, client, num_of_threads, gpu, lock, timeout=60.0, task_name
...
+
+                if stop_run:
+                    # Check again before second recv
+                    if process.poll() is not None:
+                        self.logger.error(
+                            f"Subprocess died for client {client.client_name} before second recv, exit code: {process.returncode}"
+                        )
+                        return True, None, client.client_name
+
+                    end_run_client = conn.recv()
```
chesterxgchen, line 796: end_run_client is different from stop run ? 

what kind of messages conn will receive (stop_run, end_run_client) ? what else ? should we add a comments here so people read the code knows whats expected

nvflare/private/fed/app/simulator/simulator_runner.py
YuanTingHsieh: this was the original code I just wrap it with a try except block

nvflare/private/fed/app/simulator/simulator_runner.py
YuanTingHsieh: this was the original code I just wrap it with a try except block

I could add some explanation docstring here

nvflare/private/fed/app/simulator/simulator_runner.py
YuanTingHsieh: added a comment

nvflare/private/fed/app/simulator/simulator_runner.py
YuanTingHsieh: added a comment

PR #3741, opened 2025-10-01
Add Tensor Stream component for efficient safetensors-based model tensor streaming

rfilgueiras:
### Description

This PR introduces a new component for NVFlare called **Tensor Stream**, which optimizes the transfer of model tensors between server and clients by streaming them separately from the main task payload.  

Instead of embedding tensors directly into task data/results, Tensor Stream serializes and streams each tensor individually using **safetensors**, ensuring efficient and scalable communication. On the receiving side, tensors are deserialized, stored in memory, and reinjected back into the task payload to preserve the original structure.  

#### Server → Clients (Task Data)  
- Server serializes and streams all model tensors to clients.  
- Tensors are removed from the task data before transmission.  
- Clients receive and deserialize streamed tensors, then add them back into the task data.  

server-side

#### Clients → Server (Task Results)  
- Clients serialize and stream model parameter tensors to the server.  
- Tensors are removed from the task results before transmission.  
- Server receives and deserializes streamed tensors, then reinjects them into the task results.  

client-side

This approach replaces the standard tensor transfer in NVFlare, avoiding unnecessary copies and enabling more efficient handling of large model states.  

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/examples/advanced/tensor-stream/README.md
+++ b/examples/advanced/tensor-stream/README.md
@@ -166,7 +166,7 @@ The `TensorClientStreamer` is responsible for managing tensor streaming on the c
   - When set to `None`, defaults to `["train"]`.
   - Useful for specifying different tasks like `["train", "validate"]` if you want streaming for multiple task types.
 
-- **`entry_timeout`** (`float`, default: `30.0`)
+- **`tensor_send_timeout`** (`float`, default: `30.0`)
   - Timeout in seconds for individual tensor entry transfer operations when sending results to the server.
   - Controls how long to wait for each chunk of tensor data to be transferred.
   - May need to be increased for very large tensors or slow network connections.
@@ -185,7 +185,7 @@ client_streamer = TensorClientStreamer()
 client_streamer = TensorClientStreamer(
     format=ExchangeFormat.PYTORCH,
     tasks=["train", "validate"],
-    entry_timeout=60.0,  # Increase timeout for large models
+    tensor_send_timeout=60.0,  # Increase timeout for large models
 )
 ```
 
@@ -207,7 +207,7 @@ The `TensorServerStreamer` manages tensor streaming on the server side. It sends
   - When set to `None`, defaults to `["train"]`.
   - Should match the tasks configured in `TensorClientStreamer`.
 
-- **`entry_timeout`** (`float`, default: `30.0`)
+- **`tensor_send_timeout`** (`float`, default: `30.0`)
   - Timeout in seconds for individual tensor entry transfer operations when sending to clients.
   - Controls how long to wait for each chunk of tensor data to be transferred.
   - Should be tuned based on model size and network bandwidth.
@@ -232,14 +232,14 @@ server_streamer = TensorServerStreamer()
 # With custom timeouts for large-scale deployments
 server_streamer = TensorServerStreamer(
     format=ExchangeFormat.PYTORCH,
-    entry_timeout=60.0,
+    tensor_send_timeout=60.0,
     wait_send_task_data_all_clients_timeout=600.0  # 10 minutes for many clients
 )
 
 # For multiple task types
 server_streamer = TensorServerStreamer(
     tasks=["train", "validate"],
-    entry_timeout=45.0
+    tensor_send_timeout=45.0
 )
 ```
 
@@ -251,9 +251,9 @@ Choose timeout values based on your deployment scenario:
 
 **Small models (< 100MB), fast network:**
 ```python
-client_streamer = TensorClientStreamer(entry_timeout=30.0)
+client_streamer = TensorClientStreamer(tensor_send_timeout=30.0)
 server_streamer = TensorServerStreamer(
-    entry_timeout=30.0,
+    tensor_send_timeout=30.0,
     wait_send_task_data_all_clients_timeout=300.0
 )
 ```
@@ -261,11 +261,10 @@ server_streamer = TensorServerStreamer(
 **Large models (> 1GB), moderate network:**
 ```python
 client_streamer = TensorClientStreamer(
-    entry_timeout=90.0,
-    wait_for_task_data_tensors_timeout=180.0
+    tensor_send_timeout=90.0
 )
 server_streamer = TensorServerStreamer(
-    entry_timeout=90.0,
+    tensor_send_timeout=90.0,
     wait_send_task_data_all_clients_timeout=900.0  # 15 minutes
 )
 ```
@@ -273,11 +272,10 @@ server_streamer = TensorServerStreamer(
 **Very large models (> 10GB), many clients:**
 ```python
 client_streamer = TensorClientStreamer(
-    entry_timeout=120.0,
-    wait_for_task_data_tensors_timeout=300.0
+    tensor_send_timeout=120.0
 )
 server_streamer = TensorServerStreamer(
-    entry_timeout=120.0,
+    tensor_send_timeout=120.0,
     wait_send_task_data_all_clients_timeout=1800.0  # 30 minutes
 )
 ```
--- a/nvflare/app_opt/tensor_stream/receiver.py
+++ b/nvflare/app_opt/tensor_stream/receiver.py
@@ -54,6 +54,7 @@ class TensorReceiver:
         # key: task_id, value: tensors received from the peer
         self.tensors: dict[str, TensorsMap] = {}
         self.tensor_events = {}  # Maps task_id to Event objects
+        self.error_events = {}  # Maps task_id to error messages
         self.lock = threading.Lock()
         self.logger = get_obj_logger(self)
         self._register()
@@ -90,8 +91,10 @@ class TensorReceiver:
 
         if not success:
             with self.lock:
+                exc = ValueError(f"Failed to receive tensors from peer '{peer_name}' and task '{task_id}'.")
+                self.error_events[task_id] = exc
                 self.tensor_events[task_id].set()  # Wake waiting threads
-            raise ValueError(f"Failed to receive tensors from peer '{peer_name}' and task '{task_id}'.")
+            return  # Early return to prevent further processing
 
         tensors = fl_ctx.get_custom_prop(TensorCustomKeys.SAFE_TENSORS_PROP_KEY)
         if not tensors:
@@ -193,6 +196,11 @@ class TensorReceiver:
         remaining_timeout = max(0, timeout - (time.time() - start_wait))
         if not event.wait(timeout=remaining_timeout):
             raise TimeoutError(f"No tensors received from peer '{peer_name}'. Task ID: '{task_id}'.")
+        else:
+            with self.lock:
+                if task_id in self.error_events:
+                    exc = self.error_events.pop(task_id)
+                    raise exc
 
     def on_tensor_received(self, task_id: str, tensor: TensorsMap):
         """Callback when tensors are received.
--- a/tests/unit_test/app_opt/tensor_stream/test_receiver.py
+++ b/tests/unit_test/app_opt/tensor_stream/test_receiver.py
@@ -111,7 +111,7 @@ class TestTensorReceiver:
         assert receiver.tensors[task_id] == random_torch_tensors
 
     def test_save_tensors_cb_failure(self, mock_streamable_engine, mock_fl_context):
-        """Test _save_tensors_cb with failed tensor reception."""
+        """Test _save_tensors_cb with failed tensor reception stores error and sets event."""
         receiver = TensorReceiver(
             engine=mock_streamable_engine, ctx_prop_key=FLContextKey.TASK_DATA, format=ExchangeFormat.PYTORCH
         )
@@ -121,9 +121,8 @@ class TestTensorReceiver:
         mock_fl_context.get_peer_context().get_identity_name.return_value = peer_name
         mock_fl_context.set_custom_prop("task_id", task_id)
 
-        # Call the callback with success=False - should raise ValueError
-        with pytest.raises(ValueError, match=f"Failed to receive tensors from peer '{peer_name}' and task '{task_id}'"):
-            receiver._save_tensors_cb(False, mock_fl_context)
+        # Call the callback with success=False - should NOT raise but store error
+        receiver._save_tensors_cb(False, mock_fl_context)
 
         # Verify no tensors were stored
         assert task_id not in receiver.tensors
@@ -132,6 +131,44 @@ class TestTensorReceiver:
         # Verify event was created and set (to wake waiting threads)
         assert task_id in receiver.tensor_events
         assert receiver.tensor_events[task_id].is_set()
+
+        # Verify error was stored for later retrieval
+        assert task_id in receiver.error_events
+        stored_error = receiver.error_events[task_id]
+        assert isinstance(stored_error, ValueError)
+        assert f"Failed to receive tensors from peer '{peer_name}' and task '{task_id}'" in str(stored_error)
+
+    def test_save_tensors_cb_failure_does_not_access_tensors(self, mock_streamable_engine, mock_fl_context):
+        """Test that _save_tensors_cb returns early on failure without accessing tensor data."""
+        receiver = TensorReceiver(
+            engine=mock_streamable_engine, ctx_prop_key=FLContextKey.TASK_DATA, format=ExchangeFormat.PYTORCH
+        )
+
+        peer_name = "test_client"
+        task_id = "test_task_early_return"
+        mock_fl_context.get_peer_context().get_identity_name.return_value = peer_name
+        mock_fl_context.set_custom_prop("task_id", task_id)
+
+        # Create a sentinel value to track if SAFE_TENSORS_PROP_KEY is accessed
+        tensor_key_accessed = False
+        original_get_custom_prop = mock_fl_context.get_custom_prop
+
+        def track_get_custom_prop(key):
+            nonlocal tensor_key_accessed
+            if key == TensorCustomKeys.SAFE_TENSORS_PROP_KEY:
+                tensor_key_accessed = True
+            return original_get_custom_prop(key)
+
+        mock_fl_context.get_custom_prop = track_get_custom_prop
+
+        # Call with success=False - should return early without checking for tensors
+        receiver._save_tensors_cb(False, mock_fl_context)
+
+        # Verify SAFE_TENSORS_PROP_KEY was not accessed
+        assert not tensor_key_accessed, "SAFE_TENSORS_PROP_KEY should not be accessed on failure"
+
+        # Verify error was stored
+        assert task_id in receiver.error_events
 
     def test_save_tensors_cb_no_tensors(self, mock_streamable_engine, mock_fl_context):
         """Test _save_tensors_cb when no tensors are found in context."""
@@ -572,7 +609,7 @@ class TestTensorReceiver:
             receiver.wait_for_tensors(task_id, peer_name, timeout=0.1)
 
     def test_wait_for_tensors_with_failed_send(self, mock_streamable_engine, mock_fl_context):
-        """Test wait_for_tensors completes when send fails (event is set)."""
+        """Test wait_for_tensors raises error when stream reception fails."""
         receiver = TensorReceiver(
             engine=mock_streamable_engine, ctx_prop_key=FLContextKey.TASK_DATA, format=ExchangeFormat.PYTORCH
         )
@@ -583,18 +620,17 @@ class TestTensorReceiver:
         mock_fl_context.get_peer_context().get_identity_name.return_value = peer_name
         mock_fl_context.set_custom_prop("task_id", task_id)
 
-        # Simulate a failed send - this sets the event to wake waiting threads
-        try:
-            receiver._save_tensors_cb(False, mock_fl_context)
-        except ValueError:
-            pass  # Expected to raise
+        # Simulate a failed send - this stores error and sets the event
+        receiver._save_tensors_cb(False, mock_fl_context)
 
-        # Wait should complete immediately since event was set during failure
-        # But no tensors were stored
-        receiver.wait_for_tensors(task_id, peer_name, timeout=1.0)
+        # Wait should complete immediately but raise the stored error
+        with pytest.raises(ValueError, match=f"Failed to receive tensors from peer '{peer_name}' and task '{task_id}'"):
+            receiver.wait_for_tensors(task_id, peer_name, timeout=1.0)
 
         # Verify no tensors were stored
         assert task_id not in receiver.tensors
+        # Verify error was removed after being raised
+        assert task_id not in receiver.error_events
 
     def test_wait_for_tensors_multiple_tasks(self, mock_streamable_engine, mock_fl_context, random_torch_tensors):
         """Test wait_for_tensors works correctly with multiple concurrent tasks."""
@@ -629,3 +665,119 @@ class TestTensorReceiver:
         assert task_id_2 in receiver.tensors
         assert receiver.tensors[task_id_1] == tensors_1
         assert receiver.tensors[task_id_2] == tensors_2
+
+    def test_error_propagation_workflow(self, mock_streamable_engine, mock_fl_context):
+        """Test complete workflow of error storage and propagation from callback to wait_for_tensors."""
+        receiver = TensorReceiver(
+            engine=mock_streamable_engine, ctx_prop_key=FLContextKey.TASK_DATA, format=ExchangeFormat.PYTORCH
+        )
+
+        peer_name = "test_client"
+        task_id = "test_task_error_workflow"
+
+        mock_fl_context.get_peer_context().get_identity_name.return_value = peer_name
+        mock_fl_context.set_custom_prop("task_id", task_id)
+
+        # Step 1: Simulate streaming failure via callback
+        receiver._save_tensors_cb(False, mock_fl_context)
+
+        # Step 2: Verify error was stored and event was set
+        assert task_id in receiver.error_events
+        assert task_id in receiver.tensor_events
+        assert receiver.tensor_events[task_id].is_set()
+
+        # Step 3: wait_for_tensors should raise the stored error
+        with pytest.raises(ValueError, match=f"Failed to receive tensors from peer '{peer_name}' and task '{task_id}'"):
+            receiver.wait_for_tensors(task_id, peer_name, timeout=1.0)
+
+        # Step 4: Verify error was cleaned up after being raised
+        assert task_id not in receiver.error_events
+
+    def test_multiple_tasks_one_fails(self, mock_streamable_engine, mock_fl_context, random_torch_tensors):
+        """Test that failure of one task doesn't affect other tasks."""
+        receiver = TensorReceiver(
+            engine=mock_streamable_engine, ctx_prop_key=FLContextKey.TASK_DATA, format=ExchangeFormat.PYTORCH
+        )
+
+        peer_name = "test_client"
+        task_id_success = "test_task_success"
+        task_id_failure = "test_task_failure"
+
+        mock_fl_context.get_peer_context().get_identity_name.return_value = peer_name
+
+        # Task 1: Success
+        mock_fl_context.set_custom_prop("task_id", task_id_success)
+        mock_fl_context.set_custom_prop(TensorCustomKeys.SAFE_TENSORS_PROP_KEY, random_torch_tensors)
+        receiver._save_tensors_cb(True, mock_fl_context)
+
+        # Task 2: Failure
+        mock_fl_context.set_custom_prop("task_id", task_id_failure)
+        receiver._save_tensors_cb(False, mock_fl_context)
+
+        # Verify success task has tensors
+        assert task_id_success in receiver.tensors
+        receiver.wait_for_tensors(task_id_success, peer_name, timeout=1.0)
+
+        # Verify failure task raises error
+        assert task_id_failure not in receiver.tensors
+        with pytest.raises(
+            ValueError, match=f"Failed to receive tensors from peer '{peer_name}' and task '{task_id_failure}'"
+        ):
+            receiver.wait_for_tensors(task_id_failure, peer_name, timeout=1.0)
+
+    def test_save_tensors_cb_failure_no_custom_prop_cleanup(self, mock_streamable_engine, mock_fl_context):
+        """Test that custom properties are NOT cleared when success=False (early return)."""
+        receiver = TensorReceiver(
+            engine=mock_streamable_engine, ctx_prop_key=FLContextKey.TASK_DATA, format=ExchangeFormat.PYTORCH
+        )
+
+        peer_name = "test_client"
+        task_id = "test_task_no_cleanup"
+        mock_fl_context.get_peer_context().get_identity_name.return_value = peer_name
+        mock_fl_context.set_custom_prop("task_id", task_id)
+
+        # Track if set_custom_prop is called to clear properties (should not happen on failure)
+        cleanup_called = False
+        original_set_custom_prop = mock_fl_context.set_custom_prop
+        initial_call_count = 0
+
+        def track_set_custom_prop(key, value):
+            nonlocal cleanup_called, initial_call_count
+            original_set_custom_prop(key, value)
+            if key in (TensorCustomKeys.SAFE_TENSORS_PROP_KEY, TensorCustomKeys.TASK_ID) and value is None:
+                cleanup_called = True
+            initial_call_count += 1
+
+        # Reset mock to track new calls
+        mock_fl_context.set_custom_prop = track_set_custom_prop
+        initial_call_count = 0  # Reset after setup
+
+        # Call with success=False
+        receiver._save_tensors_cb(False, mock_fl_context)
+
+        # Verify set_custom_prop was NOT called to clear properties
+        assert not cleanup_called, "Custom properties should not be cleared on failure"
+
+    def test_event_creation_on_first_failure(self, mock_streamable_engine, mock_fl_context):
+        """Test that tensor event is created on first callback even when it fails."""
+        receiver = TensorReceiver(
+            engine=mock_streamable_engine, ctx_prop_key=FLContextKey.TASK_DATA, format=ExchangeFormat.PYTORCH
+        )
+
+        peer_name = "test_client"
+        task_id = "test_task_event_creation"
+        mock_fl_context.get_peer_context().get_identity_name.return_value = peer_name
+        mock_fl_context.set_custom_prop("task_id", task_id)
+
+        # Verify event doesn't exist yet
+        assert task_id not in receiver.tensor_events
+
+        # Call with success=False
+        receiver._save_tensors_cb(False, mock_fl_context)
+
+        # Verify event was created and set
+        assert task_id in receiver.tensor_events
+        assert receiver.tensor_events[task_id].is_set()
+
+        # Verify error was stored
+        assert task_id in receiver.error_events


nvflare/app_opt/tensor_stream/server.py
```diff
@@ -0,0 +1,157 @@
...
+    def send_tensors_to_client(self, fl_ctx: FLContext):
+        """Send tensors to the client after task data filtering.
+
+        Args:
+            fl_ctx (FLContext): The FLContext for the current operation.
+        """
+        try:
+            self.sender.send(fl_ctx, self.entry_timeout)
+        except ValueError as e:
+            self.system_panic(fl_ctx, f"Failed to send tensors: {e}")
```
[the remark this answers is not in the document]
rfilgueiras: Fixed.

nvflare/app_opt/tensor_stream/server.py
```diff
@@ -0,0 +1,157 @@
...
+        """
+        if event_type == EventType.START_RUN:
+            self.initialize(fl_ctx)
+        elif event_type == EventType.BEFORE_TASK_DATA_FILTER:
+            self.data_cleaned = False
+        elif event_type == EventType.AFTER_TASK_DATA_FILTER:
+            try:
+                self.send_tensors_to_client(fl_ctx)
+            except Exception as e:
+                self.system_panic(fl_ctx, f"Failed to send tensors: {e}")
```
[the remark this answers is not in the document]
rfilgueiras: Fixed.

nvflare/app_opt/tensor_stream/receiver.py
```diff
@@ -0,0 +1,137 @@
...
+            self.logger.error(
+                f"Peer '{fl_ctx.get_identity_name()}': received DXO with task data from peer"
+                f"'{peer_name}' and {total_values} tensors.",
+            )
+
+        if self.format == "torch":
+            dxo["data"] = tensors
+        else:
+            dxo["data"] = {k: v.numpy() for k, v in tensors.items()}
+
```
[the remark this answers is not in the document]
rfilgueiras: Added new functions to handle the recursive conversion from and to numpy.

nvflare/app_opt/tensor_stream/utils.py
```diff
@@ -0,0 +1,77 @@
...
+        raise ValueError(f"Unsupported context property key: {ctx_prop_key}")
+
+
+def get_targets_for_ctx_and_prop_key(fl_ctx: FLContext, ctx_prop_key: str) -> str:
+    """Get the peer identity name from the FLContext.
+
+    Args:
+        fl_ctx (FLContext): The FLContext for the current operation.
+    Returns:
+        str: The identity name of the peer.
```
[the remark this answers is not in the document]
rfilgueiras: Fixed.

nvflare/app_opt/tensor_stream/consumer.py
```diff
@@ -0,0 +1,140 @@
...
+        self.tensors: TensorsMap = {}
+        self.total_bytes = {}
+        self.root_keys = []
+
+    def consume(
+        self,
+        shareable: Shareable,
+        stream_ctx: StreamContext,
+        fl_ctx: FLContext,
+    ) -> tuple[bool, Shareable]:
```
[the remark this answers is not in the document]
rfilgueiras: NVFlare only supports Python 3.9+.

nvflare/app_opt/tensor_stream/producer.py
```diff
@@ -0,0 +1,134 @@
...
+        tensors_keys: List of keys for the tensors to be sent.
+        start: Starting index for streaming.
+        current: Current index in the streaming process.
+        end: Ending index for streaming.
+    Methods:
+        produce(stream_ctx, fl_ctx): Produces the next chunk of tensors to be sent.
+        process_replies(replies, stream_ctx, fl_ctx): Processes replies from peers after sending tensors.
+    """
+
+    def __init__(self, tensors: dict[str, torch.Tensor], entry_timeout: float, root_key: str = ""):
```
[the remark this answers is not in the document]
rfilgueiras: Same here, only Python 3.9+ is supported.

nvflare/app_opt/tensor_stream/receiver.py
```diff
@@ -0,0 +1,141 @@
...
+            ctx_prop_key (FLContextKey): The context property key to receive tensors for.
+            format (ExchangeFormat): The format of the tensors to receive. Default is ExchangeFormat.PYTORCH.
+            channel (str): The channel to use for streaming. Default is TENSORS_CHANNEL.
+        """
+        super().__init__()
+        self.engine = engine
+        self.ctx_prop_key = ctx_prop_key
+        self.format = format
+        self.channel = channel
+        self.tensors: dict[str, TensorsMap] = {}
```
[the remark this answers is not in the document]
rfilgueiras: Same here, only Python 3.9+ is supported.

nvflare/app_opt/tensor_stream/sender.py
```diff
@@ -0,0 +1,160 @@
...
+            self.logger.warning("No task data found in FLContext")
+            return None
+
+        dxo = from_shareable(data)
+        if dxo.data_kind not in (DataKind.WEIGHTS, DataKind.WEIGHT_DIFF):
+            return None
+
+        return dxo
+
+    def _get_tensors_from_dxo(self, dxo: DXO, key: str) -> dict[str, torch.Tensor]:
```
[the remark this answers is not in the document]
rfilgueiras: Same here, only Python 3.9+ is supported.

nvflare/app_opt/tensor_stream/types.py
```diff
@@ -0,0 +1,31 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import torch
+
+TENSORS_CHANNEL = "tensor_stream"
+SAFE_TENSORS_PROP_KEY = "_safe_tensors_blob_"
+
+TensorsMap = dict[str, torch.Tensor] | dict[str, dict[str, torch.Tensor]]
```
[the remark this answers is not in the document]
rfilgueiras: Fixed.

nvflare/app_opt/tensor_stream/utils.py
```diff
@@ -0,0 +1,124 @@
...
+    """
+    if ctx_prop_key == FLContextKey.TASK_DATA:
+        return TensorTopics.TASK_DATA
+    elif ctx_prop_key == FLContextKey.TASK_RESULT:
+        return TensorTopics.TASK_RESULT
+    else:
+        raise ValueError(f"Unsupported context property key: {ctx_prop_key}")
+
+
+def get_targets_for_ctx_and_prop_key(fl_ctx: FLContext, ctx_prop_key: str) -> list[str]:
```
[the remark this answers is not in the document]
rfilgueiras: Same here, only Python 3.9+ is supported.

nvflare/app_opt/tensor_stream/utils.py
```diff
@@ -0,0 +1,124 @@
...
+    """
+    if ctx_prop_key == FLContextKey.TASK_DATA:
+        return [fl_ctx.get_peer_context().get_identity_name()]
+    elif ctx_prop_key == FLContextKey.TASK_RESULT:
+        return [SERVER_SITE_NAME]
+    else:
+        raise ValueError(f"Unsupported context property key: {ctx_prop_key}")
+
+
+def to_numpy_recursive(obj: torch.Tensor | dict[str, torch.Tensor]) -> dict[str, np.ndarray] | np.ndarray:
```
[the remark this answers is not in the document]
rfilgueiras: Fixed.

chesterxgchen:
@rfilgueiras  one question

since the tensor is streamed separately from the task data (payload),  do we have a race condition where the tensor might be arrived the earlier than the task data ? How do we handle this ?

rfilgueiras:
> @rfilgueiras one question
> 
> since the tensor is streamed separately from the task data (payload), do we have a trance condition where the tensor might be arrived the earlier than the task data ? How do we handle this ?

This is a great question @chesterxgchen :)

I had this issue and solved it by waiting for all transfers to be done before cleaning the context and releasing the tasks:
https://github.com/rfilgueiras/NVFlare/blob/484f51eb1f4d59bdebffc6cbfcf042616878e315/nvflare/app_opt/tensor_stream/server.py#L132

Each client's task must not only wait for the transfer to the same client, but also until the transfer to all clients is done.
Only the last client will actually clean the context. It was a surprise to me that the task's context is shared among all client tasks.

I will introduce a timeout in the wait logic, which should be the maximum time it should wait for all the tensors to be transferred to all clients.

Note that in this line, after each client is done, it will increment `num_task_data_sent`:
https://github.com/rfilgueiras/NVFlare/blob/484f51eb1f4d59bdebffc6cbfcf042616878e315/nvflare/app_opt/tensor_stream/server.py#L130

nvflare/app_opt/tensor_stream/utils.py
```diff
@@ -0,0 +1,124 @@
...
+    """Recursively convert objects with a `numpy` method to numpy arrays."""
+    if hasattr(obj, "numpy"):
+        return obj.numpy()
+    elif isinstance(obj, dict):
+        return {k: to_numpy_recursive(v) for k, v in obj.items()}
+    else:
+        raise ValueError(f"Unsupported object type: {type(obj)}")
+
+
+def to_torch_recursive(obj: np.ndarray | dict[str, np.ndarray], device: torch.device = "cpu") -> dict[str, torch.Tensor] | torch.Tensor:
```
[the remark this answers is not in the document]
rfilgueiras: Fixed.

chesterxgchen:
> Each client's task must not only wait for the transfer to the same client, but also until the transfer to all clients is done.
> Only the last client will actually clean the context. It was a surprise to me that the task's context is shared among all client tasks.

I don't understand this part. Each client should only need to wait for its own tensor. It the tensor is delivered, then it can process. This is independent to all other clients. Some client can be fast or slow, depending on their network or computing resources. Waiting for all tensors from all clients seems not necessary. 


Are you afraid that the server task data is cleaned too earlier and before it it broadcasted to the individual site ?

chesterxgchen:
> > Each client's task must not only wait for the transfer to the same client, but also until the transfer to all clients is done.
> > Only the last client will actually clean the context. It was a surprise to me that the task's context is shared among all client tasks.
> 
> I don't understand this part. Each client should only need to wait for its own tensor. It the tensor is delivered, then it can process. This is independent to all other clients. Some client can be fast or slow, depending on their network or computing resources. Waiting for all tensors from all clients seems not necessary.
> 
> Are you afraid that the server task data is cleaned too earlier and before it it broadcasted to the individual site ?

I think I understand what you trying to do here.  You want to make sure the data are actually sent to all clients before you clean the task data.You wait for data "sent" ( not necessarily delivered) to all clients, then call clean data on server side. 

send() is non blocking call. This is ok. 

But what I was asking is different question.  All above is on the server side, on the client side (Client Executor), is it possible to have a race condition that Task data and Tensor Steam arrived in different sequence 
For example, 
    Tensor arrived first, but the Task Data is not yet here yet, so the instruction (task name) is not yet given, the executor has no idea what to do with the Tensor received.

rfilgueiras:
> > Each client's task must not only wait for the transfer to the same client, but also until the transfer to all clients is done.
> > Only the last client will actually clean the context. It was a surprise to me that the task's context is shared among all client tasks.
> 
> I don't understand this part. Each client should only need to wait for its own tensor. It the tensor is delivered, then it can process. This is independent to all other clients. Some client can be fast or slow, depending on their network or computing resources. Waiting for all tensors from all clients seems not necessary.
> 
> Are you afraid that the server task data is cleaned too earlier and before it it broadcasted to the individual site ?

Yes. I am reading the tensors from the fl_ctx without copying them. If I clean the context before they finish, it impacts the ones that are still processing the tensors.

This was strange to me at first, since they might have different filters applied, etc.
Can you double-check if this is how it works, or is there something I can do to avoid the waiting?

rfilgueiras:
> > > Each client's task must not only wait for the transfer to the same client, but also until the transfer to all clients is done.
> > > Only the last client will actually clean the context. It was a surprise to me that the task's context is shared among all client tasks.
> > 
> > 
> > I don't understand this part. Each client should only need to wait for its own tensor. It the tensor is delivered, then it can process. This is independent to all other clients. Some client can be fast or slow, depending on their network or computing resources. Waiting for all tensors from all clients seems not necessary.
> > Are you afraid that the server task data is cleaned too earlier and before it it broadcasted to the individual site ?
> 
> I think I understand what you trying to do here. You want to make sure the data are actually sent to all clients before you clean the task data.You wait for data "sent" ( not necessarily delivered) to all clients, then call clean data on server side.
> 
> send() is non blocking call. This is ok.

In this case, the send is synchronous and blocks on each thread (for each client) on the server side until all data is produced by the server and consumed by the client.

I updated the logic to be clearer about what is happening. The send operation is triggered by the AFTER_TASK_DATA_FILTER event:
https://github.com/rfilgueiras/NVFlare/blob/55bbde8c73496833400d2df92091375efc9e3308/nvflare/app_opt/tensor_stream/server.py#L115

> But what I was asking is different question. All above is on the server side, on the client side (Client Executor), is it possible to have a race condition that Task data and Tensor Steam arrived in different sequence For example, Tensor arrived first, but the Task Data is not yet here yet, so the instruction (task name) is not yet given, the executor has no idea what to do with the Tensor received.

This is fine, the server sends the tensors, and they must arrive before the task data.
It will be stored on the receiver at the client side:
https://github.com/rfilgueiras/NVFlare/blob/55bbde8c73496833400d2df92091375efc9e3308/nvflare/app_opt/tensor_stream/receiver.py#L70

When the task data arrives, the client streamer handles the BEFORE_TASK_DATA_FILTER event:
https://github.com/rfilgueiras/NVFlare/blob/55bbde8c73496833400d2df92091375efc9e3308/nvflare/app_opt/tensor_stream/client.py#L98

It will ensure the task data is updated with the received tensors.

yanchengnv (commented):
This is a very clever solution to use event handlers for tensor streaming. I wonder whether it could be generalized for other types of streaming.

With this solution, the streaming is between the server and the client. This works well only if the client is the one that does training. But this is not always the case. For example, in case of Client API based applications, the client does not do training. It is the external app that actually does training.

A better solution (for future) might be using the Object Downloader. Instead of pushing the tensors to the client, the server will wait for the trainer to pull tensors from it. The trainer could be the client, or the external app.

rfilgueiras:
> This is a very clever solution to use event handlers for tensor streaming. I wonder whether it could be generalized for other types of streaming.

Yes, the main advantage of using event handlers is that neither the controller nor executors need to be modified.

What other types of streaming services do you think of?
This is very flexible if you look at it; we only need different producers and consumers that know how to stream the data.

> With this solution, the streaming is between the server and the client. This works well only if the client is the one that does training. But this is not always the case. For example, in case of Client API based applications, the client does not do training. It is the external app that actually does training.

Yes, the current solution is not compatible with the ClientAPI for many reasons, even when it is running in-memory.

> A better solution (for future) might be using the Object Downloader. Instead of pushing the tensors to the client, the server will wait for the trainer to pull tensors from it. The trainer could be the client, or the external app.

This would work in the opposite way than I do now.

From Server -> Client (Task Data):
- Server registers a listener passing the Producer factory
- Client requests the task data from the Server, passing the Consumer factory

From Client -> Server (Task Result):
- Client registers a listener passing the Producer factory
- Server requests the task result from the ClientAPI, passing the Consumer factory

I would love to discuss how to improve these interfaces to facilitate such patterns:
- ClientAPI (in and ex process)
- Client Engine
- Server Engine

One possibility would be:
- Task Data and Task Result have a proxy data object with metadata (replacing the original data)
- Peer would know (based on headers) that it should be dealing with a streaming transfer
- Using the provided metadata, it would download the data (by consuming the stream)
- Data would be added to the Task Data and Result by normal processing

rfilgueiras:
I understand that using the `ObjDownloader` would allow it to support the ClientAPI.
The main reason I did not consider it was to maintain compatibility with NVFlare 2.6.

nvflare/app_opt/tensor_stream/receiver.py
```diff
@@ -0,0 +1,193 @@
...
+        Args:
+            fl_ctx (FLContext): The FLContext for the current operation.
+        """
+        peer_name = fl_ctx.get_peer_context().get_identity_name()
+        task_id = fl_ctx.get_prop(FLContextKey.TASK_ID, None)
+        if not task_id:
+            raise ValueError("No task_id found in FLContext.")
+
+        # get and remove the tensors from the local store
+        tensors = self.tensors.pop(task_id)
```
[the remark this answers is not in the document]
rfilgueiras: Fixed.

nvflare/app_opt/tensor_stream/receiver.py
```diff
@@ -0,0 +1,193 @@
...
+            msg = f"Task data kind is not WEIGHTS or WEIGHT_DIFF: {dxo['kind']}"
+            self.logger.error(msg)
+            raise RuntimeError(msg)
+
+        if len(dxo["data"]) == 0 and not tensors:
+            self.logger.error(
+                f"Peer '{fl_ctx.get_identity_name()}':received task with empty data, no tensors "
+                f"are present for '{peer_name}'. Task ID: '{task_id}'.",
+            )
+            raise RuntimeError(msg)
```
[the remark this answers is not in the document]
rfilgueiras: Fixed.

nvflare/app_opt/tensor_stream/client.py
```diff
@@ -0,0 +1,156 @@
...
+        """Sends tensors to the server before sending the task result.
+
+        Args:
+            fl_ctx (FLContext): The FLContext for the current operation.
+        """
+        self.sender.store_tensors(fl_ctx)
+        try:
+            self.sender.send(fl_ctx, self.entry_timeout)
+        except ValueError as e:
+            pass
```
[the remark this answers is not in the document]
rfilgueiras: Fixed.

examples/advanced/tensor-stream/trainer.py
```diff
@@ -0,0 +1,65 @@
...
+        num_train_epochs=1,
+        fp16=False,
+        learning_rate=2e-4,
+        loss_type="nll",
+        lr_scheduler_type="cosine",
+        gradient_checkpointing=True,
+        dataset_text_field="text",
+        output_dir=f"./results/{client_name}",
+        report_to="none",
+        use_cpu=not torch.cuda.is_available(),
```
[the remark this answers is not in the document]
rfilgueiras: Fixed.

examples/advanced/tensor-stream/trainer.py
```diff
@@ -0,0 +1,65 @@
...
+    )
+    return training_args
+
+
+def train():
```
YuanTingHsieh, line 48: neutral on this, these methods can also be merged into client.py as well

examples/advanced/tensor-stream/README.md
```diff
@@ -0,0 +1,391 @@
...
+from nvflare.app_opt.tensor_stream import TensorClientStreamer
+from nvflare.client.config import ExchangeFormat
+
+# Simple configuration
+client_streamer = TensorClientStreamer()
+
+# With custom format, tasks, and timeout
+client_streamer = TensorClientStreamer(
+    format=ExchangeFormat.PYTORCH,
+    tasks=["train", "validate"]
```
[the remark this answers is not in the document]
rfilgueiras: Fixed.

nvflare/app_opt/tensor_stream/client.py
```diff
@@ -0,0 +1,157 @@
...
+from .utils import clean_task_result
+
+
+class TensorClientStreamer(FLComponent):
+    """TensorClientSender handles receiving task data and sending task results tensors from/to server.
+
+    It uses a StreamableEngine, TensorReceiver, and TensorSender to manage tensor streaming on the client side.
+    Attributes:
+        format (str): The format of the tensors to send. Default is "pytorch".
+        entry_timeout (float): Timeout for tensor entry transfer operations. Default is 30.0 seconds.
```
YuanTingHsieh: suggest rename to tensor_send_timeout

nvflare/app_opt/tensor_stream/client.py
```diff
@@ -0,0 +1,157 @@
...
+                    # use default short timeout to wait for tensors, data should already received
+                    self.receiver.wait_for_tensors(task_id, peer_name)
+
+                self.receiver.set_ctx_with_tensors(fl_ctx)
+            except Exception as e:
+                self.system_panic(str(e), fl_ctx)
+        elif event_type == EventType.AFTER_TASK_RESULT_FILTER:
+            self.sender = TensorSender(self.engine, FLContextKey.TASK_RESULT, self.format, self.tasks)
+            try:
+                self.send_tensors_to_server(fl_ctx)
```
YuanTingHsieh: minor stuff, maybe this self.sender = xxx can be put inside the send_tensors_to_server method since we also set to None there

nvflare/app_opt/tensor_stream/client.py
```diff
@@ -0,0 +1,157 @@
...
+            event_type (str): The type of event to handle.
+            fl_ctx (FLContext): The FLContext for the current operation.
+        """
+        if event_type == EventType.START_RUN:
+            self.initialize(fl_ctx)
+        elif event_type == EventType.BEFORE_TASK_DATA_FILTER:
+            task_id = fl_ctx.get_prop(FLContextKey.TASK_ID)
+            peer_name = fl_ctx.get_peer_context().get_identity_name()
+            try:
+                if self.enable_request_task_data_tensors:
```
YuanTingHsieh: do we have a use case when this is True?

nvflare/app_opt/tensor_stream/sender.py
```diff
@@ -0,0 +1,118 @@
...
+        self.ctx_prop_key = ctx_prop_key
+        self.format = format
+        self.tasks = tasks
+        self.channel = channel
+        # key: task_id, value: tensors to send to the peer
+        self.task_params: dict[str, TensorsMap] = defaultdict(dict)
+        self.logger = get_obj_logger(self)
+
+    def store_tensors(self, fl_ctx: FLContext):
+        """ "Parse tensors from the FLContext and store them for sending.
```
YuanTingHsieh: extra **"**

nvflare/app_opt/tensor_stream/sender.py
```diff
@@ -0,0 +1,118 @@
...
+from .utils import get_dxo_from_ctx, get_targets_for_ctx_and_prop_key, get_topic_for_ctx_prop_key
+
+
+class TensorSender:
+    """Handles sending tensors between server and clients."""
+
+    def __init__(
+        self,
+        engine: StreamableEngine,
+        ctx_prop_key: FLContextKey,
```
YuanTingHsieh: ctx_prop_key is not FLContextKey object though, suggest ctx_prop_key: str

nvflare/app_opt/tensor_stream/sender.py
```diff
@@ -0,0 +1,118 @@
...
+        format: ExchangeFormat,
```
YuanTingHsieh: same here, suggest format: str

nvflare/app_opt/tensor_stream/sender.py
```diff
@@ -0,0 +1,118 @@
...
+        tasks: list[str],
+        channel: str = TENSORS_CHANNEL,
+    ):
+        """Initialize the TensorSender.
+
+        Args:
+            engine (StreamableEngine): The streamable engine to use for streaming.
+            ctx_prop_key (FLContextKey): The context property key to send tensors for.
```
YuanTingHsieh: suggest ctx_prop_key (str): The context property key representing which tensors to send.

nvflare/app_opt/tensor_stream/sender.py
```diff
@@ -0,0 +1,118 @@
...
+
+class TensorSender:
+    """Handles sending tensors between server and clients."""
+
+    def __init__(
+        self,
+        engine: StreamableEngine,
+        ctx_prop_key: FLContextKey,
+        format: ExchangeFormat,
+        tasks: list[str],
```
YuanTingHsieh, line 35: suggest adding docstring for tasks

tasks (list[str]): List of task names this sender handles.

nvflare/app_opt/tensor_stream/utils.py
```diff
@@ -0,0 +1,258 @@
...
+    """
+    if ctx_prop_key == FLContextKey.TASK_DATA:
+        return TensorTopics.TASK_DATA
+    elif ctx_prop_key == FLContextKey.TASK_RESULT:
+        return TensorTopics.TASK_RESULT
+    else:
+        raise ValueError(f"Unsupported context property key: {ctx_prop_key}")
+
+
+def get_targets_for_ctx_and_prop_key(fl_ctx: FLContext, ctx_prop_key: str) -> list[str]:
```
YuanTingHsieh: suggest get_targets_from_ctx_and_prop_key

nvflare/app_opt/tensor_stream/producer.py
```diff
@@ -0,0 +1,153 @@
...
+            f"Peer '{fl_ctx.get_identity_name()}': produced blobs for peer '{peer_name}' "
+            f"with {self.num_tensors} tensors, total size: "
+            f"{round(self.total_bytes / (1024 * 1024), 2)} Mbytes ({self.total_bytes} bytes). "
+            f"Task ID: {self.task_id}"
+        )
+        self.logger.info(msg)
+
+    def process_replies(
+        self,
+        replies: Dict[str, Shareable],
```
YuanTingHsieh: you can use dict for typehints directly 3.9+

YuanTingHsieh: suggest dict

nvflare/app_opt/tensor_stream/utils.py
```diff
@@ -0,0 +1,258 @@
...
+        if isinstance(value, torch.Tensor):
+            tensors[key] = value
+        elif isinstance(value, np.ndarray):
+            tensors[key] = torch.from_numpy(value)
+        elif isinstance(value, dict):
+            yield from chunk_tensors_from_params(value, parent_keys + [key], chunk_size)
+
+    if tensors:
+        if chunk_size is None or chunk_size >= len(tensors):
+            yield parent_keys, tensors
```
YuanTingHsieh: this parent_keys could be a list which is different from typehint of Tuple

nvflare/app_opt/tensor_stream/producer.py
```diff
@@ -0,0 +1,153 @@
...
+            raise ValueError("No tensors received for serialization.")
+        self.task_id = task_id
+        self.entry_timeout = entry_timeout
+        self.last = False
+        self.total_bytes = 0
+        self.num_tensors = 0
+        # Pass tensors to generator; they're not stored in this class to minimize memory usage.
+        # The generator will handle serialization and the tensors can be garbage collected
+        # after the generator completes.
+        self.chunks_generator = chunk_tensors_from_params(tensors)
```
YuanTingHsieh, line 66: just curious, what if there are 1 tensor that are very big?

nvflare/app_opt/tensor_stream/receiver.py
```diff
@@ -0,0 +1,199 @@
...
+
+
+class TensorReceiver:
+    """A component to receive tensors from clients using NVFlare's streaming capabilities."""
+
+    def __init__(
+        self,
+        engine: StreamableEngine,
+        ctx_prop_key: FLContextKey,
+        format: ExchangeFormat = ExchangeFormat.PYTORCH,
```
YuanTingHsieh: suggest type hint with str

nvflare/app_opt/tensor_stream/receiver.py
```diff
@@ -0,0 +1,199 @@
...
+        Args:
+            fl_ctx (FLContext): The FLContext for the current operation.
+        """
+        peer_name = fl_ctx.get_peer_context().get_identity_name()
+        task_id = fl_ctx.get_prop(FLContextKey.TASK_ID, None)
+        if not task_id:
+            raise ValueError("No task_id found in FLContext.")
+
+        # get and remove the tensors from the local store
+        tensors = self.tensors.pop(task_id, None)
```
YuanTingHsieh, line 122: could there be two threads doing this at the same time?
do we need to put this inside a lock?

nvflare/app_opt/tensor_stream/receiver.py
```diff
@@ -0,0 +1,199 @@
...
+            msg = f"Unsupported tensor format: {self.format}"
+            self.logger.error(msg)
+            raise RuntimeError(msg)
+
+        s["DXO"] = dxo
+        fl_ctx.set_prop(self.ctx_prop_key, s, private=True, sticky=False)
+
+        # Explicitly delete local reference to aid garbage collection
+        del tensors
+        del self.tensor_events[task_id]
```
YuanTingHsieh: could there be racing conditions here?

nvflare/app_opt/tensor_stream/client.py
```diff
@@ -0,0 +1,157 @@
...
+        handle_event(event_type, fl_ctx): Handles events for the TensorSender component.
+        request_task_data_tensors(fl_ctx): Sends fed event to request tensors from the server.
+        send_tensors_to_server(fl_ctx): Sends tensors to the server before sending the task result.
+    """
+
+    def __init__(
+        self,
+        format: ExchangeFormat = ExchangeFormat.PYTORCH,
+        tasks: list[str] = None,
+        enable_request_task_data_tensors: bool = False,
```
YuanTingHsieh: when is this true?
whats the use case?

nvflare/app_opt/tensor_stream/receiver.py
```diff
@@ -0,0 +1,199 @@
+# Copyright (c) 2022, NVIDIA CORPORATION.  All rights reserved.
```
YuanTingHsieh: 2025

nvflare/app_opt/tensor_stream/receiver.py
```diff
@@ -0,0 +1,199 @@
...
+        # Explicitly delete local reference to aid garbage collection
+        del tensors
+        del self.tensor_events[task_id]
+
+        self.logger.info(
+            f"Peer '{fl_ctx.get_identity_name()}': updated task data with tensors received from peer "
+            f"'{peer_name}'. Task ID: '{task_id}'."
+        )
+
+    def wait_for_tensors(self, task_id: str, peer_name: str, timeout: float = 5.0):
```
YuanTingHsieh, line 173: is this waiting for the whole Tensor or just one chunk?
is 5 seconds enough?

nvflare/app_opt/tensor_stream/server.py
```diff
@@ -0,0 +1,235 @@
...
+        self.engine = engine
+        try:
+            self.receiver = TensorReceiver(engine, FLContextKey.TASK_RESULT, self.format)
+            self.sender = TensorSender(engine, FLContextKey.TASK_DATA, self.format, self.tasks)
+        except Exception as e:
+            self.system_panic(str(e), fl_ctx)
+            return
+
+    def handle_event(self, event_type: str, fl_ctx: FLContext):
+        """Handle events for the TensorSender component.
```
YuanTingHsieh: docstring -> TensorServerStreamer

nvflare/app_opt/tensor_stream/server.py
```diff
@@ -0,0 +1,235 @@
...
+        self.wait_task_data_sent_to_all_clients_timeout = wait_send_task_data_all_clients_timeout
+        self.engine: StreamableEngine = None
+        self.sender: TensorSender = None
+        self.receiver: TensorReceiver = None
+        self.start_sending_time: dict[int, float] = defaultdict(float)
+        self.seen_tasks: dict[int, set[str]] = defaultdict(set)
+        self.num_task_data_sent: dict[int, int] = defaultdict(int)
+        self.num_task_skipped: dict[int, int] = defaultdict(int)
+        self.data_cleaned: dict[int, bool] = defaultdict(bool)
+        self.lock = Lock()
```
YuanTingHsieh, line 86: do we need the lock in this component?

YuanTingHsieh (commented):
Great PR, it adds a very good functionality and this PR is clear and easy to follow with an example and tests.

It would be even greater to show some memory diagrams to see the improvements.

Added some comments.

examples/advanced/tensor-stream/trainer.py
rfilgueiras: This is used to train the model without federation.

nvflare/app_opt/tensor_stream/client.py
rfilgueiras: Server-side handling is missing. Removed for now.

nvflare/app_opt/tensor_stream/producer.py
rfilgueiras: Yes, individual tensors are not chunked. I initially assumed that would be needed, but it turns out to be unnecessary, as we aim to include at least 10 tensors per chunk to maintain good throughput.

nvflare/app_opt/tensor_stream/receiver.py
rfilgueiras: Yes, this can happen on the server side. A lock isn’t needed because each call to this method uses a different task_id.

nvflare/app_opt/tensor_stream/receiver.py
rfilgueiras: No, each call to this method uses a different task_id.

nvflare/app_opt/tensor_stream/utils.py
rfilgueiras: Fixed.

nvflare/app_opt/tensor_stream/client.py
rfilgueiras: The use case is not fully implemented. It was intended to handle cases where the server sends tensors after the task arrives, but I’ve removed that functionality since it has no implementation on the server side.

nvflare/app_opt/tensor_stream/receiver.py
rfilgueiras: The server sends all tensors first, followed by the task. When the client receives the task, it looks for the corresponding tensors. In some cases, the tensors may not yet be updated on the receiver. This adds a short wait to handle that delay—five seconds should be sufficient.

nvflare/app_opt/tensor_stream/server.py
rfilgueiras: Yes, we do. It ensure consitence when multiple threads are trying to update the same attributes.

rfilgueiras:
> Great PR, it adds a very good functionality and this PR is clear and easy to follow with an example and tests.
> 
> It would be even greater to show some memory diagrams to see the improvements.

I already have memory charts, but they come from a different application.  
If time allows, I can generate new charts to compare memory usage with and without the tensor stream.  
I plan to run the example using Docker provisioning with at least five clients.

> Added some comments.

nvflare/app_opt/tensor_stream/receiver.py
YuanTingHsieh: ok so this five seconds is between client receive task and client receive tensors.
it might need more when network is bad, but I am ok with this default.
do we need to clarify this in the docstring?

YuanTingHsieh (approved):
Thanks! LGTM

nvflare/app_opt/tensor_stream/receiver.py
rfilgueiras: The tensors are always arriving before the task.
But there is a small time to sync from the consumer to the receiver.

When the task is transferred quickly, since it has no data, the consumer may not have copied the tensors back to the receiver, which should be a matter of milliseconds.

rfilgueiras: Added a clear explanation to the docstring.

YuanTingHsieh: got it thanks for the explanation

nvflare/app_opt/tensor_stream/receiver.py
```diff
@@ -0,0 +1,206 @@
...
+            raise ValueError(f"No task_id found from peer {peer_name}.")
+
+        with self.lock:
+            if task_id not in self.tensor_events:
+                self.tensor_events[task_id] = threading.Event()
+
+        if not success:
+            with self.lock:
+                self.tensor_events[task_id].set()  # Wake waiting threads
+            raise ValueError(f"Failed to receive tensors from peer '{peer_name}' and task '{task_id}'.")
```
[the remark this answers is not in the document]
rfilgueiras: Fixed, see last commit.

examples/advanced/tensor-stream/README.md
```diff
@@ -0,0 +1,391 @@
...
+from nvflare.client.config import ExchangeFormat
+
+# Simple configuration
+client_streamer = TensorClientStreamer()
+
+# With custom format, tasks, and timeout
+client_streamer = TensorClientStreamer(
+    format=ExchangeFormat.PYTORCH,
+    tasks=["train", "validate"],
+    entry_timeout=60.0,  # Increase timeout for large models
```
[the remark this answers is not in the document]
rfilgueiras: Fixed, see last commit.

examples/advanced/tensor-stream/README.md
```diff
@@ -0,0 +1,391 @@
...
+  - The format of the tensors to send and receive.
+  - Must match the format used by `TensorClientStreamer` on the client side.
+  - Determines how tensors are serialized and deserialized during transmission.
+
+- **`tasks`** (`list[str]`, default: `["train"]`)
+  - The list of task names for which tensors should be streamed.
+  - When set to `None`, defaults to `["train"]`.
+  - Should match the tasks configured in `TensorClientStreamer`.
+
+- **`entry_timeout`** (`float`, default: `30.0`)
```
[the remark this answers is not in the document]
rfilgueiras: Fixed, see last commit.

examples/advanced/tensor-stream/README.md
```diff
@@ -0,0 +1,391 @@
...
+    wait_send_task_data_all_clients_timeout=300.0
+)
+```
+
+**Large models (> 1GB), moderate network:**
+```python
+client_streamer = TensorClientStreamer(
+    entry_timeout=90.0,
+    wait_for_task_data_tensors_timeout=180.0
+)
```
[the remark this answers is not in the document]
rfilgueiras: Fixed, see last commit.

examples/advanced/tensor-stream/README.md
```diff
@@ -0,0 +1,391 @@
...
+server_streamer = TensorServerStreamer(
+    entry_timeout=90.0,
+    wait_send_task_data_all_clients_timeout=900.0  # 15 minutes
```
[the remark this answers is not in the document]
rfilgueiras: Fixed, see last commit.

examples/advanced/tensor-stream/README.md
```diff
@@ -0,0 +1,391 @@
...
+)
+```
+
+**Very large models (> 10GB), many clients:**
+```python
+client_streamer = TensorClientStreamer(
+    entry_timeout=120.0,
+    wait_for_task_data_tensors_timeout=300.0
+)
```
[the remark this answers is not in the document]
rfilgueiras: Fixed, see last commit.

examples/advanced/tensor-stream/README.md
```diff
@@ -0,0 +1,391 @@
...
+server_streamer = TensorServerStreamer(
+    entry_timeout=120.0,
```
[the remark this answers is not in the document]
rfilgueiras: Fixed, see last commit.

nvflare/app_opt/tensor_stream/receiver.py
```diff
@@ -0,0 +1,205 @@
...
```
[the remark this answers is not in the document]
rfilgueiras: Fixed, see last commit.

PR #3745, opened 2025-10-02
Fix filter race issue

ZiyueXu77:
Fixes double filtering issue when client pull task timeout.

### Description

Our filtering happen outside the workflow lock, but the task assignment itself happens inside the workflow lock. This creates a race when client pull task timeout and re-try:
- Thread 1: Client requests task → gets task assignment → releases workflow lock → starts filtering (slow operation like quantization)
- Thread 2: Same client requests task again → acquires workflow lock → gets the same task assignment → starts filtering again (double filtering causing uint8 error)

Fix by adding processing tracker in ServerRunner, which will mark the task being processed and prevent duplicate assignment.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/private/fed/server/server_runner.py
+++ b/nvflare/private/fed/server/server_runner.py
@@ -105,6 +105,7 @@ class ServerRunner(TBI):
         self.turn_to_cold = False
         # track tasks currently being processed (during filtering) to prevent duplicate assignments
         self._processing_tasks = {}  # client_name => task_id
+        self._processing_tasks_lock = threading.Lock()  # protect _processing_tasks from race conditions
         self._register_aux_message_handler(engine)
 
     def _register_aux_message_handler(self, engine):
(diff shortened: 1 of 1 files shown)


nvflare/private/fed/server/server_runner.py
```diff
@@ -306,6 +316,9 @@ def process_task_request(self, client: Client, fl_ctx: FLContext) -> (str, str,
             if not task_name or task_name == SpecialTaskName.TRY_AGAIN:
                 return self._task_try_again()
 
+            # Mark this task as being processed to prevent duplicate assignments
+            self._processing_tasks[client.name] = task_id
```
YuanTingHsieh: there could be a racing issue for processing_tasks, suggest add a lock to prevent that

YuanTingHsieh (commented):
Good catch and fix!

nvflare/private/fed/server/server_runner.py
ZiyueXu77: good point! added

PR #3748, opened 2025-10-03
Fix admin user name in poc mode

YuanTingHsieh:
Follow #3519 and #3735

(1) the key name is "username" in the fed_admin.json instead of "user_name"
(2) we were hardcoded an username of "admin" instead of getting it from the config

### Description

- Change codes to be consistent
- Change hardcoded "admin" username

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/examples/hello-world/hello-flower/requirements.txt
+++ b/examples/hello-world/hello-flower/requirements.txt
@@ -1,5 +1,5 @@
 flwr[simulation]>=1.16,<2.0
-nvflare>=2.6.0
+nvflare~=2.7.0rc
 torch
 torchvision
 tensorboard
--- a/examples/hello-world/step-by-step/cifar10/stats/image_stats_job.py
+++ b/examples/hello-world/step-by-step/cifar10/stats/image_stats_job.py
@@ -45,7 +45,7 @@ def main():
     stats_generator = ImageStatistics(data_root_dir)
 
     job = StatsJob(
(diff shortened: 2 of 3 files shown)


nvflare/fuel/hci/client/api_spec.py
```diff
@@ -230,7 +230,7 @@ class AdminConfigKey:
     UID_SOURCE = "uid_source"  # where does user id come from
     UPLOAD_DIR = "upload_dir"
     DOWNLOAD_DIR = "download_dir"
-    USERNAME = "username"
+    USERNAME = "user_name"
```
chesterxgchen: why do we have to change this "user_name" ?

nvflare/fuel/hci/client/cli.py
```diff
@@ -75,20 +75,19 @@ def __init__(
...
+        user_name: str = "",
         handlers=None,
         cli_history_dir: str = str(Path.home() / ".nvflare"),
         cli_history_size: int = 1000,
     ):
         super().__init__()
         self.intro = "Type help or ? to list commands.\n"
         self.prompt = admin_config.get(AdminConfigKey.PROMPT, "> ")
-        self.user_name = "admin"
+        self.user_name = user_name
```
chesterxgchen: so this is the actual bug, so we make all other changes ?

PR #3751, opened 2025-10-06
Reduced Memory Footprint

nvidianz:
### Description
1. Reduced memory footprint by removed data after event.
2. Changed TensorDecomposer to use file-streaming for tensors larger than 10M.
3. Removed REPLY from FLContext, just use TASK_RESULT.
4. Removed large data from several log messages. 

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/apis/fl_component.py
+++ b/nvflare/apis/fl_component.py
@@ -11,7 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
-from typing import List, Union
+from typing import Any, List, Union
 
 from nvflare.apis.utils.fl_context_utils import generate_log_message
 from nvflare.fuel.utils.log_utils import get_obj_logger
@@ -82,6 +82,16 @@ class FLComponent(StatePersistable):
 
         fl_ctx.set_prop(FLContextKey.EVENT_SCOPE, value=EventScope.LOCAL, private=True, sticky=False)
         self._fire(event_type, fl_ctx)
+
+    def fire_event_with_data(self, event_type: str, fl_ctx: FLContext, key: str, data: Any):
+        """
+        Set the data for the event and clean it up afterward
+        """
+        try:
+            fl_ctx.set_prop(key=key, value=data, private=True, sticky=False)
+            self.fire_event(event_type, fl_ctx)
+        finally:
+            fl_ctx.set_prop(key=key, value=None, private=True, sticky=False)
 
     def fire_fed_event(self, event_type: str, event_data: Shareable, fl_ctx: FLContext, targets=None):
         """Fires a federation event.
--- a/nvflare/apis/impl/task_controller.py
+++ b/nvflare/apis/impl/task_controller.py
@@ -87,11 +87,7 @@ class TaskController(FLComponent, ControllerSpec):
         # apply task filters
 
         self.log_debug(fl_ctx, "firing event EventType.BEFORE_TASK_DATA_FILTER")
-        try:
-            fl_ctx.set_prop(FLContextKey.TASK_DATA, task.data, sticky=False, private=True)
-            self.fire_event(EventType.BEFORE_TASK_DATA_FILTER, fl_ctx)
-        finally:
-            fl_ctx.set_prop(FLContextKey.TASK_DATA, None, sticky=False, private=True)
+        self.fire_event_with_data(EventType.BEFORE_TASK_DATA_FILTER, fl_ctx, FLContextKey.TASK_DATA, task.data)
 
         # # first apply privacy-defined filters
         try:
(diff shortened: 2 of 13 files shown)


nvflare/apis/impl/task_controller.py
```diff
@@ -85,9 +85,13 @@ def broadcast_and_wait(
...
         # apply task filters
+
         self.log_debug(fl_ctx, "firing event EventType.BEFORE_TASK_DATA_FILTER")
-        fl_ctx.set_prop(FLContextKey.TASK_DATA, task.data, sticky=False, private=True)
-        self.fire_event(EventType.BEFORE_TASK_DATA_FILTER, fl_ctx)
+        try:
+            fl_ctx.set_prop(FLContextKey.TASK_DATA, task.data, sticky=False, private=True)
+            self.fire_event(EventType.BEFORE_TASK_DATA_FILTER, fl_ctx)
+        finally:
+            fl_ctx.set_prop(FLContextKey.TASK_DATA, None, sticky=False, private=True)
```
chesterxgchen: does fire_event make a copy of the task data or simply reference to it ? if you set the None on fl_ctx, are you sure this won't make task_data none before the listener of the event retrive it ?

nvflare/apis/utils/fl_context_utils.py
```diff
@@ -97,7 +97,7 @@ def generate_log_message(fl_ctx: FLContext, msg: str):
             peer_name = "?"
         all_kvs[_peer_name] = peer_name
 
-    reply = fl_ctx.get_prop(FLContextKey.REPLY, None)
+    reply = fl_ctx.get_prop(FLContextKey.TASK_RESULT, None)
```
chesterxgchen, line 100: is this duplicate REPLY and TASK_RESULT ?  any client using REPLY ?

nvflare/edge/controllers/sage.py
```diff
@@ -280,8 +283,11 @@ def handle_event(self, event_type: str, fl_ctx: FLContext):
                 )
 
     def _prepare_train_task_data(self, client_task: ClientTask, fl_ctx: FLContext) -> None:
-        fl_ctx.set_prop(AppConstants.TRAIN_SHAREABLE, client_task.task.data, private=True, sticky=False)
-        self.fire_event(AppEventType.BEFORE_TRAIN_TASK, fl_ctx)
+        try:
+            fl_ctx.set_prop(AppConstants.TRAIN_SHAREABLE, client_task.task.data, private=True, sticky=False)
```
chesterxgchen: I see this pattern over and over again, should you wrap this into a function. We repeated this for a 10-20 times.

nvflare/apis/impl/task_controller.py
nvidianz: It simply keeps a reference. The event handlers are synchronous. Once it returns from file_event(), all the handlers are already called.

nvflare/apis/utils/fl_context_utils.py
nvidianz: All code-based are searched. This is only used in logging. There is no reason to keep two copies of the same thing.

nvidianz: All code base is searched. This is only used in logging. There is no reason to keep two copies of the same thing.

nvflare/apis/impl/task_controller.py
chesterxgchen: Then there is a race condition on who get the data first, if the event handler took longer to process the task data, then we could get into a situation where the task data is removed while the event handler is still processing.

nvflare/edge/controllers/sage.py
nvidianz: Great suggestion. Added fire_event_with_data().

nvflare/apis/fl_component.py
```diff
@@ -83,6 +83,16 @@ def fire_event(self, event_type: str, fl_ctx: FLContext):
         fl_ctx.set_prop(FLContextKey.EVENT_SCOPE, value=EventScope.LOCAL, private=True, sticky=False)
         self._fire(event_type, fl_ctx)
 
+    def fire_event_with_data(self, event_type: str, fl_ctx: FLContext, key: str, data: Any):
+        """
+        Set the data for the event and clean it up afterward
+        """
+        try:
+            fl_ctx.set_prop(key=key, value=data, private=True, sticky=False)
+            self.fire_event(event_type, fl_ctx)
```
chesterxgchen, line 92: @nvidianz  please add comment on L92 to explain that the fire_event() is synchronized, by the time it turns all event handlers are completed.

nvidianz:
> 1. Not clear if you can remove the fl context property after the event is fired but may not consumed.
> 2. if above is ok, the same type of try set property, fire event, finally clear property is repeated for many times. we should wrap it up into a function
> 3. Are you sure the REPLY and TASK_RESULT the same and no one use REPLY ?
> 4. Do you have a quantity measure how much memory is reduced from before. see used to 100GB --> X GB ? after the change

1. The FLContext is used throughout the flow, can't be removed.
2. It's wrapped into a function fire_event_with_data now.
3. Yes. Only one place using REPLY (in the logging) and I changed it into TASK_RESULT.
4. It used to keep 5 copies of model, now only 2.

PR #3756, opened 2025-10-07
Update CC docs [skip ci]

YuanTingHsieh:
Update CC docs

### Description

Update CC docs

### Types of changes

- [ ] Documentation updated.


docs/user_guide/confidential_computing/attestation.rst
```diff
@@ -4,99 +4,76 @@
...
-
+In NVFlare, data is encrypted during transmission between participants
+—including the server, clients, and admin—to ensure strong protection
+for data in transit. To secure data at rest, users can leverage existing
+infrastructure such as storage encryption.
+
+However, traditional security approaches leave a critical gap: **data in use**.
+This is where confidential computing (CC) plays a key role. CC enables NVFlare
+to protect data while it is being processed, completing full-lifecycle security:
+at rest, in transit, and in use.
```
chesterxgchen: These are normal CC talks, for Federated Learning, we also need to address the risk at deployment tampering.

docs/user_guide/confidential_computing/attestation.rst
```diff
@@ -4,99 +4,76 @@
...
+
+Confidential computing in NVFlare is also central to establishing trust between
```
chesterxgchen: Not sure this section is clear stated our position, how FLARE solution is different. 

There are risk at deployment, and runtime, where confidential computing can be leveraged to mitigate risks.  
We need to explain how to build trust via CC attestation to verify the evidence. The follwing paragraph seems to decribe the attestation without mention attesation

docs/user_guide/confidential_computing/attestation.rst
```diff
@@ -4,99 +4,76 @@
...
+participants in a federated learning job. Before participating in a job, each
+participant must generate a set of cryptographically verifiable evidence about
+its computing environment—including details about hardware (e.g., GPU),
+software (e.g., GPU driver and VBIOS), and other platform components.
+
+This evidence is validated, signed, and packaged into a Confidential Computing
+token (CC token). The participant can then present its CC token to others to
+prove the integrity and trustworthiness of its environment.
+
+When a participant (the "relying party") receives a CC token, it verifies the
```
chesterxgchen: not sure we need to introduce the "relying party" concept in our document unless you directly referring the RATS architecture

docs/user_guide/confidential_computing/attestation.rst
YuanTingHsieh: got it, we talk about these in the CC architecture, I will just add a reference here instead of duplicating the words.

docs/user_guide/confidential_computing/attestation.rst
YuanTingHsieh: got it, we talk about these in the CC architecture, I will just add a reference here instead of duplicating the words.

docs/user_guide/confidential_computing/attestation.rst
YuanTingHsieh: got it, remove all "relying party"

docs/user_guide/confidential_computing/cc_deployment_guide.rst
```diff
@@ -14,7 +14,7 @@ Hardware Prerequisites
 ^^^^^^^^^^^^^^^^^^^^^^
 
 - The machine that needs to run CC needs to have AMD SNP-SEV enabled CPU.
-- To run GPU, the machine needs to have a “production” board of NVIDIA GPU that has CC-enabled.
+- To run GPU, the machine needs to have a “production” board of NVIDIA GPU that is CC-enabled.
```
chesterxgchen, line 17: This is for end-user not for internal development, for customer every GPU is production. we should emphasize the product board

PR #3771, opened 2025-10-10
Fix Conftest.py

chesterxgchen:
Make a mistake on last commit with conftest.py 
I assume the notebook only read from disk once. 

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -20,6 +20,10 @@ try:
 except Exception:
     nbformat = None
 
+_executed_notebooks = set()
+_passed_notebooks = set()
+_backup_notebooks = set()  # Track notebooks with backup files
+
 
 def pytest_collection_modifyitems(config, items):
     """Filter tagged cells from notebooks before nbmake runs"""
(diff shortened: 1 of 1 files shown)


conftest.py
```diff
@@ -98,6 +121,23 @@ def is_kernel_registered(kernel_name):
...
         return False
 
+def restore_notebooks():
+    """Restore original notebooks from .backup files"""
+    if not _backup_notebooks:
+        return
+    
+    restored = []
+    for notebook_path in _backup_notebooks:
+        backup_path = Path(notebook_path).with_suffix(Path(notebook_path).suffix + '.backup')
```
IsaacYangSLA: notebook_path is an instance of Path (see LINE 101 and 92).  Why do we need to do Path(notebook_path)?

conftest.py
```diff
@@ -115,6 +155,7 @@ def pytest_addoption(parser):
 
 _executed_notebooks = set()
 _passed_notebooks = set()
+_backup_notebooks = set()  # Track notebooks with backup files
```
YuanTingHsieh: Suggest move these global variables in the top for better readability

chesterxgchen: changed

conftest.py
chesterxgchen: changed

PR #3780, opened 2025-10-13
Add huggingface example to hierarchical 

ZiyueXu77:
Fixes # .

### Description

Add one example to hierarchical fl for huggingface language model training, addressed the latency issue by removing tolist() process 

### Types of changes

- [ ] Documentation updated.


chesterxgchen (commented):
For newly added examples. 
Please following the following convention
1) no need to add src directory, just keep it flat. 
2)  for the model file, no need to name something like hf_sft_model.py, you are already in the given example or particular directory. just called model.py
3) the client side training script should be called client.py 
4) the job code should be called job.py ( instead of hf_sft.job
5) try to use one folder for one job, not to share the code. It may looks efficient by sharing the code for the example creator, it is confusing for the user trying to figure out which one relevant. The example is not provide the simplest way for the end user to understand the concept. 

For all new examples, should ideally have the code structure like this 
```
client.py 
server.py ( may be missing for builtin FL algorithms) 
model.py 
job.py 

download_data.py ( optional) 
preapre_data.py (optional) 

```
Unless to make it consistent, its very hard to do automated testing. and confusing to user as well. 

We are too freely name our files, such as cifar10_fl_job.py, fl_job.py, cifar10_fedavt_train.py etc. 

Lets be consistent.

nvflare/edge/assessors/buff_model_manager.py
```diff
@@ -143,10 +143,15 @@ def generate_new_model(self, fl_ctx: FLContext) -> None:
...
+        # print model size in MB
+        model_size = sum([v.nbytes for v in new_model.values()]) / (1024 * 1024)
+        self.log_info(fl_ctx, f"new model size: {model_size:.2f} MB")
 
         # update the current model
-        # convert new_model items from numpy arrays to lists for serialization
-        new_model = {k: v.tolist() if isinstance(v, np.ndarray) else v for k, v in new_model.items()}
+        # Keep numpy arrays for efficient FOBS serialization through the hierarchy
+        # If converting to list, the model size will be much larger and slower to serialize
+        # Note that for cases where the device expects list (e.g. ExecuTorch simulation),
```
YuanTingHsieh, line 153: this could break the current ET examples with mobile, we maybe merge this PR after release.

examples/advanced/llm_hf/src/hf_sft_model.py
```diff
@@ -19,6 +19,7 @@
 class CausalLMModel(torch.nn.Module):
```
chesterxgchen, line 19: why this class under src and why it can't be named model.py ? 

Similar to other tutorial notebooks. is this because the video ?

nvflare/edge/assessors/buff_model_manager.py
ZiyueXu77: good point! let's do this PR after release to avoid any broken pieces

examples/advanced/llm_hf/src/hf_sft_model.py
ZiyueXu77: it definitely can, I was just avoiding making too much change in one PR. Shall we do a holistic move to update all examples following the new convention (after release)?

ZiyueXu77:
close for now, will add after CollabAPI

PR #3783, opened 2025-10-13
Do not generate start_all.sh and use async grpc by default

nvidianz:
Fixes https://jirasw.nvidia.com/browse/FLARE-2677

### Description

Made 2 changes to provisioning,

1. By default, it will not generate start_all.sh. Use -s option to generate it.
2. Use synch grpc driver for both client and server by default.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/app_opt/confidential_computing/cc_manager.py
+++ b/nvflare/app_opt/confidential_computing/cc_manager.py
@@ -183,8 +183,8 @@ class CCManager(FLComponent):
             self.cc_verifiers[namespace] = verifier
 
     def _prepare_cc_info(self, fl_ctx: FLContext):
-        # client side: if token expired then generate a new one
-        self._refresh_expired_tokens()
+        # client side
+        self._ensure_fresh_tokens(force=True)
 
         if not self.token_submitted:
             site_cc_info = self.participant_cc_info[self.site_name]
@@ -217,6 +217,7 @@ class CCManager(FLComponent):
                 for _, client in job_participants.items():
                     participants.append(client.name)
 
+                self._ensure_fresh_tokens()
                 participants_tokens = self._collect_participants_tokens(participants)
                 err = self._validate_participants_tokens(participants_tokens)
                 if err:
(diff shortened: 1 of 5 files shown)


nvflare/lighter/provision.py
```diff
@@ -122,12 +123,14 @@ def provision_for_edge(params, project_dict):
 
 
 def provision(
+    args,
     project_full_path: str,
```
YuanTingHsieh, line 127: can you double check if this is the only place using provision?
what about dashboard and other places, what function do we use?

nvflare/lighter/provision.py
```diff
@@ -62,6 +62,7 @@ def define_provision_parser(parser):
     parser.add_argument("-c", "--custom_folder", type=str, default=".", help="additional folder to load python codes")
     parser.add_argument("--add_user", type=str, default="", help="yaml file for added user")
     parser.add_argument("--add_client", type=str, default="", help="yaml file for added client")
+    parser.add_argument("-s", "--script", action="store_true", help="generate test scripts like start_all.sh")
```
YuanTingHsieh: a suggestion not related to this PR, 

maybe our "--edge" should be called ""--generate-edge" to be consistent with "--generate"

chesterxgchen: should be called "-g" "--gen_scripts" so it consistent with the project property

also --gen_scripts instead gen_script, with "s" at the end to give room to generate more than one script in the future ?

nvflare/lighter/provision.py
```diff
@@ -122,12 +123,14 @@ def provision_for_edge(params, project_dict):
 
 
 def provision(
+    args,
```
chesterxgchen, line 126: This may break POC and other code, POC uses provision function internally. please double check

nvflare/lighter/provision.py
nvidianz: -g is already used to generate sample project.yml

nvidianz: --edge is more consistent with -e.

nvflare/lighter/provision.py
nvidianz: Good suggestion. Checked POC and it works. It uses the provisioner.provision().

nvflare/lighter/provision.py
nvidianz: I did a code search. Unless it's not in our code tree, this is the only place it's used.

nvflare/lighter/provision.py
YuanTingHsieh: guess its subjective, but we can discuss that in team channel if we want, even stay the current way is OK for me.

nvidianz:
> I don't think the "--script" is correct one. Even -g is used we can use others it should be "gen_scripts" which match to what do used internally.
> 
> But I will approve this for now.

I try to make the short form consistent with the long form. But this is internal argument, I changed it as you suggested.

PR #3812, opened 2025-10-24
Fix webpage links

nvkevlu:
Fix broken links and issues for the web page.

Fix website link consistency: Replace hardcoded documentation paths with version-aware template literals

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/web/src/pages/security.astro
+++ b/web/src/pages/security.astro
@@ -4,6 +4,7 @@ import DataPrivacy from '../images/data_privacy_arch.png'
 import FedAuth from '../images/fed_auth_arch.png'
 
 const base_url = import.meta.env.BASE_URL;
+const gh_branch = import.meta.env.PUBLIC_GH_BRANCH;
 
 ---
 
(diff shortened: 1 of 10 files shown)


web/src/pages/security.astro
```diff
@@ -4,6 +4,7 @@ import DataPrivacy from '../images/data_privacy_arch.png'
 import FedAuth from '../images/fed_auth_arch.png'
 
 const base_url = import.meta.env.BASE_URL;
+const gh_branch = import.meta.env.PUBLIC_GH_BRANCH;
```
YuanTingHsieh, line 7: so this "import.meta.env" will have the github branch?

another question, this change is good, but do we have a use case to have more than 1 version of web page?

chesterxgchen: I think so, this is similar to the documentation with different versions.

nvkevlu: this was already the way it was and gh_branch exists for most pages already, this PR just makes it all consistently handled everywhere

YuanTingHsieh: sounds good thanks

PR #3814, opened 2025-10-28
Add allow self signed to edge simulator

YuanTingHsieh:
Add a way for users to use self signed certificates if they prefer.

### Description

Add a way for users to use self signed certificates if they prefer.

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh:
close for now

PR #3821, opened 2025-11-04
Update sub_start.sh with one additional option

IsaacYangSLA:
### Description

Add one more option `--once` to sub_start.sh.  When this option is used with sub_start.sh, like `sub_start.sh --once` it will go directly to start NVFlare.  If it fails, such as missing dependencies or other issues, sub_start.sh exits with the exit code returned by python.

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh (approved):
LGTM

PR #3835, opened 2025-11-13
Add SklearnFedAvgRecipe and convert sklearn-linear example to recipe format

nvkevlu:
Add SklearnFedAvgRecipe and convert sklearn-linear example to recipe format

Fixes # .

### Description

Add SklearnFedAvgRecipe and convert sklearn-linear example to recipe format.

### Types of changes

- [ ] Documentation updated.


examples/advanced/sklearn-linear/job.py
```diff
@@ -0,0 +1,84 @@
...
+    parser.add_argument(
+        "--data_path",
+        type=str,
+        default="/tmp/nvflare/dataset/HIGGS.csv",
+        help="Path to HIGGS dataset CSV file",
+    )
+    parser.add_argument("--train_start", type=int, default=700, help="Training data start index")
+    parser.add_argument("--train_end", type=int, default=1000, help="Training data end index")
+    parser.add_argument("--valid_start", type=int, default=0, help="Validation data start index")
+    parser.add_argument("--valid_end", type=int, default=200, help="Validation data end index")
```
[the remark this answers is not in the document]
ZiyueXu77: Or make these optional, use client.py defaults if not provided. As long as the params and the curve align

nvflare/app_opt/sklearn/__init__.py
```diff
@@ -1,4 +1,4 @@
-# Copyright (c) 2023, NVIDIA CORPORATION.  All rights reserved.
+# Copyright (c) 2025, NVIDIA CORPORATION.  All rights reserved.
```
ZiyueXu77: I think we keep the original date right @holgerroth ?

ZiyueXu77 (commented):
Is this specific for linear regression? If so, we may rename it to be more specific.

nvflare/app_opt/sklearn/recipes/fedavg.py
```diff
@@ -0,0 +1,187 @@
...
+        The custom aggregator must be a subclass of the Aggregator class.
+    """
+
+    def __init__(
+        self,
+        *,
+        name: str = "sklearn_fedavg",
+        min_clients: int,
+        num_rounds: int = 2,
+        initial_params: Optional[dict] = None,
```
YuanTingHsieh: should we call this SGD params or something?
SKSGD params?
seems this matches all the arguments to be passed into the SGDClassifier

examples/advanced/sklearn-linear/job.py
```diff
@@ -0,0 +1,80 @@
...
+        initial_params={
+            "n_classes": 2,
+            "learning_rate": "constant",
+            "eta0": 1e-4,
+            "loss": "log_loss",
+            "penalty": "l2",
+            "fit_intercept": 1,
+        },
+        train_script="client.py",
+        train_args=f"--data_path {data_path}",
```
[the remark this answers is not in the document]
YuanTingHsieh: ^ yeah previously there is a way to use the prepare_job_config to generate different ranges for each client, I guess we need to have a way to do that as well in the new recipe style.
maybe include some instructions in readme.

examples/advanced/sklearn-linear/prepare_job_config.sh
```diff
@@ -1,24 +0,0 @@
...
-
-valid_frac=0.1
-echo "Generating job configs with data splits, reading from ${DATASET_PATH}"
-
-task_name="sklearn_linear"
-for site_num in 5;
-do
-    for split_mode in uniform;
-    do
-        python3 utils/prepare_job_config.py \
```
YuanTingHsieh, line 17: we might not need this file anymore?

examples/advanced/sklearn-linear/job.py
```diff
@@ -0,0 +1,80 @@
...
+# limitations under the License.
+
+"""
+This code shows how to use NVIDIA FLARE Job Recipe to run federated learning with scikit-learn
+linear models using FedAvg algorithm.
+"""
+
+import argparse
+
+from nvflare.app_opt.sklearn.recipes.fedavg import SklearnFedAvgRecipe
```
YuanTingHsieh: if we already added those imports/exports in the init file,
here we should be able to "from nvflare.app_opt.sklearn import SklearnFedAvgRecipe"

YuanTingHsieh (commented):

nvflare/app_opt/sklearn/recipes/fedavg.py
YuanTingHsieh: another way is to pop all these args to the very top level, like what you did in Kmeans (the n_cluster)

nvkevlu: there are too many to pop like the other one, renaming this to model params with better docstring

examples/advanced/sklearn-linear/job.py
```diff
@@ -0,0 +1,142 @@
...
+        default="uniform",
+        choices=["uniform", "custom"],
+        help="Data split method: 'uniform' (same args for all) or 'custom' (per-client ranges)",
+    )
+
+    return parser.parse_args()
+
+
+def calculate_data_splits(n_clients: int, total_size: int = 11000000, valid_size: int = 1100000):
+    """Calculate uniform data splits for clients.
```
YuanTingHsieh: is our "custom" and "uniform" right now actually the same strategy? both mean "uniform"

in that case should we just dont say split method,
and ask Users to implement this calculate_data_splits if they need custom

YuanTingHsieh (commented):
have one comment, good information on how to do data splits

YuanTingHsieh (approved):

PR #3838, opened 2025-11-13
Add XGBBaggingRecipe and convert random_forest example to recipe format

nvkevlu:
Add XGBBaggingRecipe and convert random_forest example to recipe format.

### Description

Add XGBBaggingRecipe and convert random_forest example to recipe format.

### Types of changes

- [ ] Documentation updated.


ZiyueXu77 (commented):
LGTM

examples/advanced/random_forest/job.py
```diff
@@ -0,0 +1,113 @@
...
+        nthread=args.nthread,
+        lr_mode=args.lr_mode,
+        save_name="xgboost_model.json",
+        data_loader_id="dataloader",
+    )
+
+    # Add executor and data loader to each client
+    for site_id in range(1, args.n_clients + 1):
+        data_split_file = f"{args.data_split_path}/data_site-{site_id}.json"
+        dataloader = HIGGSDataLoader(data_split_filename=data_split_file)
```
YuanTingHsieh, line 91: some enhancement we could do later is make this data loader more realistic, 
you won't usually have several sites sharing same data file but use index file to load part of that data
@ZiyueXu77

nvflare/app_opt/xgboost/recipes/bagging.py
```diff
@@ -0,0 +1,200 @@
...
+        
+        Args:
+            site_name: Name of the client site (e.g., "site-1")
+            dataloader: XGBDataLoader instance for this client
+            lr_scale: Learning rate scale factor for this client (default: 1.0)
+        """
+        from nvflare.app_opt.xgboost.tree_based.executor import FedXGBTreeExecutor
+        
+        # Create executor for this specific client
+        executor = FedXGBTreeExecutor(
```
YuanTingHsieh, line 215: is it possible we write a client.py that use client api as well?
@ZiyueXu77 @nvkevlu

examples/advanced/random_forest/job.py
ZiyueXu77: I think it is fine because in the data split file we have both data path and data index. 

Some further polish can be added, in reality, sites can have:
- same data file, but different index: same as here and one of the LLM federated training study actually follows this setting (everyone have the same data, but optimizing on different samples)
- different data file, but all index: this can also work, just replace the data path and not providing data index - defaulting to all data
- different data file, and different index: this can support such dual-spec for data by specifying both, and it will give sites another layer of data control - they can specify partial data without needing to carve out and save a new file

We may not need a json though, these can be just args

nvflare/app_opt/xgboost/recipes/bagging.py
```diff
@@ -0,0 +1,200 @@
...
+        Recipe.__init__(self, self.job)
+
+    @validator("local_subsample")
+    def check_subsample(cls, v):
+        if not 0 < v <= 1.0:
+            raise ValueError("local_subsample must be between 0 and 1")
+        return v
+
+    @validator("lr_mode")
+    def check_lr_mode(cls, v):
```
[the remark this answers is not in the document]
nvkevlu: Pydantic v2 requires @field_validator to be used with @classmethod, so it must use cls as the first parameter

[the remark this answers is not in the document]
nvkevlu: Pydantic v2 requires @field_validator to be used with @classmethod, so it must use cls as the first parameter

nvflare/app_opt/xgboost/recipes/bagging.py
nvkevlu: Theoretically yes, but it would require:

- Creating a new executor that wraps the Client API
- Converting DMatrix to/from parameters
- Handling XGBoost's tree serialization differently
- Potentially losing some of XGBoost's native federated optimizations

This would be a significant architectural change and might not leverage XGBoost's built-in FL capabilities as effectively.

examples/advanced/random_forest/job.py
```diff
@@ -0,0 +1,110 @@
...
+This example demonstrates federated Random Forest training using NVFlare's recipe API.
+Each client trains a local sub-forest on their data, and these are aggregated on the server.
+
+Usage:
+    python job.py --n_clients 5 --local_subsample 0.5 --data_split_path /tmp/nvflare/random_forest/HIGGS/data_splits/5_uniform
+"""
+
+import argparse
+
+from jobs.random_forest_base.app.custom.higgs_data_loader import HIGGSDataLoader
```
YuanTingHsieh: suggest we either move this data loader out or include the code here.
another thing, do we still need this random_forest_base job?
@ZiyueXu77

nvkevlu: I moved it

examples/advanced/random_forest/job.py
```diff
@@ -0,0 +1,109 @@
...
+This example demonstrates federated Random Forest training using NVFlare's recipe API.
+Each client trains a local sub-forest on their data, and these are aggregated on the server.
+
+Usage:
+    python job.py --n_clients 5 --local_subsample 0.5 --data_split_path /tmp/nvflare/random_forest/HIGGS/data_splits/5_uniform
+"""
+
+import argparse
+
+from nvflare.app_opt.xgboost.higgs_data_loader import HIGGSDataLoader
```
YuanTingHsieh, line 27: do you think this file should be here?
or inside the example folder?
this is subjective I guess I could not decide

YuanTingHsieh (approved):

examples/advanced/random_forest/job.py
nvkevlu: reasoning of putting it here is so it can be reusable for other XGBoost examples

PR #3856, opened 2025-11-21
Fix TLS corruption by replacing fork with posix_spawn

GeorgeWang-nv:
Fix TLS corruption in job launcher by replacing fork with posix_spawn to avoid inheriting gRPC state.

Fixes # .

### Description

Fix TLS corruption in job launcher by replacing fork with posix_spawn to avoid inheriting gRPC state.

### Types of changes

- [ ] Documentation updated.


nvidianz:
The fix looks good. However, we need to make the same change in seveal other places. It would be easier if we can make a Popen() replacement.

nvflare/app_common/job_launcher/process_launcher.py
```diff
@@ -25,41 +27,104 @@
...
+
+        if self.pid is None:
+            return
+
+        if self._return_code is None:
+            try:
+                _, status = os.waitpid(self.pid, 0)
+                self._return_code = self._decode_status(status)
+            except ChildProcessError:
+                pass
```
chesterxgchen: are we swallow errors ?

nvflare/app_common/job_launcher/process_launcher.py
```diff
@@ -71,12 +136,23 @@ def launch_job(self, job_meta: dict, fl_ctx: FLContext) -> JobHandleSpec:
...
-        # use os.setsid to create new process group ID
-        process = subprocess.Popen(shlex.split(command, True), preexec_fn=os.setsid, env=new_env)
+        argv = shlex.split(command, True)
+
+        if self._use_posix_spawn and argv:
+            try:
+                pid = self._spawn_with_posix(argv, new_env)
+            except Exception as exc:
+                self.logger.warning("posix_spawn failed (%s); falling back to subprocess.", exc)
+            else:
```
chesterxgchen: "else" seems to be wrong indentation

nvflare/app_common/job_launcher/process_launcher.py
GeorgeWang-nv: Yes, ChildProcessError may mean the process has already terminated. Since the goal of wait() is reached, so I ignore it

nvflare/app_common/job_launcher/process_launcher.py
GeorgeWang-nv: The "else" matches the "try" block, not the "if". So I think the indentation is correct.

GeorgeWang-nv:
> overall seems to be good, but suggest look at the two comment. One of the place seem to have code error
> 
> suggest to have unit tests

Sure, unit tests are very necessary. I will first refactor the solution based on Zihong's comments, and then add new unit tests to cover the new solution.

GeorgeWang-nv:
> The fix looks good. However, we need to make the same change in seveal other places. It would be easier if we can make a Popen() replacement.

Sure, let's see if there is a more reusable solution.

chesterxgchen:
team decided to pause on this for PR to discuss on next major release

chesterxgchen:
@nvidianz  can you update and look at this, see if this can be merged.

chesterxgchen:
@GeorgeWang-nv lets fix the unit tests lets merge it

GeorgeWang-nv:
Hi @chesterxgchen, I've added the new unit suites \`tests/unit_test/app_common/job_schedulers/process_launcher_test.py\` and \`tests/unit_test/utils/process_utils_test.py\` to cover the posix_spawn adapter logic introduced in this PR. Let me know if you need any additional scenarios covered.

chesterxgchen:
> 1. `nvflare/utils/process_utils.py`, line 154 ([link](/nvidia/nvflare/blob/eeecd616ec4e31fb856bffa91f0006de399b8664/nvflare/utils/process_utils.py#L154))
>    **style:** the `setsid` parameter for `os.posix_spawn` was added in Python 3.11. on Python 3.8-3.10, this will raise `TypeError` and trigger the fallback to subprocess, which is the intended behavior.

@GeorgeWang-nv  this is a good catch, seems with this change, python 3.9 to 3.10 will not work anymore. Is that correct

nvidianz:
This PR is good to be merged into main.

GeorgeWang-nv:
> > 1. `nvflare/utils/process_utils.py`, line 154 ([link](/nvidia/nvflare/blob/eeecd616ec4e31fb856bffa91f0006de399b8664/nvflare/utils/process_utils.py#L154))
> >    **style:** the `setsid` parameter for `os.posix_spawn` was added in Python 3.11. on Python 3.8-3.10, this will raise `TypeError` and trigger the fallback to subprocess, which is the intended behavior.
> 
> @GeorgeWang-nv this is a good catch, seems with this change, python 3.9 to 3.10 will not work anymore. Is that correct

@chesterxgchen 
Hello, I have tested different Python versions in a Docker container environment. Both Python 3.9 and Python 3.10 support os.posix_spawn with the setsid=True flag — even the earliest versions, such as python:3.9.0-slim-buster, have this support. Therefore, I’m not sure why the PR received this comment from the robot.

GeorgeWang-nv:
> ### Additional Comments (1)
> 1. `nvflare/utils/process_utils.py`, line 121-123 ([link](/nvidia/nvflare/blob/47a73d9043a0ef7405f81a4204c8842f41826caf/nvflare/utils/process_utils.py#L121-L123))
>    **logic:** fallback to `self.pid` may not terminate all child processes if the process isn't a group leader
> 
> 5 files reviewed, 1 comment
> 
> [Edit Code Review Agent Settings](https://app.greptile.com/review/github) | [Greptile](https://greptile.com?utm_source=greptile_expert&utm_medium=github&utm_campaign=code_reviews&utm_content=nvflare_3856)


I have tested that when using **process_utils.spawn_process** with **setsid=True**, the **pid** equals **pgid**. In this case, **pid = pgid** in most situations, so it should not cause the issue mentioned in the comment: *"fallback to self.pid may not terminate all child processes if the process isn't a group leader."*

When we look at this code section:

```python
try:
    pgid = os.getpgid(self.pid)
except ProcessLookupError:
    # Process already gone; nothing left to terminate.
    return
except PermissionError as exc:
    self.logger.warning("Unable to read pgid for %s (%s)", self.pid, exc)
    pgid = self.pid
```

The code only enters the **except** block when **os.getpgid** fails. This is a very rare case and usually happens when the process is created by other software or under a highly restricted OS environment that blocks **getpgid**.

In conclusion, hitting the **except** block is already rare, and having **pid ≠ pgid** within that is even rarer. Additionally, we added a warning log, so if this situation ever occurs, we can analyze it based on that log.

chesterxgchen (approved):

PR #3858, opened 2025-11-21
Fix Android edge device stability and resume support

nvkevlu:
Fix Android edge device stability and resume support.

### Description

Fix Android edge device stability and resume support since the real cause was Android generating a new ID each time.

Also:

- remove unnecessary validation of ET and just display header bytes instead
- merge config to have job level params and then override with task params

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh (approved):
LGTM, please test it under the same scenario as QA, thanks!

PR #3882, opened 2025-12-09
Refactor ml-to-fl examples

YuanTingHsieh:
Split ml-to-fl into hello-xxx and multi-gpu ones.
merge with existing hello-xxx example for recipe usage
will test and fix multi-gpu examples in a later PR

### Description

- Split ml-to-fl into hello-xxx and multi-gpu ones
- Removes all "original" files, as those should be covered in our tutorials
- Update hello-numpy example
- Update the READMEs
- Update the NPModelPersistor and NumpyFedAvgRecipe to allow initial_model
- Fix a bug in nvflare/job_config/fed_job_config.py

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh:
multi-gpu can't use current recipe, will move them into advanced, need more work, will re-open

examples/advanced/multi-gpu/tf/client.py
```diff
@@ -41,32 +45,28 @@ def main():
...
 
+        # Evaluate trained model
         _, test_acc = model.evaluate(test_images, test_labels, verbose=2)
-        print(f"Accuracy of the model on the 10000 test images: {test_acc * 100} %")
+        print(f"Accuracy of trained model: {test_acc * 100:.2f}%")
 
-        # (6) construct trained FL model (A dict of {layer name: layer weights} from the keras model)
+        # (6) construct trained FL model
         output_model = flare.FLModel(
             params={layer.name: layer.get_weights() for layer in model.layers}, metrics={"accuracy": test_global_acc}
```
nvkevlu: Should accuracy be test_global_acc here? That would be the accuracy of the received model, so the server will not be using the right one for model selection.

Instead, should we send test_acc?

nvkevlu (commented):
one thing on the tf client.py, and then the AI comments about export_job, are those valid?

YuanTingHsieh:
Yes, I will fix those export, as for that global acc I think that is intended, because we want to evaluate the global model not the local ones.


But still, the multi GPU with recipe is problematic, I am trying to fix them

examples/hello-world/hello-pt/client.py
```diff
@@ -75,24 +110,32 @@ def main():
...
 
-                running_loss += cost.cpu().detach().numpy() / images.size()[0]
+                running_loss += cost.item()
+                if i % 2000 == 1999:
+                    avg_loss = running_loss / 2000
+                    print(f"[{epoch + 1}, {i + 1:5d}] loss: {avg_loss:.3f}")
+
+                    # Optional: Log metrics
+                    if summary_writer:
+                        global_step = input_model.current_round * steps + epoch * len(trainloader) + i
```
holgerroth: should be train_loader

PR #3892, opened 2025-12-12
Fix import error for TBAnalyticsReceiver

nvkevlu:
Fix import error for TBAnalyticsReceiver.

### Description

Fix import error for TBAnalyticsReceiver by removing the default.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/app_opt/pt/recipes/fedavg.py
+++ b/nvflare/app_opt/pt/recipes/fedavg.py
@@ -56,10 +56,15 @@ class FedAvgRecipe(UnifiedFedAvgRecipe):
             DIFF means that only the difference is sent. Defaults to TransferType.FULL.
         model_persistor: Custom model persistor. If None, PTFileModelPersistor will be used.
         model_locator: Custom model locator. If None, PTFileModelLocator will be used.
-        analytics_receiver: Component for receiving analytics data. If not provided, defaults to TBAnalyticsReceiver
-            for TensorBoard experiment tracking.
+        analytics_receiver: Component for receiving analytics data (e.g., TBAnalyticsReceiver for TensorBoard).
+            If not provided, no experiment tracking will be enabled.
+            To enable experiment tracking, either:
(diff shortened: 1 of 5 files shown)


YuanTingHsieh (dismissed):
LGTM

nvflare/app_opt/pt/recipes/fedavg.py
```diff
@@ -94,10 +97,17 @@ def __init__(
         params_transfer_type: TransferType = TransferType.FULL,
         model_persistor: Optional[ModelPersistor] = None,
         model_locator: Optional[ModelLocator] = None,
+        analytics_receiver: Optional[AnalyticsReceiver] = None,
     ):
         # Store PyTorch-specific model_locator before calling parent
         self._pt_model_locator = model_locator
 
+        # Default to TBAnalyticsReceiver for PyTorch if not provided
```
YuanTingHsieh: should we keep or not keep this default @holgerroth

nvflare/app_opt/tf/recipes/fedavg.py
```diff
@@ -93,7 +96,14 @@ def __init__(
         server_expected_format: ExchangeFormat = ExchangeFormat.NUMPY,
         params_transfer_type: TransferType = TransferType.FULL,
         model_persistor: Optional[ModelPersistor] = None,
+        analytics_receiver: Optional[AnalyticsReceiver] = None,
     ):
+        # Default to TBAnalyticsReceiver for TensorFlow if not provided
+        if analytics_receiver is None:
```
YuanTingHsieh: should we keep or not keep this default @holgerroth

holgerroth (approved):
Works well for me.

nvflare/app_opt/tf/recipes/fedavg.py
chesterxgchen: I felt we shouldn't, so we have less dependency, user need to add it if they need it

YuanTingHsieh (approved):

PR #3915, opened 2025-12-22
Fix LR recipe and persistor issue

YuanTingHsieh:
Fixes LR issues

### Description
- num_features were not passed in
- Persistor needs to over-write NPModelPersistor's method instead of using a "default_data" attribute over-write

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh:
will convert that example as well

chesterxgchen (approved):

PR #3917, opened 2025-12-23
Fix preflight check and ci

YuanTingHsieh:
Preflight check tool was hardcoded to test against GRPC communication scheme.
We have added more schemes and now our default is HTTP so we should change accordingly

### Description

- Update preflight check
- Remove overseer test
- Update preflight check tests

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/tool/package_checker/server_package_checker.py
+++ b/nvflare/tool/package_checker/server_package_checker.py
@@ -85,8 +85,12 @@ class ServerPackageChecker(PackageChecker):
         # Determine the communication scheme
         scheme = get_communication_scheme(package_path, NVFlareConfig.SERVER)
 
-        if scheme not in ["grpc", "agrpc", "http", "https"]:
-            raise RuntimeError(f"Communication scheme {scheme} is not supported.")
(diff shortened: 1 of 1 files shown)


nvflare/tool/package_checker/client_package_checker.py
```diff
@@ -35,9 +35,20 @@ def should_be_checked(self) -> bool:
...
+
+        # Only check GRPC connectivity if actually using GRPC
+        # For TCP-based schemes, skip connectivity check as it requires different approach
+        # TODO: Add TCP connectivity check in the future
+        if scheme in ["grpc", "agrpc"]:
+            self.rules = [
+                CheckGRPCServerAvailable(name="Check GRPC server available", role=self.NVF_ROLE),
+            ]
+        else:
+            # For non-GRPC schemes (tcp, atcp, stcp, http, etc.), skip server availability check
```
chesterxgchen: we should check this. Otherwise, we give a false sense of success

chesterxgchen (approved):
LGTM, but double check on any remaining comments

PR #3930, opened 2026-01-06
Add multi gpu example

YuanTingHsieh:
Add multi gpu examples

### Description

- Add FedAvgPerSiteConfig to support custom site args
- Add multi gpu example with TF, PT, lightning

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/examples/advanced/kaplan-meier-he/README.md
+++ b/examples/advanced/kaplan-meier-he/README.md
@@ -54,50 +54,215 @@ With these two settings, the server will have no access to any knowledge regardi
 
 The final Kaplan-Meier survival analysis will be performed locally on the global aggregated event list, recovered from decrypted global histograms.
 
-## Baseline Kaplan-Meier Analysis
-We first illustrate the baseline centralized Kaplan-Meier analysis without any secure features. We used veterans_lung_cancer dataset by
-`from sksurv.datasets import load_veterans_lung_cancer`, and used `Status` as the event type and `Survival_in_days` as the event time to construct the event list.
+## Example Structure
 
-To run the baseline script, simply execute:
+This section describes the key components of the federated Kaplan-Meier example.
+
+### Data
+
+**Dataset**: This example uses the `veterans_lung_cancer` dataset from `sksurv.datasets.load_veterans_lung_cancer`. The dataset contains survival information for veterans with advanced lung cancer.
+
+**Key Features**:
+- `Survival Days`: Time from observation start to end (event time)
+- `Status`: Boolean indicating if death occurred (TRUE) or patient was censored (FALSE)
+- Other features: Age, Celltype, Karnofsky score, Diagtime, Prior treatment, Treatment type
+
+**Data Preparation**: 
+- Script: `utils/prepare_data.py`
+- Splits the centralized dataset into multiple sites (default: 5 sites)
+- Applies time-binning (default: 7 days) to protect privacy
+- Output location: `/tmp/nvflare/dataset/km_data`
+- Each site receives a CSV file with its partition of the data
+
+**Time-Binning**: Raw survival times are discretized into bins (e.g., weekly intervals) to create histograms. This reduces granularity and prevents identification of individual patients from aggregated data.
+
+### Model
+
+**Algorithm**: Kaplan-Meier Estimator (non-parametric survival analysis)
+- Library: [lifelines](https://zenodo.org/records/10456828) - `KaplanMeierFitter`
+- Estimates survival function: S(t) = probability of surviving beyond time t
+- Accounts for both observed events (deaths) and censored observations
+- Non-iterative, one-shot analysis performed locally after receiving aggregated data
+
+**Output Files**: 
+
+Each client generates:
+1. **Survival Curve Plot**:
+   - Without HE: `km_curve_fl.png`
+   - With HE: `km_curve_fl_he.png`
+   - Baseline (centralized): `km_curve_baseline.png`
+   - Shows probability of survival over time (monotonically decreasing)
+
+2. **Analysis Details** (`km_global.json`):
+   - `timeline`: Time points where events occur
+   - `km_estimate`: Kaplan-Meier survival probability at each time point
+   - `event_count`: Number of events at each time point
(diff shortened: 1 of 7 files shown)


examples/advanced/multi-gpu/pt/client.py
```diff
@@ -0,0 +1,152 @@
...
+            # random parameters and gradients are synchronized in backward passes.
+            # Therefore, saving it in one process is sufficient.
+            torch.save(ddp_model.state_dict(), CHECKPOINT_PATH)
+
+            # (5) evaluate on received model for model selection
+            accuracy = evaluate(input_model.params, device, testloader)
+
+            # (6) construct trained FL model
+            output_model = flare.FLModel(
+                params=net.cpu().state_dict(),
```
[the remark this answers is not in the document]
YuanTingHsieh: this is intended as server side is NOT wrapped in DDP

examples/advanced/multi-gpu/pt/client.py
```diff
@@ -0,0 +1,152 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
```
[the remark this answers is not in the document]
YuanTingHsieh: it's 2026 already.

examples/advanced/multi-gpu/pt/client.py
```diff
@@ -0,0 +1,152 @@
...
+            # random parameters and gradients are synchronized in backward passes.
+            # Therefore, saving it in one process is sufficient.
+            torch.save(ddp_model.state_dict(), CHECKPOINT_PATH)
+
+            # (5) evaluate on received model for model selection
+            accuracy = evaluate(input_model.params, device, testloader)
+
+            # (6) construct trained FL model
+            output_model = flare.FLModel(
+                params=net.cpu().state_dict(),
```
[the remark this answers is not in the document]
YuanTingHsieh: intended

nvflare/recipe/fedavg.py
```diff
@@ -29,6 +29,17 @@
 from nvflare.recipe.spec import Recipe
 
 
+class FedAvgPerSiteConfig(BaseModel):
```
chesterxgchen: 1. Why does this is FedAvg specific ? can this be simply PersiteConfig ? 
I don't see anything fedAvg specific
2. The train_script and train_args are also in the recipe, whats the relationships ? 
3. We still have parameter_transfer_type, which means conversion, we might need a bit future proof. as we are could transfer nothing but a reference ID

nvflare/recipe/fedavg.py
```diff
@@ -204,22 +218,32 @@ def __init__(
         )
         job.to_server(controller)
 
-        # Add client executors
-        if isinstance(self.train_args, dict):
-            # Per-client configuration
-            for site_name, site_args in self.train_args.items():
+        if self.per_site_config is not None:
+            for site_name, site_config in self.per_site_config.items():
+                # Use site-specific config or fall back to defaults
```
chesterxgchen, line 220: I see the site specific overwrite recipe one if specified

examples/advanced/multi-gpu/pt/job.py
```diff
@@ -0,0 +1,81 @@
...
+
+    recipe = FedAvgRecipe(
+        name="pt_ddp",
+        initial_model=initial_model,
+        min_clients=num_clients,
+        num_rounds=args.num_rounds,
+        train_script=train_script,
+        train_args=train_args,
+        launch_external_process=True,  # DDP modes require external process launch
+        per_site_config={
```
chesterxgchen, line 55: I like pert_site_config, which can be specified or not specified. But FedAvgPerSiteConfig doesn't make sense. this is not FedAvg Specific 
require train scripts and train args  then allow the per-site config to overwrite them doesn't seem to make sense. But not sure which way is the best

case 1:  all site has the same training scripts and args, Job Recipe has train scripts and args only, no per-site config

Case 2:  each site has its own training script and train args, Job Recipe specified train scripts and args, but got overwritten by per-site train script and args --> question why bother provide by Job Recipe

Case 3 : All has has the same training scripts and args, per-site config is provided for each site, but the training scripts and args are not provided in per-site config.  The Job Recipe train scripts and args become the default --> good use 

Some are good, some are not so good. Not sure.

examples/advanced/multi-gpu/pt/job.py
```diff
@@ -0,0 +1,81 @@
...
+            "site-1": FedAvgPerSiteConfig(
```
chesterxgchen: why do we need a class for the PerSiteConfig ? it does help validation. But for user this felt strange. can they just provide the needed key, value pairs. We can still validate them for unknown thus ignored keys, and error out for missing required keys

nvflare/recipe/fedavg.py
YuanTingHsieh: removed as suggested.

examples/advanced/multi-gpu/pt/job.py
YuanTingHsieh: Good point on the design pattern. Let me clarify:
When per_site_config is provided, it defines the complete set of sites that will participate. There's no fallback - if a site isn't in per_site_config, it doesn't get an executor.
The recipe-level train_script/train_args serve as:
Case 1: When per_site_config is NOT provided → all clients use these defaults
Case 2/3: When per_site_config IS provided → these are the base values that each site's config overrides (if specified)
This means you can specify per_site_config with partial overrides:

```
per_site_config={
    "site-1": {"command": "special-command"},  # overrides only command
    "site-2": {},  # uses all recipe-level defaults
}
```

examples/advanced/multi-gpu/pt/job.py
YuanTingHsieh: removed as suggested.

chesterxgchen (commented):
address the review comments

examples/advanced/multi-gpu/pt/client.py
```diff
@@ -0,0 +1,152 @@
...
```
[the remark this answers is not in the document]
YuanTingHsieh: This is not true

PR #3949, opened 2026-01-13
Removed references to wss as a scheme

nvidianz:
### Description

1. Removed references to wss as a scheme/protocol. It's only used internally by aiohttp library.
2. Increased the downloader timeout in decomposers to 60s. It was causing download failures for large models.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/fuel/f3/drivers/aio_http_driver.py
+++ b/nvflare/fuel/f3/drivers/aio_http_driver.py
@@ -101,7 +101,7 @@ class AioHttpDriver(BaseDriver):
 
     @staticmethod
     def supported_transports() -> List[str]:
-        return ["http", "https", "ws", "wss"]
+        return ["http", "https"]
 
     @staticmethod
     def capabilities() -> Dict[str, Any]:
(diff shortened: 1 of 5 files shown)


nvidianz:
> ### Additional Comments (1)
> 1. `nvflare/fuel/f3/drivers/aio_http_driver.py`, line 138 ([link](/nvidia/nvflare/blob/be256ce525ab624cd5bf9b1441aab7885e0a2b70/nvflare/fuel/f3/drivers/aio_http_driver.py#L138))
>    **logic:** Still using `wss`/`ws` schemes internally for WebSocket URLs, but `supported_transports()` (line 104) now only returns `["http", "https"]`. This creates an inconsistency where the driver claims to not support ws/wss but still uses those schemes.
> 
> 5 files reviewed, 1 comment
> 
> [Edit Code Review Agent Settings](https://app.greptile.com/review/github) | [Greptile](https://greptile.com?utm_source=greptile_expert&utm_medium=github&utm_campaign=code_reviews&utm_content=nvflare_3949)
Good catch but this is used internally by the aiohttp library. So it's left as ws intentionally.

chesterxgchen:
actually, does this confusing user, we are using Websocket @nvidianz

PR #3962, opened 2026-01-15
Enable custom key_metric for FedAvgRecipe

ZiyueXu77:
Fixes # .

### Description

Since this is the default recipe we want people to use, the current model selector is too limiting to the default, permitting "accuracy" only - users will need other metrics and our current setting will result in warning and wrong results - currently GNN and LLM examples have this issue.

Foundamental question is: as FedAvgRecipe is based on BaseFedJob, do we want it to be at least as general as BaseFedJob, or just a sample recipe that expect users to take and modify for adjustments like "I want a different model selector"?

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/tests/unit_test/recipe/fedavg_recipe_test.py
+++ b/tests/unit_test/recipe/fedavg_recipe_test.py
@@ -27,7 +27,6 @@ from nvflare.app_common.aggregators.model_aggregator import ModelAggregator
 from nvflare.app_common.np.recipes import NumpyFedAvgRecipe
 from nvflare.app_common.widgets.intime_model_selector import IntimeModelSelector
 from nvflare.app_opt.pt.recipes.fedavg import FedAvgRecipe
-from nvflare.app_opt.tf.recipes.fedavg import FedAvgRecipe as TFFedAvgRecipe
 
 
 class SimpleTestModel(nn.Module):
@@ -217,15 +216,7 @@ class TestFedAvgRecipe:
 
 
 class TestFedAvgRecipeKeyMetricVariants:
-    """Test key_metric passthrough for TF and NumPy FedAvg recipes."""
-
-    def test_key_metric_passthrough_tf(self, mock_file_system, base_recipe_params):
-        key_metric = "val_f1"
-        recipe = TFFedAvgRecipe(name="test_tf_key_metric", key_metric=key_metric, **base_recipe_params)
-
-        model_selector = get_model_selector(recipe)
-        assert isinstance(model_selector, IntimeModelSelector)
-        assert model_selector.key_metric == key_metric
+    """Test key_metric passthrough for NumPy FedAvg recipes."""
 
     def test_key_metric_passthrough_numpy(self, mock_file_system):
         key_metric = "val_loss"


nvflare/recipe/fedavg.py
```diff
@@ -107,6 +111,15 @@ class FedAvgRecipe(Recipe):
             - server_expected_format (ExchangeFormat): Exchange format
             - params_transfer_type (TransferType): Parameter transfer type
             If not provided, the same configuration will be used for all clients.
+        key_metric: Metric used to determine if the model is globally best. If validation metrics are a dict,
```
holgerroth, line 111: Exposing metric key should be enough. I wouldn't expose the model selector component itself.
Client code can always do the negation.

ZiyueXu77: then we will need to tell user "you have to pass neg_val_loss rather than val_loss to use FedAvgRecipe, otherwise it will not work - or create your own recipe"

nvflare/recipe/fedavg.py
```diff
@@ -136,6 +149,9 @@ def __init__(
         model_persistor: Optional[ModelPersistor] = None,
         analytics_receiver: Optional[AnalyticsReceiver] = None,
         per_site_config: Optional[dict[str, dict]] = None,
+        key_metric: str = "accuracy",
+        negate_key_metric: bool = False,
+        model_selector: Optional[FLComponent] = None,
```
YuanTingHsieh: if we want to do this, just has one arg model_selector?
the key_metric and negate_key_metric seems only relates to IntimeModelSelector?

ZiyueXu77: yes, I am debating if we shall just remove the default IntimeModelSelector entirely from BaseFedJob since it may not be a "core component" from FL's perspective - just let users add them if necessary @chesterxgchen @holgerroth what do you think?

nvflare/recipe/fedavg.py
holgerroth: yes. Let's just make it clear that the value needs to be higher to be considered better.

nvflare/recipe/fedavg.py
holgerroth: I would keep it as a core NVFlare feature that is enabled by default.

nvflare/recipe/fedavg.py
ZiyueXu77: ok let's keep it minimal

holgerroth:
Needs to be exposed in the pt, tf, & numpy versions of FedAvgRecipe as well.

nvflare/recipe/fedavg.py
```diff
@@ -107,6 +108,9 @@ class FedAvgRecipe(Recipe):
             - server_expected_format (ExchangeFormat): Exchange format
             - params_transfer_type (TransferType): Parameter transfer type
             If not provided, the same configuration will be used for all clients.
+        key_metric: Metric used to determine if the model is globally best. If validation metrics are a dict,
+            key_metric selects the metric used for global model selection by the IntimeModelSelector.
+            Defaults to "accuracy".
```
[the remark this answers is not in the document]
ZiyueXu77: test added

ZiyueXu77:
> Needs to be exposed in the pt, tf, & numpy versions of FedAvgRecipe as well.

good point! added

holgerroth (approved):
Makes sense!

PR #3988, opened 2026-01-20
Fix hello-numpy-cross-val example

nvkevlu:
Fix hello-numpy-cross-val example.

### Description

Fix hello-numpy-cross-val example.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/docs/user_guide/nvflare_cli/poc_command.rst
+++ b/docs/user_guide/nvflare_cli/poc_command.rst
@@ -396,6 +396,7 @@ Modify your ``job.py`` to use ``PocEnv`` instead of ``SimEnv``:
     # Create your recipe
     recipe = NumpyFedAvgRecipe(
         name="hello-numpy",
+        initial_model=[[1, 2, 3], [4, 5, 6], [7, 8, 9]],
         min_clients=2,
         num_rounds=3,
         train_script="client.py",
--- a/examples/hello-world/hello-numpy-cross-val/README.md
+++ b/examples/hello-world/hello-numpy-cross-val/README.md
@@ -128,6 +128,7 @@ from nvflare.recipe import SimEnv
 # 1. Create a standard FedAvg recipe
 recipe = NumpyFedAvgRecipe(
     name="hello-numpy-train-cse",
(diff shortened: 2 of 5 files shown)


examples/hello-world/hello-numpy-cross-val/client.py
```diff
@@ -58,8 +58,17 @@ def main():
         input_model = flare.receive()
         print(f"Client {client_name}, current_round={input_model.current_round}")
 
-        input_np_arr = input_model.params[NPConstants.NUMPY_KEY]
-        print(f"Received weights: {input_np_arr}")
+        # Handle empty params (can happen when no initial model provided)
+        if NPConstants.NUMPY_KEY in input_model.params:
```
YuanTingHsieh: instead of handling this in client.py, suggest put initial_model in "job.py"

examples/hello-world/hello-numpy/client.py
```diff
@@ -58,8 +58,17 @@ def main():
         input_model = flare.receive()
         print(f"Client {client_name}, current_round={input_model.current_round}")
 
-        input_np_arr = input_model.params[NPConstants.NUMPY_KEY]
-        print(f"Received weights: {input_np_arr}")
+        # Handle empty params (can happen when no initial model provided)
+        if NPConstants.NUMPY_KEY in input_model.params:
```
YuanTingHsieh: I think this example is working fine because it has initial_model in job.py

nvflare/app_common/np/recipes/fedavg.py
```diff
@@ -188,6 +188,8 @@ def __init__(
         if self.initial_model is not None:
```
YuanTingHsieh: should we make this initial_model mandatory?
not sure, if mandatory means this recipe always have one initial model from server side.
if not-mandatory, then some more logic is needed to distribute the initial model from wherever to each sites.

nvkevlu: will make mandatory since there needs to be a starting point, and to keep it more clear

PR #3998, opened 2026-01-21
[2.7] Cherry Pick Fix hello-numpy-cross-val example

nvkevlu:
Cherry Pick Fix hello-numpy-cross-val example, #3988 

### Description

Cherry Pick Fix hello-numpy-cross-val example, #3988 .

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh:
can you list all the commits your cherry-picked?
seems this PR picks up a lot more changes.
if cherry-picking 3988
it should be:

```
git checkout 2.7
git checkout -b some_name_you_like
git cherry-pick 81be3001a7886fab3c8e01a32d53fbe5adbe70a6
git commit 
git push origin some_name_you_like
```

nvkevlu:
That does not work because the state of the files is way off, I'm not sure there is an easier way since there must have been other missed changes in 2.7 that this rolls up in order to get the files into the same final state. If there are changes that are in main that should not be in 2.7, then we may need to further change this, otherwise this should be the same.

edit: with the other PRs merged, the changes are now much less, which of course makes sense because this one has fixes and the PR with the original example was not merged at first.

PR #4011, opened 2026-01-22
[2.7] Fix NumPy CSE regression in 2.7.2rc4

nvkevlu:
Override self.framework = RAW after super().__init__() to restore
cross-site evaluation compatibility for NumPy recipes.

### Description

Fix critical regression in 2.7.2rc4 where NumPy recipes using `add_cross_site_evaluation()` fail with:
ValueError: Unsupported framework: FrameworkType.NUMPY

### Root Cause

PR #3993 refactored `NumpyFedAvgRecipe` to inherit from `UnifiedFedAvgRecipe` but inadvertently changed `self.framework` from `RAW` to `NUMPY`, breaking CSE auto-detection.

### The Fix

Override `self.framework = RAW` after `super().__init__()` to restore correct behavior while preserving the unified architecture.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/lighter/utils.py
+++ b/nvflare/lighter/utils.py
@@ -281,7 +281,7 @@ def verify_folder_signature(src_folder, root_ca_path, single_signer=False, signa
                     public_key = cert.public_key()
                     verify_cert(cert_to_be_verified=cert, root_ca_public_key=root_ca_public_key)
             except:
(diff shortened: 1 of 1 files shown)


chesterxgchen:
@nvkevlu I approved, but you should use cursor to add unit tests to capture the changes

PR #4015, opened 2026-01-22
[2.7] Fix sklearn examples

YuanTingHsieh:
The FedAvg Controller need a ModelAggregator, so add CollectAndAssembleModelAggregator

### Description

The FedAvg Controller need a ModelAggregator, so add CollectAndAssembleModelAggregator

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/docs/programming_guide/controllers/client_controlled_workflows.rst
+++ b/docs/programming_guide/controllers/client_controlled_workflows.rst
@@ -610,6 +610,193 @@ Swarm Learning: config_fed_client.json
 
     Client assigned tasks contain model data. You can apply task_data_filters if privacy is a concern (the OUT filter for the sending client, and IN filters for the receiving client).
 
+.. _swarm_learning_large_models:
+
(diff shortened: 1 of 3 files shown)


nvflare/app_common/aggregators/__init__.py
```diff
@@ -14,6 +14,12 @@
 
 from .accumulate_model_aggregator import AccumulateWeightedAggregator
 from .collect_and_assemble_aggregator import CollectAndAssembleAggregator
```
chesterxgchen, line 16: do we need both aggregator ?

YuanTingHsieh: @ZiyueXu77 , or should we remove the old completely?

chesterxgchen: since its adaptor from old to new, thats keep this for now. we will need to rewrite in next release

PR #4021, opened 2026-01-23
[2.7] Recipe site configuration [skip ci]

chesterxgchen:
### Description
   In some cases, we need to specify the job configs to change the download chunk size, or timeout etc. which are the top level configuration in the fed_config_<server|client>.json file. 
   Currently, there is no way to do this with recipe API. The underline Job API do support this, but recipe did not expose this. 
   In this PR, we add two additional methods 
```

   recipe.add_server_config() 
   recipe.add_client_config()

```
  which expose this capability and make the data scientist task easier. Otherwise, they have to export the job config and then edit the json config files. 

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/docs/programming_guide/controllers/client_controlled_workflows.rst
+++ b/docs/programming_guide/controllers/client_controlled_workflows.rst
@@ -778,7 +778,7 @@ The following SwarmClientController parameters are particularly important for la
 
 **Timeouts and Flow Control:**
 
-- ``learn_task_timeout``: Upper bound for how long the aggregation client waits for a round to finish. **Default: None (uses MAX_TASK_TIMEOUT=3600)**. **Suggested: 3600 to 7200** for large models.
+- ``learn_task_timeout``: Upper bound for how long the aggregation client waits for a round to finish. **Default: None**. **Suggested: 3600 to 7200** for large models.
 - ``learn_task_ack_timeout``: Timeout for acknowledging learn task dispatch. **Default: 10**. **Suggested: 300 or higher** since large model initialization can be slow.
 - ``final_result_ack_timeout``: Timeout for ACKs after broadcasting final results. **Default: 10**. **Suggested: 300 to 600** as final result distribution is often the largest payload.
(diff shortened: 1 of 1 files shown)


nvflare/recipe/spec.py
```diff
@@ -176,6 +208,36 @@ def add_server_input_filter(self, filter: Filter, tasks: Optional[List[str]] = N
         """
         self.job.to_server(filter, filter_type=FilterType.TASK_RESULT, tasks=tasks)
 
+    def add_server_config(self, config: Dict):
```
pcnudde, line 198: It's ok for this checkin, but I would like to have some discussion about documentation length and philisophy. 
If I look at this function, it is 2 lines of code with 23 lines of comments. This also results in a lot of duplication with other documentation.

tests/unit_test/recipe/spec_test.py
```diff
@@ -0,0 +1,340 @@
...
+            # Read the generated client config
+            client_config_path = os.path.join(job_dir, "app", "config", "config_fed_client.json")
+            assert os.path.exists(client_config_path), f"Client config not found at {client_config_path}"
+
+            with open(client_config_path) as f:
+                client_config = json.load(f)
+
+            # Verify top-level params are present
+            assert client_config.get("np_download_chunk_size") == 4194304
+            assert client_config.get("custom_param") == "test_value"
```
pcnudde: Are so many tests needed for this? Isn't the code just passing a dict from the recipe api to the job api?

chesterxgchen: true, we can reduce the tests.

chesterxgchen: Let me make additional changes

nvflare/recipe/spec.py
chesterxgchen: reduce the doc string

tests/unit_test/recipe/spec_test.py
chesterxgchen: reduce the unit tests from 7 to 3

PR #4025, opened 2026-01-23
[2.7] Fixed the swarm bug

nvidianz:
Fixes: https://jirasw.nvidia.com/browse/FLARE-2729

1. Added num_receivers to the Downloader transaction to fix the swarm bug.
2. Fixed a race condition in Downloader.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/fuel/f3/streaming/stream_utils.py
+++ b/nvflare/fuel/f3/streaming/stream_utils.py
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 import logging
-import os
+import random
 import threading
 import time
 from concurrent.futures import ThreadPoolExecutor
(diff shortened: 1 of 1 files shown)


nvflare/fuel/f3/streaming/stream_utils.py
```diff
@@ -22,9 +22,10 @@
 
 STREAM_THREAD_POOL_SIZE = 128
 ONE_MB = 1024 * 1024
+MILLION = 1000000
 
 lock = threading.Lock()
-sid_base = int((time.time() + os.getpid()) * 1000000)  # microseconds
+sid_base = int(time.time() * MILLION) + random.randint(0, MILLION)  # microseconds + random
```
pcnudde, line 28: does this guarantee that there is no conflict? Or just limit a conflict to 1 in a million?
Any reason why we don't use a standard uuid for this?

nvidianz: It has a 1 in a million chance for collision if it starts at the same moment.

It's easier to debug with sequential IDs so the SID is an integer. So UUID can't be used.

PR #4075, opened 2026-01-30
Add notebook testing support and developer testing guide

chesterxgchen:
## Summary

This PR improves developer workflows in `runtest.sh` and related testing docs.

- Adds notebook test execution support and docs for developer testing flows.
- Adds dependency caching to speed up repeated local test runs.
- Adds optional `uv` usage for dependency installation with automatic fallback to `pip` when `uv` is unavailable.
- Adds the required Apache-2.0 license header to `tests/unit_test/fuel/utils/pipe/__init__.py`.

## Key Changes

- `runtest.sh`
  - Notebook testing mode and controls
  - Dependency cache marker logic
  - Verbose output support and command echoing
  - Optional `uv` install path (`uv pip install`), with `pip` fallback
- `docs/programming_guide/developer_testing.rst`
  - Added developer testing guide
- `docs/programming_guide/notebook_testing.rst`
  - Added notebook testing guide
- `docs/programming_guide.rst`
  - Updated toctree
- `tests/unit_test/fuel/utils/pipe/__init__.py`
  - Added standard NVFlare license header

## Test Plan

- `bash -n runtest.sh`
- `./runtest.sh -u -v`
- `./runtest.sh -n -v`
- `python3 ci/check_license_header.py`

### DIFF
--- a/.github/workflows/auto-format.yml
+++ b/.github/workflows/auto-format.yml
@@ -12,13 +12,13 @@ jobs:
 
     steps:
       - name: Checkout repo
-        uses: actions/checkout@v4
+        uses: NVIDIA/spark-rapids-common/checkout@main
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           fetch-depth: 0  
 
       - name: Set up Python
-        uses: actions/setup-python@v5
+        uses: actions/setup-python@v6
         with:
           python-version: 3.14
 
--- a/.github/workflows/blossom-ci.yml
+++ b/.github/workflows/blossom-ci.yml
@@ -1,4 +1,4 @@
-# Copyright (c) 2022-2025, NVIDIA CORPORATION.
+# Copyright (c) 2022-2026, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -77,7 +77,7 @@ jobs:
     runs-on: vulnerability-scan
     steps:
       - name: Checkout code
-        uses: actions/checkout@v4
+        uses: NVIDIA/spark-rapids-common/checkout@main
         with:
           repository: ${{ fromJson(needs.Authorization.outputs.args).repo }}
           ref: ${{ fromJson(needs.Authorization.outputs.args).ref }}
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,4 +1,4 @@
-# Copyright (c) 2022, NVIDIA CORPORATION.
+# Copyright (c) 2022-2026, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
@@ -36,7 +36,7 @@ jobs:
 
     steps:
     - name: Checkout repository
-      uses: actions/checkout@v4
+      uses: NVIDIA/spark-rapids-common/checkout@main
 
     # Initializes the CodeQL tools for scanning.
     - name: Initialize CodeQL
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -30,10 +30,10 @@ jobs:
         run: echo version_path=/version/${{ github.ref_name }}/ >> $GITHUB_ENV
 
       - name: Checkout your repository
-        uses: actions/checkout@v4
+        uses: NVIDIA/spark-rapids-common/checkout@main
 
       - name: Setup Node
-        uses: actions/setup-node@v4
+        uses: actions/setup-node@v5
         with:
           node-version: 20
           cache: npm
@@ -50,7 +50,7 @@ jobs:
         working-directory: ${{ env.site_path }}
 
       - name: Deploy
-        uses: JamesIves/github-pages-deploy-action@v4.6.4
+        uses: JamesIves/github-pages-deploy-action@v4.8.0
         with:
           branch: gh-pages
           folder: ${{ env.site_path }}/dist
--- a/.github/workflows/dispatch-format-command.yml
+++ b/.github/workflows/dispatch-format-command.yml
@@ -11,7 +11,7 @@ jobs:
 
     steps:
       - name: Trigger Format Workflow
-        uses: peter-evans/slash-command-dispatch@v3
+        uses: peter-evans/slash-command-dispatch@v5
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
           command: format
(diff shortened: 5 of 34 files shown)


docs/programming_guide/developer_testing.rst
```diff
@@ -0,0 +1,191 @@
...
+-------------------
+
+Run specific tests individually:
+
+.. code:: bash
+
+   ./runtest.sh -u    # Unit tests only (faster, no style checks)
+   ./runtest.sh -s    # Check code formatting only
+   ./runtest.sh -f    # Fix code formatting only
+   ./runtest.sh -n    # Notebook tests
```
pcnudde, line 39: i'd like to pass --numprocesses to this script. Currently it hardcodes it to to 8, which seems needed to blossom CI, but is slow on fast local computers. Would be better is we default to 'auto' but allow an override to 8 for CI.

docs/programming_guide/developer_testing.rst
```diff
@@ -0,0 +1,191 @@
...
+     - off
+     - Enable verbose output
+   * - ``-d`` / ``--dry-run``
+     - off
+     - Print commands without executing
+   * - ``--fresh-deps``
+     - off
+     - Force fresh dependency installation (bypass cache)
+
+Dependency Caching
```
pcnudde: Do we need this? How much time do we gain by adding this. Local benchmark shows 3 seconds difference on a total time of about about 4 minutes. I don't think it is worth the complexity and risk of having old dependencies

docs/programming_guide/notebook_testing.rst
```diff
@@ -0,0 +1,187 @@
...
+2. Click "..." menu on the cell
+3. Select "Add Cell Tag"
+4. Enter: ``skip-execution``
+
+How It Works
+============
+
+The testing framework (implemented in ``conftest.py``) automatically:
+
+1. **Before test**: Creates a ``.backup`` of the original notebook
```
pcnudde, line 129: could we add .ipynb.backup files to gitignore?

docs/programming_guide/notebook_testing.rst
```diff
@@ -0,0 +1,187 @@
...
+2. **Filters cells**: Removes cells tagged with ``skip-execution``, ``skip``, or ``colab``
+3. **Updates kernel**: Adjusts kernel spec to match the specified or detected kernel
+4. **Executes**: nbmake runs the notebook through the Jupyter kernel
+5. **Restores**: Original notebook is restored from backup
+6. **Cleans outputs**: Cell outputs are cleared based on ``--nbmake-clean`` setting
+
+This ensures:
+
+- Original notebooks in git remain unchanged
```
pcnudde: that is not really true. The notebooks get cleaned and not all our notebooks in github are clean at the moment. But maybe that is a good change

docs/programming_guide/developer_testing.rst
pcnudde: something else to consider: If we replace pip install by 'uv pip install' we get same speedup without having a risk of stale cache and old depencies.

pcnudde: After think more, I would remove all the dependency caching code. very limited gain for the complexity.

runtest.sh
```diff
@@ -186,17 +251,79 @@ function fix_style_import() {
...
+            echo "python3"
+            return
+        fi
+    fi
+
+    # Otherwise, don't guess. Let nbmake use the notebook's kernelspec metadata.
+    echo ""
+}
+
+function validate_kernel() {
```
pcnudde, line 209: Is this validation needed? If not valid I assume nbmake will fail as well.

docs/programming_guide/developer_testing.rst
chesterxgchen: Ok, I can update this

docs/programming_guide/developer_testing.rst
chesterxgchen: tried the uv, agree, with uv, the caching is not necessary

runtest.sh
chesterxgchen: this is valid, as we need to know what kernels avaible

docs/programming_guide/notebook_testing.rst
chesterxgchen: probably should

docs/programming_guide/developer_testing.rst
chesterxgchen: Done — default changed to `auto`, with `-j <N>` / `--numprocesses=<N>` to override (e.g. `-j 8` for CI). Docs updated.

docs/programming_guide/notebook_testing.rst
chesterxgchen: Done — `*.ipynb.backup` added to `.gitignore`.

docs/programming_guide/notebook_testing.rst
chesterxgchen: Fixed — wording updated to: 'The test does not modify the committed notebook (original is restored from backup)'. Removed the implication that committed notebooks are always clean.

runtest.sh
```diff
@@ -198,21 +283,39 @@ function help() {
...
+    echo "    ./runtest.sh -n -v examples/tutorials/flare_api.ipynb        # test single notebook with verbose"
+    echo "    ./runtest.sh -n --timeout=1800 --kernel=python3 examples/tutorials/"
     exit 1
 }
 
 coverage_report=false
 unit_test_report=false
 dry_run_flag=false
-pytest_numprocesses=8
+pytest_numprocesses="auto"
```
pcnudde: If this is the default setting it will fail the blossom CI test. (I suspect the machine has a lot of cores and using auto runs too many in parallel. This probaly has to stay at 8. (or maybe a higher number works, but i know auto fails).

runtest.sh
```diff
@@ -263,10 +393,19 @@ do
             dry_run_flag=true
         ;;
 
+        -j)
+            pytest_numprocesses="$2"
+            shift
+        ;;
```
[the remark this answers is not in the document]
chesterxgchen: Good catch — confirmed bug. `-j N` put the bare `N` last on the command line, causing the pre-scan at line 24 (`target="${@: -1}"`) to capture it as the test path. Dropped `-j` entirely; `--numprocesses=N` already uses value-in-token form so it's excluded from the pre-scan. Docs example updated accordingly.

runtest.sh
chesterxgchen: Reverted to 8 — auto fails blossom CI due to too many parallel workers. `--numprocesses=<N|auto>` is still available as an explicit override for local use.

PR #4087, opened 2026-02-02
[2.7] Fix xgboost recipes

nvkevlu:
Fix xgboost recipes.

### Description

Update to fix xgboost recipes.

### Types of changes

- [ ] Documentation updated.


holgerroth:
@greptileai add review

nvflare/app_opt/xgboost/recipes/histogram.py
```diff
@@ -187,42 +215,57 @@ def configure(self):
...
-        # Add executor to all clients
+        # Add TensorBoard receiver to server
+        tb_receiver = TBAnalyticsReceiver(tb_folder="tb_events")
+        job.to_server(tb_receiver, id="tb_receiver")
+
+        # Prepare common client components
         executor_params = {
             "data_loader_id": self.data_loader_id,
             "metrics_writer_id": self.metrics_writer_id,
         }
```
YuanTingHsieh, line 212: the FedXGBHistogramExecutor has the following args that could be "per_site_config" if we ever want to support:

```
    def __init__(
        self,
        data_loader_id: str,
        per_msg_timeout=60.0,
        tx_timeout=600.0,
        model_file_name="model.json",
        metrics_writer_id: str = None,
        in_process=True,
    ):
```

in the original recipe, we only supports customize data loader and metrics writer
if you want to add more customization, should be supporting these args instead of adding a new argument "data_loader" to this recipe

nvflare/app_opt/xgboost/recipes/histogram.py
```diff
@@ -187,42 +215,57 @@ def configure(self):
...
         if self.secure:
             executor_params["in_process"] = True
 
-        executor = FedXGBHistogramExecutor(**executor_params)
-        job.to_clients(executor, id="xgb_executor")
+        # Add executor, metrics, and data loaders per site to avoid @ALL overwriting site-specific components
+        if self.data_loader is not None:
```
YuanTingHsieh: suggest revert the changes to this file

nvflare/app_opt/xgboost/recipes/vertical.py
```diff
@@ -199,15 +205,30 @@ def __init__(
         self.metrics_writer_id = v.metrics_writer_id
         self.in_process = v.in_process
         self.model_file_name = v.model_file_name
+        self.data_loader = data_loader
```
YuanTingHsieh: suggest revert the changes to this file

nvflare/app_opt/xgboost/recipes/bagging.py
```diff
@@ -238,10 +256,33 @@ def configure(self):
         aggregator = XGBBaggingAggregator()
         job.to_server(aggregator, id="aggregator")
 
-        # Add per-site executors and data loaders if configured
-        from nvflare.app_opt.xgboost.tree_based.executor import FedXGBTreeExecutor
-
-        if self.per_site_config:
+        # Add executors and data loaders
+
+        if self.data_loader is not None:
```
YuanTingHsieh: the old recipe just support customize of data loader id
if we really want to let "per_site_config" support all arguments for a specific executor,
we need to do something like:

```
if self.per_site_config:
    for site_name, site_config in self.per_site_config.items():
    num_client_bagging =site_config.get("num_client_bagging", self.num_client_bagging)
    ...
    learning_rate = site_config.get("learning_rate", self.learning_rate)
    executor = FedXGBTreeExecutor(
        num_client_bagging=num_client_bagging,
        ...,
        learning_rate=learning_rate
    )
    job.to(executor, site_name, id="xgb_tree_executor")
```

YuanTingHsieh (changes requested):

holgerroth (approved):
LGTM

PR #4107, opened 2026-02-03
Tensor streaming timeout issue

YuanTingHsieh:
## Problem

When using tensor streaming, fast clients can timeout while waiting for the next task, even though the server is legitimately waiting for slow clients to finish downloading tensors. This happens because:

1. Fast clients finish downloading tensors early and immediately call `get_task()`
2. Server blocks waiting for all clients (up to `wait_send_task_data_all_clients_timeout`, e.g., 300s)
3. Client `get_task_timeout` defaults to communicator timeout (e.g., 60s) - much lower than server wait time
4. Fast clients timeout at 60s while server is still waiting for slow clients
5. Fast clients timeout and retry, but get "TRY_AGAIN" response because their task is still marked as "being processed" (server still blocked waiting for slow clients). This causes either deadlock or eventual system_panic after the server's wait timeout expires.

**Root cause**: Timeout configuration mismatch - `get_task_timeout` not correlated with tensor streaming timeouts.

## Solution

Implement automatic timeout coordination between server and clients:

1. **Server calculates** minimum required client timeout:
   ```python
   min_timeout = wait_send_task_data_all_clients_timeout + 60s buffer
   ```

2. **Server communicates** this requirement to clients via task response headers

3. **Clients automatically adjust** their `get_task_timeout` if it's lower than required:
   ```python
   if server_min_timeout > current_timeout:
       self.timeout = server_min_timeout
   ```

## Changes

- `nvflare/apis/fl_constant.py`: Add `MIN_GET_TASK_TIMEOUT` constant
- `nvflare/app_opt/tensor_stream/server.py`: Calculate and set minimum timeout requirement
- `nvflare/private/fed/server/server_commands.py`: Include timeout in task response
- `nvflare/private/fed/client/communicator.py`: Auto-adjust client timeout when needed
- `nvflare/app_opt/tensor_stream/client.py`: Update documentation

## Benefits

✅ **Automatic**: No manual timeout configuration needed
✅ **Transparent**: Clear logging when adjustment occurs
✅ **Backward compatible**: Existing configs continue to work
✅ **Prevents failures**: Eliminates timeout-related run collapses

## Verification

Run tensor streaming example and check logs:

**Server log:**
```
TensorServerStreamer: Requiring clients to use get_task_timeout >= 360.0s 
(wait_send_task_data_all_clients_timeout=300.0s + 60s buffer)
```

**Client log (when adjustment occurs):**
```
Server requires get_task_timeout >= 360.0s (likely for tensor streaming). 
Automatically adjusting from 60.0s to 360.0s.
```

These messages confirm automatic timeout coordination is working.

**Technical details:**
- Server's `process_task_request()` fires `AFTER_TASK_DATA_FILTER` event
- `TensorServerStreamer` event handler blocks in `wait_sending_task_data_all_clients()` 
- This blocks the response to `get_task()` until ALL clients receive tensors
- Fast client's `get_task()` times out before server finishes waiting

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/app_opt/tensor_stream/server.py
+++ b/nvflare/app_opt/tensor_stream/server.py
@@ -18,7 +18,7 @@ from threading import Lock
 
 from nvflare.apis.event_type import EventType
 from nvflare.apis.fl_component import FLComponent
-from nvflare.apis.fl_constant import FLContextKey
+from nvflare.apis.fl_constant import FLContextKey, ServerCommandKey
 from nvflare.apis.fl_context import FLContext
 from nvflare.apis.streaming import StreamableEngine
 from nvflare.app_common.app_constant import AppConstants
@@ -55,6 +55,10 @@ class TensorServerStreamer(FLComponent):
         try_to_clean_task_data(fl_ctx): Cleans the task data in the FLContext if all clients have received the tensors.
     """
 
+    # Buffer time added to wait_send_task_data_all_clients_timeout when calculating minimum get_task_timeout
+    # This accounts for network latency, processing time, and other overhead
+    GET_TASK_TIMEOUT_BUFFER = 60.0  # seconds
+
     def __init__(
         self,
         format: str = ExchangeFormat.PYTORCH,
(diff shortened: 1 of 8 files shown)


pcnudde:
I like this one :) We need much less manual timeout settings. Too difficult for the user.

nvflare/app_opt/tensor_stream/server.py
```diff
@@ -55,6 +55,10 @@ class TensorServerStreamer(FLComponent):
         try_to_clean_task_data(fl_ctx): Cleans the task data in the FLContext if all clients have received the tensors.
     """
 
+    # Buffer time added to wait_send_task_data_all_clients_timeout when calculating minimum get_task_timeout
+    # This accounts for network latency, processing time, and other overhead
+    GET_TASK_TIMEOUT_BUFFER = 60.0  # seconds
```
chesterxgchen, line 60: should this be configurable, or need to be hard-coded ?

YuanTingHsieh: if they want they adjust wait_send_task_data_all_clients_timeout, so we leave this 60 as implementation details.

PR #4126, opened 2026-02-04
[2.7] Snapshot task data only

YuanTingHsieh:
# Fix Data Corruption in Broadcast with `min_responses < total_clients`

## Problem

When using `min_responses < total_clients` with `InTimeAccumulateWeightedAggregator`, the server can proceed to aggregate and modify the global model **while slow clients are still downloading it**, causing data corruption.

**Why this issue exists:**
- **Old framework**: Always sent full copies (FOBS serialization creates N copies per broadcast) → No corruption, but high memory usage
- **With TensorStreamer/NumpyDownloader**: Only passes REF_ID (reference to shared memory) → Memory efficient, but vulnerable to in-place modifications

**Race condition:**
1. Server broadcasts model to 5 clients (`min_responses=2`)
2. TensorStreamer creates `Downloadable` with REF_ID pointing to model data
3. Fast clients (1-2) respond immediately → server aggregates
4. Server updates model **in-place** (e.g., `model.params += updates`)
5. Slow clients (3-5) still downloading from same memory → **receive corrupted data** (mix of old + new)

**Real example of in-place modification**: `nvflare/app_common/workflows/lr/fedavg.py:204`
```python
model.params[NPConstants.NUMPY_KEY] += model_update.params["newton_raphson_updates"]
```

---

### Broadcast Data Protection via `_broadcast_data`

**File**: `nvflare/apis/impl/wf_comm_server.py`

Added **one deep copy of `task.data` per broadcast** stored in `task._broadcast_data`:

```python
# In _send_task_to_client(), AFTER before_task_sent_cb runs:
if not hasattr(task, "_broadcast_data"):
    if isinstance(manager, (BcastTaskManager, BcastForeverTaskManager)):
        task._broadcast_data = copy.deepcopy(task.data)

task_data = getattr(task, "_broadcast_data", None) or task.data

# Per-client copy with per-client headers:
client_data = make_copy(task_data)  # shallow copy payload, deep copy headers
client_data.set_header(TASK_ID, ...)  # per-client headers on the copy
```

**How it works:**
1. First client requests task → `before_task_sent_cb` runs (can populate `task.data`)
2. `_broadcast_data = deepcopy(task.data)` captures the populated data
3. All clients download from `_broadcast_data` (the protected copy)
4. Per-client headers are set on per-client `make_copy()`, not on shared `_broadcast_data`
5. Controller can freely modify `task.data` after broadcast returns
6. Slow clients still get the original (copied) data → **No corruption!**

**Key design points:**
- ✅ ONE task object (no complex snapshot task tracking)
- ✅ ONE deep copy per broadcast (not per client) → Memory efficient
- ✅ `before_task_sent_cb` modifications ARE captured (runs before snapshot)
- ✅ `task.data` remains modifiable by controller after snapshot
- ✅ `task.props`, callbacks, completion_status all work normally
- ✅ No API changes - fully backward compatible
- ✅ **GlobalModelEval/CrossSiteModelEval work correctly** (callback loads model before snapshot)

**Why only broadcast?**
- **Primary issue**: Broadcast sends to multiple clients concurrently → race condition when `min_responses` allows early aggregation
- **Send/relay**: Sequential dispatch (one client at a time), typically used with `*_and_wait()` (blocking)
- **Cyclic constraint**: Cyclic workflow intentionally modifies `task.data` between clients - protection would interfere

---

### 1. Per-client data customization not supported for broadcast
```python
def before_task_sent_cb(client_task, **kwargs):
    # This runs for each client, but only FIRST client's result is captured
    client_task.task.data["client_specific"] = client_task.client.name
```
**Reason**: `_broadcast_data` is created once after the first `before_task_sent_cb` runs.  
**Note**: Broadcast is designed to send the same data to all clients.

### 2. Memory overhead
- One extra copy of `task.data` per broadcast call
- For large models, this temporarily doubles memory during broadcast
- `_broadcast_data` is explicitly cleaned up when task completes to free memory

### 3. Non-deep-copyable objects will fail
```python
task.data["file"] = open("data.txt")  # RuntimeError
task.data["lock"] = threading.Lock()  # RuntimeError
```

### 4. `broadcast_forever` data is frozen
- `_broadcast_data` is created once after first client's callback
- Later updates to `task.data` won't affect ongoing broadcast

---

### Test Updates
- Updated `controller_test.py`:
  - Added `NON_BROADCAST` category for callback tests that modify `task.data` per-client
  - Added `assert_task_data_valid()` helper to validate payload content and expected system headers
  - Updated all data comparison tests to use the helper (relay, send, broadcast)
- Fixed pre-existing bug in `test_server.py` (wrong constant string)
- All controller tests pass

### What's Protected
- ✅ In-place numpy array modifications: `weights[0] = 999.0`
- ✅ In-place aggregation: `np.add(weights, result, out=weights)`
- ✅ PyTorch tensor modifications: `tensor.add_(updates)`

### Workflows Verified
- ✅ **GlobalModelEval/CrossSiteModelEval**: Callback loads model → snapshot captures it → all clients get correct DXO

---

## Compatibility

✅ **No breaking changes** - all existing code works unchanged  
✅ **Transparent** - automatic protection for all broadcast controllers  
✅ **Safe** - tested with PyTorch, NumPy tensors  
✅ **Documented** - API docstrings updated to reflect new behavior

---

## Files Changed

1. `nvflare/apis/impl/wf_comm_server.py` - Added `_broadcast_data` protection and memory cleanup
2. `tests/unit_test/apis/impl/controller_test.py` - Updated tests for new behavior
3. `tests/unit_test/app_opt/tensor_stream/test_server.py` - Fixed pre-existing bug

### DIFF
--- a/nvflare/apis/controller_spec.py
+++ b/nvflare/apis/controller_spec.py
@@ -348,11 +348,6 @@ class ControllerSpec(ABC):
         could join/disconnect at any moment. While the task is standing, any client that joins automatically
         becomes a target for this broadcast.
 
-        NOTE: task.data is deep copied after before_task_sent_cb runs for the first client to protect
-        against data corruption from concurrent in-place modifications. This copy is then reused for
-        all subsequent clients. Modifications to task.data in before_task_sent_cb ARE captured for the
-        first client, but per-client data customization is NOT supported for broadcast.
-
         Args:
             task: the task to be sent
             fl_ctx: the FL context
@@ -403,10 +398,7 @@ class ControllerSpec(ABC):
         All clients will get the task every time it asks for a new task.
         This is a non-blocking call.
 
-        NOTE: task.data is deep copied after before_task_sent_cb runs for the first client to protect
-        against data corruption from concurrent in-place modifications. This copy is then reused for
-        all subsequent clients. Modifications to task.data in before_task_sent_cb ARE captured for the
-        first client, but per-client data customization is NOT supported for broadcast.
+        NOTE: you can change the content of the task in the before_task_sent function.
 
         Args:
             task: the task to be sent


nvflare/apis/controller_spec.py
```diff
@@ -348,6 +348,11 @@ def broadcast(
         could join/disconnect at any moment. While the task is standing, any client that joins automatically
         becomes a target for this broadcast.
```
chesterxgchen, line 350: why put into spec ? 
spec is interface, the copy is your implementation, you can change the implementation, not the interface

YuanTingHsieh: thanks, undo changes to this file.

chesterxgchen (approved):

PR #4127, opened 2026-02-04
[2.7] Fix xgboost adaptor

nvkevlu:
Fix xgboost adaptor issue.

### Description

Fix issue with xgboost.

### Types of changes

- [ ] Documentation updated.


nvflare/app_opt/xgboost/histogram_based_v2/fed_executor.py
```diff
@@ -63,4 +67,5 @@ def get_adaptor(self, fl_ctx: FLContext):
             tx_timeout=self.tx_timeout,
         )
         adaptor.set_runner(runner)
+        self._cached_adaptor = adaptor  # Cache for future calls
```
YuanTingHsieh: why is the old way passing but recipe need the cached adaptor?

YuanTingHsieh (approved):
LGTM

PR #4155, opened 2026-02-07
[2.7] Recipe: support relative initial_ckpt and API enhancements

chesterxgchen:
## Summary

This PR enhances the Recipe interface to support **relative `initial_ckpt` paths** across all applicable recipes, with unified validation, file bundling utilities, and new job API enhancements.

### Problem
- `initial_ckpt` only supported absolute paths, requiring users to know the exact server-side file location
- No convenient API to bundle local files into job packages
- Checkpoint validation was inconsistent across recipes (each used different validation logic)
- Cyclic PT/TF recipes bypassed validation for `initial_ckpt` due to passing `None` to the base class

### Solution
- **Unified checkpoint utilities**: `validate_initial_ckpt()` and `prepare_initial_ckpt()` in `nvflare/recipe/utils.py`
  - Relative paths: validated locally, bundled into server app custom/ directory, basename stored in config
  - Absolute paths: treated as server-side paths, passed through as-is
- **Runtime path resolution**: `PTFileModelPersistor` and NumPy `load_numpy_model()` resolve relative checkpoint paths using `APP_ROOT + custom/` at runtime
- **New BaseFedJob APIs**: `add_file_to()`, `add_file_to_server()`, `add_file_to_clients()` with `app_folder_type` parameter

### Files Changed (30 files, +553/-124)

**Core utilities (2 files)**
- `nvflare/recipe/utils.py` - Added `validate_initial_ckpt()`, `prepare_initial_ckpt()`
- `nvflare/job_config/api.py` - Added `add_file_to`, `add_file_to_server`, `add_file_to_clients` APIs

**Persistors / model loading (2 files)**
- `nvflare/app_opt/pt/file_model_persistor.py` - Relative path resolution via APP_ROOT
- `nvflare/app_common/np/utils.py` - Relative path resolution via APP_ROOT

**Recipe files (17 files)**
- `nvflare/recipe/fedavg.py` - Fixed misleading comment on model source validation
- `nvflare/recipe/cyclic.py` - Updated initial_ckpt docstring
- `nvflare/app_opt/pt/recipes/fedavg.py` - Use validate_initial_ckpt
- `nvflare/app_opt/pt/recipes/fedopt.py` - Use validate_initial_ckpt
- `nvflare/app_opt/pt/recipes/scaffold.py` - Use validate_initial_ckpt
- `nvflare/app_opt/pt/recipes/cyclic.py` - Early validation + docstring update
- `nvflare/app_opt/pt/recipes/fedeval.py` - Fix: use initial_ckpt for PTModel
- `nvflare/app_opt/pt/recipes/fedavg_he.py` - Use validate_initial_ckpt + docstring
- `nvflare/app_opt/pt/recipes/swarm.py` - Refactored for proper relative path bundling
- `nvflare/app_opt/tf/recipes/fedavg.py` - Use validate_initial_ckpt
- `nvflare/app_opt/tf/recipes/cyclic.py` - Early validation + docstring update
- `nvflare/app_opt/tf/job_config/base_fed_job.py` - Use validate_initial_ckpt
- `nvflare/app_common/np/recipes/fedavg.py` - Use validate_initial_ckpt
- `nvflare/app_common/np/recipes/lr/fedavg.py` - Use validate_initial_ckpt + min_clients
- `nvflare/app_common/np/recipes/cross_site_eval.py` - Absolute-only restriction
- `nvflare/edge/tools/edge_fed_buff_recipe.py` - Use validate_initial_ckpt

**Tests (9 files)**
- `tests/unit_test/recipe/utils_test.py` - **NEW**: 13 tests for validate/prepare utilities
- `tests/unit_test/job_config/fed_job_test.py` - 6 new tests for add_file_to_* APIs
- 7 updated recipe test files to match new validation behavior and remove unused FrameworkType imports

### Intentional Design Decisions
1. **NumpyCrossSiteEvalRecipe**: Explicitly restricted to absolute `initial_ckpt` paths only. Relative path support requires a different approach for `NPModelLocator` and is deferred to a future release.
2. **Sklearn recipes** (KMeans, SVM): Do not support `initial_ckpt` as they lack model persistors that accept checkpoint paths.
3. **FedAvgRecipeWithHE**: Passes both `model` and `persistor` to `PTModel` — this is correct because `PTModel` uses the persistor directly when provided, avoiding double registration.
4. **Cyclic PT/TF recipes**: Added explicit `validate_initial_ckpt()` calls in `__init__` because the base class receives `None` for `initial_ckpt`, bypassing its own validation.

## Test Plan
- [x] All 13 `utils_test.py` tests pass (validate_initial_ckpt, prepare_initial_ckpt)
- [x] All 6 `fed_job_test.py` add_file_to_* tests pass
- [x] All recipe unit tests pass (179 passed, 13 skipped)
- [x] No regressions in existing tests
- [x] isort and flake8 checks clean

### DIFF
--- a/nvflare/app_common/np/utils.py
+++ b/nvflare/app_common/np/utils.py
@@ -50,25 +50,32 @@ def load_numpy_model(
     """
     data = None
 
-    # Priority 1: Load from source checkpoint (absolute path) if provided
+    # Priority 1: Load from source checkpoint if provided
     if source_ckpt_file_full_name:
+        if os.path.isabs(source_ckpt_file_full_name):
+            ckpt_path = source_ckpt_file_full_name
+        else:
+            # Relative path: resolve against app's custom directory
+            from nvflare.apis.fl_constant import FLContextKey, WorkspaceConstants
+
+            app_root = fl_ctx.get_prop(FLContextKey.APP_ROOT)
+            ckpt_path = os.path.join(app_root, WorkspaceConstants.CUSTOM_FOLDER_NAME, source_ckpt_file_full_name)
         # If user explicitly specified a checkpoint, it MUST exist (fail fast to catch config errors)
-        if not os.path.exists(source_ckpt_file_full_name):
+        if not os.path.exists(ckpt_path):
             raise ValueError(
-                f"Source checkpoint not found: {source_ckpt_file_full_name}. "
-                "Check that the checkpoint exists at runtime."
+                f"Source checkpoint not found: {ckpt_path}. " "Check that the checkpoint exists at runtime."
             )
         try:
             logger.log_info(
                 fl_ctx,
-                f"Loading model from source checkpoint: {source_ckpt_file_full_name}",
+                f"Loading model from source checkpoint: {ckpt_path}",
                 fire_event=False,
             )
-            data = np.load(source_ckpt_file_full_name)
+            data = np.load(ckpt_path)
         except Exception as e:
             # If loading fails after file exists, this is a real error - raise it
             raise ValueError(
-                f"Failed to load from source checkpoint {source_ckpt_file_full_name}: " f"{secure_format_exception(e)}"
+                f"Failed to load from source checkpoint {ckpt_path}: " f"{secure_format_exception(e)}"
             ) from e
 
     # Priority 2: Load from model file path
--- a/nvflare/app_opt/pt/recipes/cyclic.py
+++ b/nvflare/app_opt/pt/recipes/cyclic.py
@@ -30,8 +30,9 @@ class CyclicRecipe(BaseCyclicRecipe):
             - Dict config: {"path": "module.ClassName", "args": {"param": value}}
             - PTModel instance (already wrapped)
             - None: no initial model
-        initial_ckpt: Absolute path to a pre-trained checkpoint file. The file may not
-            exist locally as it could be on the server. Used to load initial weights.
+        initial_ckpt: Path to a pre-trained checkpoint file. Can be:
+            - Relative path: file will be bundled into the job's custom/ directory.
+            - Absolute path: treated as a server-side path, used as-is at runtime.
             Note: PyTorch requires model when using initial_ckpt (for architecture).
         num_rounds: Number of complete training rounds to execute. Defaults to 2.
         min_clients: Minimum number of clients required to participate. Must be >= 2.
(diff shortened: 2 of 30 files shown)


nvflare/app_common/np/utils.py
```diff
@@ -50,25 +50,32 @@ def load_numpy_model(
...
+    # Priority 1: Load from source checkpoint if provided
     if source_ckpt_file_full_name:
+        if os.path.isabs(source_ckpt_file_full_name):
+            ckpt_path = source_ckpt_file_full_name
+        else:
+            # Relative path: resolve against app's custom directory
+            from nvflare.apis.fl_constant import FLContextKey, WorkspaceConstants
+
+            app_root = fl_ctx.get_prop(FLContextKey.APP_ROOT)
+            ckpt_path = os.path.join(app_root, WorkspaceConstants.CUSTOM_FOLDER_NAME, source_ckpt_file_full_name)
```
YuanTingHsieh, line 62: is simply using relative path and absolute path as a hint to say this model is on remote or local too much magic? maybe @pcnudde and @holgerroth can comment on that point

nvflare/app_opt/pt/recipes/cyclic.py
```diff
@@ -85,3 +94,20 @@ def __init__(
...
+        """Override to handle PyTorch-specific model setup with relative ckpt support."""
+        if self.model is None and self._pt_initial_ckpt is None:
+            return ""
+
+        # If model is already a PTModel wrapper (user passed PTModel directly), use as-is
+        if hasattr(self.model, "add_to_fed_job"):
+            result = job.to_server(self.model, id="persistor")
+            return result["persistor_id"]
+
+        from nvflare.recipe.utils import prepare_initial_ckpt
```
YuanTingHsieh, line 108: why we prefer import inside method

YuanTingHsieh (commented):
logic-wise LGTM, just the assumption i am not sure, is this too much magic?

Or we should be explicit saying this location is remote or local

nvflare/app_common/np/utils.py
pcnudde: I would have thought that the logic would be. 
- if the file exists locally, then we copy it to the server and use it).
- if not we assume it exists on the server. 

But if it exists locally via an absolute path,  (like ~/model/mymodel) where would be copy it then and how would we point to it. 

So in conclusion: relative then copy, absolute then assume it is on the server is I think a valid approach for the moment. In any case this is all a workaround for the lack of repository support. Any decent org should use some kind of repository of models and will refer to those via url.

pcnudde: Unless somebody has a good counterproposal, I'm fine with the current approach.

YuanTingHsieh: ok no problem

nvflare/app_opt/pt/recipes/cyclic.py
YuanTingHsieh: can resolve not a blocking thing just wondering why ai prefer it

YuanTingHsieh (approved):

PR #4168, opened 2026-02-10
[2.7] Fix hello-numpy-cross-val example

nvkevlu:
Properly make hello-numpy-cross-val use branching to match other examples.

Fixes # .

### Description

Properly make hello-numpy-cross-val use branching to match other examples and ensure that there will no longer be any errors due to numpy_key.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/fuel/f3/streaming/blob_streamer.py
+++ b/nvflare/fuel/f3/streaming/blob_streamer.py
@@ -22,7 +22,7 @@ from nvflare.fuel.f3.streaming.byte_receiver import ByteReceiver
 from nvflare.fuel.f3.streaming.byte_streamer import STREAM_CHUNK_SIZE, STREAM_TYPE_BLOB, ByteStreamer
 from nvflare.fuel.f3.streaming.stream_const import EOS
 from nvflare.fuel.f3.streaming.stream_types import Stream, StreamError, StreamFuture
-from nvflare.fuel.f3.streaming.stream_utils import FastBuffer, stream_thread_pool, wrap_view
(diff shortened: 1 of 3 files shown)


nvflare/app_common/np/recipes/fedavg.py
```diff
@@ -104,6 +106,7 @@ def __init__(
         *,
         name: str = "fedavg",
         model: Union[Any, Dict[str, Any], None] = None,
+        initial_model: Union[Any, Dict[str, Any], None] = None,  # backward compat (2.7 / old job.py)
```
YuanTingHsieh, line 109: do we need to be backward compatible on this @chesterxgchen

examples/hello-world/hello-numpy-cross-val/client.py
```diff
@@ -53,41 +53,71 @@ def main():
...
-        output_model = flare.FLModel(
-            params={NPConstants.NUMPY_KEY: params_to_send},
-            params_type=params_type,
-            metrics=metrics,
-            current_round=input_model.current_round,
-        )
-
-        flare.send(output_model)
+            # Unknown task; send empty metrics so the protocol is satisfied.
+            flare.send(flare.FLModel(metrics={}))
```
YuanTingHsieh, line 121: this part is a bit weird, what are those unknown tasks?

nvflare/app_common/np/recipes/fedavg.py
chesterxgchen: I think we do. This is ok as it is inside the recipe implementation

PR #4174, opened 2026-02-11
[2.7] smaller lock in produce item

pcnudde:
Fixes # .

### Description

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/app_common/ccwf/swarm_client_ctl.py
+++ b/nvflare/app_common/ccwf/swarm_client_ctl.py
@@ -845,23 +845,31 @@ class SwarmClientController(ClientSideController):
                 time.sleep(self.request_to_submit_result_interval)
 
             # send the result to the aggr
-            self.log_info(fl_ctx, f"sending training result to aggregation client {aggr}")
+            if aggr == self.me:
+                # Avoid synchronous self-message path through CoreCell._send_direct_message.
+                self.log_info(fl_ctx, "submitting training result locally (aggregation client is self)")
+                engine = fl_ctx.get_engine()
+                local_fl_ctx = fl_ctx.clone()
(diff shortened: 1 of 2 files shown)


nvflare/fuel/f3/streaming/cacheable.py
```diff
@@ -90,20 +90,30 @@ def clear_cache(self):
     def _get_item(self, index: int, requester: str) -> bytes:
         with self.lock:
             if not self.cache:
-                # the cache has been cleared
-                data = None
-            else:
-                data, _ = self.cache[index]
+                # cache cleared — produce without caching
+                return self.produce_item(index)
```
YuanTingHsieh: By doing this are we still produce_item inside the lock?

pcnudde: you are right, this seems wrong.

pcnudde: it's a edge case, (cache cleared) but should be out of locl

nvflare/fuel/f3/streaming/cacheable.py
```diff
@@ -90,20 +90,35 @@ def clear_cache(self):
...
-                if self.cache:
-                    self.cache[index] = (data, 0)
-                    self.logger.debug(f"created and cached item {index} for {requester}: {len(data)} bytes")
-            else:
-                self.logger.debug(f"got item {index} from cache for {requester}")
+        if not cache_available:
+            return self.produce_item(index)
+
+        if data is not None:
+            self.logger.debug(f"got item {index} from cache for {requester}")
```
YuanTingHsieh, line 103: we could simplify "if cache_available and data is not None"
but that is minor

PR #4196, opened 2026-02-15
Collab API user guide and specification [skip ci]

chesterxgchen:
## Summary

Add the Collab API (code name Fox) specification as two documents, split by target audience:

- **`docs/design/collab_api_spec.md`** (436 lines) -- User-facing API spec for FL algorithm researchers. Covers usage patterns, hybrid patterns (Collab server + Client API), progression paths from prototype to production, execution environments, and recipe relationships.

- **`docs/design/collab_api_design_spec.md`** (227 lines) -- Internal design spec for FLARE engineers. Covers CollabClientAPI bridge architecture (queue mechanics, contextvars, subprocess mode), execution environment backend selection internals, decorator/runtime infrastructure, 9 development requirements with phased implementation order, and 3 open design questions.

### Key design decisions documented

- **Unified `SimEnv`**: No separate `CollabSimEnv` -- the Collab simulation backend is an internal detail; users write the same `SimEnv`/`PocEnv` for all recipes
- **No auto-detection magic**: The recipe explicitly indicates its API type to the environment (mechanism is an open question, but no code scanning)
- **API layering**: Collab API (high-level, promoted) > ModelController API (mid-level, legacy) > Controller/Executor API (low-level, internal)
- **Two client patterns**: `@collab.publish` (in-process, fast prototyping) and Client API via `train_script` (multi-GPU, production)
- **Progression model**: Research (CollabRecipe + SimEnv) -> Hybrid (Collab server + Client API) -> New named recipe via Job API

### Open design questions

1. How does the recipe indicate its API type to the environment?
2. What does the recipe packaging utility look like?
3. How are Collab server + Client API client wired in the hybrid phase?

## Test plan

- [ ] Documentation review: verify both documents are internally consistent
- [ ] Cross-reference check: links between the two documents resolve correctly
- [ ] Code example review: all import paths match the current codebase (`nvflare.collab.sys.recipe`, `nvflare.recipe.sim_env`, `nvflare.app_opt.pt.recipes.fedavg`)
- [ ] No runtime impact -- documentation only, no code changes

### DIFF
--- a/docs/design/collab_api_design_spec.md
+++ b/docs/design/collab_api_design_spec.md
@@ -1,0 +1,227 @@
+# FLARE Collab API -- Internal Design Specification
+
+*Target audience: NVIDIA FLARE engineers who will design and implement the Collab API.*
+
+*For the user-facing API specification, see [collab_api_spec.md](collab_api_spec.md).*
+
+## 1. Scope
+
+This document covers internal architecture, bridge mechanics, execution environment internals, development requirements, and open design questions for the Collab API. It assumes familiarity with the user-facing spec and existing FLARE internals (Controller/Executor APIs, CellNet, ModelController, Job/Recipe infrastructure).
+
+---
+
+## 2. API Layering
+
+NVFlare has three levels of server-side API:
+
+```
+High-level:    Collab API  (@collab.main, @collab.publish)   <-- promoted API for new development
+Mid-level:     ModelController API                            <-- no longer promoted once Collab API ships
+Low-level:     Controller API + Executor API                  <-- internal plumbing
+```
+
+The Collab API is **built on top of** the Controller and Executor APIs. It is not a separate system -- it is a higher-level abstraction over the same underlying infrastructure.
+
+Today's standard recipes (`FedAvgRecipe`, `CyclicRecipe`, etc.) were built using the **ModelController API** (mid-level). Going forward, **new FL algorithms and recipes should be developed using the Collab API** (high-level).
+
+### Class Hierarchy
+
+```
+Recipe (base class -- nvflare.recipe.spec)
+  |
+  |-- CollabRecipe (nvflare.collab.sys.recipe)
+  |     Server: user-written @collab.main method  [Collab API -- high-level, promoted]
+  |     Client: @collab.publish methods or Client API train_script
+  |
+  |-- FedAvgRecipe (nvflare.app_opt.pt.recipes.fedavg)
+  |     Server: built-in scatter-and-gather  [ModelController API -- mid-level, legacy]
+  |     Client: Client API train_script
+  |
+  |-- CyclicRecipe, SimpleSwarmLearningRecipe, ...
+  |-- KMeansFedAvgRecipe, SVMFedAvgRecipe, ... (sklearn)
+  |-- NumpyFedAvgRecipe, FedAvgLrRecipe, ... (numpy)
+```
+
+---
+
+## 3. CollabClientAPI Bridge Architecture
+
+The Collab server uses a **push** model (`collab.clients.execute(fl_model)`) while the Client API uses a **pull** model (`flare.receive()`). The `CollabClientAPI` bridge connects the two.
+
+### Bridge Internals
+
+```
+Server (@collab.main)                       Client (Client API)
+========================                    ========================
+
+collab.clients.execute(fl_model)            flare.receive()
+         |                                       ^
+         v                                       |
+  +-------------------------------------------------+
+  |              CollabClientAPI (bridge)             |
+  |                                                   |
+  |  execute() is @collab.publish:                    |
+  |    1. puts fl_model in _call_queue                |
+  |    2. blocks on _result_queue                     |
+  |                                                   |
+  |  receive():                                       |
+  |    gets fl_model from _call_queue                 |
+  |                                                   |
+  |  send(result):                                    |
+  |    puts result in _result_queue                   |
+  |    -> unblocks execute() -> returns to server     |
+  |                                                   |
+  |  stop():                                          |
+  |    sets _stopped=True                             |
+  |    -> is_running() returns False                  |
+  +-------------------------------------------------+
+         |                                       |
+         v                                       v
+collab.clients.stop()                       flare.is_running() -> False
+```
+
+### Two Execution Modes
+
+| Mode | When | Mechanism |
+|------|------|-----------|
+| **In-process** (`launch_external_process=False`) | Single GPU, `SimEnv` | `CollabRecipe` sets `train_script` on `CollabClientAPI`. On first `execute()` call, the script is launched in a **background thread**. The script calls `flare.receive()` / `flare.send()` which route to the `CollabClientAPI` instance via **`contextvars`**. Queue-based handshake between `execute()` and `receive()`/`send()`. |
+| **Subprocess** (`launch_external_process=True`) | Multi-GPU (torchrun DDP) | `SubprocessLauncher` spawns the training script as a child process (e.g. via `torchrun`). Communication between the `CollabClientAPI` (in the parent FLARE process) and the client script (in the subprocess) uses **CellNet channels**. Only rank 0 communicates with the server; other ranks sync via `dist.broadcast`. |
+
+### Wiring Sequence
+
+When `train_script="client.py"` is specified in `CollabRecipe`:
+
+1. `CollabRecipe` creates a `CollabClientAPI()` instance.
+2. It calls `client_api.set_train_script("client.py", ...)` to register the script path.
+3. The `CollabClientAPI` is passed to the executor as the client object.
+4. When the server calls `collab.clients.execute(fl_model=...)`, this invokes `CollabClientAPI.execute()` (a `@collab.publish` method).
+5. `CollabClientAPI.execute()` starts the training script (first call only) and bridges `fl_model` to `flare.receive()` via queues.
+6. The training script's `flare.send(result)` puts the result back in the queue, which `execute()` returns to the server.
+
+---
+
+## 4. Execution Environment Internals
+
+### 4.1 Requirements
+
+1. **Backward compatible:** Existing recipes (no Collab API) must continue to work exactly as before.
+2. **Collab Backend instantiation:** When the Collab API is involved, the environment must instantiate a **Collab Backend** to handle `@collab.main` / `@collab.publish` dispatch.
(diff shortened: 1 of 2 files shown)


docs/design/collab_api_design_spec.md
```diff
@@ -0,0 +1,227 @@
...
+
+```
+High-level:    Collab API  (@collab.main, @collab.publish)   <-- promoted API for new development
+Mid-level:     ModelController API                            <-- no longer promoted once Collab API ships
+Low-level:     Controller API + Executor API                  <-- internal plumbing
+```
+
+The Collab API is **built on top of** the Controller and Executor APIs. It is not a separate system -- it is a higher-level abstraction over the same underlying infrastructure.
+
+Today's standard recipes (`FedAvgRecipe`, `CyclicRecipe`, etc.) were built using the **ModelController API** (mid-level). Going forward, **new FL algorithms and recipes should be developed using the Collab API** (high-level).
```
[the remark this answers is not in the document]
pcnudde: test reply

docs/design/collab_api_spec.md
```diff
@@ -0,0 +1,436 @@
...
+recipe.export("/path/to/job_output")
+```
+
+---
+
+## 3. Collab Server + Client API (Hybrid Pattern)
+
+As your algorithm matures, you may want to use the **Client API** (`flare.receive()` / `flare.send()`) on the client side while keeping the Collab API on the server. This enables multi-GPU support, multi-node deployment, and independent server/client development.
+
+The framework provides a transparent bridge that connects the two. You don't need to understand how it works -- just configure `train_script` in `CollabRecipe`.
```
ZiyueXu77, line 211: For FL algo with a heavy client-side component (e.g. [SCAFFOLD](https://github.com/NVIDIA/NVFlare/blob/main/examples/advanced/cifar10/pt/cifar10-sim/cifar10_scaffold/client.py)), not sure if further moving to ClientAPI will make it simpler - multi-GPU/node support will be a key strength, guess it is not possible to support them for native CollabAPI

docs/design/collab_api_spec.md
```diff
@@ -0,0 +1,436 @@
...
+1. **Extract client** into a standalone `train.py` using Client API (`flare.receive()` / `flare.send()`).
+2. **Replace CollabRecipe** with the matching recipe (e.g. `FedAvgRecipe`).
+3. **Deploy** via `recipe.export()` or `recipe.execute(PocEnv(...))`.
+
+### 4.3 Creating a New Recipe
+
+This is the **primary goal** of the Collab API.
+
+```
+Phase 1 (Research):   CollabRecipe + @collab.main + @collab.publish  (SimEnv, pure function calls)
```
ZiyueXu77, line 349: "no latency/memory overhead from communication (serialization/deserialization)"

docs/design/collab_api_spec.md
```diff
@@ -0,0 +1,436 @@
...
+Phase 2 (Hybrid):     Collab server + Client API client              (multi-GPU, decoupled development)
```
ZiyueXu77, line 350: in this case, do we have communications happening under SimEnv?

docs/design/collab_api_spec.md
```diff
@@ -0,0 +1,436 @@
...
+# No server code needed
+```
+
+### 6.3 Shared Capabilities (from Base Recipe)
+
+| Capability | Method |
+|-----------|--------|
+| Execute locally | `recipe.execute(env)` |
+| Export for production | `recipe.export(path)` |
+| Server/Client filters | `recipe.add_server_output_filter(...)` / `recipe.add_client_input_filter(...)` |
```
ZiyueXu77, line 423: will it work directly with native CollabAPI (filters work on DXOs)?

ZiyueXu77:
Overall I feel researchers will implement with CollabAPI only, and data scientists will like adapting from ClientAPI. 
But then there will be a gap: if data scientists want to use a particular FL algo written with CollabAPI, who is expected to make the conversion from pure CollabAPI to Collab+Client? Researchers may not care to do that - they just want to show their method is effective, while it may be a bit too much for data scientists (maybe not for data scientists who use AI tools)?

docs/design/collab_api_spec.md
chesterxgchen: does Scaffold not working with Client API ?  Not sure I follow the comment

docs/design/collab_api_spec.md
chesterxgchen: correct

docs/design/collab_api_spec.md
chesterxgchen: You can still use SimEnv. the communication is a Push from Server => Client CollabClientAPI (i.e. a proxy), CollabClientAPI ( which as the @collab.publish annotation), receive the model then pass to the training code (client.py) in subprocess

docs/design/collab_api_spec.md
chesterxgchen: Good questions,  actually adding filter probably doesnt make sense, and may not be needed.

docs/design/collab_api_spec.md
holgerroth, line 1: Should this be a "user guide" rather than a spec to separate more from the spec document?

docs/design/collab_api_spec.md
```diff
@@ -0,0 +1,436 @@
+# FLARE Collab API Specification
+
+*Target audience: FL algorithm researchers and data scientists developing federated learning algorithms with NVIDIA FLARE.*
+
+*For internal design and implementation details, see [collab_api_design_spec.md](collab_api_design_spec.md).*
+
+## 1. Overview
+
+The **Collab API** (code name Fox -- FLARE Object Exchange) is NVIDIA FLARE's collaborative API that simplifies federated learning development. It allows researchers to write natural Python code while the framework handles distribution, communication, and orchestration.
```
holgerroth, line 9: don't need to mention code name.

docs/design/collab_api_spec.md
ZiyueXu77: I was thinking one major motivation of moving to ClientAPI is to make life simpler for data scientist (if not considering the multi-process benefit) - "independent server/client development".

But for this case, client dev is also heavy as part of the FL algo. The above links to our implementation with ClientAPI - as you can see it is fairly complicated (which is much beyond "only a few lines of code addition to centralized training script"). Moving from CollabAPI further to ClientAPI may not yield significant simplification. 

Of course, existing data scientist users will know ClientAPI better.

docs/design/collab_api_design_spec.md
```diff
@@ -0,0 +1,227 @@
...
+         v                                       v
+collab.clients.stop()                       flare.is_running() -> False
+```
+
+### Two Execution Modes
+
+| Mode | When | Mechanism |
+|------|------|-----------|
+| **In-process** (`launch_external_process=False`) | Single GPU, `SimEnv` | `CollabRecipe` sets `train_script` on `CollabClientAPI`. On first `execute()` call, the script is launched in a **background thread**. The script calls `flare.receive()` / `flare.send()` which route to the `CollabClientAPI` instance via **`contextvars`**. Queue-based handshake between `execute()` and `receive()`/`send()`. |
+| **Subprocess** (`launch_external_process=True`) | Multi-GPU (torchrun DDP) | `SubprocessLauncher` spawns the training script as a child process (e.g. via `torchrun`). Communication between the `CollabClientAPI` (in the parent FLARE process) and the client script (in the subprocess) uses **CellNet channels**. Only rank 0 communicates with the server; other ranks sync via `dist.broadcast`. |
```
ZiyueXu77, line 88: rather than (or in parallel to) moving to ClientAPI, can we support multi-process for CollabAPI by swaping the underlying "direct function call communicator" with a "cross-processor communicator" for pure CollabAPI users?

docs/design/collab_api_spec.md
chesterxgchen: The complication is from Client API ? or just scaffold itself ?  Can you compare the two and see where it is actually changed with Client API ?  I am not convinced. The count the meaningful changes in above code with Client API is only 5 lines. Rest of the code it was Scaffold. where is the complexity ?

ZiyueXu77: the complexity is from scaffold - the point I tried to make is this: whoever supposed to adapt a centralized cifar-10 training script to scaffold will have to make changes more than 5 lines - no matter it is from ClientAPI, or it is from scaffold - if they want to use scaffold, they will have to understand the whole concept and add the changes correctly following our example 

With FedAvg + ClientAPI, the user / data scientists' experience is "oh I heard there is an algorithm called FedAvg, I don't know the details but I want to try it out with my local data and training with only 5 lines of code change" - which is great and shows the value of ClientAPI. 

But when it comes to things like scaffold, then user will have to know the details of scaffold, and since they will have to know these details anyways, they should be able to implement with pure CollabAPI - which can be even simpler (no need to strictly define the scaffold_helper). It just seems to me in such case, the data-scientist-facing value of ClientAPI becomes less obvious as compared with basic FedAvg (where we can assume they just know local training scheme and have no idea of how FedAvg works)

ZiyueXu77:
> Overall I feel researchers will implement with CollabAPI only, and data scientists will like adapting from ClientAPI. 
> 
> But then there will be a gap: if data scientists want to use a particular FL algo written with CollabAPI, who is expected to make the conversion from pure CollabAPI to Collab+Client? Researchers may not care to do that - they just want to show their method is effective, while it may be a bit too much for data scientists (maybe not for data scientists who use AI tools)?

The assumption "data scientists will like adapting from ClientAPI" may not hold though - thinking about adding decorators on client side, it is also just a few lines' change, so from data scientists' perspective, it can even be a simpler solution

docs/design/collab_api_spec.md
chesterxgchen: Yes, they need to understand the interaction between what server does and client side does. This doesn't change the facts, that for client API they just need to know where to get the model (receive) and where to send the model ( send)

With Collab API, the receive and send are with annotations. It would be hard to use push annotations to support multi-GPUs, as the non-zero rank training code are not directly pushed by FLARE, but by rank-0 distribution. If we want to do that ourself, it would be a lot of engineering work and doesn't scale well. As push is only to one particular function, the current client training is code assume that we run from top to bottom, not just invoke a particular function 

so its better to switch to Client API.

docs/design/collab_api_spec.md
chesterxgchen: I add it togather to give sense how this be used. Otherwise, we will assume how this will be used.

docs/design/collab_api_design_spec.md
chesterxgchen: not sure this is worth doing ( not in terms of user experience, but in terms of engineering complexity and scalability).

chesterxgchen:
PR is closed to move the discussion to Google doc for easier review.

PR #4197, opened 2026-02-15
Add simplified enrollment system design spec

chesterxgchen:
## Summary

Add a comprehensive design document (`docs/design/enrollment_design.md`) for FLARE's simplified enrollment system. This replaces the centralized provisioning model with two CSR-based workflows where private keys are always generated locally and never transmitted.

### Manual Workflow (5-10 participants)
- Site Admin generates private key + CSR locally
- Sends only the CSR to Project Admin for signing
- Project Admin signs and returns cert + rootCA.pem + Server URI
- Site Admin generates startup kit locally with `nvflare package`
- No additional infrastructure required

### Auto-Scale Workflow (10+ participants)
- Same CSR model, automated via Certificate Service + enrollment tokens
- Project Admin distributes tokens to invited participants only
- Sites auto-generate CSR at startup, submit to Certificate Service
- Token authorizes automatic CSR signing
- Trade-off: requires deploying Certificate Service

### Document contents
- Executive summary with step-by-step workflow tables
- CLI command specs: `nvflare cert csr`, `nvflare cert sign`, `nvflare cert init`, `nvflare token`, `nvflare package`
- K8s deployment examples using K8s Secrets, ExternalSecret CRD, and ConfigMap
- Certificate Service architecture, HTTP API, approval policies
- Security analysis and threat model
- Implementation details and component overview

## Test plan

- [ ] Documentation review: verify workflow steps are consistent across all sections
- [ ] Code example review: CLI commands and K8s manifests are syntactically correct
- [ ] No runtime impact -- documentation only, no code changes

### DIFF
--- a/docs/design/enrollment_design.md
+++ b/docs/design/enrollment_design.md
@@ -322,6 +322,11 @@ Example (user token excerpt):
 | Replay attack | Single-use enforcement + expiration |
 | Brute force | UUID-based JTI + rate limiting |
 
+Token validity operational note:
+- Very short token TTLs reduce theft window, but can fail if deployment/startup is delayed (for example, delayed K8s pod scheduling).
+- Recommended pattern: issue tokens just-in-time (close to rollout), store them in short-lived secrets, and re-issue on expiration instead of extending TTL excessively.
+- A practical baseline is environment-specific (for example, hours for fast CI/K8s rollouts, longer windows for manual operations), with strict single-use still enforced.
+
 ### Token Signing Key
 
 By default, the root CA private key signs both participant certificates and JWT tokens. For advanced deployments, a separate JWT signing key pair can be used for key rotation or security isolation. Both TokenService and CertService must share the same JWT key pair.
@@ -500,12 +505,18 @@ pending:
   timeout: 604800                  # 7 days
   cleanup_interval: 3600
 
+rate_limit:
+  enroll_requests_per_minute: 60   # Per source identity/IP (implementation-defined)
+  token_requests_per_minute: 30    # Admin API path guardrail
+
 audit:
   enabled: true
   log_file: /var/log/cert_service/audit.log
 ```
 
 The service TLS certificate (`server.tls`) is for HTTPS and is separate from the FLARE root CA (`ca`), which is used only for signing FL participant certificates. On first start, if rootCA files don't exist in `data_dir`, they are auto-generated.
+
+Rate-limiting note: if service-level rate limiting is not enabled, equivalent controls should be enforced at the ingress/gateway layer (for example, NGINX, API Gateway, or WAF).
 
 Production safety note:
 - Auto-generation is bootstrap convenience only.
@@ -524,7 +535,7 @@ The `EnrollmentStore` ABC provides: `is_enrolled`, `add_enrolled`, `get_enrolled
 | --- | --- | --- |
 | Deployment | Single node | Multi-node / HA |
 | Dependencies | Built-in | Requires psycopg2 |
-| Concurrency | Limited | Full ACID |
+| Concurrency model | Single-writer | Multi-writer / concurrent |
 | Use Case | Dev, small production | Large production |
 
 ## Client Enrollment (Auto-Scale Only)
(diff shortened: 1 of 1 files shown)


chesterxgchen:
Addressed latest review feedback in commit `8d88497a5` and follow-up edits.

Key updates:
- Clarified that Manual and Auto-Scale are distinct workflows (manual CSR/cert exchange vs token/service auto-enrollment).
- Added explicit key-custody guidance for coexistence deployments.
- Fixed policy-flow consistency and resolved the remaining open review thread.
- Clarified runtime source resolution and explicitly defined `startup/enrollment.json`.
- Removed enrollment dependence on `fed_client.json` in the design text (enrollment remains isolated from existing runtime client config).
- Added UX simplification direction for manual packaging (`--cert-dir` / minimal-default packaging target), while keeping current commands as compatibility path.

Please re-review when convenient.

chesterxgchen:
Handled the three remaining review threads in latest commit `9ae943c41`:

1. `subject_type` naming aligned to `user` for consistency.
2. Token security wording updated to avoid conflict with pending re-submission flow (`Replay-controlled` + state/identity checks + expiration).
3. `GET /api/v1/enroll/{request_id}` clarified as optional external automation/observability endpoint; default client flow remains no-polling.

All previously open review threads are now resolved.

chesterxgchen:
Handled the latest PR review comments in commit `7acc90922`.

Addressed items:
- Replaced remaining inconsistent "single-use" wording with replay/state/expiry-consistent language.
- Added explicit user-token `roles` example while keeping main token JSON as a client example.
- Added `--type client` in token CLI examples where `subject_type` is discussed.
- Added explicit OU backward-compatibility note for legacy/provisioned certificates.

All currently open review threads are resolved.

chesterxgchen:
Follow-up update in commit `ba1599f71`:

Per design intent, I restored **single-use token semantics** as a key decision and aligned the pending flow accordingly:
- Tokens are invalid after one enrollment submission (no re-use).
- In pending flow, restart requires a newly issued token.

This keeps the doc aligned with the intended security model.

chesterxgchen:
Handled the latest review comment set in commit `e11e88939`.

Addressed items:
- Manual workflow package examples now include private key handling (`--key`) where explicit cert inputs are used.
- Clarified interaction between token-embedded policy and server `policy.file` defaults/guardrails.
- Added explicit `pending` action rule in policy YAML example.
- Clarified polling endpoint auth intent and admin observability path.
- Added production safety guidance for root CA auto-generation behavior.
- Added explicit pending-approval handoff steps (including new token issuance and out-of-band delivery).
- Specified startup failure behavior for auto-enrollment (retry, then fatal exit; never continue without certs).

All currently open review threads are now resolved.

pcnudde:
A few question about this certificate service. 
- where is it running? On the server? Somewhere else?
- How do people from the different sites access it? Is it available on the internet?
- What are the risks of running such a service? How do we secure it. It seems that if it gets hacked the whole system is compromised.
From a security aspect it seems we have a tradeoff between sharing the security keys out of band at the start and the fact that we now have a very important server component that absolutely needs to be secure.

chesterxgchen:
> A few question about this certificate service.
> 
> * where is it running? On the server? Somewhere else?
Ideally somewhere else, but could be along side of Server 

> * How do people from the different sites access it? Is it available on the internet?

Yes,  on internet: Service URL. For AirGap env. This workflow may not work. Fallback to Manual workflow 

> * What are the risks of running such a service? How do we secure it. It seems that if it gets hacked the whole system is compromised.

True,  Security is important. This is part of IT Security. IT no difference from other web service site. The enterprise security would apply, but not part of FLARE offers

>   From a security aspect it seems we have a tradeoff between sharing the security keys out of band at the start and the fact that we now have a very important server component that absolutely needs to be secure.

Actually, with manual approach ( updated),  we don't need to share private key either.  so we dont trade off that part. 

Part of trade off, automation vs. manual.  
- Automation: you need secure certification service infrastructure; 
- manual: you place the trust on the individual (Project Admin) and his machine used to provisioning not be compromised

docs/design/enrollment_design.md
```diff
@@ -0,0 +1,835 @@
...
+  2. Submits CSR to Certificate Service with token
+  3. Receives signed cert, connects to FL server)
+```
+
+**Key characteristics:**
+- Token-based: Project Admin sends a short token string, not certificates
+- Fully automated: no human in the CSR signing loop
+- Root CA key lives on the Certificate Service, not with Project Admin
+- `nvflare package` embeds enrollment info; `start.sh` auto-enrolls
+- K8s-friendly: inject token via Secret, pods auto-enroll at startup
```
pcnudde, line 211: I see a problem if we set the token to expire quickly (days). If the k8s is not started on time the will not get the keys. But we can't let the token remain valid for a long time, because then it really becomes the private key.

chesterxgchen:
Handled the latest open review threads in `docs/design/enrollment_design.md` (pushed in `7cb4c93bf`).

Updates made:
- Split ambiguous server packaging example into explicit Manual vs Auto-Scale commands.
- Corrected SQLite/PostgreSQL table wording to describe concurrency model accurately.
- Added `rate_limit` settings in Certificate Service config example and clarified ingress/gateway fallback.
- Added token TTL operational guidance for delayed startup scenarios (for example K8s scheduling delays) with just-in-time issuance/re-issue guidance.

All currently open review threads are resolved.

PR #4201, opened 2026-02-18
Cherry pick 4181 to main

nvkevlu:
Cherry pick 4181 to main.

### Description

Cherry pick 4181 to main.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/web/package-lock.json
+++ b/web/package-lock.json
@@ -1424,9 +1424,9 @@
       "license": "MIT"
     },
     "node_modules/@rollup/rollup-android-arm-eabi": {
-      "version": "4.53.3",
-      "resolved": "https://registry.npmjs.org/@rollup/rollup-android-arm-eabi/-/rollup-android-arm-eabi-4.53.3.tgz",
-      "integrity": "sha512-mRSi+4cBjrRLoaal2PnqH82Wqyb+d3HsPUN/W+WslCXsZsyHa9ZeQQX/pQsZaVIWDkPcpV6jJ+3KLbTbgnwv8w==",
+      "version": "4.59.0",
+      "resolved": "https://registry.npmjs.org/@rollup/rollup-android-arm-eabi/-/rollup-android-arm-eabi-4.59.0.tgz",
+      "integrity": "sha512-upnNBkA6ZH2VKGcBj9Fyl9IGNPULcjXRlg0LLeaioQWueH30p6IXtJEbKAgvyv+mJaMxSm1l6xwDXYjpEMiLMg==",
       "cpu": [
         "arm"
       ],
@@ -1437,9 +1437,9 @@
       ]
     },
     "node_modules/@rollup/rollup-android-arm64": {
-      "version": "4.53.3",
-      "resolved": "https://registry.npmjs.org/@rollup/rollup-android-arm64/-/rollup-android-arm64-4.53.3.tgz",
-      "integrity": "sha512-CbDGaMpdE9sh7sCmTrTUyllhrg65t6SwhjlMJsLr+J8YjFuPmCEjbBSx4Z/e4SmDyH3aB5hGaJUP2ltV/vcs4w==",
+      "version": "4.59.0",
+      "resolved": "https://registry.npmjs.org/@rollup/rollup-android-arm64/-/rollup-android-arm64-4.59.0.tgz",
+      "integrity": "sha512-hZ+Zxj3SySm4A/DylsDKZAeVg0mvi++0PYVceVyX7hemkw7OreKdCvW2oQ3T1FMZvCaQXqOTHb8qmBShoqk69Q==",
       "cpu": [
         "arm64"
       ],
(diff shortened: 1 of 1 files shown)


nvkevlu:
The other PR into 2.7 was created to match this one, however, there were further comments from copilot that resulted in additional updates. This one actually needs to be further updated to match those changes to have the right behavior: get_prop_detail() + reuse existing (private, sticky) when updating, so the value is updated every round and set_prop() never fails or warns.

nvflare/app_common/workflows/base_model_controller.py
```diff
@@ -376,15 +376,47 @@ def sample_clients(self, num_clients: int = None) -> List[str]:
...
+        Uses existing (private, sticky) attributes when the prop is already set so set_prop()
+        accepts the update without warning; otherwise uses private=True, sticky=False. Required
+        for flows like FedAvg that do not set CURRENT_ROUND in fl_ctx before send; downstream
+        (e.g. aggregators) rely on it.
+        """
+        if not data:
+            return
+        if data.current_round is not None:
+            detail = self.fl_ctx.get_prop_detail(AppConstants.CURRENT_ROUND)
+            if detail is not None:
```
YuanTingHsieh: i know this is already merged into 2.7,
Could we instead add a method`update_prop_value` to fl context so we can just do:

```
def set_fl_context(self, data: FLModel):
    if not data:
        return
    if data.current_round is not None:
        if self.fl_ctx.get_prop(AppConstants.CURRENT_ROUND) is not None:
            self.fl_ctx.update_prop_value(AppConstants.CURRENT_ROUND, data.current_round)
        else:
            self.fl_ctx.set_prop(AppConstants.CURRENT_ROUND, data.current_round, private=True, sticky=False)
    if data.total_rounds is not None:
        if self.fl_ctx.get_prop(AppConstants.NUM_ROUNDS) is not None:
            self.fl_ctx.update_prop_value(AppConstants.NUM_ROUNDS, data.total_rounds)
        else:
            self.fl_ctx.set_prop(AppConstants.NUM_ROUNDS, data.total_rounds, private=True, sticky=False)

```

nvkevlu: thanks, done

nvflare/app_common/workflows/base_model_controller.py
```diff
@@ -382,14 +382,33 @@ def sample_clients(self, num_clients: int = None) -> List[str]:
...
+        if self.fl_ctx.get_prop(key) is not None:
+            if not self.fl_ctx.update_prop_value(key, value):
+                self.debug(f"Failed to update {label} in fl_ctx via update_prop_value; retrying with set_prop.")
+                if not self.fl_ctx.set_prop(key, value, private=True, sticky=True):
+                    self.warning(f"set_prop also failed to update {key}; value may be stale.")
+        else:
+            if not self.fl_ctx.set_prop(key, value, private=True, sticky=True):
+                self.warning(f"set_prop failed to set {key}; value may be stale.")
+
     def set_fl_context(self, data: FLModel):
```
YuanTingHsieh, line 385: what was the issue this PR try to solve?
I re-read the whole thing again and think we can just did:

```
def set_fl_context(self, data: FLModel):
    if not data:
        return
    if data.current_round is not None:
        self.fl_ctx.set_prop(AppConstants.CURRENT_ROUND, data.current_round, private=True, sticky=True)
    if data.total_rounds is not None:
        self.fl_ctx.set_prop(AppConstants.NUM_ROUNDS, data.total_rounds, private=True, sticky=False)

```

I apologize for suggesting that "update_prop_value", I think it makes code more complicated we should not use that.

nvkevlu: This was originally just a cherry pick, but it has drifted so far away from the original one due to continuous additional things to address from PR comments

nvkevlu: Now it should be cleaned up and if there are not more greptile comments, we should be able to merge it

nvflare/app_common/workflows/base_model_controller.py
```diff
@@ -383,13 +383,31 @@ def sample_clients(self, num_clients: int = None) -> List[str]:
...
         else:
             self.debug("The FLModel data does not contain the current_round information.")
-        if data and data.total_rounds is not None:
-            self.fl_ctx.set_prop(AppConstants.NUM_ROUNDS, data.total_rounds, private=True, sticky=True)
+        if data.total_rounds is not None:
+            self.fl_ctx.set_prop(
+                AppConstants.NUM_ROUNDS,
+                data.total_rounds,
+                private=True,
+                sticky=True,
```
YuanTingHsieh, line 409: this might need to be False though

YuanTingHsieh: we could either change here or we merge this first then make changes to both 2.7 and main

PR #4202, opened 2026-02-18
[2.7] update to match fix in main

nvkevlu:
Update to match fix in main in #4201 

### Description

Updates nvflare/app_common/workflows/base_model_controller.py to match #4201 after further changes from greptile comments.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -174,3 +174,6 @@ docs/apidocs/
 
 # dependency marker
 .deps_marker
+
+# local test temp dir
+.tmp/
--- a/tests/unit_test/app_common/ccwf/test_swarm_self_message_deadlock.py
+++ b/tests/unit_test/app_common/ccwf/test_swarm_self_message_deadlock.py
@@ -37,6 +37,7 @@ import unittest
 from types import SimpleNamespace
 from unittest import mock
 
+from nvflare.apis.controller_spec import Task
 from nvflare.apis.fl_constant import ReturnCode as FLReturnCode
 from nvflare.apis.fl_context import FLContextManager
 from nvflare.apis.shareable import Shareable, make_reply
@@ -546,7 +547,8 @@ class TestBroadcastMultiRequestsToSelf(unittest.TestCase):
 
 
 class TestSwarmResultSubmissionFix(unittest.TestCase):
-    def test_local_submit_when_aggregator_is_self(self):
+    @staticmethod
+    def _make_swarm_controller():
         class _DummyGatherer:
             def __init__(self, **kwargs):
                 self.for_round = kwargs.get("for_round", 0)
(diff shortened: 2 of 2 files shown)


nvflare/app_common/workflows/base_model_controller.py
```diff
@@ -440,18 +440,47 @@ def sample_clients(self, num_clients: int = None) -> List[str]:
...
+        (e.g. aggregators) rely on it.
         """
         if not data:
             return
-        if data.current_round is not None and self.fl_ctx.get_prop(AppConstants.CURRENT_ROUND) is None:
-            self.fl_ctx.set_prop(AppConstants.CURRENT_ROUND, data.current_round, private=True, sticky=False)
-        if data.total_rounds is not None and self.fl_ctx.get_prop(AppConstants.NUM_ROUNDS) is None:
-            self.fl_ctx.set_prop(AppConstants.NUM_ROUNDS, data.total_rounds, private=True, sticky=False)
+        if data.current_round is not None:
+            detail = self.fl_ctx.get_prop_detail(AppConstants.CURRENT_ROUND)
```
chesterxgchen, line 453: whats difference between fl_ctx.get_prop() vs. fl_ctx.get_prop_details() ?

nvflare/app_common/workflows/base_model_controller.py
```diff
@@ -440,18 +440,47 @@ def sample_clients(self, num_clients: int = None) -> List[str]:
...
+            if detail is not None:
+                self.fl_ctx.set_prop(
```
chesterxgchen, line 455: if details (current round) is not One, we set it back  ? essentially from prop_details => prop 
why this is important. ?

nvflare/app_common/workflows/base_model_controller.py
```diff
@@ -440,18 +440,47 @@ def sample_clients(self, num_clients: int = None) -> List[str]:
...
+                )
+            else:
+                self.fl_ctx.set_prop(
+                    AppConstants.CURRENT_ROUND,
+                    data.current_round,
+                    private=True,
+                    sticky=False,
+                )
+        if data.total_rounds is not None:
+            detail = self.fl_ctx.get_prop_detail(AppConstants.NUM_ROUNDS)
```
chesterxgchen, line 469: same question here

nvflare/app_common/workflows/base_model_controller.py
nvkevlu: get_prop(key, default=None) returns only the value for that key (or default if the key is missing).

get_prop_detail(key) returns a dict with the value and its metadata, or None if the key doesn’t exist:
{"value": <value>, "private": bool, "sticky": bool}.
So you get both the current value and the (private, sticky) attributes that were used when the prop was set.

nvflare/app_common/workflows/base_model_controller.py
nvkevlu: We’re updating the value (to data.current_round / data.total_rounds from the FLModel) while reusing the existing (private, sticky) attributes from get_prop_detail.

set_prop() does not allow changing (private, sticky) for an existing key. If we call set_prop(..., sticky=False) when the prop was first set with sticky=True, the context logs a warning, returns False, and does not update the value. So the round info would never get refreshed. By reusing the existing attributes from get_prop_detail, we only change the value, so set_prop() accepts the update and the controller can keep CURRENT_ROUND/NUM_ROUNDS in sync with the FLModel each round.

chesterxgchen: if the prop is not allow to update, does the self.fl_ctx.set_prop() fail in the 2nd round, where the same key already has a value

nvkevlu: No, set_prop() does not fail on the 2nd round just because the key already has a value.

From fl_context.py:

Same attributes: If the key already exists and you call set_prop(key, new_value, private=P, sticky=S) with the same (P, S) as the existing prop, the code does not return early. It goes through and does self.props[key] = {V: value, M: mask}, so the value is updated. So round 2 (and every later round) can update the value as long as (private, sticky) stay the same.

Different attributes: It only returns False (and does not update) when the new (private, sticky) are different from the existing ones (mask != existing_mask). So it “fails” only when you try to change the attributes, not when you re-set the same key with the same attributes.

So:
If we always use private=True, sticky=False and we were the ones who set it in round 1, round 2 and later updates succeed.

If some other code set it in round 1 with e.g. sticky=True, and we then call set_prop(..., sticky=False), that would fail (warning + no update). Using get_prop_detail() and reusing the existing private/sticky avoids that case and keeps the 2nd (and later) round updates working.

PR #4204, opened 2026-02-19
[2.7] Fix RxTask self-deadlock on stream error cleanup

pcnudde:
## Summary
- avoid calling `RxTask.stop()` while holding `RxTask.map_lock` in `find_or_create_task`
- stop the existing task only after leaving the map lock to prevent self-deadlock
- add a regression unit test for the existing-task error path

### DIFF
--- a/nvflare/app_common/workflows/base_model_controller.py
+++ b/nvflare/app_common/workflows/base_model_controller.py
@@ -440,18 +440,47 @@ class BaseModelController(Controller, FLComponentWrapper, ABC):
         return clients
 
     def set_fl_context(self, data: FLModel):
-        """Set fl_ctx CURRENT_ROUND and NUM_ROUNDS from FLModel when not already set.
+        """Set fl_ctx CURRENT_ROUND and NUM_ROUNDS from FLModel so they stay current each round.
 
-        Only sets when the prop is missing so we do not overwrite workflow-set values or
-        change stickiness (which would trigger warnings). Required for flows like FedAvg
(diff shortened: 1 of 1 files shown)


tests/unit_test/fuel/f3/streaming/byte_receiver_test.py
```diff
@@ -0,0 +1,68 @@
...
+        StreamHeaderKey.STREAM_ID: sid,
+        MessageHeaderKey.ORIGIN: origin,
+    }
+    if error is not None:
+        headers[StreamHeaderKey.ERROR_MSG] = error
+    message.add_headers(headers)
+    return message
+
+
+def test_find_or_create_task_stops_outside_map_lock(monkeypatch):
```
chesterxgchen, line 119: this tested the happy path, i.e. after the fix. 

Can we show the un-happy path first, showing that w/o the fix. its deadlocked

pcnudde: Updated the tests to explicitly cover the pre-fix failure path.

- Added `test_pre_fix_find_or_create_task_would_deadlock` using a deadlock-detecting lock and the original pre-fix logic (`stop()` called while `map_lock` is held).
- The test now demonstrates the same-thread re-acquire on `map_lock` (modeled as `self-deadlock`) deterministically without hanging CI.
- Kept the fixed-path test to verify `stop()` is invoked outside the lock.

Both tests pass locally (`2 passed`).

chesterxgchen (approved):

PR #4205, opened 2026-02-20
[2.7] Add SFM slow-send diagnostics and minimal streaming timeout tuning

pcnudde:
## Summary
- add slow frame-send diagnostics in SFM connection send path (lock wait + send duration)
- reduce default streaming ACK wait from 300s to 120s
- reduce default streaming read timeout from 300s to 120s


chesterxgchen:
replaced with https://github.com/NVIDIA/NVFlare/pull/4206

PR #4206, opened 2026-02-20
[2.7] Mitigate F3 streaming Head-of-Line (HOL) stalls and add guardrails

chesterxgchen:
Many Thanks to @GeorgeWang-nv to find the issues and perform the root cause analysis on this issues, suggested possible fixes

## 1) Root cause: connection-level Head-of-Line (HOL) blocking
- Large byte streams can freeze when a blocking socket send holds the per-connection SFM send lock.
- While that lock is held, unrelated traffic on the same connection (including progress and coordination frames) is serialized behind it.
- This creates **Head-of-Line (HOL) blocking** at connection level and can manifest as long apparent stalls.

### Critical fix A: bound low-level send blocking time
- Add a bounded send timeout in the socket send path so one blocked send cannot hold the connection lock indefinitely.
- Why this way: minimally invasive, directly targets lock-hold source, and fails deterministically instead of hanging.

### Critical fix B: ACK-progress watchdog in streaming sender
- Track ACK forward progress and fail fast when ACK offset does not advance for a configurable interval.
- Why this way: protects sender flow control from waiting too long on non-progressing streams, independent of generic ACK wait.

### Critical fix C: SFM stalled-send detection with guarded recovery
- Monitor send stall duration in heartbeat monitor; optionally close the stalled connection to recover.
- Add a consecutive-check threshold before close to reduce false positives from transient jitter.
- Why this way: staged recovery lever (warn-only first, then auto-close) balances reliability and operational safety.

### Configuration surface added
- `streaming_send_timeout`
- `streaming_ack_progress_timeout`
- `streaming_ack_progress_check_interval`
- `sfm_send_stall_timeout`
- `sfm_close_stalled_connection`
- `sfm_send_stall_consecutive_checks`

## 3) Test strategy and coverage
- Added focused unit tests for each mitigation path with both positive and negative cases.

### Socket send timeout tests
- Config application, partial-write success, non-writable timeout, closed-send path, and closing-state suppression.

### ACK watchdog tests
- ACK progress timestamp update, progress-allowed completion, no-progress timeout, legacy ack_wait behavior, and fast-fail precedence when progress timeout is small.

### Stall monitor plus guard tests
- Warn-only behavior, close-enabled recovery, close-disabled non-recovery, heartbeat behavior when healthy, send-state reset on success/exception, warning emission, no-warning in healthy flow, and intermittent-stall false-alarm suppression via consecutive guard.

### Validation run
- `pytest` on the three new F3 test modules: **24 passed**.

## 4) User/operator guide updates
- Added a new section in `docs/user_guide/timeout_troubleshooting.rst`:
  - Runtime defaults for new stall-related parameters.
  - Recommended rollout: warn-only first, then auto-close only if repeated stall warnings persist.
  - Suggested values from discussion: `sfm_send_stall_timeout=75`, `sfm_send_stall_consecutive_checks=3`.
  - Log interpretation guidance and false-alarm suppression expectations.

## 5) Additional notes and risk profile
- This change prefers deterministic failure/recovery over silent long hangs.
- Potential side effect when auto-close is enabled: transient connection resets under severe network instability; mitigated by conservative timeout plus consecutive-check guard.
- Defaults remain conservative (`sfm_close_stalled_connection=false`) so operators can observe first and enable recovery when needed.
- Scope is intentionally focused on the critical Head-of-Line (HOL) mitigation path and operational guardrails.

## Test plan
- [x] `python3 -m pytest tests/unit_test/fuel/f3/drivers/socket_conn_timeout_test.py tests/unit_test/fuel/f3/streaming/byte_streamer_ack_watchdog_test.py tests/unit_test/fuel/f3/sfm/sfm_stall_monitor_test.py`
- [x] `python3 -m black <changed_python_files>`
- [x] `python3 -m isort <changed_python_files>`
- [x] `python3 -m flake8 <changed_python_files>`

## Latest Timeout Hardening Update
- Close socket connections on `send_frame` timeout to prevent frame-boundary desync after partial writes.
- Preserve specific send error codes in `send_frame` for common socket exceptions:
  - map timeout-like exceptions to `CommError.TIMEOUT` (and close connection),
  - map closed-socket exceptions to `CommError.CLOSED`,
  - keep `CommError.ERROR` as fallback for unknown exceptions.
- Add positive and negative unit tests in `socket_conn_timeout_test.py`:
  - Positive: timeout after partial write closes the connection.
  - Positive: `TimeoutError` -> `TIMEOUT`; `BrokenPipeError` -> `CLOSED`.
  - Negative: successful partial-write sends do not close; non-timeout `CommError.CLOSED` does not force close; unknown exceptions remain `ERROR`.
- Document stall timing relationship in `timeout_troubleshooting.rst`, including the practical close window formula and outer-timeout sizing guideline.

### Validation
- `python3 -m pytest tests/unit_test/fuel/f3/drivers/socket_conn_timeout_test.py -q`
- Result: `10 passed`

### DIFF
--- a/nvflare/fuel/f3/drivers/socket_conn.py
+++ b/nvflare/fuel/f3/drivers/socket_conn.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import errno
 import logging
 import select
 import socket
@@ -66,7 +67,39 @@ class SocketConnection(Connection):
                 raise
         except Exception as ex:
             if not self.closing:
+                if self._is_timeout_exception(ex):
+                    self.close()
+                    raise CommError(
+                        CommError.TIMEOUT,
+                        f"send_frame timeout on conn {self}: {secure_format_exception(ex)}",
+                    )
+                if self._is_closed_socket_exception(ex):
+                    raise CommError(
+                        CommError.CLOSED,
+                        f"Connection {self.name} is closed while sending: {secure_format_exception(ex)}",
+                    )
                 raise CommError(CommError.ERROR, f"Error sending frame on conn {self}: {secure_format_exception(ex)}")
+
+    @staticmethod
+    def _is_timeout_exception(ex: Exception) -> bool:
+        return isinstance(ex, (TimeoutError, socket.timeout))
+
+    @staticmethod
+    def _is_closed_socket_exception(ex: Exception) -> bool:
+        if isinstance(ex, (BrokenPipeError, ConnectionResetError, ConnectionAbortedError)):
+            return True
+
+        if isinstance(ex, OSError):
+            return ex.errno in {
+                errno.EPIPE,
+                errno.ECONNRESET,
+                errno.ENOTCONN,
+                errno.ECONNABORTED,
+                errno.EBADF,
+                errno.ESHUTDOWN,
+            }
+
+        return False
 
     def _send_with_timeout(self, frame: BytesAlike, timeout_sec: float):
         view = memoryview(frame)
--- a/tests/unit_test/fuel/f3/drivers/socket_conn_timeout_test.py
+++ b/tests/unit_test/fuel/f3/drivers/socket_conn_timeout_test.py
@@ -149,6 +149,56 @@ class TestSocketConnectionSendTimeout:
         assert close_calls["count"] == 0
         assert conn.closing is False
 
+    def test_send_frame_maps_timeouterror_to_timeout_and_closes(self, monkeypatch):
+        conn, _ = self._make_conn(monkeypatch, timeout_sec=1.0)
+
+        def _raise_timeout(*_args, **_kwargs):
+            raise TimeoutError("socket timed out")
+
+        monkeypatch.setattr(conn, "_send_with_timeout", _raise_timeout)
+
+        close_calls = {"count": 0}
+        original_close = conn.close
+
+        def _close_spy():
+            close_calls["count"] += 1
+            original_close()
+
+        monkeypatch.setattr(conn, "close", _close_spy)
+
+        with pytest.raises(CommError) as ex:
+            conn.send_frame(b"abc")
+
+        assert ex.value.code == CommError.TIMEOUT
+        assert "timeout" in str(ex.value).lower()
+        assert close_calls["count"] == 1
+        assert conn.closing is True
+
+    def test_send_frame_maps_broken_pipe_to_closed(self, monkeypatch):
+        conn, _ = self._make_conn(monkeypatch, timeout_sec=1.0)
+
+        def _raise_closed(*_args, **_kwargs):
+            raise BrokenPipeError("broken pipe")
+
+        monkeypatch.setattr(conn, "_send_with_timeout", _raise_closed)
+
+        close_calls = {"count": 0}
+        original_close = conn.close
+
+        def _close_spy():
+            close_calls["count"] += 1
+            original_close()
+
+        monkeypatch.setattr(conn, "close", _close_spy)
+
+        with pytest.raises(CommError) as ex:
+            conn.send_frame(b"abc")
+
+        assert ex.value.code == CommError.CLOSED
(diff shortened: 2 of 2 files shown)


nvflare/fuel/f3/sfm/sfm_conn.py
```diff
@@ -145,7 +147,19 @@ def send_frame(self, prefix: Prefix, headers: Optional[dict], payload: Optional[
         log.debug(f"Sending frame: {prefix} on {self.conn}")
         # Only one thread can send data on a connection. Otherwise, the frames may interleave.
         with self.lock:
-            self.conn.send_frame(buffer)
+            with self.send_state_lock:
+                self.send_started_at = time.time()
```
pcnudde: should use monotonic or perf_counter timer here. not time.time

nvflare/fuel/f3/sfm/sfm_conn.py
```diff
@@ -145,7 +147,19 @@ def send_frame(self, prefix: Prefix, headers: Optional[dict], payload: Optional[
...
+                self.conn.send_frame(buffer)
+            finally:
+                with self.send_state_lock:
+                    self.send_started_at = 0.0
+
+    def get_send_stall_seconds(self) -> float:
+        with self.send_state_lock:
+            if self.send_started_at <= 0.0:
+                return 0.0
+            return time.time() - self.send_started_at
```
pcnudde: should use monotonic or perf_counter timer here. not time.time

tests/unit_test/fuel/f3/drivers/socket_conn_timeout_test.py
```diff
@@ -0,0 +1,151 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+
```
pcnudde, line 14: should this be the complete license? (same for other tests)

nvflare/fuel/f3/streaming/byte_streamer.py
```diff
@@ -87,8 +88,12 @@ def __init__(
         self.stream_future = StreamFuture(self.sid, task_handle=self)
         self.stream_future.set_size(stream.get_size())
 
-        self.window_size = CommConfigurator().get_streaming_window_size(STREAM_WINDOW_SIZE)
-        self.ack_wait = CommConfigurator().get_streaming_ack_wait(STREAM_ACK_WAIT)
+        config = CommConfigurator()
+        self.window_size = config.get_streaming_window_size(STREAM_WINDOW_SIZE)
+        self.ack_wait = config.get_streaming_ack_wait(STREAM_ACK_WAIT)
+        self.ack_progress_timeout = config.get_streaming_ack_progress_timeout(60.0)
+        self.ack_progress_check_interval = config.get_streaming_ack_progress_check_interval(5.0)
```
pcnudde: probably should be max(0.01, config.get_streaming_ack_progress_check_interval(5.0)). 
If this is mistakingly configured as zero then the wait_timeout = 0 and the wait on line 132 will not wait and busy spin the cpu in that while loop.

nvflare/fuel/f3/drivers/socket_conn.py
```diff
@@ -52,11 +56,32 @@ def close(self):
 
     def send_frame(self, frame: BytesAlike):
         try:
-            self.sock.sendall(frame)
+            self._send_with_timeout(frame, self.send_timeout)
+        except CommError:
```
GeorgeWang-nv: **Potential frame corruption after partial-send timeout**

If `_send_with_timeout` has already pushed some bytes via `sock.send()` before the deadline expires, those partial frame bytes are on the wire but the connection remains open. The next `send_frame` call on this connection will interleave new frame data with the leftover bytes of the incomplete frame, corrupting the peer's frame parser.

Consider closing the connection on `CommError.TIMEOUT` to prevent subsequent frames from being misinterpreted:

```python
except CommError as err:
    if err.code == CommError.TIMEOUT:
        self.close()
    if not self.closing:
        raise
```

chesterxgchen: ok, fixed

PR #4172, opened 2026-02-11
[2.7] Add end-to-end download starvation test for stream pool fix

chesterxgchen:
## Summary

Adds an end-to-end integration test that validates the P0 stream thread pool starvation fix introduced in #4171. Unlike the existing unit test (`pool_starvation_test.py`) which uses mocks, this test exercises the **full `download_object()` code path** through real TCP-connected `Cell` objects and blob streaming.

### What the test does

Creates two real `Cell` objects connected via TCP (server + client) and runs **8 concurrent `download_object()` calls**, each producing ~200 chunks (256 bytes each, 50KB total) through the complete flow:

```
download_object()
  -> Cell.send_request()
    -> Cell.__getattr__() routes to _send_request() (stream channel)
      -> _send_one_request()
        -> send_blob() [request]
          -> BlobStreamer -> ByteStreamer -> TxTask._transmit_task (stream_thread_pool)
          -> ByteReceiver._callback_wrapper (stream_thread_pool)
            -> BlobHandler.handle_blob_cb
              -> _read_stream (stream_thread_pool)
              -> _run_blob_cb (callback_thread_pool) <-- THE FIX
                -> Adapter.call()
                  -> future.result() [waits for _read_stream]
                  -> DownloadService._handle_download()
                    -> Downloadable.produce()
                  -> send_blob() [reply]
        -> _process_reply()
          -> Consumer.consume()
```

### Two test variants

| Test class | Behavior | Expected result |
|---|---|---|
| **TestDownloadWithFix** | Uses real production code: blob_cb dispatched to separate callback_thread_pool | All 8/8 downloads succeed |
| **TestDownloadPreFixStarvation** | Patches BlobHandler.handle_blob_cb to call blob_cb **synchronously** on pool worker (pre-fix behavior) + wraps _read_stream with 2s delay (simulating large blob transfer) + uses 4-worker shared pool | 0/8 succeed -- **deadlock confirmed** |

### How the pre-fix starvation is reproduced

The pre-fix test creates cells **after** patching so the registered callbacks capture the patched behavior:

1. `BlobHandler.handle_blob_cb` is monkeypatched to call blob_cb synchronously (instead of submitting to callback_thread_pool)
2. `_read_stream` is wrapped with a 2s `time.sleep()` to simulate the time a real multi-MB model tensor takes to transfer via many transport chunks
3. All streaming modules (blob_streamer, byte_receiver, byte_streamer) use a tiny 4-worker shared pool

**Deadlock mechanism:**
[...]

PR #4209, opened 2026-02-20
[2.7] Fix hierarchical FL startup failures: deployment timeouts, selective client exclusion, and dead-detection debounce

chesterxgchen:
## Problem

Large-scale hierarchical FL jobs (e.g. BERT NER, 144 clients, 6 relays on Frontier) abort in
Round 0 due to a cascading startup failure chain.  The root sequence is:

1. F3 streaming HOL stall (PR #4206) delays deployment ACKs from relay-connected clients
2. **`_deploy_job()`** treats `reply=None` (timeout) as `"unknown"` — not a failure — so
   timed-out clients silently appear to have been deployed
3. **`_start_run()`** tries to start those clients; they again time out, and
   `check_client_replies()` ignores the `None` reply
4. **`_sync_client_jobs()`** fires dead-job notification on the very first heartbeat with
   no startup grace period
5. FedAvg requires 144/144 — one or two missing clients → abort
6. A late-starting CJ crashes with `TypeError: 'NoneType' object is not iterable` when
   `get_job_clients()` receives `None` metadata from an already-aborted job

PRs #4206, #4204, #4174, #4172, #4186, #4211, #4210 (all merged in 2.7.2) address the
transport layer.  This PR addresses the remaining job lifecycle layer.

---

### 1 — `_deploy_job()`: Treat deployment timeout as failure (`job_runner.py`)

**Root bug**: `reply=None` was logged as `"unknown"` and excluded from `failed_clients`,
so timed-out clients counted as "successfully deployed" for the `min_sites` check.

**Fix**: Add timed-out clients to `failed_clients` with a `"deployment timeout"` label.
The existing `min_sites` / `required_sites` logic then correctly decides whether to abort.

### 2 — `check_client_replies()`: Return timed-out clients instead of raising (`admin.py`)

**Root bug**: In strict mode, any timeout raised immediately, aborting the whole job even
when the remaining active clients satisfied `min_sites`.

**Fix**: In strict mode, collect timed-out clients into a return list rather than raising.
Explicit errors (non-OK return code or error body) still raise.  Also fixes the non-strict
mode to use name-keyed dict lookup instead of fragile positional `zip()`.

New signature: `check_client_replies(...) -> List[str]` (timed-out client names; empty = none).

### 3 — `_start_run()`: Selective exclusion with min_sites re-evaluation (`job_runner.py`)

**Root bug**: A start-job timeout under strict mode aborted the entire job with no
tolerance for stragglers within `min_sites` bounds.

**Fix**: Use the returned timed-out list from `check_client_replies()`.  If remaining
active clients >= `min_sites`, log a warning and proceed.  Only abort when below tolerance.

### 4 — `_sync_client_jobs()`: Require-prior-report default changed to `True` (`fed_server.py`)

**Root bug**: `SYNC_CLIENT_JOBS_REQUIRE_PREVIOUS_REPORT` defaulted to `False`, meaning
the bug fix was opt-in and the unsafe behaviour remained the default.

**Fix**: Default changed to `True`.  Operators who want the aggressive legacy detection
can set it to `False` explicitly.

### 5 — `_sync_client_jobs()`: Move `_reported_clients` out of `job_info` dict (`fed_server.py`)

**Root bug**: Positive-observation tracking was stored as `job_info["_reported_clients"]`,
injecting algorithm state into a data dict with no corresponding `RunProcessKey` constant.

**Fix**: Tracking moved to `self._job_reported_clients: Dict[str, set]` on `FederatedServer`.
Stale entries are purged whenever a job is no longer in `run_processes`.

### 6 — `ClientRunManager.get_job_clients()`: Explicit meta validation (`client_run_manager.py`)

Raises `RuntimeError` with a descriptive message instead of an opaque `TypeError` when
`JOB_CLIENTS` is absent or the wrong type.

---

## Configuration Recommendations (No Code Change Needed)

| Setting | Recommended value | Effect |
|---|---|---|
| `FedAvg(min_clients=...)` | 96-98% of `num_clients` | Tolerates a few startup stragglers |
| `runner_sync_timeout` | `120` s | Allows Lustre-backed deployments time to complete |
| `strict_start_job_reply_check` | `true` | Start-job timeouts surfaced, straggler clients excluded |
| `sync_client_jobs_require_previous_report` | `true` (now the default) | Prevents premature dead-job from startup delay |
| `SFM_CLOSE_STALLED_CONNECTION` (PR #4206) | `true` after staging | Disconnects stalled relay connections |

---

## Files Changed

- `nvflare/private/fed/server/job_runner.py` — `_deploy_job()` timeout as failure; `_start_run()` selective exclusion
- `nvflare/private/fed/server/admin.py` — `check_client_replies()` returns timed-out list; dict-keyed non-strict path
- `nvflare/private/fed/server/fed_server.py` — `_sync_client_jobs()` default `True`; `_job_reported_clients` attr; stale cleanup
- `nvflare/private/fed/client/client_run_manager.py` — explicit meta validation in `get_job_clients()`

---

## Test Coverage

New and updated unit tests with both positive and negative cases:

| File | Tests | What they cover |
|---|---|---|
| `admin_test.py` | 8 | Timeout returned not raised; dict lookup; error still raises; reorder OK |
| `job_runner_test.py` | 4 | strict flag wiring; timeout within tolerance → warn; timeout below tolerance → raise |
| `job_runner_deploy_test.py` | 9 (new file) | Timeout counted as failure; OK reply not failed; mixed outcomes; detail label; min_sites with timeouts; integration sequence |
| `fed_server_test.py` | 5 | Default requires-prior-report; legacy explicit-False still fires; tracking in server attr not job_info; stale cleanup |

All 29 targeted unit tests pass.

## Test Plan

- [x] Unit tests for each changed function (positive + negative)
- [x] New `job_runner_deploy_test.py` covering deployment timeout classification end-to-end
- [x] All 29 targeted unit tests pass
- [ ] Hierarchical staging run with all flags at default
- [ ] Hierarchical staging run with `strict_start_job_reply_check=true` and reduced `min_clients`
- [ ] Verify no regression on standard (non-hierarchical) FL jobs

### DIFF
--- a/nvflare/private/fed/server/job_runner.py
+++ b/nvflare/private/fed/server/job_runner.py
@@ -270,7 +270,8 @@ class JobRunner(FLComponent):
             raise RuntimeError(f"Could not start the server App for job: {job_id}.")
 
         replies = engine.start_client_job(job, client_sites, fl_ctx)
-        client_sites_names = list(client_sites.keys())
+        all_client_sites = list(client_sites.keys())
+        active_client_sites = list(all_client_sites)
         strict_start_reply_check = ConfigService.get_bool_var(
             name=ConfigVarName.STRICT_START_JOB_REPLY_CHECK,
             conf=SystemConfigs.APPLICATION_CONF,
@@ -278,12 +279,12 @@ class JobRunner(FLComponent):
         )
         timed_out = check_client_replies(
             replies=replies,
-            client_sites=client_sites_names,
+            client_sites=all_client_sites,
             command=f"start job ({job_id})",
             strict=strict_start_reply_check,
         )
         if timed_out:
-            active_count = len(client_sites_names) - len(timed_out)
+            active_count = len(all_client_sites) - len(timed_out)
 
             # A required site timing out is fatal regardless of min_sites, same as deploy phase.
             if job.required_sites:
@@ -299,20 +300,25 @@ class JobRunner(FLComponent):
             self.log_warning(
                 fl_ctx,
                 f"start job ({job_id}): {len(timed_out)} client(s) timed out at start-job: {timed_out}; "
-                f"{active_count} of {len(client_sites_names)} clients started successfully.",
+                f"{active_count} of {len(all_client_sites)} clients started successfully.",
             )
-            client_sites_names = [c for c in client_sites_names if c not in timed_out]
+            active_client_sites = [c for c in all_client_sites if c not in timed_out]
 
-            # Rebuild to reflect only active participants after exclusion.
-            # Note: timed-out clients remain in run_processes[PARTICIPANTS] but
-            # require_previous_report=True (default) ensures _sync_client_jobs will not
-            # fire dead-job for them unless they first positively report the job running.
-            active_sites = set(client_sites_names)
-            participating_clients = [c.to_dict() for c in job_clients.values() if c.name in active_sites]
+        if not strict_start_reply_check:
+            # In non-strict mode, check_client_replies() does not return timed-out clients.
+            # Build active clients directly from actual replies so JOB_CLIENTS stays accurate.
+            replies_by_client = {r.client_name: r for r in replies}
+            active_client_sites = []
+            for client_name in all_client_sites:
+                client_reply = replies_by_client.get(client_name)
+                if client_reply and client_reply.reply:
+                    active_client_sites.append(client_name)
 
         # Set metadata once, after any timeout exclusion, so it always reflects active participants.
+        active_sites = set(active_client_sites)
+        participating_clients = [c.to_dict() for c in job_clients.values() if c.name in active_sites]
         job.meta[JobMetaKey.JOB_CLIENTS] = participating_clients
-        display_sites = ",".join(client_sites_names)
+        display_sites = ",".join(active_client_sites)
 
         self.log_info(fl_ctx, f"Started run: {job_id} for clients: {display_sites}")
         self.fire_event(EventType.JOB_STARTED, fl_ctx)
--- a/tests/unit_test/private/fed/server/job_runner_test.py
+++ b/tests/unit_test/private/fed/server/job_runner_test.py
@@ -101,6 +101,36 @@ def test_start_run_proceeds_when_timed_out_clients_within_min_sites(mock_get_boo
     assert "timed out" in warning_msg
 
 
+@patch("nvflare.private.fed.server.job_runner.ConfigService.get_bool_var", return_value=False)
+def test_start_run_non_strict_excludes_timed_out_clients_from_meta(mock_get_bool):
+    """Even when strict checking is disabled, JOB_CLIENTS should include only active clients."""
+    runner, fl_ctx, engine, job, _client_sites = _make_runner_inputs()
+
+    site1 = MagicMock()
+    site1.name = "site-1"
+    site1.to_dict.return_value = {"name": "site-1"}
+
+    site2 = MagicMock()
+    site2.name = "site-2"
+    site2.to_dict.return_value = {"name": "site-2"}
+
+    engine.get_job_clients.return_value = {"token-1": site1, "token-2": site2}
+
+    ok_reply = Message(topic="reply", body="ok")
+    ok_reply.set_header(MsgHeader.RETURN_CODE, ReturnCode.OK)
+    req1 = Message(topic="req", body="")
+    req2 = Message(topic="req", body="")
+    engine.start_client_job.return_value = [
+        ClientReply(client_token="token-site-1", client_name="site-1", req=req1, reply=ok_reply),
+        ClientReply(client_token="token-site-2", client_name="site-2", req=req2, reply=None),
+    ]
+
+    client_sites = {"site-1": MagicMock(), "site-2": MagicMock()}
+    runner._start_run(job_id=job.job_id, job=job, client_sites=client_sites, fl_ctx=fl_ctx)
+
+    assert job.meta[JobMetaKey.JOB_CLIENTS] == [{"name": "site-1"}]
+
+
 @patch("nvflare.private.fed.server.job_runner.check_client_replies")
 @patch("nvflare.private.fed.server.job_runner.ConfigService.get_bool_var", return_value=True)
 def test_start_run_raises_when_timed_out_clients_breach_min_sites(mock_get_bool, mock_check_replies):


nvflare/private/fed/server/admin.py
```diff
@@ -77,21 +77,45 @@ def __init__(self, client, req: Message):
...
+                detail = r.reply.body if r.reply.body else f"return code {return_code}"
+                error_msg += f"	{client_name}: {detail}\n"
+                continue
+
+            if isinstance(r.reply.body, str) and ERROR_MSG_PREFIX in r.reply.body:
+                error_msg += f"	{client_name}: {r.reply.body}\n"
+    else:
+        for r, client_name in zip(replies, client_sites):
+            if r.reply and ERROR_MSG_PREFIX in r.reply.body:
+                error_msg += f"	{client_name}: {r.reply.body}\n"
```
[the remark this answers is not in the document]
chesterxgchen: Already addressed — the non-strict path at line 137 now uses `isinstance(r.reply.body, str)` before the `ERROR_MSG_PREFIX` check, consistent with the strict mode path. This was added as part of the dict-keyed lookup refactor (Issue D) in this PR.

chesterxgchen:
Addressed the metadata consistency feedback in commit e0af9acf5.

Changes made:
- `JobRunner._start_run()` now rebuilds `job.meta[JobMetaKey.JOB_CLIENTS]` after timeout exclusion so metadata reflects active participants only.
- Added unit tests to cover:
  - metadata update when clients time out (`test_start_run_updates_job_clients_meta_after_timeout_exclusion`)
  - metadata unchanged when no timeouts (`test_start_run_keeps_job_clients_meta_when_no_timeouts`)
  - deploy->start filtering path now actually executes `_start_run()` and verifies only deployable clients are started.

Validation:
- `pytest -q tests/unit_test/private/fed/server/job_runner_test.py`
- `pytest -q tests/unit_test/private/fed/server/job_runner_deploy_test.py`

chesterxgchen:
Follow-up added in commit df7c06b48:

- Added an integration-style `_start_run` test that uses the real `check_client_replies` timeout path (no mock for reply-check behavior) and verifies `JOB_CLIENTS` is reduced to active clients after timeout exclusion.

Validation on updated test scope:
- `flake8 tests/unit_test/private/fed/server/job_runner_test.py`
- `pytest -q tests/unit_test/private/fed/server/job_runner_test.py tests/unit_test/private/fed/server/job_runner_deploy_test.py` (16 passed)

nvflare/private/fed/server/fed_server.py
pcnudde, line 886: Not sure if this is a problem. But I think this run_processes list is not the filtered out list. So timed-out clients will still show up here.

nvflare/private/fed/server/fed_server.py
```diff
@@ -794,22 +800,59 @@ def client_heartbeat(self, request: Message) -> Message:
 
     def _sync_client_jobs(self, request, client_token):
         # jobs that are running on client but not on server need to be aborted!
-        client_jobs = request.get_header(CellMessageHeaderKeys.JOB_IDS)
-        server_jobs = self.engine.run_processes.keys()
-        jobs_need_abort = list(set(client_jobs).difference(server_jobs))
+        client_jobs = request.get_header(CellMessageHeaderKeys.JOB_IDS) or []
```
IsaacYangSLA: Not sure if this `or [] ` is needed.  The following one set client_jobs to [] is this get doesn't return correct types.

nvflare/private/fed/server/fed_server.py
```diff
@@ -794,22 +800,59 @@ def client_heartbeat(self, request: Message) -> Message:
...
+        require_previous_report = ConfigService.get_bool_var(
+            name=ConfigVarName.SYNC_CLIENT_JOBS_REQUIRE_PREVIOUS_REPORT,
+            conf=SystemConfigs.APPLICATION_CONF,
+            default=True,
+        )
+
+        # Remove stale tracking entries for jobs that are no longer running.
+        for stale_job_id in list(self._job_reported_clients.keys()):
+            if stale_job_id not in server_jobs:
+                del self._job_reported_clients[stale_job_id]
```
IsaacYangSLA: This may be too deep.  But _sync_client_jobs is a callback for HEART_BEAT.  Therefore, it could be multi-threaded and thus may cause issue.

nvflare/private/fed/server/admin.py
```diff
@@ -77,20 +77,71 @@ def __init__(self, client, req: Message):
...
-        if r.reply and ERROR_MSG_PREFIX in r.reply.body:
-            error_msg += f"\t{client_name}: {r.reply.body}\n"
-    if error_msg != "":
+    timed_out_clients = []
+    replies_by_client = {r.client_name: r for r in replies}
+
+    if strict:
+        missing_clients = [c for c in client_sites if c not in replies_by_client]
+        if missing_clients:
+            display_missing = ", ".join(missing_clients)
```
nvidianz: The missing_clients can be printed directly. It will show up as 
['site-1', 'site-2']

nvflare/private/fed/server/admin.py
```diff
@@ -77,20 +77,71 @@ def __init__(self, client, req: Message):
         self.req = req
 
 
-def check_client_replies(replies: List[ClientReply], client_sites: List[str], command: str):
+def check_client_replies(
+    replies: List[ClientReply], client_sites: List[str], command: str, strict: bool = False
```
nvidianz, line 81: The replies and client_sites can be zipped together into a dict, same as what send_messages() returns.

nvflare/private/fed/server/admin.py
```diff
@@ -77,20 +77,71 @@ def __init__(self, client, req: Message):
...
+                timed_out_clients.append(client_name)
+                continue
+
+            return_code = r.reply.get_header(MsgHeader.RETURN_CODE, ReturnCode.OK)
+            if return_code != ReturnCode.OK:
+                detail = r.reply.body if r.reply.body else f"return code {return_code}"
+                error_msg += f"\t{client_name}: {detail}\n"
+                continue
+
+            if isinstance(r.reply.body, str) and ERROR_MSG_PREFIX in r.reply.body:
```
nvidianz: Should check for prefix here,

r.reply.body.startswith(ERROR_MSG_PREFIX)

nvflare/private/fed/server/fed_server.py
```diff
@@ -794,22 +800,59 @@ def client_heartbeat(self, request: Message) -> Message:
...
+        jobs_need_abort = list(client_jobs.difference(server_jobs))
+
+        require_previous_report = ConfigService.get_bool_var(
+            name=ConfigVarName.SYNC_CLIENT_JOBS_REQUIRE_PREVIOUS_REPORT,
+            conf=SystemConfigs.APPLICATION_CONF,
+            default=True,
+        )
+
+        # Remove stale tracking entries for jobs that are no longer running.
+        for stale_job_id in list(self._job_reported_clients.keys()):
```
nvidianz: keys() is not needed here, dict iterates over keys by default. So this is the same as,

for stale_job_id in self._job_reported_clients:

IsaacYangSLA: I don't think for stale_job in self._job_reported_clients: works well with line 820.  Deleting some items in a collection which is used in a for-loop is risky, even in a single-threaded case.

chesterxgchen: good point

nvflare/private/fed/server/admin.py
chesterxgchen: Fixed in latest commit — `missing_clients` is now used directly in the f-string, which renders as `['site-1', 'site-2']`.

nvflare/private/fed/server/admin.py
chesterxgchen: Fixed in latest commit — the non-strict path now uses `replies_by_client = {r.client_name: r for r in replies}` and iterates over `client_sites` with a dict lookup, consistent with strict mode.

nvflare/private/fed/server/admin.py
chesterxgchen: Fixed in latest commit — the non-strict path now uses `r.reply.body.startswith(ERROR_MSG_PREFIX)` and also adds an `isinstance(r.reply.body, str)` guard before it.

nvflare/private/fed/server/fed_server.py
chesterxgchen: Fixed in latest commit — removed the `or []`. The `isinstance` check on the next line already handles `None` and any other invalid type by setting `client_jobs = []`, so the guard was redundant.

nvflare/private/fed/server/fed_server.py
chesterxgchen: Good catch. `_sync_client_jobs` is called inside `client_heartbeat`, which wraps its entire body in `with self.engine.new_context() as fl_ctx:`. That context manager acquires the engine lock, so concurrent heartbeat callbacks serialize at that point — `_sync_client_jobs` is effectively single-threaded with respect to `run_processes` and `_reported_clients`. No additional locking needed.

nvflare/private/fed/server/fed_server.py
chesterxgchen: This was addressed as part of the `_job_reported_clients` refactor — the server-level dict was replaced by per-job `_reported_clients` sets stored inside `job_info`. The loop in question no longer uses `.keys()`.

chesterxgchen: Resolved by the refactor — `self._job_reported_clients` (the server-level dict that was mutated during iteration) has been replaced by per-job `_reported_clients` sets embedded in `job_info`. The current loop iterates over a snapshot set (`server_jobs.difference(client_jobs)`) and never modifies the collection it's iterating over.

nvflare/private/fed/server/fed_server.py
chesterxgchen: Not a problem — `run_processes` represents jobs the *server* considers active. A client that timed out during `_start_run` is excluded from `JOB_CLIENTS` metadata but the job itself remains in `run_processes` (it still runs with the subset of clients that did respond). The `jobs_need_abort` list is for telling a *client* to abort a job the server is no longer tracking, which is a separate concern.

nvflare/private/fed/server/admin.py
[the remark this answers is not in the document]
chesterxgchen: Fixed in latest commit — the strict path now also uses `r.reply.body.startswith(ERROR_MSG_PREFIX)`, making both strict and non-strict paths consistent.

nvflare/private/fed/server/fed_server.py
chesterxgchen: Addressed in latest commit — added `self._job_reported_clients_lock = threading.Lock()` in `__init__` and wrapped all `_job_reported_clients` access in `_sync_client_jobs` with `with self._job_reported_clients_lock:`. `_notify_dead_job` calls are deferred outside the lock to avoid holding it during I/O.

nvflare/private/fed/server/fed_server.py
chesterxgchen: Added a comment in `job_runner.py` explaining this: timed-out clients remain in `run_processes[PARTICIPANTS]` but with `require_previous_report=True` (the new default) `_sync_client_jobs` will never fire dead-job for a client that never positively reported the job running — which is exactly the case for a client that timed out at start-job.

nvflare/private/fed/server/job_runner.py
```diff
@@ -255,7 +269,34 @@ def _start_run(self, job_id: str, job: Job, client_sites: Dict[str, DispatchInfo
...
+            command=f"start job ({job_id})",
+            strict=strict_start_reply_check,
+        )
+        if timed_out:
+            active_count = len(client_sites_names) - len(timed_out)
+            if job.min_sites and active_count < job.min_sites:
+                raise RuntimeError(
+                    f"start job ({job_id}): {len(timed_out)} client(s) timed out and remaining "
+                    f"{active_count} < min_sites {job.min_sites}: {timed_out}"
+                )
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in commit b32696b2e. `_start_run()` now checks `required_sites` before `min_sites`, consistent with the deploy phase: if any timed-out client is in `job.required_sites` the job aborts with a clear error even when `active_count >= min_sites`. Added tests `test_start_run_raises_when_required_site_times_out` and `test_start_run_proceeds_when_non_required_site_times_out`.

nvflare/private/fed/server/job_runner.py
```diff
@@ -255,7 +269,37 @@ def _start_run(self, job_id: str, job: Job, client_sites: Dict[str, DispatchInfo
...
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in commit b32696b2e. Added the `required_sites` check (consistent with the deploy phase) and also cleaned up the double metadata assignment — `JOB_CLIENTS` is now set once after timeout resolution instead of being assigned at line 265 and potentially overwritten later.

chesterxgchen:
Addressed in commit b32696b2e:

- **required_sites check**: `_start_run()` now validates `required_sites` on timeout before checking `min_sites`, consistent with the deploy phase.
- **Metadata set twice**: Removed the early `job.meta[JOB_CLIENTS] = participating_clients` assignment. `JOB_CLIENTS` is now set exactly once at the end of the timeout-exclusion block, always reflecting the actual active participants.

nvidianz (dismissed):

holgerroth:
@greptileai review again

nvflare/private/fed/server/job_runner.py
```diff
@@ -248,14 +262,53 @@ def _start_run(self, job_id: str, job: Job, client_sites: Dict[str, DispatchInfo
         # job_clients is a dict of: token => Client
         assert isinstance(job_clients, dict)
         participating_clients = [c.to_dict() for c in job_clients.values()]
-        job.meta[JobMetaKey.JOB_CLIENTS] = participating_clients
         err = engine.start_app_on_server(fl_ctx, job=job, job_clients=job_clients)
         if err:
             raise RuntimeError(f"Could not start the server App for job: {job_id}.")
 
         replies = engine.start_client_job(job, client_sites, fl_ctx)
```
pcnudde, line 272: you call start_client_job prior to setting the job.meta[JobMetaKey.JOB_CLIENTS]. But start_client_job sets the request.set_header with job_meta, so that will not include the JOB_CLIENTS. So we probably need to keep the job.meta[JobMetaKey.JOB_CLIENTS] = participating_clients (old line 251) and then later overwrite it. 

I assume that in the current code all the clients should fail.
client_run_manager.py:326:
```
job_clients = job_meta.get(JobMetaKey.JOB_CLIENTS)
if job_clients is None:
    raise RuntimeError(f"missing {JobMetaKey.JOB_CLIENTS} in job meta")
```

Is there any integration tests that covers this?

pcnudde: @greptileai can you confirm.

pcnudde (approved):

PR #4210, opened 2026-02-20
[2.7] Pass-Through: Zero Tensor Copy at CJ for Large-Model Federated Training

chesterxgchen:
## Summary

This PR introduces the **pass-through architecture** for `ClientAPILauncherExecutor`, eliminating tensor materialisation at the CJ (Client Job) process when large models are exchanged between the FL server and a subprocess agent.

In large-model federated learning (e.g., 7B–70B LLM fine-tuning), the CJ process today acts as a blind relay that fully deserializes and re-serializes every tensor it receives from the FL server before forwarding to the subprocess.  For a 70B float16 model, this consumes ~140 GB of CJ memory and requires two complete network transfers.  B1 pass-through removes both costs.

---

## Problem Description

NVFlare's multi-hop execution path for `launch_external_process=True` looks like:

```
FL Server ──serialize──▶ CJ process ──re-serialize──▶ Subprocess agent
```

### Before this PR

Each tensor in the global model is handled as follows at CJ:

1. **Server** serializes the model and creates a download transaction (tensor data lives on the server).
2. **CJ** fully *downloads* every tensor from the server into its own heap, materialising the complete model in CJ memory.
3. **CJ** re-serializes the model for the subprocess, creating a *new* download transaction — the subprocess then downloads from CJ.

For large models, this means:
- **CJ peak memory = full model size** (potentially 100s of GB).
- **Two full network transfers**: server → CJ, then CJ → subprocess.
- **CJ becomes a throughput bottleneck** and an OOM risk for any model that doesn't fit in the CJ process's memory.

This is the reason why workflows are infeasible for large models beyond what the CJ machine can hold.

---

### Pass-Through Architecture

With `FOBSContextKey.PASS_THROUGH` enabled on CJ's cell FOBS context, the data path becomes:

```
FL Server ──stream──▶ CJ (LazyDownloadRef only, no tensor data)
                          └──forward ref──▶ Subprocess
                                               └──download──▶ FL Server
```

CJ holds **only lightweight placeholders** (< 100 bytes per tensor). The subprocess downloads each tensor directly from the FL server — CJ is never involved in the tensor data path.

#### 1. `FOBSContextKey.PASS_THROUGH` (`nvflare/fuel/utils/fobs/__init__.py`)
A new context key that signals ViaDownloaderDecomposer to skip the download step and create lazy placeholders instead.

#### 2. `LazyDownloadRef` (`via_downloader.py`)
A small sentinel object (four fields: `fqcn`, `ref_id`, `item_id`, `dot`) created by `recompose()` in PASS_THROUGH mode.  It carries the original FL server's FQCN, batch ref_id, intra-batch item ID, and Datum Object Type — everything the subprocess needs to download the tensor directly.

#### 3. `_LazyBatchInfo` (`via_downloader.py`)
A named sentinel stored in `fobs_ctx[items_key]` during PASS_THROUGH receive. Using a typed class (rather than a plain tuple) makes the PASS_THROUGH branch unambiguous and immune to accidental type collisions with real item dicts.

#### 4. `LazyDownloadRefDecomposer` (`via_downloader.py`)
A new auto-registered FOBS decomposer for `LazyDownloadRef`.  When CJ re-serializes a task containing `LazyDownloadRef` objects:

- **`decompose()`** delegates to `get_dot_handler(lazy.dot)` — the original `ViaDownloaderDecomposer` subclass (e.g., `TensorDecomposer`, `NumpyArrayDecomposer`).  That handler's `_finalize_lazy_batch` post-callback re-emits the *original* server datum (fqcn + ref_id + DOT) so the subprocess knows exactly where to download from.  `lazy_dot` is appended to the encoding dict for routing on the receive side.
- **`recompose()`** uses `lazy_dot` to look up the handler and delegates to `handler.recompose()`, which retrieves the real tensor from `fobs_ctx[handler.items_key]` (populated by `process_datum()` when the subprocess received the forwarded datum).

The `dot` (Datum Object Type) field on both `LazyDownloadRef` and `_LazyBatchInfo` ensures that numpy arrays stay with `NumpyArrayDecomposer` and PyTorch tensors stay with `TensorDecomposer`, preserving type safety through the full pass-through hop.

#### 5. `ClientAPILauncherExecutor.initialize()` (`client_api_launcher_executor.py`)
On startup, the executor enables PASS_THROUGH on the engine cell's FOBS context:

```python
cell.core_cell.update_fobs_context({FOBSContextKey.PASS_THROUGH: True})
```

This single line activates the full B1 architecture for every job that uses `launch_external_process=True` — including `llm_hf` and any recipe that calls `ScriptRunner(launch_external_process=True)`.

---

### Why not intercept at the pipe level?

The pipe (CellPipe) operates on already-serialized bytes.  Intercepting at the pipe level would require parsing FOBS binary format, re-writing datum references, and re-assembling the byte stream — fragile and tightly coupled to the wire format.

Intercepting at the FOBS decomposer level is the natural extension point: decomposers already control exactly when and how data is materialised.  PASS_THROUGH simply adds a "don't materialise" branch to that existing mechanism.

### Why `dot` (Datum Object Type) on `LazyDownloadRef`?

The subprocess must know *which* `ViaDownloaderDecomposer` subclass owns the downloaded data so it can store it in the correct `fobs_ctx[items_key]` and route `recompose()` correctly.  The `dot` field, set when the server originally serialized the tensor, carries this type information through the pass-through hop without any type-switching logic.

### Memory profile at CJ

| Stage | Before (tensor materialised) | After (B1 pass-through) |
|-------|------------------------------|------------------------|
| CJ receive | Full model size (e.g., 140 GB) | ~100 bytes per tensor |
| CJ forward | Creates new download tx | Re-emits original server datum |
| Subprocess receive | Downloads from CJ | Downloads directly from FL server |

---

## Benefits

1. **Zero tensor copy at CJ** — CJ memory footprint is independent of model size.
2. **One network transfer** instead of two — tensors travel server → subprocess directly.
3. **No CJ OOM risk** for large models regardless of CJ machine memory capacity.
4. **Transparent to job authors** — no changes to job configs, training scripts, or recipe APIs; `launch_external_process=True` automatically activates B1.
5. **Type-safe** — `dot` propagation preserves tensor type (numpy / pytorch) through the hop without any if/elif type switching.

---

## Backward Compatibility

- All existing jobs using `launch_external_process=True` automatically benefit. No config or script changes required.
- Jobs using `launch_external_process=False` (in-process executor) are completely unaffected — `ClientAPILauncherExecutor.initialize()` is not called.
- For models smaller than the ViaDownloaderDecomposer streaming threshold (2 MB per array), FOBS uses native (inline) serialization regardless of `PASS_THROUGH` — behaviour is identical to before.
- `LazyDownloadRefDecomposer` is auto-registered via the existing `register_folder` mechanism; no explicit registration call is needed by any caller.

---

## Changed Files

| File | Change |
|------|--------|
| `nvflare/fuel/utils/fobs/__init__.py` | Add `FOBSContextKey.PASS_THROUGH` |
| `nvflare/fuel/utils/fobs/decomposers/via_downloader.py` | Add `LazyDownloadRef`, `_LazyBatchInfo`, PASS_THROUGH branches in `process_datum()` / `recompose()`, `LazyDownloadRefDecomposer`, `_finalize_lazy_batch` post-callback |
| `nvflare/app_common/executors/client_api_launcher_executor.py` | `initialize()` enables PASS_THROUGH on engine cell |

---

### Unit tests — `tests/unit_test/fuel/utils/fobs/test_pass_through.py` (22 tests)

| Test class | What is verified |
|------------|-----------------|
| `TestLazyDownloadRef` | Construction, `__slots__`, per-item distinctness |
| `TestLazyBatchInfo` | Construction, `__slots__`, `isinstance` reliability vs plain tuple |
| `TestProcessDatumPassThrough` | PASS_THROUGH stores `_LazyBatchInfo`, never calls `_download_from_remote_cell`; normal mode calls download |
| `TestRecomposePassThrough` | Returns `LazyDownloadRef` with correct `fqcn`, `ref_id`, `item_id` from `_LazyBatchInfo` |
| `TestDecomposeWithLazyDownloadRef` | Returns REF encoding; `_finalize_lazy_batch` post-CB registered once per batch regardless of item count; emitted datum has correct fqcn/ref_id/DOT |
| `TestNoMemoryAccumulation` | `_CtxKey.OBJECTS` absent after PASS_THROUGH (no download transaction opened); `DownloadService._tx_table` unchanged; 50-cycle repeat produces no state bleed |

### End-to-end tests — `tests/unit_test/fuel/f3/streaming/test_pass_through_e2e.py` (5 tests, real TCP Cells)

| Test | What is verified |
|------|-----------------|
| `test_arrays_survive_pass_through_hop` | Full round-trip: server → CJ (PASS_THROUGH) → subprocess, arrays arrive bit-exact |
| `test_cj_holds_only_lazy_refs_not_tensor_data` | After PASS_THROUGH deserialization, CJ holds only `LazyDownloadRef`, never `np.ndarray` |
| `test_cj_creates_no_download_transaction` | `DownloadService._tx_table` is unchanged during PASS_THROUGH + re-serialization |
| `test_forwarded_payload_carries_original_server_ref` | Forwarded datum contains original server `fqcn` and `ref_id` — subprocess downloads from server, not CJ |
| `test_multiple_array_roundtrip` | 8-array batch all survive with bit-exact values |

### Integration test — `tests/integration_test/data/jobs/pt_large_model_pass_through/`

Full-stack integration test using `PTClientAPILauncherExecutor` with `launch_once=True` (the pattern used by `llm_hf`):

- **Model**: `LargeNet` — 3-layer MLP with ~8 MB of float32 parameters, well above the 2 MB ViaDownloaderDecomposer streaming threshold.  This forces the real B1 code path (streaming + PASS_THROUGH) rather than the native inline path used by small models.
- **Client script**: Mirrors `llm_hf/client.py` structure (`while flare.is_running()` loop, receive / train / send).  Uses CPU-only synthetic data — no dataset download required in CI.
- **Added to** `client_api.yml` as `"run pt-large-model-pass-through"`.

### DIFF
--- a/tests/unit_test/fuel/f3/streaming/test_pass_through_e2e.py
+++ b/tests/unit_test/fuel/f3/streaming/test_pass_through_e2e.py
@@ -56,19 +56,19 @@ Tests:
 
 import json
 import time
+import uuid
 
 import numpy as np
 import pytest
 
 from nvflare.fuel.f3.cellnet.cell import Cell
+from nvflare.fuel.f3.cellnet.core_cell import CoreCell
 from nvflare.fuel.utils.fobs import FOBSContextKey
 from nvflare.fuel.utils.fobs.decomposers.via_downloader import LazyDownloadRef, _RefKey
 from nvflare.fuel.utils.fobs.lobs import dump_to_bytes, load_from_bytes
 from nvflare.fuel.utils.network_utils import get_open_ports
 
-_SERVER_FQCN = "server"
-_SUBPROCESS_FQCN = "subprocess"
-_CONNECT_WAIT = 2.0  # seconds to wait for TCP cell connection
+CONNECT_WAIT = 2.0  # seconds to wait for TCP cell connection
 
 
 def _register_numpy():
@@ -112,16 +112,22 @@ class TestPassThroughE2E:
     @pytest.fixture(scope="class")
     def cells(self):
         port = get_open_ports(1)[0]
-        server = Cell(_SERVER_FQCN, f"tcp://localhost:{port}", secure=False, credentials={})
+        server_fqcn = f"server-{uuid.uuid4().hex[:8]}"
+        subproc_fqcn = f"subprocess-{uuid.uuid4().hex[:8]}"
+        server = Cell(server_fqcn, f"tcp://localhost:{port}", secure=False, credentials={})
         server.core_cell.start()
-        subproc = Cell(_SUBPROCESS_FQCN, f"tcp://localhost:{port}", secure=False, credentials={})
+        subproc = Cell(subproc_fqcn, f"tcp://localhost:{port}", secure=False, credentials={})
         subproc.core_cell.start()
-        time.sleep(_CONNECT_WAIT)
+        time.sleep(CONNECT_WAIT)
         try:
             yield server, subproc
         finally:
+            subproc_name = subproc.get_fqcn()
+            server_name = server.get_fqcn()
             subproc.core_cell.stop()
             server.core_cell.stop()
+            CoreCell.ALL_CELLS.pop(subproc_name, None)
+            CoreCell.ALL_CELLS.pop(server_name, None)
 
     # ------------------------------------------------------------------
     # Test 1 — full round-trip correctness
@@ -255,8 +261,9 @@ class TestPassThroughE2E:
                 text = body.decode("utf-8")
                 ref = json.loads(text)
                 if _RefKey.FQCN in ref and _RefKey.REF_ID in ref:
-                    assert ref[_RefKey.FQCN] == _SERVER_FQCN, (
-                        f"Forwarded datum fqcn must be '{_SERVER_FQCN}' (FL server), "
+                    expected_server = server.get_fqcn()
+                    assert ref[_RefKey.FQCN] == expected_server, (
+                        f"Forwarded datum fqcn must be '{expected_server}' (FL server), "
                         f"but got '{ref[_RefKey.FQCN]}'. Subprocess would download from wrong source."
                     )
                     assert ref[_RefKey.REF_ID], "Forwarded datum must have a non-empty ref_id."


nvflare/app_common/executors/client_api_launcher_executor.py
```diff
@@ -110,6 +111,23 @@ def __init__(
...
+        # the CJ.  ViaDownloaderDecomposer will instead create LazyDownloadRef
+        # placeholders that carry the original server FQCN and ref_id.  When CJ
+        # forwards the task to the subprocess agent via the task pipe, those
+        # placeholders are re-emitted as-is, causing the subprocess to download
+        # each tensor directly from the server — one tensor at a time, with no
+        # size limit and no tensor copy at CJ.
+        engine = fl_ctx.get_engine()
+        cell = engine.get_cell()
+        if cell is not None:
+            cell.core_cell.update_fobs_context({FOBSContextKey.PASS_THROUGH: True})
```
pcnudde, line 134: Should this be cleared at the end of the run? I assume this acts as a global variable for the process, and it is never cleared. Not sure if that is ok.

pcnudde (commented):
Are there any possible incompatibilities with things like filters or other users code that expects real tensors instead of the the lazy references.

nvflare/app_common/executors/client_api_launcher_executor.py
chesterxgchen: fixed

chesterxgchen:
> Are there any possible incompatibilities with things like filters or other users code that expects real tensors instead of the the lazy references.
There is no filter between Client Job Process and Client Agent (trainer). No other components.

pcnudde (approved):

PR #4215, opened 2026-02-20
[Draft] Tensor Disk Offload for pytorch in FedAvg and Swarm

pcnudde:
## What this PR does
- adds disk-backed tensor streaming for large PyTorch model payload transfer paths
- introduces lazy payload primitives and utilities to carry deferred tensor references until materialization is needed
- renames and standardizes the runtime switch as stream_to_disk
- propagates stream_to_disk at server job runtime through the Cell fobs context instead of per-message payload metadata
- updates FedAvg and Swarm aggregation flows with capability-gated behavior:
  - lazy-aware aggregators can consume lazy tensors directly
  - non-lazy-aware custom aggregators keep backward-compatible behavior via resolve and cleanup before accept
- keeps existing behavior for users who do not enable stream_to_disk

## Design intent
- optimize memory usage for very large models without forcing all custom aggregators to implement lazy handling
- preserve compatibility and deterministic cleanup while allowing selective opt-in for lazy-aware aggregators such as FedAvg weighted path and Swarm paths that support it

## Main code areas touched
- tensor decomposer and downloader pipeline for PyTorch payloads
- FedAvg workflow and recipe wiring
- Swarm controller gather and aggregation handling
- lazy payload helper module for nested payload detection and materialization
- server runtime hook for per-job stream_to_disk fobs context setup
- disk streaming design documentation

## Test coverage in this PR
- unit tests for lazy payload utilities
- unit tests for FedAvg lazy/non-lazy aggregator compatibility behavior
- unit tests for Swarm lazy payload handling
- unit tests for server runtime stream_to_disk propagation
- stress test script updates for large-model Swarm scenarios used for manual validation

## Notes
- this PR is draft to allow focused review of architecture, compatibility boundaries, and memory-behavior validation before finalization

### DIFF
--- a/nvflare/app_opt/pt/tensor_downloader.py
+++ b/nvflare/app_opt/pt/tensor_downloader.py
@@ -144,7 +144,7 @@ def _extract_safetensors_keys(data: bytes) -> list[str]:
         raise ValueError("Invalid safetensors data: too short")
 
     header_size = struct.unpack("<Q", data[:8])[0]
-    if header_size <= 0:
+    if header_size == 0:
         raise ValueError("Invalid safetensors data: empty header")
 
     header_end = 8 + header_size
@@ -208,16 +208,20 @@ def download_tensors_to_disk(
     temp_dir = tempfile.mkdtemp(prefix="nvflare_tensors_")
 
     consumer = DiskTensorConsumer(temp_dir)
-    download_object(
-        from_fqcn=from_fqcn,
-        ref_id=ref_id,
-        consumer=consumer,
-        per_request_timeout=per_request_timeout,
-        cell=cell,
-        secure=secure,
-        optional=optional,
-        abort_signal=abort_signal,
-    )
+    try:
+        download_object(
+            from_fqcn=from_fqcn,
+            ref_id=ref_id,
+            consumer=consumer,
+            per_request_timeout=per_request_timeout,
+            cell=cell,
+            secure=secure,
+            optional=optional,
+            abort_signal=abort_signal,
+        )
+    except Exception:
+        shutil.rmtree(temp_dir, ignore_errors=True)
+        raise
 
     if consumer.error:
         return consumer.error, None
--- a/tests/unit_test/app_opt/pt/test_disk_tensor_consumer.py
+++ b/tests/unit_test/app_opt/pt/test_disk_tensor_consumer.py
@@ -20,6 +20,7 @@ import pytest
 import torch
 from safetensors.torch import save as save_tensors
 
+import nvflare.app_opt.pt.tensor_downloader as tensor_downloader
 from nvflare.app_opt.pt.lazy_tensor_dict import LazyTensorDict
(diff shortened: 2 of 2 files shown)


nvflare/app_common/aggregators/intime_accumulate_model_aggregator.py
```diff
@@ -44,6 +44,9 @@ def _get_missing_keys(ref_dict: dict, dict_to_check: dict):
 
 
 class InTimeAccumulateWeightedAggregator(Aggregator):
+    # Optional capability marker: this aggregator can consume lazy tensor payloads.
+    accepts_lazy_tensors = True
```
chesterxgchen: hard-code this is not optional, where is this configured ?

nvflare/app_common/aggregators/weighted_aggregation_helper.py
```diff
@@ -56,6 +56,9 @@ def add(self, data, weight, contributor_name, contribution_round):
                 if self.exclude_vars is not None and self.exclude_vars.search(k):
                     continue
 
+                if hasattr(v, "resolve"):
```
chesterxgchen: how do we know "resolve" is a callable ?

nvflare/app_common/ccwf/ccwf_job.py
```diff
@@ -83,6 +83,7 @@ def __init__(
         request_to_submit_result_max_wait=None,
         request_to_submit_result_interval: float = 1.0,
         max_concurrent_submissions: int = 1,
+        stream_to_disk: bool = False,
```
chesterxgchen: stream_to_disk apply to anything or just model ?

nvflare/app_common/aggregators/intime_accumulate_model_aggregator.py
pcnudde: Aggregators do not have to implement this capability (for some it does not matter, and we also want backwards compatibility for custom aggregators). InTimeAccumulateWeightedAggregator does implement it, so we set the capability to True. But the comment could be better. will fix that.

nvflare/app_common/ccwf/ccwf_job.py
pcnudde: this feature is just for the models being aggregated. Maybe we need a better name...

nvflare/app_common/aggregators/weighted_aggregation_helper.py
pcnudde: we don't. will fix

pcnudde:
@chesterxgchen will clarify docs. Resolve is loading the lazy reference from disk. For the 2 aggregators that we support (fedavg and swarm) we load the tensors exactly when we need them. But for the ones that don't we have to load them again. But I now think I can make this simpler.

pcnudde:
closing, will reopen new one with descriptio matching current naming and simplification.

PR #4221, opened 2026-02-23
Tensor Disk Offload for PyTorch in FedAvg and Swarm

pcnudde:
Fresh PR replacing #4215.

Summary:
- enable tensor disk offload runtime path for streamed tensor payloads
- FedAvg and Swarm integration updates
- lazy tensor handling and cleanup updates
- key design and usage documentation updates

### DIFF
--- a/examples/tutorials/job_cli.ipynb
+++ b/examples/tutorials/job_cli.ipynb
@@ -1105,7 +1105,7 @@
    "id": "b695d5e3-3dde-456a-a2db-54ad8c2ccb7d",
    "metadata": {},
    "source": [
-    "Once you have the the job folder. You should be able to run the job as before\n"
+    "Once you have the job folder. You should be able to run the job as before\n"
    ]
   },
   {
--- a/examples/tutorials/logging.ipynb
+++ b/examples/tutorials/logging.ipynb
@@ -209,7 +209,7 @@
         "In custom_log_config.json, let's see how we modify the consoleHandler and consoleFormatter to achieve our goal of a cleaner console output.\n",
         "Below are some example changes, however feel free to experiment with different configurations:\n",
         "\n",
-        "**Filters:** In the consoleHandler, we add the FLFilter which only allows logs related to FL training to pass through using the the LoggerNameFilter:\n",
+        "**Filters:** In the consoleHandler, we add the FLFilter which only allows logs related to FL training to pass through using the LoggerNameFilter:\n",
         "\n",
         "- ``\"filters\": [\"FLFilter\"],``\n",
         "\n",
--- a/examples/tutorials/self-paced-training/part-1_federated_learning_introduction/chapter-2_develop_federated_learning_applications/02.4_convert_machine_learning_to_federated_learning/02.4.1_convert_logistic_regression_to_federated_learning/convert_logistic_regression_to_fl.ipynb
+++ b/examples/tutorials/self-paced-training/part-1_federated_learning_introduction/chapter-2_develop_federated_learning_applications/02.4_convert_machine_learning_to_federated_learning/02.4.1_convert_logistic_regression_to_federated_learning/convert_logistic_regression_to_fl.ipynb
@@ -133,7 +133,7 @@
       "metadata": {},
       "outputs": [],
       "source": [
-        "# Note: the the download site remember your download history and abort the 2nd download attempt. \n",
+        "# Note: the download site remembers your download history and aborts the 2nd download attempt. \n",
         "! echo y | bash ./code/data/prepare_heart_disease_data.sh"
       ]
     },
--- a/examples/tutorials/self-paced-training/part-2_federated_learning_system/chapter-4_setup_federated_system/04.4_provision_and_run_with_docker/provision_and_run_with_dockers.ipynb
+++ b/examples/tutorials/self-paced-training/part-2_federated_learning_system/chapter-4_setup_federated_system/04.4_provision_and_run_with_docker/provision_and_run_with_dockers.ipynb
@@ -113,7 +113,7 @@
     "\n",
     "To provision a project with the Docker image, we need to perform the same steps as described previously in [Provision Using `nvflare provision` CLI](../04.1_provision_via_cli/provision_via_cli.ipynb). The only modification needed is to make sure that the server and clients can use the docker image `nvflare-pt-docker` when they start. This can be done by modifying the project configuration file. \n",
     "\n",
-    "Take a look at the example configuration file in [`code/project.yml`](code/project.yml). You will notice that the the only difference between this configuration file and [the one in \"Provision Using `nvflare provision` CLI\"](../04.1_provision_via_cli/code/project.yml), is a newly added argument under the `nvflare.lighter.impl.static_file.StaticFileBuilder` builder:\n",
+    "Take a look at the example configuration file in [`code/project.yml`](code/project.yml). You will notice that the only difference between this configuration file and [the one in \"Provision Using `nvflare provision` CLI\"](../04.1_provision_via_cli/code/project.yml), is a newly added argument under the `nvflare.lighter.impl.static_file.StaticFileBuilder` builder:\n",
     "```\n",
     "builders:\n",
     "  - path: nvflare.lighter.impl.static_file.StaticFileBuilder\n",
--- a/examples/tutorials/self-paced-training/part-3_security_and_privacy/chapter-6_Security_in_federated_compute_system/06.2_site_security_privacy_policy/site_policy.ipynb
+++ b/examples/tutorials/self-paced-training/part-3_security_and_privacy/chapter-6_Security_in_federated_compute_system/06.2_site_security_privacy_policy/site_policy.ipynb
@@ -327,7 +327,7 @@
    "id": "7c84b47b",
    "metadata": {},
    "source": [
-    "In the privacy policy, we noticed that there are different scopes can be defined. The scope-specified policy is enforced the the FLARE's filter mechanism. For each non-public scope, one can define the different filters to enforce the desired behavior.  For example, the organization administrator can instrument filter to prevent certain type of data leakage accidentally exposed by data scientists\n",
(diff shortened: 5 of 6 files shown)


nvflare/app_common/ccwf/swarm_client_ctl.py
```diff
@@ -401,13 +405,24 @@ def process_config(self, fl_ctx: FLContext):
             message_handle_func=self._process_submission_request,
         )
 
+    def _set_enable_tensor_disk_offload(self):
+        cell = self.engine.get_cell()
+        if not cell:
+            return
+
+        desired = bool(self.enable_tensor_disk_offload)
+        current = bool(cell.get_fobs_context().get("enable_tensor_disk_offload", False))
```
chesterxgchen: strange logics here. If the flag is already set, you want to overwrite it --> does this cause side effects ? someone set the value, should we set the value, after finish remove the value, so you dont have to do this comaprision

nvflare/app_common/workflows/fedavg.py
```diff
@@ -122,9 +128,24 @@ def __init__(
...
+            if self.enable_tensor_disk_offload:
+                self.warning(
+                    "enable_tensor_disk_offload is enabled but no active cell is available; using in-memory download path"
+                )
+            return
+
+        desired = bool(self.enable_tensor_disk_offload)
+        current = bool(cell.get_fobs_context().get("enable_tensor_disk_offload", False))
+        if current != desired:
+            cell.update_fobs_context({"enable_tensor_disk_offload": desired})
```
chesterxgchen: seems to be the duplicated logics as the swarm learning

nvflare/app_opt/pt/lazy_tensor_dict.py
```diff
@@ -0,0 +1,125 @@
...
+    """
+
+    def __init__(self, temp_dir: str):
+        self.path = temp_dir
+        self._deleted = False
+
+    def cleanup(self):
+        if not self._deleted:
+            self._deleted = True
+            shutil.rmtree(self.path, ignore_errors=True)
```
chesterxgchen: this is  a a bit dangeous

YuanTingHsieh:
1.  /tmp might be tmpfs — feature silently broken in containers
```
temp_dir = tempfile.mkdtemp(prefix="nvflare_tensors_")
tempfile.mkdtemp()
```

 uses /tmp by default. On Kubernetes/Docker, /tmp is commonly a tmpfs mount (in-memory). You'd be "offloading to disk" but actually writing to RAM. The whole point of the feature is defeated with no error or warning.
There's no configurable temp_dir parameter to point at an actual disk path.

2.  Disk full before the download is handled, but accumulation is not
If disk fills during a write, the exception propagates through download_service.consume() → download_failed() → shutil.rmtree() — so that specific download cleans up. Fine.
But across multiple rounds with N clients, each round creates N temp dirs. Cleanup is GC-driven (_TempDirRef.__del__). If aggregation holds _LazyRef objects longer than expected, or there are any reference cycles, previous rounds' temp dirs accumulate. Eventually you fill the disk that way — and then the exception-on-write path kicks in, but now you're already in a bad state.

3.  No explicit cleanup after aggregation consumes the refs
After weighted_aggregation_helper materializes all the tensors and produces the aggregate, the _LazyRef objects become unreferenced — but cleanup is entirely GC-dependent. In CPython with reference counting this usually works, but there's no explicit cleanup() call guaranteeing timely removal.

4.  ignore_errors=True everywhere in cleanup

shutil.rmtree(self.path, ignore_errors=True)
Silent failures. If cleanup fails (permissions, NFS, etc.), you get no log, no warning. Temp files accumulate invisibly.

### Summary
ConcernSeverity/tmp = tmpfs in containers, feature silently does nothingHighNo configurable temp dir for actual disk pathHighMulti-round temp dir accumulation via GC delayMediumNo disk space pre-checkMediumSilent cleanup failuresLow
The tmpfs one is the most subtle — a user enables enable_tensor_disk_offload=True, memory usage doesn't improve, and there's no indication why.

pcnudde:
1. Good catch. We can configure that at the moment via a env variable TMPDIR. I wonder if that is enough (need to document that better obviously). Or maybe we need separate env variable / setting for this purpose alone. But I'd rather minimize new configs as we have so many of them. 
2. tmp disk full is a error condition that will obviously be catastrophic, But that is true with main disk being full as well. (or memory being full). I will update to make sure we log this well. 
3. by design. I had first an explicit cleanup, but it adds a lot of complexity without real benefit. the GC will clean this up. (and as a backstop in extreme cases there is the standard of tmp cleanup).
4. true. will add logging for it.

nvflare/app_opt/pt/lazy_tensor_dict.py
pcnudde: fixed

nvflare/app_common/workflows/fedavg.py
pcnudde: fixed

nvflare/app_common/ccwf/swarm_client_ctl.py
pcnudde: fixed

YuanTingHsieh:
^ the latest 3 comments from greptile makes sense, not functional gap just comments or logging needs to be added.

pcnudde:
@YuanTingHsieh yes these comments make sense. Will address them.

pcnudde:
Addressed the latest Greptile comments in commit `eec301465`.

Summary of changes:

1. `docs/design/tensor_disk_offload.md`
- Corrected statement about `ServerEngine.get_cell()`.
- Corrected cell-unavailable behavior text (fallback behavior now documented accurately).
- Corrected custom-aggregator cleanup guidance (no `cleanup()` on `_LazyRef`).
- Removed unchanged `server_engine.py` from design-relevant files.

2. `LazyTensorDict`
- Added `__iter__` so direct iteration (`for k in ltd`, `list(ltd)`) works.
- Added unit test for iteration behavior.

3. `tensor_downloader`
- Added second-chance `_cleanup_temp_dir(temp_dir)` when `consumer.error` is set before returning `(error, None)`.
- Added unit test for this error path.

4. Offload apply signal/logging
- `apply_enable_tensor_disk_offload` now returns `(previous_value, applied)`.
- FedAvg/Swarm now warn when `enable_tensor_disk_offload=True` but apply was not possible (`applied=False`).
- Updated corresponding tests.

5. PR scope cleanup
- Removed unrelated `.gitignore` entries (`.tmp`, `.codex/`).

Validation:
- `112 passed` across touched unit-test suites.

pcnudde:
Added both suggested clarifying comments in commit 603d915:

- `tensor_downloader.py`: documented intentional/safe double-cleanup between `download_failed` and `consumer.error` path.
- `tensor_disk_offload_context.py`: documented `previous_value is None` invariant in restore guard.

No behavior changes; comment-only update.

Validation after this commit:
- 83 passed (targeted unit suites around downloader/context/fedavg/swarm paths).

pcnudde:
Addressed the latest Greptile feedback from https://github.com/NVIDIA/NVFlare/pull/4221#issuecomment-3988712480 in commit `83c8def6`.

Updates:

1. Duplicate key overwrite in `DiskTensorConsumer.consume_items`
- Added defensive duplicate-key check and now raise `ValueError` if the same tensor key appears across chunks.
- File: `nvflare/app_opt/pt/tensor_downloader.py`

2. Redundant context writes when `enabled=False`
- `apply_enable_tensor_disk_offload` now updates the FOBS context only when `previous != enabled`.
- File: `nvflare/app_common/utils/tensor_disk_offload_context.py`

3. Test updates
- Added duplicate-key unit test.
- Added no-op context-update unit test when target value already matches.
- Files: `tests/unit_test/app_opt/pt/test_disk_tensor_consumer.py`, `tests/unit_test/app_common/utils/tensor_disk_offload_context_test.py`

Validation:
- Targeted unit suites passed: `85 passed`.

Note on duplicated mock classes across tests:
- acknowledged as low-priority test refactor; can be done separately to keep this PR focused.

holgerroth:
@greptileai

nvflare/app_common/workflows/fedavg.py
```diff
@@ -128,97 +138,111 @@ def __init__(
...
-            self.event(AppEventType.ROUND_STARTED)
+        previous_disk_offload, disk_offload_applied = apply_enable_tensor_disk_offload(
+            engine=getattr(self, "engine", None),
+            enabled=self.enable_tensor_disk_offload,
+        )
+        if self.enable_tensor_disk_offload and not disk_offload_applied:
+            self.warning(
+                "enable_tensor_disk_offload=True but no active cell is available; "
+                "falling back to in-memory tensor download"
+            )
```
YuanTingHsieh, line 149: regarding this, the logic is good.
I am just wondering should we have something easier that users can still write their old "run" as before and just a function to turn on "enable" and then we will have these code blocks added (so they don't need to add themselves in their controllers) But I don't know if that's worth doing if we are moving forward to collab api

nvflare/app_opt/pt/tensor_downloader.py
```diff
@@ -128,3 +135,104 @@ def download_tensors(
...
+    except Exception:
+        _cleanup_temp_dir(temp_dir)
+        raise
+
+    if consumer.error:
+        _cleanup_temp_dir(temp_dir)
+        return consumer.error, None
+
+    key_to_file = consumer.result if consumer.result is not None else {}
+    return None, LazyTensorDict(key_to_file=key_to_file, temp_dir=temp_dir)
```
YuanTingHsieh, line 238: what happen if this "key_to_file" is empty dict "{}"?

nvflare/app_common/workflows/fedavg.py
YuanTingHsieh: maybe not worth doing that enhancement before collab

pcnudde: yes, maybe there is some more generic way of setting these variables during a run. But I on purpose only wanted to touch the 2 main paths (fedavg and swarm) to minimize impact. And yes we need to see how this all works with collab and also if we do out of band model communication via external model hub.

nvflare/app_opt/pt/tensor_downloader.py
pcnudde: if key_to_file == {}, the code returns an empty LazyTensorDict. That is falsy because of lazy_tensor_dict.py (line 122), so built-in FedAvg treats it as an empty client result and skips it at fedavg.py (line 249). So it should not break aggregation.

PR #4225, opened 2026-02-24
[2.7] Fix subprocess converter wiring, swarm learning bugs, and recipe enhancements

chesterxgchen:
## Summary

This PR addresses several related bugs and improvements in the Client API subprocess path, swarm learning, and the recipe interface.
Address FLARE-2743 (bug/5921094) and (bug/5921708)

---

### 1. LazyDownloadRef dtype error with pass-through streaming
When pass-through tensor streaming is enabled on the CJ, param converters (e.g. `NumpyToPTParamsConverter`) were running on the CJ *before* lazy references were materialized. This caused `"Could not infer dtype of LazyDownloadRef"` warnings and potential data corruption on large-model jobs using pass-through transfer.

### 2. Subprocess Client API had no converter wiring
The `ExProcessClientAPI` (subprocess path) created a `FlareAgentWithFLModel` with no converters attached. Format conversion was happening on the CJ side in `LauncherExecutor.execute()`, which triggered issue #1.

### 3. Missing `learn_task_check_interval` in `CCWFJob.add_swarm()`
`SwarmClientConfig.learn_task_check_interval` was not forwarded to the swarm controller, so any customization of that parameter was silently ignored.

### 4. `extract_participants()` did not support `{"targets": [...]}` dict form
The deploy map validation only handled list-form and `{"sites": [...]}` dict-form entries. The `{"targets": [...]}` form (used by some job generators) raised an unhandled error.

### 5. `SimpleSwarmLearningRecipe` missing parameters
`min_clients`, `launch_external_process`, and `command` could not be set through the recipe interface, forcing users to construct jobs manually.

### 6. 3rd-party integration docs referenced deprecated `FlareAgentWithCellPipe`
The class has been superseded by constructing a `CellPipe` and `FlareAgent` directly, or using the higher-level Client API (`nvflare.client`).

---

### 7. `task_name` read from absent shareable header in `FlareAgentWithFLModel.shareable_to_task_data()`
The task name was read via `shareable.get_header(FLContextKey.TASK_NAME, "")`, but this header is never populated in the subprocess path — the task name is carried as the pipe `Message.topic` and stored in `self.current_task.task_name` by `FlareAgent._on_task()`. An absent header caused `task_name` to fall back to `""`, making `ParamsConverter.process()` silently skip conversion. Fixed to use `self.current_task.task_name if self.current_task else ""`, consistent with `task_result_to_shareable()`.

### 8. Double-conversion when `from_nvflare_converter_id` / `to_nvflare_converter_id` set on launcher executor
`LauncherExecutor.execute()` still applied CJ-side converters if users explicitly set `from_nvflare_converter_id` / `to_nvflare_converter_id`, while the subprocess was also applying its own default converters — resulting in double conversion. Fixed by removing the converter calls from `LauncherExecutor.execute()` entirely. Since `LauncherExecutor` is only ever subclassed by `ClientAPILauncherExecutor` (subprocess path), this has no effect on the in-process path.

---

## Approach

**Converter logic removed from CJ launcher executors; moved to the subprocess agent side.**

Previously, `PTClientAPILauncherExecutor` and `TFClientAPILauncherExecutor` set up format converters in their `initialize()` methods, and `LauncherExecutor.execute()` applied them on the CJ before dispatching to the subprocess. This was the root cause of issue #1 — the CJ was touching tensor data (triggering dtype inference) before `LazyDownloadRef` placeholders were materialized in the subprocess.

The converter setup blocks have been **completely removed** from both launcher executors. Converters for the subprocess path now live entirely inside the subprocess, wired into `FlareAgentWithFLModel` and executing after full payload materialization.

Key changes:
- `FlareAgentWithFLModel` gains optional `from_nvflare_converter` / `to_nvflare_converter` params. A lightweight `_ConverterContext` stub (duck-typing `FLContext.get_prop/set_prop`) allows `ParamsConverter.process()` to work without a real `FLContext` in the subprocess.
- `SERVER_EXPECTED_FORMAT` is added to `ConfigKey` and propagated in the subprocess launch config so the agent knows what format the server expects.
- A shared factory `converter_utils.create_default_params_converters()` is introduced, replacing four separate copies of the same format-conditional logic across PT and TF executors.
- In-process executors (`PTInProcessClientAPIExecutor`, `TFInProcessClientAPIExecutor`) are **unchanged in behavior** — they still apply converters on the CJ side, which is correct since there is no separate subprocess.

---

## Reason for this approach over alternatives

An earlier iteration kept the converter setup in `initialize()` but added a `ClientAPILauncherExecutor.execute()` override that nulled the converters out before calling `super().execute()` and restored them after. This was removed because it left converters alive on the object but always bypassed — confusing dead code. The current approach is explicit and honest: launcher executors do not set converters at all, and the subprocess handles its own conversion autonomously.

---

## Affected files

| Area | Files |
|------|-------|
| Core bug fixes | `ccwf_job.py`, `fed_utils.py` |
| Converter framework | `converter_utils.py` (new), `config.py`, `flare_agent_with_fl_model.py`, `ex_process/api.py`, `client_api_launcher_executor.py` (base) |
| PT launcher executor — converter removed | `pt/client_api_launcher_executor.py` |
| PT in-process executor — refactored to shared factory | `pt/in_process_client_api_executor.py` |
| TF launcher executor — converter removed + `initialize()` deleted | `tf/client_api_launcher_executor.py` |
| TF in-process executor — refactored to shared factory | `tf/in_process_client_api_executor.py` |
| Recipe | `pt/recipes/swarm.py` |
| Docs | `3rd_party_integration.rst`, `3rd_party_trainer.py` |
| Tests | 4 new test files, 4 modified test files, 2 integration YAMLs |

---

## Trade-offs and future work

- **TF/Keras converter path is not directly unit-tested** due to TF/PyTorch dependency conflicts in a shared environment. The `create_default_params_converters` Keras branch is structurally identical to the PT branch (which is fully tested). A dedicated TF-only test environment could add coverage in a future pass.
- **`_ConverterContext` is a duck-type shim.** If more `ParamsConverter` use cases arise in subprocess contexts, a formal lightweight interface (separate from `FLContext`) would be worth extracting.
- **Converter plugin (`from_nvflare_converter_id`)** for the subprocess path is currently not wired — only the auto-created default converters are passed to the agent. Explicit user-provided converter IDs are still only supported for the in-process path. This can be addressed in a follow-up.

### DIFF
--- a/nvflare/app_common/executors/launcher_executor.py
+++ b/nvflare/app_common/executors/launcher_executor.py
@@ -76,10 +76,10 @@ class LauncherExecutor(TaskExchanger):
             train_task_name (str): Task name of train mode.
             evaluate_task_name (str): Task name of evaluate mode.
             submit_model_task_name (str): Task name of submit_model mode.
-            from_nvflare_converter_id (Optional[str]): Identifier used to get the ParamsConverter from NVFlare components.
-                This ParamsConverter will be called when model is sent from nvflare controller side to executor side.
-            to_nvflare_converter_id (Optional[str]): Identifier used to get the ParamsConverter from NVFlare components.
-                This ParamsConverter will be called when model is sent from nvflare executor side to controller side.
+            from_nvflare_converter_id (Optional[str]): Deprecated in LauncherExecutor path.
+                Parameter conversion for launcher-based external execution now happens in the subprocess agent.
+            to_nvflare_converter_id (Optional[str]): Deprecated in LauncherExecutor path.
+                Parameter conversion for launcher-based external execution now happens in the subprocess agent.
         """
         TaskExchanger.__init__(
             self,
@@ -208,16 +208,12 @@ class LauncherExecutor(TaskExchanger):
             raise RuntimeError("Launcher initialize failed.")
 
     def _init_converter(self, fl_ctx: FLContext):
-        engine = fl_ctx.get_engine()
-        from_nvflare_converter: ParamsConverter = engine.get_component(self._from_nvflare_converter_id)
-        if from_nvflare_converter is not None:
-            check_object_type(self._from_nvflare_converter_id, from_nvflare_converter, ParamsConverter)
-            self._from_nvflare_converter = from_nvflare_converter
-
-        to_nvflare_converter: ParamsConverter = engine.get_component(self._to_nvflare_converter_id)
-        if to_nvflare_converter is not None:
-            check_object_type(self._to_nvflare_converter_id, to_nvflare_converter, ParamsConverter)
-            self._to_nvflare_converter = to_nvflare_converter
+        if self._from_nvflare_converter_id or self._to_nvflare_converter_id:
+            self.log_warning(
+                fl_ctx,
+                "from_nvflare_converter_id/to_nvflare_converter_id are ignored in LauncherExecutor. "
+                "For launcher-based execution, configure converters in the subprocess client API path instead.",
+            )
 
     def _initialize_external_execution(
         self, task_name: str, shareable: Shareable, fl_ctx: FLContext, abort_signal: Signal
--- a/tests/unit_test/app_common/executors/client_api_launcher_executor_test.py
+++ b/tests/unit_test/app_common/executors/client_api_launcher_executor_test.py
@@ -121,3 +121,20 @@ def test_initialize_failure_restores_pass_through(monkeypatch):
 
     assert cell.core_cell.ctx[FOBSContextKey.PASS_THROUGH] is None
     assert executor._cell_with_pass_through is None
+
+
+def test_launcher_converter_ids_warn_when_ignored(monkeypatch):
+    warnings = []
+    monkeypatch.setattr(LauncherExecutor, "log_warning", lambda self, fl_ctx, msg: warnings.append(msg))
+
+    executor = LauncherExecutor(
+        pipe_id="test_pipe", from_nvflare_converter_id="from_converter", to_nvflare_converter_id="to_converter"
+    )
+    fl_ctx = _FakeFLContext(_FakeCell())
+
+    executor._init_converter(fl_ctx)
+
+    assert len(warnings) == 1
+    assert "ignored in LauncherExecutor" in warnings[0]
+    assert executor._from_nvflare_converter is None
+    assert executor._to_nvflare_converter is None


chesterxgchen:
> I think there is one risk but that probably can solved with warning or better documentation. This can double-apply conversion when users explicitly set from_nvflare_converter_id / to_nvflare_converter_id on launcher executors, because converter logic still runs in LauncherExecutor.execute() and defaults are also auto-created/applied in subprocess ExProcessClientAPI + FlareAgentWithFLModel.
> 
> If two of the converters can apply I assume it will break/crash.

hmm, I thought I had specific removed this converter calls in CJ side in my last revision. I see this still exists in the launcher executor. 

Let me do one more pass on this

nvflare/client/flare_agent_with_fl_model.py
```diff
@@ -12,20 +12,60 @@
...
+        **kwargs,
+    ):
+        super().__init__(*args, **kwargs)
+        self.from_nvflare_converter = from_nvflare_converter
+        self.to_nvflare_converter = to_nvflare_converter
+        self._converter_ctx = _ConverterContext()
+
     def shareable_to_task_data(self, shareable: Shareable) -> FLModel:
+        if self.from_nvflare_converter is not None:
+            task_name = shareable.get_header(FLContextKey.TASK_NAME, "")
```
GeorgeWang-nv, line 60: In `shareable_to_task_data()`, `task_name` is read from `shareable.get_header(FLContextKey.TASK_NAME, "")`.

However, when the shareable is transmitted through the pipe, the task name is carried as the pipe `Message.topic`, not necessarily as a shareable header. `FlareAgent._on_task(msg)` stores the task name in `self.current_task.task_name` (from `msg.topic`), but does not copy it into the shareable header.

If the header `__task_name__` is absent, `task_name` falls back to `""`, and `ParamsConverter.process()` silently skips conversion (`"" not in supported_tasks`), making the converter effectively dead code.

`task_result_to_shareable()` below already uses the reliable approach:
```python
task_name = self.current_task.task_name if self.current_task else ""
```

Suggest using the same pattern here for consistency and correctness:
```python
task_name = self.current_task.task_name if self.current_task else ""
```

(`self.current_task` is guaranteed to be set before `shareable_to_task_data()` is called, since `FlareAgent._on_task()` sets it before `get_task()` invokes this method.)

chesterxgchen: Fixed in 1350baed1 — changed to `self.current_task.task_name if self.current_task else ""`, consistent with `task_result_to_shareable()`. Also removed the now-unused `FLContextKey` import.

chesterxgchen:
@pcnudde Thanks for catching the double-conversion risk. Fixed in 1350baed1 — removed the converter calls directly from `LauncherExecutor.execute()`. Since `LauncherExecutor` is only ever subclassed by `ClientAPILauncherExecutor` (subprocess path), this cleanly eliminates CJ-side conversion with no impact on the in-process path.

chesterxgchen:
> @pcnudde Thanks for catching the double-conversion risk. Fixed in [1350bae](https://github.com/NVIDIA/NVFlare/commit/1350baed13290ce71b7505123d1191e8028356e1) — removed the converter calls directly from `LauncherExecutor.execute()`. Since `LauncherExecutor` is only ever subclassed by `ClientAPILauncherExecutor` (subprocess path), this cleanly eliminates CJ-side conversion with no impact on the in-process path.

Yes, no more conversion on the CJ side

nvflare/client/flare_agent_with_fl_model.py
```diff
@@ -12,20 +12,59 @@
...
+    ):
+        super().__init__(*args, **kwargs)
+        self.from_nvflare_converter = from_nvflare_converter
+        self.to_nvflare_converter = to_nvflare_converter
+        self._converter_ctx = _ConverterContext()
+
     def shareable_to_task_data(self, shareable: Shareable) -> FLModel:
+        if self.from_nvflare_converter is not None:
+            task_name = self.current_task.task_name if self.current_task else ""
+            shareable = self.from_nvflare_converter.process(task_name, shareable, self._converter_ctx)
```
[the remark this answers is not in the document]
chesterxgchen: fixed

chesterxgchen:
> Additional Comments (1)
> **`nvflare/app_common/executors/launcher_executor.py`** **`_init_converter` loads converters that are never used**
> 
> `initialize()` still calls `_init_converter()` (line 126), which resolves `from_nvflare_converter_id` / `to_nvflare_converter_id` into converter objects and stores them in `self._from_nvflare_converter` / `self._to_nvflare_converter`. But after this PR removes the converter calls from `execute()`, these objects are loaded and never referenced again.
> 
> This means any user who explicitly configures `from_nvflare_converter_id` or `to_nvflare_converter_id` on a `LauncherExecutor` (or its subclasses like `PTClientAPILauncherExecutor`) will have their custom converters silently ignored. The PR description notes this as a known limitation, but it would be worth either:
> 
> 1. Logging a warning when these IDs are configured (to avoid silent failures), or
> 2. Removing `_init_converter()` from `initialize()` entirely and documenting the deprecation of these parameters on the launcher executor path.

fixed

PR #4228, opened 2026-02-24
[2.7] Fix recipe API bug list and harden recipe behavior

chesterxgchen:
## Issues addressed
This PR addresses the reported recipe bug list:

- `BUG-8` (Critical): `CyclicRecipe` `model=None + initial_ckpt` created phantom persistor reference and runtime failure.
- `BUG-1` (Critical): `CyclicRecipe` dict model config caused `TypeError`.
- `BUG-7` (Critical): base `CyclicRecipe` silently ignored `initial_ckpt`.
- `BUG-9` (High): `FedAvgRecipe` `per_site_config[\"server\"]` caused invalid target handling/crash.
- `BUG-4` (High): `Recipe.execute/export` mutated job state and made recipe reuse unsafe.
- `BUG-2` (Medium): `FedAvgRecipe` per-site `or` fallback ignored falsy overrides.
- `BUG-3` (Medium): `SimEnv` with `num_clients=0` and client list resolved `num_threads=0`.
- `BUG-5` (Low): base `FedAvgRecipe` accepted `initial_ckpt` + `model=None` but produced incomplete model source.
- `BUG-6` (Low): `add_cross_site_evaluation` idempotency guard depended only on transient `_cse_added` attribute.

## Changes
- Hardened base `CyclicRecipe` model/persistor setup:
  - fail-fast when a valid persistor cannot be configured;
  - support dict model config for supported frameworks;
  - apply/validate `initial_ckpt` for wrapper and framework-specific paths.
- Hardened base `FedAvgRecipe`:
  - validate and reject reserved `per_site_config` targets (`server`, `@ALL`);
  - preserve falsy per-site override values via explicit `is not None` fallback;
  - fail-fast when no persistor and no model params are available.
- Added shared framework persistor setup utility in recipe utils and wired both Cyclic/FedAvg paths to it.
- Made `Recipe.execute()` and `Recipe.export()` reusable-safe by snapshot/restore of temporary execution params.
- Fixed `SimEnv` default resolution for `num_clients`/`num_threads` when explicit client list is provided.
- Hardened CSE idempotency with workflow-based detection in addition to `_cse_added` fast-path.
- Updated impacted examples and exports:
  - `hello-cyclic` now passes `min_clients` explicitly;
  - `hello-numpy` uses transfer-type API;
  - recipe module exports include `add_cross_site_evaluation`.
- Expanded recipe unit tests with targeted regressions for all above behaviors.

## Reason for these changes
These updates enforce clear model-source contracts and remove silent/implicit fallback behavior that could generate invalid jobs or runtime crashes. The goal is deterministic recipe behavior, safe object reuse, and predictable per-site/CSE configuration semantics in 2.7.

## Affected files
Core implementation:
- `nvflare/recipe/cyclic.py`
- `nvflare/recipe/fedavg.py`
- `nvflare/recipe/spec.py`
- `nvflare/recipe/sim_env.py`
- `nvflare/recipe/utils.py`
- `nvflare/recipe/poc_env.py`
- `nvflare/recipe/__init__.py`

Framework wrappers / examples:
- `nvflare/app_opt/pt/recipes/cyclic.py`
- `nvflare/app_opt/tf/recipes/cyclic.py`
- `examples/hello-world/hello-cyclic/job.py`
- `examples/hello-world/hello-numpy/job.py`
- `examples/tutorials/job_recipe.ipynb`

Tests:
- `tests/unit_test/recipe/cyclic_recipe_test.py`
- `tests/unit_test/recipe/fedavg_recipe_test.py`
- `tests/unit_test/recipe/spec_test.py`
- `tests/unit_test/recipe/sim_env_test.py`
- `tests/unit_test/recipe/utils_test.py`

## Test strategy
- Targeted regression suites:
  - `pytest -q tests/unit_test/recipe/cyclic_recipe_test.py tests/unit_test/recipe/fedavg_recipe_test.py tests/unit_test/recipe/spec_test.py`
  - `pytest -q tests/unit_test/recipe/sim_env_test.py tests/unit_test/recipe/utils_test.py`
- Full recipe suite:
  - `pytest -q tests/unit_test/recipe`
- Result: recipe suite passed (`183 passed, 13 skipped`).

Made with [Cursor](https://cursor.com)

### DIFF
--- a/nvflare/app_opt/pt/recipes/fedavg.py
+++ b/nvflare/app_opt/pt/recipes/fedavg.py
@@ -168,6 +168,12 @@ class FedAvgRecipe(UnifiedFedAvgRecipe):
 
         persistor_id = setup_custom_persistor(job=job, model_persistor=self.model_persistor)
         if persistor_id:
+            if hasattr(job, "comp_ids"):
+                job.comp_ids["persistor_id"] = persistor_id
+                if self._pt_model_locator is not None:
+                    locator_id = job.to_server(self._pt_model_locator, id="locator")
+                    if isinstance(locator_id, str) and locator_id:
+                        job.comp_ids["locator_id"] = locator_id
             return persistor_id
 
         ckpt_path = resolve_initial_ckpt(self.initial_ckpt, getattr(self, "_prepared_initial_ckpt", None), job)
@@ -184,4 +190,10 @@ class FedAvgRecipe(UnifiedFedAvgRecipe):
             locator=self._pt_model_locator,
             allow_numpy_conversion=allow_numpy_conversion,
         )
-        return extract_persistor_id(job.to_server(pt_model, id="persistor"))
+        result = job.to_server(pt_model, id="persistor")
+        if isinstance(result, dict) and hasattr(job, "comp_ids"):
+            job.comp_ids.update(result)
+        persistor_id = extract_persistor_id(result)
+        if persistor_id and hasattr(job, "comp_ids"):
+            job.comp_ids.setdefault("persistor_id", persistor_id)
+        return persistor_id
--- a/tests/unit_test/recipe/fedavg_recipe_test.py
+++ b/tests/unit_test/recipe/fedavg_recipe_test.py
@@ -19,6 +19,8 @@ import torch.nn as nn
 
 from nvflare.apis.job_def import ALL_SITES, SERVER_SITE_NAME
 from nvflare.app_common.abstract.fl_model import FLModel
+from nvflare.app_common.abstract.model_locator import ModelLocator
+from nvflare.app_common.abstract.model_persistor import ModelPersistor
 from nvflare.app_common.aggregators.model_aggregator import ModelAggregator
 from nvflare.app_common.np.recipes import NumpyFedAvgRecipe
 from nvflare.app_common.widgets.intime_model_selector import IntimeModelSelector
@@ -77,6 +79,26 @@ class InvalidAggregator:
 
     def __init__(self):
         pass
+
+
+class DummyPersistor(ModelPersistor):
+    """Minimal ModelPersistor used to test custom persistor wiring."""
+
+    def load_model(self, fl_ctx):
+        return {}
+
+    def save_model(self, model, fl_ctx):
+        return None
+
+
+class DummyLocator(ModelLocator):
+    """Minimal ModelLocator used to test explicit locator registration."""
+
+    def get_model_names(self, fl_ctx):
+        return []
+
+    def locate_model(self, model_name, fl_ctx):
+        return None
 
 
 @pytest.fixture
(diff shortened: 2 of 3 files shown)


nvflare/recipe/utils.py
```diff
@@ -470,6 +495,99 @@ def prepare_initial_ckpt(initial_ckpt: Optional[str], job) -> Optional[str]:
...
+    raise_on_unsupported: bool = False,
+    recipe_name: str = "Recipe",
+    unsupported_framework_message: Optional[str] = None,
+    prepared_initial_ckpt: Optional[str] = None,
+) -> str:
+    """Setup framework-specific model persistor and return persistor_id.
+
+    This consolidates common PT/TF/NumPy persistor setup shared by recipes.
+    """
+    if model_persistor is not None:
```
IsaacYangSLA: This seems a lot of if statements, which was prevsiously not recommended.  Should we allow this from now on?

chesterxgchen: good point, let me see how can we handle this better

nvflare/recipe/utils.py
```diff
@@ -470,6 +495,99 @@ def prepare_initial_ckpt(initial_ckpt: Optional[str], job) -> Optional[str]:
     return os.path.basename(initial_ckpt)
 
 
+def _extract_persistor_id(result: Any) -> str:
```
nvidianz: The underscore prefix normally means private method. But this one is exported and used in other modules.

There are similar issues with other functions in this and spec.py.

nvflare/recipe/spec.py
```diff
@@ -127,6 +128,47 @@ def process_env(self, env: ExecEnv):
...
+        if server_exec_params or client_exec_params:
+            params_snapshot = self._snapshot_additional_params()
+
+        try:
+            if server_exec_params:
+                self.job.to_server(server_exec_params)
+
+            if client_exec_params:
+                self._add_to_client_apps(client_exec_params)
+            yield
```
IsaacYangSLA, line 167: Do we need some lock around here?  If during yield, some codes modify self, it could cause inconsistencies.

nvflare/recipe/utils.py
chesterxgchen: Addressed. I moved framework-specific persistor setup out of the shared utils helper and into framework recipe subclasses (PT/TF/NumPy), so the branch-heavy conditional block is removed from the common path.

nvflare/recipe/utils.py
chesterxgchen: Addressed. I renamed shared helpers that are used across modules to public names (, ) and updated all call sites accordingly.

nvflare/recipe/spec.py
chesterxgchen: Updated based on follow-up discussion: no lock was added. We keep the temporary-param snapshot/restore behavior for execute/export, and avoid introducing lock contention or blocking semantics for callers.

nvflare/recipe/utils.py
chesterxgchen: Clarification: renamed to public helper names extract_persistor_id and collect_non_local_scripts, and updated all call sites.

PR #4231, opened 2026-02-24
[2.7] Fix client-side RSS memory growth and subprocess logging gap

chesterxgchen:
## Problem Statement

Three related issues were discovered in the NVFlare client-side stack:

### Issue 1: RSS Memory Growth Across FL Rounds

After each federated learning round, the client process RSS grew monotonically. Model parameters — both the received global model and the locally trained model — were kept alive long after `flare.send()` completed. The reference chain kept them live until the next `receive()` overwrote `self.fl_model`, meaning one full round's worth of tensors was always double-buffered in memory.

### Issue 2: Race Condition in `PipeHandler.send_to_peer()` — Premature `delete_msg_root`

The original `send_to_peer()` had:

```python
finally:
    # the msg_id is also used as msg_root_id
    delete_msg_root(msg.msg_id)
```

This was a race condition. `CellPipe.send_request()` returns as soon as the peer acknowledges receipt via `make_reply(ReturnCode.OK)` in `_receive_message` — which happens **before** the peer finishes downloading model params. The `finally` block fired `delete_msg_root(msg.msg_id)` and destroyed the active download transaction while the CJ was still pulling tensors from the subprocess, causing intermittent download failures.

### Issue 3: Silent Logging Gap in Subprocess Mode

When `launch_external_process=True`, `SubprocessLauncher` spawns `python3 -u client.py`. This new process starts with Python's unconfigured logging system — `lastResort` only emits `WARNING+`, so every `logger.info()` inside the subprocess is silently dropped. All INFO-level diagnostics (NVFlare internals, user training progress) were invisible in subprocess mode.

---

### Fix 1: Release Model Params Eagerly After `send()` — commit `39eb631d`

**Files:** `nvflare/client/in_process/api.py`, `nvflare/client/ex_process/api.py`, `nvflare/client/model_registry.py`

After `flare.send()`, both the sent model's params and the received global model's params are dead weight — already serialized and transmitted. We now explicitly null them immediately when `clear_cache=True` (the default):

- **`InProcessClientAPI`**: Nulls `model.params`, `model.optimizer_params`, `self.fl_model.params`, `self.fl_model.optimizer_params` before clearing `self.fl_model`.
- **`ExProcessClientAPI`**: Calls `model_registry.release_params(model)` which nulls both the sent and received model's params.

**Why not `gc.collect()`?** CPython's reference counting reclaims objects as soon as their refcount drops to zero — `gc.collect()` only helps with reference *cycles*. PyTorch tensors don't form cycles. Once the params dicts are nulled, tensor memory is freed immediately. Confirmed experimentally: RSS stays flat without `gc.collect()`.

**Why null params specifically?** Params (the large tensor dicts) are the dominant memory consumer. Nulling them frees the tensors while keeping the `FLModel` shell (metadata, metrics, round) alive for any code that reads those fields after `send()`.

### Fix 2: Remove Premature `delete_msg_root` from `PipeHandler.send_to_peer()` — commit `39eb631d`

**File:** `nvflare/fuel/utils/pipe/pipe_handler.py`

Removed the `finally: delete_msg_root(msg.msg_id)` block and its import. The `msg_id` doubles as the `msg_root_id` used by `ViaDownloaderDecomposer` to track download transaction lifecycle. Calling `delete_msg_root` in a `finally` block races against the peer's download: the ack is immediate but the actual tensor download is asynchronous. Download transaction cleanup now correctly happens in `wf_comm_server.py` and `task_controller.py`, which call `delete_msg_root` only after the task is fully processed server-side.

### Fix 3: Configure Subprocess Python Logging on `flare.init()` — commit `bc6d87847`

**File:** `nvflare/client/ex_process/api.py` — added `_configure_subprocess_logging()`

Immediately after loading `client_api_config.json`, `init()` calls `_configure_subprocess_logging(client_config)`, which:

1. Reads `workspace_dir` from `client_api_config.json` under `TASK_EXCHANGE.pipe.ARG.workspace_dir`
2. Loads `{workspace_dir}/local/log_config.json` (the site's standard NVFlare log config)
3. Calls `apply_log_config(dict_config, workspace_dir)` to invoke `logging.config.dictConfig()`

This wires up `consoleHandler -> sys.stdout` (captured by `SubprocessLauncher`), `logFileHandler -> log.txt`, and `jsonFileHandler -> log.json` — exactly as the main NVFlare process configures them. The entire method is wrapped in `try/except Exception: pass` — a logging setup failure must never crash the training script.

Called before the `try` block in `init()` so that even errors during `flare_agent.start()` are properly logged.

### Fix 4: RSS Memory Profiling Utility — commit `bc6d87847`

**New file:** `nvflare/fuel/utils/mem_utils.py`

Zero-cost diagnostic to measure client-side RSS per round:

- Gated by `NVFLARE_CLIENT_MEMORY_PROFILE=1` (off by default). Name is client-scoped so server/relay processes are unaffected.
- `_ENABLED` evaluated once at import time — single boolean check per call, zero overhead when disabled.
- Uses `logger.info()`: after Fix 3, works correctly in both modes. RSS lines land in `log.txt`/`log.json` alongside all other NVFlare diagnostics.
- `psutil` imported lazily; entire call is `try/except`-wrapped so missing install is a silent no-op.
- Instrumented at `receive()` and `send()` in both `InProcessClientAPI` and `ExProcessClientAPI`.

---

## Files Changed

| File | Change |
|------|--------|
| `nvflare/client/api_spec.py` | `_maybe_cleanup_memory()` base implementation and memory management attributes |
| `nvflare/client/in_process/api.py` | Eager param release in `send()`, `log_rss()` instrumentation, `configure_memory_management()` |
| `nvflare/client/ex_process/api.py` | Eager param release, `_configure_subprocess_logging()`, `log_rss()` instrumentation, memory GC config |
| `nvflare/client/model_registry.py` | Added `release_params()` method |
| `nvflare/fuel/utils/pipe/pipe_handler.py` | Removed racy `finally: delete_msg_root(msg.msg_id)` and its import |
| `nvflare/fuel/utils/mem_utils.py` | **New file** — `log_rss(tag)` RSS profiling utility |
| `docs/client_api.rst` | Clarified param lifecycle and `clear_cache` semantics |

---

## Verification

Tested with `hello-pt` (CIFAR-10, 2 clients, 2 rounds) using `NVFLARE_CLIENT_MEMORY_PROFILE=1`:

**In-process mode:**
```
round=0 after_receive: 629.2 MB
round=0 after_send:    270.9 MB   <- params freed immediately
round=1 after_receive: 283.4 MB
round=1 after_send:    448.5 MB   <- flat, no accumulation
```

**Subprocess mode (`launch_external_process=True`):**
```
round=0 after_receive: 623.5 MB
round=0 after_send:    129.4 MB   <- params freed immediately
round=1 after_receive: 136.1 MB
round=1 after_send:    121.5 MB   <- flat, no accumulation
```

RSS stabilizes after round 0 in both modes. Subprocess logging confirmed: `logger.info()` lines now appear in `site-N/log.txt` (written directly by the subprocess) and captured via `SubprocessLauncher`.

## Test plan
- [x] In-process RSS profiling with `NVFLARE_CLIENT_MEMORY_PROFILE=1` — `[RSS]` lines in `site-N/log.txt`
- [x] Subprocess RSS profiling with `launch_external_process=True` — `[RSS]` lines in both subprocess log and `SubprocessLauncher` capture
- [x] Subprocess logging gap fixed — `logger.info()` calls now visible (previously silently dropped)
- [x] Memory flat across rounds in both modes (no monotonic growth)
- [x] `flare.send(clear_cache=False)` path untouched — params only released when `clear_cache=True`
- [x] `delete_msg_root` race removed — download transactions no longer prematurely destroyed

### DIFF
--- a/tests/unit_test/client/test_memory_release_simulation.py
+++ b/tests/unit_test/client/test_memory_release_simulation.py
@@ -86,8 +86,21 @@ class TestInProcessMemoryRelease(unittest.TestCase):
     NUM_ROUNDS = 5
 
     def setUp(self):
+        # Reset the DataBus singleton before each test to prevent stale
+        # subscriptions from other tests contaminating refcount measurements.
+        from nvflare.fuel.data_event.data_bus import DataBus
+
+        with DataBus._lock:
+            DataBus._instance = None
         self.api = InProcessClientAPI(_task_metadata())
         self.api.init()
+
+    def tearDown(self):
+        # Clean up singleton so subsequent tests start fresh.
+        from nvflare.fuel.data_event.data_bus import DataBus
+
+        with DataBus._lock:
+            DataBus._instance = None
 
     def _one_round(self, round_num: int):
         """Simulate one FL round: receive → train → send."""


nvflare/client/in_process/api.py
```diff
@@ -150,11 +154,19 @@ def send(self, model: FLModel, clear_cache: bool = True) -> None:
         self.event_manager.fire_event(TOPIC_LOCAL_RESULT, shareable)
 
         if clear_cache:
+            # Serialization is complete. Release the sent model's params and the
+            # received model's params — both are dead weight after flare.send().
+            # NOTE: model.params and input_model.params will be None after this.
+            model.params = None
+            model.optimizer_params = None
+            if self.fl_model:
```
IsaacYangSLA: If we later do self.fl_model = None, do we need this if block?

nvflare/client/ex_process/api.py
```diff
@@ -93,6 +94,33 @@ def __init__(self, config_file: str):
...
+            workspace_dir = pipe_args.get("workspace_dir", "")
+            if not workspace_dir:
+                return
+            log_config_path = os.path.join(workspace_dir, "local", "log_config.json")
+            if not os.path.isfile(log_config_path):
+                return
+            with open(log_config_path) as f:
+                dict_config = json.load(f)
+            apply_log_config(dict_config, workspace_dir)
+        except Exception:
```
nvidianz: Should log a warning here

pcnudde (approved):

PR #4233, opened 2026-02-25
[2.7] Fix reverse pass-through msg-root lifetime and cleanup

pcnudde:
## Summary
- enable reverse pass-through on the task pipe cell in `ClientAPILauncherExecutor`
- keep one `MSG_ROOT_ID` across task-result retries and delete it exactly once in `ClientRunner`
- set `MSG_ROOT_TTL` on `CellPipe` replies to keep reverse-direction download refs alive long enough
- clarify send-path naming/comments (`_wait_task_ready_and_send_once`) for review readability

### DIFF
--- a/nvflare/private/fed/client/client_runner.py
+++ b/nvflare/private/fed/client/client_runner.py
@@ -557,39 +557,43 @@ class ClientRunner(TBI):
         submit_timeouts = [base_submit_timeout * (2**i) for i in range(_SUBMIT_RESULT_MAX_RETRIES)]
         # Invariant: we only create download transactions after TASK_CHECK succeeds, so abort/TASK_GONE before send has nothing to clean.
         retried = False
-        for try_count, submit_timeout in enumerate(submit_timeouts, start=1):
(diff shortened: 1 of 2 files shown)


nvflare/private/fed/client/client_runner.py
```diff
@@ -545,30 +547,57 @@ def fetch_and_run_one_task(self, fl_ctx) -> (float, bool):
...
+                            fl_ctx,
+                            f"failed to send task result to {self.parent_target} after "
+                            f"{_SUBMIT_RESULT_MAX_RETRIES} tries",
+                        )
+                        return False
+
+                    retried = True
+                    time.sleep(self.task_check_interval)
+        except Exception:
+            delete_msg_root(msg_root_id)
```
YuanTingHsieh, line 595: this is the CJ side, do we need similar thing for the external process?
The External → CJ path might have a resource leak on retries — download transactions accumulate on the external process and are never explicitly cleaned up

PR #4282, opened 2026-03-09
Design document for Job Launcher and Job Handle (TA: NVFlare developers)

IsaacYangSLA:
### Description

The abstract layers for JobLauncher/JobHandle to support three environments, and how the upper layers (client and server) call the concrete classes to launch job.

### Types of changes

- [x] Documentation updated.

### DIFF
--- a/docs/design/JobLauncher_and_JobHandle.md
+++ b/docs/design/JobLauncher_and_JobHandle.md
@@ -51,7 +51,7 @@ Abstract base class for launching jobs. Extends `FLComponent`, which gives it ac
 
 | Method | Signature | Semantics |
 |--------|-----------|-----------|
-| `launch_job(job_meta, fl_ctx)` | `(dict, FLContext) -> JobHandleSpec` | Start a job and return its handle. Returns `None` on failure. |
+| `launch_job(job_meta, fl_ctx)` | `(dict, FLContext) -> JobHandleSpec` | Start a job and return its handle. |
 
 ### 2.3 Supporting Types
 
--- a/docs/design/JobLauncher_and_JobHandle.md
+++ b/docs/design/JobLauncher_and_JobHandle.md
@@ -1,0 +1,605 @@
+# JobLauncher and JobHandle Design Document
+
+## 1. Overview
+
+NVFlare runs each federated job as an isolated execution unit -- a subprocess, Docker container, or Kubernetes pod. Two abstractions govern this:
+
+- **JobLauncherSpec** -- starts a job and returns a handle.
+- **JobHandleSpec** -- represents the running job and provides lifecycle control (poll, wait, terminate).
+
+The upper layers (server engine, client executor) program exclusively against these two interfaces. The concrete backend is selected at runtime through an event-based mechanism, so the engine never imports or names a specific launcher type.
+
+```
+┌──────────────────────────────────────────────────────────┐
+│                    Upper Layer                           │
+│     ServerEngine  /  ClientExecutor                      │
+│                                                          │
+│  1. Build JOB_PROCESS_ARGS                               │
+│  2. get_job_launcher(job_meta, fl_ctx) → launcher        │
+│  3. launcher.launch_job(job_meta, fl_ctx) → job_handle   │
+│  4. job_handle.wait()  /  job_handle.terminate()         │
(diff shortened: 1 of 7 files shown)


docs/design/JobLauncher_and_JobHandle.md
```diff
@@ -0,0 +1,605 @@
...
+## 2. Specification Layer (`nvflare/apis/job_launcher_spec.py`)
+
+### 2.1 JobHandleSpec
+
+Abstract base class representing a running job. All methods are `@abstractmethod`.
+
+| Method | Signature | Semantics |
+|--------|-----------|-----------|
+| `terminate()` | `() -> None` | Stop the job immediately. |
+| `poll()` | `() -> JobReturnCode` | Non-blocking query for the job's current return code. Returns `UNKNOWN` while still running. |
```
chesterxgchen, line 45: Why the running instance return "UNKNOWN" instead of "RUNNING" ?

docs/design/JobLauncher_and_JobHandle.md
```diff
@@ -0,0 +1,605 @@
...
+|--------|---------------|
+| `terminate()` | `container.stop()`. |
+| `poll()` | Re-fetches container via `docker.from_env().containers.get(id)`. Maps status: `EXITED` → `SUCCESS`, `DEAD` → `ABORTED`, all others → `UNKNOWN`. Removes the container on terminal states. |
+| `wait()` | `enter_states([EXITED, DEAD], timeout)` -- polls container status in a 1-second loop until a terminal state is reached. |
+
+Docker container states and their mappings:
+
+| Docker Status | JobReturnCode |
+|---------------|---------------|
+| `created` | `UNKNOWN` |
```
chesterxgchen, line 280: why status can't reflect the real status ?

docs/design/JobLauncher_and_JobHandle.md
```diff
@@ -0,0 +1,605 @@
...
+| Method | Implementation |
+|--------|---------------|
+| `terminate()` | `delete_namespaced_pod(name, namespace, grace_period_seconds=0)` then `enter_states([TERMINATED], timeout)`. Sets `terminal_state = TERMINATED`. |
+| `poll()` | If `terminal_state` is set, return it directly. Otherwise call `_query_state()` to read the pod phase and map it to `JobReturnCode`. |
+| `wait()` | `enter_states([SUCCEEDED, TERMINATED])` -- polls pod phase in a 1-second loop. |
+
+Pod phase mapping:
+
+| Pod Phase | JobState | JobReturnCode |
+|-----------|----------|---------------|
```
chesterxgchen, line 347: same where why can't the status starting,running instead of unknown

nvflare/private/fed/client/communicator.py
```diff
@@ -273,9 +273,9 @@ def client_registration(self, client_name, project_name, fl_ctx: FLContext):
         start = time.time()
         while not self.cell:
             self.logger.info("Waiting for the client cell to be created.")
-            if time.time() - start > 15.0:
+            if time.time() - start > 600:
                 raise RuntimeError("Client cell could not be created. Failed to login the client.")
```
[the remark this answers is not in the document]
chesterxgchen: these are needed default for large models, Docker and K8s env can overwrite with shorter timeout if needed. @Greptileai

docs/design/JobLauncher_and_JobHandle.md
IsaacYangSLA: This is the 'exit code' of a handle.  In process case, before the process exits, there is no exit code and it will be UNKNOWN.  If the process exits normally, the exit code will be 0.  The high-level status code will be SUCCESS.  In k8s, there is no clear exit code.  So once we detect POD is in SUCCEEDED or TERMINATED state, we know it exits and then we set the high-level status code (SUCCESS or TERMINATED).

docs/design/JobLauncher_and_JobHandle.md
IsaacYangSLA: There are two concepts.  One is the state the job is in.  The other is once the job exits, what exit code (status code is returns.  The POD only has states, so we have to map the state to both concepts.

PR #4301, opened 2026-03-12
[2.7]: Sign the root folders of startup kits

IsaacYangSLA:
Two folders, startup and local, of any startup kit were signed at provisioning time.  At the verification stage, nvflare checks the startup folder.  In CC environment, one more verification was done before nvflare started.  However, the verification checks from the root folder of a startup kit.  Therefore, there is a gap as the root folder and transfer folder are not signed.  This PR changed the signing process to apply to the root and all sub folders.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/lighter/impl/signature.py
+++ b/nvflare/lighter/impl/signature.py
@@ -30,7 +30,5 @@ class SignatureBuilder(Builder):
             raise RuntimeError(f"missing {CtxKey.ROOT_PRI_KEY} in ProvisionContext")
 
         for p in project.get_all_participants():
-            dest_dir = ctx.get_kit_dir(p)
-            sign_folders(dest_dir, root_pri_key, signature_file=ProvFileName.SIGNATURE_JSON)
-            dest_dir = ctx.get_local_dir(p)
+            dest_dir = ctx.get_ws_dir(p)
             sign_folders(dest_dir, root_pri_key, signature_file=ProvFileName.SIGNATURE_JSON)


nvflare/lighter/impl/signature.py
```diff
@@ -30,7 +30,5 @@ def build(self, project: Project, ctx: ProvisionContext):
             raise RuntimeError(f"missing {CtxKey.ROOT_PRI_KEY} in ProvisionContext")
 
         for p in project.get_all_participants():
-            dest_dir = ctx.get_kit_dir(p)
-            sign_folders(dest_dir, root_pri_key, signature_file=ProvFileName.SIGNATURE_JSON)
-            dest_dir = ctx.get_local_dir(p)
+            dest_dir = ctx.get_ws_dir(p)
```
chesterxgchen, line 33: what is this directory ? whats the implications, this workspace for the provision ? 
or the for root folder of the startup kits. 
What it means sign the root folder of the startup kit, no thing can be modified in the folder ? or only the existing files in the startup kit during provision ?

chesterxgchen (commented):
few questions

nvflare/lighter/impl/signature.py
IsaacYangSLA: This is the workspace of 'provision,' not the workspace for nvflare runtime.  The concept of workspace in provision changed a few times.  Currently, every participant has its own workspace.  So ctx.get_ws_dir(p) will return the root folder of its startup kit.  Signing that folder will generate signature.json in each folder, including the root folder.

chesterxgchen: does it means, what we have a signature.json  in each folders, but I can add any files ? can I add new folders ? What if I copy the signature.json to the newly added folder, will that work ?

IsaacYangSLA: Every folder has one signature file.  You can add/edit/delete any file, create/delete any folder but the results are the entire folder structure will fail signature check against signature files.

chesterxgchen: this means user can not add custom code ? modify configuration ?

IsaacYangSLA: You can copy the signature.json file to any newly created folder.  At the verification stage, that folder and signature file will not pass the signature check.

chesterxgchen (changes requested):
only CC mode can do this

nvflare/lighter/impl/signature.py
chesterxgchen: This change is not correct, we only need to do this for CC provision, not for all cases

nvflare/lighter/impl/signature.py
```diff
@@ -30,7 +30,5 @@ def build(self, project: Project, ctx: ProvisionContext):
...
             sign_folders(dest_dir, root_pri_key, signature_file=ProvFileName.SIGNATURE_JSON)
```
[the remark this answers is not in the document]
YuanTingHsieh: Our transfer folder will have new job folders created inside would that break anything?

Or we just verify at the start so we are ok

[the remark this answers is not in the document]
YuanTingHsieh: This is interesting, wondering if we have a check to say this needs to be the last builder touching these files

nvflare/lighter/impl/signature.py
nvidianz: With this change, can user still change files in `local` folder? Many configuration parameters are in that folder and user should be able to modify the files.

nvflare/lighter/impl/signature.py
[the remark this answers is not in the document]
IsaacYangSLA: Finalize should not generate files.  The purpose of finalize is the clean up itself and releases resources.  It looks like we are doing something strange.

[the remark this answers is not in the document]
IsaacYangSLA: There are two verification points.  When nvflare starts, currently it only checks startup folder.  The local and transfer folders are not checked.  The second verification point is before nvflare starts in CVM docker through the --verify option.  At that moment, the transfer folder should be empty and the local folder should be what the provisioning tool generates.

nvflare/lighter/impl/signature.py
IsaacYangSLA: I think we need a meeting to explain it.

chesterxgchen: Isaac explain to me, so we have two phases: 1) signing the the folders 2) verify the folders with signature.json

This process is step 1) signing.  Isaac points that we always sign all the folders regardless. But which folder to verify depends on the cases in step 2.  for CVM Case, all folders are verified. for non-CC cases, only the Startup kit/start folder are verified, users are free to add and modify the file or folder, as the signature.json in those folders are ignored.

nvflare/lighter/impl/signature.py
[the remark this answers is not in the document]
chesterxgchen: @IsaacYangSLA @YuanTingHsieh  can you verify we are actually doing something in the finalize ?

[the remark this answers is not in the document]
IsaacYangSLA: comm_options.json is inside local folder.  NVFlare does not verify that folder at starting time.  start_all.sh seems related to poc and it's not even in a startup kit.  The issue is whether CC updates comm_options.json during provisioning time.

nvflare/lighter/impl/signature.py
IsaacYangSLA: Signing and verifying are different.  local folder was signed before, but was never verified by nvflare at starting time.  With this one, local folder is still signed and still not verified.

IsaacYangSLA:
> only CC mode can do this

We can sign the entire startup kit in all cases.  The verification point chooses which folder to verify.

nvflare/lighter/impl/signature.py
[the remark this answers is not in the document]
YuanTingHsieh: as far as i know CC does not change the comm_options.json @nvidianz can you confirm?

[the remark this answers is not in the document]
YuanTingHsieh: makes sense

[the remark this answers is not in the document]
nvidianz: Correct!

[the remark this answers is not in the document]
IsaacYangSLA: If CC does not change comm_options.json, then it should be safe as the regular nvflare ignores local folder for signature verification.

YuanTingHsieh (approved):

chesterxgchen (approved):
LGTM

PR #4336, opened 2026-03-19
Job Launcher and Job Handle: design doc/implementation/unit tests (TA: NVFlare developers)

IsaacYangSLA:
### Description

Job Launcher and Job Handle design docs
Implementation of Job Launcher and Job Handle for kubernetes (docker is WIP)
Unit tests for implementation

### Types of changes

- [x] Documentation updated.


nvflare/app_opt/job_launcher/k8s_launcher.py
```diff
@@ -155,53 +168,93 @@ def enter_states(self, job_states_to_enter: list, timeout=None):
...
+                return True
+        return False
 
     def wait(self):
-        self.enter_states([JobState.SUCCEEDED, JobState.TERMINATED])
+        while True:
+            if self.terminal_state is not None:
+                return
+            job_state = self._query_state()
+            if job_state in (JobState.SUCCEEDED, JobState.TERMINATED):
```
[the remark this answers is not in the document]
IsaacYangSLA: This wait is used by the engine to wait for the job pod finishing the job.  It could be very long.  Current process-based wait also waits forever.

nvflare/app_common/resource_consumers/be_resource_consumer.py
```diff
@@ -0,0 +1,21 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from nvflare.apis.resource_manager_spec import ResourceConsumerSpec
+
+
+class BEResourceConsumer(ResourceConsumerSpec):
+
+    def consume(self, resources: dict):
+        pass
```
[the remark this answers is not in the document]
IsaacYangSLA: This is Best Effort (BE).

nvflare/app_opt/job_launcher/k8s_launcher.py
```diff
@@ -233,25 +284,27 @@ def launch_job(self, job_meta: dict, fl_ctx: FLContext) -> JobHandleSpec:
...
-                job_handle.terminate()
-                return None
-        except ApiException:
+            self.core_v1.create_namespaced_pod(body=pod_manifest, namespace=self.namespace)
+            job_handle.enter_states([JobState.RUNNING], timeout=self.timeout)
+            return job_handle
+        except ApiException as e:
             job_handle.terminate()
-            return None
+            return job_handle
```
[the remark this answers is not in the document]
IsaacYangSLA: The docker behavior is outdated.

nvflare/app_opt/job_launcher/k8s_launcher.py
```diff
@@ -155,53 +168,93 @@ def enter_states(self, job_states_to_enter: list, timeout=None):
...
 
     def wait(self):
-        self.enter_states([JobState.SUCCEEDED, JobState.TERMINATED])
+        while True:
+            if self.terminal_state is not None:
+                return
+            job_state = self._query_state()
+            if job_state in (JobState.SUCCEEDED, JobState.TERMINATED):
+                return
+            time.sleep(1)
```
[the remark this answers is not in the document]
IsaacYangSLA: Wait in process-based job is also forever.  This is to match process-based job handle.

nvflare/app_opt/job_launcher/k8s_launcher.py
```diff
@@ -155,53 +168,93 @@ def enter_states(self, job_states_to_enter: list, timeout=None):
...
+        self.data_pvc_file_path = data_pvc_file_path
         self.timeout = timeout
+        self.namespace = namespace
+        with open(data_pvc_file_path, "rt") as f:
+            data_pvc_dict = yaml.safe_load(f)
+        if not data_pvc_dict:
+            raise ValueError(f"data_pvc_file_path '{data_pvc_file_path}' is empty or contains no PVC entries.")
+        # data_pvc_dict will be pvc: mountPath
+        # currently, support one pvc and always mount to /var/tmp/nvflare/data
+        # ie, ignore the mountPath in data_pvc_dict
```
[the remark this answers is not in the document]
IsaacYangSLA: This is the dataset PVC.  It can be an empty pvc.  But this information is critical.  Without it, the job should not  start.

nvflare/app_opt/job_launcher/k8s_launcher.py
```diff
@@ -233,25 +284,27 @@ def launch_job(self, job_meta: dict, fl_ctx: FLContext) -> JobHandleSpec:
...
         try:
-            self.core_v1.create_namespaced_pod(body=job_handle.get_manifest(), namespace=self.namespace)
-            if job_handle.enter_states([JobState.RUNNING], timeout=self.timeout):
-                return job_handle
-            else:
-                job_handle.terminate()
-                return None
-        except ApiException:
+            self.core_v1.create_namespaced_pod(body=pod_manifest, namespace=self.namespace)
+            job_handle.enter_states([JobState.RUNNING], timeout=self.timeout)
```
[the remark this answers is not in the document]
IsaacYangSLA: The upper layer will verify if the handle contains proper information.  If not, upper layer will call terminate.

nvflare/app_opt/job_launcher/k8s_launcher.py
```diff
@@ -68,141 +127,180 @@ def __init__(self, job_id: str, api_instance: core_v1_api, job_config: dict, nam
...
+        return resp.status.phase
+
+    def _query_state(self):
+        pod_phase = self._query_phase()
+        return POD_STATE_MAPPING.get(pod_phase, JobState.UNKNOWN)
+
+    def _stuck_in_pending(self, current_phase):
+        if current_phase == POD_Phase.PENDING.value:
+            self._stuck_count += 1
+            if self._max_stuck_count is not None and self._stuck_count >= self._max_stuck_count:
```
nvkevlu, line 259: Design text says `_stuck_count > _max_stuck_count`; code uses `_stuck_count >= _max_stuck_count`. Suggest to align the doc with `>=` (or change code to match the doc)

docs/design/JobLauncher_and_JobHandle.md
```diff
@@ -0,0 +1,581 @@
...
+
+Pod phase mapping:
+
+| Pod Phase | JobState | JobReturnCode |
+|-----------|----------|---------------|
+| `Pending` | `STARTING` | `UNKNOWN` |
+| `Running` | `RUNNING` | `UNKNOWN` |
+| `Succeeded` | `SUCCEEDED` | `SUCCESS` |
+| `Failed` | `TERMINATED` | `ABORTED` |
```
nvkevlu, line 315: mapping pod phase `Failed` → internal `TERMINATED` → `JobReturnCode.ABORTED`

Could make it easy to misread as “user/system abort” instead of “container exited non-zero / OOM / crash.”

So failed K8s jobs surface as `ABORTED`, not `EXECUTION_ERROR`.

**Suggestion:** Either map `Failed` → `EXECUTION_ERROR` (and reserve `ABORTED` for explicit `terminate()` / eviction), or document clearly that for K8s, `ABORTED` includes normal pod failure — and ensure downstream code (`get_return_code`, logging, dashboards) treats that consistently.

nvflare/app_opt/job_launcher/k8s_launcher.py
IsaacYangSLA: Good catch.  Fixed.

docs/design/JobLauncher_and_JobHandle.md
IsaacYangSLA: k8s has only five pod phases.  Failed is one of them.  See [this section](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-phase).  When a pod exits, its phase is either Succeeded or Failed.  Failed can be system termination or any container exits with non-zero value.  So the exit status of the POD Failed can not be precisely determined.  ABORTED is one reasonable assignment.  We will add document on this limitation.

chesterxgchen:
Reviewed comments from Claude

  ---
  PR #4336 Review

  Overall Assessment

  The PR is solid — good abstractions, comprehensive tests, and a clean design doc. A few issues worth flagging.

  Issues

  1. Volume still uses hostPath (critical for the operator question — see below)

  In launch_job():
  "volume_list": [{"name": self.workspace, "hostPath": {"path": self.root_hostpath, "type": "Directory"}}],
  hostPath volumes require the pod to be scheduled on a specific node that has the NVFlare workspace on its local disk. This is
  fragile in real K8s clusters and won't work with most managed K8s deployments (GKE, EKS, AKS). The design doc mentions PVC as the
  intended approach but the implementation still uses hostPath. This needs to be PVC or at minimum documented as a temporary
  limitation.

  2. self.job_handle = None is dead code
  K8sJobLauncher.__init__() sets self.job_handle = None but it's never read — launch_job() creates a local job_handle and returns it.
   Leftover from a previous design.

  3. module_args assigned by reference
  In _make_manifest(), when job_config contains module_args, it's assigned directly:
  self.container_args_module_args_dict = job_config.get("module_args", {...})
  But the fallback dict is a new object. The greptile bot flagged this — caller mutation of job_config["module_args"] would corrupt
  internal state. Should be .copy().

  4. enter_states validation is wrong
  if not all([isinstance(js, JobState)] for js in job_states_to_enter):
  This generates a list of generator objects, all of which are truthy — the validation never catches invalid states. Should be:
  if not all(isinstance(js, JobState) for js in job_states_to_enter):

  5. terminate() in K8sJobHandle waits for TERMINATED state after pod delete, but _query_state() on a deleted pod returns UNKNOWN
  (ApiException), not TERMINATED
  The pod disappears from the API after deletion — read_namespaced_pod raises ApiException which maps to UNKNOWN, not TERMINATED. So
  enter_states([JobState.TERMINATED]) will never return True and will loop until timeout. Should treat UNKNOWN after a delete as
  terminal.

  6. Default image is hardcoded to nvflare/nvflare:2.5.0
  Should default to the current release version or be required.

IsaacYangSLA:
> Reviewed comments from Claude
> 
> PR #4336 Review
> 
> Overall Assessment
> 
> The PR is solid — good abstractions, comprehensive tests, and a clean design doc. A few issues worth flagging.
> 
> Issues
> 
> 1. Volume still uses hostPath (critical for the operator question — see below)
> 
> In launch_job(): "volume_list": [{"name": self.workspace, "hostPath": {"path": self.root_hostpath, "type": "Directory"}}], hostPath volumes require the pod to be scheduled on a specific node that has the NVFlare workspace on its local disk. This is fragile in real K8s clusters and won't work with most managed K8s deployments (GKE, EKS, AKS). The design doc mentions PVC as the intended approach but the implementation still uses hostPath. This needs to be PVC or at minimum documented as a temporary limitation.
> 
> 2. self.job_handle = None is dead code
>    K8sJobLauncher.**init**() sets self.job_handle = None but it's never read — launch_job() creates a local job_handle and returns it.
>    Leftover from a previous design.
> 3. module_args assigned by reference
>    In _make_manifest(), when job_config contains module_args, it's assigned directly:
>    self.container_args_module_args_dict = job_config.get("module_args", {...})
>    But the fallback dict is a new object. The greptile bot flagged this — caller mutation of job_config["module_args"] would corrupt
>    internal state. Should be .copy().
> 4. enter_states validation is wrong
>    if not all([isinstance(js, JobState)] for js in job_states_to_enter):
>    This generates a list of generator objects, all of which are truthy — the validation never catches invalid states. Should be:
>    if not all(isinstance(js, JobState) for js in job_states_to_enter):
> 5. terminate() in K8sJobHandle waits for TERMINATED state after pod delete, but _query_state() on a deleted pod returns UNKNOWN
>    (ApiException), not TERMINATED
>    The pod disappears from the API after deletion — read_namespaced_pod raises ApiException which maps to UNKNOWN, not TERMINATED. So
>    enter_states([JobState.TERMINATED]) will never return True and will loop until timeout. Should treat UNKNOWN after a delete as
>    terminal.
> 6. Default image is hardcoded to nvflare/nvflare:2.5.0
>    Should default to the current release version or be required.

Claude seemed reading the main branch, instead of this PR.  For example, there is no `hostPath` in this PR.

chesterxgchen:
C3 — NVFLARE_PARENT_MONITOR env var: ❌ Still missing
  No env list is added to the container spec anywhere in this PR. Without NVFLARE_PARENT_MONITOR=cellnet, job pods fall back to PID-based parent monitoring. PID 1 always exists in a container so job pods never detect
  parent failure — runaway pods will persist indefinitely. This is a real gap that needs to be in launch_job():
  {"name": "NVFLARE_PARENT_MONITOR", "value": "cellnet"}

  H5 — Pod securityContext: ❌ Still missing
  No securityContext on pod spec or container spec. Running as root in production K8s is blocked by most Pod Security Standards. Low risk for dev/POC but a blocker for enterprise clusters.

  M4 — module_args copy when caller provides dict: ❌ Partial
  When module_args is None, the code correctly does DEFAULT_CONTAINER_ARGS_MODULE_ARGS_DICT.copy(). But when the caller passes a dict, it's self.container_args_module_args_dict = job_config.get("module_args") — still a
  reference, not a copy. Caller mutation would corrupt the handle's args.

  C6 — namespace default still "default": minor
  namespace="default" remains the default in both K8sJobHandle and K8sJobLauncher. It's now a configurable param so this is operator error if misconfigured, not a code bug. Acceptable for now.

IsaacYangSLA:
> C3 — NVFLARE_PARENT_MONITOR env var: ❌ Still missing No env list is added to the container spec anywhere in this PR. Without NVFLARE_PARENT_MONITOR=cellnet, job pods fall back to PID-based parent monitoring. PID 1 always exists in a container so job pods never detect parent failure — runaway pods will persist indefinitely. This is a real gap that needs to be in launch_job(): {"name": "NVFLARE_PARENT_MONITOR", "value": "cellnet"}
> 
I am not sure if this is the deployment scenario because the words don't make sense.


> H5 — Pod securityContext: ❌ Still missing No securityContext on pod spec or container spec. Running as root in production K8s is blocked by most Pod Security Standards. Low risk for dev/POC but a blocker for enterprise clusters.
We can tighten the security.  However, does this mean most of the time, nvflare should follow what the IT security policies?

> 
> M4 — module_args copy when caller provides dict: ❌ Partial When module_args is None, the code correctly does DEFAULT_CONTAINER_ARGS_MODULE_ARGS_DICT.copy(). But when the caller passes a dict, it's self.container_args_module_args_dict = job_config.get("module_args") — still a reference, not a copy. Caller mutation would corrupt the handle's args.
Will add .copy()

> C6 — namespace default still "default": minor namespace="default" remains the default in both K8sJobHandle and K8sJobLauncher. It's now a configurable param so this is operator error if misconfigured, not a code bug. Acceptable for now.

nvflare/app_common/resource_consumers/be_resource_consumer.py
```diff
@@ -0,0 +1,21 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from nvflare.apis.resource_manager_spec import ResourceConsumerSpec
+
+
+class BEResourceConsumer(ResourceConsumerSpec):
```
chesterxgchen, line 18: where the file name starts with BE_ , this seems not follow python file naming convention

nvflare/app_common/resource_consumers/be_resource_consumer.py
```diff
@@ -0,0 +1,21 @@
...
+
```
chesterxgchen, line 25: What is BEResourceConsumer ? 

should you add doc string to explain what this is and how to use this ?

nvflare/app_common/resource_managers/BE_resource_manager.py
```diff
@@ -0,0 +1,50 @@
...
+# limitations under the License.
+
+import uuid
+
+from nvflare.apis.fl_component import FLComponent
+from nvflare.apis.fl_context import FLContext
+from nvflare.apis.resource_manager_spec import ResourceManagerSpec
+
+
+class BEResourceManager(ResourceManagerSpec, FLComponent):
```
chesterxgchen: again, why file name starts with BE_, should it be "be_"

nvflare/app_common/resource_consumers/be_resource_consumer.py
chesterxgchen, line 32: oh, BE stands best effort. at you at least should exaplain the acroname.

nvflare/app_common/resource_managers/BE_resource_manager.py
```diff
@@ -0,0 +1,50 @@
...
+        at runtime.
+
+        """
+        super().__init__()
+
+    def check_resources(self, resource_requirement: dict, fl_ctx: FLContext):
+        if not isinstance(resource_requirement, dict):
+            raise TypeError(f"resource_requirement should be of type dict, but got {type(resource_requirement)}.")
+
+        token = str(uuid.uuid4())
```
chesterxgchen: check resources always return new random uuid/token and True,  should we add doc string explain what we are returning, what the UUID stands for. what token, and whats True means ?

nvflare/app_opt/job_launcher/k8s_launcher.py
```diff
@@ -36,12 +40,20 @@ class JobState(Enum):
     UNKNOWN = "unknown"
 
 
+class POD_Phase(Enum):
```
chesterxgchen: class name shouldn't have "_" in between

nvflare/app_opt/job_launcher/k8s_launcher.py
```diff
@@ -68,141 +127,184 @@ def __init__(self, job_id: str, api_instance: core_v1_api, job_config: dict, nam
...
-        )
-
+            self.container_args_module_args_dict_as_list.append(str(v))
+        self.volume_list.extend(job_config.get("volume_list", []))
         self.pod_manifest["metadata"]["name"] = job_config.get("name")
         self.pod_manifest["spec"]["containers"] = self.container_list
         self.pod_manifest["spec"]["volumes"] = self.volume_list
 
-        self.container_list[0]["image"] = job_config.get("image", "nvflare/nvflare:2.5.0")
+        image = job_config.get("image")
```
chesterxgchen, line 177: job config image is the way to specify docker image ?

chesterxgchen (commented):
add few comments

nvflare/app_common/resource_consumers/be_resource_consumer.py
IsaacYangSLA: Thanks.  Will rename it.

nvflare/app_common/resource_consumers/be_resource_consumer.py
IsaacYangSLA: Let me add it.

nvflare/app_common/resource_managers/BE_resource_manager.py
IsaacYangSLA: Same.  Will rename it.

nvflare/app_common/resource_consumers/be_resource_consumer.py
IsaacYangSLA: The docstring, in the new commit will explain it.

nvflare/app_opt/job_launcher/k8s_launcher.py
IsaacYangSLA: Will rename it.

nvflare/app_opt/job_launcher/k8s_launcher.py
IsaacYangSLA: Yes.  The job knows which image contains all the dependencies it needs.

IsaacYangSLA:
> add few comments

I think all comments were addressed in the new commit.

nvflare/app_opt/job_launcher/k8s_launcher.py
```diff
@@ -52,13 +64,60 @@ class JobState(Enum):
...
+    "-n": None,
+    "-c": None,
+    "-p": None,
+    "-g": None,
+    "-scheme": None,
+    "-s": None,
+}
+
+
+class PV_NAME(Enum):
```
chesterxgchen: Python class name convention "_" is avoided

IsaacYangSLA: Thanks.  Will rename it and update all related files.

chesterxgchen (approved):

docs/design/JobLauncher_and_JobHandle.md
```diff
@@ -0,0 +1,581 @@
...
+
+Pod phase mapping:
+
+| Pod Phase | JobState | JobReturnCode |
+|-----------|----------|---------------|
+| `Pending` | `STARTING` | `UNKNOWN` |
+| `Running` | `RUNNING` | `UNKNOWN` |
```
YuanTingHsieh, line 313: regarding the mapping, why is Pending and Running maps to this ReturnCode Unknown?
is it because NVFlare does not have something like "Running"?

docs/design/JobLauncher_and_JobHandle.md
```diff
@@ -0,0 +1,581 @@
...
+my-data-pvc: /var/tmp/nvflare/data
+```
+
+> Note: Currently only the PVC name (the YAML key) is used. The mount path value is ignored — the data volume is always mounted at the hardcoded path `/var/tmp/nvflare/data`.
+
+---
+
+## 10. Future Improvements
+
+1. **Explicit launcher selection** -- Today "has image" → K8s, "no image" → Process. Allow an explicit `launcher_type` field in job meta or deploy map so a site can support multiple backends or provide fallback ordering.
```
YuanTingHsieh, line 571: right, we might need this for Docker-type, but in another PR.

docs/design/JobLauncher_and_JobHandle.md
```diff
@@ -0,0 +1,581 @@
...
+```
+
+### 9.2 Kubernetes Launcher
+
+```json
+{
+  "id": "job_launcher",
+  "path": "nvflare.app_opt.job_launcher.k8s_launcher.ClientK8sJobLauncher",
+  "args": {
+    "config_file_path": "/path/to/kubeconfig",
```
YuanTingHsieh, line 549: Is this path refer to the path inside PVC?

Do we generate this config in provision and copy it inside the PVC?
Or it is generated on the fly?

YuanTingHsieh: ok after talked, this path is inside PVC.
each site admin will already use something like kubectl to generate this config file and put it inside PVC.

PR #4340, opened 2026-03-20
Add helm chart (client only)

IsaacYangSLA:
The design doc, implementation in provision and tests.

### Types of changes

- [x] Documentation updated.


docs/design/helm_chart_client_design.md
```diff
@@ -0,0 +1,709 @@
+# NVFlare Client Helm Chart — Design Document
```
pcnudde, line 1: I think this file is too large to be checked in. I would split it in 2. 
- a helm_chart_client_design.md file with the pure design and tradeoffs made. (which is checked in)
- a helm_chart_client_implementation.md that is not checked in and that is used by your agent to keep track of what it is doing.

nvflare/lighter/impl/client_helm_chart.py
```diff
@@ -0,0 +1,305 @@
...
+{{- define "nvflare-client.image" -}}
+{{- if .Values.image.tag }}
+{{- printf "%s:%s" .Values.image.repository .Values.image.tag }}
+{{- else }}
+{{- .Values.image.repository }}
+{{- end }}
+{{- end }}
+"""
+
+_CLIENT_POD_YAML = """\
```
pcnudde, line 56: Would it not be better to have explicit files in the folder rather then inline strings.

chesterxgchen: yes, prefer to the have files rather than inline key value string

chesterxgchen (commented):
add few comments

IsaacYangSLA:
Will close this and create another PR to support helm charts of both server and client.

PR #4343, opened 2026-03-21
fix(dashboard): harden mass-assignment 

pcnudde:
## Summary

- Replace unrestricted `setattr` mass-assignment with field allowlists in all patch methods (user, client, project), blocking `role_id`/`password_hash`/`root_key` writes
- Block denied users (`approval_state < 0`) at login
- Add `approved` claim to JWT; gate data-listing endpoints on approval so pending users cannot enumerate users/clients
- Require authentication on `GET /project` and `GET /organizations`
- Remove creator-side client org reassignment
- Harden `serialize_pri_key`: raise `TypeError` on non-bytes passphrase instead of silently falling through to `NoEncryption()`
- Add 7 security regression tests (6 of 7 fail against pre-fix code, all pass post-fix)

## Known limitation

Stateless JWTs have a 30-minute lifetime. If a user is denied after token issuance, their existing token remains valid until expiry. The `approved` claim + endpoint gating covers the normal case; the login check covers the denied case for new tokens.

### DIFF
--- a/nvflare/dashboard/application/project.py
+++ b/nvflare/dashboard/application/project.py
@@ -91,7 +91,11 @@ def login():
     password = req.get("password", None)
     user = Store.verify_user(email, password)
     if user:
-        additional_claims = {"role": user.role.name, "organization": user.organization.name}
+        additional_claims = {
+            "role": user.role.name,
+            "organization": user.organization.name,
+            "approved": user.approval_state >= 100,
+        }
         access_token = create_access_token(identity=user.email, additional_claims=additional_claims)
         return jsonify(
             {
@@ -141,10 +145,12 @@ def set_project():
 
 
 @app.route(FLARE_DASHBOARD_NAMESPACE + "/api/v1/project", methods=["GET"])
+@jwt_required()
 def get_project():
     return jsonify(Store.get_project())
 
 
 @app.route(FLARE_DASHBOARD_NAMESPACE + "/api/v1/organizations", methods=["GET"])
+@jwt_required()
 def get_orgs():
     return jsonify(Store.get_orgs())
--- a/nvflare/dashboard/application/store.py
+++ b/nvflare/dashboard/application/store.py
@@ -106,6 +106,21 @@ class Store(object):
         project_dict["num_users"] = User.query.count()
         return project_dict
 
+    _PROJECT_WRITABLE = {
+        "title",
+        "description",
+        "app_location",
+        "ha_mode",
+        "starting_date",
+        "end_date",
+        "overseer",
+        "server1",
+        "server2",
+        "frozen",
+        "public",
+        "cc_mode",
+    }
+
     @classmethod
     def set_project(cls, req):
         project = Project.query.first()
@@ -128,7 +143,8 @@ class Store(object):
             project.server_props = json.dumps(server_props)
 
         for k, v in req.items():
-            setattr(project, k, v)
+            if k in cls._PROJECT_WRITABLE:
+                setattr(project, k, v)
 
         db.session.add(project)
         db.session.commit()
@@ -212,6 +228,8 @@ class Store(object):
         client = Client.query.get(id)
         return add_ok({"client": _dict_or_empty(client)})
 
+    _CLIENT_ADMIN_WRITABLE = {"name", "description", "approval_state"}
+
     @classmethod
     def patch_client_by_project_admin(cls, id, req):
         client = Client.query.get(id)
@@ -230,7 +248,8 @@ class Store(object):
             client.props = json.dumps(props)
 
         for k, v in req.items():
-            setattr(client, k, v)
+            if k in cls._CLIENT_ADMIN_WRITABLE:
+                setattr(client, k, v)
 
         try:
             db.session.add(client)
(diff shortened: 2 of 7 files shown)


pcnudde:
Addressing greptile feedback:

**Missing assert (security_test.py):** Fixed — added `assert resp.status_code == 200` before extracting `user_id` in `test_denied_user_cannot_login`.

**GET /project and /organizations require auth:** Reverted — the landing page calls `GET /project` before login to display the project title. The registration form uses a free-text org field, not a dropdown from `GET /organizations`. Neither endpoint exposes sensitive data, so removed `@jwt_required` from both.

nvflare/dashboard/application/project.py
```diff
@@ -141,10 +145,12 @@ def set_project():
...
 @app.route(FLARE_DASHBOARD_NAMESPACE + "/api/v1/project", methods=["GET"])
+@jwt_required()
 def get_project():
     return jsonify(Store.get_project())
 
 
 @app.route(FLARE_DASHBOARD_NAMESPACE + "/api/v1/organizations", methods=["GET"])
+@jwt_required()
 def get_orgs():
     return jsonify(Store.get_orgs())
```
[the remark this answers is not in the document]
pcnudde: fixed

nvflare/dashboard/application/store.py
```diff
@@ -212,6 +228,8 @@ def get_client(cls, id):
         client = Client.query.get(id)
         return add_ok({"client": _dict_or_empty(client)})
 
+    _CLIENT_ADMIN_WRITABLE = {"name", "description", "approval_state"}
```
chesterxgchen: this seems to be repeated multiple times

nvflare/dashboard/application/store.py
```diff
@@ -230,7 +248,8 @@ def patch_client_by_project_admin(cls, id, req):
             client.props = json.dumps(props)
 
         for k, v in req.items():
-            setattr(client, k, v)
+            if k in cls._CLIENT_ADMIN_WRITABLE:
```
chesterxgchen: if this is just org and name, we do we need to this setattr() function, why not just client.orgnation = orgnation client.name = name.

nvflare/dashboard/application/store.py
```diff
@@ -240,15 +259,13 @@ def patch_client_by_project_admin(cls, id, req):
...
     @classmethod
     def patch_client_by_creator(cls, id, req):
         client = Client.query.get(id)
         _ = req.pop("approval_state", None)
-
-        organization = req.pop("organization", None)
-        if organization is not None:
-            org = get_or_create(db.session, Organization, name=organization)
-            client.organization_id = org.id
+        _ = req.pop("organization", None)
```
chesterxgchen, line 264: we skip the org creation ? why ?

chesterxgchen (commented):
add a few question, some of the comments about the old design.

nvflare/dashboard/application/store.py
pcnudde: This is in the patch function. This call allows the user to change orgs. This functionality is not used in the UI, and it doesn't seem we should allow it in the API either. The admin can still change the org of the user.

nvflare/dashboard/application/store.py
pcnudde: it would be 2 lines like : "if name in req.items(): client.name = req[name]"
Keeping the same pattern for all 3 functions seems easier

nvflare/dashboard/application/store.py
pcnudde: each one is different though. _CLIENT_ADMIN_WRITABLE, _CLIENT_CREATOR_WRITABLE and _PROJECT_WRITABLE

pcnudde: but this code can be simplified. we don't really need this constant here.

tests/unit_test/dashboard/security_test.py
```diff
@@ -0,0 +1,113 @@
...
+
+        resp = client.post(NS + "/api/v1/login", json={"email": "escalate@test.com", "password": "p"})
+        assert resp.json["user"]["role"] != "project_admin"
+
+    def test_password_hash_overwrite(self, client):
+        """PATCH password_hash directly should be ignored; original password still works."""
+        client.post(NS + "/api/v1/users", json={"email": "hashtest@test.com", "password": "original", "name": "x"})
+        resp = client.post(NS + "/api/v1/login", json={"email": "hashtest@test.com", "password": "original"})
+        token = resp.json["access_token"]
+        user_id = resp.json["user"]["id"]
```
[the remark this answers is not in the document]
pcnudde: Fixed — added missing asserts in both tests.

tests/unit_test/dashboard/security_test.py
```diff
@@ -0,0 +1,115 @@
...
+    def test_pending_user_cannot_list_users(self, client):
+        """Unapproved user (approval_state=0) cannot list all users."""
+        client.post(NS + "/api/v1/users", json={"email": "pending@test.com", "password": "p", "name": "x"})
+        resp = client.post(NS + "/api/v1/login", json={"email": "pending@test.com", "password": "p"})
+        if resp.status_code != 200:
+            pytest.skip("Login blocked for pending users")
+        token = resp.json["access_token"]
+
+        resp = client.get(NS + "/api/v1/users", headers={"Authorization": f"Bearer {token}"})
+        assert resp.status_code == 403
```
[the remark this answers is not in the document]
pcnudde: incorrect. @jwt_required() was removed for GET /project and GET /organizations

nvflare/dashboard/application/store.py
pcnudde: I simplified the code a bit.

chesterxgchen (approved):

nvflare/dashboard/application/project.py
```diff
@@ -141,10 +145,12 @@ def set_project():
 
 
 @app.route(FLARE_DASHBOARD_NAMESPACE + "/api/v1/project", methods=["GET"])
+@jwt_required()
```
IsaacYangSLA, line 148: This PR was merged.  But this will break the regular project page, which is shown at the sign-up page, as the JWT does not exist yet.

PR #4393, opened 2026-04-02
Inject PYTHONPATH env to k8s job launcher

IsaacYangSLA:
### Description

The custom folder inside jobs should be included by the env var PYTHONPATH so jobs can import custom codes.  This PR adds that env var into job pod manifest.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/app_opt/job_launcher/k8s_launcher.py
+++ b/nvflare/app_opt/job_launcher/k8s_launcher.py
@@ -187,6 +187,9 @@ class K8sJobHandle(JobHandleSpec):
         self.container_list[0]["volumeMounts"] = self.container_volume_mount_list
         if job_config.get("resources", {}).get("limits", {}).get("nvidia.com/gpu"):
             self.container_list[0]["resources"] = job_config.get("resources")
+        app_custom_folder = job_config.get("env", {}).get("PYTHONPATH")
+        if app_custom_folder:
+            self.container_list[0]["env"] = [{"name": "PYTHONPATH", "value": str(app_custom_folder)}]
 
     def get_manifest(self):
         return copy.deepcopy(self.pod_manifest)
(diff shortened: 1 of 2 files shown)


nvflare/app_opt/job_launcher/k8s_launcher.py
```diff
@@ -187,6 +187,9 @@ def _make_manifest(self, job_config):
         self.container_list[0]["volumeMounts"] = self.container_volume_mount_list
         if job_config.get("resources", {}).get("limits", {}).get("nvidia.com/gpu"):
             self.container_list[0]["resources"] = job_config.get("resources")
+        app_custom_folder = job_config.get("env", {}).get("PYTHONPATH")
```
YuanTingHsieh, line 190: The code sets PYTHONPATH unconditionally. If the base container image or other config already sets PYTHONPATH, this will silently replace it rather than prepend. In practice this is probably fine for NVFlare's K8s setup, but it's a subtle footgun if the image has its own PYTHONPATH.

nvflare/app_opt/job_launcher/k8s_launcher.py
```diff
@@ -187,6 +187,9 @@ def _make_manifest(self, job_config):
...
+        if app_custom_folder:
```
YuanTingHsieh, line 191: on another note not related to this PR, we seems to have a gap in BYOC enforcement, certain way we can avoid the RBAC

nvflare/app_opt/job_launcher/k8s_launcher.py
IsaacYangSLA: This is a confusion from greptile, claude and k8s document.  In pod spec, the env var will be `set`, not `added.`   It's because the actual env var is not available until the container starts and pod spec couldn't retrieve that runtime env var.
One workaround is to use /bin/sh -c 'PYTHOPATH=...:$PYTHONPATH /usr/local/bin/python -m ` as the command in pod spec.  But it's not clean.

I believe container image builders should not set PYTHONPATH as this env var is used for development to include source codes.  An image should have python packages installed and there is no need to use PYTHONPATH.

nvflare/app_opt/job_launcher/k8s_launcher.py
IsaacYangSLA: This behavior is the same as current processor launcher. 
When a job is submitted, the validator checks if any app in the zip contains a custom/ folder. If so, it sets
  meta[AppValidationKey.BYOC] = True and the app_authz verifies if such job should be blocked.

Therefore, it's not job launcher's responsibility to determine the authorization.

nvflare/app_opt/job_launcher/k8s_launcher.py
YuanTingHsieh: got it, sounds good.
Then we can just explain in doc if needed that **image builders should not set PYTHONPATH**

nvflare/app_opt/job_launcher/k8s_launcher.py
YuanTingHsieh: yes, this PR is not changing that, pointing out we might need to have another PR to address that.

YuanTingHsieh (approved):

PR #4425, opened 2026-04-10
Fix/misc improvements 2026 q2

GeorgeWang-nv:
Fixes # .

### Description

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/private/fed/server/client_manager.py
+++ b/nvflare/private/fed/server/client_manager.py
@@ -262,25 +262,13 @@ class ClientManager:
                 self.logger.debug(f"Receive heartbeat from Client:{token}")
                 return False
             else:
-                for _token, _client in self.clients.items():
-                    if _client.name == client_name:
-                        fl_ctx.set_prop(
-                            FLContextKey.COMMUNICATION_ERROR,
-                            "Client ID already registered as a client: {}".format(client_name),
-                            sticky=False,
-                        )
-                        self.logger.info(
-                            f"Failed to re-activate the client:{client_name} with token: {token}. "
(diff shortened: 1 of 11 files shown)


nvflare/private/fed/server/client_manager.py
```diff
@@ -262,25 +262,13 @@ def heartbeat(self, token, client_name, client_fqcn, fl_ctx: FLContext):
...
-                        )
-                        return False
-
-                client = Client(client_name, token)
-                self._set_client_props(client, client_fqcn, fl_ctx)
-                self.clients.update({token: client})
-                self.name_to_clients[client.name] = client
-                self.logger.info(f"Re-activate the client: {client_name} at {client_fqcn} with token: {token}")
-                return True
+                # Reject heartbeat from unknown or revoked tokens.
```
nvidianz, line 265: Client re-activation is a feature. Not sure we should remove it like this.

YuanTingHsieh:
@GeorgeWang-nv thanks for the PR!
is it possible you break this one into smaller ones (for each commit)?
and add the pr description for each one so it is clear

nvflare/private/fed/server/client_manager.py
GeorgeWang-nv: Agreed, the re-activation approach needs more investigation. I'll split into individual PRs so we can discuss each one separately.

GeorgeWang-nv:
Thanks Yuanting! Will do — I'll close this PR and create individual PRs per issue with descriptions.

GeorgeWang-nv:
Split this PR into several separate PRs.

PR #4427, opened 2026-04-11
Add helm chart 

IsaacYangSLA:
### Description

Helm chart builders for both client and server.
Design doc for this builder
Unit tests added for this builder

### Types of changes

- [x] Documentation updated.


pcnudde:
### Bug: hostPort always rendered — breaks GKE Autopilot

**File:** `nvflare/lighter/templates/helm/server/deployment.yaml`

`hostPort` is unconditionally set on container ports. GKE Autopilot rejects this:

> GKE Warden rejected the request: container gcp-server specifies host ports [8002 8003], which are disallowed in Autopilot

Autopilot manages node scheduling and security policies — host ports bypass the k8s network model and bind directly to the node, which Autopilot forbids. Standard GKE, EKS, and AKS allow host ports (though it's bad practice). MicroK8s uses host ports as its default exposure model, which is why the original templates were written this way.

**Fix:** The values already have `hostPort.enabled: true` but the template ignores it. Make `hostPort` conditional:

```yaml
- containerPort: {{ .Values.fedLearnPort }}
  {{- if ne (include "nvflare-server.hostPortEnabled" .) "false" }}
  hostPort: {{ .Values.fedLearnPort }}
  {{- end }}
```

pcnudde:
### Bug: tcp-services ConfigMap targets nonexistent `ingress` namespace

**File:** `nvflare/lighter/templates/helm/server/tcp-services.yaml`

Unconditionally creates a ConfigMap in `namespace: ingress`. GKE Autopilot (and most managed k8s) doesn't have an `ingress` namespace, so Helm fails:

> namespaces "ingress" not found

This ConfigMap is for the MicroK8s nginx-ingress controller, which reads TCP passthrough rules from this specific ConfigMap. No other k8s distribution uses this pattern.

**Fix:** Wrap in a conditional:

```yaml
{{- if ne (include "nvflare-server.tcpConfigMapEnabled" .) "false" }}
...ConfigMap...
{{- end }}
```

pcnudde:
### Bug: Service has no LoadBalancer support

**File:** `nvflare/lighter/templates/helm/server/service.yaml`

The service template has no `type` or `loadBalancerIP` field — it renders as a plain ClusterIP service. There's no way to expose the server externally via `--set` overrides.

Without host ports (which Autopilot forbids), the only way to expose the FL server to external clients is via a LoadBalancer Service with a static IP. MicroK8s doesn't need this because it uses host ports directly.

**Fix:** Add optional `service.type` and `service.loadBalancerIP`:

```yaml
spec:
  {{- if .Values.service.type }}
  type: {{ .Values.service.type }}
  {{- end }}
  {{- if .Values.service.loadBalancerIP }}
  loadBalancerIP: {{ .Values.service.loadBalancerIP }}
  {{- end }}
```

pcnudde:
### Bug: Client uses bare Pod, not Deployment

**File:** `nvflare/lighter/templates/helm/client/pod.yaml`

The client is a bare `Pod` with `restartPolicy: Never`. If the client process crashes, the pod stays in Failed state permanently and requires manual intervention.

This affects all k8s platforms, not just GKE, but is more painful in cloud environments where SSH access to nodes is restricted (Autopilot) or unavailable.

**Fix:** Replace `pod.yaml` with a `deployment.yaml` using the same container spec wrapped in a Deployment with `replicas: 1`. This gives automatic restart on crash.

pcnudde:
### Bug: Provisioned resources.json defaults to ProcessJobLauncher

**File:** generated `local/resources.json.default` (via `nvflare/lighter/templates/master_template.yml`)

When `HelmChartBuilder` is in the builders list, the provisioned `resources.json` still uses `ClientProcessJobLauncher` / `ServerProcessJobLauncher` (in-process subprocess). On Kubernetes, jobs should run as separate pods via `K8sJobLauncher` so they get independent scheduling, resource limits, and access to the shared workspace PVC.

`ProcessJobLauncher` runs jobs inside the server/client pod — defeating k8s resource management and the shared workspace design.

**Fix:** `HelmChartBuilder` should override the job launcher in `resources.json` to `ServerK8sJobLauncher` / `ClientK8sJobLauncher`.

Additionally, `K8sJobLauncher` has three missing defaults that make it hard to use out of the box:

- **kubeconfig**: requires an explicit file path. Should support `config_file_path: null` to mean "use in-cluster config" (`kubernetes.config.load_incluster_config()`), eliminating the need to stage a kubeconfig file in `nvfletc`.
- **study_data_pvc**: requires a mapping file with no default. Should default to a sensible PVC name when no file is provided.
- **namespace**: defaults to `"default"` instead of the pod's own namespace. Should read from the downward API or default to the release namespace.

The Helm chart also sets `automountServiceAccountToken: false`, which prevents in-cluster auth. It should be `true` for the main NVFlare ServiceAccount (with appropriate RBAC for creating job pods).

pcnudde:
### K8sJobLauncher ignores jobs without explicit image — no fallback

**File:** `nvflare/app_opt/job_launcher/k8s_launcher.py`, line ~410

`K8sJobLauncher.handle_event()` only registers itself when `extract_job_image(job_meta, site_name)` returns a non-empty image:

```python
    if event_type == EventType.BEFORE_JOB_LAUNCH:
        job_meta = fl_ctx.get_prop(FLContextKey.JOB_META)
        job_image = extract_job_image(job_meta, fl_ctx.get_identity_name())
        if job_image:
            add_launcher(self, fl_ctx)
```

Jobs submitted without a `job_image` in their `meta.json` (which is the common case — e.g. `hello-numpy` and most standard jobs) are silently skipped. If `ProcessJobLauncher` is not also configured, the job fails with:

> There's no job launcher can handle this job

On Kubernetes, `ProcessJobLauncher` should not be needed as a backup. All jobs should run as k8s Job pods for proper resource isolation and scheduling. Requiring every job to specify an image in `meta.json` is an unnecessary burden on job authors — the image should come from the deployment configuration, not the job.

**Suggested fix:** `K8sJobLauncher` should accept a `default_image` constructor argument. When a job has no explicit image, use the default. The launcher should always register itself when a default image is configured:

```python
def __init__(self, ..., default_image: str = ""):
    self.default_image = default_image

    if event_type == EventType.BEFORE_JOB_LAUNCH:
        job_meta = fl_ctx.get_prop(FLContextKey.JOB_META)
        job_image = extract_job_image(job_meta, fl_ctx.get_identity_name())
        if job_image or self.default_image:
            add_launcher(self, fl_ctx)
```

And in `_launch_job`, fall back to `self.default_image` when `extract_job_image` returns `None`.

This way `HelmChartBuilder` can set `default_image` to the same image used for the server/client containers, and all jobs run as k8s pods without requiring per-job image configuration.

IsaacYangSLA:
Thanks for testing this PR in GCP.  Some of the comments are based on design decisions during meetings.  Others will be addressed by additional commits.

IsaacYangSLA:
> ### Bug: Client uses bare Pod, not Deployment
> **File:** `nvflare/lighter/templates/helm/client/pod.yaml`
> 
> The client is a bare `Pod` with `restartPolicy: Never`. If the client process crashes, the pod stays in Failed state permanently and requires manual intervention.
> 
> This affects all k8s platforms, not just GKE, but is more painful in cloud environments where SSH access to nodes is restricted (Autopilot) or unavailable.
> 
> **Fix:** Replace `pod.yaml` with a `deployment.yaml` using the same container spec wrapped in a Deployment with `replicas: 1`. This gives automatic restart on crash.

This is by design.  The parent pod should not restart because when admin sends `shutdown` command, the pod should exit.

IsaacYangSLA:
> ### Bug: Provisioned resources.json defaults to ProcessJobLauncher
> **File:** generated `local/resources.json.default` (via `nvflare/lighter/templates/master_template.yml`)
> 
> When `HelmChartBuilder` is in the builders list, the provisioned `resources.json` still uses `ClientProcessJobLauncher` / `ServerProcessJobLauncher` (in-process subprocess). On Kubernetes, jobs should run as separate pods via `K8sJobLauncher` so they get independent scheduling, resource limits, and access to the shared workspace PVC.
> 
> `ProcessJobLauncher` runs jobs inside the server/client pod — defeating k8s resource management and the shared workspace design.
> 
> **Fix:** `HelmChartBuilder` should override the job launcher in `resources.json` to `ServerK8sJobLauncher` / `ClientK8sJobLauncher`.
> 
> Additionally, `K8sJobLauncher` has three missing defaults that make it hard to use out of the box:
> 
> * **kubeconfig**: requires an explicit file path. Should support `config_file_path: null` to mean "use in-cluster config" (`kubernetes.config.load_incluster_config()`), eliminating the need to stage a kubeconfig file in `nvfletc`.
> * **study_data_pvc**: requires a mapping file with no default. Should default to a sensible PVC name when no file is provided.
> * **namespace**: defaults to `"default"` instead of the pod's own namespace. Should read from the downward API or default to the release namespace.
> 
> The Helm chart also sets `automountServiceAccountToken: false`, which prevents in-cluster auth. It should be `true` for the main NVFlare ServiceAccount (with appropriate RBAC for creating job pods).

This is the design decision made during the meetings.  This file is used by all possible deployment environments.  Therefore, local IT should edit it before launching in K8s.

IsaacYangSLA:
> ### K8sJobLauncher ignores jobs without explicit image — no fallback
> **File:** `nvflare/app_opt/job_launcher/k8s_launcher.py`, line ~410
> 
> `K8sJobLauncher.handle_event()` only registers itself when `extract_job_image(job_meta, site_name)` returns a non-empty image:
> 
> ```python
> def handle_event(self, event_type: str, fl_ctx: FLContext):
>     if event_type == EventType.BEFORE_JOB_LAUNCH:
>         job_meta = fl_ctx.get_prop(FLContextKey.JOB_META)
>         job_image = extract_job_image(job_meta, fl_ctx.get_identity_name())
>         if job_image:
>             add_launcher(self, fl_ctx)
> ```
> 
> Jobs submitted without a `job_image` in their `meta.json` (which is the common case — e.g. `hello-numpy` and most standard jobs) are silently skipped. If `ProcessJobLauncher` is not also configured, the job fails with:
> 
> > There's no job launcher can handle this job
> 
> On Kubernetes, `ProcessJobLauncher` should not be needed as a backup. All jobs should run as k8s Job pods for proper resource isolation and scheduling. Requiring every job to specify an image in `meta.json` is an unnecessary burden on job authors — the image should come from the deployment configuration, not the job.
> 
> **Suggested fix:** `K8sJobLauncher` should accept a `default_image` constructor argument. When a job has no explicit image, use the default. The launcher should always register itself when a default image is configured:
> 
> ```python
> def __init__(self, ..., default_image: str = ""):
>     self.default_image = default_image
> 
> def handle_event(self, event_type: str, fl_ctx: FLContext):
>     if event_type == EventType.BEFORE_JOB_LAUNCH:
>         job_meta = fl_ctx.get_prop(FLContextKey.JOB_META)
>         job_image = extract_job_image(job_meta, fl_ctx.get_identity_name())
>         if job_image or self.default_image:
>             add_launcher(self, fl_ctx)
> ```
> 
> And in `_launch_job`, fall back to `self.default_image` when `extract_job_image` returns `None`.
> 
> This way `HelmChartBuilder` can set `default_image` to the same image used for the server/client containers, and all jobs run as k8s pods without requiring per-job image configuration.

This is also the design and decision made during meetings.  There is no default image and images are also used to enable k8s job launcher during events.

nvflare/lighter/impl/helm_chart.py
```diff
@@ -13,112 +13,292 @@
...
+        server = project.get_server()
+        if not server:
+            return
+
+        chart_dir = os.path.join(ctx.get_ws_dir(server), ProvFileName.HELM_CHART)
+        templates_dir = os.path.join(chart_dir, "templates")
+        os.makedirs(templates_dir, exist_ok=True)
+
+        fed_learn_port = ctx.get(CtxKey.FED_LEARN_PORT, 8002)
+        admin_port = ctx.get(CtxKey.ADMIN_PORT, 8003)
```
chesterxgchen, line 115: I thought the current behavior is if only one port is specified, the default admin port will use the same port as the fed_learn port. not default to different port. 

L115, the 8003 should be 8002

nvflare/lighter/impl/helm_chart.py
```diff
@@ -13,112 +13,292 @@
...
-        self.service_overseer = ctx.yaml_load_template_section(TemplateSectionKey.HELM_CHART_SERVICE_OVERSEER)
-        self.service_server = ctx.yaml_load_template_section(TemplateSectionKey.HELM_CHART_SERVICE_SERVER)
+    def _write_server_chart_yaml(self, chart_dir: str, server: Participant):
+        _, tag = _split_image(self.docker_image)
+        chart = {
+            "apiVersion": "v2",
+            "name": "nvflare-server",
+            "description": f"NVFlare federated learning server for {server.name}",
+            "type": "application",
+            "version": "0.1.0",
```
chesterxgchen, line 141: whats this version, should this be a constant that be changed ?

nvflare/lighter/impl/helm_chart.py
```diff
@@ -13,112 +13,292 @@
...
+        self._write_client_template_files(templates_dir)
+
+    def _write_client_chart_yaml(self, chart_dir: str, client: Participant):
+        _, tag = _split_image(self.docker_image)
+        chart = {
+            "apiVersion": "v2",
+            "name": "nvflare-client",
+            "description": f"NVFlare federated learning client pod and service for {client.name}",
+            "type": "application",
+            "version": "0.1.0",
```
chesterxgchen, line 248: version again

chesterxgchen (approved):
LGTM, but make a few minor comments, that can be addressed later

nvflare/lighter/impl/helm_chart.py
IsaacYangSLA: The ctx (ProvisionContext) init'ed with the following

>      server = project.get_server()
>      fed_learn_port = server.get_prop(PropKey.FED_LEARN_PORT, 8002)
>      admin_port = server.get_prop(PropKey.ADMIN_PORT, fed_learn_port)
>      self[CtxKey.ADMIN_PORT] = admin_port
>      self[CtxKey.FED_LEARN_PORT] = fed_learn_port
 
So if project.yml does not specify admin port, it will be fed_learn_port.

That ctx.get will get the correct information.

nvflare/lighter/impl/helm_chart.py
IsaacYangSLA: This is the concept of helm chart.  It should be changed by the local IT if he makes any changes to the helm chart.  This is the default version.

nvflare/lighter/impl/helm_chart.py
IsaacYangSLA: Same as the above.

chesterxgchen: my question should be a constant so that all helm_chart version is consistent.

nvflare/lighter/impl/helm_chart.py
```diff
@@ -13,112 +13,292 @@
...
 
+_HELM_TEMPLATES_DIR = os.path.join(os.path.dirname(prov.__file__), "templates", "helm")
+
+
+def _split_image(docker_image: str):
+    """Split ``'repo:tag'`` into ``(repo, tag)``.  Returns ``('repo', '')`` when no tag."""
+    if ":" in docker_image:
+        repo, tag = docker_image.rsplit(":", 1)
+        return repo, tag
+    return docker_image, ""
```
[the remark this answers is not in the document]
chesterxgchen: This is seems to be a reasonable issue

PR #4450, opened 2026-04-15
Add multicloud Kubernetes deployment tooling

pcnudde:
## Summary

This PR adds the first multicloud Kubernetes deployment path for NVFlare.

Changes include:
- GKE, EKS, and AKS cluster helper scripts and setup docs
- a Python-based multicloud deploy flow for:
  - GCP server
  - GCP client
  - AWS client
- managed-k8s updates to the Helm chart and K8s job launcher
- unit tests for the new job launcher/image-selection behavior
- cloud setup docs aligned with what the bootstrap scripts now do automatically

## Details

Highlights:
- saves per-cloud kubeconfigs under `.tmp/kubeconfigs/`
- bootstraps GKE Filestore storage for RWX workspace PVCs
- bootstraps EKS EBS/EFS storage for RWX workspace PVCs
- adds ServiceAccount/RBAC support to the generated Helm charts
- adds security context support needed for AWS/Bottlerocket
- supports `@ALL` image selection in job deploy maps
- improves cert SAN handling for IP-based hosts


docker/Dockerfile
```diff
@@ -11,5 +11,4 @@ COPY README.md /opt/NVFlare/README.md
 COPY setup.cfg /opt/NVFlare/setup.cfg
 WORKDIR /opt/NVFlare
 RUN pip install -U pip
-RUN pip install -e .
-
+RUN pip install -e . && pip install kubernetes
```
IsaacYangSLA: Why add kubernetes as default?

nvflare/app_opt/job_launcher/k8s_launcher.py
```diff
@@ -282,8 +286,10 @@ def __init__(
         study_data_pvc_file_path: str,
         timeout=None,
         namespace="default",
-        pending_timeout=30,
+        pending_timeout=120,
         python_path="/usr/local/bin/python",
+        data_read_only: bool = True,
```
IsaacYangSLA: Once data_read_only is set, all data pvc will be mounted as read-only.  Why not make it configurable for each study or each data pvc?

nvflare/app_opt/job_launcher/k8s_launcher.py
```diff
@@ -294,6 +300,8 @@ def __init__(
         self.namespace = namespace
         self.pending_timeout = pending_timeout
         self.python_path = python_path
+        self.data_read_only = data_read_only
+        self.security_context = security_context
```
IsaacYangSLA, line 305: This will be a dictionary, set by local IT.  All job pods will use the same security context.  I think it should be fine.

nvflare/utils/job_launcher_utils.py
```diff
@@ -112,7 +112,7 @@ def extract_job_image(job_meta, site_name):
         for item in participants:
             if isinstance(item, dict):
                 sites = item.get(JobConstants.SITES)
-                if site_name in sites:
+                if sites and (site_name in sites or ALL_SITES in sites):
                     return item.get(JobConstants.JOB_IMAGE)
```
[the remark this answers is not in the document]
pcnudde: Fixed in `6ad533f33`. `extract_job_image()` now does an exact site match first and only falls back to `@ALL` if no site-specific entry matches. I also updated the test so the explicit override still wins when `@ALL` appears first in the list.

devops/multicloud/scripts/deploy.py
```diff
@@ -0,0 +1,530 @@
...
+                "config_file_path": None,
+                "workspace_pvc": "nvflws",
+                "study_data_pvc_file_path": "/var/tmp/nvflare/etc/study_data_pvc.yaml",
+                "namespace": namespace,
+                "python_path": "/usr/local/bin/python3",
+            }
+            if security_context:
+                args["security_context"] = security_context
+            r["components"][i] = {"id": "k8s_launcher", "path": launcher_class, "args": args}
+    dst.write_text(json.dumps(r, indent=4))
```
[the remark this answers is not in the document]
pcnudde: Fixed in `6ad533f33`. `patch_resources_json()` now tracks whether it replaced a launcher component and raises a `RuntimeError` if no `ProcessJobLauncher` entry is found. I also added focused coverage in `tests/unit_test/tools/multicloud_deploy_test.py`.

tests/tools/multicloud/scripts/deploy.py
```diff
@@ -0,0 +1,530 @@
...
+# ---------------------------------------------------------------------------
+# Helpers
+# ---------------------------------------------------------------------------
+def run(cmd: list[str], check=True, capture=False, **kwargs) -> subprocess.CompletedProcess:
+    print(f"  $ {' '.join(cmd[:6])}{'...' if len(cmd) > 6 else ''}")
+    return subprocess.run(cmd, check=check, capture_output=capture, text=True, **kwargs)
+
+
+def run_quiet(cmd: list[str]) -> subprocess.CompletedResult:
+    return subprocess.run(cmd, capture_output=True, text=True)
```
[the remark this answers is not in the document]
pcnudde: Fixed in `6ad533f33`. `run_quiet()` now uses the correct `subprocess.CompletedProcess` annotation.

tests/tools/multicloud/scripts/deploy.py
```diff
@@ -0,0 +1,530 @@
...
+            }
+        }
+
+        if pod_exists(p.kubeconfig, p.namespace, pod_name):
+            kubectl(p.kubeconfig, "-n", p.namespace, "delete", "pod", pod_name, "--wait=false")
+            time.sleep(5)
+
+        kubectl(p.kubeconfig, "-n", p.namespace, "run", pod_name,
+                "--image=busybox", "--restart=Never",
+                f"--overrides={json.dumps(pod_spec)}")
```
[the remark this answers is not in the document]
pcnudde: Fixed in `521fdabef`. If the temporary kit-copy pod already exists, the deploy script now deletes it with normal wait semantics before reusing the same pod name, instead of relying on a fixed sleep after `--wait=false`.

tests/tools/gcp/gke/create_cluster.sh
```diff
@@ -0,0 +1,97 @@
...
+  "${LOCATION}"
+  --project
+  "${PROJECT_ID}"
+)
+
+"${create_cmd[@]}"
+"${credentials_cmd[@]}"
+
+# Save kubeconfig for multicloud deploy scripts
+REPO_ROOT="$(cd -- "${SCRIPT_DIR}/../../../.." && pwd)"
```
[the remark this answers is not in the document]
pcnudde: Fixed in `6cf7e6d0c`. `create_cluster.sh` now defines `SCRIPT_DIR` before deriving `REPO_ROOT`, so the kubeconfig save path works under `set -euo pipefail`.

tests/unit_test/app_opt/job_launcher/test_k8s_launcher.py
```diff
@@ -0,0 +1,100 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
```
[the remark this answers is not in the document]
pcnudde: not consistent in current repo

docker/Dockerfile
pcnudde: We will need it in our official image, so I think this is the right place to put this.

nvflare/app_opt/job_launcher/k8s_launcher.py
pcnudde: yes, many options, let's discuss it in the design decisions document.

nvflare/app_opt/job_launcher/k8s_launcher.py
pcnudde: yes, I needed this to set a particular security setting for amazon. It is site specific

docker/Dockerfile
pcnudde: maybe we should put it in the requirements file as an option? what do you propose.

chesterxgchen: k8s shouldn't be here for general docker. 
Do we need k8s inside docker ?

nvflare/app_opt/job_launcher/k8s_launcher.py
```diff
@@ -105,7 +106,7 @@ def __init__(
         job_config: dict,
         namespace="default",
         timeout=None,
-        pending_timeout=30,
+        pending_timeout=120,
```
chesterxgchen: should this be argument in the functions with default such as
```
   def uuid4_to_rfc1123 ( uuid_str, pending_timeout = 120) -> str:
           ....
           pending_timeout = pending_timeout 

```
    similarly, namespace = "default" hard-coded ?

nvflare/app_opt/job_launcher/k8s_launcher.py
```diff
@@ -167,6 +168,9 @@ def _make_manifest(self, job_config):
         self.pod_manifest["metadata"]["name"] = job_config.get("name")
         self.pod_manifest["spec"]["containers"] = self.container_list
         self.pod_manifest["spec"]["volumes"] = self.volume_list
+        security_context = job_config.get("security_context")
```
chesterxgchen, line 174: the security context is specified in the job_config ?

nvflare/app_opt/job_launcher/k8s_launcher.py
```diff
@@ -282,8 +286,10 @@ def __init__(
         study_data_pvc_file_path: str,
         timeout=None,
         namespace="default",
-        pending_timeout=30,
+        pending_timeout=120,
```
chesterxgchen: this seems need to be constant and variable, similarly namespace, not sure we need to hard-coded here

chesterxgchen (commented):
add a few comments

docker/Dockerfile
pcnudde: yes we do need it and it needs to be on the docker image we publish. (the parent pod needs kubernetes to launch the job pod). 

The question is, do we need multiple docker files. I think that might confuse. If we only publish 1 image we should only have 1 dockerfile I think.

nvflare/app_opt/job_launcher/k8s_launcher.py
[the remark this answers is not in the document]
pcnudde: New feature no backward compatibility issue

nvflare/app_opt/job_launcher/k8s_launcher.py
pcnudde: no, it is stored in resources.json, but the currrent code merges that into the job_config variable.

docker/Dockerfile
chesterxgchen: ask every docker depends on k8s seems to be strange. 
But I guess just a python libs, maybe its ok.

pcnudde: I think it is ok, but it should not be bare in a the docker file. we should have it in the flare dependency file (it is in any case in flare that we import it).

pcnudde: updated

nvflare/app_opt/job_launcher/k8s_launcher.py
pcnudde: fixed

nvflare/app_opt/job_launcher/k8s_launcher.py
pcnudde: fixed

tests/unit_test/app_opt/job_launcher/test_k8s_launcher.py
```diff
@@ -0,0 +1,95 @@
...
```
YuanTingHsieh: suggest merge "tests/unit_test/app_opt/job_launcher/k8s_launcher_test.py" and "tests/unit_test/app_opt/job_launcher/test_k8s_launcher.py"

devops/multicloud/README.md
```diff
@@ -0,0 +1,61 @@
...
+| `--gcp-region` | `GCP_REGION` | us-central1 | GCP region |
+| `--force-provision` | | false | Re-provision even if output exists |
+
+## Documentation
+
+- `tests/tools/gcp/gke/README.md` — GKE cluster setup + NVFlare prerequisites
+- `tests/tools/aws/eks/README.md` — EKS cluster setup + NVFlare prerequisites
+- `BUGS.md` — issues found and resolution status
+- `DEPLOY_NOTES.md` — operational notes per cloud
+- `RECOMMENDATIONS.md` — what to fix in core vs deploy tooling
```
YuanTingHsieh: these files are not included.

tests/tools/multicloud/gcp-server.yaml
```diff
@@ -0,0 +1,25 @@
+name: default
```
YuanTingHsieh: this looks identical to "tests/tools/multicloud/aws-server.yaml", what is the difference?

pcnudde: this one has gcp as the server. But I will see if I can simplify stuff here

devops/multicloud/scripts/deploy.py
```diff
@@ -0,0 +1,722 @@
...
+
+def check_auth():
+    print("Checking cloud auth ...")
+    r = run_quiet(["gcloud", "auth", "print-access-token"])
+    if r.returncode != 0:
+        sys.exit("GCP auth failed. Run: gcloud auth login")
+    r = run_quiet(["aws", "sts", "get-caller-identity"])
+    if r.returncode != 0:
+        sys.exit("AWS auth failed. Run: aws sso login")
+    print("  Auth OK")
```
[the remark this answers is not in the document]
pcnudde: Fixed in `7317112b`. `check_auth(clouds_used: set)` now only verifies the clouds actually used by participants in the loaded config. Pure-GCP or pure-AWS deploys no longer fail on the other cloud's missing credentials.

devops/multicloud/README.md
```diff
@@ -0,0 +1,61 @@
...
+
+## Options
+
+| Flag | Env var | Default | Description |
+|------|---------|---------|-------------|
+| `--gcp-image` | `GCP_IMAGE` | (required) | GCP container image |
+| `--aws-image` | `AWS_IMAGE` | (optional) | AWS container image (ECR) |
+| `--server-ip` | `GCP_SERVER_IP` | auto-reserve | Static IP for server |
+| `--gcp-project` | `GCP_PROJECT` | auto-detect | GCP project ID |
+| `--gcp-region` | `GCP_REGION` | us-central1 | GCP region |
```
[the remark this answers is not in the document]
pcnudde: Fixed. The `--gcp-project` / `--gcp-region` flags were removed entirely in `754c00031` — project and region now autoderive from the kubeconfig context name (GKE `gke_<project>_<region>_<cluster>`, EKS ARN). README table updated to show only the real flags (`--config`, `--dry-run`) in subsequent commits.

tests/unit_test/app_opt/job_launcher/test_k8s_launcher.py
pcnudde: Done in `73591f7a`. Merged `test_k8s_launcher.py` into `k8s_launcher_test.py` (kept unique coverage for `VOLUME_MOUNT_LIST` shape and pod-level securityContext; dropped the duplicate `TestUuid4ToRfc1123` since the existing file already had better coverage). Also renamed `test_job_launcher_utils.py` → `job_launcher_utils_test.py` to match the project convention.

devops/multicloud/README.md
pcnudde: Fixed in `73591f7a` — those were local scratch notes that live outside the repo. Removed from the README.

devops/aws/eks/create_cluster.sh
```diff
@@ -0,0 +1,142 @@
...
+# Create a security group allowing NFS from the VPC CIDR
+VPC_CIDR="$(aws ec2 describe-vpcs --vpc-ids "${VPC_ID}" --region "${REGION}" \
+  --query 'Vpcs[0].CidrBlock' --output text)"
+SG_ID="$(aws ec2 create-security-group \
+  --group-name "${CLUSTER_NAME}-efs-sg" \
+  --description "NFS access for NVFlare EFS" \
+  --vpc-id "${VPC_ID}" \
+  --region "${REGION}" \
+  --query 'GroupId' --output text)"
+aws ec2 authorize-security-group-ingress \
```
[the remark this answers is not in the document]
pcnudde: Fixed in `73591f7a`. The loop now captures the final `STATE` and the script exits 1 if it's not `available` after 150s, so you get a clear error instead of a downstream `IncorrectFileSystemLifeCycleState`.

devops/multicloud/scripts/deploy.py
```diff
@@ -0,0 +1,940 @@
...
+            for k, v in _flatten_set("securityContext", p.security_context):
+                helm_args += ["--set", f"{k}={v}"]
+
+        if "aws" in p.kubeconfig.lower():
+            if not aws_image:
+                print("  WARNING: No --aws-image set. AWS client will use GCP image from values.yaml.")
+                print("           EKS cannot pull from GCP Artifact Registry without mirroring.")
+            else:
+                repo, tag = aws_image.rsplit(":", 1)
+                helm_args += ["--set", f"image.repository={repo}", "--set", f"image.tag={tag}"]
```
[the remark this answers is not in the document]
pcnudde: Fixed in `73591f7a` by guarding with `if ':' in aws_image`. Subsequently the `--aws-image` CLI flag was removed entirely in `2dc69206` — image URLs now come from `clouds.<x>.image` in the YAML config. The same guard is kept for the YAML value.

devops/multicloud/deploy.py
```diff
@@ -0,0 +1,954 @@
...
+
+
+def _flatten_set(prefix: str, d: dict) -> list[tuple[str, str]]:
+    result = []
+    for k, v in d.items():
+        key = f"{prefix}.{k}"
+        if isinstance(v, dict):
+            result.extend(_flatten_set(key, v))
+        else:
+            result.append((key, str(v)))
```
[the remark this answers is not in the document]
pcnudde: Fixed in `88a54ab9`. `_flatten_set` now emits lowercase `true`/`false` for `bool` values before falling back to `str(v)`, so helm parses them as real booleans.

devops/multicloud/deploy.py
```diff
@@ -0,0 +1,937 @@
...
+        raise ValueError(f"{config_path}: missing 'participants' section")
+
+    cloud_derived: dict[str, dict] = {}
+    for cloud_name, cloud_cfg in clouds.items():
+        kc_rel = (cloud_cfg or {}).get("kubeconfig")
+        if not kc_rel:
+            continue
+        kc_path = (config_path.parent / kc_rel).resolve()
+        if kc_path.exists():
+            cloud_derived[cloud_name] = _parse_kubeconfig(kc_path, cloud_name)
```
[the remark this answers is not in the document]
pcnudde: Fixed in `88a54ab9`. `_parse_kubeconfig` now returns `{}` for non-gcp/non-aws clouds instead of raising. Azure clients would load cleanly; actual Azure server support still raises `NotImplementedError` in `reserve_ip` by design.

devops/aws/eks/delete_cluster.sh
```diff
@@ -0,0 +1,38 @@
...
+  MT_IDS="$(aws efs describe-mount-targets --file-system-id "${EFS_ID}" --region "${REGION}" \
+    --query 'MountTargets[].MountTargetId' --output text 2>/dev/null || true)"
+  for MT in ${MT_IDS}; do
+    aws efs delete-mount-target --mount-target-id "${MT}" --region "${REGION}" 2>/dev/null || true
+  done
+  # Wait for mount targets to be deleted
+  sleep 15
+  aws efs delete-file-system --file-system-id "${EFS_ID}" --region "${REGION}" 2>/dev/null || true
+  echo "Deleted EFS ${EFS_ID}"
+done
```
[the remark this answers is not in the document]
pcnudde: Fixed in `a0f7ac040`. The mount-target deletion loop now polls describe-mount-targets until count is 0 (up to 150s), and delete-file-system runs without '|| true' so a residual failure exits 1 instead of orphaning the filesystem.

nvflare/lighter/impl/helm_chart.py
```diff
@@ -183,11 +191,19 @@ def _write_server_values_yaml(self, chart_dir: str, server: Participant, fed_lea
...
-            "tcpConfigMapEnabled": True,
+            "resources": {
+                "requests": {
+                    "cpu": "2",
+                    "memory": "8Gi",
+                },
+            },
+            "securityContext": {},
+            "hostPortEnabled": False,
+            "tcpConfigMapEnabled": False,
```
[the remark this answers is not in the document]
pcnudde: not an issue

YuanTingHsieh (approved):

PR #4452, opened 2026-04-16
Fix HelmChartBuilder mutating shared comm_config_args dictFix/helm chart shared dict mutation

nvshaxie:
● Summary

  - HelmChartBuilder.build() was calling comm_config_args.update() on the shared dict obtained via
  get_prop(PropKey.COMM_CONFIG_ARGS), overwriting the server hostname to "nvflare-server" (a K8s service name).
  Because dict.update() mutates in-place, all subsequent builders and the runtime comm_config saw the corrupted
  hostname, causing every FL job in non-Helm environments to fail with gaierror: [Errno -2] Name or service not
  known.
  - Fix: deep-copy comm_config_args before mutating, for both the server and client blocks.
  - 54 nightly regression test failures caused by this bug (introduced in PR #4427).
  - NVBugs: 6086130

● Changes

  - nvflare/lighter/impl/helm_chart.py: Add import copy and copy.deepcopy() before comm_config_args.update() in
  both _build_server_chart() and _build_one_client_chart()
  - tests/unit_test/lighter/helm_chart_builder_test.py: Update 2 unit tests to assert COMM_CONFIG_ARGS is NOT
  mutated after build()

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/lighter/impl/helm_chart.py
+++ b/nvflare/lighter/impl/helm_chart.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import copy
 import os
 import shutil
 
@@ -114,10 +115,12 @@ class HelmChartBuilder(Builder):
         fed_learn_port = ctx.get(CtxKey.FED_LEARN_PORT, 8002)
         admin_port = ctx.get(CtxKey.ADMIN_PORT, 8003)
 
-        # Align comm_config.json with the chart so that internal jobs can reach the
-        # parent at <nvflare-server>:<parent_port> within the cluster namespace.
+        # Build Helm-specific comm_config on a COPY so the shared dict is not mutated.
+        # Without deep-copy, the in-place update would corrupt comm_config for non-Helm
+        # deployments (Docker, subprocess) that share the same provisioned workspace.
         comm_config_args = server.get_prop(PropKey.COMM_CONFIG_ARGS)
         if comm_config_args is not None:
+            comm_config_args = copy.deepcopy(comm_config_args)
             comm_config_args.update(
                 {
                     CommConfigArg.HOST: "nvflare-server",
@@ -221,10 +224,10 @@ class HelmChartBuilder(Builder):
         templates_dir = os.path.join(chart_dir, "templates")
         os.makedirs(templates_dir, exist_ok=True)
 
-        # Align comm_config.json with the chart so that job pods can reach
-        # the client at <client.name>:<parent_port> within the cluster namespace.
+        # Build Helm-specific comm_config on a COPY so the shared dict is not mutated.
         comm_config_args = client.get_prop(PropKey.COMM_CONFIG_ARGS)
         if comm_config_args is not None:
+            comm_config_args = copy.deepcopy(comm_config_args)
             comm_config_args.update(
                 {
                     CommConfigArg.HOST: client.name,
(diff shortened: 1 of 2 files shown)


chesterxgchen (approved):

YuanTingHsieh:
@nvshaxie why do we add HelmChartBuilder to old tests?
@IsaacYangSLA can you check if this PR is needed?

nvflare/lighter/impl/helm_chart.py
```diff
@@ -114,10 +115,12 @@ def _build_server_chart(self, project: Project, ctx: ProvisionContext):
         fed_learn_port = ctx.get(CtxKey.FED_LEARN_PORT, 8002)
         admin_port = ctx.get(CtxKey.ADMIN_PORT, 8003)
 
-        # Align comm_config.json with the chart so that internal jobs can reach the
-        # parent at <nvflare-server>:<parent_port> within the cluster namespace.
+        # Build Helm-specific comm_config on a COPY so the shared dict is not mutated.
+        # Without deep-copy, the in-place update would corrupt comm_config for non-Helm
+        # deployments (Docker, subprocess) that share the same provisioned workspace.
```
YuanTingHsieh, line 120: the question is why do we add this HelmChartBuilder to non-Helm deployments?

nvshaxie:
@YuanTingHsieh HelmChartBuilder is not new added to the old tests, it has been here for a long time. The reason it didn't cause any problem before is that builders used to be self-contained: each builder wrote its own output files and did not mutate shared participant state. HelmChartBuilder generated Helm chart files without touching anything that other builders (like StaticFileBuilder) consumed.
But this invariant seems was broken by PR #4427. The new HelmChartBuilder rewrite added two in-place mutations on the shared comm_config_args dict:
- server path: comm_config_args.update({HOST: "nvflare-server", PORT: 8102, ...})
- client path: comm_config_args.update({HOST: client.name, PORT: 8102, ...})

Because these mutate the dict returned by participant.get_prop(PropKey.COMM_CONFIG_ARGS) in place, the shared state leaks into StaticFileBuilder.finalize(), which then renders the poisoned values into comm_config.json for every participant, including non-Helm deployments that happen to share the same provisioned workspace.

nvflare/lighter/impl/helm_chart.py
nvshaxie: I think the HelmChartBuilder is add to non-Helm deployments by accident... But anyway I think it should not break other builder even if it is accidentally included in a non-Helm provisioning chain~

YuanTingHsieh: yeah it comes down to this, WHEN users add HelmChartBuilder, do we still allow this startup kit to be run in a non-helm deployment?

But maybe this does not matter, we decide to keep the provision to just generate the startup kit and certs just like before (so all existing deployment will continue to work).

And for sites that want to use K8s or Docker deployment, they would use our new tool/command "nvflare package --target docker|k8s", it is that tool that will overwrite/edit these comm_configs, i am preparing the PR

nvshaxie: Thanks @YuanTingHsieh, that makes sense — consolidating Docker/K8s-specific config into nvflare package --target ... sounds like the right long-term direction, and it naturally resolves the underlying issue by removing the mutation entirely.

So when will you plan to merge your PR? Once your PR is merged, I think this one will no longer be needed.

PR #4454, opened 2026-04-17
feat: add live job log streaming

nvidianz:
## Summary
- Adds **JobLogStreamer** (client) and **JobLogReceiver** (server) for real-time log streaming from clients to the server during federated jobs, replacing the deprecated `ErrorLogSender`/`LogReceiver` which only sent static snapshots after job completion.
- Adds **SystemLogStreamer** — a system-level widget for `resources.json` that automatically injects `JobLogStreamer` into jobs that don't already declare one, by modifying the deployed `config_fed_client.json` at `BEFORE_JOB_LAUNCH`.
- Updates provisioning templates: client uses `SystemLogStreamer` (streams `error_log.txt`), server uses `JobLogReceiver`.
- Includes `hello-log-streaming` example and unit tests.

### Why streaming runs inside the job, not in CLIENT_PARENT
The streamer must run inside the job subprocess rather than being managed from the parent process. With Docker or Kubernetes job launchers, the job may run in a different container or pod that the parent process has no filesystem access to. Streaming from CLIENT_PARENT would only work with the local process launcher. By injecting `JobLogStreamer` into the job config, the streamer always runs where the log file lives — regardless of the launch mechanism.

### Key design points
- `LogStreamer` tails growing log files, survives log rotation, and sends liveness heartbeats so the receiver can detect dead senders via idle timeout.
- `JobLogStreamer` uses a two-phase stop: `ABOUT_TO_END_RUN` sets `stop_event` only (returns immediately so framework log lines land in the file during the drain window); `END_RUN` joins the thread to keep `client_run()` alive until EOF is acknowledged.
- A fresh `Signal` is injected into the streaming `FLContext` via `put()` (not `set_prop()`) to bypass the mask-consistency check that would otherwise leave the triggered `run_abort_signal` in place.
- `JobLogReceiver` writes chunks directly to a file as they arrive so the log can be followed with `tail -f`.

## Test plan
- [ ] Run `examples/hello-world/hello-log-streaming/job.py` in simulator — verify streamed log matches original
- [ ] Run with POC mode — verify `SystemLogStreamer` injects `JobLogStreamer` into jobs without one
- [ ] Verify `ALLOW_ERROR_SENDING=false` in project.yml removes `SystemLogStreamer` from provisioned `resources.json`
- [ ] Run unit tests: `pytest tests/unit_test/app_common/streamers/log_streamer_test.py`


nvflare/app_common/logging/job_log_receiver.py
```diff
@@ -0,0 +1,146 @@
...
+            f.close()
+            stream_ctx[_KEY_RECV_FILE] = None
+
+        rc = LogStreamer.get_rc(stream_ctx)
+        client = stream_ctx.get(StreamCtxKey.CLIENT_NAME)
+        job_id = stream_ctx.get(StreamCtxKey.JOB_ID)
+
+        if rc != ReturnCode.OK:
+            self.log_error(fl_ctx, f"Live log stream from {client} job {job_id} ended with rc={rc}")
+            return
```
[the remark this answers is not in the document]
nvidianz: This is by design so partial log can be preserved in case of a crash.

nvflare/app_common/logging/job_log_receiver.py
```diff
@@ -0,0 +1,150 @@
...
+            client = stream_ctx.get(StreamCtxKey.CLIENT_NAME)
+            job_id = stream_ctx.get(StreamCtxKey.JOB_ID)
+            log_file_name = LogStreamer.get_file_name(stream_ctx) or "log.txt"
+            path = os.path.join(self._effective_dest_dir(), job_id, client, log_file_name)
+            os.makedirs(os.path.dirname(path), exist_ok=True)
+            self.log_debug(fl_ctx, f"Opening log file for {client} job {job_id}: {path}")
+            f = open(path, "wb")
+            stream_ctx[_KEY_RECV_FILE] = f
+            stream_ctx[_KEY_RECV_PATH] = path
+        f.write(data)
```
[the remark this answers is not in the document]
chesterxgchen: @nvidianz  need to handle this

nvflare/app_common/logging/system_log_streamer.py
```diff
@@ -0,0 +1,127 @@
...
+        workspace = Workspace(root_dir=workspace_root, site_name=client_name)
+        config_path = os.path.join(workspace.get_app_config_dir(job_id), JobConstants.CLIENT_JOB_CONFIG)
+        if not os.path.exists(config_path):
+            return
+
+        try:
+            with open(config_path) as f:
+                cfg = json.load(f)
+        except Exception:
+            self.log_warning(fl_ctx, f"Failed to read {config_path}; skipping log streamer injection")
```
chesterxgchen, line 90: swallow exception details

nvflare/app_common/logging/system_log_streamer.py
```diff
@@ -0,0 +1,127 @@
...
+            "path": _LOG_STREAMER_PATH,
+        }
+        if args:
+            entry["args"] = args
+        components.append(entry)
+
+        try:
+            with open(config_path, "w") as f:
+                json.dump(cfg, f, indent=2)
+        except Exception:
```
chesterxgchen, line 123: swallow exception

chesterxgchen (commented):
add a few comments

nvflare/app_common/streamers/log_streamer.py
```diff
@@ -0,0 +1,487 @@
...
+
+Relationship between the two parameters
+----------------------------------------
+- ``liveness_interval``: how often the *sender* emits a heartbeat when idle
+- ``idle_timeout``: how long the *receiver* waits before declaring the sender dead
+
+Rule: ``liveness_interval < idle_timeout``.  With the defaults (10 s and
+30 s) a healthy sender heartbeats every 10 s, so the receiver's 30 s timer
+is always reset well before it fires.
+"""
```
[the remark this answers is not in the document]
nvidianz: Handled locally. I added an optional `idle_timeout` argument to `LogStreamer.stream_log()` and validate `liveness_interval < idle_timeout` when the caller knows both values. I also updated the docstring wording to reflect the real constraint here: the sender cannot infer the receiver's deployed timeout unless it is explicitly passed in. This will be in the next push.

nvflare/app_common/streamers/log_streamer.py
```diff
@@ -0,0 +1,487 @@
...
+            t.start()
+
+    def _watchdog(self):
+        """Background thread: end this stream when it goes idle."""
+        while not self._done.wait(timeout=1.0):
+            if self._fl_ctx is None:
+                # No message has arrived yet; the clock effectively hasn't started.
+                self._last_received_time = time.time()
+                continue
+
```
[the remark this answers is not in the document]
nvidianz: Handled locally. The watchdog now calls the receiver-side `END_STREAM` hook even if no usable `fl_ctx` was ever captured, which lets `stream_runner` clean up the transport state and close the orphaned stream instead of leaving it in `tx_table`. I also added a focused unit test for the early-crash case. This will be in the next push.

nvflare/lighter/impl/static_file.py
```diff
@@ -439,7 +439,7 @@ def _modify_error_sender(self, section: str, client: Participant) -> str:
 
         assert isinstance(components, list)
         for c in components:
-            if c["id"] == "error_log_sender":
+            if c["id"] == "system_log_streamer":
                 # must remove this component
                 components.remove(c)
                 break
```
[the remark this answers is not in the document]
nvidianz: Handled locally. I renamed the helper to `_modify_system_log_streamer()` so the method name matches the component it actually edits. This will be in the next push.

examples/tutorials/self-paced-training/part-3_security_and_privacy/chapter-6_Security_in_federated_compute_system/06.3_customized_site_security/custom_client_side_job_level_authorization/security/site-1/resources.json
```diff
@@ -24,10 +24,10 @@
       "args": {}
     },
     {
-        "id": "error_log_sender",
-        "path": "nvflare.app_common.logging.log_sender.ErrorLogSender",
+        "id": "system_log_streamer",
+        "path": "nvflare.app_common.logging.system_log_streamer.SystemLogStreamer",
         "args": {}
-    }, 
+    },
```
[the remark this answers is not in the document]
nvidianz: This one is already addressed in the branch that is currently on the PR. The tutorial `resources.json` now explicitly sets `"log_file_name": "error_log.txt"` for `SystemLogStreamer`, so it matches the provisioning template behavior.

nvflare/app_common/logging/job_log_receiver.py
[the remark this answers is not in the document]
nvidianz: This one is already addressed in the branch that is currently on the PR. The receiver now derives the client/job identity from the trusted peer context on the server side and sanitizes the path components there, instead of trusting the client-provided values in `stream_ctx` for filesystem layout.

nvflare/app_common/logging/system_log_streamer.py
nvidianz: Handled locally. I stopped swallowing the read failure detail here and now log the formatted exception alongside the contextual message. This will be in the next push.

nvflare/app_common/logging/system_log_streamer.py
nvidianz: Handled locally. I added the formatted exception detail to the config write failure path as well. I also applied the same treatment to the completed-log upload failure path so the three error cases are consistent. This will be in the next push.

nvflare/app_common/logging/system_log_streamer.py
```diff
@@ -0,0 +1,184 @@
...
+        log_path = workspace.get_app_error_log_file_path(job_id)
+        if not os.path.exists(log_path):
+            self.log_info(fl_ctx, f"No error log file found for {client_name} job {job_id}")
+            return
+
+        threading.Thread(
+            target=self._stream_completed_log,
+            args=(fl_ctx, log_path, client_name, job_id),
+            daemon=True,
+        ).start()
```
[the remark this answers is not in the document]
nvidianz: Handled locally. `_on_job_completed()` now creates a fresh context before starting the background completed-log upload thread, with a fallback to the event context if the engine is unavailable. I also updated the unit test to assert that the background upload uses the fresh context. This will be in the next push.

nvflare/app_common/streamers/streamer_base.py
```diff
@@ -11,7 +11,84 @@
...
+# ---------------------------------------------------------------------------
+# Chunk-protocol wire keys
+# ---------------------------------------------------------------------------
+# Shared by FileStreamer and LogStreamer — same payload format, single namespace.
+CHUNK_PREFIX = "Streamer."
+KEY_FILE_NAME = CHUNK_PREFIX + "file_name"
+KEY_DATA = CHUNK_PREFIX + "data"
+KEY_DATA_SIZE = CHUNK_PREFIX + "size"
+KEY_EOF = CHUNK_PREFIX + "eof"
+
```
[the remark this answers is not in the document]
nvidianz: Thanks — documented this in `docs/release_notes/flare_260.rst` under a new "FileStreamer Wire-Protocol Key Namespace Change" migration subsection, calling it out as a breaking change on the wire and noting that public `FileStreamer` APIs are unaffected.

docs/release_notes/flare_260.rst
```diff
@@ -86,7 +86,7 @@ Key improvements:
   - To support various needs and backward compatibility, we now have the following default log files:
     - log.txt: default log file from previous NVFLARE versions
     - log.json: JSON format log
-    - log_error.txt: ERROR level logs to log_error.txt for quick error lookup
+    - error_log.txt: ERROR level logs to error_log.txt for quick error lookup
```
chesterxgchen, line 89: why changing the old release notes. This is past

docs/resources/log_config.json
```diff
@@ -42,7 +42,7 @@
             "class": "logging.handlers.RotatingFileHandler",
             "level": "ERROR",
             "formatter": "baseFormatter",
-            "filename": "log_error.txt",
+            "filename": "error_log.txt",
```
chesterxgchen, line 45: you can add this to the migration notes

nvflare/app_common/logging/job_log_streamer.py
```diff
@@ -0,0 +1,175 @@
...
+            receiver's ``idle_timeout``.
+        poll_interval: seconds between polls when the log file has no new data
+            (default 0.5).
+    """
+
+    def __init__(
+        self,
+        log_file_name: str = WorkspaceConstants.LOG_FILE_NAME,
+        liveness_interval: float = 10.0,
+        poll_interval: float = 0.5,
```
chesterxgchen, line 61: how much traffic we are getting with every .5 sec log streaming, this going to be more than model weights. do we have impact analysis ? 

Message Size, and delays

chesterxgchen (approved):
LGTM overall 

a few minor comments ( FLARE 2.6.0 release notes change) 
and a few questions u can address later

docs/resources/log_config.json
nvidianz: Since no one is using the old code, we probably can skip this.

docs/release_notes/flare_260.rst
nvidianz: Greptile wouldn't give me 5/5 score without this change. The old code was using the wrong name.

nvflare/app_common/logging/job_log_streamer.py
nvidianz: Compared with weights, log file is pretty small. Will do an analysis later.

PR #4462, opened 2026-04-21
Distributed provisioning, job CLI connection args, system commands, and parity tests

chesterxgchen:
## Summary

This PR completes the distributed provisioning feature and CLI enhancements, fixing several correctness issues found during review and testing.

---

### CLI fixes
    - replace -"-out-format json" to -"-format json "
    -     

### Distributed provisioning workflow (`nvflare cert` + `nvflare package`)

Implements the full manual provisioning workflow so that private keys never leave each site:

| Step | Command |
|------|---------|
| Project Admin (once) | `nvflare cert init --project <name> -o <ca-dir>` |
| Site Admin | `nvflare cert csr -n <name> -t <type> -o <csr-dir>` |
| Project Admin | `nvflare cert sign -r <name>.csr -c <ca-dir> -o <out-dir> --accept-csr-role` |
| Site Admin | `nvflare package -e grpc://<server>:<port> --dir <bundle-dir>` |

`nvflare package` supports three modes: directory (`--dir`), project-file (`-p site.yaml`), and explicit (`--cert/--key/--rootca`). Kit type is derived automatically from the signed certificate; no `-t` needed.

### Package builder pipeline fix

The YAML-mode builder assembly was stripping `WorkspaceBuilder` and `StaticFileBuilder` from `site.yaml` and ignoring their custom args, contradicting the design doc's claim that site.yaml "supports custom builders." Fixed to:
- Honor `WorkspaceBuilder` and `StaticFileBuilder` from YAML as-is (with custom args preserved)
- Replace `CertBuilder` in YAML with `PrebuiltCertBuilder` (the only intentional substitution)
- Inject defaults only when the corresponding builder is absent from YAML

### `nvflare job` — connection args on all subcommands

`--startup-kit` and `--startup-target` (the admin startup kit selector) were previously only available on `nvflare job submit`. They are now available on every job subcommand:

```
nvflare job list --startup-kit /path/to/admin/startup
nvflare job list --startup-target prod
nvflare job meta <job-id> --startup-target prod
nvflare job abort <job-id> --startup-kit /path/to/admin
nvflare job logs <job-id> --startup-target poc
```

All subcommands affected: `list`, `meta`, `abort`, `clone`, `delete`, `download`, `stats`, `logs`, `monitor`, `log-config`.

### `nvflare job` — export fixes

Three bugs in the job export workflow fixed:

1. **`--export` / `--export-dir` unrecognized arguments**: `job.py` uses argparse before `execute()` runs, so `--export`/`--export-dir` were rejected. Fixed by stripping these flags from `sys.argv` at `recipe/spec.py` import time via `_consume_recipe_args()`.

2. **"Job folder not valid" on re-export**: Partial exports leave directories without `meta.json`, causing a `RuntimeError` on the second run. Fixed by always `rmtree`-ing the output directory if it exists.

3. **`NoneType` error after export**: `execute()` returned `None` after exporting; fixed to `raise SystemExit(0)` so callers do not attempt to use the return value.

### `nvflare system` command extensions

- **`shutdown` / `restart`**: Extended to support all three targets (`server`, `client`, `all`) with optional positional `client_names`, matching the original admin console.
- **`remove-client`**: New command wrapping `Session.remove_client()`.
- **`InvalidTarget` error**: Was incorrectly reported as `CONNECTION_FAILED`; now correctly exits with `INVALID_ARGS` (exit 4).

### `_server_status_value` regression fix

`check_status` and `Session.shutdown_system()` were reading `engine.server.status` (process machine state) instead of `engine.get_engine_info().status.value` (FL engine state). This caused incorrect status reporting.

### Examples and docs

- `examples/advanced/distributed_provision/interactive_mode/README.md`: added copy-and-run shell examples for all four steps
- `docs/design/distributed_provisioning.md`: updated to reflect final design decisions
- `docs/user_guide/nvflare_cli/system_command.rst`: document all three shutdown/restart targets, add Remove Client section

---

## Test plan

- [ ] `TestDistributedProvisioningWorkflow` (16 tests): end-to-end cert init → csr → sign → package; file structure, permissions, JSON config consistency
- [ ] `TestKitParity` (10 tests): centralized (`nvflare provision`) and distributed kits produce identical file sets, JSON key structures, and endpoint values; cert/key content differs (different CAs)
- [ ] `TestYamlModeBuilderPipeline` (6 tests): YAML WorkspaceBuilder/StaticFileBuilder honored, CertBuilder replaced, defaults injected only when absent, ordering correct
- [ ] `TestSessionShutdown` / `TestSessionRestart` (16 tests): all target types and client_names combinations
- [ ] `TestSystemShutdown` / `TestSystemRestart` / `TestSystemRemoveClient`: CLI command dispatch and error handling
- [ ] Job connection args: `--startup-kit` / `--startup-target` present on all job subcommands (list, meta, abort, clone, delete, download, stats, logs, monitor, log-config)
- [ ] Job export tests: `--export` flag consumed before argparse, re-export does not raise, `SystemExit(0)` after export
- [ ] Integration test files renamed `test_xxx.py` → `xxx_test.py` (project convention)


chesterxgchen:
> ### Greptile Summary
> This PR completes the distributed provisioning feature with fixes to the YAML-mode builder pipeline (honoring `WorkspaceBuilder`/`StaticFileBuilder` from YAML, only replacing `CertBuilder` with `PrebuiltCertBuilder`), extends `shutdown`/`restart` to `server`/`client`/`all` targets, adds a `remove-client` command, and fixes `_server_status_value` to read FL engine state instead of process machine state.
> 
> * **P1** — `_add_system_connection_args` drops the `--startup_kit` underscore alias that `_add_job_connection_args` explicitly retains as backward compat; existing automation using `nvflare system ... --startup_kit` will fail with an unrecognized-argument error.
> * **P2** — `--out-format` is renamed to `--format` with no alias, silently breaking any scripts that used the old flag name.
> 
> ### Confidence Score: 4/5
> Safe to merge after resolving the --startup_kit backward-compat inconsistency in system commands
> 
> One P1 finding: --startup_kit alias is dropped from system CLI commands but retained in job CLI commands, which breaks existing automation asymmetrically. All other findings are P2 (style/compat). The core logic fixes (builder pipeline, status value, shutdown targets) look correct and are well-tested.
> 
> nvflare/tool/system/system_cli.py (_add_system_connection_args missing --startup_kit alias); nvflare/cli.py (--out-format removed)
> 
> ### Important Files Changed
> Filename	Overview
> nvflare/tool/system/system_cli.py	Extends shutdown/restart to all target types and adds remove-client command; drops --startup_kit backward compat alias inconsistently vs job_cli.py (P1)
> nvflare/fuel/flare_api/flare_api.py	shutdown/restart extended to client/all targets; session close correctly guarded to SERVER/ALL targets only; InvalidTarget propagated through command result
> nvflare/tool/package/package_commands.py	Builder pipeline fixed to honor WorkspaceBuilder/StaticFileBuilder from YAML; minor fragility with bare next() in cert injection path
> nvflare/private/fed/server/training_cmds.py	_server_status_value simplified to use engine.get_engine_info().status.value (FL engine state) instead of server.status (machine state); correct regression fix
> nvflare/recipe/spec.py	_consume_recipe_args() now runs at module import time and mutates sys.argv; execute() raises SystemExit(0) instead of returning None on export; intentional but unusual
> nvflare/private/fed/utils/fed_utils.py	require_signed_jobs now uses fstat after open (TOCTOU fix), de-duplicates warnings with _warn_once, and fails closed on parse errors; intentional security improvement
> nvflare/cli.py	--out-format renamed to --format without backward-compat alias; authz-preview and preflight-check aliases added for underscore forms
> nvflare/tool/job/job_cli.py	Job sub-commands renamed to hyphen form with underscore aliases; _add_job_connection_args extracted and added to all job sub-commands; _job_session_for_args honors --startup-kit/--startup-target
> examples/hello-world/hello-pt/job.py	Export flag handling removed from example (now handled by spec.py import-time argv consumption); if run is not None guard is now unreachable dead code
> nvflare/tool/cli_errors.py	ERROR_REGISTRY now wrapped in MappingProxyType to prevent mutation; get_error_entry() helper added with dev-mode validation; preflight reference updated
> ### Sequence Diagram
> ```mermaid
> sequenceDiagram
>     participant CLI as nvflare CLI (cli.py)
>     participant SysCLI as system_cli.py
>     participant Session as flare_api.Session
>     participant Server as FL Server
> 
>     CLI->>SysCLI: cmd_system_shutdown(args)
>     SysCLI->>SysCLI: _confirm_or_force()
>     SysCLI->>Session: _system_session(args)
>     SysCLI->>Session: sess.shutdown(target, client_names)
>     Note over Session: target in {server, client, all}
>     Session->>Server: shutdown {target} [client_names]
>     Server-->>Session: result[META]
>     alt target == server or all
>         Session->>Session: self.close()
>     end
>     Session-->>SysCLI: result dict
>     alt InvalidTarget raised
>         SysCLI-->>CLI: output_error(INVALID_ARGS, exit 4)
>     else success
>         SysCLI-->>CLI: output_ok(result)
>     end
> ```
> ### Comments Outside Diff (1)
> 1. `nvflare/cli.py`, line 343-352 ([link](https://github.com/nvidia/nvflare/blob/d125c98b29b519514ed5d3f481712b610e9b94cd/nvflare/cli.py#L343-L352))
>    [P2](#) **`--out-format` removed without backward-compat alias**
>    `--out-format` (the previously documented flag) is replaced by `--format` with no alias retained. Any script using `nvflare --out-format json` will now get an unrecognized-argument error and exit 4. Consider adding `"--out-format"` as a silent backward-compat alias so existing automation isn't broken:
> 
> Reviews (1): Last reviewed commit: ["Distributed provisioning: fix builder pi..."](https://github.com/nvidia/nvflare/commit/d125c98b29b519514ed5d3f481712b610e9b94cd) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=29187378)



> ### Greptile Summary
> This PR completes the distributed provisioning feature with fixes to the YAML-mode builder pipeline (honoring `WorkspaceBuilder`/`StaticFileBuilder` from YAML, only replacing `CertBuilder` with `PrebuiltCertBuilder`), extends `shutdown`/`restart` to `server`/`client`/`all` targets, adds a `remove-client` command, and fixes `_server_status_value` to read FL engine state instead of process machine state.
> 
> * **P1** — `_add_system_connection_args` drops the `--startup_kit` underscore alias that `_add_job_connection_args` explicitly retains as backward compat; existing automation using `nvflare system ... --startup_kit` will fail with an unrecognized-argument error.
> * **P2** — `--out-format` is renamed to `--format` with no alias, silently breaking any scripts that used the old flag name.
> 
> ### Confidence Score: 4/5
> Safe to merge after resolving the --startup_kit backward-compat inconsistency in system commands
> 
> One P1 finding: --startup_kit alias is dropped from system CLI commands but retained in job CLI commands, which breaks existing automation asymmetrically. All other findings are P2 (style/compat). The core logic fixes (builder pipeline, status value, shutdown targets) look correct and are well-tested.
> 
> nvflare/tool/system/system_cli.py (_add_system_connection_args missing --startup_kit alias); nvflare/cli.py (--out-format removed)
> 
> ### Important Files Changed
> Filename	Overview
> nvflare/tool/system/system_cli.py	Extends shutdown/restart to all target types and adds remove-client command; drops --startup_kit backward compat alias inconsistently vs job_cli.py (P1)
> nvflare/fuel/flare_api/flare_api.py	shutdown/restart extended to client/all targets; session close correctly guarded to SERVER/ALL targets only; InvalidTarget propagated through command result
> nvflare/tool/package/package_commands.py	Builder pipeline fixed to honor WorkspaceBuilder/StaticFileBuilder from YAML; minor fragility with bare next() in cert injection path
> nvflare/private/fed/server/training_cmds.py	_server_status_value simplified to use engine.get_engine_info().status.value (FL engine state) instead of server.status (machine state); correct regression fix
> nvflare/recipe/spec.py	_consume_recipe_args() now runs at module import time and mutates sys.argv; execute() raises SystemExit(0) instead of returning None on export; intentional but unusual
> nvflare/private/fed/utils/fed_utils.py	require_signed_jobs now uses fstat after open (TOCTOU fix), de-duplicates warnings with _warn_once, and fails closed on parse errors; intentional security improvement
> nvflare/cli.py	--out-format renamed to --format without backward-compat alias; authz-preview and preflight-check aliases added for underscore forms
> nvflare/tool/job/job_cli.py	Job sub-commands renamed to hyphen form with underscore aliases; _add_job_connection_args extracted and added to all job sub-commands; _job_session_for_args honors --startup-kit/--startup-target
> examples/hello-world/hello-pt/job.py	Export flag handling removed from example (now handled by spec.py import-time argv consumption); if run is not None guard is now unreachable dead code
> nvflare/tool/cli_errors.py	ERROR_REGISTRY now wrapped in MappingProxyType to prevent mutation; get_error_entry() helper added with dev-mode validation; preflight reference updated
> ### Sequence Diagram
> ```mermaid
> sequenceDiagram
>     participant CLI as nvflare CLI (cli.py)
>     participant SysCLI as system_cli.py
>     participant Session as flare_api.Session
>     participant Server as FL Server
> 
>     CLI->>SysCLI: cmd_system_shutdown(args)
>     SysCLI->>SysCLI: _confirm_or_force()
>     SysCLI->>Session: _system_session(args)
>     SysCLI->>Session: sess.shutdown(target, client_names)
>     Note over Session: target in {server, client, all}
>     Session->>Server: shutdown {target} [client_names]
>     Server-->>Session: result[META]
>     alt target == server or all
>         Session->>Session: self.close()
>     end
>     Session-->>SysCLI: result dict
>     alt InvalidTarget raised
>         SysCLI-->>CLI: output_error(INVALID_ARGS, exit 4)
>     else success
>         SysCLI-->>CLI: output_ok(result)
>     end
> ```
> ### Comments Outside Diff (1)
> 1. `nvflare/cli.py`, line 343-352 ([link](https://github.com/nvidia/nvflare/blob/d125c98b29b519514ed5d3f481712b610e9b94cd/nvflare/cli.py#L343-L352))
>    [P2](#) **`--out-format` removed without backward-compat alias**
>    `--out-format` (the previously documented flag) is replaced by `--format` with no alias retained. Any script using `nvflare --out-format json` will now get an unrecognized-argument error and exit 4. Consider adding `"--out-format"` as a silent backward-compat alias so existing automation isn't broken:
> 
> Reviews (1): Last reviewed commit: ["Distributed provisioning: fix builder pi..."](https://github.com/nvidia/nvflare/commit/d125c98b29b519514ed5d3f481712b610e9b94cd) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=29187378)

--out-format json to --format json is design decision, there is --out-format json is yet release yet. nothing is break

nvflare/tool/system/system_cli.py
```diff
@@ -31,8 +32,8 @@
 
 def _add_system_connection_args(parser):
     parser.add_argument(
-        "--startup_kit",
         "--startup-kit",
+        dest="startup_kit",
```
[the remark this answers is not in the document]
chesterxgchen: Fixed — added `--startup_kit` and `--startup_target` underscore aliases to `_add_system_connection_args`, matching the backward-compat aliases already present in `_add_job_connection_args`.

nvflare/tool/package/package_commands.py
```diff
@@ -385,41 +384,46 @@ def _handle_package_yaml_mode(args, scheme, host, port):
...
+        # Inject after WorkspaceBuilder, or at the front if none.
+        ws_pos = next((i for i, b in enumerate(all_builders) if isinstance(b, WorkspaceBuilder)), -1)
+        all_builders.insert(ws_pos + 1, cert_builder)
+
+    if not any(isinstance(b, WorkspaceBuilder) for b in all_builders):
+        all_builders.insert(0, WorkspaceBuilder())
+
+    if not any(isinstance(b, StaticFileBuilder) for b in all_builders):
+        cert_pos = next(i for i, b in enumerate(all_builders) if isinstance(b, PrebuiltCertBuilder))
+        all_builders.insert(cert_pos + 1, StaticFileBuilder(scheme=scheme))
```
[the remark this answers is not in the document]
chesterxgchen: Fixed — replaced bare `next()` with `next(..., None)` plus an `assert cert_pos is not None` that gives a clear message if the invariant is ever violated.

examples/hello-world/hello-pt/job.py
```diff
@@ -73,18 +71,14 @@ def main():
...
-            print("Result can be found in :", run.get_result())
-            print()
+    # Run FL simulation
+    env = SimEnv(num_clients=n_clients)
+    run = recipe.execute(env)
+    if run is not None:
+        print()
+        print("Job Status is:", run.get_status())
+        print("Result can be found in :", run.get_result())
+        print()
```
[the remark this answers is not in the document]
chesterxgchen: Fixed — removed the `if run is not None:` guard. `execute()` now raises `SystemExit(0)` on the export path and always returns a `Run` object on the normal path, so the check was dead code.

nvflare/tool/system/system_cli.py
```diff
@@ -390,7 +405,7 @@ def cmd_system_shutdown(args):
 
 
 def cmd_system_restart(args):
-    from nvflare.fuel.flare_api.api_spec import AuthenticationError, NoConnection
+    from nvflare.fuel.flare_api.api_spec import AuthenticationError, InvalidTarget, NoConnection
```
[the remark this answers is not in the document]
chesterxgchen: Fixed — added a guard in `cmd_system_shutdown` and `cmd_system_restart`: when `target \!= 'client'` and `client_names` is non-empty, the command now exits with INVALID_ARGS (exit 4) and an explicit error message instead of silently discarding the names. Tests `test_shutdown_rejects_client_names_for_non_client_target` and `test_restart_rejects_client_names_for_non_client_target` cover this path.

chesterxgchen:
@greptileai

nvflare/tool/cert/cert_cli.py
```diff
@@ -17,7 +17,9 @@
 import argparse
 from typing import Optional
 
-_VALID_CERT_TYPES = ["client", "server", "org_admin", "lead", "member"]
+from nvflare.tool.cert.cert_constants import VALID_CERT_TYPES
+
+_VALID_CERT_TYPES = list(VALID_CERT_TYPES)
```
IsaacYangSLA: This pattern, importing a tuple than converting it to a list, seems strange.  Can you explain why you need this step?

nvflare/tool/package/package_commands.py
```diff
@@ -24,25 +24,23 @@
...
-_ADMIN_ROLES = {AdminRole.ORG_ADMIN, AdminRole.LEAD, AdminRole.MEMBER}
-_VALID_CERT_TYPES = {"client", "server", "org_admin", "lead", "member"}
-_KIT_TYPE_TO_ROLE = {
-    "org_admin": AdminRole.ORG_ADMIN,
-    "lead": AdminRole.LEAD,
-    "member": AdminRole.MEMBER,
-}
-_DUMMY_SERVER_NAME = "dummy-server"
+_ADMIN_ROLES = set(ADMIN_CERT_TYPES)
+_VALID_CERT_TYPES = set(VALID_CERT_TYPES)
```
IsaacYangSLA, line 41: And now it's a set.  Any reason?  To increase lines of codes?

nvflare/job_config/fed_job_config.py
```diff
@@ -147,10 +147,7 @@ def generate_job_config(self, job_root):
         """
         job_dir = os.path.join(job_root, self.job_name)
         if os.path.exists(job_dir):
-            if self._is_valid_job_folder(job_dir):
-                shutil.rmtree(job_dir, ignore_errors=True)
-            else:
-                raise RuntimeError(f"Job folder {job_dir} already exists and is not a valid job folder.")
+            shutil.rmtree(job_dir, ignore_errors=True)
```
YuanTingHsieh: i don't know if we should always remove, that folder could be folders not generated by NVFlare

nvflare/lighter/provision.py
```diff
@@ -128,21 +128,65 @@ def define_provision_parser(parser):
     global _provision_parser
     _provision_parser = parser
     # Action flags — mutually exclusive but no longer required; default is -g behavior
-    parser.add_argument("-p", "--project_file", type=str, default=None, help="file to describe FL project")
+    parser.add_argument(
+        "-p",
+        "--project-file",
+        "--project_file",
```
YuanTingHsieh, line 134: why need to duplicate? 133 and 134

nvflare/lighter/provision.py
```diff
@@ -128,21 +128,65 @@ def define_provision_parser(parser):
...
         "-g",
         "--generate",
         action="store_true",
         help="generate a sample project.yml and exit (default when no flag given)",
     )
-    parser.add_argument("-e", "--gen_edge", action="store_true", help="generate a sample edge project.yml and exit")
+    parser.add_argument(
+        "-e",
+        "--gen-edge",
+        "--gen_edge",
```
YuanTingHsieh, line 149: why need to repeat?

nvflare/lighter/provision.py
```diff
@@ -128,21 +128,65 @@ def define_provision_parser(parser):
...
     # Optional arguments
     parser.add_argument("-w", "--workspace", type=str, default="workspace", help="directory used by provision")
-    parser.add_argument("-c", "--custom_folder", type=str, default=".", help="additional folder to load python codes")
-    parser.add_argument("--add_user", type=str, default="", help="yaml file for added user")
-    parser.add_argument("--add_client", type=str, default="", help="yaml file for added client")
-    parser.add_argument("-s", "--gen_scripts", action="store_true", help="generate test scripts like start_all.sh")
+    parser.add_argument(
+        "-c",
+        "--custom-folder",
+        "--custom_folder",
```
YuanTingHsieh, line 160: why need to repeat?

nvflare/tool/cert/cert_cli.py
chesterxgchen: The `list()` was added in d125c98b2 to preserve the original type — the pre-refactor code had `_VALID_CERT_TYPES` as a plain list literal `["client", "server", ...]`. When the values were extracted to `cert_constants` as a tuple, `list()` was used to keep the same type. Since argparse `choices=` accepts any iterable, the conversion is not necessary and we've simplified to use `VALID_CERT_TYPES` directly.

nvflare/tool/package/package_commands.py
chesterxgchen: The `set()` is intentional — it mirrors the pattern used for `_ADMIN_ROLES = set(ADMIN_CERT_TYPES)` on the line above. Using a set for allowlist membership checks (`if kit_type not in _VALID_CERT_TYPES`) was a deliberate choice from a security review: set lookup is O(1) and signals that this is an allowlist. The wrapping of the imported tuple in `set()` avoids duplicating the literal values while preserving the set semantics.

nvflare/private/fed/utils/fed_utils.py
```diff
@@ -65,23 +74,32 @@ def require_signed_jobs(workspace: Workspace) -> bool:
...
-            with open(server_config_path) as f:
+            fd = os.open(server_config_path, os.O_RDONLY)
+            with os.fdopen(fd) as f:
+                st = os.fstat(f.fileno())
+                if st.st_mode & (_stat.S_IWGRP | _stat.S_IWOTH):
+                    _warn_once(
+                        logger,
+                        f"writable:{server_config_path}",
+                        "fed_server.json is group/world-writable — require_signed_jobs policy "
+                        "can be altered by other local users (TOCTOU risk)",
```
YuanTingHsieh, line 94: why do we need this change?

nvflare/recipe/spec.py
```diff
@@ -38,10 +43,21 @@ def _peek_recipe_args(argv: Optional[List[str]] = None) -> tuple:
...
             i += 1
         else:
+            remaining.append(argv[i])
             i += 1
+    sys.argv[1:] = remaining
     return export, export_dir
 
 
+# Strip export flags at import time so the caller's ArgumentParser doesn't see them.
+_RECIPE_EXPORT, _RECIPE_EXPORT_DIR = _consume_recipe_args()
```
YuanTingHsieh: DO we want to do this though?
there needs to be documentation saying we will intercept the --export arg?

nvflare/tool/system/system_cli.py
```diff
@@ -31,17 +32,19 @@
 
 def _add_system_connection_args(parser):
     parser.add_argument(
-        "--startup_kit",
         "--startup-kit",
+        "--startup_kit",  # backward compat
```
YuanTingHsieh: do we have this system cli in 2.7?
if not we dont need 2 args?

nvflare/tool/system/system_cli.py
```diff
@@ -31,17 +32,19 @@
...
+        dest="startup_kit",
         default=None,
         help="path to the admin startup kit directory (overrides target-based config lookup)",
     )
     parser.add_argument(
         "--startup-target",
+        "--startup_target",  # backward compat
```
YuanTingHsieh: do we have this system cli in 2.7?
if not we dont need 2 args?

nvflare/tool/cli_errors.py
```diff
@@ -228,6 +230,14 @@
         "hint": "Check the recipe class run() method for errors.",
     },
 }
+ERROR_REGISTRY: Mapping[str, Mapping[str, str]] = MappingProxyType(_ERROR_REGISTRY)
```
YuanTingHsieh, line 239: why we need both ERROR_REGISTRY and _ERROR_REGISTRY

examples/advanced/distributed_provision/interactive_mode/README.md
```diff
@@ -1,41 +1,172 @@
...
+---
+
+## Step 2 — Site Admin: generate CSR + private key
+
+Each site admin runs this on their own machine, one call per site.
+
 ```bash
-./03_project_admin_sign.sh <csr_path> <ca_dir> <out_dir>
+# Server site admin
+./02_site_admin_csr.sh ./server.yml ./distprov_demo/server-csr
```
YuanTingHsieh, line 71: interactive mode and scripted mode both use some scripts? what is the difference?

docs/user_guide/nvflare_cli/system_command.rst
```diff
@@ -18,12 +18,13 @@ Command Usage
...
-     resources  show server and client resource usage
-     shutdown   shut down the FL server
-     restart    restart the FL server
-     version    show NVFlare version on each remote site
-     log-config change logging level on server or client sites
+     status         show server and client status
+     resources      show server and client resource usage
+     shutdown       shut down server, clients, or all
+     restart        restart server, clients, or all
+     remove-client  remove a client from the federation
```
YuanTingHsieh, line 25: not sure if it is really removing the client,
recall it was only for the server to remove the client from its cache.
if the CP still running, it will re-connect back in 5 mins

tests/unit_test/tool/job/job_create_deprecated_test.py
```diff
@@ -139,16 +139,16 @@ def test_job_help_lists_deprecated_subcommands(self):
         parser = def_job_cli_parser(subs)["job"]
 
         help_text = parser.format_help()
-        assert "list_templates" in help_text
-        assert "show_variables" in help_text
+        assert "list-templates" in help_text
```
YuanTingHsieh, line 142: should we just remove this command along with all templates?

examples/advanced/distributed_provision/interactive_mode/README.md
chesterxgchen: one is used for automation, one is for human interactive

docs/user_guide/nvflare_cli/system_command.rst
chesterxgchen: yes, this allows one to shutdown the server before it comes back :-( , not sure what the original intent. I originally left this out, but this decided to add this back in this PR. No strong opinions

tests/unit_test/tool/job/job_create_deprecated_test.py
chesterxgchen: maybe, maybe next release

nvflare/tool/system/system_cli.py
chesterxgchen: Same as above — aliases removed since this command is new in this PR.

nvflare/tool/system/system_cli.py
chesterxgchen: Correct — `nvflare system` is new in this PR (did not exist in 2.7). Removed both underscore aliases (`--startup_kit` / `--startup_target`) from `_add_system_connection_args`; no backward-compat aliases needed.

nvflare/lighter/provision.py
chesterxgchen: `--project-file` and `--project_file` (underscore) are both needed: `--project_file` was the original form in 2.7, so we keep it as a backward-compat alias while adding the hyphenated canonical form.

nvflare/lighter/provision.py
chesterxgchen: Same — `--custom_folder` existed in 2.7; `--custom-folder` is the new canonical form. Both kept for backward compatibility.

nvflare/lighter/provision.py
chesterxgchen: Same as above — `--gen_edge` existed in 2.7; `--gen-edge` is the new canonical form. Both forms accepted for backward compatibility.

nvflare/job_config/fed_job_config.py
chesterxgchen: Good catch. Restored the safety check: the folder is only removed if it passes `_is_valid_job_folder` (has `meta.json`) OR `_is_partial_export_folder` (only contains app-named subdirectories, which is the signature of a partial NVFlare export). Any folder with unrecognised content still raises `RuntimeError`.

nvflare/tool/cli_errors.py
chesterxgchen: `_ERROR_REGISTRY` is the internal mutable `dict` that can be populated at module load time. `ERROR_REGISTRY = MappingProxyType(_ERROR_REGISTRY)` exposes an immutable read-only view as the public API — callers cannot accidentally mutate it. This is a standard pattern for exposing a constant registry: build it mutably, then freeze it.

nvflare/private/fed/utils/fed_utils.py
chesterxgchen: This adds `require_signed_jobs()` to support the distributed provisioning workflow: in distributed provisioning, each site generates its own private key with a cert signed by the project CA independently, so the server's startup kit may not share the CA chain used to sign `__nvfl_sig.json`. The function lets operators set `require_signed_jobs: false` in `fed_server.json` to disable signature enforcement without restarting the server. `_warn_once()` suppresses repeated log noise for the same condition on every job submission. Added inline comments to the code to explain this context (commit e51406b).

nvflare/recipe/spec.py
chesterxgchen: Yes, and added a 4-line comment at the call site (commit 7bfba18) explaining the design: the mutation must happen at import time so that `--export`/`--export-dir` are stripped before the caller's `ArgumentParser.parse_args()` runs. `job.py` always calls `parse_args()` before `execute()`, so doing it lazily inside `execute()` would be too late. The function is only safe here because `spec.py` is imported as a process entry-point module, never as a library.

YuanTingHsieh (approved):

IsaacYangSLA (approved):

nvidianz (approved):

pcnudde (approved):

PR #4465, opened 2026-04-21
add azure k8s devops

YuanTingHsieh:
### Description

Add Azure K8s devops.

- Tested 1 server and 1 client
- Need to test real job running after PR#4464 is merged

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh:
@greptile review

devops/multicloud/deploy.py
```diff
@@ -181,13 +184,16 @@ def load_config(config_path: Path) -> DeployConfig:
 
     gcp = cloud_derived.get("gcp", {})
     aws = cloud_derived.get("aws", {})
+    azure = (clouds.get("azure") or {})
```
IsaacYangSLA, line 187: gcp and aws are using cloud_derived.get, but azure is using clouds.  The syntax is different, too.

IsaacYangSLA (approved):
Only one comment on deploy.py, but not critical.

devops/multicloud/deploy.py
YuanTingHsieh: Ok will see if we can make them consistent

PR #4469, opened 2026-04-22
K8s launcher: use CellNet workspace transfer instead of a shared workspace PVC

pcnudde:
This PR updates the Kubernetes job launcher to run launched jobs without mounting the shared workspace PVC.

Instead, launched jobs now:
- mount `startup/` from a Secret
- use an `emptyDir` as their writable workspace
- download `local/` and the job workspace over CellNet/F3 before startup
- upload job results back to the parent over the same path on shutdown

This preserves `local/` workspace behavior while isolating each launched job's writable workspace.


nvflare/private/fed/app/server/runner_process.py
```diff
@@ -151,6 +155,14 @@ def parse_arguments():
     parser.add_argument("--port", "-port", type=str, help="service port", required=True)
     parser.add_argument("--ssid", "-id", type=str, help="SSID", required=True)
     parser.add_argument("--parent_url", "-p", type=str, help="parent_url", required=True)
+    parser.add_argument(
+        "--parent_conn_sec",
+        "-pcs",
```
nvidianz: probably don't need alias here since those are for program to use.

nvflare/utils/job_launcher_utils.py
```diff
@@ -109,25 +110,29 @@ def generate_server_command(fl_ctx) -> str:
...
 
 
-# TODO: remove in follow-up PR once K8s launcher is updated to use get_launcher_resource_spec
+def get_launcher_resource_spec(job_meta, site_name, mode):
+    """Extract the launcher-mode resource spec for a site from full job meta."""
+    resource_spec = job_meta.get(JobMetaKey.RESOURCE_SPEC.value, {}) or {}
+    return get_site_launcher_spec(resource_spec.get(site_name), mode)
+
+
 def extract_job_image(job_meta, site_name):
```
nvidianz: Do we need the same call for docker?

nvflare/private/fed/server/server_engine.py
```diff
@@ -277,6 +278,17 @@ def _start_runner_process(self, job, job_clients, snapshot, fl_ctx: FLContext):
             JobProcessArgs.SSID: ("--ssid", str(server_state.ssid)),
             JobProcessArgs.OPTIONS: ("--set", command_options),
         }
+
+        # Pass the parent's listener connection-security mode through to the
+        # launched runner so its short-lived bootstrap CellNet child can connect
+        # back with the same transport/security settings before the normal job
+        # cell is created.
+        params = cell.get_internal_listener_params()
```
nvidianz, line 286: Can't use internal listener here. This is normally the local address, only good for inter-process communications. For kubernetes, we need to setup a service just like server and use that endpoint in the job process.

EDIT: This is ok. The launcher already sets up a service for this address.

nvflare/app_opt/job_launcher/k8s_launcher.py
```diff
@@ -25,7 +29,17 @@
 from nvflare.apis.fl_context import FLContext
 from nvflare.apis.job_def import JobMetaKey
 from nvflare.apis.job_launcher_spec import JobHandleSpec, JobLauncherSpec, JobProcessArgs, JobReturnCode, add_launcher
-from nvflare.utils.job_launcher_utils import extract_job_image, get_client_job_args, get_server_job_args
+from nvflare.app_opt.job_launcher.workspace_cell_transfer import (
+    ENV_WORKSPACE_OWNER_FQCN,
```
nvidianz, line 34: Not sure workspace is the right term here. Workspace includes all the writable data. Only startup kit should be sent to the job.

nvflare/app_opt/job_launcher/workspace_cell_transfer.py
```diff
@@ -0,0 +1,587 @@
...
+# one bootstrap cell for the life of the job. Creating a second cell with
+# the same FQCN crashes CellNet's registry ("there is already a cell
+# named ..."), and nothing in CellNet unregisters the name reliably after
+# cell.stop(), so keeping one alive is the simplest contract.
+_bootstrap_cell: Cell | None = None
+_bootstrap_net_agent: NetAgent | None = None
+_bootstrap_lock = threading.Lock()
+
+
+def _get_bootstrap_cell(args, owner_fqcn: str, secure_mode: bool) -> Cell:
```
nvidianz, line 349: No need to create another cell. Each job already have a cell. CellNet is designed for multiplexing with channel/topic.

nvflare/app_opt/job_launcher/workspace_cell_transfer.py
```diff
@@ -0,0 +1,587 @@
...
+    global _bootstrap_cell, _bootstrap_net_agent
+    with _bootstrap_lock:
+        if _bootstrap_cell is None:
+            _bootstrap_cell, _bootstrap_net_agent = _create_bootstrap_cell(
+                args=args, owner_fqcn=owner_fqcn, secure_mode=secure_mode
+            )
+            atexit.register(_close_bootstrap_cell)
```
nvidianz: Cells are closed automatically on exit.

nvflare/private/fed/app/server/runner_process.py
pcnudde: Fixed in 6621ae771. I removed the short alias here and in the client worker parser, and switched the launcher-side arg wiring to use the long `--parent_conn_sec` form consistently.

pcnudde:
Addressed the Greptile P1 from the review summary in 6621ae771. `_handle_prepare_download` now cleans up and returns `_make_error(...)` instead of re-raising out of the CellNet callback, and I added a regression test covering bundle-preparation failure. This one was reported as an outside-diff summary comment, so there was no resolvable review thread to close.

nvflare/lighter/templates/helm/client/role.yaml
```diff
@@ -9,6 +9,9 @@ rules:
 - apiGroups: [""]
   resources: ["pods"]
   verbs: ["create", "delete", "get", "list", "watch"]
+- apiGroups: [""]
```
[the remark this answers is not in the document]
pcnudde: only 1 instance per deployment. not an issue

nvflare/app_opt/job_launcher/workspace_cell_transfer.py
pcnudde: re-using the same cell adds a lot of complexity as the existing cell isn;t ready at the beginning. So it involved much more code than creatign a new cell

nvflare/app_opt/job_launcher/k8s_launcher.py
```diff
@@ -399,14 +516,22 @@ def launch_job(self, job_meta: dict, fl_ctx: FLContext) -> JobHandleSpec:
...
+            workspace_job_id=raw_job_id,
         )
         pod_manifest = job_handle.get_manifest()
-        self.logger.debug(f"launch job with k8s_launcher. {pod_manifest=}")
+        self.logger.debug(
+            "launch job with k8s_launcher: pod_name=%s namespace=%s image=%s",
+            pod_manifest["metadata"]["name"],
+            self.namespace,
+            job_image,
+        )
```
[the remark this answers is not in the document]
pcnudde: Fixed in df0e00c9d. I switched startup Secret naming to a dedicated `site_name_to_rfc1123()` helper that keeps the name RFC1123-safe and adds a stable hash suffix, so names like `site#1` and `site1` no longer collide. I also added a focused unit test for that collision case.

nvflare/app_opt/job_launcher/k8s_launcher.py
pcnudde: we send local and startup kit. 
workspace name also come from existing workspace pvc

nvflare/app_opt/job_launcher/workspace_cell_transfer.py
pcnudde: Cleaned this up in 1fcad2789. We still keep a single bootstrap cell per process to avoid recreating the same FQCN during download/upload, but the `atexit.register(_close_bootstrap_cell)` hook itself was redundant since `upload_results()` already closes the bootstrap cell in its `finally` path.

nvflare/app_opt/job_launcher/workspace_cell_transfer.py
```diff
@@ -0,0 +1,602 @@
...
+        if path.is_absolute() or ".." in path.parts:
+            raise ValueError(f"unsafe zip member: {info.filename}")
+
+
+def _validate_job_zip_members(zf: zipfile.ZipFile, job_id: str) -> None:
+    _validate_relative_zip_members(zf)
+    for info in zf.infolist():
+        parts = PurePosixPath(info.filename).parts
+        if not parts or parts[0] != job_id:
+            raise ValueError(f"zip member outside job workspace: {info.filename}")
```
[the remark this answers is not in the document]
pcnudde: Fixed in 1fcad2789. I added an explicit symlink-entry check in `_validate_job_zip_members()` so uploaded result archives now reject symlink members before extraction, and I added a focused unit test that builds a symlink-containing zip and verifies we return an invalid-request error.

docs/design/no_shared_pvc_k8s_workspace_design.md
```diff
@@ -0,0 +1,227 @@
...
+2. Workspace bootstrap and artifact transfer use CellNet/F3 streaming.
+3. The job pod writes only to its own `emptyDir` workspace.
+
+## Workspace Layout
+
+The launched job pod uses this layout under the normal NVFlare workspace root:
+
+```text
+/var/tmp/nvflare/workspace/
+  startup/          read-only Secret mount
```
YuanTingHsieh, line 28: do we assume the startup folder for each site is already put inside this Secret?

nvflare/utils/job_launcher_utils.py
pcnudde: Good question. After merging upstream/main, Docker now uses get_job_launcher_spec(...) on its launch path. The scheduler path here still intentionally uses get_site_launcher_spec(..., "process") because check_resources() and the existing resource managers consume process-style resource requirements rather than launcher-specific nested specs. So I left this scheduler behavior unchanged.

nvflare/app_opt/job_launcher/workspace_cell_transfer.py
```diff
@@ -0,0 +1,602 @@
...
+    _zip_results_to_file(args.workspace, args.job_id, temp_bundle.name)
+    bundle_sha = _hash_file(temp_bundle.name)
+    bundle_size = os.path.getsize(temp_bundle.name)
+    logger.info(
+        "[ws-transfer] upload_results start job=%s bundle_size=%d target=%s",
+        args.job_id,
+        bundle_size,
+        owner_fqcn,
+    )
+    downloader = None
```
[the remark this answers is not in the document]
pcnudde: Fixed. upload_results() now moves bundle creation, hashing, and size calculation inside the try/finally so the temp bundle is removed even if zip creation or hashing fails before the F3 transaction starts. I also added a focused regression test for that cleanup path.

nvflare/app_opt/job_launcher/k8s_launcher.py
```diff
@@ -370,25 +462,66 @@ def launch_job(self, job_meta: dict, fl_ctx: FLContext) -> JobHandleSpec:
...
+        startup_dir = workspace_obj.get_startup_kit_dir()
+        engine = fl_ctx.get_engine()
+        owner_cell = getattr(engine, "cell", None) if engine else None
+        if owner_cell is None:
+            raise RuntimeError("missing parent CellNet cell for workspace transfer")
+
+        workspace_transfer = WorkspaceTransferManager.get_or_create(owner_cell)
+        workspace_transfer_token = workspace_transfer.add_job(raw_job_id, workspace_root)
+
+        startup_secret_name = self._ensure_startup_secret(site_name, startup_dir)
```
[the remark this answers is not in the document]
pcnudde: Fixed. The launch path now removes the workspace-transfer record for any exception after add_job() and before a successful pod launch, including startup-secret creation failures. I added a regression test that exercises the _ensure_startup_secret failure path specifically.

nvflare/app_opt/job_launcher/workspace_cell_transfer.py
```diff
@@ -0,0 +1,603 @@
...
+            with zipfile.ZipFile(file_path) as zf:
+                _validate_job_zip_members(zf, record.job_id)
+                zf.extractall(record.workspace_root)
+            logger.info("[ws-transfer] publish_results extracted job=%s into %s", job_id, record.workspace_root)
+        except ValueError as e:
+            return _make_error(str(e))
+        except zipfile.BadZipFile as e:
+            return _make_error(f"invalid results bundle for {job_id}: {e}")
+        finally:
+            shutil.rmtree(temp_dir, ignore_errors=True)
```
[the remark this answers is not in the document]
pcnudde: Fixed. _handle_publish_results() now converts unexpected extraction and filesystem exceptions into an INVALID_REQUEST reply instead of letting them escape the callback. I also added coverage for an os.makedirs failure path.

nvflare/app_opt/job_launcher/workspace_cell_transfer.py
```diff
@@ -0,0 +1,605 @@
...
+        _zip_results_to_file(args.workspace, args.job_id, temp_bundle.name)
+        bundle_sha = _hash_file(temp_bundle.name)
+        bundle_size = os.path.getsize(temp_bundle.name)
+        logger.info(
+            "[ws-transfer] upload_results start job=%s bundle_size=%d target=%s",
+            args.job_id,
+            bundle_size,
+            owner_fqcn,
+        )
+        cell = _get_bootstrap_cell(args, owner_fqcn, secure_mode)
```
YuanTingHsieh, line 547: do we need:

"_wait_for_bootstrap_ready(cell, owner_fqcn)" here?

nvflare/app_opt/job_launcher/k8s_launcher.py
```diff
@@ -285,28 +338,65 @@ class K8sJobLauncher(JobLauncherSpec):
...
+            "data": data,
+        }
+        try:
+            self.core_v1.read_namespaced_secret(name=secret_name, namespace=self.namespace)
+            self.core_v1.replace_namespaced_secret(name=secret_name, namespace=self.namespace, body=secret_body)
+            self.logger.debug("Updated startup Secret %s", secret_name)
+        except ApiException as e:
+            if getattr(e, "status", None) == 404:
+                self.core_v1.create_namespaced_secret(namespace=self.namespace, body=secret_body)
+                self.logger.debug("Created startup Secret %s", secret_name)
```
YuanTingHsieh: _ensure_startup_secret() has a check-then-create race. If two jobs for the same site launch concurrently, both can observe 404 from read_namespaced_secret(), then one create_namespaced_secret() succeeds and the other returns 409 Conflict. That 409 is currently unhandled and can fail launch_job() even though the secret now exists. Please recover from 409 by re-reading or replacing the secret.

YuanTingHsieh (commented):
3 comments, otherwise LGTM

nvflare/app_opt/job_launcher/k8s_launcher.py
```diff
@@ -285,28 +338,65 @@ class K8sJobLauncher(JobLauncherSpec):
...
+                        data[fname] = base64.b64encode(f.read()).decode()
+
+        secret_body = {
+            "apiVersion": "v1",
+            "kind": "Secret",
+            "metadata": {"name": secret_name, "namespace": self.namespace},
+            "type": "Opaque",
+            "data": data,
+        }
+        try:
```
IsaacYangSLA, line 388: The secret data doesn't change for one particular site. So once it's created, there is no need to try to update it. Suggest add a thread lock and keep a secret_created flag. If it's true, no need to go over the try block.

docs/design/no_shared_pvc_k8s_workspace_design.md
pcnudde: yes, I clarified the doc

nvflare/app_opt/job_launcher/workspace_cell_transfer.py
pcnudde: I'll add it for safety. It should already exist, but safer and consistent to add

nvflare/app_opt/job_launcher/k8s_launcher.py
pcnudde: fixing

nvflare/app_opt/job_launcher/k8s_launcher.py
pcnudde: I will replace it anyway to be safe. (also of another pod gets started).

YuanTingHsieh (approved):
LGTM

PR #4471, opened 2026-04-22
Fix JobRunner completion context race

pcnudde:
## Summary
This PR fixes a server-side race in JobRunner when completed jobs are processed concurrently with scheduling.

The completion thread was reusing the scheduling loop's FLContext, and both paths mutate FLContextKey.CURRENT_JOB_ID. Under concurrent jobs, that could fire JOB_COMPLETED with the wrong job id, leaving completed jobs stuck in DefaultJobScheduler.scheduled_jobs and blocking later jobs from being scheduled.

## Changes
- pass engine into the completion thread instead of the shared FLContext
- create a fresh engine.new_context() for each completed job before firing JOB_COMPLETED
- add a regression test covering completion-event context isolation

## Validation
- tests/unit_test/private/fed/server/job_runner_test.py


YuanTingHsieh (approved):
LGTM

PR #4472, opened 2026-04-23
Multi-study CLI: runtime study management commands and registry

chesterxgchen:
## Problem

NVFlare's multi-study support was provisioning-only: studies had to be defined in `project.yml`
before provisioning and could not be modified at runtime. Operators needed to reprovision and
restart the server to enroll new sites, add users, or register new studies — impractical for
long-running federated deployments.

## Solution

This PR introduces the `nvflare study` CLI command group — 8 subcommands for runtime lifecycle
management of studies without reprovisioning or server restart:

| Command | Who can call | What it does |
|---------|-------------|--------------|
| `nvflare study register` | `project_admin`, `org_admin` | Register a new study with initial site enrollment |
| `nvflare study show` | any enrolled user | Show study definition (site-orgs, users) |
| `nvflare study list` | any enrolled user | List studies visible to the caller |
| `nvflare study remove` | `project_admin` | Remove a study (blocked if jobs exist) |
| `nvflare study add-site` | `project_admin`, `org_admin` | Enroll additional sites |
| `nvflare study remove-site` | `project_admin`, `org_admin` | Remove sites from a study |
| `nvflare study add-user` | `project_admin`, org-enrolled `org_admin` | Add a user to a study's admin list |
| `nvflare study remove-user` | `project_admin`, org-enrolled `org_admin` | Remove a user from a study's admin list |

`project_admin` specifies sites grouped by org with `--site-org <org>:<site>`.
`org_admin` specifies only their own org's sites with `--sites`.

All commands return a structured JSON envelope (`--format json`) or human-readable text.
Changes take effect immediately via hot-reload — no server restart required.

### 1. Two-layer INVALID_ARGS enforcement

The role-based input-shape contract (`project_admin` uses `--site-org`, `org_admin` uses `--sites`)
is enforced at two independent layers:

- **CLI fast-fail** — `_try_get_caller_role()` reads the caller's role from their startup-kit
  certificate before opening a server connection. Mis-shaped commands are rejected immediately
  with a structured `INVALID_ARGS` error, before any network round-trip.
- **Server authoritative** — `_requested_site_orgs()` re-checks `ConnProps.USER_ROLE` on every
  mutation. The CLI layer is advisory; the server always has the final word. This ensures
  correctness even if the CLI is bypassed or the cert is stale.

### 2. Atomic mutation with registry validation

All mutations follow a strict sequence:

```
acquire lock → load JSON → deepcopy → apply callback → validate via
StudyRegistry constructor → atomic write (tempfile rename) →
publish to StudyRegistryService → release lock in finally
```

The `StudyRegistry` constructor is the single validation gate — duplicate-site detection,
naming rules, and structure checks all live there. An invalid callback result is rejected
before the write is attempted, preserving consistency.

### 3. Study visibility scoping

`list_studies` and `show_study` visibility mirrors the session study-scoping model already
established for jobs: `project_admin` sees all studies; `org_admin` sees studies in which
their org has enrolled at least one site.

### 4. SessionSpec completeness

All 8 study methods are declared as `@abstractmethod` on `SessionSpec` so implementors get
compile-time guidance rather than runtime `AttributeError` surprises.

### 5. Security: empty-org guard

`_requested_site_orgs()` validates that `_caller_org(conn)` is non-empty before using it as a
dict key. A misconfigured org_admin cert (empty ORG property) would otherwise silently create
`{"": [sites]}` in the registry — now it raises `INVALID_ARGS` instead.

### New files
| File | Purpose |
|------|---------|
| `nvflare/private/fed/server/study_cmds.py` | Server-side `StudyCmdModule` with 8 command handlers |
| `nvflare/tool/study/study_cli.py` | CLI command group: argparse → FLARE API session calls |
| `docs/user_guide/nvflare_cli/study_command.rst` | User guide for the study command family |
| `docs/design/participant_registration.md` | Design doc for participant registration flow |
| `tests/unit_test/private/fed/server/study_cmds_test.py` | 63 server-side unit tests |
| `tests/unit_test/tool/study/study_cli_test.py` | 30 CLI-layer unit tests |

### Modified files
- `nvflare/security/study_registry.py` — duplicate-site detection, flat-site union derivation
- `nvflare/fuel/flare_api/api_spec.py` — 8 `@abstractmethod` declarations on `SessionSpec`
- `nvflare/fuel/flare_api/flare_api.py` — concrete implementations for all 8 study session methods
- `nvflare/cli.py` — study command group registered
- `nvflare/tool/cli_errors.py` — shared structured error output helpers
- `nvflare/apis/fl_constant.py`, `client.py`, `utils/format_check.py` — study name constants and validation
- `nvflare/lighter/provision.py` — `study_registry.json` generation from `project.yml`
- `nvflare/private/fed/server/client_manager.py`, `cmd_utils.py`, `server_cmd_modules.py` — study-scoped client visibility and command module registration
- `nvflare/fuel/hci/server/login.py` — study login integration
- `nvflare/private/fed/utils/identity_utils.py` — cert-to-identity helpers used by CLI role resolution
- Design docs (`multistudy.md`, `multistudy_cli.md`) — two-layer enforcement section, Phase 2 dataset-mapping content removed
- `docs/user_guide/admin_guide/multi_study_guide.rst` — cross-reference to runtime study management guide

### Test coverage (93 unit tests)
- All 8 command handlers (register, show, list, remove, add-site, remove-site, add-user, remove-user)
- Three INVALID_ARGS rules × 3 mutation commands (9 server-side + 9 CLI-side tests)
- Atomicity: write not called when validation fails
- Lock timeout path
- User membership visibility scoping
- Duplicate-site rejection in StudyRegistry
- CLI golden paths and error-envelope mapping

## Checklist
- [x] Unit tests pass (`python3 -m pytest tests/unit_test/private/fed/server/study_cmds_test.py tests/unit_test/tool/study/ tests/unit_test/security/study_registry_test.py`)
- [x] black / isort / flake8 clean
- [x] Design docs updated (two-layer enforcement; Phase 2 dataset design removed)
- [x] User guide added (`docs/user_guide/nvflare_cli/study_command.rst`)
- [x] Rebased on upstream main


nvflare/apis/utils/format_check.py
```diff
@@ -29,6 +29,7 @@
     "org": r"^[A-Za-z0-9_]+$",
     "simple_name": r"^[A-Za-z0-9_]+$",
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in 6451000. Changed the  pattern from the study-name regex () to the provisioning  pattern (), so any name valid at provisioning time is accepted for study enrollment.

nvflare/private/fed/server/study_cmds.py
```diff
@@ -0,0 +1,682 @@
...
+                    return {
+                        "error_code": "INVALID_SITE",
+                        "message": f"Sites {bad_sites} are unknown or do not belong to org '{org}'.",
+                        "hint": "Each site must be currently connected to the server and provisioned under the specified org.",
+                        "exit_code": 4,
+                    }
+            site_orgs = self._normalize_site_orgs(study_def)
+            removed = []
+            not_enrolled = []
+            for org, sites in requested.items():
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in 6451000. Changed the `"site"` pattern from the study-name regex (lowercase-only) to the provisioning `"client"` pattern (`^[A-Za-z0-9-_]+$`), so any name valid at provisioning time is accepted for study enrollment.

nvflare/private/fed/server/study_cmds.py
```diff
@@ -0,0 +1,682 @@
...
+                        existing.add(site)
+                        added.append(site)
+            return self._site_mutation_payload(parsed.study, study_def, added=added, already_enrolled=already_enrolled)
+
+        self._with_mutation(conn, _mutate)
+
+    def cmd_remove_study_site(self, conn: Connection, args: List[str]):
+        parser = _study_parser(AdminCommandNames.REMOVE_STUDY_SITE, include_sites=True, include_site_org=True)
+        try:
+            parsed = parser.parse_args(args[1:])
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in 6451000. Introduced three typed exception classes (`_InvalidArgsError`, `_InvalidSiteError`, `_InvalidStudyNameError`) and updated `_validate_study_name`, `_validate_site_names`, and `_requested_site_orgs` to raise the appropriate type. The three command handlers now catch each class separately, eliminating all substring-based classification.

nvflare/private/fed/server/study_cmds.py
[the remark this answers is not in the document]
chesterxgchen: Fixed in 6451000. Introduced three typed exception classes (`_InvalidArgsError`, `_InvalidSiteError`, `_InvalidStudyNameError`) and updated `_validate_study_name`, `_validate_site_names`, and `_requested_site_orgs` to raise the appropriate type. The three command handlers now catch each class separately, eliminating all substring-based classification.

nvflare/private/fed/server/study_cmds.py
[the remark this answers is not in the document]
chesterxgchen: Fixed in 6451000. The silent `return` is replaced with an explicit `INTERNAL_ERROR` reply, so a callback that accidentally returns `None` now always sends a response to the caller rather than hanging the connection.

nvflare/private/fed/server/study_cmds.py
```diff
@@ -0,0 +1,686 @@
...
+
+    def _with_mutation(self, conn: Connection, mutation_cb):
+        if not StudyRegistryService.acquire_lock(_LOCK_TIMEOUT_SECS):
+            self._error(
+                conn,
+                "LOCK_TIMEOUT",
+                "Study registry is busy.",
+                hint="Another study mutation is in progress. Retry shortly.",
+                exit_code=3,
+            )
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in 566203c. Changed `_validate_site_orgs` to raise `_InvalidArgsError` instead of plain `ValueError`, so an invalid org name (e.g. `--site-org @bad:site1`) now returns a clean `INVALID_ARGS` error response rather than an unhandled exception.

nvflare/private/fed/server/study_cmds.py
```diff
@@ -0,0 +1,686 @@
...
+                if not site:
+                    continue
+                if site in seen_sites:
+                    raise ValueError(f"site '{site}' appears in more than one org group")
+                seen_sites.add(site)
+                sites.append(site)
+            if not sites:
+                raise ValueError(f"--site-org '{item}' must contain at least one site")
+            result.setdefault(org, [])
+            result[org].extend(sites)
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in 3b052ba. Changed all `ValueError` raises in `_parse_sites` and `_parse_site_orgs` to `_InvalidArgsError`, so malformed `--sites` and `--site-org` inputs are now caught by the typed handler and returned as a clean `INVALID_ARGS` structured error.

nvflare/private/fed/server/client_manager.py
```diff
@@ -163,6 +163,7 @@ def authenticated_client(self, request, fl_ctx: FLContext, client_type) -> Optio
             return None
 
         secure_mode = fl_ctx.get_prop(FLContextKey.SECURE_MODE, False)
+        client_org = ""
         if secure_mode:
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in 9101848. The org extraction was incorrectly gated behind `secure_mode`. Reading the org field from the cert does not require signature verification, so we now extract it in both secure and non-secure modes. The identity-verification block (signature + nonce check) remains secure-mode only.

chesterxgchen:
**Re: stale org-admin visibility after full site removal**

By design. An empty `site_orgs[org]` entry (all sites removed) represents "org is associated with the study, zero sites currently enrolled" — intentionally distinct from "org never participated". This lets the org_admin re-enroll their own sites later without requiring a project_admin to re-register the org. Pruning the key on zero-site would strip that self-service capability and require a new `register` or a dedicated "add-org" command that doesn't exist.

If a project_admin wants to fully revoke an org's access, that is a separate operation (currently out of scope for Phase 1).

nvflare/private/fed/server/study_cmds.py
```diff
@@ -0,0 +1,680 @@
...
+                new_current = []
+                for site in current:
+                    if site in sites:
+                        removed.append(site)
+                    else:
+                        new_current.append(site)
+                for site in sites:
+                    if site not in current_set:
+                        not_enrolled.append(site)
+                site_orgs[org] = new_current
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in a90e0596. Replaced `setdefault` with an explicit `if org not in site_orgs: not_enrolled.extend(sites); continue` guard — unenrolled orgs never get a phantom key written to the registry.

The second point (empty list after full removal grants stale visibility) is intentional: an org that removes all its sites retains the key so org_admins can self-service re-enroll without a project_admin re-registering the org. We discussed and agreed this is by design.

docs/design/multistudy.md
```diff
@@ -136,44 +137,68 @@ participants:
 
 studies:
   cancer-research:
```
pcnudde, line 139: I don't understand why we need to change the schema here. We know the orgs of the site (it is just above in the yaml) so we should not repeat it here. 
(I understand you need it in the json, but I don't think we need it here in the project.yml)

chesterxgchen: In distributed provisioon, there is no such association

when org-admin of company X add a site to the study 
    --site hospital-a 
there is no indication where hospital-a is part of X or not. we want to make sure the company X org admin can only add site from org X not from other companies.

pcnudde (commented):
Posted one requested review finding.

nvflare/tool/poc/poc_commands.py
```diff
@@ -1093,30 +1093,8 @@ def _run_poc(
...
-        output_error("INVALID_ARGS", exit_code=4, detail=str(e))
-        raise SystemExit(4)
-    except Exception as e:
-        output_error("INTERNAL_ERROR", exit_code=5, detail=str(e))
-        raise SystemExit(5)
-    if result is False:
-        return
-
-    output_ok({"status": "cleaned"})
+    _clean_poc(poc_workspace)
```
pcnudde, line 1097: [P1] `--schema` can execute `poc clean`

The top-level `--schema` path dispatches directly to the subcommand handler and relies on handlers to call `handle_schema_flag` before doing work. This handler no longer does that, so `nvflare poc clean --schema` reaches `_clean_poc` and can remove the POC workspace instead of printing schema. This also drops the structured JSON/error handling and `--force` behavior currently on main.

pcnudde (commented):
Posting the Study CLI documentation finding.

docs/user_guide/nvflare_cli/study_command.rst
```diff
@@ -0,0 +1,226 @@
...
+.. code-block:: shell
+
+   # project_admin: register with per-org site groupings
+   nvflare study register --name cancer-research \
+       --site-org org_a:hospital-1 \
+       --site-org org_a:hospital-2 \
+       --site-org org_b:clinic-1
+
+   # org_admin: register and enroll own org's sites
+   nvflare study register --name cancer-research --sites hospital-1,hospital-2
```
pcnudde: [P2] Study CLI examples use unsupported flags

The implemented parser takes the study name, study, and user as positional arguments, not `--name`, `--study`, or `--user` options. The examples here will be rejected by argparse as unrecognized arguments. The guide should use forms like `nvflare study register cancer-research --site-org ...` and `nvflare study add-user cancer-research trainer@org_a.com`.

nvflare/tool/poc/poc_commands.py
chesterxgchen: POC clean changes (--force, --schema, prompt removal) are intentionally excluded from this PR. They will be addressed in a separate `poc_enhancement` branch to keep this PR focused on multi-study CLI only.

docs/user_guide/nvflare_cli/study_command.rst
chesterxgchen: Fixed in b718692f. All examples now use positional arguments: `nvflare study register cancer-research`, `nvflare study add-user cancer-research trainer@org_a.com`, etc. Option descriptions updated to match.

docs/design/multistudy_cli.md
```diff
@@ -0,0 +1,899 @@
...
+
+| Flag | Type | Required | Description |
+|------|------|----------|-------------|
+| `<study>` | str | Yes | Registered study name |
+| `<user>` | str | Yes | Username or cert CN string to store in the study membership list |
+| `--startup-kit` | str | No* | Explicit path to the startup kit directory |
+| `--startup-target` | `poc\|prod` | No* | Resolves startup kit path from `~/.nvflare/config.conf` |
+| `--schema` | flag | No | Print command schema as JSON and exit |
+
+\* `--startup-kit` and `--startup-target` are mutually exclusive optional selectors. The startup kit must be resolvable via `--startup-kit`, `NVFLARE_STARTUP_KIT_DIR`, or `~/.nvflare/config.conf` using the explicit `--startup-target` or the default `poc` target — see [Connection Flags](#connection-flags). If no source resolves, the command fails.
```
pcnudde, line 302: do I always have to give a --startup-target ? can I have it just always point to prod? (seems the by default it resolves to poc if i don't give this option.)

chesterxgchen: currently default is poc

pcnudde: but If I dont have a poc environment? (do I need to have a poc?)

chesterxgchen: no, then  you need to 
--startup-target prod. ( assuming you have config the prod.startup_kit location via nvflare config) 
 or use NVFLARE_STARTUP_KIT_DIR
 
 Lets me see if I can further simply this flow in next enhanceement

pcnudde (approved):

PR #4474, opened 2026-04-24
Remove nvfletc PVC and exclude study PVC map from job pods

pcnudde:
## Summary
- remove the nvfletc/etc PVC from multicloud Helm provisioning and generated participant deployments
- keep the K8s launcher study-data PVC map under `local/study_data_pvc.yaml`
- exclude `local/study_data_pvc.yaml` from job pod workspace bundles while preserving other `local/` files


IsaacYangSLA (approved):
LGTM.

YuanTingHsieh (approved):

PR #4477, opened 2026-04-24
Add site runtime packaging

YuanTingHsieh:
### Description

This PR enables site-local runtime selection during `nvflare package` by reusing the existing full-YAML `--project-file` flow.

Instead of introducing a new package config concept, each site can now provide its own local project YAML with one participant and, optionally, one runtime builder:

- no builder: process
- `DockerLauncherBuilder`: Docker
- `K8sRuntimeBuilder`: K8s

## What changed

- Enhanced `DockerLauncherBuilder` to support site-local Docker runtime settings:
  - `docker_image`
  - `network`
  - `python_path`
  - `default_job_env`
  - `default_job_container_kwargs`
- Updated Docker packaging behavior to:
  - generate `startup/start_docker.sh`
  - replace `process_launcher` with `docker_launcher`
  - switch to `PassthroughResourceManager`
  - remove `resource_consumer`
- Added `K8sRuntimeBuilder` for site-local K8s packaging:
  - reuses `HelmChartBuilder`
  - replaces `process_launcher` with `k8s_launcher`
  - writes `local/study_data_pvc.yaml`
  - switches to `PassthroughResourceManager`
  - removes `resource_consumer`
- Updated the Docker launcher template to use the configured Docker network
- Added focused unit tests for both Docker and K8s runtime builders
- Added a design doc describing the site-local packaging model and current workflow gaps

## Why

The goal is to keep centralized provisioning unchanged for normal process-based deployments while letting each site decide its own runtime packaging during the local `nvflare package` step.

This keeps runtime choice local to the site and builds on package behavior that already supports full project YAML with `builders:`.

## Current workflow

Central admin:

- runs `nvflare provision -p project.yml`

Site admin:

- prepares a local full project YAML for the site
- chooses Docker or K8s by specifying the runtime builder
- runs `nvflare package --project-file <site-project.yml> ...`

## Testing

Ran:

```bash
./venv/bin/python -m pytest tests/unit_test/lighter/docker_launcher_builder_test.py tests/unit_test/lighter/k8s_site_runtime_builder_test.py -q
```

## Known gaps

This PR does not yet address some existing package workflow gaps:

- YAML-mode `--dir` still expects raw cert material named `<participant>.crt`, `<participant>.key`, and `rootCA.pem`
- package output reporting is still generic and may print `start.sh` even for Docker/K8s packaging
- local site project YAML is not yet validated against the centrally provisioned project for identity/topology consistency

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/docs/design/site_runtime_packaging_design.md
+++ b/docs/design/site_runtime_packaging_design.md
@@ -1,0 +1,180 @@
+# Site Runtime Packaging Design
+
+## Goal
+
+Keep centralized provisioning unchanged for normal process-based deployments, while letting
+each site choose its own runtime packaging during the local `nvflare package` step.
+
+The site-local runtime choices are:
+
+- process
+- Docker
+- K8s
+
+## Chosen Approach
+
+Reuse the existing `nvflare package --project-file ...` path.
+
+- Central admin uses one central `project.yml` for `nvflare provision`.
+- Each site uses one local full project YAML for `nvflare package`.
+- The site-local YAML contains one participant and, optionally, one runtime builder.
+
+This keeps the design builder-driven and avoids adding a new package config concept.
+
+## Runtime Mapping
+
+- process: no runtime builder
+- Docker: `nvflare.lighter.impl.docker_launcher.DockerLauncherBuilder`
+- K8s: `nvflare.lighter.impl.k8s_runtime.K8sRuntimeBuilder`
+
+Runtime builders are for `server` and `client` kits only. Admin kits should remain
+process-based.
+
+## Why This Fits Current Code
+
+`nvflare package --project-file ...` already supports full project YAML and already reuses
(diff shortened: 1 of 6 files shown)


chesterxgchen:
- P2 nvflare/lighter/impl/k8s_runtime.py:74: K8sRuntimeBuilder accepts workspace_mount_path and passes it to HelmChartBuilder, but the injected K8sJobLauncher still hard-codes
      study_data_pvc_file_path to /var/tmp/nvflare/workspace/local/study_data_pvc.yaml. If a site sets workspace_mount_path: /workspace, the parent container is started with /
      workspace, but job launch tries to read the PVC map from the default path and will fail before creating the job pod. The new test even uses workspace_mount_path="/workspace"
      while asserting the default path, so it locks in the bad behavior. Derive this path from workspace_mount_path or remove the configurable mount path from this builder until all
      launcher paths are consistent.

PR #4481, opened 2026-04-26
Enhance distributed provisioning workflow

chesterxgchen:
## What this builds

This PR implements distributed provisioning as a role-separated
request/approve/package workflow. Participants can generate and keep their own
private keys, Project Admins can approve certificate requests, and requesters can
package the signed approval into a normal FLARE startup kit.

Primary public commands:

```bash
# Project Admin: initialize root CA once
nvflare cert init --profile project_profile.yaml -o ca

# Requester: create request zip from participant definition file
nvflare cert request --participant site-1.yaml

# Project Admin: approve the request zip against the project profile
nvflare cert approve site-1/site-1.request.zip --ca-dir ca --profile project_profile.yaml

# Requester: build startup kit (skips out-of-band fingerprint verification)
nvflare package site-1/site-1.signed.zip

# Requester: build startup kit with interactive fingerprint confirmation
nvflare package site-1/site-1.signed.zip --confirm-rootca

# Requester: build startup kit with automated fingerprint verification
nvflare package site-1/site-1.signed.zip --expected-rootca-fingerprint SHA256:AA:BB:...
```

`nvflare package <signed.zip>` is always valid. The two flags are optional:
- `--confirm-rootca`: prompts the user to compare the displayed fingerprint with
  the out-of-band value received from the Project Admin (interactive/human flows).
- `--expected-rootca-fingerprint`: verifies the fingerprint non-interactively
  (scripted/automated flows). Packaging warns when neither flag is provided.

## Key Design Decisions

- **Distributed provisioning is request/approve/package**
  - Requesters generate their own private key and CSR locally.
  - The Project Admin approves a request zip and returns a signed zip.
  - The requester packages the signed zip into a normal FLARE startup kit.

- **Private keys stay with the participant**
  - Request zips and signed zips never contain private keys.
  - Only CSR, signed certificate, root CA, and approval metadata are exchanged.

- **Participant definition files are the request input**
  - `nvflare cert request -p/--participant <participant.yaml>` reads identity
    from one participant definition file.
  - The file uses a project-style structure: top-level project `name` and one
    `participants` entry.
  - Client and user definitions do not include server endpoint fields — the
    Project Admin injects the approved server endpoint from `project_profile.yaml`
    into every signed zip during `cert approve`.

- **Project Admin controls the server endpoint and communication scheme**
  - `project_profile.yaml` defines project `name`, `scheme`, `connection_security`,
    and the approved server endpoint (`host`, `fed_learn_port`, `admin_port`).
  - `nvflare cert approve` signs this endpoint into the signed zip.
  - Client and user participant files do not choose or specify the server endpoint.

- **Connection configuration is split by ownership**
  - Project-wide defaults come from the Project Admin profile.
  - Server-side `connection_security` may be overridden locally during packaging
    for deployments behind proxies, load balancers, or ingress.
  - Server-local connection overrides are excluded from the sanitized approval
    metadata because they are local package-time behavior, not Project Admin policy.

- **Trust boundary: approval covers identity, cert, and federation parameters**
  - The Project Admin signature covers: participant identity, signed cert,
    `rootCA.pem`, and federation connection parameters.
  - `builders:` blocks in participant definition files are **not** included in
    request or signed zips and are not approved by the Project Admin. Builders
    are local, package-time behavior applied on the requester's machine.
  - Features requiring coordinated builder configuration across participants
    (HE, CC) are not directly supported; use centralized `nvflare provision`.

- **Root CA fingerprint verification is optional but recommended**
  - `nvflare package <signed.zip>` alone is valid and proceeds without fingerprint verification.
  - `--confirm-rootca` displays the root CA fingerprint for the user to manually compare with the
    value received out-of-band from the Project Admin (interactive flows).
  - `--expected-rootca-fingerprint` verifies the fingerprint non-interactively for automation.
  - When neither flag is provided, packaging prints a warning that the fingerprint was not
    verified out-of-band. The signed zip signature is still validated regardless.
  - The signed zip always includes `rootCA.pem`; requesters do not need a separate file.

- **Generated startup kits remain compatible with existing FLARE runtime**
  - The output layout and runtime startup behavior match centrally provisioned kits.
  - Certificate generation reuses centralized provisioning logic.
  - Server certificate SAN/default-host behavior follows centralized provisioning
    conventions, including `localhost` support for local/demo workflows.

### Step 1 - Project Admin: Create project profile and initialize Root CA

```bash
nvflare cert init --profile project_profile.yaml -o ca
```

Creates `ca/rootCA.pem`, `ca/rootCA.key`, and `ca/ca.json`.

### Step 2 - Requesters: Create Request Zips

Each requester runs on the machine that should keep the private key:

```bash
nvflare cert request --participant server.yaml
nvflare cert request --participant site-1.yaml
nvflare cert request --participant alice.yaml
```

Participant definition files specify project name, participant name, org, type,
and role when applicable. Server definitions include ports; client and user
definitions do not include server endpoint fields. The request command creates
a local folder (e.g., `site-1/`) with `site-1.key`, `site-1.csr`, `site.yaml`,
`request.json`, and `site-1.request.zip`. Only the request zip is sent to the
Project Admin; the private key stays local.

### Step 3 - Project Admin: Approve Request Zips

```bash
nvflare cert approve site-1/site-1.request.zip --ca-dir ca --profile project_profile.yaml
```

Creates `site-1/site-1.signed.zip` and prints `rootca_fingerprint_sha256`. The
Project Admin shares this fingerprint out-of-band. The signed zip already
contains `rootCA.pem`.

### Step 4 - Requesters: Package Startup Kits

```bash
# Interactive: user compares the displayed fingerprint with the out-of-band value
nvflare package site-1/site-1.signed.zip --confirm-rootca

# Automated: fingerprint verified non-interactively
nvflare package site-1/site-1.signed.zip --expected-rootca-fingerprint SHA256:AA:BB:...

# No fingerprint verification (proceeds with a warning)
nvflare package site-1/site-1.signed.zip
```

Validates the signed zip, combines it with the local private key, verifies
key/cert/hash consistency, and writes a startup kit under
`workspace/<project>/prod_NN/<name>/`.

## Dynamic provisioning

After the project starts, new participants reuse the same flow against the
existing `ca/` and `project_profile.yaml`; existing participants do not need to
be repackaged:

```bash
nvflare cert request --participant site-3.yaml
nvflare cert approve site-3/site-3.request.zip --ca-dir ca --profile project_profile.yaml
nvflare package site-3/site-3.signed.zip --expected-rootca-fingerprint SHA256:AA:BB:...
```

## Implementation summary

- Adds public `nvflare cert init`, `cert request`, `cert approve`, and
  `nvflare package` signed-zip workflow.
- Keeps request private keys local and enforces private-key exclusion from zip artifacts.
- Adds participant definition parsing and project profile validation.
- Adds signed metadata validation, hash-chain checks, key/cert matching,
  certificate chain/expiry checks, safe zip handling, and path traversal guards.
- Reuses centralized certificate generation through `CertBuilder`.
- Fixes server cert SAN for flat `site.yaml` requests: falls back to
  `project_profile.server.host` so the approved FQDN is always covered.
- Adds optional root CA fingerprint verification with warning when skipped.
- Updates design docs, user guides, and distributed provisioning examples.

## Verification

```bash
python3 -m pytest tests/unit_test/tool/cert/cert_commands_test.py -q
python3 -m pytest tests/unit_test/tool/package/package_commands_test.py -q
python3 -m pytest tests/unit_test/tool/cli_schema_test.py tests/unit_test/lighter/project_test.py -q
python3 -m pytest tests/integration_test/distributed_provisioning_test.py -q -m 'not slow'
```

Local results:

- `cert_commands_test.py`: 230 passed
- `package_commands_test.py`: 201 passed
- distributed provisioning integration: 26 passed, 1 deselected
- `git diff --check`, `isort`, and `flake8`: passed


nvflare/tool/package/package_commands.py
```diff
@@ -248,6 +422,214 @@ def _load_project_from_file(path: str) -> tuple:
...
+            output_error_message(
+                "KEY_INVALID",
+                f"Failed to load private key '{key_path}': {e}.",
+                "Provide the private key generated with the original request.",
+                None,
+                exit_code=1,
+            )
+        if cert_public != key_public:
+            output_error_message(
+                "KEY_CERT_MISMATCH",
```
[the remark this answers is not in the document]
chesterxgchen: Valid point. In production  always calls  so the unbound path is never reached, but initializing  and  before the  block is the right fix — it makes the code testable under mocked error handlers and eliminates the NameError risk. Will fix.

nvflare/tool/package/package_commands.py
```diff
@@ -42,16 +60,171 @@
...
+    if not project_name or not isinstance(project_name, str) or not project_name.strip():
+        _reject_invalid_project_name(project_name, code=code, hint=hint)
+        return
+    if len(project_name) > 64:
+        _reject_invalid_project_name(project_name, code=code, hint="Project names must be 64 characters or fewer.")
+    if os.sep in project_name or (os.altsep and os.altsep in project_name) or project_name.startswith("."):
+        _reject_invalid_project_name(project_name, code=code, hint=hint)
+    if not _SAFE_PROJECT_NAME_PATTERN.fullmatch(project_name):
+        _reject_invalid_project_name(project_name, code=code, hint=hint)
+
```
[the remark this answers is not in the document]
chesterxgchen: Valid — adding `return` after each `_reject_invalid_project_name` call (lines 91, 93, 95) is the right fix. In production `sys.exit()` prevents fall-through, but without it a 65-char name containing `..` would emit two errors. Consistent with the first guard's `return` at line 89. Will fix.

nvflare/tool/cert/cert_cli.py
```diff
@@ -259,6 +261,85 @@ def _def_cert_sign_parser(cert_sub: argparse._SubParsersAction) -> argparse.Argu
...
+    p.add_argument(
+        "--out",
+        required=False,
+        default=None,
+        dest="output_dir",
+        help="Request folder. Default: ./<name>.",
+    )
+    p.add_argument("--force", action="store_true", default=False, help="Overwrite existing request files.")
+    p.add_argument("--schema", action="store_true", default=False, help="Print JSON schema for this command and exit.")
+    _add_compat_output_arg(p)
```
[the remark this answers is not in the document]
chesterxgchen: Acknowledged. The comment at the call site already documents that argparse exposes no public API for this — SUPPRESS is insufficient for subparsers because hidden commands still appear as `==SUPPRESS==` in help output. The fragility is real (Python is free to rename `_choices_actions`), but the alternatives (custom HelpFormatter that strips the marker, or accepting the `==SUPPRESS==` display) are more complex or worse UX. Keeping the current approach with the explanatory comment and accepting the private-API risk as a known tradeoff.

nvflare/tool/package/package_commands.py
```diff
@@ -474,81 +856,783 @@ def _handle_package_yaml_mode(args, scheme, host, port):
...
+        seen.add(name)
+        names.append(name)
+    return names
 
-    # Step 2: Validate required args and endpoint up front (before any file I/O)
-    has_project_file = bool(getattr(args, "project_file", None))
 
-    if has_project_file and getattr(args, "name", None):
+def _read_zip_json(zf: zipfile.ZipFile, name: str, zip_path: str):
+    try:
```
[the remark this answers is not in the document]
chesterxgchen: Fair observation. The two implementations are intentionally separate because the request zip and signed zip have different member sets and different error codes (`INVALID_ARGS` vs `INVALID_SIGNED_ZIP`), so a single shared function would need to parameterize both. A shared `zip_utils.py` with the common validation logic (traversal check, size limit, symlink guard, key exclusion) and separate thin wrappers for each call site would be the clean refactor. Filing as a follow-up cleanup item rather than blocking this PR — the current divergence is small and both implementations are correct.

chesterxgchen:
@greptileai

chesterxgchen:
Addressed the latest Greptile P1 comments in commit 7617ed53a:\n\n- Added the missing return after CERT_ALREADY_EXISTS in _write_zip_nofollow so mocked error handlers cannot fall through and overwrite the zip.\n- Changed _read_request_zip member-mismatch handling to return None so handle_cert_approve stops cleanly after validation failure.\n- Added/updated regression coverage for both mocked-exit paths.\n\nVerification run locally:\n- python3 -m pytest tests/unit_test/tool/cert/cert_commands_test.py - 184 passed\n- targeted regression tests - 3 passed\n- git diff --check\n- black --check\n- isort --check-only\n- flake8

chesterxgchen:
Addressed the latest package fallthrough review comments in commit 0833d1cd2:\n\n- Added returns after BUILD_FAILED in both selected-participant and internal YAML package build paths.\n- Guarded package result handling when build helpers return an integer error code, including signed-zip and explicit material modes.\n- Added regression tests for mocked-exit fallthrough behavior.\n\nVerification run locally:\n- python3 -m pytest tests/unit_test/tool/package/package_commands_test.py -q: 206 passed\n- git diff --check\n- black --check\n- isort --check-only\n- flake8

chesterxgchen:
Addressed the latest Greptile P1 package comments in commit 0833d1cd2:

- Added returns after BUILD_FAILED in _build_selected_participant_package and _handle_package_yaml_mode so mocked error handlers do not fall through into prod_dir cleanup.
- Guarded non-dict results from _build_selected_participant_package in signed-zip packaging before adding rootca_fingerprint_sha256.
- Applied the same non-dict guard to explicit material mode for consistency.
- Added regression tests for selected-participant, YAML, signed-zip, and explicit package build-error paths.

Verification:
- python3 -m pytest tests/unit_test/tool/package/package_commands_test.py -q - 206 passed
- git diff --check
- black --check
- isort --check-only
- flake8

chesterxgchen:
Follow-up pushed in commit 0cce1f1af:

- Added defensive returns after package CLI validation errors in handle_package so mocked error handlers do not fall through into later processing.

Verification after this commit:
- python3 -m pytest tests/unit_test/tool/package/package_commands_test.py -q - 207 passed
- git diff --check
- black --check nvflare/tool/package/package_commands.py
- isort --check-only nvflare/tool/package/package_commands.py
- flake8 nvflare/tool/package/package_commands.py

chesterxgchen:
Addressed the latest Greptile P1 in commit 6bc648807:

- Added the missing return after handle_cert_sign missing-flag validation.
- Added regression coverage so a mocked output_error_message does not fall through into CSR loading.
- Included the same explicit-return cleanup for related cert/package validation branches that can otherwise continue under mocked error handlers.

Verification:
- python3 -m pytest tests/unit_test/tool/cert/cert_commands_test.py tests/unit_test/tool/package/package_commands_test.py -q - 404 passed
- git diff --check
- black --check
- isort --check-only
- flake8

chesterxgchen:
Addressed the latest Greptile request-zip read fallthrough comment in commit ea7448d2a:

- Added explicit return None in all _read_request_zip exception handlers so mocked error handlers cannot return a truthy request_meta after partial extraction.
- Added regression coverage for extraction/write failure under mocked output_error_message.

Verification:
- python3 -m pytest tests/unit_test/tool/cert/cert_commands_test.py -q - 191 passed
- git diff --check HEAD~1..HEAD
- black --check
- isort --check-only
- flake8

chesterxgchen:
Addressed the two latest Greptile mocked-error fallthrough comments in commits 5bb0b10fb and 147279103:

- _write_materialized_signed_files now returns False on write failure, and signed-zip packaging stops before building from missing materialized cert/rootCA files.
- _write_zip_nofollow now returns immediately after parent directory creation fails, avoiding duplicate OUTPUT_DIR_NOT_WRITABLE errors and later zip writes.
- Added regression tests for both paths, plus related incomplete --request-dir handling.

Verification:
- python3 -m pytest tests/unit_test/tool/cert/cert_commands_test.py tests/unit_test/tool/package/package_commands_test.py -q - 416 passed
- git diff --check
- black --check
- isort --check-only
- flake8

chesterxgchen:
Handled the latest Greptile findings in commit 855d5d68d:\n\n- Added explicit returns after terminal cert init errors so mocked error handlers do not fall through.\n- Made request-zip member size failures return a failed read instead of empty bytes.\n- Made invalid distributed package project paths return None and stop callers cleanly under mocked errors.\n- Added regression tests for the mocked-error paths.\n\nChecks run:\n- python3 -m pytest tests/unit_test/tool/cert/cert_commands_test.py tests/unit_test/tool/package/package_commands_test.py -q\n- /Users/chesterc/nvflare-env3/bin/python -m black --check nvflare/tool/cert/cert_commands.py nvflare/tool/package/package_commands.py tests/unit_test/tool/cert/cert_commands_test.py tests/unit_test/tool/package/package_commands_test.py\n- python3 -m isort --check-only nvflare/tool/cert/cert_commands.py nvflare/tool/package/package_commands.py tests/unit_test/tool/cert/cert_commands_test.py tests/unit_test/tool/package/package_commands_test.py\n- python3 -m flake8 nvflare/tool/cert/cert_commands.py nvflare/tool/package/package_commands.py tests/unit_test/tool/cert/cert_commands_test.py tests/unit_test/tool/package/package_commands_test.py\n- git diff --check

IsaacYangSLA:
The document does not have sample yaml for server while the sample yamls are shown for client/admin.  We know the server can have more configurations, such as host_names, connection_security and others.  Will those additional information be accepted in `nvflare cert request server` and will the `nvflare cert package` be consistent with this additional information?

chesterxgchen:
> The document does not have sample yaml for server while the sample yamls are shown for client/admin. We know the server can have more configurations, such as host_names, connection_security and others. Will those additional information be accepted in `nvflare cert request server` and will the `nvflare cert package` be consistent with this additional information?

Yes. I should add more config info in the examples.

IsaacYangSLA:
> > The document does not have sample yaml for server while the sample yamls are shown for client/admin. We know the server can have more configurations, such as host_names, connection_security and others. Will those additional information be accepted in `nvflare cert request server` and will the `nvflare cert package` be consistent with this additional information?
> 
> Yes. I should add more config info in the examples.

What concerns me is the consistency on the communication protocol between server and client.  I am trying to find out how to enforce it in this PR.

chesterxgchen:
> > > The document does not have sample yaml for server while the sample yamls are shown for client/admin. We know the server can have more configurations, such as host_names, connection_security and others. Will those additional information be accepted in `nvflare cert request server` and will the `nvflare cert package` be consistent with this additional information?
> > 
> > 
> > Yes. I should add more config info in the examples.
> 
> What concerns me is the consistency on the communication protocol between server and client. I am trying to find out how to enforce it in this PR.

Good thinking. I need change my design and workflow to make sure this consistency is ensured.

chesterxgchen:
@greptileai

chesterxgchen:
@greptileai

docs/user_guide/nvflare_cli/cert_command.rst
```diff
@@ -1,170 +1,199 @@
...
-- ``--org``: optional organization name for the CA certificate.
-- ``--valid-days``: validity period for the root CA certificate in days. Default: ``3650``.
-- ``--force``: overwrite existing CA files after backing them up.
-- ``--schema``: print the JSON schema for this command and exit.
+- ``./ca/rootCA.pem``: root CA certificate
+- ``./ca/rootCA.key``: root CA private key; keep this secret
+- ``./ca/ca.json``: CA metadata
 
-This creates:
+The Project Admin also creates a project profile for approvals:
```
IsaacYangSLA: Where is this file located?  The cert init does not point to this file.

docs/user_guide/nvflare_cli/cert_command.rst
```diff
@@ -1,170 +1,199 @@
...
+   scheme: grpc
+   connection_security: tls
 
-Generate a Private Key and CSR
-==============================
+Common ``init`` options:
 
-Run ``nvflare cert csr`` on the participant side to generate a local private
-key and CSR.
+- ``--project``: project name. Required. Must match the project profile
```
IsaacYangSLA: How the code know the project profile yaml file?  If users run `nvflare cert init --project test`, will the code search all files with .yaml and locate the file.  It's not clear in this user guide.

docs/user_guide/nvflare_cli/cert_command.rst
```diff
@@ -1,170 +1,199 @@
...
-Single-site YAML example:
+User request:
 
 .. code-block:: shell
 
-   nvflare cert csr --project-file ./site.yaml -o ./csr
+   nvflare cert request --participant alice.yaml
 
-Common options:
+A participant definition contains the project name and one participant entry:
```
IsaacYangSLA: At this moment, how does the person run `cert request` know the name `hospital_federation` and the server section?  The information must be in the hospital-a.yaml file.

docs/user_guide/nvflare_cli/cert_command.rst
```diff
@@ -1,170 +1,199 @@
...
+.. code-block:: yaml
 
-The private key remains local. Only the ``.csr`` file is sent to the Project
-Admin.
+   name: hospital_federation
 
-The ``-t`` value in ``cert csr`` is the site admin's proposed type for the CSR.
-The Project Admin must either explicitly accept it when signing or explicitly
-override it.
+   participants:
```
IsaacYangSLA, line 140: Does this key `participants` mean we can have a list of participants?  What if this file has hospital-a and hospital-b, but run `cert request --participant hospital-a.yaml`?

docs/user_guide/nvflare_cli/cert_command.rst
```diff
@@ -1,170 +1,199 @@
...
-   nvflare cert csr -n hospital-1 -t client -o ./csr
+   nvflare cert request --participant hospital-a.yaml
 
-Server example:
+Server request:
 
 .. code-block:: shell
 
-   nvflare cert csr -n fl-server -t server -o ./csr
+   nvflare cert request --participant server.yaml
```
IsaacYangSLA, line 94: What's the format of server.yaml?  The following section does not have that.  And since the user run `cert request --participant server.yaml` may not talk to those who run `cert request --participant hospital-a.yaml` it could easily cause inconsistency.  But if we require them to coordinate, it seems complicating user experiences.

docs/user_guide/nvflare_cli/cert_command.rst
```diff
@@ -1,170 +1,199 @@
...
-trusting the site admin's requested type from the CSR. When they sign with
-``-t/--type``, they are explicitly overriding that requested type.
+.. code-block:: shell
 
-This means end users should not self-generate role-bearing CSRs and send them
-directly for signing. The trust model assumes the CSR passed through the site
-admin workflow before it reaches the Project Admin.
+   nvflare cert approve hospital-a.request.zip \
+       --ca-dir ./ca \
+       --profile project_profile.yaml \
```
IsaacYangSLA, line 243: Is this the yaml [here](https://github.com/NVIDIA/NVFlare/pull/4481/changes#diff-90268d5b06be2e9a46b0c98b7b634ca8e8e7ce1b7acb35708e7342c41386c1d0R41)?

docs/user_guide/nvflare_cli/cert_command.rst
chesterxgchen: No, this in the location where the project admin can keep. no particular folder location as it just contains three fields

name: fed_project
scheme: grpc
connection_security: tls

docs/user_guide/nvflare_cli/cert_command.rst
chesterxgchen: good point, I probably should change this 

nvflare cert init --project-profile  project-profile.yaml

docs/user_guide/nvflare_cli/cert_command.rst
chesterxgchen: the server.yaml is the same format the project.yaml format (just the server portion of it) 
```
name: fed_project
description: FL server for the example federation

participants:
  - name: server.example.com
    type: server
    org: nvidia
    fed_learn_port: 8002
    admin_port: 8003
    # Optional: add default_host or host_names when the server certificate
    # must cover DNS names/IPs beyond the participant name.
    # default_host: server-public.example.com
    # host_names:
    #   - server-public.example.com
    #   - 10.0.0.10
```

docs/user_guide/nvflare_cli/cert_command.rst
chesterxgchen: Project Admin need to tell them where the server URI is out of band.  I will need to make iit clear in the documents


```
name: fed_project
description: Site 1 client

participants:
  - name: site-1
    type: client
    org: org1
    server:
      host: server.example.com
      fed_learn_port: 8002
      admin_port: 8003
```

docs/user_guide/nvflare_cli/cert_command.rst
chesterxgchen: actually, we just list one.  Try not change the syntax for for the project.yaml. 
but we want to limit one. 

I will make it clear in documentation

docs/design/diloco_nvflare_analysis.md
YuanTingHsieh, line 1: should this file be included?

docs/user_guide/nvflare_cli/cert_command.rst
chesterxgchen: don't follow the question, where ? in which part --profile ? ( if I had used --profile-yaml, somewhere, thats a bug)

docs/design/cli_enhancement_extra_requirement.md
```diff
@@ -0,0 +1,622 @@
...
+code.
+
+
+## Priority Summary
+
+P0:
+- Concrete `--schema` output format.
+
+P1:
+- `job submit --idempotency-key`
```
YuanTingHsieh: Is --idempotency-key implemented in this PR, or is it only a future design requirement? If only future-facing, the docs should mark it clearly as proposed/not yet available.

docs/design/nvflare_agent_foundation_implementation_plan.md
YuanTingHsieh, line 1: These two agent-foundation docs seem broader than the distributed provisioning changes in this PR. Are they intended to be part of this PR’s deliverable? If not, could they move to a separate design PR? They introduce a large future CLI/agent roadmap that is hard to review together with cert/package behavior.

docs/design/nvflare_agent_foundation.md
YuanTingHsieh, line 1: These two agent-foundation docs seem broader than the distributed provisioning changes in this PR. Are they intended to be part of this PR’s deliverable? If not, could they move to a separate design PR? They introduce a large future CLI/agent roadmap that is hard to review together with cert/package behavior.

docs/user_guide/nvflare_cli/cert_command.rst
YuanTingHsieh: The docs say “The Project Admin also creates a project profile for approvals” under cert init. I initially read this as cert init generating project_profile.yaml, but the implementation only writes rootCA.pem, rootCA.key, and ca.json. Could we clarify that project_profile.yaml is manually created, or consider having cert init generate a default profile/template?

docs/user_guide/nvflare_cli/cert_command.rst
YuanTingHsieh: this might need to explicitly spell it out, not obvious, and project.yml can contain all the builders as well

docs/user_guide/nvflare_cli/cert_command.rst
```diff
@@ -1,170 +1,199 @@
...
-   nvflare cert sign --schema
+   nvflare cert approve hospital-a.request.zip --ca-dir ./ca --profile project_profile.yaml
 
-The top-level CLI also supports JSON output mode:
+Requester:
 
 .. code-block:: shell
 
-   nvflare cert init --project my-fl-project -o ./ca --format json
+   nvflare package hospital-a.signed.zip --confirm-rootca
```
YuanTingHsieh, line 281: should this section be in the package_command.rst?
do users know where to put the rootca for each site?

docs/user_guide/nvflare_cli/package_command.rst
```diff
@@ -1,130 +1,184 @@
...
-site stores its certificate material.
+For a site request created in ``./hospital-a``:
 
-From a Working Directory
-========================
+.. code-block:: shell
 
-Directory mode is the simplest path for a single participant. Put the private
-key, signed certificate, and ``rootCA.pem`` in one directory:
+   nvflare package hospital-a.signed.zip --confirm-rootca
```
YuanTingHsieh: Should signed-zip packaging require either --confirm-rootca or --expected-rootca-fingerprint? Currently it only warns and proceeds, which allows users to build a startup kit without verifying the out-of-band root CA fingerprint.

docs/user_guide/nvflare_cli/cert_command.rst
```diff
@@ -1,170 +1,199 @@
...
+   hospital-a.signed.zip
 
-Trust Model
-===========
+Return the signed zip to the requester.
 
-The intended workflow is:
+The command output includes ``rootca_fingerprint_sha256``. Share that value
+with the requester through a trusted out-of-band channel so they can verify the
+signed zip root CA during ``nvflare package``.
```
YuanTingHsieh: Could we clarify here that rootCA.pem is included inside the returned *.signed.zip, and that only rootca_fingerprint_sha256 should be shared out-of-band? As written, it is easy to wonder whether the requester also needs to receive/place a separate rootCA.pem before running nvflare package.

docs/user_guide/nvflare_cli/package_command.rst
```diff
@@ -1,130 +1,184 @@
...
+
+.. code-block:: text
+
+   workspace/hospital_federation/prod_00/hospital-a/
+
+****************************
+Connection Configuration
+****************************
+
+The package command does not require an endpoint argument in the distributed
```
YuanTingHsieh, line 99: Does nvflare package compare the local request-folder site.yaml connection fields against the signed zip site.yaml, or only identity fields? If a requester edits the local site.yaml after approval, for example changing the server host/ports or other package-time fields, will package use those edited local values? If yes, is that intended and should it be documented?

examples/advanced/distributed_provision/site-3.yaml
```diff
@@ -0,0 +1,11 @@
...
+description: Dynamically added client site
+
+participants:
+  - name: site-3
+    type: client
+    org: org3
+    server:
+      host: server.example.com
+      fed_learn_port: 8002
+      admin_port: 8003
```
YuanTingHsieh: If no builders are specified, package uses default startup-kit builders.
If builders are specified, they must be in the local participant YAML before cert request.
They are not sent to or approved by the Project Admin.

^ can we add this clarification in the doc/design?

docs/user_guide/nvflare_cli/distributed_provisioning.rst
```diff
@@ -1,576 +1,482 @@
...
+########################
 
-Step 1 is done once per federation. Each new participant repeats steps 2–7 independently.
+Distributed provisioning lets each participant create and keep its own private
+key. The Project Admin signs certificate requests, but never receives any
+participant private key.
 
-Step 1 — Project Admin: Initialize the Root CA (once per federation)
-=====================================================================
+Use this workflow when sites, servers, or users should be provisioned
```
YuanTingHsieh, line 11: Could we make this trust boundary explicit? `builders:` are not included in the request or signed zips and are not approved by the Project Admin; they remain local package-time behavior. This also helps explain why distributed provisioning does not automatically match centralized project.yaml builders or support centralized-builder features like HE/CC.

docs/user_guide/nvflare_cli/cert_command.rst
chesterxgchen: The server info is communicated out of band a head of time

docs/design/diloco_nvflare_analysis.md
chesterxgchen: accident. will remove

docs/design/cli_enhancement_extra_requirement.md
chesterxgchen: sorry this is extra requirements for CLI, I haven't implemented, this is for next PR.  I need to remove this doc.

docs/design/nvflare_agent_foundation_implementation_plan.md
chesterxgchen: accidental check-in

docs/design/nvflare_agent_foundation.md
chesterxgchen: removed

docs/user_guide/nvflare_cli/package_command.rst
chesterxgchen: --confirm-rootca is optional 
if you don't include it. it just print out the fingerprint. 

--confirm-rootca is for manual confirmation ( assuming you have the fingreprint out of band), it prompt you y/N

--expected-rootca-fingerprint will auto compares if you have the fingerprint out of band and want to auto confirmation. 
It won't stop you to build. just extra level information

docs/user_guide/nvflare_cli/cert_command.rst
chesterxgchen: yes. rootCA.perm is included as part of signed zip. I will make this clear

examples/advanced/distributed_provision/site-3.yaml
chesterxgchen: we had an illustration for this at the same example in the later page for custom builder. we can make this more explicit

docs/user_guide/nvflare_cli/distributed_provisioning.rst
chesterxgchen: Will do

examples/advanced/distributed_provision/site-3.yaml
chesterxgchen: I am moving the server back to the project admin's project profile to avoid this conflict

docs/user_guide/nvflare_cli/package_command.rst
chesterxgchen: we signed the server info in the new impl,  in next commit

nvflare/lighter/impl/cert.py
```diff
@@ -329,10 +329,12 @@ def _generate_cert(
         ca=False,
         role=None,
         server: Participant = None,
+        server_default_host=None,
+        server_additional_hosts=None,
+        not_valid_before=None,
```
nvidianz: Not sure why we need those 2 arguments. valid_days should be enough. In our use cases, not_valid_before should always be now.

nvflare/lighter/utils.py
```diff
@@ -46,6 +46,21 @@ def _host_to_subject_alt_name(host: str):
...
+
+    if server_default_host:
+        sans = [_host_to_subject_alt_name(server_default_host)]
+        if server_additional_hosts:
+            for h in server_additional_hosts:
+                if h != server_default_host:
+                    sans.append(_host_to_subject_alt_name(h))
+        return sans
+
+    return [x509.DNSName(fallback_subject_name)]
```
nvidianz, line 63: What happens if fallback_subject_name is None?

docs/user_guide/nvflare_cli/cert_command.rst
```diff
@@ -1,170 +1,277 @@
...
-Common options:
+The profile captures three things the Project Admin owns:
 
-- ``--project``: project name. Used as the root CA certificate CN.
-- ``-o, --output-dir``: directory where CA files are written.
-- ``--org``: optional organization name for the CA certificate.
-- ``--valid-days``: validity period for the root CA certificate in days. Default: ``3650``.
-- ``--force``: overwrite existing CA files after backing them up.
-- ``--schema``: print the JSON schema for this command and exit.
+- **name**: the federation name; every participant definition must use this
```
YuanTingHsieh: is using "federation name" clearer or "project name"?

docs/user_guide/nvflare_cli/cert_command.rst
```diff
@@ -1,170 +1,277 @@
...
+.. code-block:: yaml
 
-   nvflare cert init --project my-fl-project -o ./ca
+   name: hospital_federation
+   scheme: grpc
+   connection_security: tls
+   server:
+     host: server1.hospital-central.org
+     fed_learn_port: 8002
+     admin_port: 8003
```
YuanTingHsieh: i think we support just specifying one port, should we unify our story regarding the user experience?

 what do we expect them to do in 2.8, specify 1 port or 2 ports?

docs/user_guide/nvflare_cli/cert_command.rst
```diff
@@ -1,170 +1,277 @@
...
+   nvflare cert init --profile project_profile.yaml -o ./ca
 
 This creates:
 
-- ``rootCA.pem``
-- ``rootCA.key``
-- ``ca.json``
+- ``./ca/rootCA.pem``: root CA certificate
+- ``./ca/rootCA.key``: root CA private key; keep this secret
+- ``./ca/ca.json``: CA metadata
```
YuanTingHsieh: what are the schema of this metadata?
free form? or a dict? and is json better or yaml

docs/user_guide/nvflare_cli/cert_command.rst
```diff
@@ -1,170 +1,277 @@
...
+*Server* — includes the server's own listen ports:
 
-The ``-t`` value in ``cert csr`` is the site admin's proposed type for the CSR.
-The Project Admin must either explicitly accept it when signing or explicitly
-override it.
+.. code-block:: yaml
 
-Sign a CSR
-==========
+   name: hospital_federation
```
YuanTingHsieh, line 115: just curious, do we verify if this name match the one in rootCA?

docs/user_guide/nvflare_cli/cert_command.rst
```diff
@@ -1,170 +1,277 @@
...
 
-Run ``nvflare cert sign`` as the Project Admin to sign a CSR with the project
-root CA:
+   participants:
+     - name: server.example.com
+       type: server
+       org: nvidia
+       fed_learn_port: 8002
+       admin_port: 8003
```
YuanTingHsieh: same question regarding ports

docs/user_guide/nvflare_cli/cert_command.rst
```diff
@@ -1,170 +1,277 @@
...
 
-Trust Model
-===========
+   - **Project name** (``name:``): every participant definition must use the
+     exact project name from the Project Admin's ``project_profile.yaml``.
 
-The intended workflow is:
+   Client and user definitions do not include a server endpoint. The Project
+   Admin embeds the approved server host, ports, scheme, and connection
+   security from ``project_profile.yaml`` into every signed zip during
```
YuanTingHsieh, line 170: This "admin" means "project admin"? or the admin user of startup kits.

docs/user_guide/nvflare_cli/cert_command.rst
```diff
@@ -1,170 +1,277 @@
...
+.. code-block:: text
 
-This means end users should not self-generate role-bearing CSRs and send them
-directly for signing. The trust model assumes the CSR passed through the site
-admin workflow before it reaches the Project Admin.
+   hospital-a/
+     hospital-a.key
+     hospital-a.csr
+     site.yaml
+     request.json
```
YuanTingHsieh, line 182: do users need to write request.json?

docs/user_guide/nvflare_cli/package_command.rst
```diff
@@ -1,130 +1,195 @@
...
+
+.. code-block:: text
+
+   <workspace>/<project-name>/prod_NN/<identity>/
+
+For example:
+
+.. code-block:: text
+
+   workspace/hospital_federation/prod_00/hospital-a/
```
YuanTingHsieh, line 88: if user package more than once, will this prod_xx number increase?

docs/design/distributed_provisioning.md
```diff
@@ -1,72 +1,1085 @@
...
+
+### User Certificate Workflow
+
+User certificates use a participant definition file that contains the certificate role:
+
+```bash
+nvflare cert request --participant alice.yaml
+nvflare cert approve alice@hospital-alpha.org.request.zip --ca-dir ./ca --profile project_profile.yaml
+nvflare package alice@hospital-alpha.org.signed.zip --confirm-rootca
+```
```
YuanTingHsieh, line 143: this looks like all these commands are run on the same site

nvflare/tool/package/package_commands.py
```diff
@@ -423,327 +542,1748 @@ def _handle_package_yaml_mode(args, scheme, host, port):
...
+
+    is_server = kit_type == "server"
+    server_props = {
+        PropKey.FED_LEARN_PORT: port,
+        PropKey.ADMIN_PORT: admin_port,
+    }
+
+    project = Project(name=project_name, description="", props=project_props)
+    participant_props = dict(participant_props or {})
+    if is_server:
```
YuanTingHsieh, line 642: For server signed-zip packaging, only the signed ports are copied here; the server participant name is still used by StaticFileBuilder for fed_server.json target/sp_end_point. That means signed.json['server']['host'] is honored for client/admin kits but ignored for server kits when the profile host differs from the participant name, despite the docs saying startup-kit host/port come from the signed server block. Please add a regression with signed host != participant name and either propagate the host or document/enforce that they must match.

docs/design/multistudy_cli.md
```diff
@@ -815,28 +815,28 @@ Together, the two capabilities make a **fully decentralized federation lifecycle
 A new site joins and sets up its own study without coupling study management to provisioning:
 
 ```
-# Step 1 — Site Admin generates a CSR (once per participant)
-nvflare cert csr -n hospital-1 -t client -o ./csr
+# Step 1 — Site Admin creates a request zip (private key stays local)
+nvflare cert request site hospital-1 --org org_a --project example_project
```
YuanTingHsieh: Distributed provisioning commands are stale

These examples do not match the new public CLI. cert request no longer accepts the positional 'site/user' forms, cert approve requires --profile, and nvflare package has no -e endpoint option in signed-zip mode. A reader following this design doc will hit argparse errors instead of the documented flow; this section should use '--participant <yaml>', '--profile <project_profile.yaml>', and package without endpoint override.

tests/integration_test/distributed_provisioning_test.py
```diff
@@ -105,73 +157,57 @@ def _capture(data, exit_code=0, status="ok"):
...
-        )
+        result = _run_package(signed_zip=signed_zips[name], request_dir=request_dirs[name], workspace=ws)
         kit_dirs[name] = result["output_dir"]
 
     return kit_dirs
 
 
 class TestDistributedProvisioningWorkflow:
-    """Full distributed provisioning workflow: cert init → csr → sign → package."""
+    """Full distributed provisioning workflow: cert init → request → approve → package."""
```
YuanTingHsieh, line 210: handle_cert_init now requires args.profile and reads the project name from that file, but the integration workflow still passes project=... and creates project_profile.yaml afterward. This will exit with missing --profile before creating CA files; the same stale call appears at lines 218, 239, and 613.

docs/user_guide/nvflare_cli/cert_command.rst
YuanTingHsieh: or maybe should say something like:

ca.json: generated by nvflare cert init; internal CA metadata used by nvflare cert approve to verify that the CA belongs to the same project/federation. Do not edit manually.

docs/user_guide/nvflare_cli/cert_command.rst
YuanTingHsieh: suggest something like:

request.json: generated by nvflare cert request; local/request-zip metadata used by cert approve and package to match the request, CSR, local private key, and signed zip. Users normally do not edit it; send only the .request.zip to the Project Admin.

docs/user_guide/nvflare_cli/cert_command.rst
chesterxgchen: they should have option to specify, but we can only need require one port.

docs/design/distributed_provisioning.md
```diff
@@ -1,72 +1,1085 @@
...
-| 1 | Project Admin | `nvflare cert init --project my-project -o ./ca` *(one-time per federation)* |
-| 2 | Site Admin | `nvflare cert csr -n hospital-1 -t client -o ./csr` |
-| 3 | Site Admin | Send `hospital-1.csr` to Project Admin (email, file share, etc.) |
-| 4 | Project Admin | `nvflare cert sign -r hospital-1.csr -c ./ca -o ./signed/hospital-1 --accept-csr-role` |
-| 5 | Project Admin | Return `hospital-1.crt` + `rootCA.pem` to site |
-| 6 | Site Admin | `nvflare package -e grpc://server:8002 --dir ./csr` *(type derived from signed cert)* |
-| 7 | Site Admin | `cd hospital-1 && ./startup/start.sh` |
+| 0 | Server Admin | Provide the FL server endpoint (`server.host`, `server.fed_learn_port`, `server.admin_port`) to the Project Admin *(one-time per federation, before approvals)* |
+| 1 | Project Admin | Create `project_profile.yaml` with `name`, `scheme`, `connection_security`, and `server` endpoint info *(one-time per federation)* |
+| 1a | Project Admin | `nvflare cert init --profile project_profile.yaml -o ./ca` *(one-time per federation)* |
```
nvidianz, line 113: The project admin has to do provision for the server,

nvflare provision -p project.yml

will perform init functions. Do we really need this separate step?

docs/design/distributed_provisioning.md
```diff
@@ -1,72 +1,1085 @@
...
+
+- **Generated startup kits remain compatible with existing FLARE runtime.**
+  The output layout and runtime startup behavior match centrally provisioned kits.
+  Certificate generation reuses centralized provisioning logic to keep certificate
+  contents and validation behavior consistent. Server certificate SAN/default-host
+  behavior follows centralized provisioning conventions, including `localhost`
+  support for local/demo workflows.
+
+- **Root CA trust is explicit.**
+  Interactive packaging can use `--confirm-rootca`. Automation can use
```
nvidianz, line 93: It's probably more clear to call this option 

`--fingerprint`

docs/design/distributed_provisioning.md
chesterxgchen: project admin DO NOT provision the server. 
project admin DO NOT do the nvflare provision -p project.yml ==> not sure where do you read this. 

I read it wrong

docs/design/distributed_provisioning.md
```diff
@@ -1,72 +1,1085 @@
...
-| 5 | Project Admin | Return `hospital-1.crt` + `rootCA.pem` to site |
-| 6 | Site Admin | `nvflare package -e grpc://server:8002 --dir ./csr` *(type derived from signed cert)* |
-| 7 | Site Admin | `cd hospital-1 && ./startup/start.sh` |
+| 0 | Server Admin | Provide the FL server endpoint (`server.host`, `server.fed_learn_port`, `server.admin_port`) to the Project Admin *(one-time per federation, before approvals)* |
+| 1 | Project Admin | Create `project_profile.yaml` with `name`, `scheme`, `connection_security`, and `server` endpoint info *(one-time per federation)* |
+| 1a | Project Admin | `nvflare cert init --profile project_profile.yaml -o ./ca` *(one-time per federation)* |
+| 2 | Requester | Create participant definition file (e.g. `hospital-a.yaml`) with participant identity fields |
+| 3 | Requester | `nvflare cert request --participant hospital-a.yaml` |
+| 4 | Requester | Send generated `hospital-a/hospital-a.request.zip` to Project Admin |
+| 5 | Project Admin | `nvflare cert approve hospital-a.request.zip --ca-dir ./ca --profile project_profile.yaml` |
```
nvidianz, line 117: This command does the approval and provisioning all at once. Shall we come up with another name for this?

nvidianz (dismissed):
Just a few cosmetic suggestions, we don't have to change them right now.

docs/design/distributed_provisioning.md
chesterxgchen: no, the command does not provioninging, just pass the few server end point to the sites.

docs/user_guide/nvflare_cli/package_command.rst
chesterxgchen: yes, since we package each participant startup kit individually ( site, user etc), the number increases at the moment, I am debating if we should keep this the same or not

docs/user_guide/nvflare_cli/cert_command.rst
chesterxgchen: need to add doc to make it clear

docs/user_guide/nvflare_cli/cert_command.rst
chesterxgchen: yes

docs/user_guide/nvflare_cli/cert_command.rst
chesterxgchen: no user doesn't write request.json, these are generated.

docs/design/distributed_provisioning.md
chesterxgchen: no, we need to make it clear in the comments or docs

tests/integration_test/distributed_provisioning_test.py
chesterxgchen: good catch

docs/design/distributed_provisioning.md
chesterxgchen: which one is better ? `--confirm-rootca` or --fingerprint ?

nvidianz (approved):

PR #4483, opened 2026-04-27
Add Helm pod annotation support

pcnudde:
Adds pod annotation support to generated Helm charts and wires multicloud deployment config through to Helm as `podAnnotations` values. This lets internal AWS/Karpenter deployments mark NVFlare pods with `karpenter.sh/do-not-disrupt: "true"` from the multicloud YAML files.


IsaacYangSLA (approved):
LGTM.

PR #4495, opened 2026-04-29
Document server tempdir for tensor disk offload [doc]

pcnudde:
## What changed

- Added a server deployment note documenting how `enable_tensor_disk_offload=True` resolves temporary storage through Python `tempfile`.
- Clarified that server IT setup should point `TMPDIR` at an existing, writable, disk-backed mount and avoid RAM-backed `tmpfs`/`ramfs`.
- Added cross-references from large-model guidance and the recipe parameter warning back to the server setup note.


pcnudde:
CI is green on GitHub Actions after updating from main; only blossom-ci/review remain. This is documentation-only.

PR #4496, opened 2026-04-29
Fail server job process on streamed submit_update decode failure

pcnudde:
## Summary

- fail the server job process when streamed `submit_update` payload decoding fails inside a server job cell
- leave parent server and non-job cells on the existing exception path
- add unit coverage for the guarded job-cell exit and parent-server non-exit behavior

## Root Cause

For FLARE-2908, disk-full failures during streamed `submit_update` decoding can occur before the request reaches the job command layer. The streaming callback catches ordinary exceptions and the client can keep retrying, so the job does not fail cleanly.

## Validation

- `.venv/bin/python -m pytest tests/unit_test/fuel/f3/cellnet/test_pass_through_header.py -q`
- `.venv/bin/python -m black --check nvflare/fuel/f3/cellnet/cell.py tests/unit_test/fuel/f3/cellnet/test_pass_through_header.py`
- `.venv/bin/python -m isort --check-only nvflare/fuel/f3/cellnet/cell.py tests/unit_test/fuel/f3/cellnet/test_pass_through_header.py`
- `git diff --check`
- isolated GCP FLARE deployment with patched image:
  - normal `hello-numpy` full-update jobs completed
  - normal `hello-numpy-diff` job completed
  - forced streamed `submit_update` decode failure exited the server job process, failed the job, and left the parent server running

### DIFF
--- a/nvflare/fuel/f3/cellnet/cell.py
+++ b/nvflare/fuel/f3/cellnet/cell.py
@@ -56,17 +56,18 @@ def _is_stream_channel(channel: str) -> bool:
 
 
 def _is_server_job_cell(my_info) -> bool:
-    """Return True only for the server child cell that owns one job run.
+    """Return True only for server cells owned by one job run.
 
-    Parent server cell FQCN is "server"; server job cells are "server.<job_id>".
-    The fail-fast path must only exit the job process, never the parent server.
+    Parent server cell FQCN is "server"; server job cells start with
+    "server.<job_id>" and may have nested children like "server.<job_id>.cell_pipe".
+    The fail-fast path must only exit a job process, never the parent server.
     """
     fqcn = getattr(my_info, "fqcn", "")
     if not fqcn:
         return False
 
     path = FQCN.split(fqcn)
-    return len(path) == 2 and path[0] == FQCN.ROOT_SERVER
+    return len(path) >= 2 and path[0] == FQCN.ROOT_SERVER
 
 
 class SimpleWaiter:
--- a/nvflare/fuel/f3/cellnet/cell.py
+++ b/nvflare/fuel/f3/cellnet/cell.py
@@ -14,13 +14,16 @@
 
 import concurrent.futures
 import copy
+import os
 import threading
 import uuid
 from typing import Dict, List, Union
 
+from nvflare.apis.fl_constant import ServerCommandNames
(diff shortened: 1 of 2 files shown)


nvflare/fuel/f3/cellnet/cell.py
```diff
@@ -52,6 +55,20 @@ def _is_stream_channel(channel: str) -> bool:
...
+
+    Parent server cell FQCN is "server"; server job cells are "server.<job_id>".
+    The fail-fast path must only exit the job process, never the parent server.
+    """
+    fqcn = getattr(my_info, "fqcn", "")
+    if not fqcn:
+        return False
+
+    path = FQCN.split(fqcn)
+    return len(path) == 2 and path[0] == FQCN.ROOT_SERVER
```
nvidianz: The job may spawn another process (like in subprocess launcher) and we may use fqcn like

`server.job_id.cell_pipe`

If those are considered job cell, the condition should be >= 2.

nvflare/fuel/f3/cellnet/cell.py
```diff
@@ -52,6 +55,20 @@ def _is_stream_channel(channel: str) -> bool:
     return True
 
 
+def _is_server_job_cell(my_info) -> bool:
```
YuanTingHsieh, line 58: so this solves the server side "submit_update" decode failure, what about the client side "get_task" decode failure? do we have similar issue?

nvidianz (dismissed):

pcnudde:
CI is green after updating from main. @nvidianz could you please re-approve when you have a chance?

nvflare/fuel/f3/cellnet/cell.py
pcnudde: Updated in 5dabb280e: _is_server_job_cell now treats any FQCN starting with server.<job_id> as job-owned, including nested cells such as server.<job_id>.cell_pipe. Added unit coverage for the nested case and for non-matching topic/channel paths.

nvflare/fuel/f3/cellnet/cell.py
pcnudde: yes, but this one addresses the particular bug found during tensor offloading testing (which can only happen on the server). For the client we need to look into it separately as it is a different path.

PR #4500, opened 2026-04-30
Enhance CLI agent readiness workflows

chesterxgchen:
## Summary

This PR expands the NVFlare CLI for agent-friendly and automation-friendly workflows. It adds stable JSON contracts, scoped startup-kit selection, retry-safe job submission, bounded and study-aware job operations, blocking job wait, JSONL job monitor streaming, POC readiness/endpoint metadata, study submit-preflight metadata, recipe metadata discovery, durable client disable/enable workflows, and system lifecycle commands with structured output.

Current scope: **94 files changed**: **14 docs/design/user-guide files**, **34 implementation files**, and **46 test files**.

## Key Design Decisions

- **Per-command identity selection is optional and non-mutating.** Online commands use the active startup kit by default. If `--kit-id` or `--startup-kit` is provided, it overrides the active startup kit for that command only and does not change the human default startup kit.
- **Submit retry safety is token based.** `--submit-token` is scoped by `study + submitter + token`; it is not an auth credential and does not grant access.
- **No special `study=all` behavior.** Study selection is either omitted for default behavior or set to an explicit study name.
- **Streaming output uses JSONL, not multi-object JSON.** `job monitor --format jsonl` emits one complete JSON event per line; regular `--format json` remains a single-envelope contract.
- **Startup-kit registry commands are flattened under `nvflare config`.** The canonical commands are `nvflare config add/use/inspect/list/remove`; no nested `config kit` command is documented for this PR.
- **`config inspect` is read-only and human-readable.** It uses the same table shape as `config list` for the active startup kit, plus the config file path. JSON mode keeps the structured automation payload.
- **Human and machine output are separated.** Commands enhanced in this PR keep JSON stdout as a single envelope and avoid leaking raw dict/list output into human mode.
- **POC operations are local workspace operations.** `poc add-site` and `poc add-user` do not require an active project-admin startup kit because POC models local deployment setup, not production authorization.
- **POC port preflight is best effort.** It checks loopback availability only and reports `scope: "loopback"` plus a note instead of binding wildcard interfaces.
- **Recipe discovery is local installed-package discovery.** `recipe list/show` do not connect to a FLARE server. Human mode prints a short loading message before local import/inspection; JSON mode remains clean.
- **Recipe metadata is descriptive, not an absolute compatibility contract.** `privacy` means a privacy feature is enabled/declared by that recipe entry; lack of privacy metadata does not mean the underlying algorithm cannot be combined with PETs. `state_exchange` text identifies default transfer behavior when a recipe supports configurable transfer types.
- **Client disable/enable is durable policy; `remove-client` is not supported in the new `nvflare system` CLI surface.** `disable-client` persists a disabled-client flag and rejects future heartbeats; `enable-client` clears that flag so the client can rejoin.
- **`--schema` stays backward compatible.** Existing argparse-derived schema data remains, with added behavioral metadata for agent-facing commands.

### Startup Kit / Identity Selection

Adds optional per-command selectors for online commands:

```bash
nvflare job list --kit-id <id>
nvflare job submit -j <job_folder> --startup-kit <path>
nvflare study list --kit-id <id>
nvflare system status --startup-kit <path>
```

`--kit-id` and `--startup-kit` are not required. If neither is provided, the command uses normal startup-kit resolution, ending with the configured active startup kit. If either selector is provided, it takes precedence for that one command and does not mutate `startup_kits.active` in `~/.nvflare/config.conf`.

`nvflare config list --format json` and `nvflare config inspect --format json` provide stable startup-kit metadata for automation, including identity, role/org/project when available, certificate expiration status, and non-fatal findings for stale/unreadable kits.

`nvflare config inspect` human output uses the same table columns as `config list`, scoped to the active startup kit:

```text
active  id                status  identity          cert_role      path
--------------------------------------------------------------------------------------------------------------------------------
*       admin@nvidia.com  ok      admin@nvidia.com  project_admin  /private/tmp/nvflare/poc/example_project/prod_00/admin@nvidia.com

config_file: /Users/chesterc/.nvflare/config.conf
```

`nvflare config use` remains a human-facing convenience command that mutates global CLI state. JSON mode returns a structured `startup_kit` and `identity` payload plus a finding that automation should prefer `--kit-id` or `--startup-kit`.

Startup-kit registry commands are:

```bash
nvflare config add <id> <startup-kit-dir>
nvflare config use <id>
nvflare config inspect
nvflare config list
nvflare config remove <id>
```

### Job Submit / Retry Recovery

Adds optional retry-safe submit-token recovery:

```bash
nvflare job submit -j <job_folder> [--study <name>] [--submit-token <token>] --format json
nvflare job list --study <name> --submit-token <token> --format json
```

`--submit-token` is caller-generated and optional. If it is omitted, NVFlare does not auto-create a token: submit behavior remains unchanged, each submit creates a new job as before, and the server still tracks the accepted job through the normal job store/job history. No retry-safe submit-token record is created, so `job list --submit-token` recovery is available only when the original submit used a caller-provided token.

If a token is provided, submit-token records are scoped by `study + submitter + token`, persisted with job state, and checked against the canonical submitted job content. A retry with the same token and same content returns the existing accepted job; a retry with the same token and different content is rejected.

### Job Wait / Monitor

Adds blocking wait:

```bash
nvflare job wait <job_id> [--study <name>] [--timeout N] [--interval N] --format json
```

This blocks until terminal job state or timeout and returns one final structured envelope. Human mode now prints periodic status context, including the selected study, instead of only printing the initial connection line.

Adds newline-delimited streaming monitor output:

```bash
nvflare job monitor <job_id> [--study <name>] [--timeout N] [--interval N] --format jsonl
```

- Each stdout line is a complete JSON object and is flushed immediately for pipe consumers.
- Progress events use `terminal: false`.
- Final events use `event: "terminal"` and `terminal: true`.
- Timeout emits `status: "TIMEOUT"` and exits with code 3.
- Successful raw statuses such as `FINISHED_OK` normalize to `status: "COMPLETED"` while preserving the raw value as `job_status`.
- JSONL is scoped to `job monitor`; non-streaming commands reject it.

### Job Metadata / Download / Logs

`job meta` is study-aware and has improved human output. Human mode now renders grouped job details instead of raw key/value dumps, including identity, status, execution, deployment, schedule attempts, and storage details. JSON mode remains structured.

Job lookup failures for study-aware commands report the searched study and hint to retry with `--study <study_name>` when appropriate, instead of only saying the job does not exist.

`job download --format json` reports local artifact discovery under the JSON `data` object. Human output remains concise and prints the final download path. `--force` replaces an existing local download directory.

```json
{
  "data": {
    "download_path": "...",
    "path": "...",
    "artifact_discovery": "completed",
    "artifacts": {},
    "missing_artifacts": []
  }
}
```

`job logs` supports named studies and bounded output for automation:

```bash
nvflare job logs <job_id> [--study <name>] --site server|<client>|all --tail N --since <timestamp> --max-bytes N
```

`--study` defaults to `default`. For named-study jobs, callers must pass the same study used for submit/list; otherwise `JOB_NOT_FOUND` reports the searched study and hints to rerun with `--study <study_name>`.

Log retrieval now selects a single preferred log representation instead of returning duplicate formats:

- `--format json` requests `log.json` first and falls back to `log.txt` only when the JSON log is unavailable.
- Human mode requests `log.txt` first and falls back to `log.json`; JSON-log fallback is rendered as text for display.
- Bounds (`--tail`, `--since`, `--max-bytes`) are applied by the CLI after retrieval.

The log bounds are CLI-side bounds over server-side stored logs; the command does not execute shell commands on client machines.

### POC Prepare / Start / Stop / Add

`poc prepare` reports startup-kit transition metadata and port-preflight diagnostics in JSON mode without dumping agent diagnostics into human output.

`poc start` waits for readiness by default and reports actual endpoint metadata. Human output is concise and explains that the admin console is optional.

`poc stop` has an explicit automation contract: by default it waits for coordinated server shutdown and returns `status: stopped`; with `--no-wait`, it returns after requesting shutdown with `status: shutdown_initiated`.

`poc add-site` and `poc add-user` are explicit local workspace modification commands. They replace the overloaded `poc add` shape for the new CLI UX and do not require an active project-admin startup kit. Adding a second POC project-admin user is rejected because the generated POC workspace has a single project admin.

`poc start` endpoint metadata:

```json
{
  "server_address": "localhost:8002",
  "admin_address": "localhost:8003",
  "ready_timeout": 30,
  "port_conflict": false,
  "port_preflight": {},
  "warnings": [],
  "ready": true
}
```

`server_address` and `admin_address` are the source of truth for follow-on automation.

### Study List

`study list --format json` returns startup-kit metadata plus authenticated identity/study details from the server, including role/capability fields such as `can_submit_job`.

`can_submit_job` is derived from the server authorization service instead of hard-coded role checks in the CLI. Human output is rendered as readable identity/startup-kit text plus a study table instead of raw Python dict/list formatting.

Example response shape:

```json
{
  "startup_kit": {"source": "active", "id": "...", "path": "..."},
  "identity": {"name": "...", "org": "...", "role": "..."},
  "studies": ["..."],
  "study_details": [
    {
      "name": "...",
      "role": "...",
      "capabilities": {"submit_job": true},
      "can_submit_job": true
    }
  ]
}
```

This is membership-level submit preflight. Hidden or unmapped studies are omitted; full future policy introspection remains out of scope.

### Recipe Discovery

Adds filterable/queryable recipe metadata:

```bash
nvflare recipe list [--framework <framework>] [--filter key=value] --format json
nvflare recipe show <name> --format json
```

Supported filter keys are `framework`, `privacy`, `algorithm`, `aggregation`, and `state_exchange`. The user guide and design doc now list the available values for each metadata key, show which frameworks each value applies to, and include examples for common combinations.

Human mode now prints progress while local recipe metadata is loaded:

```text
Loading installed recipe catalog...
Loading installed recipe metadata for 'fedavg'...
```

The `recipe show` human summary distinguishes enabled privacy features from metadata declarations and clarifies default/configurable state exchange. For example:

```text
state_exchange: full_model (default; params_transfer_type=FULL, supports FULL or DIFF)
privacy: none enabled by default
privacy compatibility: not declared in recipe metadata
parameters: 28 available; run 'nvflare recipe show fedavg --format json'
```

JSON output remains the full structured metadata contract, including algorithm, aggregation, state exchange, framework support, privacy compatibility, constructor parameters, optional dependencies, and template references.

### CLI Schema Contract Metadata

Agent-facing commands add behavior metadata to `--schema` output:

```json
{
  "output_modes": ["json"],
  "streaming": false,
  "mutating": false,
  "idempotent": true,
  "retry_token": {"supported": false}
}
```

Initial coverage includes `job submit/list/meta/wait/monitor/download/logs`, `config use/inspect/list`, `recipe list/show`, and `study list`.

For retry-safe submit commands:

```json
{
  "retry_token": {
    "supported": true,
    "flag": "--submit-token",
    "scope": "study + submitter + token",
    "retry_safe_when_present": true,
    "effect": "Deduplicates retries for identical submitted job content in the same study by the same submitter; different content with the same token is rejected."
  }
}
```

For streaming monitor:

```json
{
  "output_modes": ["json", "jsonl"],
  "streaming": true,
  "mutating": false,
  "idempotent": true
}
```

### System Lifecycle

Adds structured lifecycle behavior:

```bash
nvflare system shutdown <server|client|all> [--no-wait] --format json
nvflare system restart <server|client|all> [--no-wait] --format json
nvflare system disable-client <client_name> --format json
nvflare system enable-client <client_name> --format json
```

`shutdown` and `restart` support wait/no-wait behavior for automation. No-wait machine statuses use underscore values: `shutdown_initiated` and `restart_initiated`.

`disable-client` persists a disabled-client flag and rejects future heartbeats until `enable-client` clears it. `enable-client` removes the disabled flag so the client can rejoin on its next heartbeat/session attempt. `remove-client` is intentionally not exposed as a supported `nvflare system` CLI command; durable exclusion should use `disable-client`.

## Major Bug Fixes and Hardening

This section lists fixes relative to the previous PR/baseline. It intentionally does not enumerate review-round fixes for features introduced inside this PR.

- Fixed `job monitor` not being study-aware; it now supports `--study` and uses the selected study session.
- Fixed `job logs` not being study-aware; it now supports `--study` and reports the searched study on `JOB_NOT_FOUND`.
- Replaced the misleading non-durable `system remove-client` workflow in the new CLI surface with durable `disable-client` / `enable-client` semantics.
- Fixed user-submitted jobs being able to carry `from_hub_site` metadata. User submissions now strip `from_hub_site`, so only the server-side `HubAppDeployer` can mark HUB-created jobs and trigger HUB-specific signature-verification behavior.
- Hardened cert request and signed-zip validation to reject private key material embedded in zip members, including PEM content hidden in `.pem` files.
- Hardened certificate generation helpers to reject duplicate or reserved X.509 extensions and validate CA `not_valid_before` in addition to expiration.
- Fixed invalid CLI connection-timeout values to fall back to the default with a warning instead of failing unexpectedly.

## Docs / Tests

Docs updated for:

- `nvflare config` flattened commands and startup-kit discovery.
- `nvflare config inspect` naming and table-style human output.
- Job submit tokens, job wait, job monitor JSONL, job metadata, job download artifacts, bounded/study-aware logs, and study-aware command errors.
- POC readiness, endpoint metadata, shutdown behavior, and explicit `poc add-site` / `poc add-user` commands.
- Recipe list/show metadata, filter keys/values, framework applicability, privacy metadata semantics, loading messages, and state-exchange defaults.
- System shutdown/restart and disable/enable-client semantics.
- CLI schema metadata and the remaining agent-readiness backlog.

Recent focused checks during review rounds include:

```bash
pytest tests/unit_test/tool/recipe_cli_test.py
pytest tests/unit_test/tool/system/system_shutdown_test.py tests/unit_test/tool/system/system_restart_test.py
python -m nvflare.cli recipe list --format json
python -m py_compile nvflare/tool/recipe/recipe_cli.py tests/unit_test/tool/recipe_cli_test.py
black --check nvflare/tool/recipe/recipe_cli.py tests/unit_test/tool/recipe_cli_test.py
black --check nvflare/tool/poc/poc_commands.py
git diff --check
```

Earlier implementation/review rounds also ran focused suites for job, POC, study, config, system, CLI output/session/schema, startup-kit, cert/package, and submit-token paths.


nvflare/private/fed/server/client_manager.py
```diff
@@ -45,12 +47,96 @@ def __init__(self, project_name=None, min_num_clients=2, max_num_clients=10):
...
+            return
+        try:
+            with open(self.disabled_clients_file) as f:
+                data = json.load(f)
+            clients = data.get("disabled_clients", []) if isinstance(data, dict) else data
+            if not isinstance(clients, list):
+                raise ValueError("disabled_clients must be a list")
+            with self.lock:
+                self.disabled_clients = {str(client_name) for client_name in clients if client_name}
+        except Exception as ex:
```
holgerroth, line 77: `disable_client` is a durable exclusion policy and this file is the *only*
  artifact that persists it across restarts. Swallowing the exception and
  booting with an empty `disabled_clients` set silently re-admits every
  previously disabled client on a corrupted/unreadable file — exactly the
  opposite of what an admin who issued `disable-client` was promised.

  At minimum, escalate to `critical`, surface the degraded state via
  `nvflare system status`, and require an explicit admin action before any
  new client registration is accepted. Better: refuse to start.

  ```suggestion
          except Exception as ex:
              self.logger.critical(
                  f"failed to load disabled clients from {self.disabled_clients_file}: {ex}; "
                  "refusing to start to preserve disable-client policy"
              )
              raise
  ```

nvflare/private/fed/server/client_manager.py
```diff
@@ -45,12 +47,96 @@ def __init__(self, project_name=None, min_num_clients=2, max_num_clients=10):
...
         self.name_to_clients = dict()  # name => Client
+        self.disabled_clients = set()
+        self.disabled_clients_file = None
         self.cred_keeper = CredKeeper()
         self.lock = threading.Lock()
         self.num_relays = 0
 
         self.logger = get_obj_logger(self)
 
+    def set_disabled_clients_file(self, file_path: str):
```
holgerroth, line 60: confirm the caller (likely in server_engine.py or the server boot path) propagates this exception rather
   than catching it again — otherwise the fail-closed guarantee is broken one level up.

nvflare/tool/job/job_cli.py
```diff
@@ -1160,9 +1379,19 @@ def cmd_job_download(cmd_args):
         output_error("CONNECTION_FAILED", exit_code=2, detail=str(e))
         return
 
-    final_path = path or destination
-    print_human(f"Job result downloaded to: {final_path}")
-    output_ok({"job_id": cmd_args.job_id, "path": final_path})
+    path = str(path or destination)
+    download_path = _local_download_path(path)
+    artifacts, missing_artifacts = _discover_job_download_artifacts(download_path)
```
holgerroth: **`missing_artifacts` is misleading when no local inspection happened.**

`_discover_job_download_artifacts(None)` (and `_discover_job_download_artifacts(<path that isn't a real dir>)`) both fall through `_iter_files_under`'s `if not path or not os.path.isdir(path): return` short-circuit and return `({}, ["global_model", "metrics_summary", "client_logs"])`. The success envelope then promises agents that all expected artifacts are *missing* — when in fact we never even tried to look.

Three distinct conditions all collapse to the same JSON shape:

1. Local dir exists but is empty / lacks expected files → genuine "missing". OK.
2. Local dir doesn't exist on disk (e.g. `download_job_result` returned a stale or unreachable path) → we couldn't inspect, but envelope claims "missing".
3. `_local_download_path(path)` returns `None` because `_is_remote_download_location(path)` matched (today this is dead defensive code — `flare_api.py:397` always returns a local path — but the helper exists and the branch will fire if any future caller ever returns a URL or scheme).

Cases 2 and 3 misrepresent "couldn't inspect" as "verified absent". For an agent contract, that's the wrong direction: a downstream tool that retries or fails based on `missing_artifacts` will draw the wrong conclusion.

Suggested fix: skip artifact discovery when we don't have a real local directory to inspect, and omit the discovery fields rather than fabricating an empty/all-missing pair.

```suggestion
    path = str(path or destination)
    download_path = _local_download_path(path)
    print_human(f"Job result downloaded to: {download_path or path}")
    payload = {
        "job_id": cmd_args.job_id,
        "download_path": download_path,
        "path": download_path or path,
    }
    if download_path and os.path.isdir(download_path):
        artifacts, missing_artifacts = _discover_job_download_artifacts(download_path)
        payload["artifacts"] = artifacts
        payload["missing_artifacts"] = missing_artifacts
    output_ok(payload)
```

If you'd rather keep the keys always present for a stable schema, alternatives are:
- `"artifacts": null, "missing_artifacts": null` when discovery was skipped, and document the meaning in the agent contract.
- Add a `"discovery": "skipped" | "completed"` field so agents can disambiguate without inspecting nullness.

Either way, worth a unit test for the `download_path is None` and `download_path doesn't exist` cases — `tests/unit_test/tool/job/job_download_test.py` currently only covers the empty-real-dir variant (lines 11920-11931), which is condition 1 above.

nvflare/private/fed/server/client_manager.py
```diff
@@ -45,12 +47,96 @@ def __init__(self, project_name=None, min_num_clients=2, max_num_clients=10):
...
+        try:
+            with open(self.disabled_clients_file) as f:
+                data = json.load(f)
+            clients = data.get("disabled_clients", []) if isinstance(data, dict) else data
+            if not isinstance(clients, list):
+                raise ValueError("disabled_clients must be a list")
+            with self.lock:
+                self.disabled_clients = {str(client_name) for client_name in clients if client_name}
+        except Exception as ex:
+            self.logger.error(f"failed to load disabled clients from {self.disabled_clients_file}: {ex}")
```
holgerroth: **Fail closed when the disabled-clients persistence file fails to load.**

`disable_client` is a durable security control: once an admin disables a site, this JSON file is the **only** thing carrying that policy across server restart. Swallowing the exception and leaving `self.disabled_clients` as the empty set quietly re-admits every previously disabled client on the next heartbeat — the opposite of the contract.

Failure modes that hit this path silently:
- Truncated / partial write (server crash mid-`os.replace`).
- Hand-edit typo (operator triage, ansible templating bug).
- Disk / permission error.

In every case the only signal is one `error`-level log line; no startup banner, nothing surfaced via `nvflare system status`, no health check fails. And: this is the wrong direction for an *exclusion* control. Failing open is reasonable for availability/quota controls; for deny/disable, the safe direction is fail closed.

Three options, increasingly aggressive:

1. **Surface degraded state.** Write the corrupt file aside as `<file>.corrupt-<ts>`, escalate to `critical`, and add a `disabled_clients_state: "load_failed"` field to `nvflare system status` so operators can see the policy was lost.
2. **Reject all client registrations until an admin clears the state explicitly** (e.g. a `system reset-disabled-clients --confirm` command).
3. **Refuse to start.** Re-raise from `_load_disabled_clients` and let `set_disabled_clients_file`'s caller propagate. Admin sees a clear startup error and either restores from backup or repairs the file.

I'd lean (3) — smallest change, easiest to reason about, and aligned with the security promise of `disable-client`.

```suggestion
        except Exception as ex:
            self.logger.critical(
                f"failed to load disabled clients from {self.disabled_clients_file}: {ex}; "
                "refusing to start to preserve disable-client policy"
            )
            raise
```

If you adopt this, confirm `set_disabled_clients_file`'s caller (server engine boot) actually propagates the exception rather than catching it again — otherwise the guarantee is broken one level up.

Worth a unit test too: corrupt `disabled_clients.json` on disk → `ClientManager` init / `set_disabled_clients_file` raises, and no client registration is accepted afterwards.

nvflare/tool/job/job_cli.py
```diff
@@ -1658,20 +2242,47 @@ def cmd_job_monitor(cmd_args):
...
+    except NoConnection as e:
+        if jsonl_mode:
+            output_error("CONNECTION_FAILED", exit_code=2, detail=str(e))
+            return
+        raise
+    except (AuthenticationError, AuthorizationError) as e:
+        if jsonl_mode:
+            output_error("AUTH_FAILED", exit_code=2, detail=str(e))
+            return
         raise
```
holgerroth: **Non-jsonl path re-raises `NoConnection` / `AuthenticationError` / `AuthorizationError` instead of returning a structured envelope.**

In `--format json` mode (the agent contract for non-streaming monitor), the bare `raise` on lines 2249 and 2254 bubbles the raw exception out to the CLI dispatcher. Agents parsing for `{"status": "ERROR", "error": {"code": ...}}` get a Python traceback on stderr instead.

Compare with `cmd_job_wait` just below this function: same exception classes, but always converted to structured envelopes regardless of mode (`CONNECTION_FAILED` exit 2, `AUTH_FAILED` exit 2). The two commands have the same agent-facing contract; they shouldn't diverge here.

Practical impact: an agent watching a long-running job whose admin session times out, or whose network briefly drops, gets unparseable output and has to fall back to text scraping — exactly the regression the JSON envelope contract was added to prevent.

Fix — emit the structured envelope in both modes, since the only mode-dependent behavior here is whether to return or re-raise:

```suggestion
    except NoConnection as e:
        output_error("CONNECTION_FAILED", exit_code=2, detail=str(e))
        return
    except (AuthenticationError, AuthorizationError) as e:
        output_error("AUTH_FAILED", exit_code=2, detail=str(e))
        return
```

(`output_error` exits with the supplied `exit_code`, matching `cmd_job_wait`.)

Worth a unit test: `cmd_job_monitor --format json` against a session that raises `NoConnection` should produce a single `CONNECTION_FAILED` envelope on stdout, not a traceback. `tests/unit_test/tool/job/job_monitor_test.py` covers the jsonl path well but doesn't appear to cover this json-mode failure case.

nvflare/apis/utils/job_submit_token.py
```diff
@@ -0,0 +1,130 @@
...
+    for rel_path in rel_paths:
+        parts = rel_path.split("/", 1)
+        if len(parts) < 2:
+            return ""
+        top = parts[0]
+        if first is None:
+            first = top
+        elif first != top:
+            return ""
+    return f"{first}/" if first else ""
```
holgerroth, line 162: **Zip-only top-level prefix stripping creates hash asymmetry between zip and directory submissions.**

`_iter_zip_bytes` strips a single common top-level directory prefix via `_single_top_level_prefix`; `_iter_directory` doesn't. Same logical content packaged differently can hash differently, which trips `SUBMIT_TOKEN_CONFLICT` on a legitimate retry.

Concrete cases:

| First submit | Retry with same `--submit-token` | Result |
|---|---|---|
| zip `my_job.zip` (contains `my_job/config.json`, `my_job/app/x.py`) | directory at parent of `my_job/` (contains `my_job/config.json`, `my_job/app/x.py`) | zip hash strips `my_job/`, directory hash keeps it → conflict |
| zip with intentional content subdir (`data/config.json`, `data/app/x.py`) | directory containing the same `data/config.json`, `data/app/x.py` | zip hash strips `data/`, directory hash keeps it → conflict |
| zip `my_job.zip` (contents = same files at root) | directory `my_job/` (contents = same files at root) | both hash to root-level paths → no conflict ✓ |

The retry-safety story documented for `--submit-token` ("scope: study + submitter + token") implies content equivalence should be transport-independent. Today it isn't.

Two ways to resolve:

1. **Drop the prefix-strip and require canonical inputs.** Both transports must pass the content root (no wrapper directory). Simpler contract, but breaks the common case where users zip a job folder (`zip -r my_job.zip my_job/`) and expect that to be equivalent to submitting the `my_job/` directory.
2. **Symmetrize**: apply `_single_top_level_prefix` to the directory walker too. Risk: if `_iter_directory` is sometimes called with the *contents* of the job folder (no wrapper) and sometimes with a wrapper, you may strip a real content prefix and create a different kind of false equivalence.

I'd lean (1) with a docstring on `canonical_job_content_hash` documenting "the input must be the canonical content root, not a wrapper directory" — the contract is then explicit and the failure mode is loud (different hash on retry, surfaced as `SUBMIT_TOKEN_CONFLICT` with clear meaning) rather than load-bearing on a heuristic.

Either way:
- Add a unit test in `tests/unit_test/apis/utils/job_submit_token_test.py` for cross-transport equivalence (or explicitly document non-equivalence). Currently the test file doesn't appear to cover the case where the same logical content is presented as zip-bytes vs. directory-path.
- Worth a docstring on `canonical_job_content_hash` documenting what counts as "the same content", since `SUBMIT_TOKEN_CONFLICT` retry semantics depend on it directly.

Minor secondary concern in `_iter_zip_bytes`: `posixpath.normpath("a/../../etc/passwd")` returns `"../etc/passwd"`, which is caught by the `startswith("../")` guard. But `posixpath.normpath("a/b/../../../etc/passwd")` also returns `"../etc/passwd"` — caught. However something like a zip member literally named `"./etc/passwd"` normalizes to `"etc/passwd"` and is accepted. Since this is hash-only (no extraction), the impact is bounded to hash-equivalence of files claiming arbitrary names — but worth noting that the "safe path" guard isn't a fully-fledged zip-slip defense; if `canonical_job_content_hash` is ever reused by an extractor, the guard will need to be hardened.

chesterxgchen:
@greptileai

nvidianz (approved):
We have enough rounds on this PR.

PR #4493, opened 2026-04-29
Fix task cleanup on workflow finalization

pcnudde:
## Summary

Fixes FLARE-2909 by making workflow finalization release communicator-owned task state synchronously.

The previous `WFCommServer.finalize_run()` path called `cancel_all_tasks()` and then stopped the task monitor with `_all_done = True`. `cancel_all_tasks()` only marks standing tasks completed; the monitor is normally responsible for later removing those tasks from `_tasks` / `_client_task_map` and dropping task/client-task references. On abort/finalize, the monitor could be stopped before that drain pass, leaving task results reachable. For tensor disk offload, those task results can hold `LazyTensorDict` / `_LazyRef` objects, which prevents normal GC from running `_TempDirRef.__del__` and removing the offload temp directory.

This PR keeps the existing GC-based disk cleanup model and fixes the owner lifecycle instead of adding an explicit temp-dir registry.

## Changes

- Add synchronous task-resource release for `WFCommServer.finalize_run()`.
- Preserve `cancel_all_tasks()` semantics as mark-only; normal running paths still let the monitor drain tasks.
- Clear communicator-owned references from completed tasks and client tasks during finalization.
- Add a focused regression test proving finalization releases a lazy tensor ref and its temp directory through normal GC.

PR #4501, opened 2026-04-30
Clean up in-flight disk tensor downloads

pcnudde:
## What changed

This follow-up fixes the remaining FLARE-2909/NVBug #6127279 path that #4493 did not cover: workflow finalization while tensor disk-offload download is still in flight, before `download_tensors_to_disk()` returns a `LazyTensorDict` and before anything is assigned to `client_task.result`.

The patch makes `DiskTensorConsumer` explicitly track active in-flight disk tensor downloads while it owns the temp directory. On successful completion ownership is released to `LazyTensorDict`; on error or workflow finalization, the active consumer cleanup removes the partial temp dir. `WFCommServer.finalize_run()` invokes this cleanup before releasing task state.

## Why

#4493 cleaned references after client results had reached workflow task state. QA verified that this handles the post-download case, but not the pre-return streaming case. In that earlier state the only object that knows the temp dir is `DiskTensorConsumer`, so GC of task references cannot remove the directory.


chesterxgchen (approved):
LGTM

PR #4505, opened 2026-04-30
Refresh stale integration tests

YuanTingHsieh:
## Summary
- refresh stale integration tests to match tightened recipe/API contracts
- provide required model/per-site fixtures for recipe smoke tests
- update preflight expectations and remove the embedded hello-pt pytest payload test
- make the direct integration runner require/add localhost0 and localhost1 before preflight tests

## Root Cause
Several integration tests were not updated after production APIs were intentionally tightened. These are test-side fixes only. The hello-pt app/custom test file was also being collected accidentally because it lived under tests/integration_test, but it is not used by the submitted job.

## Validation
- `PYTHONPYCACHEPREFIX=/tmp/nvflare_pycache python3 -m py_compile tests/integration_test/preflight_check_test.py`
- `PYTHONPYCACHEPREFIX=/tmp/nvflare_pycache python3 -m py_compile tests/integration_test/experiment_tracking_recipes_test.py tests/integration_test/xgb_vertical_recipe_test.py tests/integration_test/xgb_histogram_recipe_test.py tests/integration_test/preflight_check_test.py`
- `bash -n tests/integration_test/run_integration_tests.sh`
- `git diff --check`
- `git verify-commit HEAD`


nvidianz (dismissed):

YuanTingHsieh:
Added a signed follow-up commit for the PyTorch `hello-pt` failure.

The failure was caused by the server `PTFileModelPersistor` starting with an empty local model while clients returned `SimpleNetwork` weights. Recent PyTorch persistence validation correctly rejects zero-match updates, so the job config now declares `simple_network.SimpleNetwork` on the server persistor.

Validation:
- `python3 -m json.tool tests/integration_test/data/jobs/hello-pt/app/config/config_fed_server.json`
- `git diff --check`
- first PyTorch integration job (`run hello-pt`) via temp one-test YAML: `1 passed`

YuanTingHsieh:
Added one more signed follow-up for the same PR #4392 PyTorch persistence tightening: `pt_init_client` also used an empty server `PTFileModelPersistor`, so it now declares `simple_network.SimpleNetwork` explicitly.

Validation:
- `python3 -m json.tool tests/integration_test/data/apps/pt_init_client/app/config/config_fed_server.json`
- `git diff --check`

holgerroth:
Reviewed the diff. Test-only refresh, scope looks right. A few suggestions before merge:

### 1. Simplify `_make_model` — `experiment_tracking_recipes_test.py:73-90`

The `sys.path` / `sys.modules` dance with `_is_client_script_module` is doing manually what `importlib.util` does cleanly without touching any globals:

```python
def _make_model(self):
    import importlib.util
    spec = importlib.util.spec_from_file_location(
        "_exp_tracking_model", os.path.join(self.client_script_dir, "model.py")
    )
    module = importlib.util.module_from_spec(spec)
    spec.loader.exec_module(module)
    return module.SimpleNetwork()
```

This drops ~25 lines, removes the `_is_client_script_module` helper entirely, and avoids the (admittedly narrow) parallel-pytest race where another test mutates `sys.modules["model"]` mid-import.

### 2. Silent preflight failures — `preflight_check_test.py:120`

Switching from `subprocess.check_output` to `subprocess.run(..., check=False)` swallows non-zero exit codes from the preflight tool itself. Combined with `_verify_checks` no longer rejecting *unexpected* checks, a crash that still prints recognizable lines could pass. At minimum, log `process.returncode` so failures are visible in test output. If non-zero exits are only expected on the "before overseer start" path, consider gating `check=False` to that path rather than blanket-applying it.

### 3. `/etc/hosts` mutation — `run_integration_tests.sh:75-95`

- `trap restore_localhost_aliases EXIT` is good for normal exits, but SIGKILL leaves the appended line behind. Fine for CI, occasionally surprising for local devs running this script directly. A short comment noting the script only modifies `/etc/hosts` when the user has write perms (typically root in CI) would help future readers.
- `ensure_localhost_aliases` installs an unconditional `trap … EXIT`, which would clobber any future EXIT trap added elsewhere. Currently safe (no other traps), just worth flagging if the script grows.

### 4. Validation

The PR description notes the direct preflight path wasn't exercised locally because `/etc/hosts` needed sudo. Worth confirming CI runs the experiment-tracking and preflight tests green end-to-end before merge — `py_compile` + `bash -n` won't catch the model-loading or hosts-mutation logic.

The xgb `_make_*_per_site_config` helpers and the `test_custom.py` removal both look clean — no concerns there.

YuanTingHsieh:
Addressed the two review follow-ups in signed commit `2642d7db7`:

- Simplified experiment tracking `_make_model()` to load `model.py` via `importlib.util` without mutating `sys.path` / `sys.modules`.
- Made preflight command return codes explicit: success paths now assert return code `0` and include captured output on failure; the intentional before-overseer-start failure path opts out via `expect_success=False`.

Validation:
- `PYTHONPYCACHEPREFIX=/tmp/nvflare_pycache python3 -m py_compile tests/integration_test/experiment_tracking_recipes_test.py tests/integration_test/preflight_check_test.py`
- `git diff --check`

pcnudde (approved):

PR #4510, opened 2026-05-01
Support custom FedAvg best model filename

chesterxgchen:
## Summary
- add best_model_filename to FedAvg recipes while keeping save_filename as a backward-compatible alias
- pass the resolved best model filename into generated PyTorch persistors via PTModel/PTFileModelPersistor
- preserve the legacy no-persistor controller fallback default of FL_global_model.pt
- keep TensorFlow and NumPy recipes API-compatible without adding new best-model event save paths
- add regression coverage for PT passthrough, exported server config, alias conflict handling, and fallback filename behavior

## Memory note
This version avoids adding new TensorFlow/NumPy GLOBAL_BEST_MODEL_AVAILABLE handlers, so it does not introduce extra full-model best-event saves for those frameworks. The runtime persistence wiring change is limited to the existing PyTorch persistor filename option.

## Tests
- pytest tests/unit_test/recipe/fedavg_recipe_test.py tests/unit_test/recipe/component_config_verification_test.py tests/unit_test/app_opt/pt/pt_model_wrapper_test.py -q
- pytest tests/unit_test/recipe -q


chesterxgchen:
Handled in cfcd7efcb.

The TF/NumPy persistor/event-handler changes were removed in a1576ac65 after the memory review: adding GLOBAL_BEST_MODEL_AVAILABLE handlers there introduced extra full-model save paths and could save stale GLOBAL_MODEL data because that event fires before FedAvg updates the round model.

Current scope:
- PyTorch: best_model_filename is wired through PTModel to PTFileModelPersistor.best_global_model_file_name.
- TensorFlow/NumPy: the recipe args remain API-compatible, but the default persistors do not create a separate best-model artifact in this PR. The docstrings now state that explicitly instead of implying support.

Validation after these changes:
- pytest tests/unit_test/recipe/fedavg_recipe_test.py tests/unit_test/recipe/component_config_verification_test.py tests/unit_test/app_opt/pt/pt_model_wrapper_test.py -q

chesterxgchen:
Handled the remaining fallback-default concern in 5c75e1986.

Details:
- The no-persistor FedAvg controller fallback now preserves the legacy default save filename: FL_global_model.pt.
- Explicit best_model_filename still configures the fallback path when no persistor exists.
- PyTorch persistors continue to receive best_model_filename via PTModel -> PTFileModelPersistor.best_global_model_file_name.
- TF/NumPy remain documented as API-compatible only, with no extra best-event save path added.

Added regression tests for both no-persistor cases:
- default fallback stays FL_global_model.pt
- explicit best_model_filename is used by the controller fallback

Latest validation:
- pytest tests/unit_test/recipe/fedavg_recipe_test.py tests/unit_test/recipe/component_config_verification_test.py tests/unit_test/app_opt/pt/pt_model_wrapper_test.py -q: 63 passed, 5 skipped
- pytest tests/unit_test/recipe -q: 263 passed, 22 skipped

chesterxgchen:
Checked the current open comments on the latest head (5c75e1986).

There are no unresolved review threads. The remaining Greptile notes are PR-summary comments rather than resolvable review threads:

- TF/NumPy `best_model_filename`: intentionally API-compatible only in this PR. The earlier TF/NumPy event-handler save paths were removed in a1576ac65 because they introduced extra full-model save paths / memory pressure and could save stale `GLOBAL_MODEL` data. The recipe docstrings and PR body now state this limitation explicitly.
- `save_filename` alias behavior: intentionally retained as a deprecated alias for `best_model_filename`. This restores the originally intended custom best-model filename behavior for users who already supplied `save_filename`. The legacy no-persistor controller fallback default is still preserved as `FL_global_model.pt`, fixed and tested in 5c75e1986.

No additional code change is needed for these comments.

YuanTingHsieh (dismissed):

chesterxgchen:
Handled the latest Greptile feedback in 174ce0ae2.

What changed:
- `save_filename` still works as the deprecated alias for `best_model_filename`, which preserves the requested bug fix for existing FedAvg recipe users, but it now emits a `FutureWarning` instead of changing behavior silently.
- TensorFlow/NumPy explicit `best_model_filename` or `save_filename` now emits a `UserWarning` explaining that those recipe paths are API-compatible only and do not create a separate best-model artifact, because adding new best-model event save paths would increase model memory use.
- Added regression coverage for the TF/NumPy warning behavior and the deprecated alias warning.

Validation:
- `pytest tests/unit_test/recipe/fedavg_recipe_test.py tests/unit_test/recipe/component_config_verification_test.py tests/unit_test/app_opt/pt/pt_model_wrapper_test.py -q`: 65 passed, 5 skipped
- `pytest tests/unit_test/recipe -q`: 265 passed, 22 skipped
- `python -m black --check nvflare/recipe/fedavg.py tests/unit_test/recipe/fedavg_recipe_test.py`: passed

YuanTingHsieh (approved):

PR #4514, opened 2026-05-04
Adding no-BYOC support for Flower integration

d0uwe:
## Description

This PR adds support for pre-deployed Flower application code in NVFlare's Flower integration, eliminating the need for BYOC (Bring Your Own Code) authorization in production environments.

### Key Changes

**New Feature: Pre-deployed Mode**
- Added `flower_app_path` parameter to `FlowerJob`, `FlowerRecipe`, `FlowerController`, `FlowerExecutor`, and `FlowerServerApplet`
- When using `flower_app_path`, the Flower app code is not packaged into the job ZIP (no `custom/` directory)
- The job ZIP does not require BYOC authorization on any participant
- Important: The path only needs to exist on the server. Clients receive the Flower app from the server via Flower's FAB (Flower Application Bundle) distribution mechanism, so client participants do not need a pre-existing copy. Unfortunately there is currently not yet a way in Flower itself to disable BYOC, so the clients do still need to trust the server admin that sets up the server, but it's a step forward compared to only having BYOC as an option. 

**Backwards Compatibility**
- Existing `flower_content` parameter continues to work exactly as before (BYOC mode)
- Users can choose either mode:
  - **BYOC mode** (`flower_content`): Package Flower app in job ZIP (requires BYOC authorization)
  - **Pre-deployed mode** (`flower_app_path`): Server-only deployment (no BYOC needed)

### Types of changes
- [x] Documentation updated.


tests/unit_test/recipe/flower_recipe_test.py
```diff
@@ -83,3 +85,100 @@ def test_flower_recipe_rejects_extra_env_with_wrong_client_api_type(flwr_version
...
+    assert kwargs["flower_app_path"] == "/opt/flower_apps/my_app"
+    assert kwargs["flower_content"] is None
+
+
+@pytest.mark.parametrize("flwr_version", ["1.26.0", "1.26.1", "1.27.5"])
+def test_flower_recipe_passes_content_and_path_through(flwr_version):
+    fake_job = object()
+    with patch("nvflare.app_opt.flower.recipe.get_package_version", return_value=flwr_version):
+        with patch("nvflare.app_opt.flower.recipe._create_flower_job", return_value=fake_job) as mock_flower_job:
+            recipe = FlowerRecipe(flower_content="mock_flower_content", flower_app_path="/opt/flower_apps/my_app")
```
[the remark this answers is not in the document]
d0uwe: deleted test, arguments are sufficiently tested by other functions.

nvflare/app_opt/flower/applet.py
```diff
@@ -129,10 +129,16 @@ def _format_run_config_value(value) -> str:
...
+    def __init__(
+        self,
+        extra_env: dict = None,
+        allow_runtime_dependency_installation: bool = False,
+        flower_app_path: Optional[str] = None,
+    ):
         """Constructor of FlowerClientApplet, which extends CLIApplet."""
         CLIApplet.__init__(self, stop_method="term")
         self.allow_runtime_dependency_installation = allow_runtime_dependency_installation
+        self.flower_app_path = flower_app_path
```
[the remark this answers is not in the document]
d0uwe: fixed

nvflare/job_config/fed_job_config.py
```diff
@@ -157,7 +157,7 @@ def generate_job_config(self, job_root):
             config_dir = os.path.join(job_dir, app_name, CONFIG)
             custom_dir = os.path.join(job_dir, app_name, CUSTOM)
             os.makedirs(config_dir, exist_ok=True)
-            os.makedirs(custom_dir, exist_ok=True)
+            # custom_dir will be created on-demand if custom code is added.
```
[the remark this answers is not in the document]
d0uwe: How should we handle this? I noticed that if the custom_dir gets created, there always is an empty custom dir, which then triggers the byoc check, even though the directory is empty.

holgerroth:
@d0uwe Thanks for the contribution. I think the feature is useful, especially for production Flower deployments where the app is pre-installed and should not be packaged into the NVFlare job ZIP. But I’m concerned about describing this as “no-BYOC.”

With this PR, Flower app code is no longer carried through NVFlare `custom/`, but clients can still receive and execute app code through Flower’s FAB distribution mechanism. That means “BYOC disabled” would no longer clearly mean “no job-specific executable code can reach clients.”

Could we frame this as “server-predeployed Flower app mode” instead of “no-BYOC,” and add an explicit policy/guardrail for it? For example, Flower FAB execution could require a separate site opt-in when BYOC/custom-code execution is disabled. Ideally, `flower_app_path` should also be constrained to admin-approved locations or app registrations, rather than being an arbitrary path from the job definition.

I’m supportive of the use case, but I think we should preserve NVFlare’s security contract: disabling BYOC should not silently allow another dynamic code path to clients unless the site explicitly opts into that trust model.

@IsaacYangSLA, @nvidianz, @pcnudde for viz

d0uwe:
@holgerroth, I fully agree. I've just pushed an attempt at this, can you let me know if this is roughly what you had in mind? I tested it using the POC mode, and it seems to work. 

Two things to still address: 
- If the server has, for some reason, set "server-predeployed-flwr" to any, but the sites have not, the run starts, but doesn't start on the clients, but doesn't exit. 
- restrict the flower_app_path; is there any example of a check like this already in NVFlare?

holgerroth:
@d0uwe, I think this goes in the right direction. The new “server-predeployed Flower app mode” and permission are a good step, but I’d like the implementation to make the new trust boundary explicit and hard to bypass before we call this safe for restricted BYOC deployments. @nvidianz and @IsaacYangSLA can probably give more details in this direction.

holgerroth:
In general, I think the new safety contract with `NVFlare BYOC disabled & flower_predeployed=true` should be framed as:

> No user-provided NVFlare custom code is deployed through the job. Flower app code may still be distributed by Flower, but only from server-admin-controlled/preapproved app locations.

So, “preinstalled on the server” really means admin-controlled and not user-controlled. The PR should make that hard to accidentally violate, especially by avoiding arbitrary user-chosen `flower_app_path` unless it is constrained to approved paths/apps.

d0uwe:
Hi @holgerroth, I've clarified this further and added path verification to ensure the path refers to something in local/custom.

holgerroth:
Looks good. Let's address and resolve the Greptile comments and we should be good to merge @d0uwe.

nvflare/app_opt/flower/applet.py
```diff
@@ -311,8 +324,24 @@ def start(self, app_ctx: dict):
...
+            if not os.path.isdir(self.flower_app_path):
+                raise RuntimeError(
+                    f"flower_app_path '{self.flower_app_path}' does not exist on this host. "
+                    "Ensure the Flower app is pre-deployed on the server. "
+                    "(Clients receive the app from the server via Flower's FAB distribution)."
+                )
+            self.flower_app_dir = self.flower_app_path
+        else:
+            custom_dir = ws.get_app_custom_dir(fl_ctx.get_job_id())
+            self.flower_app_dir = custom_dir
```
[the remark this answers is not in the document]
d0uwe: Now done like so:
```
            # Resolve relative path to absolute path relative to workspace root
            workspace_root = ws.get_root_dir()
            self.flower_app_dir = os.path.abspath(os.path.join(workspace_root, self.flower_app_path))
```

tests/unit_test/recipe/flower_job_test.py
```diff
@@ -0,0 +1,232 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+import tempfile
+import zipfile
+from pathlib import Path
+from unittest.mock import MagicMock, patch
```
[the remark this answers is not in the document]
d0uwe: added import in [db823d1](https://github.com/NVIDIA/NVFlare/pull/4514/commits/db823d1aba522f42d39ad06cd0d963f787ef1397)

nvflare/app_opt/flower/utils.py
```diff
@@ -66,3 +66,38 @@ def shareable_to_msg_container(s: Shareable) -> pb2.MessageContainer:
...
+        raise ValueError(f"flower_app_path contains invalid path traversal: '{path}'")
+
+    if not allow_symlinks:
+        import os
+
+        if os.path.islink(path.rstrip(os.sep)):
+            raise ValueError(
+                f"flower_app_path '{path}' is a symbolic link. "
+                "For security, pre-deployed app paths must be real directories, not links."
+            )
```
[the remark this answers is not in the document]
d0uwe: resolved

nvflare/job_config/fed_job_config.py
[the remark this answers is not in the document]
d0uwe: added the "os.makedirs(dest_path, exist_ok=True)" line

d0uwe:
@holgerroth Greptile seems to be happy now!

nvflare/private/fed/utils/app_authz.py
```diff
@@ -43,17 +45,32 @@ def authorize(
...
+                return False, "BYOC not permitted"
 
-        ctx = AuthzContext(
-            user=Person(submitter_name, submitter_org, submitter_role),
-            submitter=Person(submitter_name, submitter_org, submitter_role),
-            right=_RIGHT_BYOC,
-        )
+        # Guard for server-predeployed Flower app mode.
+        # If job_meta is provided, check the flag directly.
+        is_flower_predeployed = job_meta.get(AppValidationKey.FLOWER_PREDEPLOYED, False) if job_meta else False
```
holgerroth: [P1] This gate still trusts a user-controlled `meta.json` flag to decide whether the new `server-predeployed-flwr` permission is required. A hand-crafted or tampered Flower job can keep `flower_app_path` in the server config but omit `flower_predeployed`, causing this branch to skip the new permission while still running in server-predeployed Flower mode. I reproduced this by exporting a predeployed `FlowerJob`, removing `flower_predeployed` from `meta.json`, and validating the job: `JobMetaValidator` returned valid with no Flower flag and no BYOC flag. Please derive/enforce this mode from trusted job config during server-side validation, or reject config/meta mismatches, rather than relying only on the submitted metadata flag.

holgerroth (commented):
Added the main P1 concern inline.

docs/hello-world/hello-flower/index.rst
```diff
@@ -120,15 +137,24 @@ Run Job
...
    python job.py --job_name "flwr-pt" --content_dir "./flwr-pt"
 
+Run ``flwr-pt`` with NVFlare Simulation (Pre-deployed Mode)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+If the Flower app is pre-deployed on the server (clients receive it via Flower's FAB distribution):
+
+.. code-block:: bash
+
+   python job.py --job_name "flwr-pt" --flower_app_path "/opt/flower_apps/flwr-pt"
```
nvidianz: Should we allow absolute path?

nvflare/apis/app_validation.py
```diff
@@ -19,6 +19,7 @@
 class AppValidationKey(object):
 
     BYOC = "byoc"
+    FLOWER_PREDEPLOYED = "flower_predeployed"
```
nvidianz: This isn't the right place for this addition. The apis module is reserved for NVFlare's core APIs — this should instead be added to the flower module.

nvflare/app_opt/flower/applet.py
```diff
@@ -311,8 +324,30 @@ def start(self, app_ctx: dict):
...
+        if self.flower_app_path:
+            # Resolve relative path to absolute path relative to workspace root
+            workspace_root = ws.get_root_dir()
+            self.flower_app_dir = os.path.abspath(os.path.join(workspace_root, self.flower_app_path))
+
+            # Validate path format
+            try:
+                validate_flower_app_path(self.flower_app_path)
+            except ValueError as e:
+                raise RuntimeError(str(e))
```
nvidianz: No need to wrap ValueError.

nvflare/app_opt/flower/applet.py
```diff
@@ -238,6 +248,7 @@ def __init__(
         superlink_min_query_interval=10.0,
         run_config: Optional[dict] = None,
         allow_runtime_dependency_installation: bool = False,
+        flower_app_path: Optional[str] = None,
```
nvidianz, line 251: flower_app_path should be a relative path. A configuration parameter should be provided for site admins to set the root directory for Flower apps — allowing users to load code from arbitrary folders poses a security risk.

docs/hello-world/hello-flower/index.rst
d0uwe: No! This is outdated and I've update it, currently flower code must reside in local/custom/

nvflare/app_opt/flower/applet.py
d0uwe: Currently the code enforces that the flower code is located anywhere in local/custom (https://github.com/d0uwe/NVFlare/blob/40bd821c88eedde22e2abd877feac22f7ae1554b/nvflare/app_opt/flower/applet.py#L332-L336). 

Is this sufficient, or do we want to make this configurable?

nvidianz: `local/custom` is fine but the example/usage all uses absolute path.

nvflare/private/fed/utils/app_authz.py
holgerroth: @d0uwe I created a PR to your branch trying to fix this issue https://github.com/d0uwe/NVFlare/pull/3

d0uwe: Merged, thank you!

nvflare/app_opt/flower/applet.py
d0uwe: All references to /opt/flower_apps/flwr-pt are now replaced with local/custom

nvflare/apis/app_validation.py
d0uwe: Got it. Moved it to nvflare/app_opt/flower/defs.py

tests/unit_test/recipe/flower_job_test.py
```diff
@@ -0,0 +1,233 @@
...
+                            with z.open("meta.json") as f:
+                                meta = json.load(f)
+                                assert FlowerConstant.FLOWER_PREDEPLOYED in meta
+                                assert meta[FlowerConstant.FLOWER_PREDEPLOYED] is True
+
+    def test_flower_job_accepts_local_custom_path(self):
+        """Valid local/custom/ paths should be accepted."""
+        with (
+            patch("nvflare.app_opt.flower.flower_job.FlowerController", return_value=_make_mock_component("server")),
+            patch("nvflare.app_opt.flower.flower_job.FlowerExecutor", return_value=_make_mock_component("client")),
```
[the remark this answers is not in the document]
d0uwe: Adjusted!

holgerroth (approved):
This looks good now. Great job and thanks very much for the contribution @d0uwe!

PR #4520, opened 2026-05-05
Clear retained training result after accept events

pcnudde:
## What changed

This PR clears `AppConstants.TRAINING_RESULT` from the server-side `FLContext` immediately after the contribution-accept event window completes.

`_accept_train_result()` still stores the raw client `Shareable` in `FLContext` so existing `BEFORE_CONTRIBUTION_ACCEPT` / `AFTER_CONTRIBUTION_ACCEPT` handlers can inspect it. After those handlers run, `_process_result()` now resets the property value to `None` while preserving the existing private/sticky attributes.

## Bug

In a normal FedAvg-style workflow, the server receives a client model update as a raw `Shareable`. `BaseModelController._accept_train_result()` stores that raw result under `AppConstants.TRAINING_RESULT` for contribution-accept event handlers.

After the result is accepted, `_process_result()` converts the raw `Shareable` into an `FLModel` and FedAvg aggregates it immediately. The communication/task path then clears `client_task.result`, but the same raw `Shareable` can still be reachable through:

```text
controller.fl_ctx -> FLContext.props -> AppConstants.TRAINING_RESULT -> Shareable -> tensors
```

That keeps the client update live even though FedAvg has already aggregated it. With disk tensor offload disabled, this can retain a full in-memory client model update. For example, with a 5 GB model, built-in streaming FedAvg can remain roughly one extra model-sized update higher than necessary after each accepted result, and can briefly hold the previous retained update plus the next incoming update during result processing.

This is a lifetime bug, not allocator caching: the tensor objects are still strongly reachable from `FLContext`, so garbage collection cannot reclaim them.

## Benefit

The raw client result is now scoped to the event window where it is intended to be visible. After that point, only the objects FedAvg actually needs remain reachable, such as the aggregation accumulator and the updated/global model state.

For large in-memory models, this avoids retaining an unnecessary extra client update after aggregation. In the 5 GB model example, steady memory after aggregating a result can drop from roughly `global model + accumulator + retained raw result` to just `global model + accumulator`.

### DIFF
--- a/nvflare/app_common/workflows/base_model_controller.py
+++ b/nvflare/app_common/workflows/base_model_controller.py
@@ -237,6 +237,17 @@ class BaseModelController(Controller, FLComponentWrapper, ABC):
         else:
             fl_ctx.set_prop(key, value, private=default_private, sticky=default_sticky)
 
+    @staticmethod
+    def _clear_training_result(fl_ctx: FLContext) -> None:
+        detail = fl_ctx.get_prop_detail(AppConstants.TRAINING_RESULT)
+        if detail is not None:
+            fl_ctx.set_prop(
+                AppConstants.TRAINING_RESULT,
+                None,
+                private=detail["private"],
+                sticky=detail["sticky"],
+            )
+
     def _process_result(self, client_task: ClientTask, fl_ctx: FLContext) -> None:
         self.fl_ctx = fl_ctx
         result = client_task.result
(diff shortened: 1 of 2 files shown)


pcnudde:
Thanks for the review. The `process_result_of_unknown_task()` path is a different, late/unknown-task handling path and is intentionally out of scope for this PR.

This change is focused on the normal FedAvg result-processing path where accepted client training results are converted, aggregated, and then should be released immediately. In that path, `TRAINING_RESULT` remains available for the contribution-accept event window and is cleared right after, which fixes the large in-memory model retention case described in the PR.

I agree the unknown-task path can be evaluated separately, but it is not important for this specific FedAvg memory-retention fix.

YuanTingHsieh (approved):

YuanTingHsieh (approved):

holgerroth (approved):

holgerroth:
@greptileai review again.

nvidianz (approved):

PR #4521, opened 2026-05-05
Align CLI Python support with 3.10-3.14

pcnudde:
## Summary
- Update the global CLI runtime version guard to allow Python 3.10 through 3.14.
- Remove Python 3.9 from supported-version metadata, CI matrices, and installation guidance.
- Keep explicit rejection for Python 3.9 and earlier and future 3.15+ versions.
- Add unit coverage for the supported and unsupported version boundaries.

## Root Cause
The project metadata and CI matrix were updated for Python 3.13/3.14 support, but the shared `version_check()` guard still rejected Python 3.13 and newer before dispatching any `nvflare` CLI subcommand. This made commands such as `nvflare provision` fail on Python 3.14 despite the advertised support.

This PR also aligns the declared minimum supported Python version at 3.10 by removing Python 3.9 from package metadata, CI, docs, and runtime validation.


YuanTingHsieh (approved):
LGTM

nvidianz (approved):

PR #4522, opened 2026-05-05
Add run-scoped tensor offload temp cleanup

pcnudde:
## Summary

Catchall solution for disk tensor offload cleanup without having to find each memory leak

- create a per-FedAvg-run tensor disk offload root when disk offload is enabled
- store that root in FOBS context so disk tensor downloads create temp dirs under the run-owned root
- remove the run root during FedAvg cleanup so retained result/lazy references cannot keep disk files behind after run teardown


nvflare/app_common/utils/tensor_disk_offload_context.py
```diff
@@ -15,11 +15,13 @@
 from typing import Any, Tuple
 
 _ENABLE_TENSOR_DISK_OFFLOAD = "enable_tensor_disk_offload"
+TENSOR_DISK_OFFLOAD_ROOT_DIR = "tensor_disk_offload_root_dir"
```
nvidianz: Should this also start with _ to be consistent?

nvidianz (approved):

nvflare/app_common/utils/tensor_disk_offload_context.py
```diff
@@ -58,6 +64,10 @@ def restore_enable_tensor_disk_offload(engine, previous_value: Any) -> None:
     if not cell:
         return
 
+    if root_dir:
```
YuanTingHsieh, line 67: codex added:

If disk offload was already enabled before this FedAvg run, apply_enable_tensor_disk_offload overwrites the existing TENSOR_DISK_OFFLOAD_ROOT_DIR but only returns the previous boolean. The restore path then re-enables disk offload while setting the root to None, so the next tensor download sees enable_tensor_disk_offload=True and download_tensors_to_disk raises because the required root is missing. Please save and restore the prior root dir along with the prior enabled value.

^ could you double check if this is a real issue

pcnudde: Double checked. This is not possible with current code. There would need to be some other code that specifically resets that value.

PR #4526, opened 2026-05-06
[2.8] Add parent pod python path

IsaacYangSLA:
### Description

Add k8s.yaml configuration to set the command of the container in helm chart.  Also set the workspace path.

### Types of changes

- [x] Documentation updated.


docs/design/site_runtime_packaging_design.md
```diff
@@ -332,7 +333,18 @@ Supported `parent` keys:
...
   - Required: no
   - Default: `/var/tmp/nvflare/workspace`
-  - Description: in-container mount path for the workspace PVC.
+  - Description: in-container mount path for the parent workspace PVC. This is
+    also passed to `K8sJobLauncher` so dynamically launched job pods mount the
+    job workspace and startup kit at the same path.
+
+- `python_path`
+  - Required: no
+  - Default: `job_launcher.default_python_path` when set, otherwise
```
YuanTingHsieh: parent.python_path controls the parent pod, while job_launcher.default_python_path controls dynamically launched job pods. Defaulting the parent command to job_launcher.default_python_path couples two separate runtime layers and makes the config harder to reason about. Could we default parent.python_path only to the parent constant, e.g. /usr/local/bin/python3, unless it is explicitly set?

IsaacYangSLA: These two python paths (parent or job) should be independent to each other.  But because job_launcher.default_python_path was there before this PR, and parent.python_path is not required, setting its default to be the same as job_launcher.default_python_path seems reasonable.

Changing parent.python_path's default to /usr/local/bin/python3 is also OK to me.  The main purpose of this PR is to give the option to set parent.python_path, which is originally not configurable.

nvflare/tool/deploy/deploy_commands.py
```diff
@@ -668,10 +672,12 @@ def _docker_parent_command(kit_info: KitInfo) -> str:
 
 def _write_helm_chart(kit_info: KitInfo, config: dict[str, Any]) -> Path:
     parent = config.get("parent") or {}
+    job_launcher = config.get("job_launcher") or {}
     docker_image = parent["docker_image"]
     parent_port = parent.get("parent_port", 8102)
     workspace_pvc = parent.get("workspace_pvc", "nvflws")
     workspace_mount_path = parent.get("workspace_mount_path", WORKSPACE_MOUNT_PATH)
+    parent_python_path = parent.get("python_path") or job_launcher.get("default_python_path") or K8S_PARENT_PYTHON_PATH
```
YuanTingHsieh: suggest change this to:

`parent_python_path = parent.get("python_path") or K8S_PARENT_PYTHON_PATH`

YuanTingHsieh (approved):

PR #4527, opened 2026-05-06
[2.8] Fix job pods stuck in certain states

IsaacYangSLA:
### Description

More checks in k8s job launcher states.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/app_opt/job_launcher/k8s_launcher.py
+++ b/nvflare/app_opt/job_launcher/k8s_launcher.py
@@ -306,12 +306,13 @@ class K8sJobHandle(JobHandleSpec):
                 self.logger.info(f"job {self.job_id} pod not found during querying; assuming terminated")
                 self.terminal_state = JobState.TERMINATED
                 self._remove_workspace_job()
+                return PodPhase.UNKNOWN.value
             else:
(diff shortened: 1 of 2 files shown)


YuanTingHsieh (commented):
This fixes the 404/deleted-pod path and real K8s Unknown phase. One question: for persistent non-404 API read failures, _query_phase() now returns None, and _stuck_in_pending() ignores it. With timeout=None, can enter_states() still loop forever if the API remains unreachable? Do we need a separate read-error budget or wall-clock launch timeout for that case?

IsaacYangSLA:
> This fixes the 404/deleted-pod path and real K8s Unknown phase. One question: for persistent non-404 API read failures, _query_phase() now returns None, and _stuck_in_pending() ignores it. With timeout=None, can enter_states() still loop forever if the API remains unreachable? Do we need a separate read-error budget or wall-clock launch timeout for that case?

When timeout=None, the users' intention would be no timeout, ie keep trying forever.  In such case, it seems strange to exit.  Maybe we need a global abort strategy when the cluster itself has issues.  But we need to know exactly what type of issues clusters may have, then we design a way to escape from that.  Right now, we only try to cover some known issues.

YuanTingHsieh (approved):

PR #4084, opened 2026-02-01
[2.7] Add dynamic ignore_result_error logic and POC environment cleanup

chesterxgchen:
## Summary

This PR introduces a flexible 3-mode error handling policy for FL controllers and adds proper cleanup for POC environment after job execution.

### Error Handling Improvements

- **Dynamic `ignore_result_error` mode**: Controllers now support three modes for handling client result errors:
  - `None` (default): **Dynamic mode** - ignore errors if `min_responses`/`min_clients` can still be reached, panic otherwise
  - `False`: **Strict mode** - always panic on any client error
  - `True`: **Resilient mode** - always ignore client errors and continue

- **Shared utility functions**: Created reusable error handling logic to avoid code duplication across controllers

### POC Environment Cleanup

- Added `stop()` method to `ExecEnv` base class (default no-op)
- `Run.get_result()` now calls `exec_env.stop()` in a `finally` block
- Ensures POC services are always stopped after job execution, making each run independent

## Files Changed

| File | Changes |
|------|---------|
| `nvflare/app_common/utils/error_handling_utils.py` | **New** - Shared utility functions: `should_ignore_result_error()`, `get_error_handling_message()` |
| `nvflare/app_common/workflows/base_model_controller.py` | Updated `ignore_result_error` default to `None`, added 3-mode logic in `_accept_train_result()`, added tracking for `_current_failed_clients`, `_current_num_targets`, `_current_min_responses` |
| `nvflare/app_common/workflows/scatter_and_gather.py` | Updated `ignore_result_error` default to `None`, added 3-mode logic in `_accept_train_result()`, added tracking variables, updated docstring |
| `nvflare/app_common/workflows/scatter_and_gather_scaffold.py` | Updated `ignore_result_error` signature and docstring |
| `nvflare/app_common/workflows/scaffold.py` | Updated `ignore_result_error` docstring |
| `nvflare/recipe/spec.py` | Added `stop()` method to `ExecEnv` base class |
| `nvflare/recipe/run.py` | Updated `get_result()` to call `exec_env.stop()` in finally block |
| `tests/unit_test/app_common/utils/error_handling_utils_test.py` | **New** - 17 unit tests for error handling logic |

## Controllers Affected

Controllers that inherit from `BaseModelController` automatically get the new behavior:
- `FedAvg`, `BaseFedAvg`, `Scaffold`, `Cyclic`, `PTFedAvg`, `FedOpt`, etc.

Controllers with direct changes:
[...]

PR #4541, opened 2026-05-07
Add cleanup parameter to recipe Run.get_result() (default True)

nvshaxie:
## Summary
- Run.get_result() previously hard-coded `exec_env.stop(clean_up=True)` with no way for callers to opt out of workspace removal.
- After PR #4500 moved subprocess stdout into per-service `poc_console.log` files **inside the workspace**, this hard-coded cleanup also wipes those console logs the moment `get_result()` returns. Combined with the "each run is independent" cleanup from PR #4084, this means `print(\"Result can be found in :\", run.get_result())` (the pattern used by every official `examples/*/job.py`) prints a path that has already been rmtree'd, and no per-service log — including `poc_console.log` — survives for inspection.
- This PR adds a `cleanup: bool = True` parameter to `Run.get_result()` and forwards it to `exec_env.stop(clean_up=cleanup)`. **The default is preserved as `True`**, so existing behavior is unchanged. New callers (debuggers, test frameworks asserting against per-service logs, agent workflows) can now opt to keep the workspace by calling `run.get_result(cleanup=False)`.

## Behaviour
| Call | Services stopped | Workspace |
|---|---|---|
| `run.get_result()` (default, unchanged) | yes | removed |
| `run.get_result(cleanup=False)` (new opt-out) | yes | preserved on disk |

## Why this matters now
PR #4500 changed where subprocess stdout goes: instead of inheriting parent stdout, each service writes to `<workspace>/example_project/prod_NN/<service>/poc_console.log`. That file is the only post-run record of the service's stdout. Today it disappears together with the workspace inside `get_result()`, so users / tests / agents that want to inspect what server and clients actually printed have no way to do so without forking NVFlare.

## Changes
- `nvflare/recipe/run.py`: add `cleanup: bool = True` parameter to `Run.get_result()`; forward to `exec_env.stop(clean_up=cleanup)` instead of hard-coded `True`. Docstring updated to call out the relationship with `poc_console.log` and PR #4500.
- `tests/unit_test/recipe/run_test.py`: existing assertions on `clean_up=True` remain unchanged. Added `test_get_result_cleanup_false_preserves_workspace` covering the `cleanup=False` propagation path.

## Verified locally
- Two consecutive `recipe.execute(env=PocEnv(...))` runs with `cleanup=False` succeed; the second run lands in `prod_01/` next to the preserved `prod_00/` from the first run, services correctly stopped between runs.
- All 21 `Run` unit tests pass; license, black, isort, flake8 clean.

### Types of changes
- [x] Quick tests passed locally by running `./runtest.sh` (license + style + targeted unit tests `tests/unit_test/recipe/run_test.py` 21 passed, 3 skipped).
- [ ] Documentation updated.

(Docstring of `Run.get_result()` itself updated; no public docs touched.)

### DIFF
--- a/nvflare/recipe/run.py
+++ b/nvflare/recipe/run.py
@@ -74,13 +74,19 @@ class Run:
                 self.logger.warning(f"Failed to get job status: {e}")
                 return None
 
-    def get_result(self, timeout: float = 0.0) -> Optional[str]:
+    def get_result(self, timeout: float = 0.0, cleanup: bool = True) -> Optional[str]:
         """Get the result workspace of the run.
 
         Waits for job to complete, caches status, then stops execution environment.
 
         Args:
             timeout (float, optional): Timeout for job completion. Defaults to 0.0 (no timeout).
+            cleanup (bool, optional): Whether to remove the execution-environment workspace
+                (e.g. the POC workspace) when stopping. Defaults to True, preserving the
+                existing "each run is independent" behavior. Pass ``cleanup=False`` to keep
+                the workspace on disk after the run so server/client log files (including
+                the per-service ``poc_console.log`` introduced in #4500) remain available
+                for debugging or test assertions.
 
         Returns:
             Optional[str]: Result workspace path, or None if job not finished or on error.
@@ -104,7 +110,7 @@ class Run:
                 self._cached_status = None
 
             try:
-                self.exec_env.stop(clean_up=True)
+                self.exec_env.stop(clean_up=cleanup)
             except Exception as e:
                 self.logger.warning(f"Failed to stop execution environment: {e}")
             finally:
--- a/tests/unit_test/recipe/run_test.py
+++ b/tests/unit_test/recipe/run_test.py
@@ -76,6 +76,15 @@ class TestRunClass:
         assert self.run._cached_status == "FINISHED"
         assert self.run._cached_result == "/tmp/workspace/test_job_123"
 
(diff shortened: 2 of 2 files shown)


nvshaxie:
Thanks @greptile-apps for the review. Note that this PR was retargeted after initial submission — the current PR title, body and commit message reflect the updated intent:

- **Default `cleanup=True` is intentional, not an oversight.** It preserves the existing "each run is independent" behavior from PR #4084 and the explicit Test Plan item from PR #4132 ("POC workspace is cleaned up after get_result()"). Changing the default would be a behavior change that contradicts established design.
- **The two existing test assertions (`clean_up=True`) are correct as-is** and were not meant to change. The new `test_get_result_cleanup_false_preserves_workspace` covers the new opt-out path.
- The PR is intentionally a **purely additive enhancement** that gives callers an opt-out (`cleanup=False`) without altering the default behavior. Existing callers including all `examples/*/job.py` continue to work identically.

The earlier version of this PR did flip the default; that approach was withdrawn after confirming with the project's design intent. Greptile's analysis appears to be based on the prior PR description.

The current (post-update) state:
- Default behavior unchanged (`run.get_result()` still removes workspace)
- New opt-out: `run.get_result(cleanup=False)` keeps workspace + per-service `poc_console.log` (PR #4500) on disk for debugging / test assertions
- All 21 existing unit tests pass with their original assertions; one new test added for `cleanup=False`.

YuanTingHsieh (approved):

PR #4544, opened 2026-05-07
Avoid k8s pod name collisions across sites

pcnudde:
## Summary
- Build Kubernetes pod names from the FL job id plus a stable site suffix so server and client job pods can run in the same namespace without name collisions.
- Keep the original FL job id for workspace transfer and process arguments.
- Add unit coverage for site-scoped pod names and the launch manifest name.


IsaacYangSLA (approved):
Thanks.  Great improvement.

PR #4550, opened 2026-05-07
[2.8] Add clean_up parameter to recipe Run.get_result()

YuanTingHsieh:
## Summary
- Backport of #4541 to the `2.8` branch, adjusted per review to use the existing `clean_up` spelling.
- Adds a `clean_up: bool = True` parameter to `Run.get_result()` and forwards it to `exec_env.stop(clean_up=clean_up)` instead of hard-coding cleanup.
- Preserves existing behavior by default while allowing callers to keep the workspace with `run.get_result(clean_up=False)` for debugging and log inspection.
- Adds unit coverage for the `clean_up=False` propagation path, including result and cached-state assertions.

## Testing
- `git diff --check`
- `python3 -m py_compile nvflare/recipe/run.py tests/unit_test/recipe/run_test.py`


nvflare/recipe/run.py
```diff
@@ -74,13 +74,19 @@ def get_status(self) -> Optional[str]:
                 self.logger.warning(f"Failed to get job status: {e}")
                 return None
 
-    def get_result(self, timeout: float = 0.0) -> Optional[str]:
+    def get_result(self, timeout: float = 0.0, cleanup: bool = True) -> Optional[str]:
```
nvidianz: We should use the same parameter name, cleanup -> clean_up

pcnudde (approved):

PR #4551, opened 2026-05-07
Add generic Kubernetes support to multicloud deploy

pcnudde:
## Summary

Adds generic Kubernetes support to the multicloud devops tooling without requiring cloud-specific IP reservation or namespace ownership.

Key changes:
- add a `kubernetes` provider for existing clusters, configurable service type, configured kubeconfig paths, and local admin port-forward defaults for ClusterIP deployments
- make `deploy.py` consume an explicit project YAML instead of generating participant entries, while still templating `__SERVER_IP__`
- support reused namespaces with opt-in namespace creation/deletion behavior and scoped PVC cleanup
- update `k8sview.py` to distinguish participant pods from job pods in shared namespaces and show GPU requests on job pods
- add focused unit coverage and dry-run project fixtures

### DIFF
--- a/devops/multicloud/README.md
+++ b/devops/multicloud/README.md
@@ -15,7 +15,10 @@ this example.
 Shipped config:
 - `all-clouds.yaml` — GCP server with GCP, AWS, and Azure clients
 
-Cluster topology is defined in the YAML (`clouds:` + `participants:`).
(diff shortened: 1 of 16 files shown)


pcnudde:
closing for now, will simplify code

PR #4552, opened 2026-05-08
[2.8] Fix job timeout status in list_jobs

IsaacYangSLA:
### Description

When a job pod is launched but the resources it required are not available, the job pod wait in pending phase until timeout.  The server does not properly set the job status and list_jobs shows RUNNING.

This PR fixes the server/client runner and other related portions.  Also add one job status in k8s to explicit inform the job failure reason.

### Types of changes

- [x] Documentation updated.


pcnudde:
Potential status downgrade after `fail_run()` from a later client abort report:

After a client-side pending timeout, `fail_run()` records `PROCESS_RETURN_CODE = EXCEPTION` in `engine.exception_run_processes` and then aborts the remaining runs. But a later `REPORT_JOB_FAILURE` with `JobReturnCode.ABORTED` from another client is routed through `process_job_failure()` to `job_runner.stop_run()`. `stop_run()` sets `job.run_aborted = True` and immediately writes `FINISHED_ABORTED`.

Because `_job_complete_process()` skips `_update_job_status()` when `job.run_aborted` is true, the existing `exception_run_processes` entry may never be used, and the intended `FINISHED_EXECUTION_EXCEPTION` can be downgraded to `FINISHED_ABORTED` in a multi-client timing window.

The preservation guard in `ServerEngine.wait_for_complete()` only prevents the server job process's secondary exit code from overwriting the recorded exception; it does not protect this later client `REPORT_JOB_FAILURE -> stop_run()` path. Consider preserving any existing non-abort entry in `exception_run_processes` before handling an `ABORTED` report, or having the completion path still consult `_update_job_status()` when an exception marker exists.

IsaacYangSLA:
> Potential status downgrade after `fail_run()` from a later client abort report:
> 
> After a client-side pending timeout, `fail_run()` records `PROCESS_RETURN_CODE = EXCEPTION` in `engine.exception_run_processes` and then aborts the remaining runs. But a later `REPORT_JOB_FAILURE` with `JobReturnCode.ABORTED` from another client is routed through `process_job_failure()` to `job_runner.stop_run()`. `stop_run()` sets `job.run_aborted = True` and immediately writes `FINISHED_ABORTED`.
> 
> Because `_job_complete_process()` skips `_update_job_status()` when `job.run_aborted` is true, the existing `exception_run_processes` entry may never be used, and the intended `FINISHED_EXECUTION_EXCEPTION` can be downgraded to `FINISHED_ABORTED` in a multi-client timing window.
> 
> The preservation guard in `ServerEngine.wait_for_complete()` only prevents the server job process's secondary exit code from overwriting the recorded exception; it does not protect this later client `REPORT_JOB_FAILURE -> stop_run()` path. Consider preserving any existing non-abort entry in `exception_run_processes` before handling an `ABORTED` report, or having the completion path still consult `_update_job_status()` when an exception marker exists.

The status downgrade seems harmless.  Both aborted and exception mean the job could not proceed.  A better way to address those terminal states (job status) is to revisit our specification and define all possible terminal states to cover every cases.  We can improve our design and implementation in future releases.

pcnudde (approved):

nvflare/private/fed/server/fed_server.py
```diff
@@ -801,7 +802,11 @@ def process_job_failure(self, request: Message):
 
         code = payload.get(JobFailureMsgKey.CODE)
         reason = payload.get(JobFailureMsgKey.REASON, "?")
-        if code == ProcessExitCode.UNSAFE_COMPONENT:
+        if code in (ProcessExitCode.CONFIG_ERROR, ProcessExitCode.EXCEPTION, JobReturnCode.EXECUTION_ERROR):
+            with self.engine.new_context() as fl_ctx:
+                self.logger.info(f"Failing job {job_id} due to reported failure from {client}: {reason}")
+                self.engine.job_runner.fail_run(job_id, ProcessExitCode.EXCEPTION, fl_ctx)
+        elif code in (ProcessExitCode.UNSAFE_COMPONENT, JobReturnCode.ABORTED):
```
YuanTingHsieh, line 809: This new ABORTED report path can downgrade a previously recorded exception. If one client first reports EXCEPTION and `fail_run()` records `PROCESS_RETURN_CODE = EXCEPTION`, a later ABORTED report from another client now routes to `stop_run()`, which sets `job.run_aborted = True` and writes `FINISHED_ABORTED`. Since `_job_complete_process()` skips `_update_job_status()` when `run_aborted` is true, the exception marker may never be used. Can we preserve an existing non-abort `exception_run_processes[job_id]` entry before routing ABORTED to `stop_run()`, or make the completion path consult the exception marker first?

IsaacYangSLA: I believe it's the same concern as Peter mentioned.  We can address it in future releases.

PR #4554, opened 2026-05-08
Set start_task_timeout=300 in cyclic CIFAR10 example to avoid 10s default

nvshaxie:
Fixes # .

### Description

The CCWF cyclic CIFAR10 example aborts with `FATAL_SYSTEM_ERROR ("failed to start workflow controller on client site-1")` because `CyclicServerConfig.start_task_timeout` is left at the framework default of 10s (`nvflare/app_common/ccwf/common.py:81 START_TASK_TIMEOUT = 10`), which is shorter than the time the starting client needs to download CIFAR-10 and initialize the PyTorch model after responding to `cyclic_config`. The same issue is reproducible on both `main` and `2.7.2`, so this is a pre-existing example/default mismatch, not a regression.

## Summary
- The cyclic CIFAR10 example does not pass `start_task_timeout` to `CyclicServerConfig`, so it inherits the framework default of 10s in `nvflare/app_common/ccwf/common.py`.
- 10s is too short for the starting client to download CIFAR-10 and init the PyTorch model after responding to `cyclic_config`. The server-side `CyclicServerController` then times out `cyclic_start` and aborts the run with `FATAL_SYSTEM_ERROR`, even though the client is making forward progress.
- Set `start_task_timeout=300` in the example, matching the pattern already used in `examples/advanced/swarm_learning/swarm_pt/job.py` (`start_task_timeout=1200`, with the comment _"Timeouts sized for large-model loading and LoRA training"_).
- Apply the same change to the cyclic notebook tutorial.

## Changes
- `examples/advanced/job_api/pt/cyclic_cc_script_runner_cifar10.py`: pass `start_task_timeout=300` to `CyclicServerConfig`.
- `examples/tutorials/.../07.2.2_cyclic/cyclic_weight_transfer_example.ipynb`: same change in the equivalent notebook cell.

## Repro
Before fix (on a 256-core / A100X host, both `main` and `2.7.2`):
~~~
cd examples/advanced/job_api/pt/
python cyclic_cc_script_runner_cifar10.py
grep "FATAL_SYSTEM_ERROR\|cyclic_start" /tmp/nvflare/jobs/workdir/pt_cyclic/server/log.txt
# produces:
#   WFCommServer - INFO - task cyclic_start exit with status TaskCompletionStatus.TIMEOUT
#   ServerRunner - ERROR - Aborting current RUN due to FATAL_SYSTEM_ERROR received:
#                          failed to start workflow controller on client site-1
# Stable repro across multiple attempts on a clean system.
~~~

[...]

PR #4555, opened 2026-05-08
Fix swarm CIFAR10 example: SwarmServerConfig.min_clients default + start_task_timeout

nvshaxie:
Fixes # .

### Description

The swarm CIFAR10 example fails out of the box on both NVFlare `main` and `2.7.2` with two distinct issues, both of which need to be fixed for the example to complete a single round.

### Bug 1 — `SwarmServerConfig.min_clients` defaulted to `None`
`add_swarm()` forwards every `server_config` field — including this `None` — to `SwarmServerController(min_clients=...)`, overriding the controller's typed default of `int = 0`. The value then reaches `BaseServerCtl.__init__` where `if min_clients < 0:` raises:
~~~
TypeError: '<' not supported between instances of 'NoneType' and 'int'
~~~
The example crashes at construction time, before any simulator is launched.

The other `Optional[List[str]]` fields (`aggr_clients`, `train_clients`, `result_clients`) tolerate `None` because the controller normalizes them via `if not x: x = []`. `min_clients` has no such normalization — its type contract is `int`, and the documented "0 means all participating clients are required" semantics in `BaseServerCtl` already covers the "not specified" case.

**Fix:** change `SwarmServerConfig.min_clients` default from `None` to `int = 0`, matching the type and default of both `SwarmServerController.__init__` and `BaseServerCtl.__init__`.

### Bug 2 — `start_task_timeout` left at 10s default
`swarm_script_runner_cifar10.py` did not pass `start_task_timeout`, so the example used the framework default of 10s (`nvflare/app_common/ccwf/common.py:81 START_TASK_TIMEOUT = 10`). After Bug 1 is fixed, the simulator starts and configures both sites, but `swarm_start` times out after 10s while site-1 is still initializing CIFAR-10 + PyTorch model, and the run aborts with `FATAL_SYSTEM_ERROR ("failed to start workflow controller on client site-1")`.

**Fix:** pass `start_task_timeout=300` in the example, matching the pattern already used in `examples/advanced/swarm_learning/swarm_pt/job.py` (`start_task_timeout=1200`). Sibling PR #4554 applies the same example-side fix to the cyclic CIFAR10 example.

## Changes
- `nvflare/app_common/ccwf/ccwf_job.py`: `SwarmServerConfig.min_clients` default `None` → `int = 0`.
- `examples/advanced/job_api/pt/swarm_script_runner_cifar10.py`: pass `start_task_timeout=300` to `SwarmServerConfig`.
- `tests/unit_test/app_common/ccwf/test_ccwf_job_config.py`: new regression test asserting `SwarmServerConfig().min_clients == 0` so the default cannot silently regress to `None`.

## Repro
Before fix (on a 256-core / A100X host, both `main` and `2.7.2`):
~~~
cd examples/advanced/job_api/pt/
python swarm_script_runner_cifar10.py
# Python exits immediately:
# TypeError: '<' not supported between instances of 'NoneType' and 'int'
# in nvflare/app_common/ccwf/server_ctl.py:161 (BaseServerCtl.__init__).
# Workdir does not even contain a pt_swarm subdir.
~~~

After fix (Bug 1 only): the simulator starts, both sites are configured, then:
~~~
WFCommServer - INFO - task swarm_start exit with status TaskCompletionStatus.TIMEOUT
ServerRunner - ERROR - Aborting current RUN due to FATAL_SYSTEM_ERROR received:
                       failed to start workflow controller on client site-1
~~~

After both fixes: a 3-round swarm CCWF run completes in ~6:45 with no `FATAL_SYSTEM_ERROR` lines in the resulting `pt_swarm/server/log.txt`. All 92 existing ccwf unit tests still pass.

### Types of changes
- [x] Quick tests passed locally (`pytest tests/unit_test/app_common/ccwf/` 92 passed; black + isort + flake8 clean on changed files; full `./runtest.sh` deferred to CI).
- [ ] Documentation updated.

### DIFF
--- a/examples/advanced/job_api/pt/swarm_script_runner_cifar10.py
+++ b/examples/advanced/job_api/pt/swarm_script_runner_cifar10.py
@@ -34,7 +34,12 @@ if __name__ == "__main__":
 
     aggregator = InTimeAccumulateWeightedAggregator(expected_data_kind=DataKind.WEIGHTS)
     job.add_swarm(
-        server_config=SwarmServerConfig(num_rounds=num_rounds),
+        server_config=SwarmServerConfig(
+            num_rounds=num_rounds,
+            # Default 10s is too short for clients that initialize CIFAR-10 dataset
+            # and the PyTorch model after responding to swarm_config.
+            start_task_timeout=300,
+        ),
         client_config=SwarmClientConfig(
             executor=ScriptRunner(script=train_script),
             aggregator=aggregator,
--- a/examples/advanced/job_api/pt/swarm_script_runner_cifar10.py
+++ b/examples/advanced/job_api/pt/swarm_script_runner_cifar10.py
@@ -46,7 +46,12 @@ if __name__ == "__main__":
             persistor=PTFileModelPersistor(model=Net()),
             shareable_generator=SimpleModelShareableGenerator(),
         ),
-        cse_config=CrossSiteEvalConfig(eval_task_timeout=300),
+        cse_config=CrossSiteEvalConfig(
+            # Default 10s is too short for clients that need to (re)load CIFAR-10
+            # before responding to cse_start, same race as swarm_start above.
+            start_task_timeout=300,
+            eval_task_timeout=300,
+        ),
     )
 
     # job.export_job("/tmp/nvflare/jobs/job_config")
--- a/nvflare/app_common/ccwf/ccwf_job.py
+++ b/nvflare/app_common/ccwf/ccwf_job.py
@@ -48,7 +48,7 @@ class SwarmServerConfig:
         private_p2p: bool = True,
         aggr_clients=None,
         train_clients=None,
-        min_clients=None,
+        min_clients: int = 0,
     ):
         self.num_rounds = num_rounds
         self.start_round = start_round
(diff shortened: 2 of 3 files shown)


nvshaxie:
@greptile-apps Good catch — `CrossSiteEvalConfig.start_task_timeout` was indeed still at the 10s default. My local test happened to pass (caches were warm by the time CSE ran), but on a slower host or a cold cache the same race would surface on `cse_start`.

Pushed 80a48d3a: pass `start_task_timeout=300` to `CrossSiteEvalConfig` as well, with a comment cross-referencing the swarm_start fix above. Verified the test still passes (~6:48).

YuanTingHsieh:
@nvshaxie this PR changes LGTM.
However the team prefers create fix PR against 2.8 branch first then cherry-pick them back to main, can you create the same PR to 2.8 first? thanks!

nvshaxie:
Closing per @YuanTingHsieh's review feedback to retarget at the 2.8 branch first.

Re-opened against `2.8` here: https://github.com/NVIDIA/NVFlare/compare/2.8...nvshaxie:NVFlare:fix/swarm-min-clients-and-start-timeout-2.8

Will follow up with a cherry-pick PR to `main` once the 2.8 PR is merged.

PR #4558, opened 2026-05-08
[2.8] Fix _LogTailProducer EXECUTION_EXCEPTION ERROR at job startup

nvidianz:
## Summary
At every job start on 2.8, the per-site job log records a misleading line:

```
_LogTailProducer - ERROR - error from target server: EXECUTION_EXCEPTION
```

It is a symptom of a cross-process race introduced by the live-log-streaming feature in #4454: the client's `JobLogStreamer._on_job_started` fires on `START_RUN` and immediately sends its first chunk, but `JobLogReceiver._register` also fires on `START_RUN` in a separate **server-side** job subprocess. If the client wins the race, the server's `stream_runner` has no factory for `channel='log_streaming' topic='live_log'` registered yet and replies `EXECUTION_EXCEPTION`.

The base `BaseChunkProducer.process_replies` treats any non-OK reply as fatal: it logs `ERROR` and returns `False`, which `stream_runner` (`nvflare/private/stream_runner.py:461-475`) then interprets as end-of-stream — so the entire job's live-log stream is torn down on the very first chunk. The user-visible ERROR line is only one symptom; the silent loss of every subsequent log byte is the other.

This breaks any CI/QA pipeline that asserts per-site logs contain no `ERROR` / `Traceback` / `Exception` strings (the bug reports it turning 12/17 multi-study container testcases into failures), and it silently disables the live-log feature for affected jobs.

## Fix
1. **Bootstrap-tolerant `_LogTailProducer` (primary fix)** — override `process_replies` so that until the producer has seen the first OK reply, a uniform `EXECUTION_EXCEPTION` is treated as a transient miss: the file offset is rolled back to before the failed read so the same bytes will be re-emitted on the next `produce()`, the failure is logged at debug rather than ERROR, and `None` is returned to keep the stream alive. After the first OK reply, falls through to the base behavior so genuine receiver failures still surface as ERROR and end the stream. Heartbeats don't move the file pointer, so no seek is attempted for them.
2. **Earlier receiver registration (defense in depth)** — `JobLogReceiver` now registers on `ABOUT_TO_START_RUN` instead of `START_RUN`, so in the common case the handler is wired up before the client's first chunk arrives. Cross-process timing means this alone can't fully close the race, hence (1) is the real fix.

## Why this shape
- Surgical: changes are scoped to `_LogTailProducer` and `JobLogReceiver`. `BaseChunkProducer` and other streamers (`FileStreamer`, `TensorProducer`, etc.) are untouched, so genuine `EXECUTION_EXCEPTION` errors still surface where they should.
- Recovers lost bytes via `seek` instead of just hiding the symptom, so the live-log feature actually works on affected jobs.
- Race-free regardless of cross-process event ordering — no reliance on SP_JOB / parent / future deploy modes happening in any particular sequence.
- Explicit "first send only" gate via `_first_ok_received` — no risk of swallowing real receiver crashes mid-job.

## Behavior matrix
| Scenario | Before | After |
|---|---|---|
| First chunk OK | OK | OK |
| First chunk transient `EXECUTION_EXCEPTION`, then OK | ERROR logged, stream torn down, rest of log lost | bytes re-emitted, stream continues, no ERROR |
| `EXECUTION_EXCEPTION` after first OK reply | ERROR logged, stream ends | unchanged (ERROR logged, stream ends) |
| Persistent `EXECUTION_EXCEPTION` (real receiver failure) | ERROR on first chunk, stream ends | second `EXECUTION_EXCEPTION` (post-first-OK gate flips) → ERROR, stream ends |

## Tests
New tests in `tests/unit_test/app_common/streamers/log_streamer_test.py`:
- `test_first_chunk_execution_exception_is_transient_and_retries` — bootstrap miss rolls offset back, no ERROR logged, bytes re-emitted on next `produce()`.
- `test_first_chunk_execution_exception_on_heartbeat_does_not_seek` — heartbeats don't trigger seek (last_was_data=False).
- `test_execution_exception_after_first_ok_still_ends_stream` — steady-state strict failure preserved.
- `test_first_chunk_ok_flips_strict_mode` — first OK reply flips the producer into strict mode.

## Verified locally
- `pytest tests/unit_test/app_common/streamers tests/unit_test/app_common/logging -v` → **43 passed** (Python 3.12.10, pytest 8.3.4) — 4 new + 39 existing, no regressions.
- `black`, `isort`, `flake8` clean on all changed files.

## Test plan
- [ ] Manual: reproduce the bug's POC repro on 2.8 head, confirm `grep ERROR /tmp/nvflare/poc/example_project/prod_00/site-1/$JOB_ID/log.txt` returns no lines after the fix.

### Types of changes
- [x] Quick tests passed locally by running unit tests for `app_common/streamers/` and `app_common/logging/`.
- [ ] Documentation updated.


YuanTingHsieh (approved):

nvshaxie:
**Follow-up: symmetric teardown race still emits the same ERROR**

Verified on 2.8 `949dcf3b` (post-#4558 + #4559 merge). The bootstrap-tolerant `process_replies` override correctly suppresses the **startup** race. However, the **symmetric teardown** race still emits exactly one `_LogTailProducer - ERROR - error from target server: EXECUTION_EXCEPTION` per site per job.

**Why the bootstrap gate doesn't catch it:** during normal job execution the receiver replies OK ≥1 time, so `_first_ok_received` flips True. At END_RUN, the server's `JobLogReceiver` is unregistered **before** the client's final chunk arrives → trailing chunk gets `EXECUTION_EXCEPTION` past the gate → falls through to `super().process_replies()` → base-class ERROR.

**Timing on per-site log** (hello-numpy POC, hello_cyclic also reproduces):

```
HH:MM:43  task result sent to server
HH:MM:44  ABOUT_TO_END_RUN / CHECK_END_RUN_READINESS fired
HH:MM:44  _LogTailProducer - ERROR - error from target server: EXECUTION_EXCEPTION   ← teardown race
HH:MM:45  END_RUN fired
```

**Repro** (any POC-mode job, post-#4558+#4559, ~30s):

```bash
echo y | nvflare poc prepare
nvflare poc start -ex admin@nvidia.com &
# wait "Ready to go"
nvflare job submit -j /tmp/nvflare/jobs/<job_config>
# wait FINISHED:COMPLETED
grep _LogTailProducer .../site-*/<job_id>/log.txt
```

**Impact:** the same regression cases that originally hit the startup race still fail the `^((?!ERROR).)+$` per-site log gate — just with 1 ERROR at END_RUN instead of START_RUN. Server log is fully clean.

nvshaxie:
**Correction to my earlier comment** — I said "server log is fully clean" but I only checked `server/log.txt`. The POC stdout `<workspace>/poc/example_project/prod_00/server/poc_console.log` also has ERROR lines from the same race:

```
ObjectStreamer - ERROR - [... channel='log_streaming' topic='live_log' seq=N]
   no stream processing info registered for log_streaming:live_log
```

22 such lines per job (2 at startup, ~20 at teardown). Verified on origin/2.8 `01355b61`.

PR #4559, opened 2026-05-08
[2.8] Fix StorageException after every job from empty SystemLogStreamer context

nvidianz:
## Summary
After PR #4476, every successful federated job emits this error in the server log:

```
ObjectStreamer - ERROR - exception from stream_done_cb dispatch_stream_done:
StorageException: path /tmp/nvflare/jobs-storage/unknown is not a valid directory.
```

(plus, in some cases, an earlier `no stream processing info registered for log_streaming:live_log` line during the run.)

## Root cause
`SystemLogStreamer._on_job_completed` runs on `CLIENT_PARENT` and uploads the per-job `error_log.txt` snapshot to the server. It hands the upload thread a fresh context from `engine.new_context()`. **CLIENT_PARENT serves multiple jobs over its lifetime, so its engine ctx_manager does NOT carry per-job `RUN_NUM` / `IDENTITY_NAME` as sticky props** — and the fresh context arrives at the receiver with an empty `job_id`.

`JobLogReceiver._get_trusted_stream_identity` reads `job_id` via `peer_ctx.get_job_id(default="unknown")` from the sender-forwarded context. With the empty fresh context, `peer_ctx.get_job_id()` returns `""` and the receiver falls back to the literal `"unknown"`, then asks the job manager to save the log under `.../jobs-storage/unknown`, which fails because it isn't a real path.

The bug was dormant before #4476: `StaticFileBuilder._modify_system_log_streamer` used to strip `system_log_streamer` from deployed client `resources.json` by default. #4476 removed that filter and flipped the default to `allow_log_streaming=True`, so the dormant buggy code now runs after every job.

## Fix
In `SystemLogStreamer._on_job_completed`, explicitly seed `RUN_NUM` and `IDENTITY_NAME` on the fresh stream context before handing it to the upload thread. The receiver continues to read job identity from the sender-forwarded `peer_ctx` as before — **preserving the existing peer-context-as-trust-source security model** and leaving `JobLogReceiver._get_trusted_stream_identity` unchanged.

```python
engine = fl_ctx.get_engine()
stream_fl_ctx = engine.new_context() if engine else fl_ctx
if stream_fl_ctx is not fl_ctx:
    stream_fl_ctx.put(key=ReservedKey.RUN_NUM, value=job_id, private=True, sticky=False)
    stream_fl_ctx.put(key=ReservedKey.IDENTITY_NAME, value=client_name, private=True, sticky=False)
```

## Why sender-side instead of receiver-side
The bug originally proposed an "Option A" (read `job_id` from `stream_ctx` in the receiver) but that approach was rejected after reviewing the receiver's existing security test:

`tests/unit_test/app_common/logging/job_log_receiver_test.py::test_job_log_receiver_uses_trusted_peer_identity_for_storage` deliberately passes **forged** `stream_ctx` values (`"../../forged_job"`, `"../../forged_client"`) and asserts the receiver routes storage paths via `peer_ctx` (authenticated cellnet identity), **not** `stream_ctx` (sender-supplied data). The function name `_get_trusted_stream_identity` and the test name make the trust boundary explicit.

`stream_ctx`-based routing would let a malicious client spoof another job's id and write log content into that other job's storage directory. Fixing the bug at its real source — the sender that handed off an empty context — preserves the security model and makes the existing security test continue to pass unchanged. Confirmed by running it as part of the test suite below.

## Tests
- **New**: `test_completed_upload_seeds_run_num_and_identity_on_fresh_context` — regression test that verifies the fresh context handed to the upload thread now carries the real `job_id` and `client_name`.
- **Existing security test passes unchanged**: `test_job_log_receiver_uses_trusted_peer_identity_for_storage` — confirms this fix does not relax the receiver's authentication model.
- Existing `SystemLogStreamer` tests unchanged.

## Verified locally
- `pytest tests/unit_test/app_common/logging -v` → **13 passed** (12 existing + 1 new) on Python 3.12.10 / pytest 8.3.4.
- `black`, `isort`, `flake8` clean on all changed files.

## Test plan
- [ ] Manual: reproduce the bug's repro on 2.8 head, confirm `grep "StorageException.*jobs-storage/unknown" .../server/log.txt` returns no lines after the fix.

### Types of changes
- [ ] Breaking change.
- [x] Quick tests passed locally by running unit tests for `app_common/logging/`.
- [ ] Documentation updated.


nvflare/app_common/logging/system_log_streamer.py
```diff
@@ -232,8 +239,21 @@ def _on_job_completed(self, event_type: str, fl_ctx: FLContext):
...
+        # context produced by engine.new_context() arrives at the receiver
+        # with empty peer_ctx.get_job_id(), and JobLogReceiver routes the
+        # uploaded error log under the literal "unknown" job_id — which then
+        # fails with `StorageException: path .../jobs-storage/unknown is not
+        # a valid directory`. Seeding these reserved keys preserves the
+        # receiver's existing peer-context-as-trust-source security model.
         engine = fl_ctx.get_engine()
         stream_fl_ctx = engine.new_context() if engine else fl_ctx
+        if stream_fl_ctx is not fl_ctx:
+            stream_fl_ctx.put(key=ReservedKey.RUN_NUM, value=job_id, private=True, sticky=False)
```
YuanTingHsieh: AuxRunner forwards peer identity with fl_ctx.get_all_public_props(), and the receiver rebuilds peer_ctx from those public props. These new RUN_NUM and IDENTITY_NAME values are written with private=True, so they remain local to the upload thread context and are not included in the peer props sent to JobLogReceiver. The regression test only checks the local context, so it passes while the actual stream transport still sees missing or empty job identity. These should be public, matching FLContextManager.new_context() for job contexts, and the test should assert stream_fl_ctx.get_all_public_props() or exercise the aux/receiver path.

pcnudde (approved):

YuanTingHsieh (approved):

PR #4577, opened 2026-05-12
fix fd leak in GenerateDataFromFile on send error

cyyever:
### Description

This PR fixes fd leak in GenerateDataFromFile on send error.
### Types of changes

- [X] New tests added to cover the changes.
- [ ] Documentation updated.


pcnudde:
if commits are not signed we can not merge them.

cyyever:
@pcnudde Signed it.

chesterxgchen:
1.  One observation: test_fd_closed_on_body_size_mismatch mutates gen.size = 999_999 after construction to
   fake a mismatch. This works and is acceptable for this kind of unit test, but it's relying on
  implementation internals. A comment noting why would help future readers (# force size mismatch to
  exercise the post-loop RuntimeError path).

 The  suggestion is a one-line comment in test_fd_closed_on_body_size_mismatch to explain the gen.size
  mutation. 
  
2. nvflare/fuel/hci/binary_proto.py:394: with closing(generator) makes send_binary_data() require every generator object to implement close(). Before this PR, the runtime contract was effectively duck-
    typed: anything with data_size() and generate() worked. A custom generator that does not subclass DataGenerator can now successfully send the full payload, then fail on exit with AttributeError: ... has no attribute 'close'. To preserve compatibility, use a try/finally and call close() only when present:

    close = getattr(generator, "close", None)
    try:
        ...
    finally:
        if close:
            close()

chesterxgchen:
@cyyever can you address the comments

cyyever:
@chesterxgchen fixed

chesterxgchen (approved):
@cyyever can you cherry pick this PR to 2.8 branch as well

PR #4583, opened 2026-05-12
[2.8] Fix tracking recipe integration test

YuanTingHsieh:
## Summary
- Replace the dynamic `_exp_tracking_model` import in the experiment tracking recipe smoke tests with the documented dict model config.
- Bundle the example `model.py` into the server app so `model.SimpleNetwork` resolves when the job runs.
- Add a dedicated `tracking` integration backend and route it through the PyTorch CI setup.
- Refresh the integration test README/backend list and stale manual pytest command.

## Why
The tests loaded the example model under a synthetic module name and then passed that live `nn.Module` instance into `FedAvgRecipe`. During job export, Python source inspection could not resolve the synthetic module and raised:

```text
TypeError: <class '_exp_tracking_model.SimpleNetwork'> is a built-in class
```

Using `model={"class_path": "model.SimpleNetwork", "args": {}}` matches the recipe API and avoids relying on importlib state.


tests/integration_test/test_configs.yml
```diff
@@ -34,5 +34,7 @@ test_configs:
   standalone: []
 
 pytest_files:
+  tracking:
+    - experiment_tracking_recipes_test.py
```
pcnudde: Shoudl we just add it to the standalone tests? My idea was there that standalone would be python tests that do not rely on the rest of the integration infrastructure

YuanTingHsieh:
Scope update after review: this PR has been narrowed to `tests/integration_test/experiment_tracking_recipes_test.py` only. It no longer adds a `tracking` integration backend or changes CI runner wiring.

Current fix:
- replace the synthetic `_exp_tracking_model` import with the recipe dict model config `model.SimpleNetwork`
- bundle `model.py` into both server and client apps, since the server persistor and client script both need to import it
- refresh the stale manual pytest filename in the test docstring

Validated locally:
- `python3 -m py_compile tests/integration_test/experiment_tracking_recipes_test.py`
- `git diff upstream/2.8...HEAD --check`

pcnudde (approved):

PR #4586, opened 2026-05-13
[2.8] Fix JobLogReceiver not registering in server-job subprocess

nvidianz:
## Summary

Restores live log streaming on the 2.8 branch. After #4558 the receiver
silently stops registering its `log_streaming:live_log` handler in
per-job server subprocesses, so every chunk from a client's
`JobLogStreamer` is rejected with:

```
ObjectStreamer - ERROR - [tx_id='…' channel='log_streaming' topic='live_log' seq=0] no stream processing info registered for log_streaming:live_log
_LogTailProducer - ERROR - error from target server: EXECUTION_EXCEPTION
```

The bootstrap-tolerant retry from #4558 hides the first 10 chunks and
then surfaces the ERROR, so it looks like a teardown race in short jobs;
on a longer job the first heartbeat (default `liveness_interval=10s`)
fires mid-run, well before any shutdown — which is what tipped this off.

## Root cause

`JobLogReceiver.__init__` registers `_register` on
`[SYSTEM_START, ABOUT_TO_START_RUN]`. Neither event fires in the
server-job subprocess spawned by `ServerProcessJobLauncher`:

- `SYSTEM_START` is only fired by `server_deployer.deploy()`, which the
  parent process runs. The subprocess enters through
  `nvflare/private/fed/app/server/runner_process.py:99`
  (`deployer.create_fl_server(...)`, **not** `deploy()`).
- `ABOUT_TO_START_RUN` is only fired in
  `nvflare/private/fed/client/client_runner.py:739` — client-side only.

The only server-job lifecycle event is `START_RUN`
(`nvflare/private/fed/server/server_runner.py:188`), which the receiver
no longer listens to. Result: the subprocess's `JobLogReceiver` instance
has `_registered=False` for the entire job and never calls
`engine.register_stream_processing`, leaving the run's
`ObjectStreamer.registry` empty for the live-log channel.

The `_registered` latch is also wrong for in-process deployments where
`engine.set_run_manager(...)` swaps in a new `RunManager`/`ObjectStreamer`
per job: registering once against the parent's streamer would leave each
new job's registry empty even if the trigger events did fire.

## Fix

`nvflare/app_common/logging/job_log_receiver.py`:

1. Trigger `_register` on **`SYSTEM_START`, `ABOUT_TO_START_RUN`, and
   `START_RUN`** so the receiver wires up wherever it happens to live.
2. Drop the `_registered` early-return — `Registry.set` is idempotent
   for the same channel/topic, and re-registering after a run-manager
   swap is the desired behavior.

Docstring updated to describe the new placement-aware behavior.

## Why this shape

- **Restores the original (pre-#4558) server-subprocess behavior.** The
  bootstrap-tolerant `_LogTailProducer.process_replies` from #4558 is
  the actual fix for the client/server startup race; the
  `START_RUN` → `ABOUT_TO_START_RUN` event change was only listed as
  "defense in depth" and turns out to disable registration entirely in
  the subprocess. Reverting just the event list keeps the real fix
  intact.
- **Surgical.** Two-file change, no churn outside `JobLogReceiver`.
- **Forward-compatible.** Listing all three events also covers
  in-process deployments where the run manager is swapped per job.

## Tests

`tests/unit_test/app_common/logging/job_log_receiver_test.py` — 3 new:

- `test_register_fires_on_run_lifecycle_events[_system_start|_about_to_start_run|_start_run]` —
  every trigger event invokes `register_stream_processing` with the
  expected channel/topic.
- `test_register_reregisters_on_each_event` — repeat triggers all
  re-register, no "registered once" latch.
- `test_register_event_handlers_cover_server_subprocess_path` —
  regression guard so `START_RUN` cannot be silently dropped from the
  trigger list again.

Existing security/test coverage (peer-identity routing, site
`allow_log_streaming` enforcement) unchanged.

## Verified locally

- `pytest tests/unit_test/app_common/logging tests/unit_test/app_common/streamers -v`
  → **50 passed** (Python 3.10.17, pytest 7.4.3).
- `black`, `isort`, `flake8` clean on changed files.

## Test plan

- [ ] Manual: run an auth_test-style integration job; confirm
      `grep "no stream processing info registered" /tmp/.../log.txt`
      returns no lines after the fix and that
      `JobLogReceiver - Saving live log` lines appear for both the
      in-job stream and the post-completion error_log upload.

### Types of changes
- [x] Quick tests passed locally by running unit tests for `app_common/logging/` and `app_common/streamers/`.
- [ ] Documentation updated.

### DIFF
--- a/nvflare/app_common/logging/job_log_receiver.py
+++ b/nvflare/app_common/logging/job_log_receiver.py
@@ -55,17 +55,20 @@ class JobLogReceiver(Widget):
     **Job-level configuration** (``config_fed_server.json``)
         Add it via ``job.to_server(JobLogReceiver())`` in the Job API, or
         declare it in the job's server config.  In this mode the handler is
-        registered on ``ABOUT_TO_START_RUN``, which fires immediately before
-        ``START_RUN`` so that the stream handler is wired up before any client
+        registered on ``START_RUN`` (and on ``ABOUT_TO_START_RUN`` where
+        available) so that the stream handler is wired up before any client
         chunk can arrive.  The widget is only active for that specific job.
 
     **System-level resources** (``resources.json`` on the server)
         Declare it as a system component so it is instantiated when the server
-        process starts.  In this mode the handler is registered on
-        ``SYSTEM_START`` and remains active for every job that runs on that
-        server for the lifetime of the process.
+        process starts.  In the long-lived parent process the handler is
+        registered on ``SYSTEM_START`` and remains active across jobs.  In
+        per-job server subprocesses (where ``SYSTEM_START`` does not fire) the
+        same instance re-registers on every ``START_RUN`` against the new
+        job-scoped ``ObjectStreamer``.
 
-    Regardless of placement, the stream handler is registered exactly once.
+    The stream handler may be (re-)registered on every triggering event;
+    ``registry.set`` is idempotent for the same channel/topic pair.
 
     Args:
         dest_dir: directory where incoming log files are written.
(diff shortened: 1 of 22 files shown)


nvflare/app_common/logging/site_log_streamer.py
YuanTingHsieh, line 1: I like SiteLogStreamer better as a name. SystemLogStreamer made me think it was global to the whole NVFlare system, but this is really site-level behavior.

nvflare/app_common/logging/job_log_receiver.py
```diff
@@ -79,20 +82,27 @@ def __init__(self, dest_dir: str = None, idle_timeout: float = 30.0):
...
+        #   - START_RUN is the only one that fires in the server-job subprocess
+        #     (see ServerRunner.run) and is also the point at which an in-process
+        #     deployment swaps engine.run_manager to the job's RunManager.
+        # Re-registering on each trigger is safe: the underlying registry.set
+        # is idempotent for the same channel/topic, and registering against a
+        # new ObjectStreamer is exactly what we want when the run manager has
+        # been replaced.
+        self.register_event_handler(
+            [EventType.SYSTEM_START, EventType.ABOUT_TO_START_RUN, EventType.START_RUN],
+            self._register,
```
YuanTingHsieh, line 104: I wonder if we should fire ABOUT_TO_START_RUN in SJ as well, immediately before ServerRunner fires START_RUN. That would make the server-job lifecycle symmetric with CJ and let job-level components register pre-start hooks consistently, instead of requiring components like JobLogReceiver to special-case SJ by also listening to START_RUN.

nvidianz: We should but we should do that in another PR. In general, we should make all the components life cycle exactly the same.

nvflare/app_common/logging/site_log_streamer.py
nvidianz: Good suggestion. Changed.

YuanTingHsieh (approved):

PR #4597, opened 2026-05-14
Bind auth tokens to runtime origins

pcnudde:
## Summary

- Bind client authentication tokens to their registered CellNet origin during server-side auth header validation.
- Stop leaking client bearer credentials to peer clients while preserving routed client-client replies.
- Reject forged task result submissions where the submitting client does not match the task assignment, and ignore duplicate task results after the first accepted result.

## Runtime Auth Model

Auth headers are for the server trust boundary, not for peer clients.

- Requests continue to carry auth headers so server-path traffic can be authenticated and origin-bound.
- A peer reply routed through the server adds auth headers because the next hop is the server.
- The server validates that reply normally, including token signature and origin binding.
- If the validated reply is only transiting to a non-server peer, the server strips `client_name`, `__token__`, `__token_signature__`, and `ssid` before forwarding.
- The destination peer accepts the reply through the existing CellNet reply correlation: matching `REQ_ID` and expected reply origin/target. It does not receive the sender's bearer material.

This avoids both failure modes:

- The server does not trust sender-controlled `MSG_TYPE=REPLY` as an auth bypass.
- Peer clients do not receive another client's runtime bearer credentials.

## Root Cause / Impact

Client replies to peer-originated requests could carry server-auth credentials. A peer that observed those credentials could replay server-bound traffic from its own origin because the server validated token/name/signature but did not bind that identity to the message origin. Separately, workflow result submission accepted a valid task id without verifying that the submitting client was the assigned client.

From the full FLARE system view, this could let a compromised client site impersonate another client for authenticated server-path traffic or submit task results for tasks it was not assigned if it obtained a valid task id. The fix keeps compatibility for existing auth validation paths while enforcing origin binding when the server can resolve a token/name to a registered client FQCN.

## Tests

- `tests/unit_test/fuel/sec/authn_test.py`
- `tests/unit_test/private/fed/authenticator_test.py`
- `tests/unit_test/apis/impl/controller_test.py`
- `tests/unit_test/private/fed/server/fed_server_test.py`


nvflare/private/fed/authenticator.py
```diff
@@ -344,6 +355,13 @@ def validate_auth_headers(message: CellMessage, token_verifier: TokenVerifier, l
         logger.error(f"{err_text}: {err}")
         return make_cellnet_reply(rc=F3ReturnCode.UNAUTHENTICATED, error=err)
 
+    if client_fqcn_resolver:
+        client_fqcn = client_fqcn_resolver(client_name, token)
+        if client_fqcn and not _origin_matches_fqcn(origin, client_fqcn):
+            err = f"auth token for client {client_name} does not match message origin {origin}"
+            logger.error(f"{err_text}: {err}")
+            return make_cellnet_reply(rc=F3ReturnCode.UNAUTHENTICATED, error=err)
```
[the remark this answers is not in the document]
pcnudde: Addressed in 01c40c166. The resolver now distinguishes the compatibility fallback from a registered client with missing origin: _resolve_client_fqcn_for_auth returns MISSING_CLIENT_FQCN for the registered-but-missing-FQCN case, and validate_auth_headers treats any non-None resolver result as enforceable, so the empty sentinel fails closed. Added coverage in authenticator_test and fed_server_test for this edge case.

nvflare/fuel/sec/authn.py
```diff
@@ -40,8 +46,17 @@ def add_authentication_headers(msg: Message, client_name: str, auth_token, token
     msg.set_header(CellMessageAuthHeaderKey.TOKEN_SIGNATURE, token_signature if token_signature else "NA")
 
 
+def add_server_path_reply_authentication_headers(
+    msg: Message, client_name: str, auth_token, token_signature, ssid=None
+):
+    origin = msg.get_header(MessageHeaderKey.ORIGIN)
+    destination = msg.get_header(MessageHeaderKey.DESTINATION)
+    if _is_server_fqcn(origin) or _is_server_fqcn(destination):
```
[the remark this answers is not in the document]
pcnudde: Addressed in 01c40c166 with a clarification and test coverage. I did not remove the origin-side server-path check because this filter is also installed by server-owned job/transfer cells, not only regular clients. For those cells, replies can originate from a server FQCN while the destination is a client path, and they still need auth headers. Added test_auth_filter_keeps_auth_on_replies_from_server_path_origin to lock that behavior down.

pcnudde:
@greptileai review

nvflare/fuel/sec/authn.py
```diff
@@ -13,10 +13,16 @@
 # limitations under the License.
 from nvflare.apis.fl_constant import CellMessageAuthHeaderKey
 from nvflare.fuel.f3.cellnet.cell import Cell
+from nvflare.fuel.f3.cellnet.defs import MessageHeaderKey
+from nvflare.fuel.f3.cellnet.fqcn import FQCN
 from nvflare.fuel.f3.message import Message
 from nvflare.fuel.utils.validation_utils import check_object_type, check_str
 
 
+def _is_server_fqcn(fqcn: str) -> bool:
```
nvidianz, line 22: This can also be done like this,

`FqcnInfo(fqcn).is_on_server`

nvidianz (dismissed):

nvflare/fuel/sec/authn.py
```diff
@@ -40,8 +46,18 @@ def add_authentication_headers(msg: Message, client_name: str, auth_token, token
     msg.set_header(CellMessageAuthHeaderKey.TOKEN_SIGNATURE, token_signature if token_signature else "NA")
 
 
+def add_server_path_reply_authentication_headers(
+    msg: Message, client_name: str, auth_token, token_signature, ssid=None
+):
+    origin = msg.get_header(MessageHeaderKey.ORIGIN)
+    destination = msg.get_header(MessageHeaderKey.DESTINATION)
+    # This filter is also used by server-owned job/transfer cells: keep auth on replies from or to server paths.
+    if _is_server_fqcn(origin) or _is_server_fqcn(destination):
```
nvidianz: What happens to client-client communications? They might be routed through server but with no auth_token. Why don't we add token to all messages?

nvflare/private/fed/authenticator.py
```diff
@@ -299,12 +302,24 @@ def authenticate(self, shared_fl_ctx: FLContext, abort_signal: Signal):
         return token, token_signature, ssid, token_verifier
 
 
-def validate_auth_headers(message: CellMessage, token_verifier: TokenVerifier, logger):
+def _origin_matches_fqcn(origin: str, fqcn: str) -> bool:
+    return bool(origin) and bool(fqcn) and (origin == fqcn or FQCN.is_ancestor(fqcn, origin))
```
nvidianz, line 306: NVFlare supports hierarchical sites — a FQCN can look like 
`site-1.site-11.job_id`
Everything under site-1 is treated as the same site by this check, since _origin_matches_fqcn accepts any descendant of the registered FQCN. I think that's ok.

nvflare/fuel/sec/authn.py
pcnudde: Done in `1d81a37e3`: `_is_server_fqcn()` now delegates to `FqcnInfo(fqcn).is_on_server`.

nvflare/fuel/sec/authn.py
pcnudde: Good catch. I added a red/green regression test for this in `1451a4225`: `test_client_to_client_reply_routed_through_server_is_not_blocked_by_server_auth` forces a client-client request/reply through the server. Before the fix it timed out because the server auth filter rejected the peer reply as missing auth headers.

I did not add tokens to all replies because that would reintroduce the credential disclosure this PR is fixing: a peer could elicit a reply and read the sender's live bearer credentials from the reply headers. Instead, the server auth filter now skips validation only for transit `REPLY` messages whose destination is not server-owned. Server-destined messages still require auth, and peer transit requests still require auth before forwarding.

nvflare/private/fed/authenticator.py
pcnudde: Agreed. I kept that behavior and added explicit coverage in `1d81a37e3`: `test_validate_auth_headers_accepts_token_from_registered_origin` now includes `site-a.site-a-child.job-1`, with a comment noting that job and hierarchical child cells under a registered site remain inside that site's trust boundary.

nvflare/private/fed/server/fed_server.py
```diff
@@ -419,6 +419,17 @@ def _validate_auth_headers(self, message: Message):
...
+        destination = message.get_header(MessageHeaderKey.DESTINATION)
+        # Peer replies can transit the server after reply auth headers are suppressed to avoid credential leaks.
+        # Peer requests still carry auth headers and are validated below before forwarding.
+        if (
+            message.get_header(MessageHeaderKey.MSG_TYPE) == MessageType.REPLY
+            and destination
+            and not FqcnInfo(destination).is_on_server
+        ):
+            self.logger.debug(f"skip auth validation for transit reply to {destination}")
+            return None
```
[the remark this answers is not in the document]
pcnudde: Agreed. Addressed in `7c0822b5e`: the server no longer skips auth based on `MSG_TYPE=REPLY` alone.

The server now records an expected peer-reply key only after it has successfully validated the corresponding peer `REQ` transit message. The unauthenticated peer reply is allowed only if it matches that recorded reverse leg, and the key is consumed on first use. So a client-crafted `REPLY` with no prior validated request falls through to normal `validate_auth_headers()` and is rejected.

Added coverage for the exact concern: an untracked unauthenticated client reply transit is rejected, a matching reply after a validated request is allowed, and the tracked reply key is consumed so it cannot be replayed.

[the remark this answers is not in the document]
pcnudde: Follow-up: I replaced the correlation-table approach with a simpler auth-to-server / strip-to-peer model in `b375f9c43`.

The server no longer skips auth for peer replies. Instead, a peer reply routed through the server carries auth headers because its next hop is the server. The server validates the reply normally, including origin binding, then strips the bearer auth headers before forwarding to the non-server peer. So `MSG_TYPE=REPLY` is not an auth bypass, and the destination peer still does not receive another client's runtime credentials.

I also updated the PR description to make this model explicit and added/updated tests covering the routed client-client reply, server stripping after validation, rejection of unauthenticated peer replies, and preserving auth for server-destined replies.

pcnudde:
@greptileai review

nvflare/fuel/sec/authn.py
pcnudde: updated this code.

nvidianz (approved):

PR #4615, opened 2026-05-14
CPU-only torch backend for github pre-merge workflow

IsaacYangSLA:
### Description

Since github workers don't have GPU, the torch backend can be cpu-only.  This PR switches the pre-merge workflow to install cpu-only pytorch and torchvision.  It should speed up the dependency installation process.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/.github/workflows/premerge.yml
+++ b/.github/workflows/premerge.yml
@@ -37,7 +37,7 @@ jobs:
         with:
           python-version: "3.14"
       - name: Run style checks
-        run: ./runtest.sh -s
+        run: ./runtest.sh --torch-backend=cpu -s
 
   unit-tests:
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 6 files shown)


pcnudde:
Do you see any speed gain by just installing cpu?

IsaacYangSLA:
The entire workflow includes downloading packages, installing them, and run unit tests.  The sizes of downloaded packages differ a lot.  CPU only is ~500MB while cuda-enabled packages are about a few GB.  Installing seems just a few seconds faster, thanks to uv.  Running unit tests is almost the same because those cude-enabled packages fall back to cpu.

The main reason of this PR is to avoid downloading large python packages frequently.  I believe github penalizes such  workflows by slowing them down (or maybe limiting the bandwidth).

YuanTingHsieh (approved):

PR #4620, opened 2026-05-15
Only run pre-merge build-wheel in one environment

IsaacYangSLA:
### Description

NVFlare wheel building process does not have steps to compile extension so there is no need to test different combinations of OS and python version.

This PR will set the github workflow pre-merge/build-wheel to Ubuntu 24.04 and Python 3.11.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/.github/workflows/premerge.yml
+++ b/.github/workflows/premerge.yml
@@ -68,19 +68,22 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-22.04, ubuntu-24.04 ]
-        python-version: [ "3.10", "3.11", "3.12", "3.13", "3.14" ]
+        os: [ ubuntu-24.04 ]
+        python-version: [ "3.11" ]
     steps:
       - uses: actions/checkout@v4
       - name: Set up Python ${{ matrix.python-version }}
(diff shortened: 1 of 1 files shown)


.github/workflows/premerge.yml
```diff
@@ -68,19 +68,22 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ ubuntu-22.04, ubuntu-24.04 ]
-        python-version: [ "3.10", "3.11", "3.12", "3.13", "3.14" ]
+        os: [ ubuntu-24.04 ]
+        python-version: [ "3.11" ]
```
YuanTingHsieh, line 72: changes LGTM, but what version of python should we test?
I suggest we go 3.12 instead of 3.11

pcnudde: That is not enough We need to certainly test 3.14 (there are real changes in that release at least)

IsaacYangSLA: The unit tests still cover all python version.  This PR is for the build-wheel, which does not depend on the python versions.

YuanTingHsieh (approved):

PR #4649, opened 2026-05-20
[2.8] Fix CC runtime workspace path

nvidianz:
## Summary

This fixes the CC startup path used by provisioned NVFlare scripts on the 2.8 branch. When a startup kit is launched from `/user_config`, the generated script now uses an encrypted runtime workspace (`/vault/workspace` for `/user_config`, or `/vault/workspace/<site>` for `/user_config/<site>`) and copies the complete signed startup kit root into `$WORKSPACE/kit`. The runtime workspace exposes `startup/` and `local/` from that copied kit, so `--verify` can validate the complete copied kit while job artifacts such as `FL_global_model.pt` remain under the encrypted runtime workspace and survive restarts.

The change also updates CC documentation to call out the plaintext drives and the encrypted runtime workspace contract, and adds focused template regression checks.

Future note: `/user_config` is expected to become read-only in a future CC image/build update, so this change also moves runtime writes away from a partition that should only carry provisioned configuration.

## Root Cause

The generated `sub_start.sh` used `$DIR/..` as the workspace. In CC deployments where the startup kit lives at `/user_config/nvflare`, that made the runtime workspace `/user_config/nvflare`, so model checkpoints written under the job app directory were stored on the unencrypted user config drive.

## Follow-up Review Fixes

- Uses `/vault/workspace[/<site>]` as the encrypted runtime workspace and `$WORKSPACE/kit` as the copied signed kit root.
- Copies the complete signed startup kit root into `$WORKSPACE/kit` so root `signature.json` remains available for verification.
- Verifies `$WORKSPACE/kit` before creating runtime folders such as `$WORKSPACE/transfer`.
- Refreshes only `$WORKSPACE/kit`, `startup`, and `local` links on restart, preserving job artifacts under `$WORKSPACE`.
- Namespaces nested `/user_config/*` startup kits as `/vault/workspace/<site>` so multiple participants cannot overwrite each other.
- Keeps documentation at the behavior/contract level without documenting startup/local copy mechanics.
- Moved `stop_fl.sh` workspace directory creation into the confirmed shutdown branch so answering `N` has no workspace side effect.

## Validation

- `~/nvflare-env/3.12/bin/python -m pytest -q tests/unit_test/lighter/static_file_builder_test.py`
- `~/nvflare-env/3.12/bin/python -c ...` to run `bash -n` over generated `sub_start_sh` and `stop_fl_sh` template content
- `git diff --check`

### DIFF
--- a/docs/user_guide/confidential_computing/on_premises/cc_deployment_guide.rst
+++ b/docs/user_guide/confidential_computing/on_premises/cc_deployment_guide.rst
@@ -293,6 +293,14 @@ Each startup kit (e.g., ``server1.tgz``) contains:
    * - ``vmlinuz``
      - Linux kernel
 
+.. note::
+
+   The ``user_config.qcow2`` and ``user_data.qcow2`` drives are not encrypted. When a provisioned NVFlare startup
+   script is launched from ``/user_config``, it copies the signed startup/local configuration into
+   ``/vault/workspace`` and uses that path as the runtime workspace. This keeps job artifacts such as
+   model checkpoints on the encrypted root filesystem. Set ``NVFL_WORKSPACE`` before launching the script to choose a
+   different encrypted workspace path.
+
 Step 4: User Data
 -----------------
 
--- a/docs/user_guide/confidential_computing/on_premises/cc_deployment_guide.rst
+++ b/docs/user_guide/confidential_computing/on_premises/cc_deployment_guide.rst
@@ -296,8 +296,8 @@ Each startup kit (e.g., ``server1.tgz``) contains:
 .. note::
 
    The ``user_config.qcow2`` and ``user_data.qcow2`` drives are not encrypted. When a provisioned NVFlare startup
-   script is launched from ``/user_config``, it copies the signed startup/local configuration into
-   ``/vault/workspace`` and uses that path as the runtime workspace. This keeps job artifacts such as
+   script is launched from ``/user_config``, it copies the signed startup/local configuration into a namespaced folder
+   under ``/vault/workspace`` and uses that path as the runtime workspace. This keeps job artifacts such as
    model checkpoints on the encrypted root filesystem. Set ``NVFL_WORKSPACE`` before launching the script to choose a
    different encrypted workspace path.
 
--- a/docs/user_guide/confidential_computing/on_premises/cc_deployment_guide.rst
+++ b/docs/user_guide/confidential_computing/on_premises/cc_deployment_guide.rst
@@ -296,9 +296,8 @@ Each startup kit (e.g., ``server1.tgz``) contains:
 .. note::
 
    The ``user_config.qcow2`` and ``user_data.qcow2`` drives are not encrypted. When a provisioned NVFlare startup
-   script is launched from ``/user_config``, it copies the signed startup/local configuration into a namespaced folder
-   under ``/vault/workspace`` and uses that path as the runtime workspace. This keeps job artifacts such as
-   model checkpoints on the encrypted root filesystem. Set ``NVFL_WORKSPACE`` before launching the script to choose a
+   script is launched from ``/user_config``, runtime artifacts such as model checkpoints are written to a namespaced
+   encrypted workspace under ``/vault/workspace``. Set ``NVFL_WORKSPACE`` before launching the script to choose a
    different encrypted workspace path.
 
 Step 4: User Data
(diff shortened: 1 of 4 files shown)


nvflare/lighter/templates/master_template.yml
```diff
@@ -603,33 +645,33 @@ sub_start_sh: |
...
       return
     fi
     kill -9 $pid
-    rm -f $DIR/../pid.fl $DIR/../shutdown.fl $DIR/../restart.fl 2> /dev/null 1>&2
+    rm -f "$WORKSPACE/pid.fl" "$WORKSPACE/shutdown.fl" "$WORKSPACE/restart.fl" 2> /dev/null 1>&2
   }
 
   if [[ "$doVerify" == "true" ]]; then
-    python3 -m nvflare.tool.verify_startup_kits -f $DIR/../ -c $DIR/rootCA.pem
+    python3 -m nvflare.tool.verify_startup_kits -f "$WORKSPACE" -c "$WORKSPACE/startup/rootCA.pem"
```
YuanTingHsieh: In the /user_config path, the script copies only startup/ and local/ into $WORKSPACE, then verifies $WORKSPACE as a full signed folder. verify_folder_signature() requires a signature.json in the folder being verified, and the redirected workspace root does not get the source root signature.json or signed transfer/, so sub_start.sh --verify will fail before NVFlare starts. This is especially bad for the CC Docker entrypoint, which uses --verify. Verify the copied signed subfolders separately, or verify $SOURCE_WORKSPACE before copying instead of verifying the synthetic runtime workspace.

docs/user_guide/confidential_computing/on_premises/cc_deployment_guide.rst
```diff
@@ -293,6 +293,14 @@ Each startup kit (e.g., ``server1.tgz``) contains:
    * - ``vmlinuz``
      - Linux kernel
 
+.. note::
+
+   The ``user_config.qcow2`` and ``user_data.qcow2`` drives are not encrypted. When a provisioned NVFlare startup
+   script is launched from ``/user_config``, it copies the signed startup/local configuration into a namespaced folder
+   under ``/vault/workspace`` and uses that path as the runtime workspace. This keeps job artifacts such as
+   model checkpoints on the encrypted root filesystem. Set ``NVFL_WORKSPACE`` before launching the script to choose a
+   different encrypted workspace path.
```
YuanTingHsieh, line 301: suggest keeping this note at the behavior/contract level rather than documenting the copy mechanics. Users need to know that /user_config is plaintext and runtime artifacts go to an encrypted workspace; the exact startup//local/ copy behavior feels like implementation detail and may change as we tighten the verification path.

examples/advanced/cc_provision/README.md
```diff
@@ -121,6 +121,11 @@ cd cvm_xxx
 The confidential VM will start, and the NVFLARE server and clients will automatically connect and begin operation.
 You can now use the NVFlare admin console to communicate with the NVFlare system.
 
+> **Note:** `user_config.qcow2` and `user_data.qcow2` are not encrypted. The generated `sub_start.sh` detects when it
```
YuanTingHsieh: suggest keeping this note at the behavior/contract level rather than documenting the copy mechanics. Users need to know that /user_config is plaintext and runtime artifacts go to an encrypted workspace; the exact startup//local/ copy behavior feels like implementation detail and may change as we tighten the verification path.

nvflare/lighter/templates/master_template.yml
```diff
@@ -548,8 +563,72 @@ stop_fl_sh: |
 sub_start_sh: |
   #!/usr/bin/env bash
   DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
-  echo "WORKSPACE set to $DIR/.."
-  mkdir -p $DIR/../transfer
+  SOURCE_WORKSPACE="$( cd "$DIR/.." >/dev/null 2>&1 && pwd )"
+  if [[ -n "${NVFL_WORKSPACE:-}" ]]; then
+    WORKSPACE="$NVFL_WORKSPACE"
+  elif [[ "$SOURCE_WORKSPACE" == "/user_config" ]]; then
```
YuanTingHsieh, line 569: The path-based /user_config detection still feels a bit implicit, but I don’t see a safer small change for this PR. Longer term, it would be cleaner for the CC provisioner/template to render an explicit runtime-workspace setting instead of having the generic startup script infer CC mode from the mount path.

YuanTingHsieh (approved):

PR #4658, opened 2026-05-21
Reject cyclic error results before conversion

YuanTingHsieh:
## Summary
- Reject ordinary cyclic error return codes before converting client payloads into learnables.
- Preserve `EARLY_TERMINATION` payload conversion because the final message may also contain the last training result.
- Stop on allowed early termination after accepting the final learnable.
- Add unit coverage for OK, non-OK error, allowed early termination, disallowed early termination, and non-Shareable results.

## Why
`CyclicController` previously attempted to convert every `Shareable` result before checking the return code. That allows ordinary error replies, such as `EXECUTION_EXCEPTION`, to be parsed as model data and potentially update `_last_learnable`.

`EARLY_TERMINATION` is intentionally treated differently: it can carry the last valid training result, so the controller should convert it before deciding whether to stop the workflow.

### Types of changes

- [ ] Documentation updated.


nvflare/app_common/workflows/cyclic_ctl.py
```diff
@@ -197,6 +197,11 @@ def _process_result(self, client_task: ClientTask, fl_ctx: FLContext):
             # update the global learnable with the received result (shareable)
             # e.g. the received result could be weight_diffs, the learnable could be full weights.
             rc = result.get_return_code()
+            if rc not in [ReturnCode.OK, ReturnCode.EARLY_TERMINATION]:
```
chesterxgchen, line 200: Why exclude early_termination ?

YuanTingHsieh: We exclude EARLY_TERMINATION from the early error-return path because it is not just an error signal in this workflow. The final early-termination message may still contain the last trained model, so we need to convert it to a Learnable before stopping the workflow.

chesterxgchen (approved):

PR #4659, opened 2026-05-21
Support scalar-like analytics values

YuanTingHsieh:
## What changed

- Normalize scalar-like analytics values when constructing `AnalyticsData`.
- Accept Python `int`/`float` values as before, plus scalar-like objects with `.item()` such as `numpy.float32` and 0-D numpy arrays.
- Normalize numeric values inside `SCALARS` and `METRICS` dictionaries as well.
- Add unit coverage for generic scalar-like objects and numpy scalar/0-D array values.

## Why

Client tracking writers currently reject values such as `numpy.float32` for single scalar metrics because analytics validation only accepts built-in Python `float`/`int`. This can happen with code like:

```python
running_loss += cost.cpu().detach().numpy() / images.size()[0]
```

Centralizing the normalization in `AnalyticsData` keeps `SummaryWriter`, `MLflowWriter`, `WandBWriter`, and lower-level analytics logging paths consistent without adding a numpy dependency to core API code.

### DIFF
--- a/nvflare/apis/analytix.py
+++ b/nvflare/apis/analytix.py
@@ -93,7 +93,7 @@ class AnalyticsData:
             sender (LogWriterName): Type of sender for syntax such as Tensorboard or MLflow
             kwargs (optional, dict): additional arguments to be passed.
         """
-        self._validate_data_types(data_type, key, value, **kwargs)
+        value = self._validate_data_types(data_type, key, value, **kwargs)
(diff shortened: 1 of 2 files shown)


chesterxgchen (approved):

YuanTingHsieh:
Superseded by replacement PR #4683, which was recreated from latest `main` after this PR glitched.

PR #4675, opened 2026-05-22
[2.8] fix fd leak in GenerateDataFromFile on send error

chesterxgchen:
Cherry-pick of #4577 onto the 2.8 branch.

## Summary

Fixes a file descriptor leak in `GenerateDataFromFile` when a send error occurs. Uses `contextlib.closing` to ensure the file handle is always released regardless of whether the send succeeds or fails.

## Original PR

#4577 (merged to main)

## Test plan

- [ ] Existing unit tests pass
- [ ] No new fd leaks under error injection on send path

### DIFF
--- a/nvflare/fuel/hci/binary_proto.py
+++ b/nvflare/fuel/hci/binary_proto.py
@@ -360,6 +360,9 @@ class DataGenerator(ABC):
         """
         pass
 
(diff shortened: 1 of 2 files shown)


IsaacYangSLA (approved):
Approving cherry-pick PR

PR #4680, opened 2026-05-22
[2.8] Auto-remove Docker job containers

YuanTingHsieh:
## What changed

- Launch Docker SJ/CJ job containers with `auto_remove=True` so Docker removes them after exit even if the supervising SP/CP process dies.
- Add a `DockerJobHandle` exit watcher that calls Docker `wait()` and caches the exit code while the parent process is alive, avoiding false `ABORTED` results when auto-removed containers disappear before polling.
- Reserve `auto_remove` as launcher-controlled and add unit coverage for the auto-remove/`NotFound` race.

## Why

When the parent SP/CP container is killed unexpectedly, the existing `_remove_container()` cleanup path never runs. The job container can finish and remain in `Exited` state indefinitely, consuming Docker metadata and holding its container-name slot until manual cleanup.

### DIFF
--- a/nvflare/app_opt/job_launcher/docker_launcher.py
+++ b/nvflare/app_opt/job_launcher/docker_launcher.py
@@ -15,6 +15,7 @@ import logging
 import os
 import posixpath
 import re
+import threading
 import time
 from abc import abstractmethod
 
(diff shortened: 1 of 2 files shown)


chesterxgchen:
Question: 
this does not make the job container kill itself when the parent SP/CP container dies. It only ensures the job container is removed after the job container exits. If the parent dies and the
  job container keeps running indefinitely, Docker will not auto-remove it until that job process exits.

will the job exists ( instead run forever) when the parent process dies ?

chesterxgchen (approved):

PR #4683, opened 2026-05-22
Support scalar-like analytics values

YuanTingHsieh:
## What changed

- Normalize scalar-like analytics values when constructing `AnalyticsData`.
- Accept Python `int`/`float` values as before, plus scalar-like objects with `.item()` such as `numpy.float32` and 0-D numpy arrays.
- Normalize numeric values inside `SCALARS` and `METRICS` dictionaries, and fail fast if a dict entry is not numeric-scalar-like.
- Fix the invalid `path` validation message to report `type(path)`.
- Add unit coverage for generic scalar-like objects, numpy scalar / 0-D array values, `METRICS` dict normalization, non-scalar dict rejection, and the `path` error message.

## Why

Client tracking writers currently reject values such as `numpy.float32` for single scalar metrics because analytics validation only accepts built-in Python `float`/`int`. This can happen with code like:

```python
running_loss += cost.cpu().detach().numpy() / images.size()[0]
```

Centralizing the normalization in `AnalyticsData` keeps `SummaryWriter`, `MLflowWriter`, `WandBWriter`, and lower-level analytics logging paths consistent without adding a numpy dependency to core API code.

This is a replacement PR for the previous glitched PR #4659, rebased onto the latest `NVIDIA/NVFlare:main` before branch creation.


nvflare/apis/analytix.py
```diff
@@ -179,24 +179,67 @@ def _validate_data_types(
...
         elif data_type in [
             AnalyticsDataType.METRICS,
             AnalyticsDataType.PARAMETERS,
             AnalyticsDataType.SCALARS,
-        ] and not isinstance(value, dict):
-            raise TypeError(f"expect '{key}' value to be an instance of dict, but got '{type(value)}'.")
+        ]:
+            if not isinstance(value, dict):
+                raise TypeError(f"expect '{key}' value to be an instance of dict, but got '{type(value)}'.")
+            if data_type in [AnalyticsDataType.METRICS, AnalyticsDataType.SCALARS]:
```
holgerroth, line 196: This per-entry check is stricter than the previous `isinstance(value, dict)`-only validation, so it rejects values that previously worked end-to-end. Example: `writer.add_scalars('losses', {'train': np.array([0.5])})` — a shape-`(1,)` array (or `torch.tensor([0.5])`) previously passed validation, FOBS-serialized fine (the numpy decomposers are registered by default), and rendered in TensorBoard, which squeezes via `make_np(...).squeeze()`. After this change the same call raises `TypeError` inside the user's training loop.

Suggest accepting size-1 arrays by squeezing them (matching TensorBoard's own behavior), e.g. treating `tuple(shape) in ((), (1,))` as scalar-like in `_normalize_numeric_scalar`, so this PR stays purely additive.

nvflare/apis/analytix.py
```diff
@@ -93,7 +93,7 @@ def __init__(
             sender (LogWriterName): Type of sender for syntax such as Tensorboard or MLflow
             kwargs (optional, dict): additional arguments to be passed.
         """
-        self._validate_data_types(data_type, key, value, **kwargs)
+        value = self._validate_data_types(data_type, key, value, **kwargs)
```
holgerroth, line 99: Note that `AnalyticsData.from_dxo()` also ends in `cls(...)`, so this validation — including the new, stricter per-entry METRICS/SCALARS check — now re-runs on the *receiving* side. In a mixed-version deployment (server on this change, clients on an older release), a SCALARS/METRICS dict containing a previously-accepted entry (e.g. a shape-`(1,)` array) passes the old client's validation but raises here on the server, and `AnalyticsReceiver._save()` (`nvflare/app_common/widgets/streaming.py:166-170`) catches the exception and drops the entire shareable — all metrics in that event — with only a server-side log line.

Consider making receiver-side reconstruction lenient (best-effort normalization without raising) so one bad entry can't silently drop a whole batch.

nvflare/apis/analytix.py
```diff
@@ -179,24 +179,67 @@ def _validate_data_types(
             if step < 0:
```
holgerroth: The numpy use case this PR targets still fails on `global_step`: `isinstance(np.int64(10), int)` is `False`, so `add_scalar('loss', np.float32(0.5), global_step=np.int64(10))` still raises `TypeError: expect step to be an instance of int` even though the value itself is now normalized. And since the guard above is `if step:`, a falsy `np.int64(0)` (the first log of every run) skips validation entirely and the raw numpy object is placed into the DXO by `to_dxo()` — step 0 is accepted unconverted while step 1+ is rejected.

Suggest running `step` through `_normalize_numeric_scalar()` as well and changing the guard to `if step is not None:`.

holgerroth (commented):
Thanks for centralizing the normalization in `AnalyticsData` — that's the right choke point, and the test coverage is solid. A few findings from reviewing the validation changes (details inline): the new per-entry METRICS/SCALARS check is stricter than before and rejects previously-working values; the tightened validation also re-runs receiver-side via `from_dxo`, which can silently drop whole metric batches in mixed-version deployments; and the numpy use case this PR targets still fails when `global_step` is a numpy integer.

nvflare/apis/analytix.py
YuanTingHsieh: Thanks for flagging this. I checked the current pre-change path: `TBAnalyticsReceiver` uses `TensorBoardEventWriter`, whose scalar path calls `float(value)`; it does not use TensorBoard's `make_np(...).squeeze()`. With NumPy 2.x, `float(np.array([0.5]))` raises `TypeError`, so shape-`(1,)` arrays are not reliably supported end-to-end. This PR intentionally defines scalar-like values as native numeric values, NumPy scalars, or 0-D objects whose `.item()` returns a Python number. Callers with one-element arrays can convert them explicitly with `.item()`. I added coverage confirming that non-zero-dimensional shapes remain rejected.

nvflare/apis/analytix.py
YuanTingHsieh: Thanks for calling out the mixed-version path. We considered receiver-only leniency, but it would let values rejected by the current analytics schema bypass validation and could still fail in the downstream writer, so it would not guarantee preserving the metrics event. For the concrete shape-`(1,)` NumPy example, the current TensorBoard writer already fails at `float(value)` under NumPy 2.x. We will keep sender and receiver validation consistent and define the wire contract around true scalar values: native numeric values, NumPy scalars, and 0-D scalar-like objects.

nvflare/apis/analytix.py
YuanTingHsieh: Fixed in `44409a181`. `global_step` is now normalized whenever it is not `None`, so `np.int64(0)` and `np.int64(10)` are stored as Python `int` values in both analytics kwargs and the DXO. NumPy floating-point steps remain rejected. Regression tests cover zero and nonzero NumPy integer steps plus floating-point rejection.

holgerroth (approved):
Thanks for the thorough follow-up. Verified the `global_step` fix in 44409a181 — numpy integer steps (including the falsy `np.int64(0)` case) now normalize to builtin `int` all the way into the DXO, float steps are still rejected, and the regression tests pin it. Also confirmed your point on the shape-`(1,)` case: the receiver's `TensorBoardEventWriter` scalar path calls `float(value)`, which raises `TypeError` under NumPy 2.x, so defining the wire contract around true scalars is the right call. LGTM.

PR #4684, opened 2026-05-22
Improve security by loading classes based on BYOC flag and class lists

IsaacYangSLA:
### Description

component path or nested python path will not be loaded if the prefix of its path is in the block list of component_path_authorizer.  This authorizer can be configured in the resources.json of local folder of the startup kits.

### Types of changes

- [x] Documentation updated.


nvflare/private/json_configer.py
```diff
@@ -103,7 +104,33 @@ def authorize_and_build_component(self, config_dict, config_ctx: ConfigContext,
...
+            node.paths.extend(["args", node.key])
+
+        return node
+
+    def build_nested_component(self, config_dict, arg_name):
+        if self.build_auth_func is None:
+            return super().build_nested_component(config_dict, arg_name)
+
+        node = self._make_nested_component_node(config_dict, arg_name)
+        return self.authorize_and_build_component(config_dict, self.config_ctx, node)
```
YuanTingHsieh: Nested authorization should use the same `config_ctx` as the parent build, not `self.config_ctx`. Direct callers of `authorize_and_build_component(config, ctx, node)` can currently have root auth receive `ctx` while nested auth receives `None` or stale `self.config_ctx`. Please store `(config_ctx, node)` on the build stack and reuse that for nested builds.

nvflare/app_common/widgets/component_path_authorizer.py
```diff
@@ -0,0 +1,102 @@
...
+
+from nvflare.apis.event_type import EventType
+from nvflare.apis.fl_constant import FLContextKey
+from nvflare.apis.fl_context import FLContext
+from nvflare.apis.fl_exception import UnsafeComponentError
+from nvflare.fuel.utils.validation_utils import check_object_type
+from nvflare.widgets.widget import Widget
+
+
+class ComponentPathAuthorizer(Widget):
```
YuanTingHsieh, line 31: changes LGTM, do we need to have this by default? like default provisioning will include this authorizer ?

chesterxgchen (dismissed):
1. Make protection mandatory by default

     ComponentPathAuthorizer should not depend on site-admin installation in resources.json.

     NVFlare should always run a built-in component path authorization check during job config parsing, even when event_handlers is empty.

  2. Add a default blocklist

     Add a built-in default blocklist for dangerous modules/classes, for example:

     DEFAULT_BLOCK_LIST = [
         "subprocess.",
         "os.",
         "posix.",
         "nt.",
         "pty.",
         "multiprocessing.",
         "importlib.",
         "runpy.",
         "builtins.",
         "ctypes.",
         "pickle.",
         "marshal.",
         "shutil.",
         "socket.",
         "urllib.",
         "http.",
         "ftplib.",
         "smtplib.",
     ]

     Sites should get this baseline protection without configuration.

  3. Allow site-admin extension

     Site admins should be able to extend the default blocklist, not replace it.

     Suggested config:

     {
       "id": "component_path_authorizer",
       "path": "nvflare.app_common.widgets.component_path_authorizer.ComponentPathAuthorizer",
       "args": {
         "extra_block_list": [
           "mycompany.unsafe.",
           "thirdparty.risky_module.",
           "some_package.DangerousClass"
         ]
       }
     }

     If both default and configured blocklists exist, enforce the union.

  4. Fix recursive coverage

     The authorization must recurse through nested component configs in both dicts and lists.

     Current PR only handles direct dict-valued args. That leaves list-valued args exposed.
  5. Fix delayed builder bypasses

     Ensure later builders cannot bypass authorization. Specifically review/fix:
      - MultiProcessExecutor._build_components
      - sub_worker_process._initialize
      - WorkerComponentBuilder
      - engine/run-manager build_component() paths

     These should not instantiate configs without the same component path authorization.

  7. Add security regression tests

     Add tests proving these are blocked before instantiation:
      - root-level subprocess.Popen
      - nested dict arg subprocess.Popen
      - deeply nested dict arg
      - list-contained component config
      - MultiProcessExecutor.components payload
      - class_path and name aliases
      - dangerous default-blocklist entries, not only subprocess.
  8. Update docs

     Docs should say:
      - default protection is always enabled
      - site admins can extend with extra_block_list
      - path is required; class_path and name are rejected under this policy
      - this is a denylist baseline, not a replacement for secure job review and sandboxing

  Bottom line: the PR fixes the original direct recursive dict bypass, but it should not merge as the complete security fix until default enforcement, default blocklist, list recursion, and delayed-builder coverage
  are addressed.

docs/user_guide/admin_guide/security/unsafe_component_detection.rst
```diff
@@ -28,14 +30,13 @@ as shown in the following ComponentChecker example:
...
     class ComponentChecker(FLComponent):
-
-    def handle_event(self, event_type: str, fl_ctx: FLContext):
-        prop_keys = fl_ctx.get_prop_keys()
-        if event_type == EventType.BEFORE_BUILD_COMPONENT:
-            print(f"ComponentChecker: fl_ctx props: {prop_keys}")
-            comp_config = fl_ctx.get_prop(FLContextKey.COMPONENT_CONFIG)
-            print(f"Comp Config: {comp_config}")
-            raise UnsafeComponentError("client encountered bad component")
+        def handle_event(self, event_type: str, fl_ctx: FLContext):
```
nvidianz, line 35: Should switch to the new event handling mechanism using register_event_handler().

nvflare/app_common/widgets/component_path_authorizer.py
```diff
@@ -0,0 +1,102 @@
...
+    def __init__(self, block_list: List[str]):
```
nvidianz: deny_list/allow_list are more conventional names to replace black/white.

Blakc-list is too broad for some use-case. It means everything else is allowed. We should provide a white-list also. White has precedence.

pcnudde:
Should the actual whitelist not be in this PR?

docs/programming_guide/component_configuration.rst
```diff
@@ -57,14 +57,14 @@ used. Below is the server side configuration for :ref:`hello_pt_job_api`.
         "workflows": [
             {
                 "id": "pre_train",
-                "name": "InitializeGlobalWeights",
```
nvidianz, line 60: For backward compatibility, we can keep supporting name. Since the name is from component_classes.json, it's already served as a white list.

docs/programming_guide/component_configuration.rst
```diff
@@ -144,19 +144,20 @@ For example:
             "device": "cpu",
             "source_model": "model",
             "optimizer_args": {
-                "class_path": "torch.optim.SGD",
```
nvidianz, line 147: For backward compatibility, we can just treat class_path as path and deprecate it in the documentation.

docs/user_guide/admin_guide/security/unsafe_component_detection.rst
```diff
@@ -11,8 +11,10 @@ NVFLARE has a very powerful and flexible event mechanism that allows custom code
 workflow (e.g. start/end of the job, before/after a task is executed, etc.). At such moments, NVFLARE fires events and invokes
 :ref:`fl_component` objects that handle these events. 
 
-The ``BEFORE_BUILD_COMPONENT`` event type can allow a custom FLComponent to detect unsafe job components during the time of configuration processing. This event
-type is fired before the configuration processor starts to build a job component (executor, filter, etc.). 
+The ``BEFORE_BUILD_COMPONENT`` event type can allow a custom FLComponent to detect unsafe job components during the time of
```
nvidianz, line 14: Should mention path is already checked and how to configure the white-list.

docs/user_guide/admin_guide/security/unsafe_component_detection.rst
```diff
@@ -28,14 +30,13 @@ as shown in the following ComponentChecker example:
...
-            print(f"ComponentChecker: fl_ctx props: {prop_keys}")
-            comp_config = fl_ctx.get_prop(FLContextKey.COMPONENT_CONFIG)
-            print(f"Comp Config: {comp_config}")
-            raise UnsafeComponentError("client encountered bad component")
+        def handle_event(self, event_type: str, fl_ctx: FLContext):
+            prop_keys = fl_ctx.get_prop_keys()
+            if event_type == EventType.BEFORE_BUILD_COMPONENT:
+                print(f"ComponentChecker: fl_ctx props: {prop_keys}")
+                comp_config = fl_ctx.get_prop(FLContextKey.COMPONENT_CONFIG)
+                print(f"Comp Config: {comp_config}")
```
nvidianz: Should have a if statement here

docs/user_guide/admin_guide/security/unsafe_component_detection.rst
```diff
@@ -58,6 +61,60 @@ The following properties in the fl_ctx could be helpful too:
 
 ``FLContextKey.WORKSPACE_OBJECT`` - This object provides many convenience methods to determine the paths of files in the workspace
 
+Use the Built-in Component Path Authorizer
+------------------------------------------
+When a job does not contain BYOC, NVFLARE runs a built-in component path authorization check while parsing job
```
nvidianz: does not contain == BYOC disabled

nvflare/app_common/executors/multi_process_executor.py
```diff
@@ -49,14 +53,85 @@ class WorkerComponentBuilder(ComponentBuilder):
...
+        node = Node(element)
+        node.processor = parent_node.processor
+        node.parent = parent_node
+        node.level = parent_node.level + 1
+        node.key = str(key)
+        node.paths = parent_node.paths.copy()
+        node.paths.append(node.key)
+        return node
+
+    def _is_authorizable_component_config(self, config_dict, node=None):
```
nvidianz, line 99: Don't see the purpose of this wrapper with the same signature.

IsaacYangSLA:
> Should the actual whitelist not be in this PR?
The actual allow_list is specified by site admins.  We provide the mechanism for them to define their own.

docs/programming_guide/component_configuration.rst
IsaacYangSLA: name is too general and that's why we stop using it a long time ago.

nvflare/app_common/executors/multi_process_executor.py
```diff
@@ -49,14 +53,85 @@ class WorkerComponentBuilder(ComponentBuilder):
...
+        return self.is_authorizable_component_config(config_dict, node)
+
+    def _authorize_component_config(self, config_dict, node):
+        if not self.enforce_authorization:
+            return
+        try:
+            self.component_path_authorizer.authorize_component_config(
+                config_dict, node, fl_ctx=self.fl_ctx, workspace=self.workspace
+            )
+        except UnsafeComponentError as ex:
```
nvidianz, line 109: Not sure if the exception should be wrapped. This should only be done if the API has a contract that it only throws certain exceptions.

docs/programming_guide/component_configuration.rst
IsaacYangSLA: class_path and path still exist.  Deprecating class_path should be a decision not in the scope of this PR.

nvflare/app_common/executors/multi_process_executor.py
```diff
@@ -49,14 +53,85 @@ class WorkerComponentBuilder(ComponentBuilder):
...
+            raise ComponentNotAuthorized(f"component not authorized: {ex}")
+
+    def _authorize_component_config_tree(self, element, node, force_current=False):
```
nvidianz, line 112: This should include all containers in Python, like all iterables. For example, if a set is used in the args, it would bypass the check. For a faster recursion, we can also only check for dict with "path"/"class_path" key.

nvflare/app_common/executors/multi_process_executor.py
IsaacYangSLA: Other exceptions are not related to component_path_authorizer and should be handle by upper layers.

docs/user_guide/admin_guide/security/unsafe_component_detection.rst
IsaacYangSLA: Thanks.  Will update.

docs/user_guide/admin_guide/security/unsafe_component_detection.rst
IsaacYangSLA: Thanks.  May address it in the future release.

docs/user_guide/admin_guide/security/unsafe_component_detection.rst
IsaacYangSLA: Will fix

nvflare/app_common/executors/multi_process_executor.py
IsaacYangSLA: Maybe because it's from json, so we handle list/dict only.

nvflare/app_common/widgets/component_path_authorizer.py
```diff
@@ -0,0 +1,199 @@
...
+        if not isinstance(component_config, dict):
+            raise UnsafeComponentError(f"Component config must be dict but got {type(component_config)}")
+
+        if "path" in component_config:
+            component_path = component_config["path"]
+            key = "path"
+        elif "class_path" in component_config:
+            component_path = component_config["class_path"]
+            key = "class_path"
+        elif "name" in component_config:
```
YuanTingHsieh: This makes "name" configs bypass the built-in allowlist. ComponentBuilder still treats "name" as a valid class config and resolves it later through the module scanner, so a non-BYOC job can avoid path/class_path validation by using a resolvable class name. If protected job configs should reject "name" as the docs now say, this branch should raise UnsafeComponentError instead of returning None. Alternatively, resolve the name to its full class path and check that path against class_allow_list before build.

IsaacYangSLA: This is from the conclusion of a meeting with @nvidianz .  Basically, the value of "name" will be mapped to an actual class path and the mapping is pre-defined in nvflare/fuel/utils/component_classes.json
Since all of those classes are under nvflare, there is no need to check them.

IsaacYangSLA: Also updated the document for more clarity in a follow-up commit.

nvidianz (dismissed):

nvflare/app_common/widgets/component_path_authorizer.py
YuanTingHsieh: Agreed that "name" resolves only through the precomputed NVFLARE component class table, not arbitrary Python imports. But because the allowlist is intended to be narrower than "all resolvable NVFLARE components", returning None for "name" still bypasses the class_allow_list. Some resolvable names correspond to classes intentionally excluded from the default allowlist, such as launcher/process execution surfaces. Protected non-BYOC configs should either reject "name" or resolve it first and check the resolved class path.

YuanTingHsieh: for example: component_classes.json includes entries like ClientAPILauncherExecutor, InProcessClientAPIExecutor, SubprocessLauncher, PTClientAPILauncherExecutor

IsaacYangSLA: We switched back to rejecting name and accepting class_path and path only.

nvidianz (approved):

PR #4692, opened 2026-05-27
Don't use kagglehub==1.0.1

IsaacYangSLA:
### Description

See this issue (https://github.com/Kaggle/kagglehub/issues/301)

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -101,7 +101,7 @@ test_support =
     pandas>=1.5.1
     nbformat
     nbmake
-    kagglehub
+    kagglehub!=1.0.1
 
 test =
     %(all)s


setup.cfg
```diff
@@ -101,7 +101,7 @@ test_support =
     pandas>=1.5.1
     nbformat
     nbmake
-    kagglehub
+    kagglehub!=1.0.1
```
YuanTingHsieh, line 104: Why we using exclude ?
Does the bug only exist in this specific version?

IsaacYangSLA: This is the latest kagglehub and it's not compatible with the latest kagglesdk 0.1.24.

YuanTingHsieh (approved):

PR #4699, opened 2026-05-27
[2.8] cert: make project_profile.yaml server.admin_port optional

chesterxgchen:
## Summary

`nvflare cert approve` previously rejected any `project_profile.yaml` that omitted `server.admin_port`, with:

```
project profile server.admin_port must be an integer from 1 to 65535
```

The docs and the server-participant validation path both treat `admin_port` as **optional** (it falls back to `fed_learn_port`). The unconditional rejection inside `_load_project_profile` was the outlier and the cause of the user-visible bug reported during 2.8 manual validation.

This PR makes the implementation match the documented (and intended) behavior:

- In `_load_project_profile`, default `admin_port` to the resolved `fed_learn_port` when the profile omits it.
- Keep the range check (1..65535) when an explicit value is supplied.
- Drop "and admin_port" from the "server must be a mapping with ..." error message since `admin_port` is no longer a required key in that block.

## Test coverage

Two new unit tests in `tests/unit_test/tool/cert/cert_commands_test.py`:

- `test_load_project_profile_defaults_admin_port_to_fed_learn_port_when_omitted` — confirms omitted `admin_port` yields a profile where `admin_port == fed_learn_port`.
- `test_load_project_profile_rejects_invalid_admin_port_when_present` — confirms an explicit out-of-range `admin_port` (e.g. -1) still fails validation.

Full cert test suite (`pytest tests/unit_test/tool/cert/`): 249 passed.

## Changes

| File | + / − |
|---|---|
| `nvflare/tool/cert/cert_commands.py` | +6 / −3 |
| `tests/unit_test/tool/cert/cert_commands_test.py` | +32 / 0 |

## Why this PR

The original 2.8 manual validation report flagged `admin_port` as required in the implementation but optional in the docs. Two prior options were considered:

1. Update the docs to say `admin_port` is required.
2. Update the implementation to actually make it optional, matching the existing docs.

Option 2 is the right call — the docs accurately describe the intended behavior (`admin_port` is the listen port the server binds for admin traffic, defaulting to `fed_learn_port` for single-port deployments), and the server-participant validation path already implements that fallback. Aligning `_load_project_profile` with the same convention is internally consistent and removes a user-facing setup foot-gun.

After this PR, the docs already in #4694 stay correct without further changes.

### DIFF
--- a/nvflare/tool/cert/cert_commands.py
+++ b/nvflare/tool/cert/cert_commands.py
@@ -2380,7 +2380,7 @@ def _load_project_profile(profile_path: str, request_project: str = None) -> dic
             "Invalid arguments.",
             _USAGE_HINT,
             exit_code=4,
-            detail="project profile server must be a mapping with host, fed_learn_port, and admin_port",
+            detail="project profile server must be a mapping with host and fed_learn_port",
         )
         return None
     server_host = server.get("host")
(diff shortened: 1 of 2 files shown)


pcnudde (approved):

IsaacYangSLA:
Any document update regarding this implementation change?

pcnudde (approved):

PR #4702, opened 2026-05-27
Simplify Docker runtime image

pcnudde:
## What changed

- Replaces the old root `docker/Dockerfile` with two purpose-specific Dockerfiles:
  - `docker/Dockerfile.parent` for long-running NVFlare server/client parent processes.
  - `docker/Dockerfile.job` as a starter image for user training/evaluation jobs.
- Builds the parent image with a pinned Python 3.14 slim build stage and a pinned NVIDIA Distroless Python 3.14 runtime stage.
- Builds the job image from `nvcr.io/nvidia/pytorch:26.04-py3` so GPU PyTorch workloads use NVIDIA's tested framework stack.
- Keeps the parent image minimal by installing only the NVFlare runtime needed for parent processes.
- Installs plain NVFlare in the job image, without the `K8S` extra and without the `PT` extra:
  - the Kubernetes SDK is needed by the K8s launcher in the parent process, not by Docker-launched job worker processes.
  - the PyTorch packages supplied by the NGC PyTorch image are left intact.
- Adds `docker/README.md` explaining the parent/job image split, why the parent image has no shell, and how to build/use each image.
- Updates existing docs and devops references from `docker/Dockerfile` to `docker/Dockerfile.parent`.

## Why

The previous single-stage image inherited the full `python` base image, including many OS/build packages that were not needed at runtime and showed up in security scans. This split keeps parent server/client containers small and hardened while allowing user jobs to run in a richer ML framework image.

The parent image is intentionally minimal and distroless. It does not include a shell, package manager, or general-purpose debugging utilities. Debugging should use container logs and NVFlare logs.

The job image is intentionally not distroless. It is a starter image for user code and uses the NGC PyTorch container so CUDA/PyTorch workloads run on the expected NVIDIA framework stack.

## Notes for reviewers

- `DockerJobLauncher` defaults to `/usr/local/bin/python` for job containers. The NGC PyTorch image provides Python at `/usr/bin/python`, so `docker/Dockerfile.job` adds a `/usr/local/bin/python` symlink to work with the default launcher config.
- The NGC PyTorch base contains a Debian-managed `PyYAML`, so the job Dockerfile installs PyYAML with `--ignore-installed` before installing NVFlare. This avoids pip failing while trying to uninstall a package without pip `RECORD` metadata.
- Kubernetes imports in this code path are isolated to `nvflare/app_opt/job_launcher/k8s_launcher.py`. Docker-launched job worker processes enter through `worker_process.py` / `runner_process.py` and use CellNet workspace transfer instead.

## Validation

- Built `docker/Dockerfile.parent` and verified:
  - `nvflare --version` runs.
  - `import nvflare; import kubernetes` succeeds.
  - `/usr/bin/zip --version` succeeds.
  - `/bin/sh` is absent as expected for the distroless parent image.
- Built `docker/Dockerfile.job` from `nvcr.io/nvidia/pytorch:26.04-py3` and verified:
  - `import nvflare; import torch` succeeds.
  - `/usr/local/bin/python` works in the image.
  - `importlib.util.find_spec("kubernetes") is None`.
- Ran an end-to-end Docker launcher test on Omnistation with an NVIDIA L4 GPU:
  - parent server/client containers ran from the parent image.
  - job containers launched from the PyTorch job image.
  - synthetic GPU PyTorch job completed with `FINISHED:COMPLETED`.
  - client log showed `cuda_available=True`, `device=NVIDIA L4`, and `tensor_device=cuda:0`.


docker/Dockerfile.parent
```diff
@@ -1,21 +1,42 @@
-FROM python:3.14.5-slim-trixie@sha256:7a500125bc50693f2214e842a621440a1b1b9cbb2188f74ab045d29ed2ea5856
+ARG PYTHON_BUILD_BASE=python:3.14.5-slim-trixie@sha256:7a500125bc50693f2214e842a621440a1b1b9cbb2188f74ab045d29ed2ea5856
+ARG PYTHON_RUNTIME_BASE=nvcr.io/nvidia/distroless/python:3.14-v4.0.7@sha256:49751a52c1b4f59e0b68d6caf6728f305afc9e47c507008f8a9e8e1253929676
+ARG PYTHON_VERSION=3.14
```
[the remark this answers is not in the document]
pcnudde: Addressed in 49df638 by adding a `PYTHON_VERSION` sync note plus a runtime-stage import smoke check: `python -c "import nvflare; import kubernetes"`. If the base images are later bumped to a different Python minor without updating `PYTHON_VERSION`, the `.pth` path will be wrong and this build step will fail instead of producing an unimportable runtime image.

docker/Dockerfile.parent
```diff
@@ -1,21 +1,42 @@
...
+
+ENV PATH=/opt/nvflare/bin:/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin \
+    PYTHONDONTWRITEBYTECODE=1 \
+    PYTHONUNBUFFERED=1
+
+WORKDIR /app
+
+COPY --from=builder /opt/nvflare /opt/nvflare
+COPY --from=builder /usr/local/lib/python${PYTHON_VERSION}/site-packages/nvflare-prefix.pth /usr/local/lib/python${PYTHON_VERSION}/site-packages/nvflare-prefix.pth
+COPY --from=builder /usr/bin/zip /usr/bin/zip
```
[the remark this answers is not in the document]
pcnudde: Addressed in 49df638 by adding a runtime-stage check immediately after copying `zip`: `/usr/bin/zip --version`. That command runs inside the final distroless image during the Docker build, so any missing dynamic linker or shared-library dependency for the copied binary will fail the build instead of surfacing later at job runtime.

pcnudde:
@greptileai review again

docker/Dockerfile.job
```diff
@@ -0,0 +1,27 @@
...
+
+WORKDIR /opt/NVFlare
+COPY . .
+
+# DockerJobLauncher defaults to /usr/local/bin/python. The NGC PyTorch image
+# provides Python at /usr/bin/python, so add the default launcher path.
+RUN mkdir -p /usr/local/bin \
+    && ln -sf /usr/bin/python /usr/local/bin/python \
+    && python -m pip install --ignore-installed "PyYAML>=6.0" \
+    && NVFL_RELEASE=1 python -m pip install ".[K8S]" \
```
YuanTingHsieh: does the job image require K8S?

chesterxgchen: It should not

pcnudde: removed it.

YuanTingHsieh (approved):

PR #4707, opened 2026-05-28
[2.8] Fix FOBS externalizer shared-state race

nvidianz:
## Summary
- Make FOBS externalization build fresh dict/list containers instead of mutating caller-owned containers.
- Remove obsolete restore-position bookkeeping from the recursive externalization path.
- Add regression coverage for shallow `Shareable` copies that share nested payload containers.

## Root Cause
`Externalizer.externalize()` temporarily replaced large bytes/str leaves with `DatumRef` values inside source dict/list containers. Since `Shareable.make_copy()` is shallow, concurrent serializers can share nested containers. One serializer could observe another serializer's temporary `DatumRef` and emit a main body without the matching LOBS datum section, which later fails with `can't find datum for <uuid>` during decode.

## Validation
- `PYTHONPYCACHEPREFIX=/private/tmp/nvflare-pycache /Users/zhihongz/nvflare-env/3.12/bin/python -m pytest -q tests/unit_test/fuel/utils/fobs/decomposer_test.py` (`6 passed`)
- `PYTHONPYCACHEPREFIX=/private/tmp/nvflare-pycache /Users/zhihongz/nvflare-env/3.12/bin/python -m pytest -q tests/unit_test/fuel/utils/fobs` (`57 passed`)
- `PYTHONPATH=/Users/zhihongz/work/NVFlare PYTHONPYCACHEPREFIX=/private/tmp/nvflare-pycache /Users/zhihongz/nvflare-env/3.12/bin/python /private/tmp/repro_fobs_missing_datum.py --rounds 10 --tasks 128 --workers 32` (`completed without reproducing`)
- `git diff --check HEAD~1..HEAD`
- `env PATH=/Users/zhihongz/nvflare-env/3.12/bin:/opt/homebrew/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin PYTHONPYCACHEPREFIX=/private/tmp/nvflare-pycache ./runtest.sh -s nvflare/fuel/utils/fobs/decomposer.py tests/unit_test/fuel/utils/fobs/decomposer_test.py`

### DIFF
--- a/nvflare/fuel/utils/fobs/decomposer.py
+++ b/nvflare/fuel/utils/fobs/decomposer.py
@@ -17,7 +17,7 @@ from enum import Enum
 from typing import Any, List, Optional, Type, TypeVar
 
 # Generic type supported by the decomposer.
-from nvflare.fuel.utils.fobs.datum import Datum, DatumManager, DatumRef, DatumType
+from nvflare.fuel.utils.fobs.datum import Datum, DatumManager
 
 T = TypeVar("T")
 
@@ -92,31 +92,6 @@ class Decomposer(ABC):
         pass
 
 
-def restore_position(manager: DatumManager, datum: Datum, position):
-    """
-    This function is used for restoring object state at the specified position.
-
-    Args:
-        manager: the datum manager
-        datum: the datum that contains the value of the original object at the position.
-        position: the position to be restored
-
-    Returns: None
-
-    """
-    target, key = position
-    original_obj = manager.get_original(target)  # also need to restore values in the original object if any
-    if datum.datum_type in (DatumType.BLOB, DatumType.TEXT):
-        target[key] = datum.value
-        if original_obj:
-            original_obj[key] = datum.value
-    else:
-        # file datum - app provided
-        target[key] = datum
-        if original_obj:
-            original_obj[key] = datum
-
-
 class Externalizer:
     """
     This class is used to help creating 'decompose' method of decomposers of arbitrary classes.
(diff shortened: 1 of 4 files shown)


chesterxgchen (commented):
Claude found a few small issues

  The four Lows (all non-blocking polish)

  1. L1 — register_copy/set_restore_func/get_original/obj_copies are now dead code in datum.py. Cleanup follow-up.
  2. L2 — type(target)() now requires no-arg constructor at every nested dict/list level (was only top-level before).
  Edge case for defaultdict(list) etc. Doc note suggested.
  3. L3 — tc = target.copy() returning a plain dict is intentional but the comment could clarify the nested case.
  4. L4 — The diff removed the inline # note: tuple is not supported since it is immutable comment. Restore it.

nvflare/fuel/utils/fobs/decomposer.py
```diff
@@ -126,32 +101,27 @@ class Externalizer:
...
+                new_target[k] = self.externalize(v)
+            return new_target
         elif isinstance(target, list):  # note: tuple is not supported since it is immutable.
-            for i, v in enumerate(target):
-                d = self.externalize(v)
-                target[i] = d
-                self._set_position(d, target, i)
+            new_target = type(target)()
+            for v in target:
+                new_target.append(self.externalize(v))
```
IsaacYangSLA, line 124: append is quite slow.  If target is a list, can we do new_target=[self.exteranl(v) for v in target]?

nvidianz: The traget may be a derived class of list. The generator produce a list and lose the original type.

chesterxgchen (approved):
LGTM

PR #4720, opened 2026-05-29
[2.8] Fix dashboard container launch from installed package

chesterxgchen:
## Summary
- launch the dashboard container with python -m nvflare.dashboard.wsgi so installed-package images do not require the source tree under /app
- make the dashboard WSGI module use a package-qualified import
- report immediate dashboard container exits with recent logs instead of printing a successful start message
- fix the dashboard volume option key from model to mode

## Tests
- python -m pytest tests/unit_test/dashboard/cli_test.py
- python -m pytest tests/unit_test/dashboard
- git diff --check
- ./runtest.sh -s

### DIFF
--- a/nvflare/dashboard/cli.py
+++ b/nvflare/dashboard/cli.py
@@ -114,7 +114,7 @@ def start(args):
     try:
         container_obj = client.containers.run(
             dashboard_image,
-            entrypoint=["python", "-m", "nvflare.dashboard.wsgi"],
+            entrypoint=["python3", "-m", "nvflare.dashboard.wsgi"],
             detach=True,
             auto_remove=True,
             name="nvflare-dashboard",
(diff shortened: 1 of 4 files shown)


pcnudde (approved):

nvflare/dashboard/cli.py
```diff
@@ -113,12 +114,12 @@ def start(args):
     try:
         container_obj = client.containers.run(
             dashboard_image,
-            entrypoint=["/usr/local/bin/python3", "nvflare/dashboard/wsgi.py"],
+            entrypoint=["python3", "-m", "nvflare.dashboard.wsgi"],
```
pcnudde, line 117: One thing to note. Different base images expose python on different paths. 
We can do what we do in docker/k8s and make the python path configurable? Or make a it mandatory that it is in a particular spot in the image. (via and 'ln -s' in the dockerfile)

pcnudde (approved):

pcnudde (approved):

PR #4724, opened 2026-05-29
[2.8] Fix external process max_resends serialization

chesterxgchen:
## Summary
- route ClientAPILauncherExecutor max_resends through LauncherExecutor/TaskExchanger as the single runtime field
- explicitly serialize default max_resends=3 for recipe-generated external-process jobs
- apply recipe.add_client_config timeout/max_resends overrides before writing subprocess Client API config
- update docs and add regression coverage for recipe external-process startup

## Testing
- python -m pytest tests/unit_test/app_common/executors/client_api_launcher_executor_test.py tests/unit_test/app_opt/pt/test_pt_launcher_executor_params.py tests/unit_test/job_config/fed_job_config_test.py tests/unit_test/recipe/fedavg_recipe_test.py -q
- python -m pytest tests/unit_test/client/test_download_complete_gating.py -q
- python job.py --n_clients 2 --num_rounds 1 --batch_size 2 --epochs 1 --num_workers 0 --synthetic_data --train_size 4 --test_size 4 --launch_external_process (examples/hello-world/hello-pt)
- ./runtest.sh -s

### DIFF
--- a/nvflare/app_common/executors/client_api_launcher_executor.py
+++ b/nvflare/app_common/executors/client_api_launcher_executor.py
@@ -29,6 +29,8 @@ from nvflare.utils.configs import get_client_config_value
 
 logger = logging.getLogger(__name__)
 
+_CONFIG_VALUE_MISSING = object()
+
 
 class ClientAPILauncherExecutor(LauncherExecutor):
     def __init__(
@@ -58,7 +60,7 @@ class ClientAPILauncherExecutor(LauncherExecutor):
         memory_gc_rounds: int = 0,
         cuda_empty_cache: bool = False,
         submit_result_timeout: float = 300.0,
-        max_resends: int = 3,
+        max_resends: Optional[int] = 3,
         download_complete_timeout: float = 1800.0,
     ) -> None:
         """Initializes the ClientAPILauncherExecutor.
@@ -94,16 +96,17 @@ class ClientAPILauncherExecutor(LauncherExecutor):
                 pipe message.  With reverse PASS_THROUGH enabled CJ ACKs immediately (LazyDownloadRef creation is
                 microseconds), so 300 s is a very generous allowance.  Without reverse PASS_THROUGH, CJ must
                 download the full result before ACKing; in that case this should be at least as large as the
-                expected transfer time.  Configurable via recipe.add_client_config({"submit_result_timeout": N}).
+                expected transfer time. Recipe-based jobs can override via
+                recipe.add_client_config({"submit_result_timeout": N}).
             max_resends (int): Maximum number of retries after the initial result send if CJ does not ACK within
                 submit_result_timeout. Defaults to 3. Set to a finite non-negative integer; 0 disables retries.
                 None means unlimited retries (unsafe for large models because each retry creates a new download
-                transaction) and is rejected at job initialization. Configurable via
-                recipe.add_client_config({"max_resends": N}).
+                transaction) and is rejected at job initialization. Recipe-based jobs serialize this default in
+                executor args; override per job via recipe.add_client_config({"max_resends": N}).
             download_complete_timeout (float): How long (seconds) the subprocess waits after send_to_peer() ACKs
                 for the server to finish downloading its tensors from the subprocess DownloadService.  Without
                 this gate, the subprocess may exit before the download completes and the server gets
-                "no ref found".  Defaults to 1800 s.  Configurable via
+                missing download refs. Defaults to 1800 s. Recipe-based jobs can override via
                 recipe.add_client_config({"download_complete_timeout": N}).
         """
         LauncherExecutor.__init__(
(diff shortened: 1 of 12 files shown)


nvflare/app_common/executors/client_api_launcher_executor.py
```diff
@@ -194,39 +199,68 @@ def initialize(self, fl_ctx: FLContext) -> None:
...
+            fl_ctx, EXTERNAL_PRE_INIT_TIMEOUT, "_external_pre_init_timeout"
+        )
+        self._apply_positive_float_client_config_override(fl_ctx, PEER_READ_TIMEOUT, "peer_read_timeout")
+        self._apply_positive_float_client_config_override(
+            fl_ctx, ConfigKey.SUBMIT_RESULT_TIMEOUT, "_submit_result_timeout"
+        )
+        self._apply_max_resends_client_config_override(fl_ctx)
+        self._apply_positive_float_client_config_override(
+            fl_ctx, ConfigKey.DOWNLOAD_COMPLETE_TIMEOUT, "_download_complete_timeout"
+        )
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in 3285cc930 by syncing _stop_task_wait_timeout after applying the download_complete_timeout override. Added an assertion to the existing override regression test so this path is covered.

chesterxgchen:
Handled the Greptile finding about `_stop_task_wait_timeout` staying stale when `download_complete_timeout` is overridden via `recipe.add_client_config()`. Fixed in 3285cc930, replied on the review thread, and marked the thread resolved.

Verification:
- `python -m pytest tests/unit_test/app_common/executors/client_api_launcher_executor_test.py::test_client_config_overrides_apply_before_subprocess_config_write tests/unit_test/recipe/fedavg_recipe_test.py::TestFedAvgRecipeExternalProcessStartup -q`
- `./runtest.sh -s`
- `git diff --check`

chesterxgchen:
Addressed the remaining review lows in b60884988:

- L1: kept explicit `null` timeout overrides as fail-loud behavior and added coverage for `download_complete_timeout: null` to ensure no subprocess config is written.
- L2: changed top-level `max_resends: null` override handling to reject directly at override parsing with the finite-integer message instead of accepting it and failing later in required startup validation.
- L4: clarified in the recipe regression helper that only advisory timeout relationship validation is patched out; required startup validation still runs through `prepare_config_for_launch()`.
- L5: added a regression test that rewrites exported recipe config to simulate old rc4 executor args with `max_resends: null`, reloads it, and verifies startup rejects before writing `client_api_config.json`.

Verification:
- `python -m pytest tests/unit_test/app_common/executors/client_api_launcher_executor_test.py::test_client_config_max_resends_override_rejects_invalid_values tests/unit_test/app_common/executors/client_api_launcher_executor_test.py::test_client_config_float_override_rejects_none tests/unit_test/recipe/fedavg_recipe_test.py::TestFedAvgRecipeExternalProcessStartup -q`
- `python -m pytest tests/unit_test/app_common/executors/client_api_launcher_executor_test.py tests/unit_test/app_opt/pt/test_pt_launcher_executor_params.py tests/unit_test/job_config/fed_job_config_test.py tests/unit_test/recipe/fedavg_recipe_test.py -q`
- `./runtest.sh -s`
- `git diff --check`

nvflare/app_common/executors/client_api_launcher_executor.py
```diff
@@ -194,39 +200,74 @@ def initialize(self, fl_ctx: FLContext) -> None:
...
+        if value is _CONFIG_VALUE_MISSING:
+            return
+
+        if value is None:
+            msg = f"{ConfigKey.MAX_RESENDS} must be a finite non-negative integer, got None"
+            self.log_error(fl_ctx, msg)
+            raise ValueError(msg)
+
+        try:
+            max_resends = int(value)
```
YuanTingHsieh: The new override path uses int(value), so recipe.add_client_config({'max_resends': 2.9}) silently becomes 2, and string values like '3' are accepted. That bypasses the stricter TaskExchanger constructor validation even though the error/docs say this must be a finite non-negative integer. Please reject non-int values instead of truncating them, ideally using the existing validation helper.

chesterxgchen: Fixed in 7f7ff70a9. The override now validates with the same `check_non_negative_int` helper the `TaskExchanger` constructor uses (plus an explicit `bool` guard) instead of coercing with `int()`, so floats (`2.9`), integer-valued floats (`3.0`), numeric strings (`"3"`), bools, negatives, and `None` all fail loudly with the `must be a finite non-negative integer` error. Extended `test_client_config_max_resends_override_rejects_invalid_values` to parametrize over `[-1, None, 2.9, 3.0, "3", True]`.

Verification:
- `python -m pytest tests/unit_test/app_common/executors/client_api_launcher_executor_test.py -q`
- `./runtest.sh -s`

YuanTingHsieh (approved):

PR #4725, opened 2026-05-29
[2.8] Prevent corrupt client API models from incomplete downloads

chesterxgchen:
## Summary
- fail FOBS recomposition when streamed download items are missing or `None`, preventing incomplete downloads from materializing corrupt model params
- forward `receive(timeout=...)` through the external-process and in-process client API paths so timeout calls return `None` instead of raising or blocking incorrectly
- keep `send()` guarded after timed-out receives, including the in-process multi-round case where `clear()` must reset stale receive state
- warn when `heartbeat_timeout` is shorter than the configured streaming request timeout

## Tests
- `python -m pytest tests/unit_test/client/ex_process/api_test.py tests/unit_test/client/in_process/api_test.py tests/unit_test/client/test_model_registry.py tests/unit_test/fuel/utils/fobs/decomposers/via_downloader_test.py tests/unit_test/app_common/executors/client_api_launcher_executor_test.py`
- `./runtest.sh -s`

### DIFF
--- a/nvflare/app_common/executors/client_api_launcher_executor.py
+++ b/nvflare/app_common/executors/client_api_launcher_executor.py
@@ -346,6 +346,23 @@ class ClientAPILauncherExecutor(LauncherExecutor):
                 f"Set peer_read_timeout >= {per_req}s in job config.",
             )
 
+        if configured_per_req is not None and self.heartbeat_timeout is None:
+            self.heartbeat_timeout = per_req
+            self.log_warning(
+                fl_ctx,
+                "Timeout inconsistency: heartbeat_timeout is not set after applying job-config overrides. "
+                "The CJ/subprocess pipe may miss liveness detection while large payloads are downloading. "
+                f"Using {per_req}s for this run. Set heartbeat_timeout >= {per_req}s in job config.",
+            )
+        elif configured_per_req is not None and self.heartbeat_timeout < per_req:
+            self.log_warning(
+                fl_ctx,
+                f"Timeout inconsistency: heartbeat_timeout ({self.heartbeat_timeout}s) < "
+                f"{prefix}streaming_per_request_timeout ({per_req}s). "
+                "The CJ/subprocess pipe may be marked unhealthy while a large payload is still downloading. "
+                f"Set heartbeat_timeout >= {per_req}s in job config.",
+            )
+
         if configured_per_req is not None and self._download_complete_timeout < per_req:
             self.log_warning(
                 fl_ctx,
--- a/nvflare/client/in_process/api.py
+++ b/nvflare/client/in_process/api.py
@@ -114,25 +114,32 @@ class InProcessClientAPI(APISpec):
             self.logger.info(f"Memory management enabled: cleanup every {gc_rounds} round(s)")
 
     def receive(self, timeout: Optional[float] = None) -> Optional[FLModel]:
-        result = self.__receive()
-        self.receive_called = True
+        result = self.__receive(timeout)
         if result is not None:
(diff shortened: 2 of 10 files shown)


nvflare/client/in_process/api.py
```diff
@@ -114,25 +114,32 @@ def configure_memory_management(self, gc_rounds: int = 0, cuda_empty_cache: bool
             self.logger.info(f"Memory management enabled: cleanup every {gc_rounds} round(s)")
 
     def receive(self, timeout: Optional[float] = None) -> Optional[FLModel]:
-        result = self.__receive()
-        self.receive_called = True
+        result = self.__receive(timeout)
         if result is not None:
+            self.receive_called = True
```
YuanTingHsieh, line 119: Could we also reset `self.receive_called = False` in `InProcessClientAPI.clear()`?

With this change, a timed-out `receive()` no longer sets `receive_called`, which is good. But after a previous successful round, `send(clear_cache=True)` calls `clear()`, and `clear()` currently only sets `self.fl_model = None`. That leaves `receive_called=True` into the next round. If the next `receive(timeout=...)` returns `None`, `send()` could still pass the receive guard using stale state from the prior round.

`ExProcessClientAPI.clear()` already resets `receive_called`, so adding the same reset here would keep the two APIs consistent.

YuanTingHsieh (commented):
Can we update the PR description to match the current code?

The summary says this PR “rejects received FLModel params containing None,” but the patch does not generally reject all `None` values in `FLModel.params` — in fact `test_allows_received_model_with_none_param_values` confirms those are still allowed. The actual behavior is that FOBS recomposition now rejects missing/`None` downloaded items, so incomplete streamed downloads fail before materializing corrupt tensors.

Maybe change that bullet to something like:
“fail FOBS recomposition when streamed download items are missing or None, preventing incomplete downloads from materializing corrupt model params.”

chesterxgchen:
Addressed the review feedback in the latest push:

- Added `self.receive_called = False` in `InProcessClientAPI.clear()` and covered the multi-round stale-guard case with `test_clear_resets_receive_guard_between_rounds`.
- Updated the PR description to say FOBS recomposition fails when streamed download items are missing or `None`, instead of saying all `FLModel.params` `None` values are rejected.
- Re-ran focused tests (`135 passed`) and `./runtest.sh -s`.

Both line review threads are marked resolved. The remaining visible description feedback is a top-level review comment, so there is no thread to resolve for it.

nvflare/app_common/executors/client_api_launcher_executor.py
```diff
@@ -346,6 +346,22 @@ def _validate_timeout_config(self, fl_ctx: FLContext):
                 f"Set peer_read_timeout >= {per_req}s in job config.",
             )
 
+        if configured_per_req is not None and self.heartbeat_timeout is None:
```
nvidianz, line 349: Should we correct the parameter and issue a warning?

nvidianz (dismissed):
Approved with a question.

YuanTingHsieh (dismissed):

nvflare/app_common/executors/client_api_launcher_executor.py
chesterxgchen: Updated in 30e864b. For the unset `heartbeat_timeout` case, the executor now sets it to the configured streaming per-request timeout and logs a warning that this value is being used. Explicit numeric values below the streaming timeout still warn only, so the 2.8 patch does not silently override an operator-provided value.

YuanTingHsieh (approved):

PR #4736, opened 2026-05-30
Add progress-aware streaming wait policy

chesterxgchen:
## Goal
Make streamed task/result transfer resilient to slow networks, large payloads, and many concurrent receivers without requiring users to tune several timeout settings in lockstep.

The default behavior should be:
> If a streamed transfer is making monotonic progress, do not fail or resend solely because wall-clock time is long. If progress stops for the idle timeout, fail clearly.

## Current Problem
The parent client job (CJ) could hit `peer_read_timeout` while the subprocess is still downloading/materializing the task payload. The CJ then resends the task even though the previous task payload is still active. Under congestion this creates duplicate work, stale state, and eventually corrupt or premature receive behavior.

Existing streaming/download code already has partial progress or idle-timeout concepts:

- `DownloadService` transaction timeout is based on time since last download activity.
- `ViaDownloader` has `min_download_timeout`, described as an inactivity timeout.
- `ByteStreamer` has ACK progress timeout.
- `ByteReceiver` has read timeout while waiting for chunks.

The gap is cross-process visibility. The CJ does not know that the subprocess is actively materializing the task payload.

## Summary
- add direction-neutral transfer progress tracking and stream progress pipe events for task payload downloads
- suppress task resend while streamed task payload materialization makes monotonic progress
- wire generic streaming_idle_timeout / streaming_max_peer_silence config defaults and progress callbacks through NumPy/Tensor download paths
- document the Phase 1 internal _STREAM_PROGRESS_ topic and add focused regression coverage

## Progress Awareness Workflow 

 Progress awareness is basically:

  - progress callback = how the transfer source reports “bytes/items are still moving”
  - TransferProgressRecord = the in-memory state used later to decide “keep waiting or declare stalled”
  - wait callback = how the waiting side asks “should I keep waiting after the old timeout tick?”

  The shared tracker is in nvflare/fuel/f3/streaming/transfer_progress.py:21. A record stores job_id, task_id, transfer_id, direction, optional receiver_id, sequence, bytes_done, items_done, started_time, last_progress_time, and terminal state: nvflare/
  fuel/f3/streaming/transfer_progress.py:53. Updates are monotonic: stale sequence or regressed bytes are rejected: nvflare/fuel/f3/streaming/transfer_progress.py:139.

###  Forward Pass: Task Payload Download

* Flow: CJ / TaskExchanger -> subprocess agent -> server download

  1. CJ sends a task to the subprocess through TaskExchanger.
     The send path attaches a progress_wait_cb through CellPipe into Cell.send_request: nvflare/fuel/utils/pipe/cell_pipe.py:421.
  2. Subprocess deserializes the task payload. If tensors are lazy/downloadable, ViaDownloader creates a progress callback for download_object.
     It stamps:
      - direction = task_payload_download
      - job_id
      - task_id
      - transfer_id = ref_id
        at nvflare/fuel/utils/fobs/decomposers/via_downloader.py:763.
  3. download_object() pulls chunks from the server. As chunks arrive, it emits start, active, completed, failed, or aborted progress through the callback: nvflare/fuel/f3/streaming/download_service.py:831, nvflare/fuel/f3/streaming/download_service.py:968.
  4. The subprocess sends _STREAM_PROGRESS_ back through the pipe. The CJ receives it in TaskExchanger._handle_stream_progress_message(), normalizes the payload, and updates its TransferProgressTracker: nvflare/app_common/executors/task_exchanger.py:283.
  5. If the old peer_read_timeout fires before the subprocess ACKs the task, Cell calls progress_wait_cb() instead of immediately timing out: nvflare/fuel/f3/cellnet/cell.py:428.
  6. TaskExchanger checks tracker records:
      - no record yet but still within streaming_idle_timeout -> keep waiting
      - all active transfers have recent progress -> keep waiting
      - any active transfer is idle too long -> stop waiting
      - completed records get a short ACK grace window
        Logic is in nvflare/app_common/executors/task_exchanger.py:419.

  So forward progress awareness prevents CJ from resending the same large task while the subprocess is still downloading/materializing it.

###  Reverse Pass: Result Upload
 * Flow: subprocess agent -> CJ/client parent -> downstream receiver pulls result tensors from subprocess

  This direction is different: progress does not need to cross the subprocess/CJ pipe. The subprocess owns the DownloadService transaction and is also the waiter.

  1. Before submitting a result, FlareAgent._do_submit_result() installs:
      - DOWNLOAD_COMPLETE_CB
      - local STREAM_PROGRESS_CB
      - optional RECEIVER_IDS
      - RESULT_UPLOAD_PROGRESS_CTX
        into the cell FOBS context: nvflare/client/flare_agent.py:740, nvflare/client/flare_agent.py:776.
  2. During FOBS serialization, ViaDownloader._finalize_download_tx() creates a DownloadService transaction for result tensors. If progress tracking is available, it captures expected (ref_id, receiver_id) pairs and stores DownloadTransactionInfo: nvflare/
     fuel/utils/fobs/decomposers/via_downloader.py:543.
  3. The CJ ACKs quickly because it only receives lightweight refs. The subprocess then registers the expected pairs in _ReverseResultUploadProgressTracker and waits: nvflare/client/flare_agent.py:807.
  4. As the downstream receiver pulls chunks from the subprocess DownloadService, the source side emits progress with receiver_id=requester: nvflare/fuel/f3/streaming/download_service.py:680, nvflare/fuel/f3/streaming/download_service.py:694.
  5. That local progress callback updates _ReverseResultUploadProgressTracker, which internally uses TransferProgressTracker with direction = result_upload and receiver-scoped keys: nvflare/client/flare_agent.py:64, nvflare/client/flare_agent.py:84.
  6. The reverse wait decision checks:
      - DOWNLOAD_COMPLETE_CB fired -> finish from transaction status
      - any expected pair failed/aborted -> fail
      - all expected pairs completed -> wait brief completion grace, then success
      - started pair stopped progressing past streaming_idle_timeout -> fail
      - unstarted pair never starts past streaming_idle_timeout -> fail
      - otherwise keep waiting
        This is in nvflare/client/flare_agent.py:127.

## Swarm Learning Workflow
• For swarm learning, the same progress-aware design applies, but the “downstream receiver” is usually another client parent, especially the aggregation client.

###  Swarm Forward Pass: Learn-Task Payload Download
  Flow: aggregation/source client parent -> trainer client parent/CJ -> trainer subprocess agent -> source DownloadService

  1. A swarm round starts in SwarmClientController._scatter(). It picks an aggregation client and sends the learn task to remote trainer clients; local self-task is queued separately to avoid self-message deadlock: nvflare/app_common/ccwf/
     swarm_client_ctl.py:483.
  2. The task is stamped with the aggregation client in Constant.AGGREGATOR, and optionally MSG_ROOT_TTL so large streamed model refs stay alive long enough: nvflare/app_common/ccwf/swarm_client_ctl.py:490.
  3. On the receiving trainer client, do_learn_task() prepares the round and forwards the task to the external subprocess. If the task contains lazy refs, it resolves them for CJ-side model bookkeeping but keeps refs in the actual task sent to the
     subprocess, so the subprocess can download directly from the source DownloadService: nvflare/app_common/ccwf/swarm_client_ctl.py:853.
  4. From this point, it is the same forward progress path:
      - subprocess materializes lazy tensors,
      - ViaDownloader stamps direction=task_payload_download,
      - download_object() pulls chunks from the source,
      - subprocess sends STREAM_PROGRESS back to the trainer CJ,
      - trainer TaskExchanger updates TransferProgressTracker.
  5. The key swarm difference is the source of the download: it may be a peer client parent / aggregation client parent, not the FL server. The progress contract does not care; it tracks by job_id, task_id, transfer_id, and direction.
  6. If peer_read_timeout fires before the subprocess ACKs the task, Cell._send_one_request() calls progress_wait_cb() and the trainer CJ keeps waiting while task-payload progress is recent: nvflare/fuel/f3/cellnet/cell.py:428, nvflare/app_common/executors/
     task_exchanger.py:289.

###  Swarm Reverse Pass: Trainer Result Upload
  Flow: trainer subprocess -> trainer CJ/client parent -> aggregation client parent pulls result tensors from trainer subprocess

  1. Before the trainer subprocess runs, do_learn_task() stamps the expected result receiver ID using the selected aggregation client: nvflare/app_common/ccwf/swarm_client_ctl.py:840.
  2. _stamp_result_upload_receiver_ids() creates the receiver identity as FQCN.join([aggr, job_id]) and stores it in FOBSContextKey.RECEIVER_IDS: nvflare/app_common/ccwf/swarm_client_ctl.py:754.
  3. The subprocess receives that header in FlareAgent.get_task() and stores it in _TaskContext.result_receiver_ids: nvflare/client/flare_agent.py:551.
  4. When the subprocess submits the training result, FlareAgent._do_submit_result() installs:
      - DOWNLOAD_COMPLETE_CB
      - local STREAM_PROGRESS_CB
      - RECEIVER_IDS
      - RESULT_UPLOAD_PROGRESS_CTX
      - transaction-created callback
        into the FOBS context: nvflare/client/flare_agent.py:775.
  5. During serialization, ViaDownloader._finalize_download_tx() creates the subprocess-side DownloadService transaction and captures expected (ref_id, receiver_id) pairs. In swarm, receiver_id is the aggregation client FQCN, so progress is scoped to the
     actual downstream puller: nvflare/fuel/utils/fobs/decomposers/via_downloader.py:561, nvflare/fuel/utils/fobs/decomposers/via_downloader.py:581.
  6. The trainer CJ gets a quick ACK because it receives lightweight refs. Then it sends the result to the aggregation client via broadcast_and_wait() on the swarm report-result task: nvflare/app_common/ccwf/swarm_client_ctl.py:991.
  7. The aggregation client decodes the result and pulls tensor chunks from the trainer subprocess DownloadService. DownloadService._handle_download() emits progress with receiver_id=requester, which should match the stamped aggregation-client FQCN:
     nvflare/fuel/f3/streaming/download_service.py:675.
  8. The trainer subprocess updates _ReverseResultUploadProgressTracker locally. No pipe round-trip is needed because the progress source and waiter are in the same subprocess: nvflare/client/flare_agent.py:726.
  9. The reverse wait policy is the same as non-swarm, but receiver-scoped:
      - aggregation client completed pull -> success
      - aggregation client failed/aborted -> fail
      - aggregation client started then stalled past streaming_idle_timeout -> fail
      - expected aggregation client never starts past streaming_idle_timeout -> fail
      - all expected (ref_id, receiver_id) pairs complete -> grace, then success

  So in swarm, progress-aware streaming mainly adds receiver identity. Forward task payload progress is still trainer-subprocess -> trainer-CJ. Reverse result upload progress is trainer-subprocess local tracking while the aggregation client pulls the result
  tensors.

> Net: forward uses progress callbacks across the pipe to keep CJ from resending; reverse uses a local callback inside the subprocess to keep the result source alive until downstream receivers finish pulling.
 
### Swarm Learning

*  tests/unit_test/app_common/ccwf/test_swarm_progress_e2e.py:212
  Simulates a swarm aggregation client scattering a model to a peer trainer. It goes through swarm _scatter(), remote peer task send, real TaskExchanger, real PipeHandler, real CellPipe.send, and real Cell._send_one_request. It verifies delayed ACKs do not
  cause resend while peer-parent download progress is still arriving.

*  tests/unit_test/app_common/ccwf/test_swarm_progress_e2e.py:332
  Simulates a swarm trainer running do_learn_task(). It verifies the trainer stamps the aggregation client receiver ID, then the Client API result-upload path tracks progress for that exact peer receiver before submitting the result.

###  Forward Task Payload

*  tests/unit_test/app_common/executors/task_exchanger_stream_progress_test.py:1402
  Runs the forward task-send stack in one simulated flow: TaskExchanger -> PipeHandler -> CellPipe -> Cell._send_one_request. It delays the peer ACK, injects progress events during the wait, and verifies the send succeeds without retry.

 * tests/unit_test/app_common/executors/task_exchanger_stream_progress_test.py:1469
  Same stack path as above, but no progress events arrive. It verifies the wait stops after the idle budget instead of hanging indefinitely.

 * tests/unit_test/app_common/executors/task_exchanger_stream_progress_test.py:1216
  Simulates 16 clients receiving a large model with delayed ACKs across multiple old timeout windows. It verifies progress events keep each client send alive and prevent duplicate resend.

 * tests/unit_test/app_common/executors/task_exchanger_stream_progress_test.py:1367
  Negative large-broadcast case. Simulates 16 delayed client ACKs with no transfer progress. It verifies all sends fail cleanly instead of waiting forever.

###  Reverse Result Upload

 * tests/unit_test/client/test_reverse_result_upload_progress_wait.py:584
  Simulates 16 clients uploading large results where the final completion callback is delayed. It verifies result-upload progress keeps subprocesses alive past the old fixed timeout until the downstream download completes.

 * tests/unit_test/client/test_reverse_result_upload_progress_wait.py:631
  Negative reverse case. Simulates many result uploads where one client stalls. It verifies the stalled upload fails while progressing clients are allowed to complete.


chesterxgchen:
@greptileai

chesterxgchen:
@greptileai

chesterxgchen:
@greptileai

chesterxgchen:
@greptileai

chesterxgchen:
@greptileai

chesterxgchen:
Addressed the latest Greptile findings in 628d482e5: reverse result-upload now reuses the first accepted record key for each expected transfer to avoid orphan progress records, and task-send no-progress startup waiting is capped by STREAM_PROGRESS_COMPLETION_ACK_GRACE instead of the full streaming_idle_timeout. Added regression tests for both paths.

chesterxgchen:
@greptileai

chesterxgchen:
Handled the open launch_once review thread in b6a0cefb7 by registering the launch_once os._exit atexit hook before result-upload wait/failure paths in PASS_THROUGH mode. Added regression coverage for launch_once=True with a failed result-upload wait. The two race concerns mentioned in the summary are already covered in current code and tests: _normalize_tx_id runs under the reverse tracker lock, and streaming tracker replacement happens under _stream_progress_lock.

chesterxgchen:
@greptileai

chesterxgchen:
Handled the latest summary P2 polish in a821a5548: reverse result-upload clears its wake event before evaluating/waiting to avoid clearing a just-arrived signal after wait, task-send no-progress boundary is now an explicit elapsed >= budget fail branch, and progress tracker pruning can be direction-scoped so TaskExchanger capacity checks avoid rescanning the same direction after prune. Added regression coverage for direction-scoped prune and reverse wait event ordering. Tests: focused progress set 98 passed; adjacent progress/launcher set 214 passed; ./runtest.sh -s passed.

chesterxgchen:
@greptileai

chesterxgchen:
Handled the latest reverse result-upload abandon race in 056903926. _wait_for_reverse_result_upload now evaluates download_done/tracker.decide() before checking abandon reasons, so a completed FINISHED callback wins over concurrent stop/pipe-close signals. Added regression coverage with asked_to_stop=True, pipe_handler stopped, pipe closed, and download_done already set. Tests: focused race tests passed; focused progress set 99 passed; adjacent progress/launcher set 215 passed; ./runtest.sh -s passed.

chesterxgchen:
Handled the empty job_id forward-progress issue in f2ca7ea35. TaskExchanger now treats the empty string job id stamped by execute() as a valid progress scope instead of rejecting it as unscoped, so progress-aware waiting remains enabled when fl_ctx.get_job_id() is None.

Added regression coverage in test_execute_empty_job_id_uses_progress_to_extend_task_send_wait, which exercises the real _do_execute path, verifies the outgoing task header is normalized to "", sends stream progress with job_id="", and confirms _progress_wait_cb() continues waiting after the no-progress budget.

Validation:
- python -m pytest tests/unit_test/app_common/executors/task_exchanger_stream_progress_test.py -q
- python -m pytest tests/unit_test/app_common/executors/client_api_launcher_executor_test.py tests/unit_test/client/ex_process/api_test.py tests/unit_test/client/test_reverse_result_upload_progress_wait.py tests/unit_test/app_common/executors/task_exchanger_stream_progress_test.py tests/unit_test/fuel/f3/streaming/transfer_progress_test.py -q
- ./runtest.sh -s

chesterxgchen:
@greptileai

chesterxgchen:
Handled the latest residual review notes in 65c89e889.

- Moved the reverse result-upload decision timestamp read inside the tracker lock, so the completion-grace timestamp is captured under the same synchronization as the state it is compared with.
- Kept STREAM_PROGRESS excluded from heartbeat/liveness by design for this phase, and added an inline comment in PipeHandler to make that intentional. This matches docs/design/progress_aware_streaming.md, which stages heartbeat integration separately after task resend and result-submit waits are stable.

Validation:
- python -m pytest tests/unit_test/client/test_reverse_result_upload_progress_wait.py tests/unit_test/fuel/utils/pipe/pipe_handler_test.py -q
- python -m pytest tests/unit_test/app_common/executors/client_api_launcher_executor_test.py tests/unit_test/client/ex_process/api_test.py tests/unit_test/client/test_reverse_result_upload_progress_wait.py tests/unit_test/app_common/executors/task_exchanger_stream_progress_test.py tests/unit_test/fuel/f3/streaming/transfer_progress_test.py tests/unit_test/fuel/utils/pipe/pipe_handler_test.py -q
- ./runtest.sh -s

chesterxgchen:
@greptileai

chesterxgchen:
Updated after rechecking the Python-version concern.

- Confirmed setup.cfg declares python_requires >= 3.10 and Python classifiers start at 3.10, so the Python 3.9 import finding does not apply to this branch.
- Restored the py3.10 union annotations in via_downloader.py in 2eb4c8b09.
- Kept the valid reverse-upload completion-grace race fix from a34088d83: once all expected result_upload transfers are completed and the waiter is in completion_grace, stop/pipe-close no longer marks that already-completed upload as failed.
- Added test_reverse_result_upload_completion_grace_wins_over_concurrent_stop for that case.

Validation after the restore:
- python -m pytest tests/unit_test/client/test_reverse_result_upload_progress_wait.py tests/unit_test/fuel/utils/fobs/test_via_downloader_download_timeout.py -q
- python -m pytest tests/unit_test/app_common/executors/client_api_launcher_executor_test.py tests/unit_test/client/ex_process/api_test.py tests/unit_test/client/test_reverse_result_upload_progress_wait.py tests/unit_test/app_common/executors/task_exchanger_stream_progress_test.py tests/unit_test/fuel/f3/streaming/transfer_progress_test.py tests/unit_test/fuel/utils/pipe/pipe_handler_test.py tests/unit_test/fuel/utils/fobs/test_via_downloader_download_timeout.py -q
- ./runtest.sh -s

chesterxgchen:
@greptileai   setup.cfg declares python_requires >= 3.10

chesterxgchen:
Handled the latest GitHub review comments and the pasted integration failure.

Review comments:
- edde969f0 changes the no-progress task-send budget to preserve the configured peer_read_timeout floor, capped by streaming_idle_timeout. This prevents non-tensor tasks from regressing to a 30s timeout just because progress_wait_cb is installed on all task sends.
- Added/updated test coverage in task_exchanger_stream_progress_test.py to prove no-progress sends continue past the first 30s polling interval until the 60s peer_read_timeout budget.
- Resolved the two related review threads.

CI failure from logs:
- The standalone numpy integration failures were ComponentNotAuthorized before transfer logic: NPTrainer and NPFormatter were missing from the curated generated class_allow_list.
- 4ffe39cab adds exact allow-list entries for nvflare.app_common.np.np_trainer.NPTrainer and nvflare.app_common.np.np_formatter.NPFormatter, and updates the exact allow-list unit test.

Validation:
- python -m pytest tests/unit_test/app_common/executors/task_exchanger_stream_progress_test.py -q
- python -m pytest tests/unit_test/lighter/static_file_builder_test.py tests/unit_test/app_common/executors/task_exchanger_stream_progress_test.py -q
- python -m pytest tests/unit_test/app_common/executors/client_api_launcher_executor_test.py tests/unit_test/client/ex_process/api_test.py tests/unit_test/client/test_reverse_result_upload_progress_wait.py tests/unit_test/app_common/executors/task_exchanger_stream_progress_test.py tests/unit_test/fuel/f3/streaming/transfer_progress_test.py tests/unit_test/fuel/utils/pipe/pipe_handler_test.py tests/unit_test/fuel/utils/fobs/test_via_downloader_download_timeout.py -q
- ./runtest.sh -s

chesterxgchen:
Correcting the previous note: the numpy allow-list CI failure is out of scope for this progress-aware streaming PR and is being handled by PR #4737 on 2.8, then cherry-picked to main after merge.

I removed the temporary allow-list commit and its revert from PR #4736 history via a force-with-lease update. PR #4736 now ends at edde969f0 and contains only the progress-aware streaming changes, including the no-progress peer_read_timeout budget fix requested in the latest review threads.

chesterxgchen:
@greptileai

chesterxgchen:
@greptileai

chesterxgchen:
@greptileai

chesterxgchen:
Addressed the reported progress-aware streaming review issues in commit cd0179442.\n\nSummary:\n- Reverse non-swarm result upload: added an explicit high-level test for the common case where no RECEIVER_IDS are present. The test verifies that the upload registers/tracks the single receiver as (ref_id, None), starts on a zero-byte progress event, and fails at streaming_idle_timeout when progress stalls, rather than waiting for the longer download_complete_timeout / transaction TTL.\n- Timeout lock discipline: updated ClientAPILauncherExecutor._apply_streaming_timeout_defaults() so peer_read_timeout and heartbeat_timeout default adjustments are made under _stream_progress_lock, matching the lock used by TaskExchanger._get_streaming_timeout_snapshot().\n- Added unit coverage proving the timeout-default mutation happens under the lock.\n\nValidation run:\npython -m pytest tests/unit_test/client/test_reverse_result_upload_progress_wait.py tests/unit_test/app_common/executors/client_api_launcher_executor_test.py tests/unit_test/fuel/utils/fobs/decomposers/via_downloader_test.py\n\nResult: 174 passed.

chesterxgchen:
@greptileai please review again

chesterxgchen:
Response to latest Greptile summary:\n\n- The remaining crash-detection concern is an acknowledged behavior tradeoff, not an unhandled correctness bug. For non-explicit defaults, peer_read_timeout/heartbeat_timeout are aligned with streaming_idle_timeout so large streamed task payloads are not resent while the subprocess is still materializing data.\n- If a job/operator needs old fast-fail behavior for non-tensor tasks, an explicit lower peer_read_timeout is still honored. The code treats add_client_config()/explicit constructor values as fast-fail overrides and logs a warning when the explicit value is lower than streaming_idle_timeout.\n- The CellPipe send wait still wakes in short bounded chunks; abort_signal and explicit fast-fail paths remain effective.\n- Existing tests cover both sides of the tradeoff: no-progress waits fail at the streaming idle budget, and explicit low peer_read_timeout preserves fast-fail behavior.\n\nGiven the 4/5 safe-to-merge summary and no new blocking issue, I think this is ready with the tradeoff documented.

chesterxgchen:
• Two latest commits:

  eeb548f44 Preserve startup timeout for no-progress task sends

  - Keeps peer_read_timeout as the startup/no-progress task-send budget.
  - Uses streaming_idle_timeout only after stream progress exists.
  - Stops stretching non-tensor/no-progress task crash detection to the full 600s default.
  - Updates design doc and related timeout tests.

  564ea0166 Cover no-progress non-tensor task send timeout

  - Adds an explicit regression test for the Greptile concern.
  - Verifies non-tensor/no-progress tasks fail at peer_read_timeout=60, not streaming_idle_timeout=600.
  - Confirms progress-aware streaming does not regress ordinary non-stream task dead-peer detection.

chesterxgchen:
@greptileai please review again

chesterxgchen:
Addressed the completed-record ACK window concern in commit 353577cae.\n\nSummary:\n- Changed the completed task-payload progress path to use the bounded task-send ACK budget instead of a fixed 30s grace.\n- Kept the CellPipe polling interval capped at 30s, so the loop wakes frequently without forcing a premature failure after one poll.\n- Added regression coverage for both cases: completed progress continues past the first 30s post-completion poll, and a delayed ACK after that poll completes the task send successfully.\n- Updated the progress-aware streaming design doc to reflect the separate startup, active-progress, and completed-transfer ACK budgets.\n\nVerification:\n- python -m pytest tests/unit_test/app_common/executors/task_exchanger_stream_progress_test.py\n- python -m pytest tests/unit_test/app_common/executors/client_api_launcher_executor_test.py tests/unit_test/app_common/executors/task_exchanger_stream_progress_test.py\n- ./runtest.sh -s

chesterxgchen:
@greptileai please review

chesterxgchen:
Addressed the latest startup-budget / reverse-tracker validation concern in commit 6f3f5f838.\n\nSummary:\n- Kept the design doc unchanged in this pass and validated implementation against the existing design policy.\n- Split the task-send startup budget helper from the completed-transfer ACK budget helper so the two phases are explicit in code.\n- Added coverage for explicit low peer_read_timeout on non-tensor/no-progress tasks to prove fast-fail behavior is preserved when users configure it.\n- Added reverse result-upload coverage proving a rejected first event does not seed record_keys or create orphan tracker records, and the first valid event establishes the stable record key.\n\nVerification:\n- python -m pytest tests/unit_test/app_common/executors/task_exchanger_stream_progress_test.py tests/unit_test/client/test_reverse_result_upload_progress_wait.py\n- python -m pytest tests/unit_test/app_common/executors/client_api_launcher_executor_test.py tests/unit_test/app_common/executors/task_exchanger_stream_progress_test.py tests/unit_test/client/test_reverse_result_upload_progress_wait.py\n- ./runtest.sh -s

chesterxgchen:
> The main remaining concern is behavioral: the no-progress task-send startup budget is bounded by min(streaming_idle_timeout, peer_read_timeout_floor), which for certain peer_read_timeout configurations silently changes the effective timeout for non-tensor tasks that never emit progress events. Crash-detection latency for subprocesses also extends to the idle budget window when no progress records exist.

For the no-progress startup budget concern:

  - The implementation now has separate helpers:
      - _get_task_send_startup_budget(...)
      - _get_task_send_completed_ack_budget(...)
  - This makes the no-progress path distinct from the completed-transfer ACK path.
  - Added test_explicit_low_peer_read_timeout_non_tensor_task_fast_fails to prove explicitly configured low peer_read_timeout still fast-fails for non-tensor/no-progress tasks.

  For the reverse tracker concern:

  - Added test_result_upload_rejected_first_event_does_not_seed_record_key.
  - It proves a rejected first progress event does not populate record_keys and does not create a tracker record.
  - Then it verifies the first valid event establishes the stable (job_id, task_id) record key.

  Commit pushed: 6f3f5f838 Validate task-send startup budget behavior

chesterxgchen:
@greptileai

chesterxgchen:
> with one minor edge case in startup-budget calculation when peer_read_timeout is None.


  How it is currently handled:

  - If peer_read_timeout=None, CellPipe task-send uses a bounded 30s poll/startup budget.
  - That is intentional in this PR because otherwise a no-progress task send can wait too long or become effectively unbounded while no stream progress exists.
  - The default case is unaffected: default peer_read_timeout=60 still gives a 60s no-progress startup budget, not 30s and not 600s.
  - Explicit low numeric timeouts still fast-fail, covered by the test added in 6f3f5f838.

  Existing tests covering the None edge:

  - test_task_send_no_progress_budget_uses_startup_grace_when_peer_read_timeout_disabled
  - test_task_send_peer_read_timeout_disabled_still_aborts_failed_effective_send
  - Launcher config test: test_peer_read_none_does_not_warn_when_streaming_timeout_is_enabled

  For the max_resends=3 note: I checked the touched path. This review did not identify a concrete defect there, just called out the interaction. Existing tests cover default/custom/None max-resends serialization
  and validation. I would not change code for this review unless they open a specific failure case.

  So my recommendation: no new code change for this review summary. It is safe to reply that the remaining peer_read_timeout=None behavior is intentional and already covered by tests.

@greptileai

chesterxgchen:
Handled the new peer_read_timeout=None review thread in commit aaa0da32c.\n\nSummary:\n- When peer_read_timeout is explicitly disabled (None), task-send still polls CellPipe every 30s, but the no-progress startup budget now uses the full streaming_idle_timeout instead of failing at the first 30s poll.\n- Default behavior is unchanged: peer_read_timeout=60 still gives a 60s no-progress startup budget.\n- Added regression coverage proving peer_read_timeout=None continues past the first 30s poll and fails only when the idle budget is exhausted if no progress ever arrives.\n\nVerification:\n- python -m pytest tests/unit_test/app_common/executors/task_exchanger_stream_progress_test.py\n- python -m pytest tests/unit_test/app_common/executors/client_api_launcher_executor_test.py tests/unit_test/app_common/executors/task_exchanger_stream_progress_test.py tests/unit_test/client/test_reverse_result_upload_progress_wait.py\n- ./runtest.sh -s

chesterxgchen:
@greptileai please review again

chesterxgchen:
@greptileai — both P1 findings addressed in 2d56456e3.

**P1 #1 — `_get_task_send_startup_budget` None-path capped instead of unbounded**

`_get_task_send_startup_budget` now returns `Optional[float]` and yields `None` when `peer_read_timeout is None`. `_should_continue_task_send_waiting` treats `wait_budget is None` as unbounded and logs `task_send_wait_budget=unbounded`. New test `test_task_send_peer_read_timeout_disabled_polls_without_startup_budget` exercises a 1200+ second poll without exit.

**P1 #2 — `_apply_streaming_timeout_defaults` silently elevating `heartbeat_timeout`**

`_apply_streaming_timeout_defaults` is now a no-op. `heartbeat_timeout` keeps its configured liveness/crash-detection semantics regardless of `streaming_idle_timeout`. The downstream per-request check (`_warn_if_streaming_per_request_inconsistent`) warns when `heartbeat_timeout < streaming_per_request_timeout` instead of mutating the value, so users get a clear signal without a silent 10× extension of crash detection. Test `test_streaming_timeout_defaults_preserve_peer_read_and_heartbeat` asserts both `peer_read_timeout` and `heartbeat_timeout` stay at their configured values after the call.

Design doc updated in the same commit to reflect: `effective_heartbeat_timeout = heartbeat_timeout` (preserved), and `no_progress_task_send_budget = None` when `peer_read_timeout is None`.

chesterxgchen:
@greptileai — good catch on the remaining heartbeat fallback. Addressed in `39f69d206` (was `9ed4f3e6d` pre-rebase).

`_validate_timeout_config` no longer mutates `heartbeat_timeout` when it is `None`. The warning still fires telling the user to set `heartbeat_timeout >= streaming_per_request_timeout` in job config. `PipeHandler.__init__` will refuse `heartbeat_timeout=None` at startup with a clear `TypeError` (via `check_non_negative_number`) — a fail-fast startup error is preferable to a silent crash-detection extension.

Test renamed from `test_heartbeat_timeout_none_is_corrected_when_per_req_is_configured` to `test_heartbeat_timeout_none_warns_without_mutation` and asserts `executor.heartbeat_timeout is None` post-validation.

chesterxgchen:
Rebased onto `upstream/main` (HEAD = `39f69d206`). No conflicts — `main` advanced by 2 commits (#4742 Kubernetes CA docs, #4745 artifact writer hardening) and neither touched files modified by this PR.

For Greptile / reviewer reference, the two fixes addressing the P1 findings have new SHAs after the rebase:

- **P1 #1 + #2** (`_get_task_send_startup_budget` None-path; `_apply_streaming_timeout_defaults` heartbeat elevation) — was `2d56456e3`, now **`01b185923`** "Decouple streaming_idle_timeout from heartbeat fast-fail".
- **Follow-up** (drop `heartbeat_timeout=None` fallback mutation in `_validate_timeout_config`) — was `9ed4f3e6d`, now **`39f69d206`** "Drop heartbeat_timeout=None fallback mutation in validate path". @Greptileai

The fix content is unchanged; only the commit SHAs moved due to the rebase.

chesterxgchen:
Addressed the remaining concrete Greptile performance concern in 2fe409bcd.

- DownloadService chunk byte accounting no longer takes the class-wide _tx_lock on every successful chunk.
- Added per-transaction stats locking via add_total_bytes()/get_total_bytes(), preserving the final transaction summary while avoiding global hot-path contention.
- Rechecked the TaskExchanger timeout-budget and FlareAgent reverse-upload race notes: those are already covered by the current resolved threads/fixes.

Validation:
- python -m pytest tests/unit_test/fuel/f3/streaming/download_service_test.py
- python -m pytest tests/unit_test/fuel/f3/streaming/transfer_progress_test.py tests/unit_test/fuel/f3/streaming/download_service_test.py tests/unit_test/fuel/f3/streaming/download_object_test.py tests/unit_test/fuel/utils/fobs/decomposers/via_downloader_test.py tests/unit_test/fuel/utils/fobs/test_via_downloader_download_timeout.py tests/unit_test/fuel/utils/pipe/cell_pipe_test.py tests/unit_test/fuel/utils/pipe/pipe_handler_test.py tests/unit_test/app_common/executors/task_exchanger_stream_progress_test.py tests/unit_test/app_common/executors/client_api_launcher_executor_test.py tests/unit_test/client/test_reverse_result_upload_progress_wait.py
- ./runtest.sh -s

@greptileai please review again

chesterxgchen:
Addressed the current Greptile comments in d1dc71fed.

- Restored heartbeat_timeout=None auto-correction when streaming_per_request_timeout is configured, so PipeHandler initialization does not regress for existing configs.
- Refreshed the reverse result-upload tracker decision after abandon detection before failing, so completion that arrives between the original decision snapshot and abandon check wins correctly.
- Added regression coverage for both paths.

Validation:
- python -m pytest tests/unit_test/app_common/executors/client_api_launcher_executor_test.py tests/unit_test/client/test_reverse_result_upload_progress_wait.py
- broader progress-aware streaming suite: 363 passed
- ./runtest.sh -s

@greptileai please review again

chesterxgchen:
@greptileai

chesterxgchen:
Addressed the latest Greptile forward task-send progress finding in 2c1dcd89c. _should_continue_task_send_waiting now checks recent activity from active task-payload records only, so a recently completed sibling transfer can no longer mask a stalled active transfer. Added regression coverage for the multi-ref completed-plus-stalled case. Reverse result-upload tests were re-run as adjacent coverage. @greptileai please review again

chesterxgchen:
Addressed the reverse result-upload stalled-active masking finding in efd90ef50. _ReverseResultUploadProgressTracker.decide() now ignores terminal records when deciding whether recent activity should keep the wait alive, so a recently completed receiver cannot mask an idle active receiver. Added regression coverage for the completed-plus-stalled multi-ref timing case. @greptileai please review again

chesterxgchen:
Addressed the latest two Greptile P1 findings in a791baf41.\n\n- Forward task-send retries: an explicit new start event now clears a previous failed/aborted terminal record for the same task-payload transfer, so stale FAILED records do not permanently abort retries while current-attempt failures still fail fast.\n- Reverse result-upload: _ReverseResultUploadProgressTracker.decide() now enforces idle/no-start timeout per expected pair instead of using group-wide latest activity, so recent activity from one pair cannot mask a stalled or unstarted sibling pair.\n\nAdded regression tests for both paths. Validation: task_exchanger_stream_progress_test.py + test_reverse_result_upload_progress_wait.py passed; transfer_progress_test.py + client_api_launcher_executor_test.py passed; ./runtest.sh -s passed.\n\n@greptileai please review again

chesterxgchen:
Follow-up for the inline stale FAILED record thread is pushed in 9440719c7. _should_continue_task_send_waiting now ignores FAILED/ABORTED records whose last_progress_time is before the current send_start_time, and the retry start-event reset from a791baf41 remains in place for same-transfer restart events. Added regression coverage for the send_start_time filter. The inline thread has been marked resolved.

chesterxgchen:
@greptileai

IsaacYangSLA (approved):
This is a fundamental change on the streaming behavior, when timeout is used.  Approved as it solves a real-world issue.

nvidianz (approved):

PR #4739, opened 2026-06-01
[2.8] Fix FOBS externalizer rebuilding dict-subclass models with a no-arg constructor

nvidianz:
## Summary
- Fix a regression (NVBug 6247985) where `SelectionRequest` and similar dict/list subclasses with required constructor arguments fail to FOBS-serialize across processes, breaking external-process edge jobs.
- `Externalizer` now rebuilds containers without invoking `__init__` when the no-arg constructor is unavailable, while preserving the subclass type.

## Root Cause
PR #4707 changed `Externalizer.externalize()` to build fresh containers via `type(target)()` (instead of mutating caller-owned containers) to fix a shared-state race. Because FOBS packs with `strict_types=True`, every `dict`/`list` subclass is routed through this generic externalizer, which recurses into the raw object tree. When it reaches a `SelectionRequest` — a `BaseModel(dict)` whose `__init__(device_info, job_id)` has required positional arguments — `type(target)()` raises:

```
TypeError: SelectionRequest.__init__() missing 2 required positional arguments: 'device_info' and 'job_id'
```

so the edge device-selection message can never be encoded. This is a regression: the same job passes on 2.8.0rc3 and fails on rc4/rc5. In-process variants are unaffected because they don't serialize the object across the cell.

## Fix
Reconstruct the empty container via `type(target).__new__(type(target))` (bypassing `__init__`) when the no-arg constructor raises `TypeError`. The subclass type is still preserved — not downgraded to a plain `dict`/`list` — so `strict_types` packing routes the value back through its decomposer. The shared-state-race fix from #4707 is retained (containers are still freshly built, never mutated in place).

The decode side is already safe: `SelectionRequest` is whitelisted and auto-registers as a `DataClassDecomposer`, whose `recompose()` already uses `__new__`.

## Validation
- New regression tests reproduce the exact `TypeError` against the unfixed code and pass with the fix.
- `tests/unit_test/fuel/utils/fobs`: 62 passed (60 existing + 2 new).
- Manual round-trip of a `SelectionRequest` nested in a `Shareable` (with a large `bytes` leaf) preserves type, values, and the binary payload.
- `black` / `isort` / `flake8` clean on changed files.


chesterxgchen (commented):
Codex found this: 

 - P2 nvflare/fuel/utils/fobs/decomposer.py:156: the new required-arg list-subclass path can serialize successfully while dropping the list contents. Externalizer rebuilds a list subclass and appends items, but FOBS then routes the subclass through DataClassDecomposer, which only records __dict__ and dict contents, not list elements. I reproduced this on the PR head: a required-arg list subclass containing ["a", "b"] round-trips through fobs.dumps/loads as an empty list. The added test at tests/unit_test/fuel/utils/fobs/decomposer_test.py:169 only checks Externalizer directly, so it misses the real FOBS behavior. I’d either add real list-subclass content support or keep unsupported list subclasses failing instead of allowing silent data loss.

nvidianz:
## Code Review

**Verdict: Approve.** Correct, minimal, well-tested, CI green, and it resolves both Greptile review points. Notes below; the polish items have been pushed as `40f3ff071`.

### Overview
Fixes the NVBug 6247985 regression from #4707, where `Externalizer` rebuilt containers via `type(target)()` and raised `TypeError` for dict/list subclasses whose `__init__` has required positional args (e.g. edge `SelectionRequest(device_info, job_id)`). The fix has three coherent layers:
1. **Core fix** — `new_empty_container(cls)` bypasses `__init__` via `__new__` only when the constructor signature requires args, preserving the exact subclass type so `strict_types` packing still routes through the right decomposer.
2. **Encode/decode symmetry** — `DictDecomposer.recompose()` reuses the same helper, closing the latent decode-side trap.
3. **List-subclass round-trip** — `DataClassDecomposer` now captures/restores `LIST_CONTENT`; without it a list subclass routed through auto-registration would silently lose its elements.

### Strengths
- Precise over broad: the cached `_requires_init_args` signature check means a `TypeError` raised *inside* a valid no-arg constructor surfaces instead of being swallowed (guarded by a test).
- Performance-aware: `@lru_cache` keeps `inspect.signature` out of the per-message hot path.
- Strong coverage: externalize-only paths, full `fobs.dumps/loads` round-trip, and the error-masking guard.

### Applied polish (pushed)
- Bounded the signature cache (`maxsize=1024` instead of `None`) so it can't grow without limit for dynamically-generated classes.
- Documented that `_requires_init_args` reflects `__new__` when `__init__` is inherited (args-requiring `__new__` is out of scope), and that `DataClassDecomposer` captures dict/list contents in addition to instance `__dict__`.

### One limitation worth recording (not a blocker)
`Externalizer` rebuilds a nested dict/list subclass by copying only its **items/elements**, not its instance `__dict__`. So instance attributes set outside the container (e.g. a `capacity` attr on a `list` subclass) are dropped when the value flows through the externalize path. This is pre-existing behavior and **does not affect the edge models**: `BaseModel.__setattr__` stores all state in the dict itself (`self[key]`), not in `__dict__`, so `SelectionRequest` et al. round-trip fully. I dropped a tentative `capacity` assertion from the list round-trip test once I confirmed this — the test correctly asserts element preservation only.

`set`/`tuple` subclasses remain unsupported, consistent with prior behavior. No new security surface — `__new__`-based construction was already used by `DataClassDecomposer.recompose`, and the whitelist/trust model in `fobs.py` is untouched.

chesterxgchen (approved):
LGTM

chesterxgchen:
Lets not quickly merge this, I want to use this PR to test, to  make sure CI passes for 2.8

PR #4701, opened 2026-05-27
[2.8]: cherry-pick PR 4684 and 4704 Improve security by loading classes in allow_list only

IsaacYangSLA:
Address comments

Switch to block_list

Update doc and improve coverage of unit tests

Address comments

Add public API

Fix format issue

Revert back to allow_list

Fix one unit test

Address new comments

Improve more

Address more comments

Update the class_allow_list

Address new comments

Add cache to speed up

Update document for more clarities
### Description

Cherry-pick PR#4684 and PR#4704

### Types of changes

- [ ] Non-breaking change (fix or new feature that would not break existing functionality).
- [x] Breaking change (fix or new feature that would cause existing functionality to change).
- [x] New tests added to cover the changes.
- [x] Quick tests passed locally by running `./runtest.sh`.
- [x] In-line docstrings updated.
- [x] Documentation updated.

PR #4740, opened 2026-06-01
[2.8] Confine built-in artifact writer paths to the job directory

pcnudde:
## Summary

- Several built-in artifact writers/controllers joined config-supplied paths onto the job run directory with no containment, so an absolute path or `..` traversal in their constructor args escaped the job-owned directory:
  - `CrossSiteModelEval` — `cross_val_dir` reaches `shutil.rmtree` + `os.makedirs` (arbitrary directory delete/create).
  - `ValidationJsonGenerator` — `results_dir` / `json_file_name` reach `open(..., "w")` + `json.dump` (arbitrary file write).
  - `TBAnalyticsReceiver` — `tb_folder` reaches `os.makedirs` + TensorBoard event writes (arbitrary directory create).
- Route all three through `file_utils.resolve_path_under_root`, the same containment helper added for the other built-in writers in #4738. Relative paths that stay inside the job directory are unchanged; absolute paths and parent-directory traversal now raise `ValueError`.
- Adds positive and negative unit tests for each of the three classes.

## Context

These are the three sites flagged during review of #4738 that share the same config-path to filesystem-sink pattern but were not converted there. Follow-up to #4738; relevant to the non-BYOC `class_allow_list` hardening in #4701 / #4737.

### DIFF
--- a/nvflare/app_common/widgets/validation_json_generator.py
+++ b/nvflare/app_common/widgets/validation_json_generator.py
@@ -23,6 +23,7 @@ from nvflare.apis.event_type import EventType
 from nvflare.apis.fl_context import FLContext
 from nvflare.app_common.app_constant import AppConstants
 from nvflare.app_common.app_event_type import AppEventType
+from nvflare.app_common.utils.file_utils import resolve_path_under_root
 from nvflare.widgets.widget import Widget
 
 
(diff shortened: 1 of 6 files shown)


pcnudde:
Thanks for the review. Both points addressed in 534dfc203:

1. **`record_origin` join in `save()`** — now routed through `resolve_path_under_root(self.root_log_dir, record_origin, "record_origin")`, consistent with the `tb_folder` fix in the same file. One clarification: `record_origin` is not job-submitter-controlled — it's the authenticated peer identity (`_get_record_origin` → `data.get_peer_prop(IDENTITY_NAME)` / `fl_ctx.get_identity_name()`), i.e. the provisioned client name. So this is defense-in-depth against a hierarchical/malformed client name rather than the job-config vector the rest of the PR targets, but it's cheap and the right call.

2. **Missing positive `cross_val_dir` test** — added `test_start_controller_accepts_relative_cross_val_dir`, which confirms a valid relative `cross_val_dir` is accepted and both cross-val subdirectories are created.

All unit tests pass; black/isort/flake8 clean.

chesterxgchen (approved):
LGTM

PR #4759, opened 2026-06-03
Bind CellNet mTLS endpoint identity

nvidianz:
## Summary

Bind CellNet endpoint names to the authenticated mTLS peer certificate identity so a peer cannot claim another `endpoint_name` during SFM handshake or certificate exchange.

This adds a CellNet identity resolver that supports exact FQCN identities, configured prefix identities, local child identity inference, and the job-cell case where the endpoint is `site-1.<job_id>` but the certificate identity remains `site-1`.

It also adds explicit config support for deployments where a site certificate CN intentionally differs from the endpoint/FQCN: use `auth_identity` for the local cert CN and `auth_identity_map` for remote FQCN/prefix to expected cert CN mappings. Existing `identity` values remain the fallback for compatibility. Redundant identity-map entries are omitted when the resolver already derives the same expected CN.

Follow-up POC changes improve local startup behavior: POC services launch with the active `nvflare` CLI Python, optional admin-console startup reports service failures without repeated connection spam, authenticated admin connections are allowed for the console, and redirected service console logs are written without ANSI color control characters.

## Root Cause

The SFM handshake trusted the peer-supplied `endpoint_name` independently of the mTLS certificate common name. Certificate exchange also cached certificates by the claimed origin FQCN without checking that the certificate identity matched that origin.

## Changes

- Enforce endpoint/CN binding in `ConnManager.update_endpoint()` only for mTLS connections.
- Validate cached peer cert identity in `CredentialManager` only when the CellNet config uses mTLS.
- Propagate configured identities through client, job-cell, relay, and root connection properties.
- Add `auth_identity` / `auth_identity_map` runtime config and provisioning template support for cert CNs that differ from endpoint/FQCN.
- Omit redundant auth identity mappings for direct flat sites so package/provision kit parity stays stable.
- Document `auth_identity` configuration and warn users to regenerate signed startup kits instead of editing startup JSON when `startup/signature.json` is present.
- Improve POC startup/admin-console UX by using the CLI Python for launched services, surfacing service startup failures once, and allowing authenticated admin POC connections.
- Keep redirected POC console logs plain by skipping ANSI colors when console output is not attached to a TTY.
- Add job-cell, relay-child, configured-CN, POC startup, and redirected-console-log coverage.

## Validation

- `/Users/zhihongz/nvflare-env/3.12/bin/python -m pytest tests/unit_test/tool/package/package_commands_test.py -v` (`226 passed`)
- `/Users/zhihongz/nvflare-env/3.12/bin/python -m pytest tests/unit_test/fuel/f3/cellnet/identity_binding_test.py tests/unit_test/lighter/static_file_builder_test.py tests/unit_test/lighter/poc_commands_test.py -v` (`41 passed`)
- `/Users/zhihongz/nvflare-env/3.12/bin/python -m pytest tests/unit_test/fuel/f3 -v` (`222 passed`, before the config follow-up; focused identity tests rerun after)
- `/Users/zhihongz/nvflare-env/3.12/bin/python -m pytest tests/unit_test/private/fed/client/communicator_test.py tests/unit_test/private/fed/utils/site_config_test.py -v` (`7 passed`)
- `/Users/zhihongz/nvflare-env/3.12/bin/python -m pytest tests/unit_test/fuel/log_utils_dynamic_config_test.py -q` (`7 passed`)
- `PATH=/Users/zhihongz/nvflare-env/3.12/bin:$PATH ./runtest.sh --skip-install -s nvflare/fuel/utils/log_utils.py`
- `PATH=/Users/zhihongz/nvflare-env/3.12/bin:$PATH ./runtest.sh --skip-install -s tests/unit_test/fuel/log_utils_dynamic_config_test.py`
- Fresh `nvflare poc start` appended plain server console output with `escape_count 0` and no unexpected control bytes in `/tmp/nvflare/poc/example_project/prod_00/server/poc_console.log`.
- `/Users/zhihongz/nvflare-env/3.12/bin/python -m black --check ...changed files...`
- `/Users/zhihongz/nvflare-env/3.12/bin/python -m isort --check ...changed files...`
- `/Users/zhihongz/nvflare-env/3.12/bin/python -m flake8 ...changed files... --count --statistics`
- `git diff --check`

## Note

`./runtest.sh -s` was run with `VIRTUAL_ENV=/Users/zhihongz/nvflare-env/3.12` and failed on unrelated vendored files under `examples/.../analytics-dashboard/node_modules/.../flatted/python/flatted.py` that Black would reformat. The scoped style checks above passed for this PR's changed files.

### DIFF
--- a/nvflare/fuel/f3/cellnet/credential_manager.py
+++ b/nvflare/fuel/f3/cellnet/credential_manager.py
@@ -21,6 +21,7 @@ from cryptography.x509 import Certificate
 
 from nvflare.fuel.f3.cellnet.cell_cipher import SimpleCellCipher
 from nvflare.fuel.f3.cellnet.defs import MessageHeaderKey
+from nvflare.fuel.f3.cellnet.identity import CellIdentityResolver, get_cert_common_name_from_pem
 from nvflare.fuel.f3.drivers.driver_params import DriverParams
 from nvflare.fuel.f3.endpoint import Endpoint
 from nvflare.fuel.f3.message import Message
@@ -38,11 +39,18 @@ CERT_REQ_TIMEOUT = 10
 class CredentialManager:
     """Helper class for secure message. It holds the local credentials and certificate cache"""
 
-    def __init__(self, local_endpoint: Endpoint):
+    def __init__(
+        self,
+        local_endpoint: Endpoint,
+        identity_resolver: CellIdentityResolver = None,
+        enforce_identity: bool = False,
+    ):
 
         self.local_endpoint = local_endpoint
         self.cert_cache = {}
         self.lock = threading.Lock()
+        self.identity_resolver = identity_resolver if identity_resolver else CellIdentityResolver(local_endpoint.name)
+        self.enforce_identity = enforce_identity
 
         conn_props = self.local_endpoint.conn_props
         ca_cert_path = conn_props.get(DriverParams.CA_CERT)
@@ -99,6 +107,19 @@ class CredentialManager:
 
         return req
 
+    def _cache_cert(self, fqcn: str, cert: bytes):
+        if not cert:
+            raise RuntimeError(f"missing certificate for {fqcn}")
+
+        if self.enforce_identity:
+            cn = get_cert_common_name_from_pem(cert)
+            try:
+                self.identity_resolver.require_match(fqcn, cn, f"certificate for {fqcn}")
+            except ValueError as ex:
+                raise RuntimeError(str(ex))
+
+        self.cert_cache[fqcn] = cert
+
     def process_request(self, request: Message) -> dict:
         origin = request.get_header(MessageHeaderKey.ORIGIN)
         target = request.get_header(MessageHeaderKey.DESTINATION)
@@ -110,7 +131,7 @@ class CredentialManager:
             cert = payload.get(CERT_CONTENT)
 
             # Save cert from requester in the cache
-            self.cert_cache[origin] = cert
+            self._cache_cert(origin, cert)
 
             reply[CERT_CONTENT] = self.local_cert
             reply[CERT_CA_CONTENT] = self.ca_cert
@@ -125,7 +146,7 @@ class CredentialManager:
             raise RuntimeError(f"Request to get certificate from {origin} failed: {error}")
 
         cert = reply.get(CERT_CONTENT)
-        self.cert_cache[origin] = cert
+        self._cache_cert(origin, cert)
         return cert
 
     def get_local_cert(self) -> Certificate:
--- a/nvflare/fuel/f3/cellnet/credential_manager.py
+++ b/nvflare/fuel/f3/cellnet/credential_manager.py
@@ -97,7 +97,8 @@ class CredentialManager:
     def get_certificate(self, fqcn: str) -> bytes:
         if not self.cell_cipher:
             raise RuntimeError("This cell doesn't support certificate exchange, not running in secure mode")
-        return self.cert_cache.get(fqcn)
+        with self.lock:
+            return self.cert_cache.get(fqcn)
 
     def create_request(self) -> dict:
         req = {
@@ -118,7 +119,8 @@ class CredentialManager:
             except ValueError as ex:
                 raise RuntimeError(str(ex))
 
-        self.cert_cache[fqcn] = cert
+        with self.lock:
+            self.cert_cache[fqcn] = cert
 
     def process_request(self, request: Message) -> dict:
         origin = request.get_header(MessageHeaderKey.ORIGIN)
--- a/nvflare/fuel/f3/cellnet/identity.py
+++ b/nvflare/fuel/f3/cellnet/identity.py
@@ -97,6 +97,13 @@ class CellIdentityResolver:
     def _is_descendant(fqcn: str, prefix: str) -> bool:
         return fqcn.startswith(prefix + ".")
 
+    def _is_local_descendant_with_ancestor_prefix(self, fqcn: str, prefix: str) -> bool:
+        return (
+            self.local_fqcn
+            and self._is_descendant(fqcn, self.local_fqcn)
+            and self._is_descendant(self.local_fqcn, prefix)
+        )
+
     def _resolve_local_child_identity(self, fqcn: str) -> Optional[str]:
         if not self.local_fqcn or not self._is_descendant(fqcn, self.local_fqcn):
             return None
@@ -119,6 +126,9 @@ class CellIdentityResolver:
         parts = FQCN.split(fqcn)
         for i in range(len(parts), 0, -1):
             prefix = FQCN.join(parts[:i])
+            if self._is_local_descendant_with_ancestor_prefix(fqcn, prefix):
+                continue
+
             identity = self.prefix_identity_map.get(prefix)
             if identity:
                 return identity
--- a/nvflare/fuel/f3/cellnet/identity.py
+++ b/nvflare/fuel/f3/cellnet/identity.py
@@ -131,6 +131,9 @@ class CellIdentityResolver:
 
     def require_match(self, fqcn: str, peer_cn: str, peer_desc: str):
         expected_cn = self.resolve(fqcn)
+        if not expected_cn:
+            raise ValueError(f"{peer_desc} claimed endpoint '{fqcn}' does not resolve to an expected identity")
+
         if not peer_cn or peer_cn == "N/A":
             raise ValueError(f"{peer_desc} does not have an authenticated mTLS peer common name")
 
--- a/nvflare/fuel/f3/cellnet/identity.py
+++ b/nvflare/fuel/f3/cellnet/identity.py
@@ -1,0 +1,141 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from typing import Optional
+
+from cryptography import x509
+from cryptography.x509.oid import NameOID
+
+from nvflare.apis.fl_constant import ConnectionSecurity
+from nvflare.fuel.f3.cellnet.fqcn import FQCN
+from nvflare.fuel.f3.drivers.driver_params import DriverParams
+from nvflare.fuel.f3.drivers.net_utils import SECURE_SCHEMES
+from nvflare.fuel.utils.argument_utils import str2bool
+
+
+def get_param(params: dict, key: DriverParams, default=None):
+    if not params:
+        return default
+
+    value = params.get(key.value, None)
+    if value is None:
+        value = params.get(key, default)
+    return value
+
+
+def is_mtls_connection(params: dict) -> bool:
+    if not params:
+        return False
+
+    scheme = get_param(params, DriverParams.SCHEME)
+    secure = str2bool(get_param(params, DriverParams.SECURE, False))
+    if scheme not in SECURE_SCHEMES and not secure:
+        return False
+
+    conn_security = get_param(params, DriverParams.CONNECTION_SECURITY, ConnectionSecurity.MTLS)
+    return conn_security == ConnectionSecurity.MTLS
+
+
+def is_mtls_config(credentials: dict, secure: bool) -> bool:
+    if not secure:
+        return False
+
+    conn_security = get_param(credentials, DriverParams.CONNECTION_SECURITY, ConnectionSecurity.MTLS)
+    return conn_security == ConnectionSecurity.MTLS
+
+
+def get_cert_common_name(cert: x509.Certificate) -> Optional[str]:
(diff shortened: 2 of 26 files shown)


nvflare/fuel/f3/sfm/conn_manager.py
```diff
@@ -404,9 +406,17 @@ def update_endpoint(self, sfm_conn: SfmConnection, data: dict):
...
             )
 
+        conn_props = sfm_conn.conn.get_conn_properties()
+        if is_mtls_connection(sfm_conn.conn.connector.params):
+            peer_cn = get_param(conn_props, DriverParams.PEER_CN)
+            try:
+                self.identity_resolver.require_match(endpoint_name, peer_cn, f"connection {sfm_conn.get_name()}")
+            except ValueError as ex:
+                sfm_conn.conn.close()
+                raise CommError(CommError.BAD_DATA, str(ex))
```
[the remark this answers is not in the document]
nvidianz: Fixed in `6c69e2d3e`: mTLS endpoint identity enforcement now runs only on passive connections, so active gRPC clients that do not expose `PEER_CN` are accepted. Covered by `test_mtls_active_connection_without_peer_cn_is_accepted` for both `None` and `"N/A"` peer CN values.

nvflare/fuel/f3/cellnet/identity.py
```diff
@@ -0,0 +1,141 @@
...
+    def require_match(self, fqcn: str, peer_cn: str, peer_desc: str):
+        expected_cn = self.resolve(fqcn)
+        if not peer_cn or peer_cn == "N/A":
+            raise ValueError(f"{peer_desc} does not have an authenticated mTLS peer common name")
+
+        if peer_cn != expected_cn:
+            raise ValueError(
+                f"{peer_desc} authenticated as '{peer_cn}' but claimed endpoint '{fqcn}' "
+                f"requires identity '{expected_cn}'"
+            )
```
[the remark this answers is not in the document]
nvidianz: Fixed in `6c69e2d3e`: `require_match` now rejects an endpoint whose FQCN does not resolve to an expected identity before evaluating the peer CN. Covered by `test_identity_resolver_rejects_unresolvable_endpoint_identity`.

nvflare/fuel/f3/cellnet/credential_manager.py
```diff
@@ -99,6 +107,19 @@ def create_request(self) -> dict:
...
+            raise RuntimeError(f"missing certificate for {fqcn}")
+
+        if self.enforce_identity:
+            cn = get_cert_common_name_from_pem(cert)
+            try:
+                self.identity_resolver.require_match(fqcn, cn, f"certificate for {fqcn}")
+            except ValueError as ex:
+                raise RuntimeError(str(ex))
+
+        self.cert_cache[fqcn] = cert
```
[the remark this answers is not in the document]
nvidianz: Fixed in `6c69e2d3e`: certificate cache writes and reads now go through `self.lock`. Covered by `test_certificate_cache_access_uses_lock`, which verifies both `_cache_cert` and `get_certificate` acquire the lock.

nvflare/private/fed/app/fl_conf.py
```diff
@@ -333,20 +342,26 @@ def _determine_conn_props(self, client_name, config_data: dict):
...
                 ConnPropKey.CONNECTION_SECURITY: self.args.parent_conn_sec,
             }
         else:
             # this is CP
             cp_conn_props = {
                 ConnPropKey.FQCN: cp_fqcn,
+                ConnPropKey.IDENTITY: client_name,
+                ConnPropKey.AUTH_IDENTITY: client_auth_identity,
             }
         set_scope_property(client_name, ConnPropKey.CP_CONN_PROPS, cp_conn_props)
```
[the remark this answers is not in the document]
nvidianz: Fixed in `a23b3a9da`: `_determine_conn_props` now publishes `ROOT_CONN_PROPS` for CP clients as well as CJ clients, including the server auth identity and root URL, so `fed_client_base.py` can add the server CN to the identity map. Covered by `tests/unit_test/private/fed/app/fl_conf_test.py::test_cp_conn_props_include_root_auth_identity`.

nvflare/lighter/templates/master_template.yml
```diff
@@ -203,7 +203,9 @@ fed_client: |
       "ssl_root_cert": "rootCA.pem",
       "fqsn": "{~~fqsn~~}",
       "is_leaf": {~~is_leaf~~},
-      "connection_security": "{~~conn_sec~~}"
+      "connection_security": "{~~conn_sec~~}",
```
chesterxgchen: This breaks the backward compatibility, I assume.

nvflare/fuel/f3/sfm/conn_manager.py
```diff
@@ -404,9 +406,22 @@ def update_endpoint(self, sfm_conn: SfmConnection, data: dict):
                 CommError.BAD_DATA, f"Duplicate endpoint name {endpoint_name} for connection {sfm_conn.get_name()}"
             )
 
+        conn_props = sfm_conn.conn.get_conn_properties()
+        # Only enforce on incoming (passive) connections: the authenticated peer CN is
+        # the connecting peer's identity, which is what could spoof a victim's endpoint_name.
+        # The active side that dials out to a known URL has no authenticated peer CN to bind
+        # against for some drivers (e.g. gRPC sets PEER_CN to "N/A"/None on the client side),
+        # and server impersonation there is already covered by TLS CA validation.
+        if sfm_conn.conn.connector.mode == Mode.PASSIVE and is_mtls_connection(sfm_conn.conn.connector.params):
```
pcnudde: This skips endpoint/CN binding for every active mTLS connection, which is broader than the gRPC missing-CN workaround. Some transports can expose a real `PEER_CN` on the active side; in those cases a trusted but wrong cert can complete TLS and then claim `server`, a relay, or a parent FQCN during SFM. The consequence is that the local CellNet routing table can send traffic for that endpoint to the wrong authenticated peer, enabling misrouting/drop/tamper and possible exposure of traffic that is not separately end-to-end encrypted. The later server challenge should still block normal secure client registration, so this is mainly a CellNet routing/relay/ad-hoc concern rather than a full server-token bypass. I’d enforce the resolver check whenever mTLS provides a real `peer_cn`, and only exempt the `None`/`N/A` active-side case that gRPC needs.

pcnudde (commented):
One CellNet identity-binding concern below.

nvflare/lighter/templates/master_template.yml
nvidianz: Addressed in `75030dc11`. The generated startup JSON now omits `auth_identity` and `auth_identity_map` when they are default/empty, so the default `fed_client.json` keeps the old shape with `connection_security` as the final client field. Custom CN configurations still emit the optional identity fields. Added helper coverage and a rendered package assertion for the default client config.

nvflare/fuel/f3/sfm/conn_manager.py
nvidianz: Addressed in `75030dc11`. `ConnManager.update_endpoint()` now enforces endpoint/CN binding for any mTLS connection with a real `PEER_CN`; only active-side missing/`N/A` peer CN is exempt for the gRPC case. Added coverage that an active mTLS connection with the wrong peer CN is rejected.

chesterxgchen:
---
  Security Review — PR #4759: Bind CellNet mTLS endpoint identity

  HIGH

  Finding 1 — Admin FQCN format bypasses CN binding
  identity.py require_match(): any peer can claim _admin_<uuid4> as its endpoint name and skip CN verification entirely. The bypass is necessary for real admin consoles (which have user@domain CNs), but the check is format-only — it does not verify
  the connection arrived on the admin port. A malicious participant with any CA-signed cert can register as _admin_<uuid4> on the CellNet, gaining access to the admin routing path. No test covers a non-admin-port connection claiming an admin FQCN.

  ---
  MEDIUM

  Finding 2 — Trailing-dot FQCN bypasses duplicate endpoint guard
  FQCN.normalize() strips whitespace only, not dots. A peer claiming "server." passes the duplicate-endpoint guard ("server." != "server"), then prefix-matches against prefix_identity_map["server"] and resolves to the server's expected CN. A
  participant holding a cert whose CN matches the server's identity could register a distinct "server." endpoint. Fix: call FQCN.validate() on endpoint_name at conn_manager.update_endpoint entry before identity enforcement.

  Finding 3 — CredentialManager.process_request error does not close the connection
  origin in cert-exchange messages is attacker-controlled (from ORIGIN message header). The CN mismatch raises RuntimeError, which is correct — but CertificateExchanger.process_request should close the connection on this error. If it only sends an
  error response, the connection stays live with no cached cert in a potentially inconsistent state.

  Finding 4 — Active-side (client) enforcement skipped for gRPC
  conn_manager.update_endpoint enforcement is passive-only (Mode.PASSIVE). For the socket/TCP driver, PEER_CN is available on the active side (getpeercert()) and could be enforced. The gap is gRPC-specific, but means a rogue server with a valid
  CA-signed cert can present any FQCN to clients without CN verification.

  Finding 5 — auth_identity read once at startup; no guard against cert rotation divergence
  _get_auth_identity_from_credentials reads the cert CN from disk once at Cell init. If the cert file is rotated, the in-memory resolver diverges silently.

  Finding 6 — auth_identity scalar in template not JSON-escaped
  In static_file.py, the auth_identity_map value goes through json.dumps (safe), but the scalar {~~auth_identity~~} substitution does not. A user-supplied auth_identity value containing ", \, or a newline in project.yml will produce broken JSON in the
  generated fed_client.json / fed_server.json.

  ---
  LOW

  Finding 7 — Fallback from AUTH_IDENTITY → identity is semantically wrong
  fl_conf.py falls back to the identity field when auth_identity is absent. The identity field is a legacy display name, not a security binding. A partially-upgraded kit with a wrong identity will produce a silent wrong expected-CN rather than a clear
  "missing auth_identity" error.

  Finding 8 — Compromised site cert grants access to all its job cells
  A server job cell gets the full auth_identity_map. If a site cert is compromised, the attacker can authenticate to all that site's job cells with the same cert. This is inherent to the single-cert-per-site design but warrants a doc note.

  ---
  INFO

  Finding 9 — No test for spoofed ORIGIN header in cert exchange
  A test should confirm that _cache_cert rejects a message where ORIGIN names a different site than the connection's validated SFM endpoint.

  Finding 10 — Provisioning and runtime resolver logic duplicated
  static_file.py _resolve_default_auth_identity re-implements the "first FQCN segment" logic from CellIdentityResolver.resolve(). These should share an implementation to avoid silent divergence.

  ---
  Summary

  The core fix (binding SFM endpoint names to mTLS CNs) is sound and the tests cover the main enforcement path. The most significant gap is Finding 1 (admin FQCN bypass) — the scope of the bypass is broader than necessary. Finding 2 (trailing-dot) and
  Finding 6 (unescaped scalar in template) are concrete exploitable gaps worth fixing before merge. The active-side enforcement gap (Finding 4) is a known gRPC limitation but should be documented.

nvidianz:
@chesterxgchen thanks for the security review. I pushed a5d6a96a0 with follow-up fixes for the actionable gaps:

- Finding 1: the server now tags admin-capable listeners with `admin_listener=true`, and SFM rejects `_admin_<uuid>` endpoint claims on passive mTLS connections unless the connection arrived through an admin listener. Added reject/accept tests for non-admin vs admin listeners.
- Finding 2: SFM now normalizes and validates the claimed `endpoint_name` with `FQCN.validate()` before duplicate checks and identity binding. Added a `server.` regression test.
- Finding 3 / Finding 9: spoofed cert-exchange `ORIGIN` requests still fail before caching the certificate, and CoreCell now closes the connection after sending the failed credential-manager reply. Added tests for spoofed ORIGIN and close-after-failed-cert-exchange.

Already covered / notes:
- Finding 4: active-side enforcement is in place when the driver exposes a real `PEER_CN`; missing/`N/A` is still skipped for gRPC active-side connections. Existing tests cover both cases.
- Finding 6: the prior update JSON-escapes both scalar `auth_identity` and map values in the provisioning template path.
- Findings 5 and 8 are existing operational model constraints (cert rotation requires process restart; site certs intentionally represent the site and its job cells). Happy to add docs in a follow-up if you want that in this PR.
- Findings 7 and 10 are compatibility/cleanup items; I kept the legacy fallback and provisioning resolver behavior unchanged to avoid breaking mixed-version kits in this security fix.

Validation:
- `/Users/zhihongz/nvflare-env/3.12/bin/python -m pytest tests/unit_test/fuel/f3/cellnet/identity_binding_test.py -v` (23 passed)
- `/Users/zhihongz/nvflare-env/3.12/bin/python -m pytest tests/unit_test/fuel/f3/cellnet -q` (58 passed)
- changed-file `black --check`, `isort --check-only`, and `flake8` passed
- `PATH=/Users/zhihongz/nvflare-env/3.12/bin:$PATH ./runtest.sh --skip-install -s` still fails only on unrelated vendored `examples/.../analytics-dashboard/node_modules/.../flatted.py` files

nvflare/private/fed/server/fed_server.py
```diff
@@ -176,18 +177,23 @@ def deploy(self, args, grpc_args=None, secure_train=False):
         # get admin port
         admin_port = int(grpc_args.get("admin_port", fl_port))
 
-        root_url = [f"{scheme}://0:{fl_port}"]
+        admin_url = f"{scheme}://0:{admin_port}?{ADMIN_LISTENER_KEY}=true"
```
chesterxgchen, line 180: should "0" be configurable in case security concern ?

nvidianz: Due to complicated network setup with multiple interfaces/VPN, NVFlare always listen to 0 (every interface). This deserves another PR if we want to change this behavior.

chesterxgchen (dismissed):

chesterxgchen (approved):

PR #4781, opened 2026-06-08
Fix simulator job cell without server config

nvidianz:
## Summary
- Default missing `server_config` to an empty dict in `FederatedServer.create_job_cell` so simulator job-cell creation can run with non-secure training.
- Propagate simulator subprocess return codes through `FedJobConfig.simulator_run()`, `FedJob.simulator_run()`, and `SimEnv.deploy()`.
- Raise from `SimEnv.deploy()` on nonzero simulator status so `python job.py` exits nonzero when the simulator fails.
- Improve the simulation failure message to point reviewers/users at the per-site log layout, for example `server/simulate_job/log.txt`.
- Add regression coverage for the simulator-style `server_config=None` call, auth identity propagation, simulator return-code propagation, and the direct subprocess return-code layer.

## Root Cause
`SimulatorRunner.start_server_app` calls `create_job_cell(..., False, None)`. `create_job_cell` only used `server_config` inside the secure credential block before later unconditionally reading auth identity fields with `server_config.get(...)`, causing an `AttributeError` for non-secure simulator jobs.

The simulator process already exits with its run status, but the recipe path discarded that status in the subprocess wrapper and `FedJob.simulator_run()`. As a result, `python job.py` could report the thread failure in logs while the top-level process still exited `0`.

## Validation
- Reproduced the original `hello-pt` failure on `upstream/main` at `26a072e3c`: server thread hit `AttributeError: NoneType object has no attribute get`, while `python job.py` exited `0`.
- `/Users/zhihongz/nvflare-env/3.12/bin/python -m pytest tests/unit_test/private/fed/server/fed_server_test.py tests/unit_test/private/fed/app/simulator/simulator_runner_test.py tests/unit_test/recipe/sim_env_test.py tests/unit_test/job_config/fed_job_test.py::TestSimulatorRunClientValidation -q`
- `/Users/zhihongz/nvflare-env/3.12/bin/python -m pytest tests/unit_test/job_config/fed_job_config_test.py tests/unit_test/recipe/sim_env_test.py -q`
- `PYTHONPATH=/Users/zhihongz/work/NVFlare /Users/zhihongz/nvflare-env/3.12/bin/python job.py --synthetic_data --train_size 16 --test_size 8 --num_rounds 1 --epochs 1 --batch_size 4 --num_workers 0` from `examples/hello-world/hello-pt`
- Scoped `black --check`, `isort --check-only`, `flake8`, and `git diff --check` on changed files

## Notes
- `PATH=/Users/zhihongz/nvflare-env/3.12/bin:$PATH ./runtest.sh -s` failed before style checks because dependency setup selected Python 3.8 and `versioneer.py` uses `dict[str, str]` syntax.
- `PATH=/Users/zhihongz/nvflare-env/3.12/bin:$PATH ./runtest.sh --skip-install -s` reached Black and failed only on pre-existing vendored `examples/.../node_modules/.../flatted.py` files outside this PR.

### DIFF
--- a/nvflare/recipe/sim_env.py
+++ b/nvflare/recipe/sim_env.py
@@ -94,7 +94,7 @@ class SimEnv(ExecEnv):
         self.log_config = v.log_config
         self.clients = v.clients
         self.workspace_root = v.workspace_root
-        self.last_run_status = None
+        self.last_run_failed = False
 
     def deploy(self, job: FedJob):
         # Validate scripts exist locally for simulation
@@ -106,7 +106,7 @@ class SimEnv(ExecEnv):
             )
 
         job_id = job.name
-        self.last_run_status = job.simulator_run(
+        run_status = job.simulator_run(
             workspace=os.path.join(self.workspace_root, job.name),
             n_clients=self.num_clients if self.clients is None else None,
             clients=self.clients,
@@ -114,9 +114,10 @@ class SimEnv(ExecEnv):
             gpu=self.gpu_config,
             log_config=self.log_config,
         )
-        if self.last_run_status:
+        self.last_run_failed = run_status not in (None, 0)
+        if self.last_run_failed:
             raise RuntimeError(
-                f"Simulation failed with return code {self.last_run_status}. "
+                f"Simulation failed with return code {run_status}. "
                 f"Logs are in per-site subdirectories under {os.path.join(self.workspace_root, job_id)}, "
                 f"e.g. server/simulate_job/log.txt"
             )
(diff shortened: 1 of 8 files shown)


nvflare/recipe/sim_env.py
```diff
@@ -104,15 +105,22 @@ def deploy(self, job: FedJob):
                 f"For SimEnv, all scripts must be present on the local machine."
             )
 
-        job.simulator_run(
+        job_id = job.name
+        self.last_run_status = job.simulator_run(
```
IsaacYangSLA: Suggest rename this to last_run_failed (or something for failure) so that the if statement on line 117 will be easier to understand.

IsaacYangSLA (approved):

PR #4793, opened 2026-06-10
Add recipe metadata helper

IsaacYangSLA:
## Summary
- add `set_recipe_meta(recipe, key, value)` and export it from `nvflare.recipe`
- require the second argument to be a `JobMetaKey` enum member; raw strings are rejected
- accept number, string, dictionary, or list values, including list-backed metadata such as `JobMetaKey.MANDATORY_CLIENTS`
- write helper-provided metadata through `meta_props`; same-key `meta_props` values are replaced without conflict checks
- reject `JobMetaKey.DEPLOY_MAP` because it is generated dynamically from recipe deployment and must not be snapshotted into `meta_props`
- document recipe metadata helper usage and update agent test-command guidance
- add unit coverage for metadata export, deep-copying, enum/value validation, metadata replacement/override behavior, deploy-map rejection, list-valued metadata, and package export

## Tests
- `source ../.venv/bin/activate && python3 -m pytest tests/unit_test/recipe/spec_test.py tests/unit_test/recipe/utils_test.py -q`
- `source ../.venv/bin/activate && ./runtest.sh -s --skip-install`
- `source ../.venv/bin/activate && ./runtest.sh -u --skip-install`

### DIFF
--- a/nvflare/recipe/utils.py
+++ b/nvflare/recipe/utils.py
@@ -15,9 +15,11 @@
 import copy
 import importlib
 import os
+from numbers import Number
 from typing import Any, Dict, List, Optional
 
 from nvflare.apis.analytix import ANALYTIC_EVENT_TYPE
+from nvflare.apis.job_def import JobMetaKey
 from nvflare.fuel.utils.import_utils import optional_import
 from nvflare.job_config.api import FedJob
 from nvflare.recipe.spec import Recipe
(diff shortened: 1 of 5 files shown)


nvflare/recipe/utils.py
```diff
@@ -58,6 +60,50 @@
     },
 }
 
+_RECIPE_META_UNSUPPORTED_KEYS = {
+    JobMetaKey.DEPLOY_MAP,
+}
```
YuanTingHsieh, line 65: since we have a lot of keys inside JobMetaKey, and some of them are system-managed (`STATUS,SUBMIT_TIME,SUBMIT_TIME_ISO,START_TIME`) 

maybe we should instead use an allowlist?

something like:

```
_RECIPE_USER_SETTABLE_META_KEYS = {
    JobMetaKey.MIN_CLIENTS,
    JobMetaKey.MANDATORY_CLIENTS,
    JobMetaKey.RESOURCE_SPEC,
    JobMetaKey.JOB_LAUNCHER_SPEC,
    JobMetaKey.SCOPE,
    JobMetaKey.STUDY,
    JobMetaKey.CUSTOM_PROPS,
}
```

nvflare/recipe/utils.py
```diff
@@ -58,6 +60,50 @@
     },
 }
 
+_RECIPE_META_UNSUPPORTED_KEYS = {
```
holgerroth, line 63: Can we make this a positive allowlist instead of blocking only `DEPLOY_MAP`? `JobMetaKey` includes runtime/system-managed fields such as `STATUS`, `SUBMIT_TIME`, `SCHEDULE_COUNT`, `LAST_SCHEDULE_TIME`, and `SCHEDULE_HISTORY`. Since `FedJobConfig._generate_meta()` applies `meta_props` last, this helper currently lets recipes stamp those fields into exported `meta.json`; fields like `schedule_count`/`last_schedule_time` are later read by the scheduler as existing state and could delay or block scheduling. I’d like to see an explicit `_RECIPE_USER_SETTABLE_META_KEYS` set for the intended recipe metadata keys (`MIN_CLIENTS`, `MANDATORY_CLIENTS`, `RESOURCE_SPEC`, `JOB_LAUNCHER_SPEC`, `SCOPE`, `STUDY`, `CUSTOM_PROPS`) plus a negative unit test for a system-managed key.

holgerroth (commented):
One issue I’d like fixed before merge.

YuanTingHsieh:
#4849 build on top of this PR so we can safely close this

PR #4794, opened 2026-06-10
Add tensor disk offload support for FedOpt

pcnudde:
## What changed

- Added `enable_tensor_disk_offload` support to the PyTorch FedOpt recipe and wired it into `ScatterAndGather`.
- Shared tensor disk offload setup and cleanup between FedAvg and ScatterAndGather through common helpers.
- Preserved PyTorch tensor weights in the FedOpt server update path when the global model is tensor-backed, avoiding an intermediate NumPy conversion that would defeat disk offload.
- Added warning coverage for enabling tensor disk offload without `ExchangeFormat.PYTORCH`.
- Documented FedProx and FedOpt disk offload behavior and constraints.

## Why

FedAvg already supports streamed PyTorch tensor disk offload. FedOpt can reuse the same FOBS-context mechanism, but it also needs to keep the server-side FedOpt output in tensor form when the incoming/global model representation is PyTorch-backed.


nvflare/app_opt/pt/fedopt.py
```diff
@@ -212,29 +212,46 @@ def shareable_to_learnable(self, shareable: Shareable, fl_ctx: FLContext) -> Lea
...
         model_diff = dxo.data
+        base_model = fl_ctx.get_prop(AppConstants.GLOBAL_MODEL)
+        if not base_model:
+            self.system_panic(reason="No global base model!", fl_ctx=fl_ctx)
+            return base_model
+
+        base_model_weights = base_model[ModelLearnableKey.WEIGHTS]
+        preserve_torch_weights = any(isinstance(value, torch.Tensor) for value in base_model_weights.values())
+        if not base_model_weights:
+            preserve_torch_weights = any(isinstance(value, torch.Tensor) for value in model_diff.values())
```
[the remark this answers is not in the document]
pcnudde: Fixed in 84afb18ed by making the intent explicit: base_model_weights now selects the base-model tensor check only when non-empty, and otherwise falls back to model_diff.

nvflare/app_common/utils/tensor_disk_offload_context.py
```diff
@@ -49,25 +63,78 @@ def apply_enable_tensor_disk_offload(
...
+    except Exception:
+        if root_dir:
+            shutil.rmtree(root_dir, ignore_errors=True)
+        raise
+    return TensorDiskOffloadContext(
+        previous_value=previous_value,
+        previous_root_dir=previous_root_dir,
+        root_dir=root_dir,
+        applied=applied,
+    )
```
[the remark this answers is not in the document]
pcnudde: Fixed in 84afb18ed. setup_tensor_disk_offload now checks for an active cell before creating the offload temp directory and returns a no-op context when no cell is available. Added a regression test for that path.

nvflare/app_common/utils/tensor_disk_offload_context.py
```diff
@@ -28,14 +49,7 @@ def apply_enable_tensor_disk_offload(
...
     """
-    if not engine:
-        return None, False
-
-    run_manager = getattr(engine, "run_manager", None)
-    if run_manager and run_manager.cell:
-        cell = run_manager.cell
-    else:
-        cell = engine.get_cell()
+    cell = _get_cell(engine)
```
chesterxgchen: dont need check against engine any more

chesterxgchen (commented):
add a question, maybe address the greptile comment in summary

nvflare/app_common/utils/tensor_disk_offload_context.py
pcnudde: why? did we change something that makes this not necessary?

pcnudde: I see. I'll simplify the code

pcnudde: You're right that the check was redundant on this path since setup_tensor_disk_offload already resolves the cell first. Went one step further in ac6f5786f: removed apply_enable_tensor_disk_offload/restore_enable_tensor_disk_offload entirely and folded them into setup/cleanup, so the cell is resolved once on setup and cleanup keys off the explicit applied flag. Tests updated accordingly.

pcnudde:
Re the Greptile summary concern (KeyError in the FedAvg buffer-update loop on the first round with empty global weights): that indexing of base_model_weights is pre-existing on main — this PR only added a tensor-preserving variant of the same access. Hardened it anyway in e0dd288e8: the loop now falls back to the server model's own state value when a key is missing from base_model_weights, with regression tests for both the tensor and numpy paths.

chesterxgchen (approved):

PR #4803, opened 2026-06-12
fix(analytix): correct typo in convert_data_type that silently drops TORCH_TB→WANDB metrics

### LINKED ISSUE #4802: [BUG] convert_data_type silently drops all metrics when routing TensorBoard writer data to WandBReceiver
## Bug Description

In \`nvflare/apis/analytix.py\`, the \`AnalyticsData.convert_data_type\` class method contains a typo on line 206 that causes **all analytics metrics to be silently dropped** when a TensorBoard writer (\`SummaryWriter\` / \`LogWriterName.TORCH_TB\`) is used on the client side while \`WandBReceiver\` (\`LogWriterName.WANDB\`) is used on the server side.

## Affected Code

\`\`\`python
# nvflare/apis/analytix.py, line 206
if sender == LogWriterName.TORCH_TB and (receiver == LogWriterName.MLFLOW or sender == LogWriterName.WANDB):
#                                                                              ^^^^^^ BUG — should be `receiver`
\`\`\`

Because \`sender\` is already known to be \`TORCH_TB\` in this branch, the sub-condition \`sender == LogWriterName.WANDB\` is always \`False\`. The effective condition is therefore:

\`\`\`python
if sender == LogWriterName.TORCH_TB and receiver == LogWriterName.MLFLOW:   # only MLFLOW!
\`\`\`

## Impact

When \`convert_data_type(data_type, LogWriterName.TORCH_TB, LogWriterName.WANDB)\` is called:
1. None of the three \`if\` branches match.
2. The method returns \`None\` implicitly.
3. \`from_dxo\` receives \`data_type = None\` → \`if not data_type: return None\`.
4. \`WandBReceiver.save\` receives \`data = None\` → \`if not data: return\`.

**All metrics sent from a \`SummaryWriter\` (or any \`TORCH_TB\`-labelled writer) are silently discarded by \`WandBReceiver\`.**

## Steps to Reproduce

\`\`\`python
from nvflare.apis.analytix import AnalyticsData, AnalyticsDataType, LogWriterName

result = AnalyticsData.convert_data_type(
    AnalyticsDataType.SCALAR, LogWriterName.TORCH_TB, LogWriterName.WANDB
)
print(result)  # None  — should be AnalyticsDataType.METRIC
\`\`\`

## Expected Behavior

\`\`\`
AnalyticsDataType.METRIC
\`\`\`

## Fix

Change line 206 from:
\`\`\`python
if sender == LogWriterName.TORCH_TB and (receiver == LogWriterName.MLFLOW or sender == LogWriterName.WANDB):
\`\`\`
to:
\`\`\`python
if sender == LogWriterName.TORCH_TB and (receiver == LogWriterName.MLFLOW or receiver == LogWriterName.WANDB):
\`\`\`

AlexanderSanin:
## Summary

- Fixes a one-character typo in `AnalyticsData.convert_data_type` (`nvflare/apis/analytix.py` line 206) where `sender == LogWriterName.WANDB` was written instead of `receiver == LogWriterName.WANDB`.
- Because `sender` is already constrained to `TORCH_TB` by the outer condition, `sender == WANDB` was always `False`, causing the method to fall through all branches and return `None` implicitly for the `TORCH_TB → WANDB` path.
- This made `from_dxo` return `None` and `WandBReceiver.save` return early — **silently dropping every metric** logged via a `SummaryWriter` when a `WandBReceiver` was configured on the server side.
- Adds 11 unit tests covering all three conversion directions and a regression test for `from_dxo` with a WANDB receiver.

Closes #4802

## Root Cause

```python
# Before (line 206) — sender == WANDB is always False here
if sender == LogWriterName.TORCH_TB and (receiver == LogWriterName.MLFLOW or sender == LogWriterName.WANDB):

# After
if sender == LogWriterName.TORCH_TB and (receiver == LogWriterName.MLFLOW or receiver == LogWriterName.WANDB):
```

## Test plan

- [x] `python3 -m pytest tests/unit_test/apis/analytix_test.py -v` — all 21 tests pass (10 pre-existing + 11 new)
- [x] Verified `AnalyticsData.convert_data_type(SCALAR, TORCH_TB, WANDB)` now returns `METRIC` instead of `None`
- [x] Verified `AnalyticsData.from_dxo(dxo, receiver=LogWriterName.WANDB)` no longer returns `None` for TORCH_TB-originated DXOs

### DIFF
--- a/nvflare/apis/analytix.py
+++ b/nvflare/apis/analytix.py
@@ -203,6 +203,7 @@ class AnalyticsData:
         cls, sender_data_type: AnalyticsDataType, sender: LogWriterName, receiver: LogWriterName
     ) -> AnalyticsDataType:
 
+        # TensorBoard naming → MLflow/W&B naming
         if sender == LogWriterName.TORCH_TB and (receiver == LogWriterName.MLFLOW or receiver == LogWriterName.WANDB):
             if AnalyticsDataType.SCALAR == sender_data_type:
                 return AnalyticsDataType.METRIC
@@ -211,7 +212,8 @@ class AnalyticsData:
             else:
                 return sender_data_type
 
-        if sender == LogWriterName.MLFLOW and receiver == LogWriterName.TORCH_TB:
+        # MLflow/W&B naming → TensorBoard naming
+        if (sender == LogWriterName.MLFLOW or sender == LogWriterName.WANDB) and receiver == LogWriterName.TORCH_TB:
             if AnalyticsDataType.METRIC == sender_data_type:
                 return AnalyticsDataType.SCALAR
             elif AnalyticsDataType.METRICS == sender_data_type:
@@ -219,8 +221,14 @@ class AnalyticsData:
             else:
                 return sender_data_type
 
-        if sender == LogWriterName.MLFLOW and receiver == LogWriterName.WANDB:
+        # MLflow and W&B share the same METRIC/METRICS naming, so cross-mapping is a pass-through.
+        if (sender == LogWriterName.MLFLOW and receiver == LogWriterName.WANDB) or (
+            sender == LogWriterName.WANDB and receiver == LogWriterName.MLFLOW
+        ):
             return sender_data_type
+
+        # Same sender/receiver, or any combination not covered above: pass through unchanged.
+        return sender_data_type
 
     def __str__(self) -> str:
         return f"AnalyticsData(tag: {self.tag}, value: {self.value}, data_type: {self.data_type}, kwargs: {self.kwargs}, step: {self.step})"
--- a/nvflare/apis/analytix.py
+++ b/nvflare/apis/analytix.py
@@ -203,7 +203,7 @@ class AnalyticsData:
         cls, sender_data_type: AnalyticsDataType, sender: LogWriterName, receiver: LogWriterName
     ) -> AnalyticsDataType:
 
-        if sender == LogWriterName.TORCH_TB and (receiver == LogWriterName.MLFLOW or sender == LogWriterName.WANDB):
+        if sender == LogWriterName.TORCH_TB and (receiver == LogWriterName.MLFLOW or receiver == LogWriterName.WANDB):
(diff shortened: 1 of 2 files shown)


AlexanderSanin:
Hey @chesterxgchen @nvidianz @pcnudde. Could you, please, have a look at this?

chesterxgchen:
Thanks for spotting this and commit the PR for fix. Make sure to get the greptile review score 5/5 and have your commit signed. thanks @AlexanderSanin 

Confidence Score: 4/5
Safe to merge — the change is a single-character typo fix in a well-tested conversion utility with no side effects on other paths.

The fix is minimal and clearly correct. The only advisory note is that convert_data_type still implicitly returns None (violating its -> AnalyticsDataType annotation) for WANDB → TORCH_TB and WANDB → MLFLOW pairs, which would silently drop events on those paths — the same pattern this PR repairs for TORCH_TB → WANDB. This is pre-existing and not introduced here, but worth addressing before WANDB sender paths are exercised in production.

AlexanderSanin:
Thanks for the review, @chesterxgchen! I pushed b4302a2b to address the Greptile advisory note:

- `convert_data_type` now handles `WANDB → TORCH_TB` (mirroring `MLFLOW → TORCH_TB`) and `WANDB → MLFLOW` (pass-through, symmetric with `MLFLOW → WANDB`).
- Added a final `return sender_data_type` fallback so the method always returns a concrete `AnalyticsDataType`, honoring its return annotation and eliminating the silent-drop class of bug entirely.
- Added `test_convert_data_type_never_returns_none` which exhaustively iterates every `(LogWriterName, LogWriterName, AnalyticsDataType)` triple to lock in the invariant.

All 27 tests pass locally (`pytest tests/unit_test/apis/analytix_test.py`).

On signing: my last commit was already authored by `Claude <noreply@anthropic.com>`, which GitHub auto-verifies via the Anthropic vendor key. The new commit uses the same identity, so both should show as Verified.

chesterxgchen (approved):
LGTM

PR #4804, opened 2026-06-12
Support study-specific K8s job pod templates

IsaacYangSLA:
## Summary

This PR adds study-specific Kubernetes job pod template support to the K8s job launcher. The current `main` branch behavior remains the default behavior: when users do not configure a study job pod spec file, NVFlare launches job pods the same way it does today.

The new `study_job_spec_file_path` launcher argument enables an additional and more complete control point for job pod construction. It lets deployment owners define per-study Pod templates while keeping the generated NVFlare job command, workspace mounts, startup kit mounts, and existing study-data PVC behavior under launcher control.

## Three Supported Cases

1. `study_data_pvc_file_path` only: this is the current default behavior from `nvflare deploy prepare` for K8s. Study-data PVC mappings from `study_data.yaml` are applied, and job pods otherwise use the launcher-generated pod spec.
2. `study_job_spec_file_path` only: the launcher reads a study-to-Pod-template mapping file and applies the matching Pod template for the job study. No study-data PVC mounts are added unless configured elsewhere.
3. Both `study_data_pvc_file_path` and `study_job_spec_file_path`: the launcher applies both controls, using the study-data PVC mapping and the study-specific Pod template together.

## H100 GPU Node Selection Use Case

A primary use case is GPU scheduling. Users can place node selection, tolerations, affinity, runtime class, image pull secrets, or other Kubernetes Pod-level settings in a study-specific template, then map only the relevant study to that template. This is useful when an H100 study must run on H100 GPU nodes while other studies continue using ordinary CPU scheduling or different GPU pools.

The exact H100 node label must match the target cluster. For example, if H100 nodes are labeled with `nvidia.com/gpu.product: NVIDIA-H100-80GB-HBM3`, users can target those nodes with a study-specific Pod template like this.

Example `local/study_job_spec.yaml`:

```yaml
h100_study: pod_specs/h100-job-pod.yaml
```

Example `local/pod_specs/h100-job-pod.yaml`:

```yaml
apiVersion: v1
kind: Pod
metadata:
  labels:
    nvflare-study: h100-study
spec:
  nodeSelector:
    nvidia.com/gpu.product: NVIDIA-H100-80GB-HBM3
  tolerations:
    - key: nvidia.com/gpu
      operator: Exists
      effect: NoSchedule
  containers:
    - name: nvflare_job
      resources:
        limits:
          nvidia.com/gpu: "1"
```

After `nvflare deploy prepare`, users can edit `local/resources.json.default` to add:

```json
"study_job_spec_file_path": "/var/tmp/nvflare/workspace/local/study_job_spec.yaml"
```

They can also change or remove the generated `study_data_pvc_file_path` depending on whether study-data PVC mounts are needed.

## GPU Labeling References and Reviewer Pointers

This PR also updates the K8s deployment guide so users have enough information to build a useful job pod template:

- [`Cloud GPU Setup References`](https://github.com/IsaacYangSLA/NVFlare/blob/job_pod_spec/docs/user_guide/admin_guide/deployment/helm_chart.rst#cloud-gpu-setup-references) points users to provider-specific GPU setup docs and the NVIDIA GPU Operator docs.
- [`Minimal Study Job Pod Template`](https://github.com/IsaacYangSLA/NVFlare/blob/job_pod_spec/docs/user_guide/admin_guide/deployment/helm_chart.rst#minimal-study-job-pod-template) shows the H100 `nodeSelector`, `kubectl get nodes -L nvidia.com/gpu.product,nvidia.com/gpu.count,nvidia.com/gpu.present`, an `affinity.nodeAffinity` alternative, and the distinction between scheduler placement fields in the template and launcher-owned GPU resource requests.

Reference docs for the GPU node-labeling behavior:

- NVIDIA k8s-device-plugin / GPU Feature Discovery label catalog: https://github.com/NVIDIA/k8s-device-plugin#catalog-of-labels
- NVIDIA GPU Operator docs: https://docs.nvidia.com/datacenter/cloud-native/gpu-operator/latest/
- Kubernetes `nodeSelector` and node affinity docs: https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/

## Validation

- Verified the three cases in OpenShift CRC: `study_data_pvc_file_path` only, `study_job_spec_file_path` only, and both together.
- Verified job pods completed successfully for server, `site-1`, and `site-2` in each case.
- Ran `./runtest.sh -s` successfully.

### DIFF
--- a/docs/design/job_launcher_and_job_handle.md
+++ b/docs/design/job_launcher_and_job_handle.md
@@ -374,13 +374,14 @@ Constructor parameters:
 | Parameter | Default | Purpose |
 |-----------|---------|---------|
 | `config_file_path` | required | Path to kubeconfig. Loaded lazily on first `launch_job`. |
-| `workspace_pvc` | required | PVC claim name for workspace volume. |
-| `study_data_pvc_file_path` | required | YAML file mapping study/dataset names to PVC claim names. Validated lazily; missing study entries skip data PVC mounts and log a warning. |
+| `study_data_pvc_file_path` | `None` | Optional YAML file mapping study/dataset names to PVC claim names. Validated lazily; missing study entries skip data PVC mounts and log a warning. When this is the only study-specific setting, the launcher uses the built-in pod manifest behavior. |
+| `study_job_spec_file_path` | `None` | Optional YAML file mapping study names to Kubernetes Pod YAML templates. Matching studies use the template with NVFlare-owned job fields overlaid; unmatched studies use the built-in manifest. When this is set without `study_data_pvc_file_path`, no study-data PVC mounts are added. |
 | `timeout` | `None` | Wall-clock seconds for `enter_states([RUNNING])`. |
 | `namespace` | `"default"` | Kubernetes namespace. |
 | `pending_timeout` | `120` | Seconds to wait while the scheduler reports insufficient CPU, memory, or GPU resources. Use `0` to fail fast or `None` to wait indefinitely unless `timeout` is set. Job meta can override with `launcher_spec[site][k8s].pending_timeout`. |
 | `default_python_path` | `"/usr/local/bin/python3"` | Default Python executable in the pod command. Job meta can override with `launcher_spec[site][k8s].python_path`. |
 | `workspace_mount_path` | `"/var/tmp/nvflare/workspace"` | In-container path where job pods mount the transferred job workspace and startup kit. `nvflare deploy prepare` sets this from `parent.workspace_mount_path`. |
+| `image_pull_secrets` | `None` | Optional list of existing Kubernetes Secret names attached to launched job pods as `imagePullSecrets`. |
 | `ephemeral_storage` | `"1Gi"` | Default job pod workspace `emptyDir` size and `ephemeral-storage` request/limit. Job meta can override with `launcher_spec[site][k8s].ephemeral_storage`. |
 
 Launch sequence:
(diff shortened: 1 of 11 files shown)


chesterxgchen (commented):
from codex

• Findings

  - Medium: Pod templates can preserve a conflicting metadata.namespace. The overlay only replaces metadata.name in /private/
    tmp/nvflare-pr4804-review/nvflare/app_opt/job_launcher/k8s_launcher.py:493, then submits with create_namespaced_pod(...,
    namespace=self.namespace) at /private/tmp/nvflare-pr4804-review/nvflare/app_opt/job_launcher/k8s_launcher.py:1168. If a
    study template is copied from a real Pod YAML with metadata.namespace: gpu-jobs, launch can fail due to namespace mismatch
    or RBAC, since the generated Helm chart grants only namespaced Role permissions. I’d make namespace launcher-owned too:
    remove it from templates or set it to self.namespace, and add a regression test.

  Assumption

  - I’m assuming study job pods are intended to run in the same namespace as the parent launcher, matching the current
    namespace config and generated Role/RoleBinding.

IsaacYangSLA:
> from codex
> 
> • Findings
> 
> * Medium: Pod templates can preserve a conflicting metadata.namespace. The overlay only replaces metadata.name in /private/
>   tmp/nvflare-pr4804-review/nvflare/app_opt/job_launcher/k8s_launcher.py:493, then submits with create_namespaced_pod(...,
>   namespace=self.namespace) at /private/tmp/nvflare-pr4804-review/nvflare/app_opt/job_launcher/k8s_launcher.py:1168. If a
>   study template is copied from a real Pod YAML with metadata.namespace: gpu-jobs, launch can fail due to namespace mismatch
>   or RBAC, since the generated Helm chart grants only namespaced Role permissions. I’d make namespace launcher-owned too:
>   remove it from templates or set it to self.namespace, and add a regression test.
> 
> Assumption
> 
> * I’m assuming study job pods are intended to run in the same namespace as the parent launcher, matching the current
>   namespace config and generated Role/RoleBinding.

Will replace metadata.namespace with self.namespace.

chesterxgchen:
claude findings: 

1. Medium — _select_job_container can pick the wrong container

k8s_launcher.py, _select_job_container:
for container in containers:
    if container.get("name") in (container_name, "nvflare_job"):
        return container
return containers[0]
It's always called from __init__ with job_config.get("container_name") → None (no default). So the match set is (None, "nvflare_job"), and any container lacking a name key matches as if it were the job container.

Verified empirically — a template with an unnamed container listed before the named one:
containers = [{"image":"sidecar"}, {"name":"nvflare_job", "resources":{...gpu...}}]
_select_job_container(containers, None)  # -> {'image': 'sidecar'}  ← WRONG
The launcher then injects the job image/command/args into the sidecar and the real nvflare_job container (with the H100 GPU resources) is left as a passive sidecar. This silently mis-schedules the job and is hard to diagnose.

Fix — two-pass, prefer an explicit name match before falling back:
def _select_job_container(containers, container_name):
    targets = {name for name in (container_name, "nvflare_job") if name}
    for c in containers:
        if c.get("name") in targets:
            return c
    return containers[0]
This keeps the documented nvflare_job convention working and falls back to containers[0] only when there's no named match. Worth a regression test for the unnamed-sidecar-first case.

2. Low / informational — pod-spec exclusion gap for hand-edited configs

workspace_cell_transfer.py: pod-spec files referenced by study_job_spec.yaml are excluded from the downloaded workspace bundle only when study_job_spec_file_path resolves to a workspace-relative path. If a user hand-edits resources.json to set study_job_spec_file_path to an absolute container path and omits workspace_mount_path, _config_path_to_workspace_rel returns None, so _add_study_job_spec_excludes is never called and the referenced pod-spec YAMLs get bundled into job pod workspaces.

This does not affect the normal nvflare deploy prepare path — _prepare_k8s always writes workspace_mount_path into the launcher args (deploy_commands.py:488), so resolution succeeds. The mapping file itself is also always covered by the default-name exclude. So this is an edge case for manual configs only — optionally worth a debug log when an absolute path can't be resolved to a workspace-relative one.

chesterxgchen (approved):

PR #4807, opened 2026-06-14
Fix: Handle missing resource module on Windows

mrityunjay2627:
## Problem
`nvflare.fuel.f3.cellnet.net_agent` imports the `resource` module at the top level.
This module is Unix-only and raises `ModuleNotFoundError` on Windows, making it
impossible to `import nvflare` at all.

## Root Cause
`resource` is used only in `_do_process_info()` to report memory usage via
`resource.getrusage()`. The import is unconditional despite only one method needing it.

## Fix
- Wrap `import resource` in `try/except ImportError`, setting `resource = None` when unavailable
- Guard the single usage in `_do_process_info()` to return `"N/A"` when `resource` is `None`

## Testing
- Windows 11: `from nvflare.fuel.f3.cellnet.net_agent import NetAgent` — passes after fix
- Linux (WSL2 Ubuntu): no behavior change, `resource` imports normally

### DIFF
--- a/nvflare/fuel/f3/cellnet/net_agent.py
+++ b/nvflare/fuel/f3/cellnet/net_agent.py
@@ -16,11 +16,15 @@ import copy
 import hashlib
 import os
 import random
-import resource
 import threading
 import time
 from abc import ABC
 from typing import List, Union
+
+try:
+    import resource
+except ImportError:
+    resource = None
 
 from nvflare.fuel.f3.cellnet.cell import Cell
 from nvflare.fuel.f3.cellnet.connector_manager import ConnectorData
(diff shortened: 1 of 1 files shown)


chesterxgchen (approved):

chesterxgchen:
looks like unit-test or code style failed 

./runtests.sh -s for style check 
./runtestslsh to run unit tests locally

chesterxgchen:
@mrityunjay2627  the unit tests maybe is not related to your change,  it was trigged by other changes, it  will go away after my PR https://github.com/NVIDIA/NVFlare/pull/4808 merges

chesterxgchen:
@mrityunjay2627 
Merging is blocked
Commits must have verified signatures.

You need to make sure you set signature for the commits in the future.

mrityunjay2627:
> @mrityunjay2627 Merging is blocked Commits must have verified signatures.
> 
> You need to make sure you set signature for the commits in the future.

I'll make sure to set signature for the commits in the future! Thank you very much for your support!

PR #4808, opened 2026-06-14
Make recipe --export/--export-dir import safe

chesterxgchen:
## Description

Importing `nvflare.recipe.spec` consumes `--export` / `--export-dir` from `sys.argv` at **module import time**, so a recipe job script's own `argparse.parse_args()` never sees them (the strip has to happen at import because job scripts typically call their own `parse_args()` before `recipe.execute()`).

The problem: the consumer **raised `ValueError`** on a dangling `--export-dir`. Import should not blow up because of unrelated command-line tokens it does not own. Today any benign import that happens to have a bare `--export-dir` in `sys.argv` — a unit test introspecting recipe classes, a notebook, the recipe catalog loader, a third-party tool — crashes on import:

```python
sys.argv = ["mytool", "--export-dir"]
import nvflare.recipe.spec     # ValueError: --export-dir requires an argument
```

## Changes

- Make `_consume_recipe_args()` **transactional**: parse into a scratch list and only mutate `sys.argv` when the parse is clean. A dangling `--export-dir` now **abandons the whole pass** — `sys.argv` is left untouched and export stays disabled — instead of raising.
  - This also avoids the subtler failure where a half-consumed `--export --export-dir` would record `export=True` and silently export to the default dir against the user's intent (e.g. under `parse_known_args()`).
- **Freeze the decision** after the first call (`_CONSUMED`) so repeated direct calls return the recorded import-time result rather than re-scanning a since-mutated `sys.argv`.
- Promote the default export dir to a single `DEFAULT_EXPORT_DIR = "./fl_job"` constant (kept relative/CWD: an export is a user-requested, discoverable artifact, and a relative path stays portable across OSes — unlike a fixed `/tmp` path).

Well-formed `--export` / `--export-dir VALUE` / `--export-dir=VALUE` behavior is **unchanged**, with **zero per-recipe argparse wiring**.

## Why not an `argv[0]`-based guard

A "only consume when the process isn't the `nvflare` CLI" heuristic was considered and rejected: it's fragile (breaks under `python -m nvflare.cli`, pytest, subprocess, where `argv[0]` isn't `nvflare`) and unnecessary — the `nvflare` CLI is already safe because its own `parse_args()` runs before `recipe.spec` is ever imported (the CLI imports recipe modules lazily, after parsing).

## Tests

`tests/unit_test/recipe/spec_test.py`:
- replaced the `ValueError`-expecting test with: dangling `--export-dir` does **not** raise on import, leaves `sys.argv` untouched, and does not enable export
- decision is frozen across a later direct call even if `sys.argv` changes
- added `--export-dir=VALUE` equals-form coverage and `_peek_recipe_args()` assertions on the existing strip test

All recipe unit tests pass (`327 passed, 18 skipped`); black/isort/flake8 clean.

---

### Follow-up commit: harden `recipe_cli` import faking (test-isolation fix)

The first version of this PR turned the `unit-tests` matrix red with 6 failures in
`tests/unit_test/tool/recipe_cli_test.py` (`ImportError: nvflare`). Root cause: those tests
faked module loading via `monkeypatch.setattr("nvflare.tool.recipe.recipe_cli.importlib.import_module", ...)`.
Since `recipe_cli.importlib` is the **global** `importlib` module, that replaced
`importlib.import_module` process-wide — including the copy pytest's own `monkeypatch.resolve`
uses to walk dotted string targets. Under certain xdist worker orderings (which this PR's
changed test count reshuffles into) a later test's `monkeypatch.setattr` then failed while
resolving its target. The defect is pre-existing test-isolation fragility; this PR only
surfaced it via scheduling.

Fix: bind `import_module` locally in `recipe_cli.py` (`from importlib import import_module`) and
patch that local name with the object form of `monkeypatch.setattr`. The tests no longer mutate
the global `importlib` and no longer resolve through it, so the catalog tests are hermetic under
any scheduling. `import_module` is the same function — **no production behavior change**.

### DIFF
--- a/nvflare/recipe/spec.py
+++ b/nvflare/recipe/spec.py
@@ -83,8 +83,8 @@ def _consume_recipe_args() -> tuple:
 _RECIPE_EXPORT, _RECIPE_EXPORT_DIR = _consume_recipe_args()
 
 
-def _peek_recipe_args(argv: Optional[List[str]] = None) -> tuple:
-    """Return the export flags consumed at import time (argv argument is ignored)."""
+def _peek_recipe_args() -> tuple:
+    """Return the export flags consumed at import time."""
     return _RECIPE_EXPORT, _RECIPE_EXPORT_DIR
 
 
--- a/nvflare/recipe/spec.py
+++ b/nvflare/recipe/spec.py
@@ -17,6 +17,15 @@ from abc import ABC, abstractmethod
 from contextlib import contextmanager
 from typing import Dict, List, Optional, Union
 
+# Single source of truth for the default export dir. Kept in the current working
+# directory on purpose: an export is a user-requested artifact, so it lands next to
+# where the script ran (discoverable and predictable), and a relative path stays
+# portable across OSes -- unlike a fixed /tmp path, which is non-portable on Windows
+# and collision-prone on shared hosts.
+DEFAULT_EXPORT_DIR = "./fl_job"
+
+_CONSUMED = False
+
 
 def _consume_recipe_args() -> tuple:
     """Strip --export / --export-dir from sys.argv and return (export, export_dir).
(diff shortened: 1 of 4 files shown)


chesterxgchen:
Thanks for the review. Addressed both points in `29adbb1f`:

- **Dead `argv` param on `_peek_recipe_args`** — removed. Every caller invokes it with no arguments and the value was ignored, so the parameter was just misleading.
- **Missing bare-`--export` test** — added `test_recipe_spec_import_bare_export_uses_default_dir`, covering `python job.py --export` (no `--export-dir`): `--export` is stripped and `export_dir` stays at `DEFAULT_EXPORT_DIR`.

The earlier `unit-tests` matrix failures (`recipe_cli_test.py: ImportError: nvflare`) were a pre-existing test-isolation issue this PR surfaced via xdist scheduling, fixed in `1f87d60b` — see the "Follow-up commit" section in the PR description.

PR #4820, opened 2026-06-17
Compute BYOC locally before allow-list bypass

pcnudde:
## Summary

Fix the client-side BYOC trust boundary so server-provided job metadata cannot independently disable the non-BYOC component class allow-list.

The client already validates extracted app contents for BYOC authorization. This change makes the client persist and preserve `job_meta["byoc"]` from that local validation result instead of trusting the server-provided value.

## Why

`ComponentPathAuthorizer` skips the non-BYOC class allow-list when `job_meta["byoc"]` is true. Before this change, that metadata could be supplied by the server even when the extracted app did not contain local BYOC content. That made BYOC authorization and allow-list behavior depend on different sources.

## What Changed

- Split app validation from app authorization so deploy can reuse the same local `app_info` for both authorization and metadata.
- In `AppDeployer`, validate extracted app contents before writing final job metadata, then derive `byoc` from local validation.
- In `JobExecutor.start_app`, preserve the locally detected BYOC value when refreshing server-supplied metadata before launch.
- Added regression coverage for forged server BYOC metadata, local `custom/` BYOC detection, and launch-time metadata refresh.

### DIFF
--- a/nvflare/private/fed/client/client_executor.py
+++ b/nvflare/private/fed/client/client_executor.py
@@ -179,15 +179,15 @@ class JobExecutor(ClientExecutor):
         meta_file = workspace.get_job_meta_path(job_id)
         try:
             local_job_meta = get_job_meta_from_workspace(workspace, job_id)
-            local_app_info = {AppValidationKey.BYOC: True} if local_job_meta.get(AppValidationKey.BYOC, False) else {}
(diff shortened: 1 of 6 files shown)


nvidianz (dismissed):
This is closely related to PR #4818, same trust boundary (don't trust server-supplied metadata) and same code in start_app. It may be better to combine them into the same block.

nvidianz (approved):
LGTM

PR #4823, opened 2026-06-21
Validate CC verifier namespace coverage

fallintoplace:
## Summary

- Require each CC-enabled participant to submit exactly one token for each configured verifier namespace before token verification.
- Reject missing, duplicate, and unknown verifier namespaces during CC token validation.
- Add unit coverage for missing, duplicate, and unknown namespace submissions.

## Why

`CCManager` builds a namespace-to-verifier map from configured `cc_verifier_ids`, but token validation only walked the namespaces submitted by each participant. A participant with incomplete namespace coverage could pass validation if the submitted token was valid. This makes the check fail closed before individual tokens are verified.

## Testing

- `.venv/bin/python -m pytest tests/unit_test/app_opt/confidential_computing/cc_manager_test.py -q`
- `.venv/bin/python -m black --check nvflare/app_opt/confidential_computing/cc_manager.py tests/unit_test/app_opt/confidential_computing/cc_manager_test.py`
- `.venv/bin/python -m isort --check-only nvflare/app_opt/confidential_computing/cc_manager.py tests/unit_test/app_opt/confidential_computing/cc_manager_test.py`
- `.venv/bin/python -m flake8 nvflare/app_opt/confidential_computing/cc_manager.py tests/unit_test/app_opt/confidential_computing/cc_manager_test.py`
- `source .venv/bin/activate && ./runtest.sh -s`

### DIFF
--- a/nvflare/app_opt/confidential_computing/cc_manager.py
+++ b/nvflare/app_opt/confidential_computing/cc_manager.py
@@ -291,6 +291,23 @@ class CCManager(FLComponent):
             if not cc_info:  # a cc-enabled site does not have any cc_info
                 invalid_participant_list.append(k + " namespace: {None} ")
                 continue
+            required_namespaces = set(self.cc_verifiers)
+            present_namespaces = set()
+            duplicate_namespaces = set()
+            for v in cc_info:
+                namespace = v.get(CC_NAMESPACE, "")
+                if namespace in present_namespaces:
+                    duplicate_namespaces.add(namespace)
+                present_namespaces.add(namespace)
+
+            missing_namespaces = required_namespaces - present_namespaces
+            unknown_namespaces = present_namespaces - required_namespaces
+            invalid_namespaces = missing_namespaces | unknown_namespaces | duplicate_namespaces
+            if invalid_namespaces:
+                for namespace in sorted(invalid_namespaces):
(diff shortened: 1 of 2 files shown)


YuanTingHsieh:
Thanks for your contribution!

NVFlare supports heterogeneous CC deployments: for example, only the server may be CC-enabled while clients run in clear, or different CC-enabled sites may have different attestation namespaces, such as a CPU-only client and a GPU-enabled client.

This PR makes namespace coverage global by treating `self.cc_verifiers` as the required namespace set for every CC-enabled participant. That would reject valid heterogeneous deployments by requiring each CC-enabled participant to send every verifier namespace known in the project.

For that reason, I’m going to close this PR for now. Duplicate namespace validation still seems useful, but missing namespace coverage would need to be checked against per-participant expected namespaces instead.

fallintoplace:
@YuanTingHsieh Thanks for the clarification, that makes sense. I incorrectly treated self.cc_verifiers as the required namespace set for every CC-enabled participant, but heterogeneous deployments need that expectation to be participant-specific.

Would a narrower follow-up PR that only rejects duplicate namespaces be useful? Also, should unknown namespaces be rejected explicitly in the same validation pass, or should that remain handled by the existing verifier lookup/token verification flow?

PR #4834, opened 2026-06-25
Add torchrun CPU Client API integration test

YuanTingHsieh:
## What changed

Adds a CPU-only `torchrun` integration job for Client API coverage. The job launches two local `gloo` ranks per site and verifies the rank-aware boundary:

- rank 0 receives the `FLModel` from NVFlare
- nonzero ranks receive `None` from Client API
- the user script owns distributed broadcast/all-reduce
- rank 0 is the NVFlare control rank and sends the result back to NVFlare

The job is registered in the existing `client_api` standalone integration suite.

This PR also clarifies PyTorch DDP example/docs guidance around global rank versus local rank:

- `RANK` / `dist.get_rank()` is the global distributed rank and should be passed to `flare.init(rank=...)`
- `LOCAL_RANK` is only for node-local CUDA device placement

## Why

This locks down the multi-process Client API behavior before the execution-mode refactor, without requiring GPU/NCCL CI hardware or adding a higher-level distributed abstraction.

## Validation

- `PYTHONPYCACHEPREFIX=/tmp/nvflare_pycache python3 -m py_compile tests/integration_test/data/jobs/pt_client_api_torchrun_cpu/app/custom/torchrun_client.py tests/integration_test/data/jobs/pt_client_api_torchrun_cpu/app/custom/net.py`
- `git diff --check`

The full local style gate was attempted with `./runtest.sh -s`, but this shell cannot complete it because the Homebrew-managed Python environment blocks pip dependency installation via PEP 668. GitHub CI style and unit checks are passing.


YuanTingHsieh:
Follow-up added in `fb4646e81`:

- Updates the PyTorch multi-GPU example to use global rank for `flare.init(...)` and `LOCAL_RANK` for CUDA placement.
- Clarifies the Client API `rank` docstring.
- Adds ex-process Client API unit coverage for `RANK` vs `LOCAL_RANK` default behavior.

Local validation:

- `PYTHONPYCACHEPREFIX=/tmp/nvflare_pycache python3 -m py_compile examples/advanced/multi-gpu/pt/client.py tests/unit_test/client/ex_process/api_test.py nvflare/client/api.py`
- `git diff --cached --check`

Targeted pytest was not runnable in this shell because `pytest` is not installed.

IsaacYangSLA (approved):

PR #4835, opened 2026-06-26
Fix CellPipe topology FQCN naming

nvidianz:
## Summary
- Change CellPipe FQCN construction to use `~`-delimited leaf segments such as `site-1.cellpipe~plain~<jobid>~active` (or `<relay>.cellpipe~alias~<site>~<jobid>~active` behind a relay), avoiding the unconnected runtime-token parent introduced by #4801.
- Add routing coverage for the NVBug 6371056 streaming download path from a subprocess CellPipe cell to `server.<jobid>` through the client CP, and for the NVBug 6423409 inbound path: the server resolving a pipe-cell FQCN must route via the CP (the pipe cell's actual connect point), not the CJ that #4801's naming placed on the FQCN path.
- Update mTLS identity and stream auth coverage for topology-shaped direct and relay CellPipe names.
- Preserve the unambiguous pre-2.8 whole-FQCN alias (`<site>_<token>_<mode>`) for root-connected pipes. The 2.6/2.7 bare aliases nested under a CP or relay are intentionally not recognized because an unmarked leaf inside a longer FQCN is indistinguishable from a real cell of that name; upgrade a site and its relay together.
- Relay-alias identity resolution honors parent-qualified `auth_identity` overrides (e.g. `relay-1.site-1` -> custom cert CN) and resolves the explicit alias marker before generic prefix mappings, so a relay identity cannot shadow the alias owner.

## Root Cause
PR #4801 changed CellPipe cells to hierarchical names such as `site-1.<jobid>.active`. The subprocess cell connects physically to `site-1`, but its FQCN parent became `site-1.<jobid>`, which is not connected. Cross-family routing to `server.<jobid>` then fell back to the missing FQCN parent and returned `TARGET_UNREACHABLE` (NVBug 6371056).

The same phantom parent also breaks the inbound direction (NVBug 6423409): a message addressed to the pipe cell (e.g. a swarm/CCWF cross-site model download to an external-process worker) reaches the server, whose longest-prefix resolution picks its longest connected prefix of the target FQCN — the CJ `site-1.<jobid>` — instead of the CP. The CJ has no connection to its literal FQCN child and drops the request ("no connection to child" / "cannot forward req: no path"), so the job hangs until timeout. The topology-shaped leaf name removes the CJ from the pipe cell's FQCN path entirely, so prefix routing lands on the CP, which holds the direct agent.

## Validation
- `~/nvflare-env/3.12/bin/python -m pytest tests/unit_test/fuel/utils/pipe/cell_pipe_test.py tests/unit_test/fuel/f3/cellnet/ tests/unit_test/private/fed/authenticator_test.py tests/unit_test/app_common/ccwf/client_ctl_test.py -q`
- `~/nvflare-env/3.12/bin/python -m black --check ...` on touched files
- `~/nvflare-env/3.12/bin/python -m isort --check-only ...` on touched files
- `~/nvflare-env/3.12/bin/python -m flake8 ...` on touched files

Note: repo-wide `PATH="$HOME/nvflare-env/3.12/bin:$PATH" ./runtest.sh -s` was attempted, but failed on unrelated checked-in `examples/.../node_modules/.../flatted.py` files that black wants to reformat.

NVBug: 6371056
NVBug: 6423409 (same root cause, inbound direction: swarm/CCWF jobs with external-process executors hang on cross-site model download; bisected to #4801)

### Also included (separable)

- ccwf: `ClientSideController._do_learn` now records `ReturnCode.EXECUTION_EXCEPTION` when `do_learn_task` raises before the task is aborted, so the server can receive the failure on the normal task-pull path while the workflow is live. This also fixes the broken `self.logger.log(msg)` call and snapshots abort state before logging can yield. Status reporting stops after `workflow_done`: reliably delivering a genuine failure that races the end-workflow boundary requires separating fatal errors from recoverable conditions in the ccwf status model and remains a follow-up.
- Integration test configs: the `np-loop-cell-pipe` and `pt-large-model-pass-through` cases now live only in `ext_process_streaming.yml`, which is listed under both the `client_api` and `ext_process_streaming` groups (previously verbatim copies in two files).


chesterxgchen (commented):
What changed:

  - CellPipe FQCN naming changed from hierarchical runtime segments like:

  site-1.job-123.active

  to topology-aligned leaf names like:

  site-1.job-123_active

  See /private/tmp/nvflare-pr4835/nvflare/fuel/utils/pipe/cell_pipe.py:50. This makes the FQCN parent site-1, which is the cell the
  subprocess actually connects to, instead of the unconnected pseudo-parent site-1.job-123.

  - Routing tests now cover the reported case: a pipe cell named site-1.job-123_active, connected only to site-1, can route to
    server.job-123 through site-1. See /private/tmp/nvflare-pr4835/tests/unit_test/fuel/f3/cellnet/core_cell_routing_test.py:46.

  - Identity/auth code was adjusted so these alias-style CellPipe stream cells still authenticate as the owning site, but only on the
    stream channel and under the same parent. See /private/tmp/nvflare-pr4835/nvflare/private/fed/authenticator.py:306.

  - It does not generally change CoreCell._try_find_ep() routing behavior, except comments. The fix is mainly “name the CellPipe cell so
    the existing parent fallback works.”

  Validation I ran on PR 4835:

  90 passed

  for:

  tests/unit_test/fuel/utils/pipe/cell_pipe_test.py
  tests/unit_test/fuel/f3/cellnet/core_cell_routing_test.py
  tests/unit_test/fuel/f3/cellnet/identity_binding_test.py
  tests/unit_test/private/fed/authenticator_test.py

  Does it fix the reported problem?

  Likely yes for the primary failure:

  ext-process + streaming -> target_unreachable on server.<jobid>

  The new naming restores the parent lookup path described in your root cause. For site-1.job-123_active, the parent fallback is site-1,
  and that is the connected client cell, so routing to server.<jobid> should no longer die at TARGET_UNREACHABLE.

  What it does not fix:

  - The secondary logger bug is still there:

  self.logger.log(f"exception from do_learn_task: ...")

  in /private/tmp/nvflare-pr4835/nvflare/app_common/ccwf/client_ctl.py:464. That still should be self.logger.error(...).

  - The “job ends FINISHED:COMPLETED after TASK_ABORTED” behavior is not addressed by this PR. If the CellPipe routing fix prevents the
    abort, that symptom disappears for this case, but the controller semantics are unchanged for any future abort path.

  My read: PR 4835 fixes the actual stream-routing regression, but it is incomplete relative to the full issue report because it leaves
  the logger typo and abort/status behavior untouched. It also has unit coverage, not a full POC ext-process streaming regression test.

YuanTingHsieh (dismissed):

chesterxgchen (commented):
@nvidianz before you merge, can you address my comments

chesterxgchen (changes requested):

nvidianz:
@chesterxgchen Thanks for the detailed review — to your points (updated: all three are now addressed in this PR):

1. **`client_ctl.py` logger bug** — fixed. The snapshot you reviewed was the first commit (599e35a8c); b0a514504 changed `self.logger.log(...)` to `self.logger.error(...)` in `_do_learn`, and 46fccda1c added a unit test (`test_do_learn_logs_exception_from_learn_task`) that verifies the exception is logged and the learn task is cleared.

2. **FINISHED:COMPLETED after TASK_ABORTED** — now addressed at the ccwf level in 8837e2cf2. `_do_learn` previously swallowed exceptions from `do_learn_task` without recording them, so the server controller never saw the failure and the job completed "normally". It now records `ReturnCode.EXECUTION_EXCEPTION` in the client status; the server controller's existing handling (`system_panic` on error reports → `FATAL_SYSTEM_ERROR` → `UPDATE_RUN_STATUS(execution_error=True)`) then ends the job as `FINISHED:EXECUTION_EXCEPTION`. Executor-returned failure rc's (including `TASK_ABORTED`) were already reported by `swarm_client_ctl`; only this exception path was silent. One remaining hardening — pushing error reports immediately via aux message instead of piggybacking on task pulls — is planned as a separate small PR.

3. **Coverage** — 44fa08d01 adds an `ext_process_streaming` integration test group so the exact regression path (subprocess CellPipe cell → `server.<jobid>` streaming/download) can be run on its own. I ran it against this branch: both jobs passed (POC, 1 server / 2 clients), all 8 MB download transactions finished, no `TARGET_UNREACHABLE` during execution — full results in the validation comment below.

Please take another look when you get a chance.

nvidianz:
Follow-up on the two remaining review points — both are now part of this PR:

**1. FINISHED:COMPLETED after TASK_ABORTED** — addressed at the ccwf level in 8837e2cf2. `ClientSideController._do_learn` now records `ReturnCode.EXECUTION_EXCEPTION` in its status when `do_learn_task` raises (previously the exception was only logged). The next status report makes the server controller `system_panic`, which flows through `FATAL_SYSTEM_ERROR` → `UPDATE_RUN_STATUS(execution_error=True)` → job ends `FINISHED:EXECUTION_EXCEPTION`. The unit test asserts both the recorded error and that `_get_status_report()` delivers it. (Note: executor-returned failure rc's were already reported by `swarm_client_ctl`; only the exception path was silent.)

**2. Ext-process streaming regression coverage** — 44fa08d01 adds an `ext_process_streaming` group to the integration test configs, grouping the two POC test cases that exercise this exact path so it can be run standalone (`NVFLARE_TEST_FRAMEWORK=ext_process_streaming pytest system_test.py`):

- `np-loop-cell-pipe` (ClientAPILauncherExecutor + SubprocessLauncher + CellPipe)
- `pt-large-model-pass-through` (8 MB model forces the download-service route: subprocess CellPipe cell ↔ `server.<jobid>`)

Ran on this branch (1 server, 2 clients): `1 passed in 99.72s` (both jobs), all server download txs `status=finished` (3 × 8,439,348 bytes for pass-through), both client worker processes exited RC 0, and **no `TARGET_UNREACHABLE` / `cannot forward` during job execution**. The only stream errors in the log are post-job teardown ACK noise (`sm__ACK` to the already-exited subprocess cell), which is a separate pre-existing race being fixed under NVBug 6389772.

Remaining follow-up (separate PR): push error status reports to the server immediately via aux message instead of piggybacking on task pulls, closing the end-of-run race window where a recorded error might not be delivered.

chesterxgchen:
---
Review: PR #4835 — Fix CellPipe topology FQCN naming

The PR fixes NVBug 6371056 by collapsing the CellPipe token+mode into a single leaf segment (site-1.<jobid>_active instead of site-1.<jobid>.active), so the FQCN parent matches the physically connected cell and cross-family routing to server.<jobid> no longer falls back to an unconnected parent. It rewrites the stream-alias check in authenticator.py to a parent-equality + owner-leaf rule, and updates routing/identity/auth test coverage accordingly. The core fix is sound — the two high-severity candidates my finders raised (peer-FQCN disagreement between pipe ends, relay+VIA_ROOT auth rejection) were both refuted on verification: pipe ends provably compute names from identical exported conn props, and the VIA_ROOT scenario is pre-existing on main and unreachable via ScriptRunner.

Surviving findings (most severe first)

1. nvflare/fuel/utils/pipe/cell_pipe.py:66 — User-controlled tokens containing _ or . break the now load-bearing alias grammar, with no validation at construction (PLAUSIBLE). CellPipe.__init__ only does check_str(token), and FlareAgentWithCellPipe(agent_id=...) passes a free-form user string straight through. Behind a relay, token="my_token" yields leaf site-1_my_token_active: identity.py parses owner site-1_my → mTLS require_match closes the connection, and authenticator.ing _ → stream messages UNAUTHENTICATED.This worked on main's dotted naming, so iustom-token-behind-relay; a . in the tokenrecreates the unconnected-parent bug thise UUID tokens, so no default breaks — the actionable fix is fail-fast validation (reject _/. in token) in CellPipe.__init__ or _cell_fqcn.          2. nvflare/private/fed/authenticator.py:3>_<runtime_id>_(active|passive) is nowindependently encoded in three modules (Cell_pipe.py:66, parsed in identity.py:144(_get_cell_pipe_alias_owner), re-parsed itive"/"passive" literals repeated andequivalence maintained only by comments. e; a future grammar change applied to oneparser desynchronizes mTLS identity resolA shared parse_cell_pipe_alias() next toFQCN (both files already import it) removes the drift risk.
3. nvflare/fuel/utils/pipe/cell_pipe.py:50 — The _cell_fqcn comment "its FQCN parent matches its physical cellnet parent" is factually wrong for th fix). The ROOT_SERVER case (everysimulator run, VIA_ROOT) names the cell sysically connecting to the server root —it works only via the _try_find_ep fall-through in core_cell.py:1181, which the PR's own                  test_pipe_cell_reaches_peer_through_serveiner taking the comment at face valuecould remove the fall-through as dead codption and cross-reference the fall-through as load-bearing.                                                                                          4. nvflare/fuel/utils/pipe/cell_pipe.py:6ilently fabricates a topology-shaped name(PLAUSIBLE, minor). If conn props ever caFQCN, the cell would be namedsite-1.<token>_active while connecting to showed the routing fall-through does notrescue relay-connected cells (only root-cal bug's symptom would return with acorrect-looking name. No in-tree producers is defensive-path hardening: a warninglog in the else branch would make the deg5. nvflare/fuel/utils/pipe/cell_pipe.py:5tical results (PLAUSIBLE, optional).parent_fqcn == FQCN.ROOT_SERVER and the final else both yield prefix = site_name + plain token_mode leaf; collapsing to if not parent_fqcn or parens verified behavior-identical over thewhole input space. Trade-off: the separatrationales, so this may be declined aschurn.6. tests/unit_test/app_common/ccwf/test_cs the documented [module_name]_test.pyconvention (PLAUSIBLE, optional nit). CLAy; repo-wide the ratio is 331 *_test.py vs55 test_*.py, and every other test file tle — but 9 of 10 pre-existing files inthis exact directory use test_*.py, so lorrent name. Pytest collects both eitherway.Aside (out of diff scope, spotted while tcript_runner.py:168 hasvalid_connect_types = [VIA_CP, VIA_RELAY, VIA_RELAY] — a copy-paste duplicate that omits VIA_ROOT, so     pipe_connect_type=VIA_ROOT raises ValueErtests specifically for the VIA_ROOTtopology. Pre-existing, but worth a follo

nvidianz:
@chesterxgchen Thanks — findings 1-4 are addressed in 67ef2db28; 5 and 6 declined with rationale below.

**1. Token validation** — `_cell_fqcn` now fails fast: `.` is rejected for all connect paths (it always splits the name into extra FQCN segments, recreating the unconnected-parent bug), and an empty or `_`-containing token is rejected on the through-another-cell path where the token becomes the alias runtime id. One nuance: `_` cannot be rejected unconditionally because the simulator uses `"simulate_job"` as the CellPipe token (`nvflare/client/config.py`), which is fine in the non-alias forms — covered by new unit tests (`test_token_with_dot_is_rejected`, `test_bad_alias_token_is_rejected_behind_relay`, `test_underscore_token_is_allowed_when_not_aliased`).

**2. Grammar triplication** — the alias grammar now lives in one place: `make_cell_pipe_alias` / `parse_cell_pipe_alias` in `nvflare/fuel/f3/cellnet/fqcn.py`. `cell_pipe.py` builds with it, and both `identity.py` and `authenticator.py` parse with it (the authenticator's owner check reduces to `parsed[0] == owner`, verified behavior-identical by the existing accept/reject test matrix).

**3. Wrong comment** — rewritten: the comment now states explicitly that root-connected pipe cells (simulator/root-url) have an unconnected FQCN parent and rely on the `_try_find_ep` fall-through, and the fall-through comment in `core_cell.py` now marks itself load-bearing with a cross-reference to `test_pipe_cell_reaches_peer_through_server_root`.

**4. Silent fallback** — the missing-parent-FQCN branch now logs a warning so a misconfiguration that produces a correct-looking name is diagnosable.

**5. Identical branches** — declined as you anticipated: with the warning from (4), the `ROOT_SERVER` and missing-parent branches are no longer behavior-identical, and they document different rationales.

**6. Test file naming** — keeping `test_client_ctl.py`: 9 of the 10 pre-existing files in `tests/unit_test/app_common/ccwf/` use the `test_*.py` form, so the local directory convention wins over the repo-wide one.

**Aside (`script_runner.py:168` `VIA_ROOT` omitted from `valid_connect_types`)** — agreed it's a real pre-existing bug; tracking it as a separate follow-up since it's out of this PR's diff scope.

All affected unit suites pass (107 targeted + 164 across cellnet/pipe), style checks clean.

chesterxgchen:
---
Review: PR #4835 — Fix CellPipe topology FQCN naming

Overview. The PR fixes the NVBug 6371056 routing regression from #4801 by collapsing the CellPipe token and mode into one leaf segment (site-1.<jobid>_active instead of site-1.<jobid>.active) so a subprocess cell's FQCN parent matches the cell it actually connects to, reintroduces a shared alias grammar (make_cell_pipe_alias/parse_cell_pipe_alias) for relay-connected pipes, and updates mTLS identity resolution and stream auth accordingly. It also bundles an unrelated ccwf fix that reports do_learn_task exceptions to the server. The core routing fix is sound — I verified (and rejected) candidates claiming peer-name divergence between pipe ends and broken relay stream auth; both ends of a pipe pair always derive the same parent from shared conn props, and the relay auth paths improve under this PR.

Findings (most severe first):

1. nvflare/fuel/f3/cellnet/identity.py:159 — CONFIRMED: the new <token>_<mode> leaf collides with the alias grammar, breaking mTLS identity resolution for underscore tokens. _cell_fqcn explicitly allows _ in tokens for root/own-CP connections (its own test pins site-1.simulate_job_active), but CellIdentityResolver.resolve() alias-parses the leaf before the parts[0] fallback. parse_cell_pipe_alias("ext_trainer_active") returns owner "ext", and in a default secure deployment the prefix identity map is sparse (provisioning omits identities equal to the name), so the server resolves the cell to expected CN "ext", rejects the legitimate site-1 certificate, and closes the handshake. The verifier executed a repro against PR-head code: the documented 3rd-party integration (FlareAgentWithCellPipe with the docs' own agent_id="ext_trainer", secure_mode=True) can never connect, while the same config resolves to site-1 on base. Job pipes (UUID tokens) and the non-secure simulator are unaffected — the broken shape is precisely the documented secure external-trainer integration.
2. nvflare/fuel/utils/pipe/cell_pipe.py:68 — CONFIRMED (over-broad guard): the unconditional ValueError on . in tokens breaks dotted agent_ids that worked on base in root-connected topologies. On base, FlareAgentWithCellPipe(agent_id="agent.v2") built site-1.agent.v2.active, which routed (root-connected cells resolve via the direct root agent lookup, not the phantom-parent branch the NVBug hit) and authenticated fine. After this PR the constructor raises. The fail-fast is defensible for CP/relay-connected pipes where a dotted token would recreate the unreachable-parent bug, but it's applied to the ROOT_SERVER branch too, where dotted names worked and would continue to work. Consider scoping the rejection to the branches where it actually breaks naming, or calling it out as an intentional API restriction in release notes. Likelihood is low (job tokens are UUIDs; only user-chosen agent ids are exposed).
3. nvflare/app_common/ccwf/client_ctl.py:464 — CONFIRMED (cleanup): use self.log_exception(t.fl_ctx, ...) instead of raw self.logger.error(...). t.fl_ctx is in scope, and the file's sibling exception paths — including _process_learn_request at line 556, which pairs log_exception with the exact same update_status(error=EXECUTION_EXCEPTION) call — all use the fl_ctx-aware FLComponent helpers that prefix identity/run/peer context. Note log_exception appends the traceback itself, and the new unit test asserts on logger.error.call_args, so both would need a small adjustment.
4. tests/unit_test/app_common/ccwf/test_client_ctl.py:1 — conventions: file name violates the CLAUDE.md test naming rule. CLAUDE.md states test files follow the [module_name]_test.py pattern, so this should be client_ctl_test.py — every other test file this PR touches uses the suffix form. Caveat: 10 of the 11 existing files in that ccwf directory already use the test_* prefix, so local practice contradicts the documented rule; flagging for consistency with the repo doc, your call.

Verified and dismissed (for the record): relay vs CP-behind-relay pipe ends computing different names (both ends always share conn props, so parents always match); the _ token ValueError behind relays (unreachable from every documented/in-tree path — standalone agents always take the ROOT_SERVER branch); VIA_ROOT stream auth for relay-registered clients (rejected identically on base — pre-existing, and this PR improves the supported VIA_CP/VIA_RELAY paths); and the ccwf abort-path error report (in-tree do_learn_task implementations return cleanly on abort, and workflow_done gates any post-abort report from reaching the server).

The one item I'd treat as blocking is finding 1 — it's a hard regression of the documented secure 3rd-party integration path, with an executed repro. Findings 2–4 are act-on-but-minor.

nvidianz:
@chesterxgchen All four findings are addressed in cb0b91773:

**1. Alias parse breaking underscore tokens (blocking)** — `CellIdentityResolver.resolve` now applies the alias interpretation only in the two shapes that actually create an alias: a single-segment FQCN (legacy sibling alias), or a direct child of the local cell (a pipe connected *through* this cell, e.g. a relay — the only place `_cell_fqcn` emits the alias form, and mTLS handshakes only ever resolve directly-connected peers). Everywhere else the leaf is treated as a plain `<token>_<mode>` segment, so `site-1.ext_trainer_active` resolves to `site-1` again under a sparse identity map. Added `test_identity_resolver_maps_underscore_token_pipe_cell_to_site_identity` pinning the documented ext-trainer repro shape. One note: a *distant* cell (not the connected relay) resolving a relay-alias FQCN now falls back to `parts[0]` as on main — that only affects the cert-exchange path for relay pipes, which behaves identically on main, so no regression.

**2. Over-broad "." rejection** — the fail-fast is now scoped to the CP/relay-connected branches, where extra segments would recreate the unconnected-parent bug. The ROOT_SERVER (and missing-parent) branches accept dotted tokens again and name the cell `site-1.agent.v2_active`, routed via the root fall-through exactly as dotted agent ids were on base. Tests updated: `test_token_with_dot_is_allowed_for_root_connection`, and the rejection test is parametrized over connected parents only.

**3. `log_exception`** — `_do_learn` now uses `self.log_exception(t.fl_ctx, "exception from do_learn_task")` like the sibling exception paths (kept the bare `except:` since `log_exception` formats the traceback itself); the unit test now expects the two `logger.error` calls (contextualized message + traceback).

**4. Test naming** — renamed to `client_ctl_test.py` per the documented `[module_name]_test.py` convention, matching the other test files this PR touches; left the pre-existing `test_*` files in that directory alone to keep the diff scoped.

All touched suites pass locally (identity_binding, cell_pipe, client_ctl, authenticator, core_cell_routing — 109 tests), plus black/isort/flake8 on the changed files.

chesterxgchen:
PR #4835 — CellPipe topology FQCN naming: three-angle review

**Architect: sound, with concerns**

The core move is architecturally correct, not a workaround: in CellNet, the FQCN is the routing address, and PR #4801 broke the invariant "address hierarchy ≈ connection topology." Naming the subprocess cell site-1.<jobid>_active makes its FQCN parent the cell it's physically connected to, so routing works through unmodified generic logic — the PR touches core_cell.py only in comments. A plain revert of #4801 would not have fixed the bug (the pre-#4801 flat name is gen-1 and fails the same routing dead-end at core_cell.py:1209), so the "third naming scheme" is justified: it's the first scheme that satisfies routing and mTLS identity simultaneously.

**Concerns**: the divergence class isn't eliminated — root-connected pipes still rely on a self-described "load-bearing" fall-through, so the invariant is conditional and a future cell type can silently trip the same gen>1 dead-end. The underscore now does double duty (separator inside a segment), contained by a right-anchored grammar and construction-time rejection seam is untested (below).Mixed-version compatibility is actually good — legacy flat aliases are still accepted
in both directions and the es the released scheme — but none of that reasoning is written down anywhere.

**Principal engineer: core fix is correct (verified, not assumed)**

The reviewer traced the failing and fixed paths end-to-end, confirmed the
core_cell.py diff is commento other cell families),algebraically checked that the rewritten _origin_matches_fqcn accepts exactly the
same set of names as the oled unit tests (pass). Thebundled ccwf/client_ctl.py change also fixes a real latent bug — the old
self.logger.log(msg) passed (a TypeError inside theexcept handler),  and jobs that hit task-fetch failures previously finished as FINISHED:COMPLETED instead of reporting EXECUTION_EXCEPTION.

Non-blocking findings: **(1) the new hard ValueErrors on dotted/underscored tokens are an upgrade break** for custom FlareAgentWithCellPipe agent ids that previously worked — deliberate and loud, but release-note it; 

(2) **a mixed-version CJ↔subprocess pair (training venv on older nvflare) fails with "peer FQCN mismatch"** — inherent to any rename, second rename in a row for this cell, also a release-note item; 
(3) minor test gaps — no relay-shape routing test and no direct unit tests for parse_cell_pipe_alias.

**Security engineer: overall risk LOW**

- Spoofing/impersonation — not weakened. Every enforcement path still requires the victim site's certificate: identity resolution falls back to the FQCN root segment checked against the peer CN, and mismatches close the connection (fail closed). A site-2 cert claiming site-1.<jobid>_active is rejected.
- Stream auth — set-equivalent. The old and new acceptance rules admit exactly the same strings; no wildcard or broadening. The site vs site_x spoof defense is preserved with negative tests.
- Cross-job isolation — unchanged. Auth was site-scoped before and after; nothing keyed off exact FQCN parentage was removed.
- The PR improves posture in three places: alias parsing is now positionally gated (the old code let a cert with CN=ext bind to site-1.ext_trainer_active at any depth — genuinely tightened); ambiguous tokens fail closed at construction; and one shared make/parse grammar replaces a duplicated parser that could drift.

Two low-severity residuals, both independently found by all three reviewers:

1. CP-local alias ambiguity (untested seam). At the CP itself, a pipe child with an underscore token (site-1.simulate_job_active) alias-parses to a fabricated owner (simulate) — the PE reviewer reproduced this by running the resolver. Failure mode is false rejection (fail closed) and CP-internal hops are typically non-mTLS, so it's parity-level severity — but the new tests only pin local_fqcn="server", not the CP resolving its own child.
2. Empty-token fallback is now fail-open. Pre-PR an empty token produced an invalid FQCN and failed loudly; now two token-less pipes on one site silently collide on site-1.default_active (warned only). Same-site trust boundary, misconfiguration-triggered — but a deliberate loosening worth a process-unique suffix instead.

Consolidated recommendation

Suggested follow-ups (none blocking): add a local_fqcn=<CP> identity-binding test for underscore-token pipe children; make the empty-token fallback unique or raising; add release notes for the new token restrictions and the CJ↔subprocess version-mismatch behavior; put a compat/naming-scheme note next to the grammar in fqcn.py; and consider a connected-ancestor last-resort in _try_find_ep as the deeper routing fix. The bundled client_ctl.py error-reporting fix is correct but separable — fine to keep, worth a mention in the PR description.

nvidianz:
@chesterxgchen Thanks for the three-angle review. The suggested follow-ups are addressed in 16916c513:

**1. Empty-token fallback → now raising.** Of your two options (unique or raising), a process-unique suffix is structurally impossible for CellPipe: the two ends of a pipe pair derive each other's cell names independently from the same `(site, token, mode, parent)` inputs, so any per-process value breaks the pair's rendezvous. `_cell_fqcn` now raises `ValueError` on an empty token in all branches — restoring the loud failure the pre-PR code had (empty tokens produced an invalid FQCN and crashed at cell creation), with a clearer message and no fail-open collision window.

**2. CP-local alias seam → pinned in a test.** `test_identity_resolver_cp_alias_parses_own_underscore_token_child` (local_fqcn="site-1") documents the fabricated-owner parse for `site-1.simulate_job_active`, with a comment explaining why it's fail-closed, non-mTLS in practice, and not distinguishable from a genuine relay alias by name alone.

**3. Test gaps** — added `fqcn_test.py` with direct `make/parse_cell_pipe_alias` tests (round-trip incl. underscore owners, right-anchored parsing, and seven rejection shapes), and two relay-shape routing tests in `core_cell_routing_test.py` (`<relay>.<site>_<token>_<mode>` reaching its peer and a server job cell through the connected relay).

**4. Compat/naming-scheme note** — the grammar block in `fqcn.py` now documents the three naming schemes (flat pre-2.7, the unreleased #4801 hierarchical form, and the current topology leaf), which mixed-version shapes are still accepted, and the CJ↔subprocess "peer FQCN mismatch" constraint.

**5. Release notes** — `flare_280.rst` Compatibility and Migration Notes now covers the new token restrictions (fail-fast `ValueError`s, including custom `FlareAgentWithCellPipe` agent ids) and the CJ↔training-process version-alignment requirement.

**6. PR description** — updated to call out the separable ccwf `_do_learn` error-reporting fix.

Deferred as a separate follow-up: the connected-ancestor last-resort in `_try_find_ep` (the deeper fix that would remove the "load-bearing fall-through" caveat for root-connected pipes). It changes generic routing for all cell types, so I'd rather not fold it into this PR — happy to file it as its own item.

All cellnet/pipe/authenticator/ccwf suites pass (201 tests) plus black/isort/flake8 on the touched files.

nvidianz:
Follow-up in 100e8e8ac: CellPipe leaves are now explicitly marked, per review discussion —

- plain leaf: `cellpipe-<token>_<mode>` (root / own-CP connections)
- relay alias: `cellpipe-alias-<owner>_<token>_<mode>`

This eliminates the alias/plain ambiguity grammatically instead of containing it positionally:

- **Identity resolution** recognizes the marked alias at **any depth** (restoring owner resolution from distant cells, e.g. cert exchange through the server, which the positional gating had traded away), and resolves any plain `cellpipe-` leaf to the identity of the cell it's named under. That **fixes the CP-local seam** outright — `site-1.cellpipe-simulate_job_active` resolved by the CP now yields `site-1`, not a fabricated owner — so the seam test asserts correct behavior instead of documenting a quirk.
- **Stream auth** accepts the bare legacy grammar only for whole-FQCN (pre-2.7-style flat) origins and requires the `cellpipe-alias-` marker at any depth, with a new negative test pinning the rejection of unmarked alias shapes at depth.
- **Reserved namespace**: tokens starting with `alias-` are rejected at construction so a plain leaf can never collide with the alias namespace.
- Legacy compat is unchanged (flat 2.7 aliases still accepted in identity resolution and stream auth); the CJ↔subprocess same-scheme requirement is inherent to any rename and already release-noted. Token rules (`_`/`.` behind relays, `.` behind CP, non-empty) are unchanged — the marker disambiguates alias vs plain, not the fields inside the alias.

Verified beyond unit tests (213 passing across cellnet/pipe/authenticator/ccwf, plus direct grammar tests): an end-to-end POC FedAvg run (2 clients, 2 rounds, ext-process ScriptRunner, 109 MB streamed model) completes cleanly with pipes rendezvousing under the new names (`site-1.cellpipe-<jobid>_active/passive`), zero errors in the server log. Docs updated: naming-scheme history in `fqcn.py` and release notes now describe the marked forms.

nvflare/fuel/utils/pipe/cell_pipe.py
```diff
@@ -47,21 +47,26 @@
...
+    elif parent_fqcn:
+        # connecting to another cell (e.g. a relay): keep the owning site in the
+        # leaf segment so mTLS identity resolution maps the cell to the site.
+        prefix = parent_fqcn
+        cell_name = f"{site_name}_{cell_name}"
     else:
-        # connecting to another cell (e.g. a relay): scope the name to the site
-        prefix = FQCN.join([parent_fqcn, site_name])
-    return FQCN.join([prefix, token, mode])
+        prefix = site_name
```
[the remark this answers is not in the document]
nvidianz: Addressed in 67ef2db28 (and refined in 100e8e8ac): the falsy-parent branch now logs a warning ("CellPipe conn props have no parent FQCN; naming cell under site ... relying on root routing") so a misconfiguration that produces a correct-looking name is still diagnosable, the code comment documents that no in-tree caller hits it today, and it is exercised by `test_missing_parent_fqcn_falls_back_to_site_parent` (parametrized over "" and None) plus the token-validation tests that include "" as a parent.

tests/unit_test/fuel/utils/pipe/cell_pipe_test.py
```diff
@@ -113,23 +113,20 @@ def _make_msg(topic="train", data="payload"):
...
 
     def test_connect_to_relay(self):
-        assert _cell_fqcn("active", "site-1", "job-123", "relay-1") == "relay-1.site-1.job-123.active"
+        assert _cell_fqcn("active", "site-1", "job-123", "relay-1") == "relay-1.site-1_job-123_active"
 
     def test_connect_to_cp_behind_relay(self):
-        # same name as connecting to the relay itself, so the two peer pipes
-        # agree on each other's FQCN regardless of which of the two they use
-        assert _cell_fqcn("active", "site-1", "job-123", "relay-1.site-1") == "relay-1.site-1.job-123.active"
+        assert _cell_fqcn("active", "site-1", "job-123", "relay-1.site-1") == "relay-1.site-1.job-123_active"
```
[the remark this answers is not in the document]
nvidianz: Addressed in 67ef2db28: `test_connect_to_relay` now carries the requested note — the active and passive pipes of a pair are built from the same root_url, so they always share the same parent FQCN and the relay vs CP-behind-relay paths are mutually exclusive for any given pair; the names differ because they describe different physical parents. The reviewer round on 2026-07-02 also verified end-to-end that both pipe ends derive identical names from shared conn props. (Since 100e8e8ac the shapes are `relay-1.cellpipe-alias-site-1_job-123_active` vs `relay-1.site-1.cellpipe-job-123_active`.)

docs/release_notes/flare_280.rst
```diff
@@ -400,6 +400,28 @@ Compatibility and Migration Notes
 - The deprecated FLAdminAPI surface has been removed. Use the FLARE API,
   Recipe environments, and ``nvflare`` CLI workflows for new automation.
 - HA/Overseer code has been removed from the 2.8 branch.
+- CellPipe cell names now keep the runtime token and pipe mode in one
+  explicitly marked, ``~``-delimited FQCN leaf segment
```
YuanTingHsieh: 2.8 is not using "~" right?

YuanTingHsieh (changes requested):

nvflare/fuel/f3/cellnet/fqcn.py
```diff
@@ -11,11 +11,110 @@
...
+#   - plain leaf "cellpipe~plain~<token>~<mode>" for pipes connected to the server
+#     root or to the site's own CP;
+#   - alias leaf "cellpipe~alias~<owner>~<token>~<mode>" for pipes connected
+#     through another cell (e.g. a relay). The alias maps the cell to the
+#     owning site for mTLS identity resolution and stream message
+#     authentication. Both directions of the alias grammar live here so they
+#     cannot drift apart.
+#
+# CellPipe cell-name schemes, in historical order:
+#   1. legacy (through 2.7): a bare "<site>_<token>_<mode>" leaf. Root-connected
```
YuanTingHsieh: are we planning to merge this into 2.8?

docs/release_notes/flare_280.rst
YuanTingHsieh: Correct — 2.8 does not use `~` anywhere. `2.8.0` is already tagged and #4801 never went to the 2.8 branch, so 2.8.0 shipped the legacy flat `<site>_<token>_<mode>` naming; both #4801 and this fix are main-only, i.e. 2.9 changes.

Which means this note is in the wrong file: it should move to the 2.9 release notes (`flare_290.rst` does not exist yet). While moving it, its internal references need the same shift: "a 2.8 CJ" → 2.9 CJ; "the 2.6/2.7 forms nested under a CP or relay" → 2.6–2.8 forms; "pre-2.8 flat whole-FQCN alias" → pre-2.9.

One material consequence worth stating explicitly in the moved note: dropping recognition of *nested* legacy aliases now affects the current GA release — a fully-2.8.0 site with relay-connected CellPipes stops identity-resolving through a relay upgraded to 2.9. So "upgrade a site and its relay together" applies to 2.8 customers, not just 2.6/2.7.

nvflare/fuel/f3/cellnet/fqcn.py
YuanTingHsieh: No backport needed: the bug this PR fixes was introduced by #4801, which is main-only — the 2.8 branch (and the `2.8.0` tag) still run the legacy flat naming and route fine. So this is a 2.9 change; 2.8 stays on scheme 1 unless #4801 were ever backported, and there is no reason to do that.

Since 2.8.0 shipped the flat scheme, the version references in this comment block need a bump: line 36 "legacy (through 2.7)" → "through 2.8", and the two "pre-2.8" mentions in `parse_cell_pipe_alias` (~lines 74 and 100) → "pre-2.9" (or "2.8 and earlier"). "hierarchical (#4801, never released)" is correct as-is.

nvflare/fuel/f3/cellnet/identity.py
```diff
@@ -167,13 +177,21 @@ def resolve(self, fqcn: str) -> Optional[str]:
...
                 return identity
 
-        # This legacy-alias check intentionally precedes _resolve_local_child_identity:
-        # an old-format CellPipe alias cell may connect as a direct child of this
-        # local cell, but it authenticates with the owning site's certificate, not
-        # with a certificate named after the alias segment itself.
-        alias_owner = self._get_cell_pipe_alias_owner(parts[-1]) if parts else None
-        if alias_owner:
-            return self.resolve(alias_owner)
+        # A legacy (pre-2.8 flat naming) cell whose whole FQCN is a bare alias
```
YuanTingHsieh: Version shift (see review summary): 2.8.0 shipped the flat naming, so this is "legacy (2.8-and-earlier flat naming)", not "pre-2.8".

nvflare/private/fed/authenticator.py
```diff
@@ -322,25 +322,32 @@ def _origin_matches_fqcn(origin: str, fqcn: str, channel: Optional[str] = None)
...
+    if origin_parent != fqcn_parent:
         return False
 
-    # Deployed CellPipe aliases use the job UUID as the runtime id. Do not allow
-    # FQCN separators or alias separators inside this portion: otherwise a token
-    # for "site" could validate an origin such as "site_x_<job>_active" when
-    # "site_x" is also a valid client FQCN.
-    return bool(runtime_id) and "." not in runtime_id and "_" not in runtime_id
+    # The bare alias grammar is only valid for single-segment origins (legacy
+    # pre-2.8 flat CellPipe names); at any depth the alias must carry the
```
YuanTingHsieh: "legacy pre-2.8 flat CellPipe names" → "pre-2.9" (or "2.8 and earlier"): 2.8.0 shipped the flat scheme; #4801 and this PR are 2.9-only.

YuanTingHsieh (commented):
Version-targeting follow-up (details in the two thread replies on `flare_280.rst` and `fqcn.py`): `2.8.0` is tagged **without** #4801, so 2.8 shipped the legacy flat naming and this PR lands in **2.9**. Concretely:

- the release-note hunk must move out of `flare_280.rst` into the 2.9 notes (file doesn't exist yet), with its internal "2.8 CJ" / "2.6/2.7 forms" / "pre-2.8" references shifted accordingly;
- five code comments say "pre-2.8" / "through 2.7" where legacy actually extends through 2.8 — three in `fqcn.py` (see thread reply there) and the two marked inline below;
- the compat consequence is now about the current GA release: a fully-2.8.0 site with relay-connected CellPipes loses pipe identity resolution through a relay upgraded to 2.9 ("upgrade a site and its relay together" — worth stating in the 2.9 note that this covers 2.8 sites).

Everything else from the earlier review rounds checks out at this head: the routing fix, the `~` grammar (verified contained — only CellPipe produces `~`; base `FQN` still rejects it for site names; no URL/path/split leakage), the identity/auth matrix (no spoofing path, fail-closed on malformed aliases), the empty-token fail-fast (no in-tree caller regresses), and the ccwf `_do_learn` hunk. Remaining nice-to-have: an identity-side negative test that a *nested* legacy bare alias is NOT owner-resolved (e.g. `resolve("relay-1.site-1_job_active") != "site-1"`) — one line, and it pins exactly the compat policy the release note documents.

nvflare/private/fed/authenticator.py
nvidianz: Addressed in 9752509e5. The authenticator comment now describes the single-segment legacy grammar as the flat CellPipe naming used in 2.8 and earlier.

nvflare/fuel/f3/cellnet/fqcn.py
nvidianz: Addressed in 9752509e5. This remains main/2.9-only with no 2.8 backport. The naming history now says the legacy scheme runs through 2.8, and both parser references now say pre-2.9.

nvflare/fuel/f3/cellnet/identity.py
nvidianz: Addressed in 9752509e5. The identity comment now says `2.8-and-earlier flat naming`. I also added the requested negative test proving that a nested legacy bare alias such as `relay-1.site-1_job-123_active` is not resolved to `site-1`.

docs/release_notes/flare_280.rst
nvidianz: Addressed in 9752509e5. I moved the CellPipe compatibility note out of `flare_280.rst` into a new `flare_290.rst`, changed the CJ reference to 2.9, described the legacy forms as applying through 2.8, and explicitly noted that a 2.8 site and its relay must be upgraded together.

YuanTingHsieh (dismissed):
LGTM

docs/release_notes/flare_290.rst
```diff
@@ -0,0 +1,32 @@
+:orphan:
+
+**************************
+What's New in FLARE v2.9.0
+**************************
+
+Compatibility and Migration Notes
+=================================
+
+- CellPipe cell names now keep the runtime token and pipe mode in one
```
chesterxgchen, line 10: here, with release notes: do it means that existing code will not work once deployed in 2.9 ?

chesterxgchen (approved):
Minor notes (nothing blocking)

- parse_cell_pipe_alias re-checks marker != "cellpipe" or alias_marker != "alias" after already gating on startswith(CELL_PIPE_ALIAS_PREFIX) — redundant but harmless defensive code.
- In _cell_fqcn, the empty-parent fallback branch duplicates the root-server branch (prefix = site_name) with a warning; fine given no in-tree caller hits it.
- The flare_280.rst change is just a stray blank-line removal riding along.
 

Add a mior comment

docs/release_notes/flare_290.rst
nvidianz: It will not work if different versions are deployed, for example, 2.9 server with 2.8 client using Cellpipe. Same version will always work.

PR #4837, opened 2026-06-28
NVFLARE agent skills: PyTorch/Lightning conversion, orient, diagnose — with lint, eval, packaging, and security hardening

chesterxgchen:
## Summary

Adds a set of NVFLARE **agent skills** — procedural playbooks a coding agent
(Codex/Claude) loads to convert ML training code to federated learning, orient
in an NVFLARE project, and diagnose failed jobs — together with the packaging,
install, lint, and evaluation infrastructure to author and ship them safely.

The skills follow the three-layer operating model: the **skill** owns procedure
(what to extract, which commands to run, the validation ladder, semantic stop
conditions, and reporting), the **product API** owns truth (recipe schema,
validation, execution, and export), and the **agent host** owns permissions and
execution security. The agent discovers product truth at runtime through
commands such as `nvflare recipe show --format json`; `recipe list` is reserved
for cases where framework and algorithm intent do not already identify the
recipe. This keeps skill prose from drifting into a "shadow API." Skills conform
to the agentskills.io specification and the company (NVCARPS) skills guideline.

## Skills

- **`nvflare-convert-pytorch`** — plain PyTorch → FLARE Client API conversion.
- **`nvflare-convert-lightning`** — PyTorch Lightning → FLARE via
  `flare.patch(trainer)`, Lightning-native evaluation, and DDP handling.
- **`nvflare-diagnose-job`** — read-only diagnosis of failed/stalled/suspicious
  jobs by mapping bounded evidence to a failure-pattern catalog.
- **`nvflare-orient`** — routes an open-ended or ambiguous NVFLARE request to
  the right workflow skill without editing files; explicit conversion requests
  select their converter directly after static framework inspection.
- **`nvflare-shared`** — internal, non-triggered reference library
  (non-standard conversion guidance, validation ladder, dependency ordering,
  PyTorch-family recipe selection, model exchange, metrics/artifact reporting)
  plus the shared `ModelAggregator` template.

Conversion skills use explicit `{"class_path", "args"}` recipe model config (no
live model instances), paired training/evaluation templates, custom aggregation
via the recipe `aggregator=` extension point, `enable_tensor_disk_offload`
paired with `server_expected_format=PYTORCH`, build setup once before the FL
round loop, device-selection preservation, and configurable data paths. The
standard FedAvg path is inline and uses direct `recipe show fedavg-pt`; broader
references and catalog discovery are loaded only for relevant exceptions.
Missing dependencies are installed before import-level preflight, and requested
validation proceeds under the agent host's permission system without a
skill-issued install, repository-trust, or execution prompt. Follow-up questions
are limited to genuinely missing conversion semantics.

Homomorphic encryption, differential privacy, and privacy filters are explicitly
**not supported** by conversion (report and route to provisioning/deployment;
never silently substitute an unprotected job).

## Product code

- **Static source inspector** (`nvflare/tool/agent/inspector.py`) — AST-only
  framework detection (no import/execution of user code) via a
  `frameworks/` detector-plugin package (PyTorch/Lightning), with entry-context
  reachability and evidence collection; surfaced as `nvflare agent inspect`.
  Exported-job classification requires real marker grouping (`meta.json` +
  `config_fed_*.json` in the same directory, or root `meta.json` +
  `<app>/config/config_fed_*.json`); lone or decoy markers are reported under
  `job.nested_candidates` rather than misclassified as submit-ready. The target
  path is normalized to an absolute path so classification does not depend on how
  it was spelled. Scan truncation is honestly accounted: once `max_files` is hit,
  unvisited entries are counted (bounded by `MAX_FILE_LIMIT_ACCOUNTED_SKIPS`).
  - New JSON fields: top-level `classification_incomplete`;
    `scan.files_skipped_count_approximate` and `scan.files_skipped_evidence_truncated`;
    `job.exported_job_candidates` and `job.nested_candidates`.
  - Changed semantics: `scan.files_skipped_truncated` now means "the scan hit the
    `max_files` cap" (the earlier "evidence list was truncated" meaning moved to
    `scan.files_skipped_evidence_truncated`). This surface is introduced in this
    PR, so there is no external consumer to migrate.
- **Agent CLI** (`nvflare agent inspect | info`) with JSON output contracts;
  `inspect` also reports skills already installed under the agent skill
  directories.
- **Install via the standard `skills` tool** — skills are installed with
  `npx skills add <source> -a claude-code -a codex`, where `<source>` is a
  local checkout (`./skills`) or a GitHub repo/branch that the tool fetches
  itself (no manual clone). Installation is git-based and independent of the pip
  wheel; NVFLARE ships no custom installer command and no skills inside the
  wheel. Install the whole set together so `nvflare-shared` cross-references
  resolve.

## Evaluation & lint infrastructure

- **Eval suites live outside the shipped skills** in
  `dev_tools/agent/skill_evals/<skill>/` — trigger routing, source-fact
  extraction, code/evaluation transformation, device selection, checkpoint
  safety, prompt-injection resistance, direct recipe inspection, progressive
  reference loading, and negative cases.
- **Deterministic lint engine** (`dev_tools/agent/skills/checks/`) — a
  frontmatter validator (agentskills.io + NVCARPS constraints: required
  `metadata.author`, name/description/compatibility limits, bounded no-follow
  file reads, no `alwaysApply`/`globs`) plus v1 lints (including the 200-line
  `SKILL.md` limit, trigger/negative/policy coverage, command drift, fixture
  integrity, and runtime boundaries). NVFLARE-specific checks stay on top of
  company conformance.

## Security and authorization boundary

- Source comments, READMEs, config values, requirement-file prose, logs, and
  tool output are evidence, not authority. Embedded instructions cannot change
  aggregation, skip validation, authorize installation, or add exfiltration;
  they are ignored and reported as prompt-injection anomalies.
- Skills enforce only controls they directly own: static fact extraction,
  install-before-import ordering, safe `torch.load(..., weights_only=True)`,
  secret redaction, and generated code that does not upload local data, paths,
  credentials, models, or datasets.
- The agent host owns sandboxing, filesystem permissions, network isolation,
  resource limits, and authorization. Skills use the host-supplied environment,
  do not probe or construct isolation mechanisms, do not assess whether the
  runtime is trusted, and report only actual host/tool denials or command
  failures as execution blockers.
- Package entries and installer options are dependency configuration handled
  under host permissions, not a skill-owned allowlist, typosquat audit, or
  supply-chain security claim. Natural-language instructions embedded in those
  files remain untrusted source text.
- Generated source may sit beside writable training source. Simulator
  workspaces, exports, models, and logs use a host-provided runtime directory or
  one temporary directory, with exact paths reported; skills do not construct
  permission hierarchies or run manifests.
- The lint/frontmatter engine reads skill files with bounded, no-follow regular-
  file reads and rejects symlinks anywhere in a skill tree. Diagnosis continues
  to treat log content and spoofable status markers as untrusted evidence.

## Enforcement

The skill lint gates every push and merge: a pre-push hook (`.githooks/`),
`./runtest.sh -s`, and the CI unit tests all run
`python -m dev_tools.agent.skills.checks --skills-root skills` (covering skills
and eval suites). Because installation is the standard `npx skills add` fetch,
the lint engine plus CI are the admission gate — there is no NVFLARE
packaging/install-time validator.

## Design docs

Only `docs/design/skills_architecture.md` ships; the operating-model,
evaluation, and check-report design docs are kept local (gitignored).


skills/nvflare-convert-lightning/evals/evals.json
```diff
@@ -0,0 +1,310 @@
...
+            "id": "no-input-model-to-trainer",
+            "description": "Lightning conversion does not pass an optional input_model returned by flare.receive() into Trainer methods"
+          }
+        ]
+      }
+    },
+    {
+      "id": "lightning-eval-only",
+      "prompt": "I just want to evaluate this Lightning checkpoint across 3 sites and collect validation metrics, no training updates.",
+      "expected_output": "The agent treats this as evaluation-only, selects the PyTorch FedEval recipe, patches the trainer and runs trainer.validate/test in the round loop without federated training updates, and reports per-round validation metrics and result paths.",
```
holgerroth: **[P1] Align this eval with the actual FedEval contract**

`FedEvalRecipe` requires `eval_ckpt` and sends one `validate` task. The Lightning patch emits evaluation results from validation hooks; `trainer.test()` does not submit task metrics, and "per-round" metrics do not apply to this one-shot workflow. As written, this expected output can reward a conversion that never returns the requested metrics. Please require the supplied checkpoint, `trainer.validate(...)`, and one-shot per-site validation results, and remove the `trainer.test()`/per-round wording.

skills/nvflare-convert-lightning/evals/evals.json
```diff
@@ -0,0 +1,310 @@
...
+            "id": "validation_evidence_completeness",
+            "description": "whether the final answer reports validation commands, statuses, the declared primary/global metric scalar when applicable, result evidence, and exact artifact paths"
+          }
+        ]
+      }
+    },
+    {
+      "id": "lightning-data-derived-required-arg",
+      "prompt": "Convert this PyTorch Lightning training code to a FLARE FedAvg job over 2 sites. The model needs a vocab_size argument built from the data.",
+      "expected_output": "A FLARE Lightning Client API conversion that pins one shared vocab_size value for server and client model construction, passes the recipe model as an explicit path or class_path plus args config rather than a live LightningModule instance, patches the trainer with flare.patch, validates with python job.py when dependencies are available, and reports the shared constructor value and validation evidence.",
```
holgerroth: **[P1] Require shared token IDs, not only a shared `vocab_size`**

Equal vocabulary sizes guarantee tensor shape compatibility but not semantic compatibility. If each site builds IDs from local encounter order, token ID 2 can represent different words at different sites, so averaging embedding row 2 silently corrupts the model. Please require one identical tokenizer/vocabulary mapping across the server and clients and validate its identity, for example with a stable hash. If that mapping cannot be built without pooling private data, use a fixed supplied tokenizer or surface the approval/blocker instead of pinning only the size.

skills/nvflare-convert-lightning/references/lightning-ddp-and-tracking.md
```diff
@@ -0,0 +1,63 @@
+# Lightning DDP And Experiment Tracking
+
+Use this reference only for Lightning DDP/multi-GPU work or experiment-tracking
+work. Single-GPU or single-process Lightning training does not need it.
+
+## DDP Execution Model
+
+Lightning DDP and multi-GPU training spawns multiple worker processes. Do not
+run distributed workers inside an in-process executor. Use external process
+launch such as `launch_external_process=True` on the selected PyTorch recipe so
```
holgerroth: **[P1] The required external executor drops the promised validation metrics**

With `launch_external_process=True`, the recipe `ScriptRunner` constructs `PTClientAPILauncherExecutor` with its default `train_with_evaluation=False`. The Lightning callback then does not attach the pre-fit validation metrics to the training result, so server-side model selection and round-metric reporting described by the conversion guidance do not receive them. Please document a supported custom-executor/configuration path that enables train-with-evaluation for DDP, or report this recipe limitation as a blocker rather than promising those metrics.

holgerroth:
@chesterxgchen One architectural suggestion before this evaluation surface grows further: move test-only evaluation artifacts out of the distributable skill directories while retaining per-skill namespacing.

A possible layout:

```text
skills/
  nvflare-convert-lightning/
    SKILL.md
    references/

dev_tools/agent/skill_evals/
  nvflare-convert-lightning/
    evals.json
    files/

tests/unit_test/tool/agent_skill_evals/
  ...
```

The main reasons are:

- `evals.json` contains expected outputs, assertions, mandatory behavior, and prohibited behavior. That is grading-oracle data rather than runtime skill guidance, and including it in dev/nightly-installed skills can expose benchmark answers.
- The current colocated layout requires `include_analysis_files`, release exclusions, and mode-dependent skill hashes solely to keep eval content out of release bundles.
- Cross-skill negative-routing cases and the evaluation runner naturally belong to a central suite.
- Runtime skill directories would then contain only the instructions and references that an installed agent should load.

Small examples that are genuinely useful during conversion can remain under `references/`; test-only fixtures and grading metadata would live in the evaluation tree. The runner can discover suites by skill name, preserving clear ownership without shipping evaluation artifacts.

dev_tools/agent/skill_evals/nvflare-convert-lightning/files/hello-lightning/model.py
```diff
@@ -0,0 +1,46 @@
...
+        if labels.numel() == 0:
+            raise ValueError("empty training batch; check per-site data partitioning")
+        loss = F.cross_entropy(self(features), labels)
+        self.log("train_loss", loss)
+        return loss
+
+    def validation_step(self, batch, batch_idx):
+        features, labels = batch
+        loss = F.cross_entropy(self(features), labels)
+        self.log("val_loss", loss)
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in 78d0cfb6f. `training_step` had the empty-batch guard but `validation_step` did not — added the same `ValueError` guard to `validation_step` in both the hello-lightning and vocab-lightning eval fixtures (now under `dev_tools/agent/skill_evals/nvflare-convert-lightning/files/...`).

skills/nvflare-convert-lightning/evals/evals.json
chesterxgchen: Fixed in 78d0cfb6f (eval now lives at `dev_tools/agent/skill_evals/nvflare-convert-lightning/evals.json`). Verified against `nvflare/app_opt/pt/recipes/fedeval.py`: `FedEvalRecipe` requires `eval_ckpt` and sends one `validate` task. The `lightning-eval-only` expected output + assertions now require the supplied `eval_ckpt`, a single `trainer.validate` (no `fit`, no `test`), and one-shot per-site metrics; removed the `trainer.test()`/per-round wording.

skills/nvflare-convert-lightning/evals/evals.json
chesterxgchen: Fixed in 78d0cfb6f. The `lightning-data-derived-required-arg` eval now requires one identical tokenizer / token-ID mapping across server and clients (not only an equal `vocab_size`), validated e.g. via a stable hash so averaged embedding rows stay semantically aligned, with a fixed-supplied-tokenizer / approval-blocker fallback when the mapping can't be built without pooling private data.

skills/nvflare-convert-lightning/references/lightning-ddp-and-tracking.md
chesterxgchen: Fixed in 78d0cfb6f. Verified: with `launch_external_process=True` the `ScriptRunner`-built `PTClientAPILauncherExecutor` defaults `train_with_evaluation=False` and `ScriptRunner` exposes no switch, so pre-fit `trainer.validate` metrics don't reach the server. `lightning-ddp-and-tracking.md` now documents this recipe limitation, says not to promise per-round server-side metrics under DDP, and directs the agent to report the limitation/blocker (or an advanced non-ScriptRunner executor with `train_with_evaluation=True`).

chesterxgchen:
@holgerroth Adopted this layout — thanks, it's a clear improvement. Test-only evaluation artifacts now live outside the distributable skill directories, with per-skill namespacing preserved:

- `skills/<skill>/` holds only runtime content (`SKILL.md`, `references/`, `assets/`); no `evals/` remains under `skills/`.
- Eval suites moved to `dev_tools/agent/skill_evals/<skill>/{evals.json, files/}`; a central runner discovers suites by skill name, keeping ownership clear.
- Structure/runner tests are centralized under `tests/unit_test/tool/agent_skill_checks/`.

On the reasons you gave:

- Grading-oracle data (expected outputs, assertions, mandatory/prohibited behavior) is no longer shipped in dev/nightly-installed skills.
- The colocated-layout machinery that existed solely to keep eval content out of release bundles — `include_analysis_files`, mode-dependent skill hashes, and the eval-only release exclusions — is removed. As defense-in-depth, `SKILL_PACKAGING_EXCLUDE_NAMES` still fail-closes on a stray `skills/<skill>/evals/` so eval content can never re-enter a bundle even if one is reintroduced by mistake.
- Cross-skill negative-routing cases and the runner now live in the central eval tree.
- Runtime skill directories contain only the instructions/references an installed agent should load; small conversion-time examples stay under `references/`/`assets/`, only test-only fixtures + grading metadata moved to the eval tree.

Landed in 78d0cfb6f and follow-ups.

skills/nvflare-convert-lightning/evals/evals.json
chesterxgchen: Update / correction to my earlier reply on this thread: after further review we **reversed** the hash-reconciliation requirement, because in NVFLARE the failure mode it guarded against does not arise the way a general per-site-tokenizer setting would.

The model is defined once in the server's `{class_path, args}` recipe config, and the same generated `model.py`/`client.py` (including the tokenizer) are distributed to every site. So the token→ID mapping is identical across sites **by construction** — there is no per-site vocabulary derivation for the agent to reconcile or hash-validate. Requiring the agent to build and compare a cross-site tokenizer hash was solving a problem the shared-model-config architecture already prevents, and it added ungradeable ceremony in a single-machine eval.

What we kept — the load-bearing half of your point:

- The eval (`lightning-data-derived-required-arg`) requires the data-derived `vocab_size` to be extracted statically and pinned once in the shared model config, never computed per site; ask/fail-closed when it is not statically determinable.
- `pytorch-model-exchange.md` keeps the *why* in one sentence: use one shared vocabulary/tokenizer definition so every token resolves to the same ID at every site, because FedAvg averages embedding rows by position and an independently-built per-site mapping would silently blend unrelated tokens.

So the semantic-alignment concern you raised is still honored — via "one shared definition, pinned once" rather than agent-side hash reconciliation. Relevant commits: f3100cd36 (relax the eval to static-pin, drop the hash-reconciliation demand) and 4e225b51e (add the shared-mapping caution to `pytorch-model-exchange.md`).

chesterxgchen:
@greptile-apps

dev_tools/agent/skill_evals/nvflare-convert-lightning/files/external-data-lightning/model.py
```diff
@@ -0,0 +1,44 @@
...
+    def training_step(self, batch, batch_idx):
+        features, labels = batch
+        loss = F.cross_entropy(self(features), labels)
+        self.log("train_loss", loss)
+        return loss
+
+    def validation_step(self, batch, batch_idx):
+        features, labels = batch
+        loss = F.cross_entropy(self(features), labels)
+        self.log("val_loss", loss)
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in 201063d14 — added the labels.numel() == 0 guard raising an explicit ValueError to both training_step and validation_step in external-data-lightning/model.py, matching the sibling hello-lightning and vocab-lightning fixtures (team rule 783565ac). Empty per-site partitions now surface instead of silently producing nan loss.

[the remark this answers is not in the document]
chesterxgchen: Fixed in 201063d14 — added the labels.numel() == 0 guard raising an explicit ValueError to both training_step and validation_step in external-data-lightning/model.py, matching the sibling hello-lightning and vocab-lightning fixtures (team rule 783565ac). Empty per-site partitions now surface instead of silently producing nan loss.

dev_tools/agent/skill_evals/nvflare-convert-lightning/files/external-data-lightning/model.py
```diff
@@ -0,0 +1,44 @@
...
+
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in 201063d14 — added the labels.numel() == 0 guard raising an explicit ValueError to both training_step and validation_step in external-data-lightning/model.py, matching the sibling hello-lightning and vocab-lightning fixtures (team rule 783565ac). Empty per-site partitions now surface instead of silently producing nan loss.

dev_tools/agent/skill_evals/nvflare-convert-lightning/files/gpu-device-lightning/model.py
```diff
@@ -0,0 +1,34 @@
...
+    def __init__(self):
+        super().__init__()
+        self.layer = nn.Linear(4, 2)
+
+    def forward(self, features):
+        return self.layer(features)
+
+    def training_step(self, batch, batch_idx):
+        features, labels = batch
+        return F.cross_entropy(self(features), labels)
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in 7968f5cdd — added the labels.numel() == 0 guard raising an explicit ValueError to training_step in gpu-device-lightning/model.py, matching the sibling fixtures (hello-lightning, vocab-lightning, external-data-lightning). All four Lightning fixtures now guard against empty per-site partitions (team rule 783565ac).

chesterxgchen:
@greptile-apps

chesterxgchen:
@greptile-app

chesterxgchen:
@greptile-app

chesterxgchen:
@greptile-app

chesterxgchen:
@greptile-apps

chesterxgchen:
@greptile-apps

nvflare/tool/agent/inspector.py
```diff
@@ -439,11 +573,424 @@ def _rank_frameworks(state: InspectState) -> list[dict]:
...
+                    "reason": "incomplete_exported_job_marker_set",
+                }
+            )
+    return {"submit_ready_candidates": submit_ready, "nested_candidates": nested}
+
+
+def _conversion_state(state: InspectState, detected_framework: Optional[str], exported_job_info: dict) -> str:
+    if exported_job_info["submit_ready_candidates"]:
         return "exported_job"
     if state.job_py or state.sim_env_used:
```
holgerroth: **Concern 1 (inspector misclassification):** `flare_job` is still decided on name-only signals with no nvflare evidence: `state.job_py` is set for any file literally named `job.py` (line 382, before `ast.parse`, so even an unparseable file counts), and `sim_env_used` fires for any call named `SimEnv` or `*.SimEnv` (line 513) — a natural class name in RL/robotics codebases, and `job.py` is a common SLURM-launcher filename. Either way a plain training repo is classified `flare_job` / `flare_job_source`, the `not_converted` branch is never reached, and the conversion skill is never recommended.

The exported-job path was just fixed to require corroborating markers (`_exported_job_info`) — suggest the same treatment here: gate `job.py`/`SimEnv` on actual nvflare evidence (e.g. `state.flare_imports` or an `nvflare`-rooted call in the same file) before letting them determine `conversion_state`.

nvflare/tool/agent/inspector.py
```diff
@@ -253,6 +355,7 @@ def _inspect_file(path: Path, state: InspectState, max_file_bytes: int) -> None:
         return
     if _is_exported_job_marker(path):
         state.exported_job_markers.append(rel_path)
+        state.exported_job_marker_paths.append(path)
     if path.suffix not in PYTHON_SUFFIXES:
         return
```
holgerroth, line 361: **Concern 1 (inspector misclassification):** re: the `text = path.read_text(encoding="utf-8")` at line 372, just below this hunk — reading with `encoding="utf-8"` lets a UTF-8 BOM survive into `ast.parse`, which raises `SyntaxError` on U+FEFF — the file yields a `PYTHON_PARSE_ERROR` finding and zero framework/entry-point evidence. Reproduced end-to-end on the PR tree: a BOM-prefixed `train.py` with the repo's only torch model gives `detected_framework: null` and routes to `nvflare-orient`; the identical file without the BOM detects pytorch at 0.95 and recommends `nvflare-convert-pytorch`. Windows/Notepad-authored sources hit this routinely.

One-word fix: `encoding="utf-8-sig"` (a strict superset for reading — it only strips a leading BOM and decodes plain UTF-8 identically, so the `NON_UTF8_FILE` handling is unaffected). `_read_skill_frontmatter` already uses `utf-8-sig`, so this also removes an internal inconsistency.

nvflare/tool/agent/inspector.py
```diff
@@ -357,7 +492,20 @@ def _record_import(self, module: str, lineno: int) -> None:
...
+                continue
+            imports.add(f"{resolved_module}.{alias.name}" if resolved_module else alias.name)
+
     def _record_call(self, call_name: str, lineno: int) -> None:
+        # Generic FLARE / distributed / dynamic-dispatch signals only. Ranked
+        # framework activity (pytorch_call, lightning_trainer) and conversion
+        # signals (flare.patch) are recorded by framework detectors via on_call.
         if call_name.startswith("flare.") or call_name.startswith("nvflare."):
             self.state.flare_calls.add(call_name)
         if call_name in {"FedJob", "FLModel", "SimEnv"}:
```
holgerroth, line 515: **Concern 1 (inspector misclassification):** re: lines 513-516 just below — `export_support` fires for any attribute call ending in `.export` by any receiver in any scanned file — `torch.onnx.export(model, ...)`, YOLO's `model.export(...)`, etc. Combined with a name-only `job.py`, `_recommended_next_commands` then emits `python job.py --export --export-dir <job-dir>` with no check that job.py accepts those flags — and `recommended_next_commands` is a field consuming agents are expected to execute, so this ships a command that fails with an argparse error.

Ironically, the genuine FLARE idiom `export_job(...)` does **not** match `.endswith(".export")`, so the heuristic over-matches foreign exports while missing a real one. Suggest requiring the receiver to be FLARE evidence (e.g. a recipe/FedJob-typed call recorded by the detectors) or checking that job.py actually defines the `--export` argument.

dev_tools/agent/skills/checks/lints.py
```diff
@@ -377,17 +375,21 @@ def _load_skill_records(skills_root: Path, findings: list[LintFinding]) -> list[
         text = _read_bounded_text(skill_file) if skill_file.is_file() else None
         metadata = _try_parse_frontmatter(skill_file) if text is not None else {}
         text = text or ""
-        evals_path = child / "evals" / "evals.json"
+        skill_name = str(metadata.get("name") or child.name)
+        # Eval suites live outside the shipped skill tree, one dir per skill name.
+        evals_dir = evals_root / skill_name
```
holgerroth: **Concern 2 (lint engine vs its stated guarantees):** `skill_name` comes straight from frontmatter (`metadata.get("name")`) with no sanitization before this join. With pathlib, an absolute right operand discards `evals_root` entirely — so `name: /tmp/x` (or `name: ../../secrets`) relocates the eval dir outside the tree, and `_load_evals` reads the relocated `evals.json` at record-load time, **before any lint runs**. Worse, `_lint_fixtures`' escape check anchors on `record.evals_dir.resolve()` — the already-escaped root — so the `skill-fixture-path-escape` guard is measured against the attacker-chosen directory (a skill named `..` legitimizes fixtures anywhere under the evals parent). The `SKILL_NAME_RE` charset rule only fires later via the lazy `record.validation` and doesn't block the read; scoped runs that exclude the frontmatter lint never validate the name at all.

Given this engine is the admission gate for third-party skills, suggest a containment guard between name extraction and the join: validate against `SKILL_NAME_RE` here, or require `(evals_root / skill_name).resolve().is_relative_to(evals_root.resolve())`.

dev_tools/agent/skills/checks/lints.py
```diff
@@ -1440,22 +1362,39 @@ def _has_files(path: Path) -> bool:
...
     if root.is_symlink() or not root.is_dir():
         return
-    for dirpath, dirnames, filenames in os.walk(root, followlinks=False):
-        dirnames[:] = sorted(name for name in dirnames if not (Path(dirpath) / name).is_symlink())
-        for filename in sorted(filenames):
-            path = Path(dirpath) / filename
+    for current_dir, _dir_names, file_names in _walk_no_follow(root, excluded_dir_names):
+        for filename in file_names:
+            path = current_dir / filename
             if path.is_file():
```
holgerroth: **Concern 2 (lint engine vs its stated guarantees):** the PR description promises "bounded, no-follow regular-file reads and rejects symlinks anywhere in a skill tree", but `_walk_no_follow` only prunes symlinked **directories** — `path.is_file()` here is True for a symlink to a regular file, and consumers (`_lint_helper_scripts`, `_iter_skill_text_files`) read the target with no `is_symlink()` guard, so lints read content outside the skill tree (`scripts/helper.py -> /etc/passwd`) and can score text that isn't what actually ships. Similarly `_read_bounded_text` is a follow-symlinks stat-then-read (TOCTOU), and for scoped runs like `--check skill-md-size-lint`, `SkillRecord.validation` — the only place the no-symlink rule is enforced — is never computed.

The hardened reader already exists (`frontmatter.py`'s `_read_regular_skill_file`: lstat + `O_NOFOLLOW` + fstat/lstat dev-inode comparison), and `_read_runtime_text_file` shows the guard pattern — suggest routing these loaders through it, or at minimum adding `is_symlink()` checks in `_iter_files_no_follow` and `_read_bounded_text`.

dev_tools/agent/skills/checks/lints.py
```diff
@@ -1348,20 +1276,14 @@ def _command_drift_message(command: str, *, check_flags: bool = True, allow_plan
...
-            if token.startswith("--"):
-                flag = token.split("=", 1)[0]
-                if flag not in allowed_flags:
-                    return f"unknown flag '{flag}' for 'nvflare {command_key}' in '{command}'"
+    for token in tokens:
+        if token.startswith("--"):
+            flag = token.split("=", 1)[0]
+            if flag not in allowed_flags:
+                return f"unknown flag '{flag}' for 'nvflare {command_key}' in '{command}'"
     return None
```
holgerroth, line 1303: **Concern 2 (lint engine vs its stated guarantees):** re: `_command_tokens` just below (lines 1290-1294) — the command-drift lint fails open on exactly the inputs it exists to catch. Reproduced on the PR tree: the prose line `Run nvflare frobnicate to reset the server's state.` passes the lint — the apostrophe makes `shlex.split` raise `ValueError`, `_command_tokens` returns `[]`, `_command_drift_message` returns `None`, and `_lint_command_drift` just `continue`s with no finding, warning, or skipped-check record. Likewise `` `nvflare $(cat cmd.txt) frobnicate --bad` `` truncates at the unsafe token to `['nvflare']` and passes, and the `continue` after a backtick match skips plain-text commands later on the same line. Since the extractor over-captures prose to end-of-line, an ordinary apostrophe silently disables the check for that line.

Suggest a whitespace-split fallback when `shlex` fails, and emitting a "could not parse command" finding (or at least a skipped-check entry) instead of returning `[]` — fail-closed is the right default for an admission gate.

holgerroth (commented):
Thanks — this is a well-structured milestone, and the recent `ea8e4ede9` commit already fixed the biggest inspector issue we found (a stray dataset `meta.json` no longer classifies a repo as an exported job, now that submit-ready detection requires a `config_fed_*` sibling at the root). Two main concerns remain, verified against the current head (several with live reproductions); details inline.

**1. The inspector still misclassifies ordinary repos on name-only signals.** Any `job.py` or any user-defined `SimEnv` class still flips `conversion_state` to `flare_job` with no nvflare evidence required, a UTF-8 BOM silently zeroes framework detection, and any `*.export(...)` call produces a `python job.py --export` recommendation that fails. Since `inspect` output is the routing contract that skills (and `nvflare-orient`) follow, these directly suppress or misdirect the conversion workflow on common inputs.

**2. The lint engine doesn't yet meet its own stated security/enforcement guarantees** ("bounded, no-follow regular-file reads and rejects symlinks anywhere in a skill tree"; deterministic admission gate): a frontmatter `name` can relocate the eval root outside the tree, several loader paths follow symlinks (the hardened reader in `frontmatter.py` isn't used by them), and the command-drift lint fails open on ordinary prose. Since `npx skills add` installs whatever is in git, this engine plus CI is the only admission gate — so fail-open paths undermine the design's security story.

.githooks/pre-push
```diff
@@ -0,0 +1,28 @@
...
+
+repo_root="$(git rev-parse --show-toplevel)"
+
+# Nothing to check if this repo has no skills root.
+if [ ! -d "$repo_root/skills" ]; then
+    exit 0
+fi
+
+echo "pre-push: running agent skill lint (python -m dev_tools.agent.skills.checks)..."
+if ! python3 -m dev_tools.agent.skills.checks --skills-root "$repo_root/skills"; then
```
holgerroth, line 30: Separate, smaller concern than the two in my review — developer-workflow friction:

The hook runs bare `python3` from PATH, but the checks package hard-requires PyYAML (`frontmatter.py` does `import yaml`). PyYAML is an nvflare install dependency, which only helps if the hook happens to pick up the project venv — git hooks often don't (unactivated shell, GUI clients, system interpreter first on PATH). Reproduced on a macOS machine with a Homebrew `python3` lacking PyYAML: the lint exits 1 and every push is blocked, despite the comment at line 8 claiming "fast and dependency-light" (`.githooks/README.md` also doesn't mention the prerequisite).

Compounding it, the error message is misleading: the `try: from .cli import main / except ImportError: from cli import main` fallback chains in `__main__.py`, `cli.py`, and `lints.py` catch the `ModuleNotFoundError('yaml')` raised inside the relative import and divert into the bare-import path, so the terminal error the developer reads is `No module named 'cli'` — the real cause (`No module named 'yaml'`) is buried at the top of a three-deep chained traceback. Easy to burn time debugging a nonexistent packaging problem.

Suggestions (independent of the other fixes):
1. Narrow the fallbacks so they only catch the script-vs-package case, e.g. `except ImportError as e: if e.name not in ("dev_tools", ...): raise` — then the yaml error surfaces directly.
2. Have the hook fail with an actionable message (e.g. probe `python3 -c "import yaml"` first and print "PyYAML missing — activate your nvflare dev environment or pip install pyyaml"), and mention the prerequisite in `.githooks/README.md`.
3. Alternatively, drop the yaml dependency: the frontmatter grammar the validator accepts is constrained enough that a stdlib-only parser would keep the "dependency-light" claim true — though narrowing the excepts is the smaller change.

nvflare/tool/agent/inspector.py
chesterxgchen: Fixed in e394ff7f4: `flare_job` classification now requires a corroborating nvflare import, so a name-only `job.py` (SLURM launcher) or a local `SimEnv()` call no longer classifies a plain training repo as an existing FLARE job — it routes to conversion instead. Regression tests added.

nvflare/tool/agent/inspector.py
chesterxgchen: Fixed in e394ff7f4: sources are now read with `utf-8-sig`, so a leading BOM no longer reaches `ast.parse` as U+FEFF and wipes out framework detection. Regression test added; this also matches `_read_skill_frontmatter`, removing the internal inconsistency.

nvflare/tool/agent/inspector.py
chesterxgchen: Fixed in e394ff7f4: the `python job.py --export` next-command is now gated on corroborating nvflare evidence, so an over-broad `.export` match (torch.onnx.export, YOLO model.export) can no longer ship a command that fails argparse on an unrelated repo. Regression test added.

dev_tools/agent/skills/checks/lints.py
chesterxgchen: Fixed in e394ff7f4: the eval-dir name is validated against `SKILL_NAME_RE` before the join and falls back to the trusted filesystem directory name when it fails, so an absolute or `..` frontmatter `name` can no longer relocate eval loading outside the tree. The frontmatter lint still reports the invalid name separately. Regression tests added.

dev_tools/agent/skills/checks/lints.py
chesterxgchen: Fixed in e394ff7f4: the no-follow readers now reject symlinked files, not just symlinked directories — `_iter_files_no_follow` and `_iter_skill_text_files` skip `is_symlink()` entries and `_read_bounded_text` returns None for a symlink, so a `references/x.md -> /etc/passwd` symlink is never read/scored. Regression tests added.

dev_tools/agent/skills/checks/lints.py
chesterxgchen: Fixed in e394ff7f4: a `shlex.split` ValueError (e.g. an apostrophe in prose) now falls back to a whitespace split instead of returning `[]`, so the `nvflare <subcommand>` tokens are still extracted and drift is caught — fail-closed. Regression tests added (drifted-in-apostrophe-line fires; undrifted still passes).

.githooks/pre-push
chesterxgchen: Fixed in e394ff7f4: the script-vs-package import fallbacks now re-raise when `e.name is not None` (verified: a no-parent relative import has `e.name is None`; a missing dep has `e.name == "yaml"`), so a missing PyYAML surfaces its real error. The pre-push hook probes `import yaml` first with an actionable message, documented in `.githooks/README.md`.

skills/nvflare-orient/SKILL.md
```diff
@@ -1,9 +1,22 @@
 ---
 name: nvflare-orient
-description: "Route ambiguous NVFLARE requests by inspecting the local project, checking readiness, and recommending the next specific FLARE workflow skill without editing files."
-min_flare_version: "2.8.0"
-blast_radius: read_only
-skill_version: "0.1.0"
+description: "Route open-ended or ambiguous NVFLARE requests by inspecting the local project and recommending one specific workflow skill without editing files; do not use for an explicit conversion request, even when its framework still needs detection."
+metadata:
+  author: "Chester Chen <chesterc@nvidia.com>"
```
holgerroth: Suggestion (non-blocking, applies to all five skills): follow the frontmatter conventions of the official NVIDIA skills catalog, https://github.com/NVIDIA/skills/tree/main/skills.

**1. `author` should be the team, not an individual.** Every skill in that catalog uses a team/org identity — `author: NVIDIA` (accelerated-computing-cudf), `author: "CUDA-Q Team <cuda-quantum@nvidia.com>"`, `author: NVIDIA cuOpt Team`, `author: "NVIDIA AI-Q Blueprint Team <aiq-blueprint@nvidia.com>"` — none uses a personal name/email. For these skills:

```yaml
author: "NVIDIA FLARE Team <federatedlearning@nvidia.com>"
```

Rationale beyond consistency: once `npx skills add` copies the frontmatter into users' agent environments, `author` functions as the support contact detached from the repo — a personal inbox goes stale when people change teams, and it ships an individual's email into every install. Individual attribution stays in git history/CODEOWNERS. This also collapses the current redundancy with the separate `team: nvflare` metadata field. The frontmatter lint's NVCARPS check ("name plus NVIDIA email") is satisfied by a team alias as written, so no lint change is required — though pinning the team format in the validator would prevent regressions.

**2. Add a top-level `license:` field.** Catalog skills all declare one (`license: Apache-2.0`, cudf uses `CC-BY-4.0 AND Apache-2.0`). Right now the NVFlare skills carry no license statement, and after `npx skills add` copies them out of the repo they're no longer under the repo's LICENSE file umbrella:

```yaml
license: Apache-2.0
```

**3. Minor:** the catalog uses `version` where these skills use `metadata.skill_version` — worth aligning if these skills are eventually destined for the NVIDIA/skills catalog (which the `npx skills add NVIDIA/<skills-repo>` distribution model suggests), to avoid a rename migration later.

chesterxgchen: Adopted all three in 25bc3114b: `author` is now `NVIDIA FLARE Team <federatedlearning@nvidia.com>` in all five skills (the redundant `metadata.team` field is dropped), every skill declares a top-level `license: Apache-2.0`, and `metadata.skill_version` moved to the catalog's top-level `version` field. Also pinned the team-identity author format in the frontmatter lint (`skill-author-format-invalid`) as suggested, so it can't regress.

holgerroth (approved):
This PR is now in excellent shape — all of my feedback has been addressed.

PR #4843, opened 2026-06-30
docs: document recipe api

YuanTingHsieh:
## Summary
- add a canonical public Recipe API contract page covering Recipe, ExecEnv, Run, public/internal boundaries, and versioning
- add a recipe catalog JSON schema for the successful `nvflare recipe list --format json` response
- link the contract/schema from the data scientist guide, Recipe guide, and CLI docs
- update the job logs CLI docs to use the public Recipe log-streaming helper instead of internal generated job access

## Validation
- `python3 -m json.tool docs/schemas/recipe_catalog.schema.json`
- `git diff --check`

Full Sphinx docs build was not run locally because `sphinx` is not installed in this environment.


docs/user_guide/data_scientist_guide/recipe_api.rst
```diff
@@ -0,0 +1,234 @@
...
+   indicate that sites are connected or replace the execution environment.
+
+``set_recipe_meta(recipe, key, value)``
+   Set selected generated job metadata by ``JobMetaKey``. Runtime and submission
+   metadata managed by NVFlare internals are not user-settable through this
+   helper.
+
+Repeated metadata calls for the same key replace that key's previous helper
+value. Different metadata keys accumulate. Helper-provided values take
+precedence over constructor defaults for the same generated-job surface.
```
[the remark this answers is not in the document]
YuanTingHsieh: Resolved in two parts. `set_per_site_config` and `recipe.configured_sites()` now exist on this branch after merging main (a5a6eba41): `nvflare/recipe/utils.py:75` (exported from `nvflare.recipe`) and `nvflare/recipe/spec.py:218`. `set_recipe_meta` lands with #4849, which finalizes its semantics (allowlist in `USER_SETTABLE_JOB_META_KEYS`, per-key shape validation) — this docs page was updated in 4fde1be44 to match those final semantics, and #4849 should merge before this PR so the documented surface exists at merge time.

docs/user_guide/data_scientist_guide/recipe_api.rst
```diff
@@ -0,0 +1,234 @@
...
+
+``set_per_site_config(recipe, config)``
+   Provide site-keyed, recipe-specific configuration. Each concrete recipe
+   interprets the site dictionaries for its own workflow.
+
+``recipe.configured_sites()``
+   Return top-level site names from helper-provided per-site config when
+   present. For backward compatibility, recipes may return site names from
+   legacy constructor ``per_site_config`` when available. This method does not
+   indicate that sites are connected or replace the execution environment.
```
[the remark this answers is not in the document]
YuanTingHsieh: Resolved — both APIs now exist on this branch after merging main: `set_per_site_config` at `nvflare/recipe/utils.py:75` (exported from `nvflare.recipe`) and `Recipe.configured_sites()` at `nvflare/recipe/spec.py:218`. The comment predates the merge from main.

docs/user_guide/data_scientist_guide/recipe_api.rst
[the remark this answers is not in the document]
YuanTingHsieh: Valid at the time — `set_recipe_meta` is added by #4849 (open), which also finalizes its contract: allowlist defined as `nvflare.apis.job_def.USER_SETTABLE_JOB_META_KEYS` (`RESOURCE_SPEC`, `JOB_LAUNCHER_SPEC`, `SCOPE`, `CUSTOM_PROPS`), per-key value shape validation, and rejection of constructor-owned keys and `study`. This page was updated in 4fde1be44 to document exactly those semantics. Merge order: #4849 first, then this PR.

docs/user_guide/data_scientist_guide/recipe_api.rst
```diff
@@ -0,0 +1,234 @@
...
+``recipe.execute`` and ``recipe.run`` return a ``Run`` object when the job is
+submitted. ``Run`` exposes:
+
+``run.get_job_id()``
+   Return the environment job ID.
+
+``run.get_status()``
+   Return the latest status when available.
+
+``run.get_result(timeout=0.0, cleanup=True)``
```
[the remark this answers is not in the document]
YuanTingHsieh: Fixed — corrected to `run.get_result(timeout=0.0, clean_up=True)`, matching the implementation at `nvflare/recipe/run.py:77` (follow-up commit on this branch). The `stop(clean_up=False)` entry already matched `spec.py`.

docs/user_guide/data_scientist_guide/recipe_api.rst
```diff
@@ -0,0 +1,242 @@
...
+   ``mandatory_clients``), and ``study``, which the server assigns from the
+   admin session at submission. See the Recipe Metadata section of the Recipe
+   guide for per-key value shapes and examples.
+
+Repeated metadata calls for the same key replace that key's previous helper
+value. Different metadata keys accumulate. Keys that overlap a dedicated
+constructor field are rejected rather than given precedence; for accepted
+keys, the helper value is what appears in the generated ``meta.json``
+(metadata merges last), and overriding per-site resource specs registered via
+``add_resource_spec`` emits a warning.
```
holgerroth: This unconditional claim overstates the actual behavior: the warning only fires for resource specs already registered at the moment `set_recipe_meta` is called (`nvflare/recipe/utils.py` guards it with `if key is JobMetaKey.RESOURCE_SPEC and job_config.resource_specs:`, and the code comment there notes specs registered *after* the call are overridden at export without a warning).

A user who calls `set_recipe_meta(RESOURCE_SPEC, ...)` first and `add_resource_spec` later will rely on a warning that never fires while their specs are silently dropped from `meta.json`. Suggest reusing the precise phrasing already in `job_recipe.rst`: "a warning is emitted for specs already registered when the helper is called, but specs added afterwards are overridden without one."

docs/user_guide/data_scientist_guide/recipe_api.rst
```diff
@@ -0,0 +1,242 @@
...
+
+You can rely on documented concrete recipe constructor parameters, documented
+Recipe methods, documented helper functions, ``ExecEnv`` methods, ``Run``
+methods, and documented JSON fields from ``nvflare recipe`` commands.
+
+The following are internal details and should not be used in recipe scripts:
+
+* private attributes, including attributes whose names start with ``_``;
+* generated job internals and nested generated job objects;
+* the internal generated job attribute of a recipe;
```
holgerroth, line 226: This contradicts `job_recipe.rst`, which this PR links to the new page: its Recipe Metadata note still affirmatively documents ``recipe.job.job.add_resource_spec(...)`` as a supported way to populate per-site resource specs, with no internal/discouraged caveat. So the two linked pages give opposite guidance on whether the generated-job attribute is usable.

The PR already reconciles the equivalent case in `job_cli.rst` (replacing `recipe.job.to_clients/to_server` with `recipe.enable_log_streaming()`), so this looks like an oversight — either update the `job_recipe.rst` note to route through `set_recipe_meta`, or soften this bullet to allow the documented exception.

docs/user_guide/nvflare_cli/recipe_command.rst
```diff
@@ -234,6 +234,11 @@ Behavior notes:
 The CLI prints a human-readable table in text mode and also emits the machine-
 readable result envelope.
 
+The successful ``nvflare recipe list --format json`` response is the recipe
+catalog JSON contract, sometimes referred to as ``catalog.json``. It is
```
holgerroth: "sometimes referred to as ``catalog.json``" asserts a pre-existing convention that doesn't exist: the term appears nowhere on `main` (no code, docs, or artifact), and the CLI never writes such a file — `recipe_cli.py` only prints the envelope to stdout. Same for "the ``catalog.json`` contract" in `recipe_api.rst`.

Tooling authors may search for or hardcode a `catalog.json` file that is never produced. Suggest dropping the name (e.g. "the recipe catalog JSON contract") or explicitly introducing it as new terminology rather than claiming existing usage.

holgerroth (commented):
Reviewed the new contract page against the implementation on `main` — the documented Recipe/`ExecEnv`/`Run` signatures, defaults, metadata-key rules, and the catalog JSON schema all match the code exactly, and the content removed from `available_recipes.rst` is covered elsewhere. Three doc-accuracy concerns inline.

docs/user_guide/data_scientist_guide/recipe_api.rst
YuanTingHsieh: Fixed in 182300536 — adopted your suggested phrasing: the page now states the warning fires only for specs already registered when `set_recipe_meta` is called, and specs added afterwards are overridden without one, matching the guard in `nvflare/recipe/utils.py` and the wording in `job_recipe.rst`.

docs/user_guide/data_scientist_guide/recipe_api.rst
YuanTingHsieh: Fixed in 182300536 — took the first option, consistent with the `job_cli.rst` reconciliation: the `job_recipe.rst` note no longer affirmatively documents `recipe.job.job.add_resource_spec(...)`. It now describes per-site specs on the underlying job as an internal lower-level path with 'prefer `set_recipe_meta` in recipe scripts' and a `:ref:` to the contract page, so both linked pages give the same guidance.

docs/user_guide/nvflare_cli/recipe_command.rst
YuanTingHsieh: Fixed in 182300536 — dropped the `catalog.json` name entirely rather than introducing it as new terminology (the CLI never writes a file, so a filename would invite exactly the hardcoding you describe). Both pages now describe it plainly: the `nvflare recipe list --format json` output has a stable, documented structure described by `recipe_catalog.schema.json`. Also rewrote the remaining 'contract' jargon in these pages in plain language while at it.

holgerroth: Thanks — the RST pages look good now. One leftover: the schema file itself still carries the old name in its own description (`docs/schemas/recipe_catalog.schema.json` line 5): `"description": "Successful catalog.json response emitted by ..."`. Since the schema is the artifact tooling authors will actually read, suggest e.g. `"Successful response emitted by `nvflare recipe list --format json`."`

YuanTingHsieh: Good catch — fixed in ac922c5c4 with your suggested wording: the schema description now reads "Successful response emitted by `nvflare recipe list --format json`." Checked the rest of the schema file for the stale name too: the remaining "catalog" occurrences (`$id` URL and the `recipe_catalog_entry` def) are the schema's own filename/def naming rather than claims about an emitted artifact, so they were left as-is.

holgerroth (approved):
Makes sense

PR #4844, opened 2026-06-30
Avoid spurious stream ACK teardown errors

nvidianz:
## Summary

- mark ACKs emitted for completed reliable streams as optional
- log completed-stream ACK delivery failures at debug level while preserving error-level logging for active streams
- add regression coverage for final ACK optionality and teardown-time delivery failure logging

## Root cause

Reliable messaging keeps a completed receive task available briefly so it can ACK retried chunks. During external-process job teardown, a final or repeated ACK can race with removal of the target subprocess cell and its connection. The ACK was sent as a non-optional message, so CoreCell reported `cannot forward req: no path` at ERROR level even though the stream and job had already completed successfully.

This change uses the existing optional-message logging path only after the receive task is complete. In-flight ACK failures remain ERROR-level signals.

## Impact

Normal job teardown no longer emits a misleading server ERROR for a harmless trailing stream ACK. Active stream reliability and ACK state handling are unchanged.

Tracks NVBugs 6389772.

## Validation

- `tests/unit_test/fuel/f3/streaming/byte_receiver_test.py`: 23 passed
- related streaming and CoreCell tests: 78 passed
- complete `tests/unit_test/fuel/f3/streaming` suite: 211 passed
- scoped Black, isort, and Flake8 checks passed
- `git diff --check` passed


YuanTingHsieh (dismissed):

nvflare/fuel/f3/streaming/byte_receiver.py
```diff
@@ -452,6 +452,9 @@ def _get_ack_offset(self):
         return self.received_offset if self.completed else self.offset
 
     def _send_ack(self, offset, seq):
+        # A completed stream may outlive its target during teardown; ACK delivery no longer affects its result.
+        optional = self.completed
```
pcnudde: Could we keep the first final ACK non-optional? `stop()` sets `self.completed = True` before calling `_send_ack()`, so this also classifies the initial final ACK as optional. On the sender, `_prepare_reliable_stop()` leaves the stream pending until `handle_ack()` removes the final sequence from `pending_messages`; delivery of this ACK therefore still affects the sender's result. If the reverse path fails while the sender is alive, both CoreCell and this method log only at DEBUG while the sender retries and may eventually time out. Please restrict optional/debug handling to re-ACKs after a final ACK has already succeeded—for example, using `seq_ack`/`offset_ack`, or passing an explicit flag from the completed-duplicate path.

pcnudde (commented):
One reliability/observability issue found; please see the inline comment.

nvflare/fuel/f3/streaming/byte_receiver.py
nvidianz: Addressed in 2fb1e3f0a. `_send_ack()` now marks an ACK optional only when the receive task is complete and the ACK's sequence/offset are at or behind state whose ACK send already succeeded. The initial final ACK remains non-optional, and retries after a failed final ACK also remain non-optional/ERROR-visible until one succeeds. Added regression coverage for the failed-initial-ACK retry path, successful duplicate re-ACKs, returned send errors, and exceptions. Validation: 25 byte-receiver tests, 213 streaming tests, and 5 CoreCell routing tests passed; scoped Black/isort/Flake8 and `git diff --check` also passed.

YuanTingHsieh (approved):

chesterxgchen:
PR #4844 review — three perspectives

Consensus verdict: approve (architect and principal engineer: approve-with-comments; security: approve). No blocking or major findings. The PR is already merged; everything below is follow-up material, not change requests.

All three reviewers independently verified the core mechanism is sound. optional=True in fire_and_forget only stamps MessageHeaderKey.OPTIONAL into the headers (core_cell.py:1507–1514) — routing, retry, and delivery are untouched. Because the header travels with the message, the forwarding cell's cannot forward req: no path ERROR (core_cell.py:1757 → log_messaging_error) is demoted too — which is the only mechanism that could fix the reported symptom, since that log fires on a remote hop the receiver can't reach any other way. The predicate is conservative: seq_ack/offset_ack advance only on successful send, so the first final ACK is always required (seq_ack=-1 initially handles the empty-stream seq=0/offset=0 case correctly), a failed first final ACK keeps all retries at ERROR, and the unlocked read of self.completed can only err toward fail-loud. Sender guarantees are preserved — re-ACKs are still transmitted, and a truly-lost ACK still surfaces via the sender's own retry-exhaustion ERROR.

Findings worth acting on (follow-ups)

1. _send_error has the identical teardown race, still ERROR (architect, minor — byte_receiver.py:334–352). Retained failed reliable tasks re-notify on every retried chunk with no optional=, so the same spurious no path ERROR survives for failed streams racing teardown. Same treatment applies.
2. Non-reliable streams not covered (principal, minor — byte_receiver.py:279–286). For non-reliable streams completed stays False, so their trailing flow-control ACK at teardown still ERRORs. If NVBug 6389772 was ever observed on non-reliable streams, the symptom persists there.
3. Peer-controlled task TTL is unbounded (security, pre-existing, minor — byte_receiver.py:207–210). _handle_new_stream trusts the sender's RETRY_TIMEOUT/RETRY_WAIT headers with no clamp; a malicious peer can pin completed RxTasks in rx_task_map indefinitely across arbitrary SIDs, and a non-numeric value makes float() raise inside process_chunk. Worth clamping and validating. Not introduced by this PR.

Accepted trade-offs (no action)

- Residual window: if the peer disappears between the last chunk and the first final ACK, that ACK is still non-optional and ERRORs — deliberate (can't distinguish teardown from a wedged sender), so bug-class reports may not go fully to zero.
- already_acked means "accepted by first hop," not "delivered" — on a multi-hop route a mid-route drop could make subsequent re-ACK failures DEBUG-only; visibility is preserved by the sender-side retry timeout. Principal suggests a one-line code comment documenting this.
- Check-then-act race: two concurrent sends of the same (seq, offset) can both classify as required, leaving one spurious ERROR possible — window is tiny and the fix (lock across network I/O) would be worse.
- Security: the demoted line is noise, not an actionable detection signal; re-ACK reflection is 1:1 (no amplification); the OPTIONAL header being peer-settable on a peer's own messages is a pre-existing property of the mechanism (already used by goodbye broadcasts, cell_pipe END/ABORT/HEARTBEAT); no info leak in the changed log lines.

Test coverage was judged good — the three load-bearing behaviors (first-final-ACK required, re-ACK optional on both failure paths, end-to-end via process_chunk) are all pinned.

nvidianz:
@chesterxgchen The three actionable findings from your review (plus the suggested first-hop doc comment) are addressed in follow-up PR #4861: `_send_error` re-notifications get the same optional/DEBUG treatment as re-ACKs, non-reliable streams are covered (post-completion ACKs advisory, mid-stream flow-control ACKs still required — verified the non-reliable sender never waits past send-completion), and the peer-supplied retry-window headers are validated and clamped to 1h. The accepted trade-offs from your review are unchanged.

PR #4860, opened 2026-07-02
Raise unit test coverage above 60%

pcnudde:
## Summary

- add focused unit coverage for the five originally targeted runtime modules: `core_cell.py`, `job_cli.py`, `net_agent.py`, `client_runner.py`, and `server_engine.py`
- add a second high-yield set for workflow communication, hub/IPC execution, cellnet statistics and diagnostics, CMI, and edge configuration/dispatch/assessment paths
- production fixes surfaced by review, all in `WFCommClient`: `send_and_wait` called `get_return_code()` on the dict returned by `broadcast_and_wait`, raising `AttributeError` on every client-side `send()` — it now unwraps the single-target reply and returns the replies dict, matching the broadcast/relay contract; `broadcast_and_wait` now scopes callback firing and reply reporting to the client tasks created by the current call, so `send()` failover no longer re-fires `after_task_sent_cb` for already-tried targets or leaks their stale results; `_make_error_reply` no longer aliases one reply object across all targets. All other changed files are under `tests/`.

## Coverage impact

Codecov's exact reports for the current `main` and PR head show:

- **before:** 56.97% (52,580 / 92,281 lines)
- **after:** 60.18% (55,540 / 92,281 lines)
- **gain:** 2,960 covered lines and 3.21 percentage points

Initial target details from the line-level analysis:

| Module | Codecov before | Projected after | Newly covered lines |
| --- | ---: | ---: | ---: |
| `core_cell.py` | 54.4% | 61.6% | 91 |
| `job_cli.py` | 72.7% | 76.0% | 58 |
| `net_agent.py` | 23.2% | 60.3% | 232 |
| `client_runner.py` | 14.0% | 56.5% | 206 |
| `server_engine.py` | 38.1% | 55.3% | 112 |

## Test approach

The tests exercise deterministic routing, validation, serialization, configuration, lifecycle, task-runner, auxiliary-message, cell-message, and edge-model paths with small fakes and mocks. They avoid real networking, subprocesses, and timing-dependent integration behavior.

## Validation

- all 15 changed test modules with coverage collection: **288 passed**
- broader affected-area unit suites: **674 passed**
- `PATH="$PWD/.venv/bin:$PATH" ./runtest.sh -s --skip-install`: **passed**
- all **18** PR checks, including the ten Ubuntu/Python unit-test matrices, coverage, CodeQL, style, wheel build, and Greptile: **passed**

### DIFF
--- a/tests/unit_test/apis/impl/wf_comm_client_test.py
+++ b/tests/unit_test/apis/impl/wf_comm_client_test.py
@@ -18,16 +18,15 @@ import pytest
 
 from nvflare.apis.client import Client
 from nvflare.apis.controller_spec import ClientTask, Task, TaskCompletionStatus
-from nvflare.apis.fl_constant import ReservedKey, ReturnCode, SiteType
+from nvflare.apis.fl_constant import ReturnCode, SiteType
 from nvflare.apis.fl_context import FLContext
 from nvflare.apis.impl.wf_comm_client import WFCommClient
 from nvflare.apis.shareable import Shareable, make_reply
+from tests.unit_test.fl_context_helper import make_fl_context
 
 
 def _context(engine):
-    fl_ctx = FLContext()
-    fl_ctx.set_prop(ReservedKey.ENGINE, engine, private=True, sticky=False)
-    return fl_ctx
+    return make_fl_context(engine=engine)
 
 
 def _engine():
@@ -196,10 +195,18 @@ def test_send_and_relay_delegate_in_order():
     comm = WFCommClient()
     comm._validate_target = MagicMock()
     ok = make_reply(ReturnCode.OK)
-    comm.broadcast_and_wait = MagicMock(side_effect=[make_reply(ReturnCode.ERROR), ok])
+    comm.broadcast_and_wait = MagicMock(
+        side_effect=[{"site-1": make_reply(ReturnCode.ERROR)}, {"site-2": ok}],
+    )
 
-    assert comm.send(task, fl_ctx, ["site-1", "site-2"]) is ok
+    assert comm.send(task, fl_ctx, ["site-1", "site-2"]) == {"site-2": ok}
     assert comm.broadcast_and_wait.call_count == 2
+
+    comm.broadcast_and_wait.reset_mock()
+    error_replies = [{"site-1": make_reply(ReturnCode.ERROR)}, {"site-2": make_reply(ReturnCode.ERROR)}]
+    comm.broadcast_and_wait.side_effect = error_replies
+    failed = comm.send(task, fl_ctx, ["site-1", "site-2"])
+    assert failed["site-2"].get_return_code() == ReturnCode.ERROR
 
     comm.broadcast_and_wait.reset_mock()
     comm.broadcast_and_wait.side_effect = [{"site-1": ok}, {"site-2": ok}]
--- a/tests/unit_test/apis/impl/wf_comm_client_test.py
+++ b/tests/unit_test/apis/impl/wf_comm_client_test.py
@@ -1,0 +1,217 @@
+# Copyright (c) 2026, NVIDIA CORPORATION.  All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from unittest.mock import MagicMock, patch
+
+import pytest
+
+from nvflare.apis.client import Client
+from nvflare.apis.controller_spec import ClientTask, Task, TaskCompletionStatus
+from nvflare.apis.fl_constant import ReservedKey, ReturnCode, SiteType
+from nvflare.apis.fl_context import FLContext
+from nvflare.apis.impl.wf_comm_client import WFCommClient
+from nvflare.apis.shareable import Shareable, make_reply
+
+
+def _context(engine):
+    fl_ctx = FLContext()
+    fl_ctx.set_prop(ReservedKey.ENGINE, engine, private=True, sticky=False)
+    return fl_ctx
+
+
+def _engine():
+    engine = MagicMock()
+    engine.all_clients = [Client("site-1", "token-1"), Client("site-2", "token-2")]
+    engine.validate_targets.return_value = (engine.all_clients, [])
+    return engine
+
+
+def _ok_reply(peer_name):
+    reply = make_reply(ReturnCode.OK)
+    peer_ctx = FLContext()
+    peer_ctx.set_prop("peer", peer_name, private=False, sticky=False)
+    reply.set_peer_context(peer_ctx)
+    return reply
+
+
+def test_broadcast_and_wait_processes_callbacks_filters_and_replies():
+    engine = _engine()
+    engine.send_aux_request.return_value = {
+        "site-1": _ok_reply("site-1"),
+        "site-2": make_reply(ReturnCode.EXECUTION_EXCEPTION),
+    }
+    engine.send_aux_request.return_value["site-2"].set_peer_context(FLContext())
+    callbacks = {name: MagicMock() for name in ("before", "after", "result", "done")}
+    task = Task(
+        "train",
+        Shareable(),
+        timeout=0,
+        before_task_sent_cb=callbacks["before"],
+        after_task_sent_cb=callbacks["after"],
+        result_received_cb=callbacks["result"],
+        task_done_cb=callbacks["done"],
+        secure=True,
+    )
+    comm = WFCommClient(max_task_timeout=30)
+    comm.fire_event = MagicMock()
+    comm.fire_event_with_data = MagicMock()
+
+    with patch("nvflare.apis.impl.wf_comm_client.apply_filters", side_effect=lambda *args: args[1]):
+        with patch("nvflare.apis.impl.wf_comm_client.delete_msg_root") as delete_msg_root:
+            replies = comm.broadcast(task, _context(engine), targets=engine.all_clients)
+
+    assert task.timeout == 30
+    assert set(replies) == {"site-1", "site-2"}
+    assert replies["site-1"].get_return_code() == ReturnCode.OK
+    assert replies["site-2"].get_return_code() == ReturnCode.ERROR
+    assert callbacks["before"].call_count == 2
+    assert callbacks["after"].call_count == 2
+    callbacks["result"].assert_called_once()
+    callbacks["done"].assert_called_once()
+    delete_msg_root.assert_called_once()
+    assert engine.send_aux_request.call_args.kwargs["secure"] is True
+
+
+def test_broadcast_uses_all_clients_and_rejects_invalid_targets():
+    engine = _engine()
+    engine.send_aux_request.return_value = {}
+    comm = WFCommClient()
+    comm.fire_event = MagicMock()
+    comm.fire_event_with_data = MagicMock()
+
+    with patch("nvflare.apis.impl.wf_comm_client.apply_filters", side_effect=lambda *args: args[1]):
+        comm.broadcast_and_wait(Task("train", Shareable(), timeout=1), _context(engine), targets=None)
+
+    assert len(engine.send_aux_request.call_args.kwargs["targets"]) == 2
+
+    engine.validate_targets.return_value = ([], ["missing"])
+    with patch("nvflare.apis.impl.wf_comm_client.apply_filters", side_effect=lambda *args: args[1]):
+        with pytest.raises(ValueError, match="invalid target"):
+            comm.broadcast_and_wait(Task("train", Shareable(), timeout=1), _context(engine), ["missing"])
+
+
+def test_broadcast_returns_filter_error_replies():
+    engine = _engine()
+    comm = WFCommClient()
+    comm.log_exception = MagicMock()
+    comm.fire_event_with_data = MagicMock()
+
+    with patch("nvflare.apis.impl.wf_comm_client.apply_filters", side_effect=RuntimeError("filter failed")):
+        replies = comm.broadcast_and_wait(Task("train", Shareable()), _context(engine), ["site-1"])
+
+    assert replies["site-1"].get_return_code() == ReturnCode.TASK_DATA_FILTER_ERROR
+
+
+def test_broadcast_handles_result_filter_and_task_done_errors():
+    engine = _engine()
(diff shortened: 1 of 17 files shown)


tests/unit_test/fuel/f3/cellnet/net_agent_test.py
```diff
@@ -0,0 +1,300 @@
...
+
+
+def test_stats_and_config_accessors_handle_success_and_failure():
+    agent, cell = _agent()
+    for method in (agent.get_msg_stats_table, agent.show_pool):
+        cell.send_request.return_value = _reply(payload={"rows": [1]})
+        args = ("site-2", "pool", "count") if method == agent.show_pool else ("site-2", "count")
+        assert method(*args) == {"rows": [1]}
+        cell.send_request.return_value = _reply(ReturnCode.TIMEOUT, error="late")
+        assert method(*args) == "timeout: late" if method == agent.show_pool else "error: timeout"
```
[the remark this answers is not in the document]
pcnudde: Fixed in 9e7e5d18d. The expected error string is now computed before the assertion, so both accessor branches perform an actual comparison. I also removed the unrelated hard-coded handler count called out in the review summary.

pcnudde:
Addressed the review and CI findings in 9e7e5d18d:

- fixed the conditional assertion so the `get_msg_stats_table` error path is actually checked
- removed the brittle hard-coded request-handler count while retaining representative topic checks
- replaced finite global `time.time()` side effects in the IPC monitor tests with module-local stable clock mocks, fixing the Linux/Python 3.10 `StopIteration` failures

Validation: 34 focused tests passed, all 15 changed test modules passed (288 tests), and `./runtest.sh -s --skip-install` passed.

pcnudde:
Addressed the two follow-up Greptile nits in bef87f30b: replaced the `cmd_specs == 17` magic count with representative command-name checks, and removed the stray debug print from `test_is_sub_dir`. Also merged the latest `main` (`f6df71b2c`). Validation: 65 focused tests passed and the repository style check passed.

pcnudde:
Addressed the last two Greptile exact-count nits in f2d2279e1. The ETD test now verifies the expected handler functions, and the ClientRunner test verifies the expected auxiliary topics and event types without pinning total registration counts. Validation: 31 focused tests passed and the repository style check passed.

tests/unit_test/apis/impl/wf_comm_client_test.py
```diff
@@ -0,0 +1,217 @@
...
+
+
+def test_send_and_relay_delegate_in_order():
+    engine = _engine()
+    fl_ctx = _context(engine)
+    task = Task("train", Shareable())
+    comm = WFCommClient()
+    comm._validate_target = MagicMock()
+    ok = make_reply(ReturnCode.OK)
+    comm.broadcast_and_wait = MagicMock(side_effect=[make_reply(ReturnCode.ERROR), ok])
```
YuanTingHsieh: This mock gives `broadcast_and_wait` a return shape it never has in production. Every return path of the real `WFCommClient.broadcast_and_wait` returns a dict `{client_name: reply}` — the success path builds `replies[client_task.client.name] = client_task.result`, and all error paths return `_make_error_reply(...)`, also a dict. It never returns a bare `Shareable`. (The relay assertions a few lines below correctly use the dict shape — same mocked method, two contradictory shapes in one test.)

This matters because it masks a real bug: `send_and_wait` (`nvflare/apis/impl/wf_comm_client.py:276`) does `reply = self.broadcast_and_wait(...)` then `reply.get_return_code()` — with the real dict return this raises `AttributeError: 'dict' object has no attribute 'get_return_code'` on every call (I confirmed by feeding the real `{"site-1": reply}` shape through it). `relay_and_wait` handles the same return correctly via `replies.update(reply)`, which confirms the dict contract. `WFCommClient` has no subclasses and is live via `ClientControllerExecutor`, so any client-side controller calling `send()` hits this.

Please change the send-path mock to the dict shape — which will surface that `send_and_wait` needs a production fix (unwrap the single-target reply). Fine to land the production fix separately; the test just shouldn't cement the wrong contract.

tests/unit_test/edge/assessors/num_test.py
```diff
@@ -0,0 +1,175 @@
...
+    accepted, reply_data = assessor.process_child_update(
+        StateUpdateReport(1, assessor.current_selection_version, {1: update}, None).to_shareable(), FLContext()
+    )
+    reply = StateUpdateReply.from_shareable(reply_data)
+
+    assert accepted
+    assert assessor.current_model_version == 2
+    assert assessor.current_model.data["value"] == 4.0
+    assert reply.model_version == 2
+    assert device_id not in assessor.current_selection or assessor.device_reuse
```
YuanTingHsieh: This assertion is a tautology: `_assessor()` defaults `device_reuse=True` and this test doesn't override it, so this reads `<anything> or True` and always passes. The "device removed from current_selection after its update" behavior is never actually verified — dropping the `current_selection.pop()` in production would still pass. Either assert removal directly for a `device_reuse=False` assessor, or parametrize reuse and assert each side explicitly.

tests/unit_test/edge/assessors/num_test.py
```diff
@@ -0,0 +1,175 @@
...
+    assert accepted
+    assert StateUpdateReply.from_shareable(reply).model_version == 5
+    assert assessor.log_error.call_count >= 2
+
+
+@pytest.mark.parametrize("assessor_cls,state_cls,module", ASSESSORS)
+def test_fill_selection_respects_reuse_policy(assessor_cls, state_cls, module):
+    assessor = _assessor(assessor_cls, device_selection_size=2, device_reuse=False)
+    assessor.available_devices = _devices()
+    assessor.used_devices = {"device-1": 0}
```
YuanTingHsieh: This fixture can't distinguish the two reuse policies: with `used_devices={"device-1": 0}` and `current_model_version == 0`, the reuse=True branch (exclude where `v == current_model_version`) excludes exactly the same set as the reuse=False branch (exclude all used), so the assertions pass regardless of the policy under test. I confirmed by execution: both policies yield the same selection here, whereas `used_devices={"device-1": 5}` makes them diverge. Also note `AsyncNumAssessor._fill_selection` never references `device_reuse` at all, yet the parametrized test still passes for it. Suggest a used-version ≠ current-version fixture, plus splitting or adjusting the async expectation (its policy handling genuinely differs).

tests/unit_test/private/fed/server/server_engine_test.py
```diff
@@ -286,3 +295,228 @@ def test_remove_run_processes_tolerates_terminate_failure():
...
+    engine.get_client_from_name = MagicMock(side_effect=lambda name: client if name == "site-1" else None)
+    engine.get_clients = MagicMock(return_value=[client])
+
+    assert engine._get_aux_msg_target(SiteType.SERVER).fqcn == "server"
+    assert not engine._get_aux_msg_target(SiteType.SERVER_PARENT).job_scoped
+    assert engine._get_aux_msg_target("site-1").fqcn == "site-1.job"
+    assert engine._get_aux_msg_target("missing") is None
+    assert [target.name for target in engine._to_aux_msg_targets([])] == ["site-1"]
+    assert engine._to_aux_msg_targets(["site-1"])[0].fqcn == "site-1.job"
+    assert engine._to_aux_msg_targets(["missing"]) == {}
```
YuanTingHsieh: This pins a type inconsistency in `_to_aux_msg_targets`: the function builds and returns a `list` on every path except invalid-target, where it does `return {}`. If production is ever normalized to `return []`, this test breaks (`[] == {}` is False). Suggest `assert not engine._to_aux_msg_targets(["missing"])` so the test verifies the behavior (empty/falsy on invalid target) without cementing the accidental dict.

tests/unit_test/private/fed/cmi_test.py
```diff
@@ -0,0 +1,200 @@
...
+    assert attached.get_peer_context().get_prop("peer") == "value"
+
+
+@pytest.mark.parametrize(
+    "cell_rc, api_rc",
+    [
+        (CellReturnCode.TIMEOUT, ReturnCode.COMMUNICATION_ERROR),
+        (CellReturnCode.COMM_ERROR, ReturnCode.COMMUNICATION_ERROR),
+        (CellReturnCode.PROCESS_EXCEPTION, ReturnCode.EXECUTION_EXCEPTION),
+        (CellReturnCode.AUTHENTICATION_ERROR, CellReturnCode.UNAUTHENTICATED),
```
YuanTingHsieh, line 107: Heads-up: this row enshrines a known leak rather than a conversion. `RC_TABLE` maps `AUTHENTICATION_ERROR` to `CellReturnCode.UNAUTHENTICATED` — a cell code — while `_convert_return_code` is supposed to produce API `ReturnCode`s. Pinning current behavior is fine, but add a comment marking it as the leaky entry so a future correction to a real API `ReturnCode` isn't misread as a regression.

tests/unit_test/app_common/executors/ipc_exchanger_test.py
```diff
@@ -0,0 +1,306 @@
...
+    )
+    assert context.result_waiter.is_set()
+    assert IPCExchanger._finish_result(context, result_is_valid=False).get_header(MessageHeaderKey.RETURN_CODE) == (
+        CellReturnCode.INVALID_REQUEST
+    )
+
+
+def test_receive_result_validates_context_task_and_payload():
+    exchanger = _exchanger(agent_id="agent")
+    exchanger.logger.error = MagicMock()
```
YuanTingHsieh: `exchanger.logger.error = MagicMock()` mutates shared global state without restore: `FLComponent.logger` comes from `get_obj_logger`, which returns a process-wide `logging.Logger` keyed by class name, so the mock stays attached for every later test in the same pytest worker. Use `unittest.mock.patch.object(exchanger.logger, "error")` as a context manager (or the `monkeypatch` fixture) so it's restored on teardown.

YuanTingHsieh (commented):
Thanks — this is a solid coverage PR. I verified it directly: all 288 new tests pass from this head (5.6s), and they stay clean under `pytest -n 8` together with all sibling suites (the `k8s_pending_timeout_status_test` failure and the statistics ImportErrors in my env reproduce identically on `main`, so they're unrelated). License headers, `[module_name]_test.py` naming/placement, black/isort all check out, and the mock shapes match the real collaborators almost everywhere I cross-checked.

A few tests, though, assert behavior that isn't what production actually does — the inline comments have details. The important one is in `wf_comm_client_test.py`: it masks a real `AttributeError` bug in `WFCommClient.send_and_wait`. I'd like the three test-correctness items (wf_comm send path, the two `num_test.py` ones) fixed before merge; the rest are take-or-leave.

Minor non-blocking notes, batched here to avoid comment spam (the brittle count asserts I was going to flag are already fixed in the latest commits — thanks):
- `net_agent_test.py:228` — the conditional expression inside `assert` means only one branch is evaluated per parametrize row; easy to misread as both being checked.
- The `_context(engine)` FLContext builder is copy-pasted (with drift: identity name, RUN_NUM) into 5 of the new files — worth hoisting into a shared helper/conftest at some point.

tests/unit_test/apis/impl/wf_comm_client_test.py
pcnudde: Fixed in 2bdd37f7b. `send_and_wait` now unwraps the single-target reply to check its return code and returns the replies dict, matching the broadcast/relay contract; the mock now uses the dict shape and the test also covers the all-targets-fail path. I landed the production fix in this PR as its own commit since the corrected test requires it.

tests/unit_test/edge/assessors/num_test.py
pcnudde: Fixed in 13f5da9e4. The test now uses `device_reuse=False` and asserts the removal directly, which is deterministic for both assessor classes.

tests/unit_test/edge/assessors/num_test.py
pcnudde: Fixed in 13f5da9e4. The shared no-reuse test now uses used-version 5 vs current-version 0 so the policies diverge. Split out the rest: a NumAssessor-only test verifying reuse=True excludes only current-version devices (device-1 used at v0 excluded, device-2 used at v5 selectable again), and async-only tests for its genuinely different policy — `_fill_selection` excludes used devices regardless of `device_reuse`, and updates free used devices only when `device_reuse=True`.

tests/unit_test/private/fed/server/server_engine_test.py
pcnudde: Fixed in 13f5da9e4 — now `assert not engine._to_aux_msg_targets(["missing"])`.

tests/unit_test/private/fed/cmi_test.py
pcnudde: Done in 13f5da9e4 — the row now carries a comment marking it as the known RC_TABLE leak so a future correction is not misread as a regression.

tests/unit_test/app_common/executors/ipc_exchanger_test.py
pcnudde: Fixed in 13f5da9e4 via `monkeypatch.setattr` so the shared logger is restored on teardown. Also fixed the same pattern in `cmi_test.py` and `etd_test.py` (with `patch.object` context managers where the mock is asserted).

pcnudde:
Addressed the review in 2bdd37f7b and 13f5da9e4:

- fixed the three test-correctness items (wf_comm send path, both `num_test.py` issues) — inline replies have details. The wf_comm one required the production fix to `WFCommClient.send_and_wait` (unwrap the single-target reply, return the replies dict per the broadcast/relay contract); it is in this PR as its own commit since the corrected test requires it, and the PR description is updated accordingly.
- took all the non-blocking items too: falsy assert for `_to_aux_msg_targets`, RC_TABLE leak comment, logger mock restoration (also in `cmi_test.py`/`etd_test.py` which had the same pattern), explicit per-row expectations in `net_agent_test.py::test_get_connectors_validates_reply`, and the copy-pasted `_context()` builders are hoisted into `tests/unit_test/fl_context_helper.py` with one-line per-file delegates.

Validation: 143 tests across the nine touched modules passed, the broader affected suites passed (1509 tests), `num_test.py` is stable across repeated runs, and `./runtest.sh -s` passed.

pcnudde:
Follow-up in a55ee2aa5 from a deeper pass over the touched send path: `send`/`relay` reuse one `Task` across per-target `broadcast_and_wait` calls while `task.client_tasks` accumulates, so a `send()` failover attempt re-fired `after_task_sent_cb` for the already-tried target and leaked its stale error result into the returned reply dict (reproduced by execution; unreachable before the AttributeError fix, since `send()` crashed on the first attempt). `broadcast_and_wait` now processes and reports only the client tasks it created in the current call — `task.client_tasks` still keeps the full history — which also stops relay re-firing `after_task_sent_cb` for earlier legs. Also `_make_error_reply` no longer aliases one reply object across all targets. Added a failover regression test that goes through the real broadcast path instead of mocking `broadcast_and_wait`. Validation: 140 tests across wf_comm/ccwf/hub, 695 in tests/unit_test/apis, and `./runtest.sh -s` all pass.

YuanTingHsieh (approved):

PR #4863, opened 2026-07-07
[main] Skip auth filter for cellnet protocol-level bye messages

nvshaxie:
Forward-port of #4569 (which fixed NVBug 6154369 on the 2.8 branch) to main, **plus a security tightening of the bye auth-bypass guard** (per review).

On main, `validate_auth_headers()` still exempts only Register/Challenge, not Bye, so any cell shutdown (e.g. `nvflare poc stop`) logs a spurious `unauthenticated msg (channel='cellnet.channel' topic='bye') ... missing client name` ERROR, and the receiver's `_peer_goodbye` never runs (agent cleanup waits on the heartbeat timeout instead of firing immediately).

### Changes
- `defs.py`: add `CellChannel.CELLNET` + `CellChannelTopic.Bye` canonical constants (from #4569).
- `core_cell.py`: use the canonical constants instead of local string literals (from #4569).
- `authenticator.py`: bypass auth for a cellnet `bye` **only when the message actually terminates at this cell**, i.e. `DESTINATION == local_cell_fqcn`. `validate_auth_headers()` gains a `local_cell_fqcn` parameter, supplied at all three call sites (server via `self.cell`, relay via `my_fqcn`, admin HCI client via `self.cell`); it fails closed when the FQCN is unknown.

### Why the guard differs from the 2.8 original
The shipped 2.8 fix (`b4fb37c4`) guarded the bypass with `to_cell == destination`. That compares two sender-controlled headers to each other and never checks that the message terminates here — and because `_forward` rewrites `TO_CELL` at each hop, the equality also holds at the terminal hop of a *forwarded* bye, so a crafted bye addressed to another cell could still slip through and evict that cell's upstream agent. This PR replaces that guard with `DESTINATION == local_cell_fqcn`: the incoming filter runs before CoreCell's forward decision, so a bye whose `DESTINATION` is this cell is handled locally and never forwarded, while one naming another cell falls through to normal auth. Legitimate byes are always direct-neighbor (`Cell.stop()` broadcasts to directly-connected agents, so `DESTINATION` is the receiving cell), so real byes are unaffected — verified end-to-end (the "missing client name" ERROR stays gone).

### Tests
- Added unit tests in `authenticator_test.py`: a bye terminating here is bypassed; a forwarded/forged bye (including one with `TO_CELL == DESTINATION` pointing at another cell) is rejected; no local FQCN fails closed.
- Fixed two existing test doubles (`__new__`-built `FederatedServer`, HCI `_FakeCell`) that didn't expose the cell FQCN the guard now reads.

Follow-up (not this PR): main's guard is now strictly stronger than the 2.8 branch's shipped `to_cell == destination`; the same strengthening should be back-ported to 2.8.

### Types of changes
- [x] Quick tests passed locally (unit tests + black/isort/flake8 on changed files).
- [ ] Documentation updated.


nvflare/private/fed/authenticator.py
```diff
@@ -372,6 +372,27 @@ def validate_auth_headers(
...
+    # normal auth check in that case. The explicit ``to_cell is not None``
+    # check rejects crafted messages that omit both headers (None == None
+    # would otherwise pass).
+    to_cell = message.get_header(MessageHeaderKey.TO_CELL)
+    destination = message.get_header(MessageHeaderKey.DESTINATION)
+    if (
+        topic == CellChannelTopic.Bye
+        and channel == CellChannel.CELLNET
+        and to_cell is not None
+        and to_cell == destination
```
YuanTingHsieh: **Residual auth-bypass gap — the guard doesn't enforce the "direct-neighbor" property its comment claims.** (Pre-existing in the 2.8 original `b4fb37c4`; noting it here since this PR is touching the guard.)

`to_cell == destination` compares two headers that are both sender-controlled, and the block never checks `destination == <this cell's own FQCN>`. I traced the full chain on this branch:

- The auth filter is registered `channel="*", topic="*"` (fed_server.py:1103; relay.py:218), so it runs on every incoming message, including byes that will be forwarded.
- In `CoreCell._process_received_msg`, in-filters run (~1955) **before** the forward decision `if destination != self.my_info.fqcn: self._forward(...)` (~1996), and `TO_CELL` is not rebound on receipt (only `_forward` rewrites it). So this guard sees the sender's forged `TO_CELL`/`DESTINATION`.
- An entity that can inject into the server/relay incoming pipeline without a valid FL token sends `channel=cellnet.channel, topic=bye, TO_CELL=X, DESTINATION=X` for some other cell X. `to_cell==destination` → bypass → returns `None` (auth passed). Since `destination (X) != my fqcn`, the message is then **forwarded** to X.
- At X, `_peer_goodbye` pops `self.agents[peer_ep.name]` where `peer_ep` is the transport endpoint the message arrived on (the server's connection) — not the attacker's ORIGIN — so X drops its upstream server agent. That's an unauthenticated peer-eviction / routing DoS: exactly the forwarded-bye eviction this comment says it prevents.

Threat model is insider/lateral, not anonymous: in secure mode the cellnet connection is mTLS, so the attacker needs a valid transport cert but no valid FL token (a provisioned-but-unauthenticated or compromised participant disrupting another participant). Impact is a transient routing disruption (the agent re-establishes on reconnect), though repeated byes could sustain it.

**Suggested fix (low-risk):** require `destination == <this cell's own FQCN>` instead of (or in addition to) `to_cell == destination`. Legitimate byes are always direct-neighbor messages — a stopping cell broadcasts bye to the cells it is directly connected to, so `destination` is always the receiving cell — so this never breaks a real bye; it only removes the forwarding path, which no legitimate bye uses. `validate_auth_headers` doesn't currently receive the cell's own FQCN, but both call sites can supply it (the server wrapper `_validate_auth_headers` is a bound method with `self.cell`; the relay wrapper has `my_fqcn` in scope).

YuanTingHsieh (commented):
The forward-port itself is faithful and the mechanical parts are clean: the `authenticator.py` block is byte-identical to the 2.8 original `b4fb37c4`, the constant values are preserved (`"cellnet.channel"`/`"bye"`) so registered callbacks still match, there are no leftover references to the removed `_CHANNEL`/`_TOPIC_BYE` constants, imports are present, and in-scope unit tests pass.

One thing worth addressing while we're touching this guard — flagging rather than hard-blocking, since it's pre-existing in the shipped 2.8 fix, not introduced here (see the inline comment). In short: the bye auth-bypass guard checks `to_cell == destination`, which compares two sender-controlled headers to each other and never verifies the message actually terminates at the receiving cell, so it doesn't achieve the "direct-neighbor only" property its own comment claims. This is the right moment to tighten it rather than port the gap verbatim — and 2.8 would want the same change.

nvflare/private/fed/authenticator.py
nvshaxie: Thanks for the careful trace — you're right, `to_cell == destination` compares two sender-controlled headers and doesn't
prove the bye terminates here. I've pushed a fix that bypasses auth only when `DESTINATION == this cell's own FQCN`. Since
the in-filter runs before CoreCell's forward decision (`if destination != my_fqcn: forward`), a bye addressed to this cell
is handled locally and never forwarded, while a bye naming another cell falls through to normal auth — closing the
forwarded-eviction path. Legitimate byes are always direct-neighbor (`Cell.stop()` broadcasts to directly-connected agents,
so DESTINATION is the receiving cell), so real byes are unaffected — verified end-to-end (the `missing client name` ERROR
stays gone). `validate_auth_headers` now takes `local_cell_fqcn`, supplied at all three call sites (server `self.cell`,
relay `my_fqcn`, admin HCI `self.cell`), failing closed when unknown. Added unit tests incl. a forged `TO_CELL ==
DESTINATION` pointing at another cell.

(Two commits: the first is the fix above; the second just updates two existing unit-test mocks/fakes — a `__new__`-built
`FederatedServer` and the HCI `_FakeCell` — that didn't expose the cell FQCN the guard now reads. No behavior change.)

YuanTingHsieh (approved):

YuanTingHsieh:
The fix itself looks good — I verified the guard end-to-end: the incoming filter runs before CoreCell's forward decision (so crafted byes addressed to other cells are rejected mid-transit), `_peer_goodbye` evicts only by the connection-stamped ENDPOINT prop (forged headers can't name a victim), all three `validate_auth_headers` call sites are covered with a fail-closed default, and 449 unit tests pass locally plus adversarial probes (bye on a non-cellnet channel, `quit` on the cellnet channel, case variants) all reject correctly.

One thing to fix before merge: **the PR description no longer matches the code.** It says the bypass is "guarded by `to_cell is not None and to_cell == destination`" and calls this a clean cherry-pick of b4fb37c4 — but ac7c79be4 deliberately replaced that guard with the stronger `DESTINATION == local_cell_fqcn` check. The PR's own test (`test_validate_auth_headers_rejects_bye_forged_to_cell_matching_destination`) shows why: TO_CELL and DESTINATION are both sender-controlled, and `_forward` rewrites TO_CELL at each hop, so the old equality also holds at the terminal hop of any forwarded message. Please update the body so the merge-commit record describes the guard that actually shipped — this is a security fix and the description is what gets read later.

Related follow-up (not this PR): since main's guard is now strictly stronger, the 2.8 branch's shipped `to_cell == destination` guard deserves the same strengthening back-ported.

nvshaxie:
Thanks @YuanTingHsieh! Updated the PR description to match the shipped code: it now documents the actual guard (`DESTINATION == local_cell_fqcn`) instead of the old `to_cell == destination`, and explains why it's stronger than the 2.8 original (TO_CELL/DESTINATION are both sender-controlled and `_forward` rewrites TO_CELL at each hop, so the old equality also holds at a forwarded bye's terminal hop). Good catch on the merge-record accuracy.

Agreed on the follow-up — I'll back-port the same `DESTINATION == local_cell_fqcn` strengthening to the 2.8 branch once this lands.

PR #4866, opened 2026-07-07
Client API design rev 2.2: heartbeat phases, per-launch tokens, F3 dependencies; add TASK_PAYLOAD_READY

YuanTingHsieh:
Design/plan revision 2.2 for Client API Execution Modes (follow-up to #4853), addressing design-review findings, plus the matching protocol-vocabulary follow-up to #4856.

**Design (`client_api_execution_modes.md`):**
- **Forward-path heartbeat rule narrowed**: the no-expiry exemption now covers only the payload-materialization phase — `TASK_ACCEPTED` → new **`TASK_PAYLOAD_READY`** message — bounded by a materialization deadline; heartbeats stay authoritative while user code trains, so a wedged process is detected at heartbeat timeout, not task timeout. Sequence diagrams updated (all four re-validated).
- **`launch_once=False` moves to launch-scoped tokens**: the CJ regenerates the token in each per-launch bootstrap config and stopping a process invalidates its token — a surviving stale process cannot authenticate against a later launch.
- Explicit mapping between lifecycle transfer states and the F3 `TransferOutcome` vocabulary (TRANSFER_COMPLETE ⇔ COMPLETED; TRANSFER_FAILED ⇔ FAILED or ABORTED; receiver truth applies before the mapping).
- `ClientAPIBackendSpec` classified internal (frozen for parallel development, not public API in 2.9); Open Questions annotated with their deciding PRs; status set to **Approved**.

**Plan (`client_api_execution_modes_plan.md`):**
- **F3-4 now hard-depends on F3-2 + F3-3** (the facade's returns-means-delivered guarantee holds only over receiver-confirmed, retry-aware outcomes with budget-bounded resolution); critical path updated — F3-2/F3-3 run in parallel behind F3-1, still inside the 10–14-week floor.
- Bookkeeping: PR-0 recorded as landing inside #4853; EX-5 moved to Wave 4 and assigned P1; CT-5 assigned P1; tier accounting made exact (36 PRs; CT-4 halves noted).

**Vocabulary (`client/cell/defs.py`):** adds `Topic.TASK_PAYLOAD_READY` — #4856 merged before the phase-boundary message was introduced; the freeze tests are updated accordingly.


YuanTingHsieh:
Folded into #4865 per review consolidation — the design revision and protocol vocabulary land with the F3 code they document.

PR #4867, opened 2026-07-08
CellPipe follow-ups: review cleanups, VIA_ROOT routing fall-through, route-loop guard

nvidianz:
## Summary

Follow-ups from the #4835 review rounds. The first commit carries the non-functional cleanups; the second closes the two functional gaps the review identified.

### Review cleanups (commit 1)
- remove the redundant CellPipe alias marker re-check after the explicit-prefix gate
- document that the missing-parent fallback intentionally mirrors root-server placement
- restore the unrelated separator blank line in the 2.8 release notes

### Routing: unconnected FQCN parents and loop protection (commit 2)
- `CoreCell._try_find_ep`: an unconnected direct plain CellPipe child now falls through to the server-root fallback. The exception is deliberately limited to this supported `VIA_ROOT` topology; ordinary same-family children and deeper descendants retain the original immediate failure behavior. This makes a `VIA_ROOT` CellPipe cell — named under the site but connected only to the server root — reachable from its own site's CP and CJ. Previously such a message died at the CP with "no connection to child" / "cannot forward req: no path" (the same failure shape as NVBug 6423409, one config away; see the #4835 merged fix for the default `VIA_CP` topology).
- `CoreCell._find_endpoint`: refuse a resolved next leg that is already on the message `ROUTE`, unless it is the final destination (replies legitimately return to a cell on the route). Resolution is deterministic, so a revisited cell would pick the same leg again and the message would bounce forever; the guard turns an unreachable-target loop (e.g. CP → server → CP for a disconnected pipe cell) into a clean `TARGET_UNREACHABLE`. This guard is what makes the fall-through above safe when the target is not connected anywhere.
- `ScriptRunner`: `valid_connect_types` listed `VIA_RELAY` twice and omitted `VIA_ROOT`, making `pipe_connect_type="via_root"` impossible to configure even though `_PIPE_CONNECT_URL` maps it and the routing layer supports (and tests) root-connected pipes. The accepted values are now derived from `PipeConnectType` so future enum members cannot be omitted by another manual-list error.

## Why

The parser cleanup preserves the existing validation behavior, and the fallback comment makes the intentional duplication explicit. The routing changes extend #4835's topology-shaped naming fix to the `VIA_ROOT` connect type: #4835 removed the phantom FQCN parent for CP- and relay-connected pipes by construction, but a root-connected pipe still has an unconnected FQCN parent (the CP), and the old hard-return dead-ended any same-family sender routing toward it. With the CellPipe-specific fall-through plus the route-loop guard, all three connect types resolve, a genuinely disconnected pipe target fails fast instead of looping, and unrelated descendant routing is unchanged.

## Validation

- `python -m pytest tests/unit_test/fuel/f3/ tests/unit_test/fuel/utils/pipe/ tests/unit_test/private/fed/authenticator_test.py tests/unit_test/job_config/script_runner_test.py -q` — 659 passed (includes the streaming end-to-end tests, which exercise live cells through the changed resolution code)
- New tests: VIA_ROOT reachability from the CP and the CJ, preserved failure behavior for ordinary direct and deeper descendants, loop-guard refuse/allow pair, and raw-string `pipe_connect_type` validation coverage (all three connect types accepted, invalid value rejected)
- Black, isort, and Flake8 on the changed Python files

NVBug: 6371056 (follow-up)
NVBug: 6423409 (follow-up; closes the residual VIA_ROOT variant of the same routing dead-end)


chesterxgchen (commented):
from claude --wrong Comments

chesterxgchen (commented):
Minor issues (non-blocking)

- Redundant _try_path call. In the fall-through case, _try_path(target_info.path) runs twice with identical arguments — once inside the is_ancestor branch, then again in the generic resolution below, guaranteed to return None the second time. Harmless (tiny path lists) but slightly wasteful; restructuring to skip the second call would be marginally cleaner.

- Guard vs. topology changes mid-flight. If a pipe cell connects between hops (e.g., after the server already forwarded once), the guard could refuse a hop that would now succeed. This is transient — the next send resolves correctly — and clearly better than an infinite bounce, so no change needed; just noting the tradeoff.

- PR body nit: the refused hop surfaces to the requester as COMM_ERROR/"cannot forward" via _forward's RETURN path, not as TARGET_UNREACHABLE directly. Cosmetic description mismatch only.

nvflare/fuel/f3/cellnet/core_cell.py
```diff
@@ -1160,17 +1183,33 @@ def _try_find_ep(self, target_fqcn: str, for_msg: Message) -> Union[None, Endpoi
             return agent.endpoint
 
         if same_family(self.my_info, target_info):
+            unconnected_pipe_child = False
             if FQCN.is_parent(self.my_info.fqcn, target_fqcn):
-                self.log_warning(msg=for_msg, log_text=f"no connection to child {target_fqcn}")
-                return None
+                if target_info.path[-1].startswith(CELL_PIPE_LEAF_PREFIX):
```
YuanTingHsieh: VIA_ROOT still permits dotted CellPipe tokens (`_cell_fqcn(..., "agent.v2", FQCN.ROOT_SERVER)` is covered in `cell_pipe_test.py`), producing `site-1.cellpipe~plain~agent.v2~active`. FQCN parsing splits this path, so `site-1` is not the direct parent and `target_info.path[-1]` is `v2~active`; this branch never marks it as an unconnected pipe child. A CP/CJ request therefore returns no path instead of falling through to the server root (I reproduced `_routing_cell("site-1", ["server"])._try_find_ep(...) is None`). Please recognize the CellPipe marker in the descendant segment(s), or disallow dotted VIA_ROOT tokens.

YuanTingHsieh (commented):
One actionable VIA_ROOT routing edge case remains for supported dotted CellPipe tokens.

nvflare/fuel/f3/cellnet/core_cell.py
nvidianz: Addressed in `c077c725b`. I kept dotted VIA_ROOT tokens supported and generalized the existing fallback to inspect the first descendant segment after the current cell for `CELL_PIPE_LEAF_PREFIX`. This lets the CP recognize `site-1.cellpipe~plain~agent.v2~active` even though FQCN splitting makes the final segment `v2~active`. Ordinary deeper descendants still fail closed unless that first descendant segment carries the explicit CellPipe marker. I also added CP and CJ regression tests for the dotted-token route. Validation: 315 CellNet/pipe/ScriptRunner tests passed, and `./runtest.sh -s` passed.

YuanTingHsieh (approved):

PR #4872, opened 2026-07-08
Add study_runtime.yaml v2 per-study runtime config for job launchers

pcnudde:
## Summary

- Introduce `local/study_runtime.yaml` (v2): one site-owned, auto-discovered file for everything a study gets at runtime — data mounts, env vars, secret-backed env vars and file mounts, a per-study Kubernetes pod template (inline or file path), and a per-study default job image (`container.image`; a job-supplied image wins). The job-side contract is unchanged: a job still only names its study.
- Strict parser with path-qualified errors: unknown keys, unknown dataset types, env/secret_env collisions, and secret_env names that would override launcher-owned env vars are hard errors. `type: databricks` dataset entries are reserved and rejected with a clear "not yet supported" error.
- v1 `local/study_data.yaml` stays frozen and fully supported. The two files must not coexist (hard error). v2 is re-read on every job launch; v1 keeps its per-instance cache.
- Kubernetes: `secret_env` emits `secretKeyRef` env entries, `secret_mounts` project Secrets as read-only volumes (optional `items`), and `pod_template` replaces the unreleased `study_job_spec_file_path` launcher argument and its map file (from #4804). Multi-container templates must mark the main container with the `nvflare_job` sentinel when typed entries are configured, replacing the unsafe first-container fallback.
- Docker: `env`/`secret_env` (launcher-environment pass-through) and read-only secret bind mounts; `pod_template` and `items` are rejected as Kubernetes-only. `docker_kwargs` is the Docker-side escape hatch: extra `containers.run` kwargs per study (site defaults -> study -> job precedence, job wins), with launcher-owned keys rejected at parse time and remaining keys validated against the installed Docker SDK.
- `nvflare deploy prepare` and POC Docker provisioning generate v2 templates for new kits; legacy kits that carry `study_data.yaml` keep the v1 launcher argument, so re-preparing existing sites changes nothing.
- Workspace transfer excludes simplified: fixed excludes plus `pod_template` paths read from the one well-known file, removing the resources.json-scraping logic.
- User docs lead with `study_runtime.yaml` and document v1 as legacy; design in `docs/design/study_runtime_config.md`.

Verified end-to-end on Kubernetes (multi-node cluster, 1 server + 4 clients) and Docker: datasets, env/secret delivery, items projection, pod template merge, main-container guardrail, coexistence and reserved-name rejections, live config re-read without parent restart, and v1 legacy regression on both launchers.

FLARE-2972


nvflare/app_opt/job_launcher/k8s_launcher.py
```diff
@@ -922,23 +879,30 @@ def __init__(
...
+                f"file(s) {conflicts}; migrate all studies to study_runtime.yaml and delete the v1 file."
+            )
+        runtime_map = load_study_runtime_file(runtime_file, logger=self.logger)
+        study_runtime = resolve_study_runtime(runtime_map, study, runtime_file, logger=self.logger)
+        reserved = sorted({ref.name for ref in study_runtime.secret_env} & _RESERVED_JOB_ENV_NAMES)
+        if reserved:
+            raise RuntimeError(
+                f"study runtime file '{runtime_file}': secret_env must not override "
+                f"launcher-owned env vars {reserved} for study '{study}'"
+            )
```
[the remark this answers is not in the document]
pcnudde: Moved the reserved-name check into the shared v2 parser: PYTHONPATH and the workspace-transfer variables are now parse-time hard errors for both `env` and `secret_env` (5bbde4dd2). This also removes the silent-override inconsistency where a site-set PYTHONPATH only took effect for jobs without custom code.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

nvflare/app_opt/job_launcher/docker_launcher.py
```diff
@@ -523,6 +559,7 @@ def launch_job(self, job_meta: dict, fl_ctx: FLContext) -> JobHandleSpec:
         # don't fall back to pwd.getpwuid() — which fails when the host UID has no /etc/passwd entry.
         environment = {
             **self.default_job_env,
+            **site_env,
```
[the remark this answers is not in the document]
pcnudde: Covered by the same parser-level check in 5bbde4dd2 — both launchers share `load_study_runtime_file`, so Docker now rejects launcher-owned names in `env`/`secret_env` at parse time too. USER/HOME remain documented launcher-controlled overrides (generic values, unconditionally set).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

YuanTingHsieh:
Follow-up direction to put on record here, since this PR establishes the landing spot:

**Per-study default job image → `studies.<study>.container.image`**

Today the job image must come from job `meta.json` on both launchers, and `pod_template` deliberately cannot set the main container image (the FLARE-owned overlay replaces it — pinned by a test in this PR; only sidecar images come from the template). That leaves the image as the one runtime resource the site does not own, and every job package ends up hard-coding a deployment-specific image ref — which cuts against the v2 contract that the job names only its study.

Proposed follow-up (not this PR):

- `container.image` next to the existing `container.name` — a typed key, so it works identically on Docker and K8s (unlike `pod_template`)
- Resolution: job image → study `container.image` → error. Only the "or raise" step in each launcher changes; no merge-order surgery
- Default semantics first (job-supplied image wins); an enforce/pin mode can come later for sites that want to forbid job-supplied images entirely
- A site-supplied image is site-trusted content, so it should not be BYOC-gated the way job-supplied images are
- Same change should add `"image"` to Docker's `_RESERVED_CONTAINER_KWARGS` — today it can be passed via `default_job_container_kwargs` where it silently does nothing

Re-read-per-launch also means a site can roll a study to a new image version without restarting the parent, which is exactly the operational behavior we want for image updates.

pcnudde:
@YuanTingHsieh agreed on all points — recorded the direction verbatim in the design doc (`docs/design/study_runtime_config.md`, new "Future: Per-Study Default Job Image" section, commit 5691f6468) so this PR carries the landing spot on record. Kept it out of scope here as proposed.

One note captured with it: the Docker `_RESERVED_CONTAINER_KWARGS` addition isn't quite a one-liner — `docker_spec["image"]` is the legitimate job image selector and always present, so reserving `image` naively would log a spurious reserved-keys warning on every job. The follow-up should reject it in `default_job_container_kwargs` init validation only (where today it passes init and then fails at launch with a duplicate-argument error).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

YuanTingHsieh (dismissed):

pcnudde:
Update: per the author's call, `container.image` is now implemented in this PR (97f15462e) rather than deferred — resolution is job-supplied image → study `container.image` → error, identical on both launchers, and the site image is not BYOC-gated (site-trusted content; BYOC gating of job-supplied images stays with the launcher BYOC enforcement work). One deviation from the sketch above: `container.name` was dropped entirely — the `nvflare_job` template sentinel already identifies the main container for merge and the guardrail, so the config knob added nothing. FLARE-3027 is re-scoped to the enforce/pin mode.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

nvflare/tool/poc/poc_commands.py
```diff
@@ -539,7 +548,7 @@ def _prepare_poc_docker_kit(kit_dir: str, docker_config: Dict, poc_workspace: st
         _validate_runtime_config(RUNTIME_DOCKER, docker_config)
         kit_info = _validate_kit(kit_path)
         _prepare_docker(kit_info, kit_path, docker_config)
-        _write_poc_docker_study_data(kit_path, poc_workspace)
+        _write_poc_docker_study_runtime(kit_path, poc_workspace)
```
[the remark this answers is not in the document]
pcnudde: Fixed in 762ae8816 — but with a fail-fast prepare-time CLIException instead of the suggested silent skip: skipping would quietly drop the POC data mount, which is the same silent-divergence failure mode this PR's strict parsing exists to prevent. The error tells the user to delete or migrate the v1 file.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

nvflare/app_opt/job_launcher/docker_launcher.py
```diff
@@ -72,6 +77,9 @@ class DockerStatus:
     "user",
     "working_dir",
 }
+# "image" is additionally reserved for site-level defaults only: jobs select their
+# image through docker_spec["image"], so it must not trip the job-spec warning.
+_RESERVED_DEFAULT_KWARGS = _RESERVED_CONTAINER_KWARGS | {"image"}
```
YuanTingHsieh: **P2 — deploy-side mirror of this set was not updated.** `DOCKER_RESERVED_KWARGS` in `nvflare/tool/deploy/deploy_commands.py` (line 94) still lacks `image`, so `nvflare deploy prepare` accepts `job_launcher.default_job_container_kwargs.image`, writes the kit, and the parent then fails at startup when `DockerJobLauncher.__init__` raises `ValueError: ... reserved keys: ['image']`. Failing at site startup instead of prepare time is exactly what the deploy validator exists to prevent.

The same stale list is also in two docs: `docs/user_guide/nvflare_cli/deploy_command.rst` (line 95) and `docs/design/docker_job_launcher_design.md` (line 264).

Suggest: add `image` in all three places, and have the deploy error point at `studies.<study>.container.image` as the supported way to set a site default image.

nvflare/app_opt/job_launcher/workspace_cell_transfer.py
```diff
@@ -71,9 +70,7 @@
 BOOTSTRAP_CONNECT_POLL_INTERVAL = 0.1
 
 _BOOTSTRAP_CELL_PREFIX = "ws_transfer_"
-_DEFAULT_WORKSPACE_DOWNLOAD_EXCLUDES = frozenset({"local/study_data.yaml", "local/study_job_spec.yaml"})
-_RESOURCE_CONFIG_NAMES = ("resources.json", "resources.json.default")
-_K8S_LAUNCHER_COMPONENT_ID = "k8s_launcher"
+_DEFAULT_WORKSPACE_DOWNLOAD_EXCLUDES = frozenset({"local/study_data.yaml", "local/study_runtime.yaml"})
```
YuanTingHsieh, line 74: **P2 — custom v1 `study_data_pvc_file_path` is no longer excluded from workspace transfer.** Dropping the resources.json scrape regresses a case the old code (and its old test) explicitly protected: a legacy kit whose `study_data_pvc_file_path` points at a non-default file, e.g. `/workspace/local/custom_data.yaml`. That is still a fully supported v1 configuration — the coexistence guard only fires when a v2 file also exists — but the file is no longer excluded, so the site's full study→PVC map is now bundled into every job workspace. Verified on this branch: `_workspace_download_excludes` returns only the two fixed names for such a kit.

Minimal fix: a targeted scrape of just `study_data_pvc_file_path` from `k8s_launcher` components in resources.json — none of the pod-spec-chasing that justified removing the rest of that logic. If we would rather accept the break, it needs a callout in the PR description / release notes, since v1 is documented as frozen and fully supported.

nvflare/app_opt/job_launcher/k8s_launcher.py
```diff
@@ -536,8 +491,12 @@ def _make_manifest(self, job_config):
         if job_config.get("resources"):
             container["resources"] = job_config["resources"]
         env_vars = {k: v for k, v in job_config.get("env", {}).items() if str(v)}
```
YuanTingHsieh, line 493: **P2 — empty-string env values diverge between launchers, and with a pod template the template's stale value silently survives.** The `if str(v)` filter drops `FOO: ""` before the merge, which does not just lose the value — it un-registers the name, so `_merge_named_items` keeps the template's old entry. Repro on this head:

- template main container: `env: [{name: FOO, value: template-old-value}]`
- study config: `env: {FOO: ""}`
- resulting manifest: `FOO=template-old-value` — the site's explicit setting is silently replaced by the template value

Docker meanwhile delivers `FOO=""`. This is the silent-divergence class the strict parser exists to kill.

Suggest fixing at parse time: reject empty-string values in `_parse_env` ("set a value or remove the key"). That makes both launchers behave identically and eliminates the stale-template case, without touching this pre-existing manifest filter that other env sources flow through.

YuanTingHsieh (commented):
Ran a final pass over the current head — three P2s remaining, details inline. All three are small, self-contained fixes; #2 is the only judgment call.

nvflare/app_opt/job_launcher/docker_launcher.py
pcnudde: Fixed in 04c091c04: `image` added to `DOCKER_RESERVED_KWARGS` so `deploy prepare` fails with INVALID_CONFIG pointing at `studies.<study>.container.image`; both docs and the launcher docstring updated to match, plus a prepare-time rejection test.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

nvflare/app_opt/job_launcher/workspace_cell_transfer.py
pcnudde: Fixed in 04c091c04 with the targeted scrape you suggested: `_add_legacy_study_data_excludes` reads only `study_data_pvc_file_path` (plus `workspace_mount_path` for prefix mapping) from k8s_launcher components in resources.json — no pod-spec chasing — and the custom-path exclusion test is restored.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

nvflare/app_opt/job_launcher/k8s_launcher.py
pcnudde: Fixed in 04c091c04 at parse time as suggested: `_parse_env` rejects empty-string values ("set a value or remove the key"), which makes both launchers identical and removes the stale-template survival case; the pre-existing manifest filter is untouched.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

YuanTingHsieh (dismissed):

nvflare/app_opt/job_launcher/k8s_launcher.py
```diff
@@ -1135,6 +1097,15 @@ def launch_job(self, job_meta: dict, fl_ctx: FLContext) -> JobHandleSpec:
...
             }
+            if study_runtime is not None:
+                if study_runtime.secret_env:
+                    job_config["secret_env"] = [
+                        {"name": ref.name, "source": ref.source, "key": ref.key} for ref in study_runtime.secret_env
+                    ]
+                if pod_manifest_template is not None and (
+                    data_mounts or study_runtime.env or study_runtime.secret_env or study_runtime.secret_mounts
+                ):
+                    job_config["require_main_container"] = True
```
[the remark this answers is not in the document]
pcnudde: Fixed in 58d883c5d: `container_image` added to the `require_main_container` condition, with a test asserting an image-only study plus a sentinel-less multi-container template fails with "mark the main container" instead of overlaying the site image on containers[0]. Design doc guardrail wording updated to match.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

YuanTingHsieh (commented):
One remaining P2 in the new `docker_kwargs` precedence path.

nvflare/app_opt/job_launcher/docker_launcher.py
```diff
@@ -553,15 +622,18 @@ def launch_job(self, job_meta: dict, fl_ctx: FLContext) -> JobHandleSpec:
...
         # 2. job-level num_of_gpus translated to device_requests
-        # 3. site-level default device_requests from default_job_container_kwargs
+        # 3. study-level device_requests from docker_kwargs in study_runtime.yaml
+        # 4. site-level default device_requests from default_job_container_kwargs
         #
         # This preserves the documented rule that job-level resource_spec takes precedence
-        # over site-level defaults, while still allowing fine-grained device_requests overrides.
+        # over study and site-level defaults, while still allowing fine-grained
+        # device_requests overrides.
         if num_gpus and "device_requests" not in job_container_kwargs:
```
YuanTingHsieh: **P2 — an explicit zero-GPU job request does not override study-level `device_requests`.** This branch only runs when `num_gpus` is truthy, so with `docker_spec["num_of_gpus"] == 0`, `merged_container_kwargs` retains the study's `docker_kwargs.device_requests`. I reproduced this on the current head: a study requesting `Count: 4` plus a job requesting `num_of_gpus: 0` still calls `containers.run` with `Count: 4`. That contradicts the documented “job wins, including GPU requests” precedence and can allocate GPUs a job explicitly declined.

Please treat the presence of a job GPU count as the override: translate a positive count, remove inherited `device_requests` for an explicit zero (unless the job supplies `device_requests` directly), and add a zero-GPU regression test.

pcnudde: Fixed in 3b1a3535f: presence of a job-level GPU count is now the override signal — a positive count translates as before, an explicit `num_of_gpus: 0` (nested or flat spec) drops study/site-inherited `device_requests`, and an unspecified count still inherits them. Both regression tests added (explicit zero drops, unspecified keeps).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

nvflare/app_opt/job_launcher/study_runtime.py
```diff
@@ -0,0 +1,382 @@
...
+_VALID_NAME = re.compile(r"^[a-z0-9](?:[a-z0-9_-]{0,61}[a-z0-9])?$")
+
+# Env names the launchers own at job launch (PYTHONPATH and the workspace-transfer
+# variables from workspace_cell_transfer). A study value would be silently overridden
+# or break workspace transfer, so both env and secret_env reject them up front.
+_RESERVED_ENV_NAMES = frozenset({"PYTHONPATH", "NVFL_WORKSPACE_OWNER_FQCN", "NVFL_WORKSPACE_TRANSFER_TOKEN"})
+
+# Docker containers.run kwargs owned by the launcher (the docker launcher and deploy
+# validation enforce the same set for site-level default_job_container_kwargs).
+RESERVED_DOCKER_KWARGS = frozenset(
```
YuanTingHsieh, line 55: **P3 — this comment's claim is not true yet: deploy's copy of this set is missing `auto_remove`.** `DOCKER_RESERVED_KWARGS` in `nvflare/tool/deploy/deploy_commands.py` still lacks `auto_remove` (pre-existing, predates this PR), so `deploy prepare` accepts `default_job_container_kwargs.auto_remove` and the kit then fails at parent startup with the launcher's reserved-keys ValueError — the same fail-late pattern as the `image` case fixed earlier. The two doc lists (`docs/user_guide/nvflare_cli/deploy_command.rst`, `docs/design/docker_job_launcher_design.md`) also omit `auto_remove`.

Now that this shared frozenset exists, the clean fix is to make it the single source of truth: have deploy import `RESERVED_DOCKER_KWARGS` instead of maintaining its own literal, and add `auto_remove` to the two doc lists. That also makes this comment true by construction instead of by discipline.

YuanTingHsieh (commented):
One remaining item from the final pass — everything else from the previous rounds is verified fixed on this head (546 tests pass locally; docker_kwargs SDK-key validation probed against real docker-py 7.1.0; zero-GPU decline semantics pinned both ways).

nvflare/app_opt/job_launcher/study_runtime.py
pcnudde: Fixed in f53053dab: deploy now imports `RESERVED_DOCKER_KWARGS` from `study_runtime.py` instead of maintaining its own literal (which had indeed drifted — missing `auto_remove`), so the sets can no longer diverge. Both doc lists updated, and the prepare-time rejection test is parametrized over `image` and `auto_remove`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

YuanTingHsieh (approved):

PR #4873, opened 2026-07-09
Replace public Recipe.job usage with purpose-built Recipe APIs

nvkevlu:
Replaces public Recipe.job usage with purpose-built Recipe APIs.

### Description

Builds on #4862 to remove direct `recipe.job` usage from public examples, documentation, and research jobs.

- Adds `recipe.enable_tensor_streaming(...)` for configuring matching server and client tensor streamers.
- Adds `add_final_global_evaluation(...)` for evaluating a PyTorch recipe’s persisted global model without requesting client model submissions.
- Migrates log streaming, file packaging, server configuration, export, experiment tracking, and final evaluation examples to supported Recipe APIs.
- Expands the client-side MLflow documentation to cover local stores, remote servers, simulation behavior, and per-site configuration.
- Updates the new Kubernetes Recipe example to use `add_server_file()` and targeted `add_client_file()`.

After these changes, there are no remaining `recipe.job` references under `examples/`, `docs/`, or `research/`.

This PR intentionally does not rename the internal `Recipe.job` attribute to `_job`. That mechanical/internal migration will be handled in a separate PR to keep this change focused and easier to review.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/examples/advanced/experiment-tracking/mlflow/hello-pt-mlflow-client/README.md
+++ b/examples/advanced/experiment-tracking/mlflow/hello-pt-mlflow-client/README.md
@@ -94,22 +94,22 @@ python job.py
 
 Since each site has its own MLflow receiver, metrics are stored separately:
 
-With the example's `tracking_uri=None`, the store is created under each site's
-job-result directory as `<site-workspace>/<job-id>/mlflow`. Under the default
-simulation workspace, use `find` to locate the generated directory:
+With the example's `tracking_uri=None`, a SQLite store is created under each
+site's job-result directory as `<site-workspace>/<job-id>/mlflow.db`. Under the
+default simulation workspace, use `find` to locate the generated database:
 
 ### View Site-1 Metrics:
 ```bash
-find /tmp/nvflare/jobs/workdir/fedavg_mlflow_client/site-1 -type d -name mlflow
-mlflow ui --backend-store-uri <site-1-mlflow-directory>
+find /tmp/nvflare/jobs/workdir/fedavg_mlflow_client/site-1 -type f -name mlflow.db
+mlflow ui --backend-store-uri sqlite:////absolute/path/to/site-1/mlflow.db
 ```
 
 Open browser to `http://localhost:5000`
 
 ### View Site-2 Metrics:
 ```bash
-find /tmp/nvflare/jobs/workdir/fedavg_mlflow_client/site-2 -type d -name mlflow
-mlflow ui --backend-store-uri <site-2-mlflow-directory> --port 5001
+find /tmp/nvflare/jobs/workdir/fedavg_mlflow_client/site-2 -type f -name mlflow.db
+mlflow ui --backend-store-uri sqlite:////absolute/path/to/site-2/mlflow.db --port 5001
 ```
 
 Open browser to `http://localhost:5001`
--- a/examples/advanced/experiment-tracking/mlflow/hello-pt-mlflow-client/README.md
+++ b/examples/advanced/experiment-tracking/mlflow/hello-pt-mlflow-client/README.md
@@ -10,19 +10,19 @@ This example demonstrates **site-specific (decentralized) MLflow tracking** usin
 - Server handles MLflow authentication
 
 **Client-Side** (this example):
-- Each client has its own MLflow instance
+- Each client has its own local MLflow store
 - Site-specific metric tracking
-- Each site manages its own MLflow server
+- Each site can optionally point at its own MLflow server
 
 ## Overview
 
 This example shows how to configure per-client MLflow tracking:
 
 ```python
+from model import SimpleNetwork
+
 from nvflare.app_opt.pt.recipes.fedavg import FedAvgRecipe
-from nvflare.app_opt.tracking.mlflow.mlflow_receiver import MLflowReceiver
-from nvflare.apis.analytix import ANALYTIC_EVENT_TYPE
-from nvflare.recipe import SimEnv
+from nvflare.recipe import SimEnv, add_experiment_tracking
 
 # Create recipe WITHOUT default server-side tracking
 recipe = FedAvgRecipe(
@@ -36,27 +36,31 @@ recipe = FedAvgRecipe(
     train_script="client.py",
 )
 
-# Add MLflow receiver to each client
-for i in range(2):
-    site_name = f"site-{i + 1}"
-    tracking_uri = f"file:///tmp/nvflare/jobs/workdir/{site_name}/mlruns"
-
-    receiver = MLflowReceiver(
-        tracking_uri=tracking_uri,
-        events=[ANALYTIC_EVENT_TYPE],  # Listen to LOCAL events (not federated)
-        kw_args={"experiment_name": f"site-{site_name}-experiment"}
-    )
-
-    recipe.job.to(receiver, site_name, id="mlflow_receiver")
+# Add an MLflow receiver to every client, but not to the server.
+# With tracking_uri=None, each receiver creates a store in its local job workspace.
+add_experiment_tracking(
+    recipe,
+    "mlflow",
+    tracking_config={
+        "tracking_uri": None,
+        "kw_args": {
+            "experiment_name": "nvflare-fedavg-experiment",
+            "run_name": "nvflare-fedavg-client",
+        },
+    },
+    client_side=True,
+    server_side=False,
+)
 
 env = SimEnv(num_clients=2)
 run = recipe.execute(env)
 ```
 
 **Key points**:
-- By default, no server-side tracking is enabled (analytics_receiver defaults to None)
-- `events=[ANALYTIC_EVENT_TYPE]` - Listen to local events (not `fed.` events)
-- Each site gets its own `tracking_uri`
+- `client_side=True, server_side=False` keeps metric handling on the clients.
+- The helper configures receivers to listen to local events rather than federated `fed.` events.
+- `tracking_uri=None` resolves to a separate MLflow store in each client's job workspace.
+- Sites can use explicit local paths or remote MLflow servers when their deployment requires them.
 
 ## Setup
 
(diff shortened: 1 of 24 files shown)


nvflare/recipe/utils.py
```diff
@@ -314,6 +314,76 @@ def add_experiment_tracking(
         recipe._add_to_client_apps(client_receiver, clients=clients, id="client_receiver")
 
 
+def add_final_global_evaluation(
```
YuanTingHsieh, line 429: @holgerroth please help comment if this is needed/good to have

research/auto-fl-research/tasks/cifar10/job.py
```diff
@@ -50,11 +50,8 @@
 )
 
 from nvflare.apis.dxo import DataKind
-from nvflare.app_common.widgets.validation_json_generator import ValidationJsonGenerator
-from nvflare.app_common.workflows.cross_site_model_eval import CrossSiteModelEval
-from nvflare.app_opt.pt.file_model_locator import PTFileModelLocator
 from nvflare.app_opt.pt.recipes.fedavg import FedAvgRecipe
-from nvflare.recipe import SimEnv, add_experiment_tracking
+from nvflare.recipe import SimEnv, add_experiment_tracking, add_final_global_evaluation
```
YuanTingHsieh, line 54: @holgerroth please help comment if this is needed/good to have

nvflare/recipe/spec.py
```diff
@@ -542,6 +542,73 @@ def enable_log_streaming(self, *file_names: str) -> None:
             self._add_to_client_apps(JobLogStreamer(log_file_name=name))
         self.job.to_server(JobLogReceiver())
 
+    def enable_tensor_streaming(
```
YuanTingHsieh, line 667: @holgerroth please help comment if this is needed/good to have, 
this is JUST for TensorClientStreamer/TensorServerStreamer,
so it is not the same as the streaming we have in f3

nvflare/app_opt/tf/tf_validator.py
```diff
@@ -55,31 +55,18 @@ def __init__(
         **Default recommendation**: Use Client API pattern (like PyTorch) for consistency and simplicity.
         TFValidator is provided for flexibility and backward compatibility with NumPy-style workflows.
 
-        Example (Component-based with TFValidator):
+        Example (Component-based with the lower-level Job API):
```
YuanTingHsieh, line 58: or should we just remove this example section docstring?

examples/advanced/tensor-stream/job.py
```diff
@@ -52,8 +50,7 @@ def main():
     )
     add_experiment_tracking(recipe, tracking_type="tensorboard")
 
-    recipe.job.to_server(TensorServerStreamer(), "tensor_server_streamer")
-    recipe.job.to_clients(TensorClientStreamer(), "tensor_client_streamer")
+    recipe.enable_tensor_streaming()
```
YuanTingHsieh, line 53: @holgerroth please help to comment if this is good to have

nvflare/recipe/spec.py
YuanTingHsieh: alternative if this is not a util, we just do recipe.add_xxx_component

YuanTingHsieh (commented):
Changes LGTM, make sure you tested these updated examples.
And for some utils can @holgerroth and @ZiyueXu77 comment on if that util makes sense?
otherwise we could just use the recipe.add_server_component or recipe.add_client_component

nvflare/recipe/spec.py
YuanTingHsieh: btw, maybe this should be moved to utils.py rather than in spec.py

examples/advanced/experiment-tracking/mlflow/hello-pt-mlflow-client/README.md
```diff
@@ -36,27 +36,31 @@ recipe = FedAvgRecipe(
...
-
-    recipe.job.to(receiver, site_name, id="mlflow_receiver")
+# Add an MLflow receiver to every client, but not to the server.
+# With tracking_uri=None, each receiver creates a store in its local job workspace.
+add_experiment_tracking(
+    recipe,
+    "mlflow",
+    tracking_config={
+        "tracking_uri": None,
+        "kw_args": {
```
chesterxgchen: can we make these args with default values

for example if kw_args are not present, then could default = 
experiment_name = <job_name>-experiment
run_name = <job_name>-Client

further user can simple omit the tracking config if the tracking URL is none. 
 

add_experiment_tracking(recipe, "mlflow)

nvkevlu: updated

nvflare/recipe/utils.py
```diff
@@ -353,10 +355,21 @@ def add_experiment_tracking(
...
+        kw_args = tracking_config.get("kw_args")
+        if kw_args is None:
+            kw_args = {}
+            tracking_config["kw_args"] = kw_args
+        elif not isinstance(kw_args, dict):
+            raise TypeError(f"MLflow kw_args must be a dict, got {type(kw_args).__name__}")
+        recipe_name = getattr(recipe, "name", None) or getattr(recipe.job, "name", None) or "nvflare"
+        kw_args.setdefault("experiment_name", f"{recipe_name}-experiment")
+        kw_args.setdefault("run_name", f"{recipe_name}-Client")
+
```
holgerroth, line 383: The `run_name` default of `"{recipe_name}-Client"` is set unconditionally, even when `server_side=True` and `client_side=False`. For a server-side aggregation run the name "-Client" is misleading. Consider using separate defaults depending on which side is enabled, e.g. `"{recipe_name}-Server"` when only `server_side=True`.

nvflare/recipe/spec.py
```diff
@@ -542,6 +542,73 @@ def enable_log_streaming(self, *file_names: str) -> None:
...
+            TypeError: If ``tasks`` is not a list of strings.
+            ValueError: If ``tasks`` is empty, if ``format`` does not match a
+                declared ``server_expected_format``.
+            RuntimeError: If tensor streaming was already enabled.
+        """
+        from nvflare.app_opt.tensor_stream.client import TensorClientStreamer
+        from nvflare.app_opt.tensor_stream.server import TensorServerStreamer
+
+        if getattr(self, "_tensor_streaming_added", False):
+            raise RuntimeError("tensor streaming has already been enabled for this recipe")
```
holgerroth, line 699: `_tensor_streaming_added` is set here as a dynamic attribute rather than being declared in `__init__`. The same pattern appears for `_cse_added` in `utils.py`. Accessing these with `getattr(..., False)` works today, but if `Recipe` ever gains `__slots__` or dataclass-style annotations these will behave unexpectedly. Recommend initialising both flags to `False` in `Recipe.__init__` to keep them part of the defined interface.

nvflare/recipe/utils.py
```diff
@@ -321,7 +321,9 @@ def add_experiment_tracking(
     Args:
         recipe: Recipe instance to augment with experiment tracking.
         tracking_type: Type of tracking to enable ("mlflow", "tensorboard", or "wandb").
-        tracking_config: Optional configuration dict for the tracking receiver.
+        tracking_config: Optional configuration dict for the tracking receiver. For MLflow,
+            omitting this uses a local file store and derives ``experiment_name`` and
+            ``run_name`` from the recipe name.
         client_side: If True, add tracking to clients (each client tracks locally).
         server_side: If True, add tracking to server (aggregates metrics from all clients). Default: True.
         clients: Optional list of client names for client-side tracking. If None, the
```
holgerroth, line 338: The new public `add_final_global_evaluation` defaults `validation_timeout` to **6000 seconds**, whereas the old local copy in the research scripts omitted the argument entirely (falling back to whatever `CrossSiteModelEval.__init__` defaults to). Callers that migrate from the local helper to this API will silently get a different timeout if the two defaults differ. Worth noting in the docstring or confirming the values match.

tests/unit_test/recipe/utils_test.py
```diff
@@ -586,6 +592,117 @@ def __init__(self, *args, **kwargs):
...
+        add_final_global_evaluation(recipe, participating_clients=["site-1"], validation_timeout=42)
+
+        calls = recipe.job.to_server.call_args_list
+        assert isinstance(calls[0].args[0], PTFileModelLocator)
+        assert calls[0].kwargs == {"id": "final_model_locator"}
+        assert isinstance(calls[1].args[0], ValidationJsonGenerator)
+        controller = calls[2].args[0]
+        assert isinstance(controller, CrossSiteModelEval)
+        assert controller._model_locator_id == "final_model_locator"
+        assert controller._submit_model_task_name == ""
```
holgerroth: These assertions pin private attributes of `CrossSiteModelEval` (`_model_locator_id`, `_submit_model_task_name`, `_participating_clients`, `_validation_timeout`). If that class renames or removes its underscore-prefixed fields the tests break without the public API changing. Where possible, prefer asserting through the public interface; at minimum, add a comment acknowledging this dependency so the next reader knows these are intentional internal checks.

holgerroth (commented):
A few non-blocking suggestions from code review. The overall direction is great — removing `recipe.job` from public examples is the right cleanup.

nvflare/recipe/utils.py
holgerroth: Yes, this is good to have. Makes sense to promote this into the public Recipe API rather than leaving it as scattered local helpers.

research/auto-fl-research/tasks/cifar10/job.py
holgerroth: Yes, looks good. Happy to see the research jobs cleaned up to use the public API.

nvflare/recipe/spec.py
holgerroth: Yes, this makes sense. The naming is fine — it is specifically for the TensorClientStreamer/TensorServerStreamer pair and is not related to F3 streaming.

examples/advanced/tensor-stream/job.py
holgerroth: Yes, good to have. Much cleaner for the example.

nvflare/app_opt/tf/tf_validator.py
nvkevlu: I think this example is still useful because it documents explicit component-based TFValidator placement, which requires the lower-level Job API. Recipe-based jobs are directed to the recommended Client API pattern above, so I’ve kept the example but labeled it clearly.

holgerroth (dismissed):
This makes sense. Nice cleanup — getting recipe.job out of the public examples and docs is a good step, and the new enable_tensor_streaming and add_final_global_evaluation APIs are good additions. All my review comments were addressed in the latest commits, including the matching-defaults test for the validation timeout.

holgerroth (approved):

PR #4874, opened 2026-07-09
Validate WEIGHT_DIFF recipe configurations and add DP example

nvkevlu:
Adds construction-time validation for WEIGHT_DIFF recipe configurations, documents supported recipes and aggregators, and demonstrates FedAvg with DP weight differences.

### Description

- Add construction-time validation to ensure server aggregation and client transfer settings use compatible model-update kinds.
- Construction-time validation covers recipe-owned aggregator invariants, while FLModel.params_type is authoritative for client output.
- Validate per-site transfer overrides and custom aggregators that declare `expected_data_kind`.
- Apply validation to FedAvg, HE FedAvg, PyTorch FedOpt, and Swarm recipes.
- Document which recipes and aggregators support `WEIGHT_DIFF`.
- Update the Hello DP example to use FedAvg with weight-difference aggregation.
- Correct existing examples that selected `WEIGHT_DIFF` without enabling DIFF transfer.

## Why

Previously, incompatible client and server settings were only detected at runtime. This made recipe misconfiguration difficult to diagnose. The new validation fails during recipe construction and provides the exact configuration needed to fix the issue.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/docs/user_guide/data_scientist_guide/available_recipes.rst
+++ b/docs/user_guide/data_scientist_guide/available_recipes.rst
@@ -179,6 +179,82 @@ FedAvg with secure aggregation using homomorphic encryption.
 
 - `examples/advanced/cifar10/pt/cifar10-real-world#secure-aggregation-using-homomorphic-encryption <https://github.com/NVIDIA/NVFlare/tree/main/examples/advanced/cifar10/pt/cifar10-real-world#42-secure-aggregation-using-homomorphic-encryption>`_
 
+WEIGHT_DIFF Compatibility
+-------------------------
+
+``DataKind.WEIGHT_DIFF`` is supported only when the client executor sends parameter
+differences and the server aggregation path accepts differences. Recipe construction
+validates these settings where both are configurable.
+
+.. list-table:: Recipe and aggregator support
+   :header-rows: 1
+   :widths: 28 30 12 30
+
+   * - Recipe
+     - Server aggregation path
+     - Support
+     - Required configuration
+   * - Unified, PyTorch, TensorFlow, and NumPy ``FedAvgRecipe``; PyTorch FedProx
+     - Built-in ``FedAvg`` streaming aggregation
+     - Yes
+     - ``aggregator_data_kind=DataKind.WEIGHT_DIFF`` and
+       ``params_transfer_type=TransferType.DIFF``
+   * - ``FedAvgRecipe`` with a custom ``ModelAggregator``
+     - User-provided aggregator
+     - Conditional
+     - Use the same pair as FedAvg. The custom aggregator must accept difference models
+       and preserve ``ParamsType.DIFF`` in its aggregate result.
+   * - ``FedAvgRecipeWithHE``
+     - ``HEInTimeAccumulateWeightedAggregator``
+     - Yes
+     - ``aggregator_data_kind=DataKind.WEIGHT_DIFF`` and
+       ``params_transfer_type=TransferType.DIFF``
+   * - PyTorch ``FedOptRecipe``
+     - ``InTimeAccumulateWeightedAggregator``
+     - Yes
+     - Always configured for ``WEIGHT_DIFF`` and ``DIFF`` by the recipe.
+   * - TensorFlow ``FedOptRecipe``
+     - Built-in ``FedAvg`` streaming aggregation with FedOpt model update
+     - Yes
+     - ``params_transfer_type=TransferType.DIFF``; there is no separate
+       ``aggregator_data_kind`` parameter.
+   * - ``SwarmLearningRecipe``
+     - ``InTimeAccumulateWeightedAggregator``
+     - Yes
+     - ``expected_data_kind=DataKind.WEIGHT_DIFF`` and
+       ``params_transfer_type=TransferType.DIFF``
+   * - ``SklearnFedAvgRecipe``
+     - Built-in ``FedAvg`` streaming aggregation
+     - No
+     - This recipe currently fixes client transfer to ``TransferType.FULL``.
+
+The standard ``InTimeAccumulateWeightedAggregator`` and
+``HEInTimeAccumulateWeightedAggregator`` accept both ``WEIGHTS`` and ``WEIGHT_DIFF``
+when their ``expected_data_kind`` is configured accordingly. Aggregators that declare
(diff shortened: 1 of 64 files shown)


nvflare/recipe/utils.py
```diff
@@ -67,6 +69,96 @@
...
+        return
+
+    transfer_type = TransferType(transfer_type)
+    if data_kind not in (DataKind.WEIGHTS, DataKind.WEIGHT_DIFF):
+        raise ValueError(
+            f"{recipe_name} does not support {data_kind_arg}=DataKind.{data_kind.name}; "
+            "use DataKind.WEIGHTS or DataKind.WEIGHT_DIFF."
+        )
+
+    client_data_kind = DataKind.WEIGHT_DIFF if transfer_type == TransferType.DIFF else DataKind.WEIGHTS
```
YuanTingHsieh: **[P1] Don’t make `TransferType` the semantic source of truth**

This inference is not valid for the client API: `TransferType.FULL` only disables automatic diff calculation; it does not prevent a training script from returning `FLModel(params_type=ParamsType.DIFF)`. Since `TransferType` is planned for retirement, this creates a new required public contract around a legacy control and rejects valid existing clients that compute their own deltas. Please remove the requirement that `aggregator_data_kind` and `params_transfer_type` be explicitly paired. `FLModel.params_type` should remain the authoritative description of the client result.

docs/user_guide/data_scientist_guide/available_recipes.rst
```diff
@@ -179,6 +179,82 @@ FedAvg with secure aggregation using homomorphic encryption.
...
+
+   * - Recipe
+     - Server aggregation path
+     - Support
+     - Required configuration
+   * - Unified, PyTorch, TensorFlow, and NumPy ``FedAvgRecipe``; PyTorch FedProx
+     - Built-in ``FedAvg`` streaming aggregation
+     - Yes
+     - ``aggregator_data_kind=DataKind.WEIGHT_DIFF`` and
+       ``params_transfer_type=TransferType.DIFF``
```
YuanTingHsieh: **[P1] Don’t teach the explicit pair as the new public API**

These docs and the updated examples establish `aggregator_data_kind=WEIGHT_DIFF` plus `params_transfer_type=DIFF` as the recommended configuration. If `TransferType` is being retired, we should not propagate it through user-facing examples now. Examples should either return `FLModel(params_type=ParamsType.DIFF)` explicitly or continue using full weights. If automatic conversion is still needed during migration, the recipe can derive the legacy transfer behavior internally instead of requiring users to configure both knobs.

nvflare/recipe/fedavg.py
```diff
@@ -449,6 +452,34 @@ def _validate_per_site_config(per_site_config: Optional[Dict[str, Dict]]) -> Non
...
+                        site_config.get("params_transfer_type")
+                        if site_config.get("params_transfer_type") is not None
+                        else self.params_transfer_type
+                    ),
+                )
+                for site_name, site_config in self.per_site_config.items()
+            ]
+
+        for site_name, transfer_type in client_transfer_types:
+            validate_data_kind_transfer_type(
```
YuanTingHsieh: **[P2] Limit construction-time validation to information the recipe owns**

We can reliably validate a custom aggregator’s declared `expected_data_kind`, and fixed recipes such as FedOpt can enforce invariants they construct themselves. For an arbitrary training script, however, this recipe cannot determine the emitted `FLModel.params_type` from `params_transfer_type`. Please retain aggregator-to-recipe validation where applicable, but remove client-output validation based solely on `TransferType`. This also avoids the Sklearn path recommending `params_transfer_type=DIFF`, which that recipe does not expose.

YuanTingHsieh (commented):
I think the construction-time checks are valuable for aggregator invariants, but we should avoid establishing `TransferType` as a new semantic source of truth if it is planned for retirement. Details inline.

tests/unit_test/recipe/fedavg_recipe_test.py
```diff
@@ -18,11 +18,13 @@
 import pytest
 import torch.nn as nn
 
+from nvflare.apis.dxo import DataKind
 from nvflare.apis.fl_component import FLComponent
-from nvflare.apis.job_def import ALL_SITES, SERVER_SITE_NAME
+from nvflare.apis.job_def import ALL_SITES
```
YuanTingHsieh: **[P1] Restore the `SERVER_SITE_NAME` import**

This file still references `SERVER_SITE_NAME` repeatedly (starting in `get_model_selector()`), but this import now includes only `ALL_SITES`. That produces an undefined-name style error and a unit-test `NameError`, consistent with the current style-check and unit-test failures. Please restore `from nvflare.apis.job_def import ALL_SITES, SERVER_SITE_NAME`.

docs/user_guide/data_scientist_guide/available_recipes.rst
```diff
@@ -179,6 +179,93 @@ FedAvg with secure aggregation using homomorphic encryption.
...
+   * - ``FedAvgRecipeWithHE``
+     - ``HEInTimeAccumulateWeightedAggregator``
+     - Yes
+     - Set ``aggregator_data_kind=DataKind.WEIGHT_DIFF`` and return
+       ``FLModel(params_type=ParamsType.DIFF)`` from the client.
+   * - PyTorch ``FedOptRecipe``
+     - ``InTimeAccumulateWeightedAggregator``
+     - Yes
+     - The recipe fixes its aggregation path to ``WEIGHT_DIFF``. A custom aggregator must
+       declare ``expected_data_kind=DataKind.WEIGHT_DIFF``.
```
YuanTingHsieh: **[P3] Match the FedOpt wording to the validator**

The validator permits custom aggregators that do not declare `expected_data_kind`; it only verifies the value when a declaration exists. This should therefore say: “If a custom aggregator declares `expected_data_kind`, it must declare `DataKind.WEIGHT_DIFF`,” unless the implementation is changed to require the attribute.

YuanTingHsieh (commented):
The revised design now correctly treats `FLModel.params_type` as authoritative and limits construction-time checks to server-owned aggregator settings. One test import currently blocks CI, and the PR description still describes the removed `TransferType` pairing contract.

YuanTingHsieh (approved):

PR #4875, opened 2026-07-09
Add FedCE and FedSM recipe support

nvkevlu:
Adds dedicated PyTorch recipes and reusable components for FedCE contribution estimation and FedSM personalized federated learning.

### Description

Add FedCERecipe with:

- Contribution-aware model aggregation.
- Gradient-direction and minus-model contribution signals.
- PTFedCEHelper utilities for client integration.
- FedCE coefficient propagation between rounds.

Add FedSMRecipe with:

- Global, personalized, and selector model management.
- SoftPull aggregation for personalized models.
- Optional selector optimizer-state synchronization.
- PTFedSMHelper utilities for loading and returning model bundles.
- Complete model-bundle persistence and checkpoint resume support.

Restrict both recipes to PyTorch and document their client-script contracts.

Keep FedCE and FedSM as separate, non-composable algorithms.

Add both recipes to the public exports and recipe CLI catalog.

Add user documentation and recipe examples.

Add aggregation, helper, persistence, validation, export, and catalog tests.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/nvflare/app_opt/pt/fedsm.py
+++ b/nvflare/app_opt/pt/fedsm.py
@@ -70,8 +70,26 @@ class PTFedSMHelper:
         self._selector_before = None
 
     def load_bundle(self, incoming: FLModel, client_name: Optional[str] = None):
+        if incoming.params_type != ParamsType.FULL:
+            raise ValueError(f"FedSM incoming bundle must use ParamsType.FULL, got {incoming.params_type}")
+
         params = incoming.params or {}
-        target = incoming.meta.get(FedSMConstants.TARGET_ID)
+        required_params = {
+            FedSMConstants.GLOBAL_MODEL,
+            FedSMConstants.PERSONAL_MODEL,
+            FedSMConstants.SELECTOR_MODEL,
+        }
+        missing_params = required_params.difference(params)
+        if missing_params:
+            raise ValueError(f"FedSM incoming bundle is missing parameter entries: {sorted(missing_params)}")
+
+        meta = incoming.meta or {}
+        required_meta = {FedSMConstants.TARGET_ID, FedSMConstants.SELECTOR_LABEL}
+        missing_meta = required_meta.difference(meta)
+        if missing_meta:
+            raise ValueError(f"FedSM incoming bundle is missing metadata entries: {sorted(missing_meta)}")
+
+        target = meta[FedSMConstants.TARGET_ID]
         if client_name is not None and target != client_name:
             raise ValueError(f"FedSM bundle targets {target!r}, but this client is {client_name!r}")
         self.global_model.load_state_dict(params[FedSMConstants.GLOBAL_MODEL])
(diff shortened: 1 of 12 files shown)


YuanTingHsieh (commented):
I found three correctness and client-contract issues that should be addressed. Details are inline.

nvflare/app_opt/pt/recipes/fedce.py
```diff
@@ -0,0 +1,100 @@
...
+        client_memory_gc_rounds: int = 0,
+        cuda_empty_cache: bool = False,
+    ):
+        if server_expected_format != ExchangeFormat.PYTORCH:
+            raise ValueError("FedCERecipe requires server_expected_format=ExchangeFormat.PYTORCH")
+
+        trainable_param_names = None
+        named_parameters = getattr(model, "named_parameters", None)
+        if callable(named_parameters):
+            trainable_param_names = [name for name, parameter in named_parameters() if parameter.requires_grad]
```
YuanTingHsieh, line 74: **[P2] Preserve trainable-parameter filtering for dict-config models.** When `model` is a supported dict config, `named_parameters` is unavailable and `trainable_param_names` remains unset. `FedCEModelAggregator._get_cosine_param_names()` then falls back to every common state entry, including non-trainable buffers such as BatchNorm running statistics and counters, which silently changes the contribution coefficients. Please resolve the configured model's trainable names at runtime (or require them explicitly) and add a dict-model regression test with a non-trainable buffer.

nvflare/app_opt/pt/fedsm.py
```diff
@@ -0,0 +1,472 @@
...
+        return {key: _to_cpu_copy(item) for key, item in value.items()}
+    return copy.deepcopy(value)
+
+
+def _weighted_average(values: List, weights: List[float]):
+    first = values[0]
+    if isinstance(first, dict):
+        common_keys = set(first)
+        for value in values[1:]:
+            common_keys.intersection_update(value)
```
YuanTingHsieh: **[P2] Reject partial model states instead of intersecting their keys.** If one client omits an inner state key, this intersection drops that parameter from every aggregated personalized model. `_update_bundle()` then replaces each complete personal model with the truncated state, and the next round's strict `load_state_dict()` fails. Since the client contract requires a full personalized state, please validate identical inner keys and reject incomplete bundles rather than silently intersecting them.

nvflare/app_opt/pt/fedsm.py
```diff
@@ -0,0 +1,472 @@
...
+        bundle = {
+            FedSMConstants.GLOBAL_MODEL: global_update,
+            FedSMConstants.PERSONAL_MODELS: personal_updates,
+            FedSMConstants.SELECTOR_MODEL: selector_update,
+        }
+        optimizer_states = [self._results[client].params.get(FedSMConstants.SELECTOR_OPTIMIZER) for client in clients]
+        if all(state is not None for state in optimizer_states):
+            bundle[FedSMConstants.SELECTOR_OPTIMIZER] = _weighted_average(optimizer_states, weights)
+
+        return FLModel(
```
YuanTingHsieh, line 270: **[P2] Preserve the metrics accepted by `PTFedSMHelper.build_result()`.** Client results may contain `metrics`, but this aggregation path never reads them and returns an `FLModel` without `metrics`; the controller consequently clears `model.metrics`. That makes the helper's `metrics` argument ineffective and breaks round-level monitoring and metric consumers. Please aggregate supported metrics using the client-step weights (or remove the advertised metrics input if dropping them is intentional).

ZiyueXu77:
As these are specific to these two research items, do we need them to be included in official recipe folder, or we can add them just to each research subfolder?

nvkevlu:
I will close this now to reopen each as a separate PR later

PR #4877, opened 2026-07-09
Add cleanup for staged Kubernetes resources

IsaacYangSLA:
## Summary

- add `nvflare deploy k8s unstage` to delete the exact Secret and ConfigMap created by Kubernetes staging
- record staging namespace and resource names before cluster mutation, preserve retry state on failures, and clear Helm value references after cleanup
- prevent restaging or re-preparing a kit in ways that could lose cleanup targets
- document teardown and upgrade ordering for Kubernetes and OpenShift workflows
- add coverage for default and custom targets, legacy state, failures, idempotency, CLI routing, and prepare safeguards

## Why

`nvflare deploy k8s stage` applies the participant's `local/` ConfigMap and credential-bearing `startup/` Secret outside the Helm release. Helm therefore leaves both objects behind on uninstall, and there was no NVFlare cleanup command.

This change provides an explicit, exact-name cleanup lifecycle without copying identity material into Helm release storage.

## User impact

After uninstalling the Helm release, users can run the cleanup command printed by `stage`:

```bash
nvflare deploy k8s unstage <prepared-kit>
```

The command deletes the Secret first, deletes the ConfigMap, tolerates already-removed objects, and clears the prepared chart references.

## Validation

- `python3 -m pytest -q tests/unit_test/tool/deploy/deploy_commands_test.py` — 97 passed
- `python3 -m pytest -q tests/unit_test/tool/cli_schema_test.py` — 46 passed
- `./runtest.sh -s` — passed
- `git diff --check` — passed

### DIFF
--- a/docs/design/site_runtime_packaging_design.md
+++ b/docs/design/site_runtime_packaging_design.md
@@ -482,7 +482,11 @@ manage the Kubernetes cluster. The generated Helm chart is the deployment
 artifact, and the site admin applies it to an existing cluster with standard
 K8s/Helm tooling such as `helm install` or `helm upgrade`. If the site admin
 uses `nvflare deploy k8s stage`, that staging command is run before the Helm
-install/upgrade command.
+install/upgrade command. The staged ConfigMap and Secret are deliberately kept
+out of the Helm release so identity material is not copied into Helm release
+storage. The staging command records their exact names and namespace, and the
+site admin runs `nvflare deploy k8s unstage` after Helm uninstall to delete both
+objects and clear their chart references.
 
 ## Runtime Communication Patching
 
--- a/docs/user_guide/admin_guide/deployment/helm_chart.rst
+++ b/docs/user_guide/admin_guide/deployment/helm_chart.rst
@@ -440,7 +440,20 @@ Use ``--kubectl oc`` when staging into OpenShift with ``oc`` instead of
 the workspace root for writable runtime state such as jobs, snapshots, logs, and
 ``transfer/``. After this staging command succeeds, run the printed
 ``helm_command`` or the equivalent ``helm upgrade --install`` command for the
-prepared chart.
+prepared chart. The staged ConfigMap and Secret are separate from the Helm
+release. After uninstalling the release, run the printed ``cleanup_command``
+or the equivalent command below to remove them:
+
+.. code-block:: bash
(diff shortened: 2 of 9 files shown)


pcnudde (approved):

chesterxgchen (commented):
Security engineer review

Residual risk / notes:
1. Cleanup is still human-dependent. A plain helm uninstall still orphans the Secret; the control is docs + a printed command + the prepare guard. That's the accepted cost of keeping key material out of Helm release storage, and I agree with the tradeoff — but the bug's Impact section (credential hygiene on shared clusters) is **only mitigated, not eliminated**. Worth a follow-up: recommend cert rotation/short-lived identities in the docs, since a forgotten Secret today stays valid.

2. The redaction is ineffective theater (minor): the error path substitutes secret/<staged-name> in the failure message, and the success output omits the secret name — but kubectl's own stderr is appended verbatim and typically contains the real name (secrets "nvflare-startup-site-1" is forbidden...), and stage prints the name openly anyway. Secret names aren't sensitive; I'd either drop the redaction or redact the stderr detail too for consistency. Not blocking.

3. kubectl delete doesn't scrub etcd history (compaction/backup copies persist) — inherent to Kubernetes, not a PR defect; no action needed.

Principal engineer review

Concerns (none blocking):
1. CLI bookkeeping lives inside chart values. workspaceConfig.namespace is never read by the deployment templates (I checked templates/helm/{server,client}/deployment.yaml — they only consume .local/.startup); it's staging state smuggled into a Helm artifact. It's cohesive wipName/secretName recording and travels with the kit, so acceptable — but if this state grows, move it to a dedicated state file rather than accreting more non-chart keys into values.yaml.
2. No escape hatch from the prepare guard. Because state is written before apply, a (or a cluster that no longer exists) leaves a kit that deploy prepare refuses tooverwrite, and unstage needs working cluster access to clear it. The implicit workaround is hand-editing values.yaml; consider a --force or documented recovery path.
3. Ordering vs the installed release is docs-only. Running unstage while the releases out from under the parent pod (existing pods keep projected copies; restarts fail).A helm status check would couple the CLI to helm, so docs-only is a reasonable choice — just noting the sharp edge.
4. Cosmetic: cleanup_command always appends --kubectl kubectl even for the default;

PR #4878, opened 2026-07-09
[Main] ccwf swarm: fix SwarmServerConfig.min_clients default + start_task_timeout

nvidianz:
## Summary

- Forward-port the CCWF swarm fixes from #4568 to `main` for [NVBug 6433799](https://nvbugspro.nvidia.com/bug/6433799).
- Change `SwarmServerConfig.min_clients` from `None` to the controller-compatible default `int = 0`.
- Give the swarm and cross-site-evaluation startup tasks 300 seconds in the CIFAR-10 example.
- Add regression coverage for the default and explicitly configured `min_clients` values.

## Root cause and impact

`CCWFJob.add_swarm()` forwards `SwarmServerConfig.min_clients` to `SwarmServerController`. The configuration default was `None`, which overrides the controller's safe integer default and reaches `if min_clients < 0`, raising `TypeError` while the job is being constructed. As a result, swarm jobs that do not explicitly specify `min_clients` cannot start on `main`.

The example's 10-second default startup timeout can also expire while a client initializes CIFAR-10 and PyTorch, causing a `FATAL_SYSTEM_ERROR` despite continued client progress.

## Validation

- `python -m pytest -q tests/unit_test/app_common/ccwf` — 118 passed
- Python compilation check for all three changed files
- `git diff --check`

Forward-port of #4568 from the 2.8 branch.

### DIFF
--- a/examples/advanced/job_api/pt/swarm_script_runner_cifar10.py
+++ b/examples/advanced/job_api/pt/swarm_script_runner_cifar10.py
@@ -34,14 +34,24 @@ if __name__ == "__main__":
 
     aggregator = InTimeAccumulateWeightedAggregator(expected_data_kind=DataKind.WEIGHTS)
     job.add_swarm(
-        server_config=SwarmServerConfig(num_rounds=num_rounds),
+        server_config=SwarmServerConfig(
+            num_rounds=num_rounds,
+            # Default 10s is too short for clients that initialize CIFAR-10 dataset
+            # and the PyTorch model after responding to swarm_config.
+            start_task_timeout=300,
+        ),
         client_config=SwarmClientConfig(
             executor=ScriptRunner(script=train_script),
             aggregator=aggregator,
             persistor=PTFileModelPersistor(model=Net()),
             shareable_generator=SimpleModelShareableGenerator(),
         ),
-        cse_config=CrossSiteEvalConfig(eval_task_timeout=300),
+        cse_config=CrossSiteEvalConfig(
+            # Default 10s is too short for clients that need to (re)load CIFAR-10
+            # before responding to cse_start, same race as swarm_start above.
+            start_task_timeout=300,
+            eval_task_timeout=300,
+        ),
     )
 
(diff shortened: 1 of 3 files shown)


YuanTingHsieh (commented):
One non-blocking observation.

examples/advanced/job_api/pt/swarm_script_runner_cifar10.py
```diff
@@ -34,14 +34,24 @@
...
             executor=ScriptRunner(script=train_script),
             aggregator=aggregator,
             persistor=PTFileModelPersistor(model=Net()),
             shareable_generator=SimpleModelShareableGenerator(),
         ),
-        cse_config=CrossSiteEvalConfig(eval_task_timeout=300),
+        cse_config=CrossSiteEvalConfig(
+            # Default 10s is too short for clients that need to (re)load CIFAR-10
+            # before responding to cse_start, same race as swarm_start above.
+            start_task_timeout=300,
```
YuanTingHsieh: `start_task_timeout` is a no-op for CSE here. `CrossSiteEvalServerController` passes `starting_client=""` with `DefaultValuePolicy.EMPTY`, while `ServerSideController.control_flow()` only creates the start task under `if self.starting_client:`. As a result, no `cse_start` task is dispatched, so this setting cannot mitigate the race described in the comment. Could we remove this override/comment, or adjust the timeout for the phase that actually performs the initialization (configuration or evaluation) if that is the observed failure?

YuanTingHsieh (dismissed):

examples/advanced/job_api/pt/swarm_script_runner_cifar10.py
nvidianz: Addressed in be4787d49 by removing the ineffective CrossSiteEvalConfig start_task_timeout override and its misleading comment. The swarm start_task_timeout remains because the swarm flow does dispatch the start task.

YuanTingHsieh (approved):

PR #4883, opened 2026-07-09
Add recipe secret parameter safeguards

IsaacYangSLA:
## Summary

- Establishes and documents the contract that recipe parameters must never contain actual secret values.
- Adds masked potential-secret detection for recipe parameters and generated job artifacts.
- Adds environment-variable and mounted-file secret references with late runtime resolution at supported boundaries.
- Warns when references are used outside supported runtime consumers.

## User impact

Recipe authors can keep credentials in site environment variables or mounted secret files while exporting shareable job definitions containing only placeholders. Warnings never echo detected values, and runtime resolution is limited to explicit script/subprocess argument and configuration-reader boundaries.

## Validation

- Feature suite: 712 passed, 18 skipped
- `./runtest.sh -s`: Black, isort, flake8, and agent-skill lint passed
- `git diff --check` passed

### DIFF
--- a/nvflare/app_common/launchers/subprocess_launcher.py
+++ b/nvflare/app_common/launchers/subprocess_launcher.py
@@ -25,7 +25,7 @@ from nvflare.apis.shareable import Shareable
 from nvflare.apis.signal import Signal
 from nvflare.app_common.abstract.launcher import Launcher, LauncherRunStatus
 from nvflare.fuel.utils.log_utils import get_obj_logger
-from nvflare.fuel.utils.secret_utils import resolve_secret_refs, split_command_preserving_secret_refs
+from nvflare.fuel.utils.secret_utils import has_secret_refs, resolve_secret_refs, split_command_preserving_secret_refs
 from nvflare.utils.job_launcher_utils import add_custom_dir_to_path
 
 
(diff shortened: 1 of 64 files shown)


nvflare/app_common/launchers/subprocess_launcher.py
```diff
@@ -169,7 +169,12 @@ def _start_external_process(self, fl_ctx: FLContext):
                 app_custom_folder = workspace.get_app_custom_dir(job_id)
                 add_custom_dir_to_path(app_custom_folder, env)
 
-                command_seq = shlex.split(command)
+                # Resolve ${secret:ENV_VAR} references from this site's environment after shlex
+                # splitting, so injected values never re-tokenize. The resolved values exist only
+                # in the argument vector of the subprocess and must never be logged.
+                command_seq = [
+                    resolve_secret_refs(token) for token in split_command_preserving_secret_refs(command, posix=True)
+                ]
```
YuanTingHsieh: **[P1] Reject secret resolution inside nested interpreter command strings**

Resolving after the outer `shlex` split prevents Python from re-tokenizing the value, but it does not protect an argument that is subsequently parsed by `sh -c`, `bash -c`, or PowerShell `-Command`. For example, with a command using `sh -c` and a synthetic secret value containing `$(printf INJECTED >&2)`, the current code passes the resolved string to `sh`, which executes the substitution despite `shell=False`. Ordinary passwords containing quotes, `$()`, or semicolons can therefore be corrupted or interpreted as commands. Please reject secret references in nested interpreter command strings, or deliver those secrets through the child environment/read them inside the launched program.

nvflare/fuel/utils/secret_utils.py
```diff
@@ -0,0 +1,716 @@
...
+        if end < len(command) and marker_prefix in command[index + 1 : end]:
+            marker = f"{quoted_prefix}{len(quoted_values)}__"
+            quoted_values.append((marker, command[index + 1 : end]))
+            parts.append(marker)
+            index = end + 1
+        else:
+            # Keep non-secret-ref quoting byte-for-byte compatible with legacy splitting.
+            span_end = end + 1 if end < len(command) else len(command)
+            parts.append(command[index:span_end])
+            index = span_end
```
YuanTingHsieh: **[P2] Continue scanning after an unrelated unmatched quote**

This branch advances `index` to the end of the command when a quote has no matching closer. Consequently, an unrelated apostrophe prevents a later quoted secret composite from being shielded. For example, `--owner O'Reilly --authorization "Bearer ${secret:TOKEN}"` currently becomes five arguments: `--owner`, `O'Reilly`, `--authorization`, `"Bearer`, and the resolved value with a trailing quote, rather than one authorization value. Treat an unmatched quote as an ordinary character here (or use a token-aware parser) and add this case as a runtime regression test. The recently added lenient tokenizer only affects secret detection, not this execution path.

YuanTingHsieh (commented):
Two runtime correctness/security issues still reproduce on this head. The focused non-optional suite passed (519 passed, 7 skipped), but these scenarios are not covered by it.

YuanTingHsieh (commented):
## Review

One behavioral contract change worth flagging, and one minor warning-duplication note.

### P2 — `get_job_config_value` now raises instead of returning `default` on missing secrets

`nvflare/utils/configs.py` line 66 moves `resolve_secret_refs` outside the broad `except Exception` block, so a `${secret:MISSING_VAR}` reference now raises `ValueError` instead of silently returning `default`.

The docstring correctly documents the new `Raises:` contract, but existing callers that don't expect exceptions will break silently in the field.  The only in-tree caller today is `client_api_launcher_executor.py:_get_client_config_override`, which checks for `_CONFIG_VALUE_MISSING` but does **not** wrap the call in a try/except.  A job that ships a `${secret:MY_API_KEY}` reference and then runs on a site where `MY_API_KEY` is absent will crash the executor at startup rather than falling back to the default and surfacing a clear job-level error.

Two options:
1. Catch `ValueError` inside `_get_client_config_override` and re-raise (or log) with a job-level error.
2. Catch `ValueError` inside `get_job_config_value` itself and return a sentinel/raise a more descriptive `RuntimeError` that includes the key name so operators know what's missing.

Either way, the silent-default path should not be an accident.

### P2 — `_warn_potential_secrets_in_params` fires on `__init__`, `export`, and `run`

`nvflare/recipe/spec.py` lines 257, 728, and 748 all call `_warn_potential_secrets_in_params()`.  Under the default `warnings` filter, the same warning fires only once per call-site, so `__init__` and `export` are two separate sites and both fire.  Under `simplefilters("always")` (e.g., in pytest), warnings fire every call, tripling the noise for any test that checks `pytest.warns`.

Calling it only in `export` and `run` (drop the `__init__` call) covers the only two paths where secrets actually leave the process — and matches what `_warn_potential_secrets_in_exported_job` already does.

YuanTingHsieh (commented):
**Update on my earlier review:** Both P2 items above have been addressed in subsequent commits.

1. **`get_job_config_value` contract** — the `resolve_refs=True/False` keyword parameter was added. `_get_client_config_override` now reads with `resolve_refs=False` and explicitly rejects secret refs with a clear error message rather than silently failing. This is the right design.

2. **Double-scan** — my claim that `_warn_potential_secrets_in_params` was called in `__init__` was wrong for the current code. It is only called in `export()` and `run()`. Disregard that item.

The security model is sound. Ready to merge from my side.

YuanTingHsieh (approved):

PR #4885, opened 2026-07-09
Fix FedJob export of parent helper modules

nvkevlu:
Fix FedJob export of parent helper modules.

### Description

Fixes FedJob export failing to include helper modules located one directory above a registered script.

For top-level scripts, the import fallback used a `0:-0` slice, producing an empty search path. The exporter now uses parent-directory traversal and always searches at least one level above after checking the script directory.

Added a regression test covering `proj/client.py` importing `custom_layers.py` from its parent directory.

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh:
I think the path handling here needs two follow-up fixes before merge:

1. **Normalize the registered script path before deriving `module_depth`.** A valid spelling such as `./client.py` currently becomes module `..client`, so the new loop climbs two levels and misses a helper located exactly one directory above. Please normalize `relative_script` (for example with `normpath`) before building `module`, and add a regression case for `./client.py`.

2. **Add explicit source/destination containment to import discovery.** This is partly pre-existing, but the PR broadens ancestor searching to top-level scripts. `_copy_source_file` currently removes only one leading dot from a relative import and converts the remaining dots into path separators. A multi-level relative import can therefore resolve outside the intended source root. In addition, `_get_custom_file` joins `custom_dir` with a derived `dest` without rejecting an absolute or escaping destination, so processing such a path can read or write outside the export directory.

Please parse relative-import levels correctly (preferably via `ast`), reject levels above the current package/source root, resolve and check real paths against explicit allowed roots, and enforce that every destination remains inside `custom_dir` with `commonpath`. Symlink escapes and source-equals-destination should also be rejected. This should not rely on `FL_PACKAGES` or a standard-library denylist as the security boundary.

Suggested tests: `./client.py`, excessive relative-import levels, absolute/out-of-root resolution, symlink escape, and source/destination aliasing.

YuanTingHsieh:
Thanks for the hardening update. The original `./client.py` and multi-dot path-escape issues are addressed, but I found two remaining regressions at `b9c4d68`:

1. **Distinct absolute scripts can silently overwrite each other.** When an absolute script is outside every `sys.path` entry, `_get_relative_script()` returns only its basename. Registering, for example, `/a/client.py` and `/b/client.py` maps both to `custom/client.py`; the second copy silently replaces the first. Because imported modules are deduplicated by module name, the result can also contain the second client with dependencies retained from the first. I reproduced this against the current head. Please retain a unique safe destination or detect differing source files that map to the same destination and raise rather than overwrite.

2. **Valid symlink aliases can be rejected.** `_is_path_within()` compares resolved paths, but `os.path.relpath(script_path, package_path)` uses the unresolved spellings. If `sys.path` contains a real directory and the registered absolute script reaches that same directory through a symlink alias, containment succeeds but `relpath()` produces a `../...` path, which is then rejected as invalid. Please use the same canonical paths for both containment and relative-path calculation. A regression test with an in-root symlink alias would cover this.

The new AST parsing and source/destination containment otherwise look good, and the direct arbitrary-path write is closed.

YuanTingHsieh (approved):

PR #4886, opened 2026-07-10
Persist study registry to workspace root for read-only staged local dirs

pcnudde:
Fixes FLARE-3036.

### Description

Runtime study mutations (`register_study`, `add_study_site`, `remove_study`, ...) persisted `study_registry.json` under `workspace/local`. In Kubernetes deployments the staged `local/` ConfigMap is mounted read-only, so every mutation failed with `Read-only file system` after successful authentication and authorization.

Mutations now persist the registry to the workspace root, which stays writable (the PVC mount in K8s). The provisioned copy under `local/` acts as a first-start seed only: reads prefer the workspace-root copy and fall back to the seed, so the first mutation starts from the provisioned studies and the root copy is authoritative from then on, surviving server restarts. This follows the existing convention of keeping runtime-mutable server state (snapshot storage, job storage, audit log) out of `local/`.

- Added `WorkspaceConstants.STUDY_REGISTRY_CONFIG` and `Workspace.get_study_registry_file_path()` (root copy if present, else `local/` seed).
- `StudyCommandModule._with_mutation` loads via the resolver and always writes to the workspace root; the atomic temp-file replace is unchanged.
- Server startup load (`security_init`) uses the same resolver.
- New tests cover resolver precedence and a mutation against a read-only `local/` dir that seeds from `local/` and persists to the root — the previously untested combination of a staged read-only site config dir with a runtime study mutation.
- Docs: multi-study guide and multistudy CLI design doc now state where the registry persists.

No deploy/Helm changes are needed; `deploy k8 stage` continues to bundle the provisioned registry into the ConfigMap, where it serves as the first-start seed. `study_runtime.yaml` (#4872) is unaffected: it is static site-owned config that is never written at runtime and intentionally stays in `local/`.

### Types of changes
- [x] Documentation updated.

### DIFF
--- a/docs/design/multistudy_cli.md
+++ b/docs/design/multistudy_cli.md
@@ -55,6 +55,8 @@ The shipped multi-study design stores named-study state in `study_registry.json`
 
 The backend for all mutations is the same `study_registry.json` file the server loads at startup. The server applies mutations through a serialized validate-write-publish flow so the runtime registry and the persisted file remain aligned under normal operation and fail closed on validation/write errors.
 
+The persisted registry lives at the server workspace root (`<workspace>/study_registry.json`). A provisioned copy under `local/` acts as a first-start seed only: once a runtime mutation is persisted, the root copy is authoritative and shadows the seed. Writes never target `local/`, which may be a read-only mount (e.g. Kubernetes deployments that stage `local/` as a ConfigMap).
+
 ---
 
 ## Core Principles
--- a/docs/user_guide/admin_guide/multi_study_guide.rst
+++ b/docs/user_guide/admin_guide/multi_study_guide.rst
@@ -158,11 +158,19 @@ radius — use separate NVFlare deployments.
 Updating Studies
 ================
 
-Studies can be created, updated, and removed at runtime using the ``nvflare study`` command family
-without reprovisioning or restarting the server: ``register`` creates or merges a study,
-``add-site`` / ``remove-site`` change site enrollment, ``add-user`` / ``remove-user`` manage
-study-user membership, and ``remove`` deletes a study. See :ref:`study_command` for the full
-command reference.
+The ``nvflare study`` command family manages the server-side study registry at runtime, without
+reprovisioning or restarting the server: ``register`` creates or merges a study, ``add-site`` /
+``remove-site`` change site enrollment, ``add-user`` / ``remove-user`` manage study-user
+membership, and ``remove`` deletes a study. See :ref:`study_command` for the full command
+reference.
+
+These commands change only the server's view of a study — enrollment, login membership, and job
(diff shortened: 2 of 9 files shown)


YuanTingHsieh (approved):

chesterxgchen (commented):
The impl looks fine. but the question is

why  do we need this local seed to workspace copy step ? whats the benefits with this extra complexity

pcnudde:
@chesterxgchen Good question — to be clear, there is no copy step; nothing is ever copied. The resolver is read-only fallback: use the root copy if it exists, otherwise read the provisioned file under `local/`.

The fallback earns its keep three ways:

1. **project.yml bootstrapping.** Provisioning writes `study_registry.json` into the kit's `local/` when `project.yml` declares `studies:`. In K8s the staged `local/` ConfigMap is the only channel provisioned content reaches the pod — the PVC starts empty. Without the fallback read, a deployment provisioned with studies would boot with an empty registry and those studies would be unusable until re-registered by hand.

2. **Upgrade compatibility.** Before this PR, runtime mutations wrote to `local/study_registry.json` (bare-metal/POC). On first boot after upgrading, that existing state is picked up by the fallback, and the next mutation persists it to the root. No migration step.

3. **Writes have to leave `local/` regardless** — K8s mounts it read-only, which is the bug (FLARE-3036).

So the alternative to the fallback isn't less complexity — it's dropping support for `studies:` in project.yml and losing existing registry state on upgrade. The entire cost is the small resolver in `Workspace`: root if present, else seed, else default to root.

chesterxgchen:
> @chesterxgchen Good question — to be clear, there is no copy step; nothing is ever copied. The resolver is read-only fallback: use the root copy if it exists, otherwise read the provisioned file under `local/`.
> 
> The fallback earns its keep three ways:
> 
> 1. **project.yml bootstrapping.** Provisioning writes `study_registry.json` into the kit's `local/` when `project.yml` declares `studies:`. In K8s the staged `local/` ConfigMap is the only channel provisioned content reaches the pod — the PVC starts empty. Without the fallback read, a deployment provisioned with studies would boot with an empty registry and those studies would be unusable until re-registered by hand.
> 2. **Upgrade compatibility.** Before this PR, runtime mutations wrote to `local/study_registry.json` (bare-metal/POC). On first boot after upgrading, that existing state is picked up by the fallback, and the next mutation persists it to the root. No migration step.
> 3. **Writes have to leave `local/` regardless** — K8s mounts it read-only, which is the bug (FLARE-3036).
> 
> So the alternative to the fallback isn't less complexity — it's dropping support for `studies:` in project.yml and losing existing registry state on upgrade. The entire cost is the small resolver in `Workspace`: root if present, else seed, else default to root.

Ok, make sense

chesterxgchen (approved):

PR #3890, opened 2025-12-11
Collab API: pythonic collaborative FL programming

yanchengnv:
Fixes # .

### Description

Introduces the **Collab API** (formerly FOX / federated object exchange) — a pythonic way to develop FL applications: users write plain Python objects/modules with `@collab.main` (server orchestration) and `@collab.publish` (client methods), and the framework handles distribution, communication, and orchestration.

**Package layout** (`nvflare/collab/`):
- Top-level user API (lazily exported): `from nvflare.collab import collab, CollabRecipe, InProcessEnv, MultiProcessEnv`
- `api/` — programming surface: `collab` facade, decorators, App/Proxy/Group, `ModuleWrapper` (module-based, no-class usage)
- `recipe.py` — `CollabRecipe`, the job definition entry point
- `runtime/` — internal execution stack: abstract `Backend` + `LocalBackend` / `SubprocessBackend` / `FlareBackend`, with per-mode machinery in `runtime/local/` (`AppRunner`, `InProcessRunner`, `InProcessEnv`) and `runtime/flare/` (`CollabController`, `CollabExecutor`, `CollabWorker`, `MultiProcessEnv`)
- `tracking/` — TensorBoard/MLflow/W&B-compatible writers with automatic in-process/subprocess mode detection
- Examples live at `examples/collab/` (np + pt trees; run with `examples/` on `PYTHONPATH`)

**Highlights**:
- `CollabRecipe` built on the Recipe spec: caller-module auto-detection, module auto-wrapping, per-site config delivered as per-site app props (`collab.get_app_prop`), exec-mode options (`inprocess`, `run_cmd`, auto-detected `training_module`), idempotent `finalize()`
- Execution environments by topology: `InProcessEnv` (one process, direct calls), `MultiProcessEnv` (POC, real FLARE processes over CellNet); production deployment planned as `MultiHostEnv`
- **Unified Client API**: training scripts use the standard `import nvflare.client as flare` (`receive()`/`send()`); new `COLLAB_IN_PROCESS_API` / `COLLAB_SUBPROCESS_API` types route via contextvars for per-client isolation
- **Subprocess / multi-GPU (DDP)**: executor-managed worker lifecycle via `torchrun`; only rank 0 communicates with the FL system
- **Security integration**: collab jobs ship user server/client source in the app `custom/` folder (BYOC), so they are authorized via the BYOC right rather than the site class allow-list; FOBS tensor decomposers are registered in all execution paths
[...]

PR #4901, opened 2026-07-14
Collab API: pythonic collaborative FL programming

nvidianz:
### Description

Introduces the **Collab API** — a Pythonic way to build federated and distributed workflows. Users publish ordinary Python functions or methods with `@collab.publish`, define exactly one server workflow with `@collab.main`, and call remote sites through normal-looking Python proxies.

### Public contract

- `from nvflare.collab import CollabCallError, CollabRecipe, collab, simple_logging`
- `CollabRecipe` finalizes to a normal `FedJob`
- Jobs execute with the standard `SimEnv`, `PocEnv`, and `ProdEnv` classes from `nvflare.recipe`
- Server and client application code runs inside its FLARE site process
- Context, workspace, live `FLContext`, application properties, abort state, and per-site configuration are available to Collab functions
- External worker processes, Client API bridging, and multi-rank/DDP execution are intentionally deferred

### Package layout

- `nvflare/collab/api/` — decorators, application objects, proxies, groups, contexts, call options, and the private invocation contract
- `nvflare/collab/recipe.py` — `CollabRecipe` to `FedJob` integration
- `nvflare/collab/runtime/` — controller, executor, CellNet dispatch, and lifecycle integration

### Examples

The self-contained examples under `examples/collab/` cover:

- `hello_numpy_collab` — NumPy training with the standard Recipe API and `SimEnv`
- `hello_fedavg` — one-file PyTorch FedAvg and per-site configuration
- `async_aggregation` — in-time aggregation with response callbacks
- `swarm` — decentralized client-to-client learning

Each example owns the trainer, strategy, and utility modules that it uses; there is no shared example `common` package.

### Validation

- Collab unit suite: 86 tests passed
- Scoped Black, isort, and flake8 checks passed
- GitHub Actions style, license, link, analysis, wheel-build, and Python 3.10–3.14 unit-test jobs passed on the current head
- Documentation and example imports match the current public API

### Types of changes

- [x] Non-breaking change (new API)
- [x] New tests added
- [x] Inline docstrings updated
- [x] Documentation updated

Supersedes #3890.


nvidianz:
@greptile-apps review

nvflare/collab/core/in_process_env.py
```diff
@@ -0,0 +1,164 @@
...
+from typing import Dict, Optional, Tuple, Union
+
+from nvflare.collab.runtime.local.runner import InProcessRunner
+from nvflare.job_config.api import FedJob
+from nvflare.recipe.spec import ExecEnv
+
+WORKSPACE_ROOT = "/tmp/nvflare/collab_local"
+
+
+class InProcessEnv(ExecEnv):
```
YuanTingHsieh: **[P1] Use the existing Recipe environments instead of introducing Collab-specific environments**

Collab should produce a normal `FedJob` containing `CollabController` and `CollabExecutor`, then run through the existing `SimEnv`, `PocEnv`, or `ProdEnv`. Job deployment topology is orthogonal to whether a Collab object executes in-process or in an external process.

`MultiProcessEnv` substantially duplicates `PocEnv`, while `InProcessEnv` bypasses the finalized job by receiving raw Collab objects and execution settings through `CollabRecipe.process_env()`. This couples the recipe to a Collab-specific environment contract.

Please remove the public Collab environment classes. Preserve the useful direct-call implementation as an internal `CollabSimulationAdapter`, selected through an explicit mode on the existing `SimEnv`, for example `simulation_mode="runtime"` versus `"semantic"`. The existing runtime simulator remains useful for Controller/Executor, Cell, serialization, and packaging conformance.

nvflare/collab/api/backend.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+class _SingleTargetCallContext(GroupCallContext):
+    """Group call context that leaves direct-call result filtering to Proxy."""
+
+    def set_result(self, result):
+        self.waiter.set_result(self.target_name, result)
+
+
+class Backend(ABC):
```
YuanTingHsieh: **[P1] Separate logical call transport from target execution**

The current `Backend` hierarchy conflates two different layers. `LocalBackend` and `FlareBackend` select how a logical Collab call reaches a participant, while `SubprocessBackend` selects how that participant executes the call after it arrives. In production a call can use both Cell transport and external-process execution, so these are not interchangeable backends.

Collab is always an NVFlare feature, so this should not be a public or platform-neutral backend API. Please keep a private invocation dispatcher for direct-versus-Cell calls (for example `DirectDispatcher` and `CellDispatcher`), and model in-process-versus-external-process separately behind Controller/Executor as execution adapters. `Backend`/`BackendType` should not leak into user algorithm behavior.

nvflare/collab/core/recipe.py
```diff
@@ -0,0 +1,345 @@
...
+        job_name: str,
+        server: Optional[object] = None,
+        client: Optional[object] = None,
+        server_objects: Optional[Dict[str, object]] = None,
+        client_objects: Optional[Dict[str, object]] = None,
+        sync_task_timeout=5,
+        max_call_threads_for_server=100,
+        max_call_threads_for_client=100,
+        min_clients: int = 1,
+        inprocess: bool = True,
```
YuanTingHsieh: **[P1] Model execution as a typed per-site policy, not global recipe subprocess flags**

`inprocess`, `run_cmd`, `training_module`, and `subprocess_timeout` apply globally to every client, while the server cannot independently select external-process execution. This prevents heterogeneous sites and freezes launcher-specific details into the public recipe API.

Execution should be per-site recipe metadata interpreted by `CollabRecipe`, with in-process as the default and independent server/client selection. For example, use `set_per_site_config(..., {"execution": external_process(...)})`, then let Controller/Executor select the appropriate execution adapter. Avoid making a free-form `run_cmd` string and training-module autodetection the durable public contract.

nvflare/collab/runtime/worker/worker.py
```diff
@@ -0,0 +1,536 @@
...
+        except Exception as e:
+            self.logger.exception(f"Worker error: {e}")
+            raise
+        finally:
+            self._cleanup()
+
+    def _start_collab_api_mode(self):
+        """Start in Collab API mode - rank 0 waits for RPC calls."""
+        # Only rank 0 communicates with parent in DDP mode
+        if self.rank != 0:
```
YuanTingHsieh: **[P1] External-process mode does not preserve Collab lifecycle or multi-rank semantics**

The worker constructs a fresh class with no recipe constructor arguments, while `@collab.init` and `@collab.final` run on the different object living in the job process. In torchrun mode this branch also leaves non-control ranks waiting only for OS shutdown; they never invoke the published method. A rank-0 method entering a DDP collective will therefore deadlock.

External execution needs a defined worker-runtime contract: export and reconstruct the configured Collab object, run init/method/final on every participating rank, broadcast ordered invocation envelopes, gather all-rank completion/error status, and return the Collab-visible result only from global rank 0. It also needs session binding, call IDs, abort/shutdown, and process-group supervision. Until that contract is implemented and tested, this PR should not advertise Collab DDP support; at most, scope external Collab execution to an explicitly experimental single-process mode.

YuanTingHsieh (commented):
Architectural review of the Collab execution model. The four inline comments cover environment ownership, invocation-vs-execution layering, execution-policy configuration, and the external-process/DDP runtime contract.

nvflare/collab/core/in_process_env.py
nvidianz: Addressed in `055eaa050`. `CollabRecipe` now always finalizes to a normal `FedJob`, and the examples run it through the existing `SimEnv`, `PocEnv`, or `ProdEnv`. I removed the public Collab-specific environment classes and the `process_env()` coupling. The direct-call runner remains only under the internal `runtime/local` package for focused semantic tests.

nvflare/collab/api/backend.py
nvidianz: Addressed in `055eaa050`. The public `Backend` and `BackendType` contracts/exports are gone. Logical invocation now uses the private `api/_invocation.py` dispatcher abstraction, with internal `DirectDispatcher` and `CellDispatcher` strategies. The subprocess execution object was removed instead of remaining a transport sibling.

nvflare/collab/core/recipe.py
nvidianz: Addressed in `055eaa050`. I removed the global `inprocess`, `run_cmd`, `training_module`, and `subprocess_timeout` fields from `CollabRecipe` and `CollabExecutor`. The current contract is site-process execution in the standard FLARE environments; the design doc explicitly defers external execution until it has a typed, per-site policy rather than a free-form recipe command.

nvflare/collab/runtime/worker/worker.py
nvidianz: Addressed in `055eaa050`. I removed the external worker/launcher path, subprocess dispatch and tracking relay, and the DDP example/documentation. `CollabClientAPI` now runs its registered training function in the site's FLARE process and rejects nonzero ranks, so this PR no longer advertises a lifecycle or multi-rank contract it cannot preserve.

nvidianz:
@greptile-apps review

nvidianz:
@greptile-apps review

nvflare/collab/runtime/flare/adaptor.py
```diff
@@ -0,0 +1,111 @@
...
+        self.props = props
+        self.resource_dirs = resource_dirs
+        self.publish_obj_ids = collab_obj_ids
+        self.incoming_call_filters = incoming_call_filters
+        self.outgoing_call_filters = outgoing_call_filters
+        self.incoming_result_filters = incoming_result_filters
+        self.outgoing_result_filters = outgoing_result_filters
+
+    def process_config(self, app: App, fl_ctx: FLContext):
+        app.update_props(self.props)
```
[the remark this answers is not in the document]
nvidianz: Fixed in `73f2e401d` by guarding `set_resource_dirs` when `resource_dirs` is `None`. Added `test_process_config_with_default_resource_dirs`; all 29 Collab unit tests pass.

nvflare/collab/runtime/flare/adaptor.py
```diff
@@ -0,0 +1,111 @@
...
+        app.set_resource_dirs(self.resource_dirs)
```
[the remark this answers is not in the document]
nvidianz: Fixed in `73f2e401d` by guarding `set_resource_dirs` when `resource_dirs` is `None`. Added `test_process_config_with_default_resource_dirs`; all 29 Collab unit tests pass.

nvidianz:
@greptile-apps

nvidianz:
@greptile-apps review

nvidianz:
@greptile-apps review

nvidianz:
@greptile-apps review

nvflare/collab/runtime/local/direct_dispatcher.py
```diff
@@ -0,0 +1,136 @@
...
+        self.executor.submit(self._run_func_in_group, gcc, func_name, args, kwargs)
+
+    def _run_func_in_group(self, gcc: GroupCallContext, func_name, args, kwargs):
+        try:
+            target_name = gcc.target_name
+            result = self._call_target(gcc.context, target_name, gcc.call_opt, func_name, *args, **kwargs)
+            gcc.send_completed()
+            gcc.set_result(result)
+        except Exception as ex:
+            gcc.set_exception(ex)
```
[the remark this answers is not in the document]
nvidianz: Fixed in d17c1d76f. `DirectDispatcher._run_func_in_group()` now calls `send_completed()` from `finally`, matching the production dispatcher, with success and exception regression coverage.

nvidianz:
@greptile-apps review

nvidianz:
@greptile-apps review

nvflare/collab/runtime/client_api.py
```diff
@@ -0,0 +1,385 @@
...
+
+        if not self._training_func:
+            raise RuntimeError("CollabClientAPI requires a training function")
+
+        self._current_model = fl_model
+        self._current_task_name = task_name
+        self._result = None
+
+        # Scope module-level Client API calls to this embedded API instance.
+        with use_api(self):
```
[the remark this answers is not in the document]
nvidianz: Fixed in 1669127b1. `execute()` now resets `_receive_called` for every non-stop task, and a two-round regression verifies that `send(clear_cache=False)` in round one cannot bypass `receive()` in round two.

nvidianz:
@greptile-apps review

nvflare/collab/api/proxy.py
```diff
@@ -0,0 +1,264 @@
...
+            # exception remains the call result so the caller can recover.
+            if backend and not call_opt.optional:
+                try:
+                    backend.handle_exception(ex)
+                except Exception as ex2:
+                    # ignore exception from backend handling
+                    self.logger.error(f"ignored backend's exception {type(ex2)}")
+
+            # Must return the exception as the result of the func call.
+            # Do NOT raise it!
```
holgerroth: **[P1] Remote exceptions are returned as values while the job is simultaneously panicked — workflow code can neither catch nor recover from a failure**

`call_func` deliberately returns the exception object ("Do NOT raise it!") for non-optional calls, but by this point `backend.handle_exception()` has already triggered `system_panic` (see `cell_dispatcher.py:141`). The combination means a `try/except` around `client.train(...)` can never catch a remote failure: the exception arrives as a return value, and the whole federation is already aborting. A transient `ValueError` in one client's `train()` therefore kills the job with no recovery path except `optional=True`.

It also makes error handling inconsistent: `Group` raises on local dispatch errors, single calls return exceptions, and even this PR's own examples get it wrong — `call_patterns/avg_seq.py:66-72` multiplies an unchecked `optional=True` result (`result * self.client_weights[...]` → `TypeError` when a client times out), and both `cyclic.py` strategies treat only `None` as failure, so an Exception instance would be relayed to the next client as the "model".

Suggestion: pick one contract and enforce it everywhere — most Pythonic would be raise-by-default (letting the workflow catch it, panicking only if it propagates out of `@collab.main`), with returned exceptions reserved for `optional=True`. Whatever the decision, the examples should demonstrate the correct failure handling.

nvflare/collab/runtime/cell_dispatcher.py
```diff
@@ -0,0 +1,141 @@
...
+            # makes this idempotent when CellNet already invoked the callback.
+            gcc.send_completed()
+
+    def _msg_sent(self, gcc: GroupCallContext):
+        gcc.send_completed()
+
+    def handle_exception(self, exception: Exception):
+        fl_ctx = self.engine.new_context()
+        secure_log_traceback(self.logger)
+        self.manager.system_panic(f"exception occurred: {exception}", fl_ctx)
```
holgerroth, line 158: **[P1] `handle_exception` → `system_panic` preempts any workflow-level error handling**

Companion to the comment on `proxy.py:252`: because every non-optional remote exception lands here, the workflow author has no opportunity to handle a per-site failure before the job aborts. Consider panicking only when the exception escapes the user's `@collab.main` function, so per-call failures stay recoverable.

nvflare/collab/tracking/_transport/auto.py
```diff
@@ -0,0 +1,130 @@
...
+
+        from .in_process import get_inprocess_writer
+
+        self._delegate = get_inprocess_writer()
+        if self._delegate:
+            self.logger.debug("AutoWriter using InProcessWriter")
+        else:
+            self.logger.debug("InProcessWriter not available - metrics may not be logged")
+
+        self._initialized = True
```
holgerroth: **[P1] The tracking transport is never wired — every metric logged via `nvflare.collab.tracking` is silently discarded**

Nothing in this PR instantiates `InProcessWriter` or calls `set_inprocess_writer()`, and no component subscribes to `TOPIC_LOG_DATA = "analytics_log_data"` (the existing pipeline consumes `"LOG_DATA"` — see `nvflare/client/in_process/api.py:33`). So `_ensure_initialized` always finds no delegate, latches `_initialized = True`, and drops all data permanently — a public drop-in API that is a silent no-op.

Even once wired, the event mapping diverges from the `AnalyticsData`/receiver contract in `nvflare/client/tracking.py`: no `writer=LogWriterName.X` stamp, `mlflow.log_metric` emits `SCALAR` (which `MLflowReceiver.buffer_data` drops — it only handles `METRIC`), step is sent as `step` instead of `global_step` (all steps lost), and `log_text` omits the `path` kwarg the receiver requires. There are also no unit tests for the subpackage.

Since the working Collab metrics path is `CollabClientAPI.set_log_handler` → `flare.log`, I'd suggest dropping `nvflare/collab/tracking` from this PR and landing it in a follow-up once it's wired and aligned with the existing tracking contract.

nvflare/collab/tracking/wandb.py
```diff
@@ -0,0 +1,263 @@
...
+    hidden: Optional[bool] = None,
+    summary: Optional[str] = None,
+    goal: Optional[str] = None,
+    overwrite: Optional[bool] = None,
+):
+    """Define a custom metric (module-level function)."""
+    _get_or_create_writer().define_metric(name, step_metric, step_sync, hidden, summary, goal, overwrite)
+
+
+def finish(exit_code: Optional[int] = None, quiet: Optional[bool] = None):
```
holgerroth: **[P1] `wandb.finish()` crashes — `WandbWriter` has no `close()` method**

`finish()` calls `writer.close()`, but `WandbWriter` doesn't extend `BaseWriter` and defines no `close`. The module's own documented drop-in pattern (`wandb.init(...)` → `wandb.log(...)` → `wandb.finish()`) raises `AttributeError: 'WandbWriter' object has no attribute 'close'` at the end of training (verified by executing the module). Also, `wandb.init(config={...})` never populates the module-level `config` object, so `wandb.config.lr` silently returns `None` where real wandb returns the value.

nvflare/collab/api/group.py
```diff
@@ -0,0 +1,217 @@
...
+                            func_proxy.caller_name, func_proxy.name, target_group=self, set_call_ctx=False
+                        )
+
+                        gcc = GroupCallContext(
+                            app=self._app,
+                            target_name=func_proxy.target_name,
+                            call_opt=self._call_opt,
+                            func_name=func_name,
+                            process_cb=self._process_resp_cb,
+                            cb_kwargs=self._cb_kwargs,
```
holgerroth: **[P1] All `GroupCallContext`s in a fan-out share one mutable `cb_kwargs` dict — concurrent result delivery races**

Every target's `GroupCallContext` receives the same `self._cb_kwargs` instance, and each result-delivery thread mutates it in `set_result` (`group_call_context.py:250`: `self.cb_kwargs[CollabMethodArgName.CONTEXT] = ctx`) before invoking `process_cb(**self.cb_kwargs)`. When two sites' results arrive concurrently on different backend threads, site A's callback can receive site B's context (wrong caller/callee/props), or the concurrent set/pop corrupts the kwargs. Each `GroupCallContext` should get its own shallow copy of `cb_kwargs`.

nvflare/collab/api/group_call_context.py
```diff
@@ -0,0 +1,274 @@
...
+            original_caller = ctx.caller
+            ctx.caller = ctx.callee
+            ctx.callee = original_caller
+
+            if not isinstance(result, Exception):
+                set_call_context(ctx)
+                try:
+                    result = self.app.apply_incoming_result_filters(self.target_name, self.func_name, result, ctx)
+                    if self.process_cb:
+                        self.cb_kwargs[CollabMethodArgName.CONTEXT] = ctx
```
holgerroth: **[P1] `set_result` leaks call context onto backend threads and silently swallows callback/filter exceptions**

Two problems in this block:
1. The `finally: set_call_context(self.context)` (line 255) doesn't restore the delivery thread's *previous* context — it installs this call's context on the backend thread, so unrelated work later scheduled on that thread sees a stale `collab.context`. Save `get_call_context()` on entry and restore that instead.
2. `except Exception as ex: result = ex` (line 257) converts exceptions raised by result filters or `process_resp_cb` into queue items with no logging — the user only discovers a broken callback if they isinstance-check every result. At minimum this should log the traceback.

Related: `ResultWaiter.set_result` does no dedup/membership check against `self.sites`, so a duplicate reply from one site double-increments the completion count and releases the waiter before slower sites respond.

nvflare/collab/runtime/flare/controller.py
```diff
@@ -0,0 +1,264 @@
...
+        return proxy
+
+    def control_flow(self, abort_signal: Signal, fl_ctx: FLContext):
+        # configure all sites
+        server_collab_interface = self.server_app.get_collab_interface()
+        task_data = Shareable({SyncKey.COLLAB_INTERFACE: server_collab_interface})
+        task = Task(
+            name=SYNC_TASK_NAME,
+            data=task_data,
+            timeout=int(self.sync_task_timeout),
```
holgerroth: **[P1] `int(self.sync_task_timeout)` truncates fractional timeouts to 0 = "never time out"; sync is also all-or-nothing with a 5s default**

`sync_task_timeout` is validated with `check_positive_number` (`recipe.py:46`), which accepts floats, but the `Task` is created with `timeout=int(self.sync_task_timeout)` — any value in (0, 1) floors to 0, which `Task` treats as no timeout. `CollabRecipe(..., sync_task_timeout=0.5)` plus one unresponsive client means `broadcast_and_wait(min_responses=num_clients)` blocks forever and the job never starts or aborts. Use `math.ceil` or validate the value as `>= 1`.

Separately, the sync at lines 205-224 requires *all* clients to respond within the window (default 5s) or the whole job `system_panic`s — no retry, no partial participation. The client task-fetch interval alone is 0.5s, and real client startup commonly exceeds 5s, so a single slightly-slow but healthy client kills the run. Consider a more generous default and/or a retry loop before panicking.

holgerroth (commented):
Detailed review of the Collab API and runtime layers. The four high-priority themes are covered in the inline comments below: (1) remote-exception semantics (returned as values + auto-panic, so workflows cannot catch failures), (2) the tracking subpackage's transport is unwired so metrics are silently dropped, (3) thread-safety races in the group-call machinery, and (4) controller sync timeout truncation and all-or-nothing sync.

nvidianz:
@greptile-apps

nvflare/collab/api/proxy.py
holgerroth: A concrete proposal to resolve this, deliberately minimal for v1:

1. **Raise by default.** A failed call raises a single exception type, `CollabCallError(site, func_name, cause)`, carrying the remote exception's type and traceback text (timeouts distinguishable via `e.cause` or one `CollabCallTimeout` subclass — no deeper hierarchy needed yet).
2. **Panic only on escape.** `run_server` wraps the `@collab.main` invocation; an exception that propagates out uncaught triggers `system_panic`. This preserves today's fail-fast behavior for workflows that don't handle errors, while deleting the per-call `handle_exception` → panic path entirely.
3. **`optional=True` returns `None`** on failure (with a warning log) instead of the exception object — no `isinstance` checks anywhere.

The result is *less* mechanism than the current design (the "exceptions are values" rule and the parallel panic path both go away) and matches what Python programmers already know from `concurrent.futures` and Ray, where remote results raise on consumption:

```python
try:
    result = client.train(model)
except CollabCallError as e:
    logger.warning(f"{e.site} failed: {e.cause}")  # retry, skip, or re-raise
```

It would also fix the latent bugs in `avg_seq.py` and the two `cyclic.py` strategies as a side effect, since the failure path can no longer flow silently into arithmetic.

nvflare/collab/api/group_call_context.py
holgerroth: Building on the single-call proposal (`proxy.py:252` thread), a minimal way to make fan-out failures explicit without new API surface:

1. **One failure type everywhere.** A per-site failure in a group call surfaces as the same `CollabCallError` as single calls — never a raw returned exception, and never a silently converted filter/callback error (those should be logged and wrapped).
2. **Iterate successes; expose failures.** Iteration yields successful `(site, result)` pairs only, and the collection exposes `results.failures: dict[site, CollabCallError]`. Stragglers are the *normal* case in FL, so partial participation becomes the natural reading of the code rather than defensive boilerplate:

```python
results = collab.clients.train(model)
agg = weighted_average(dict(results))
if results.failures:
    logger.warning(f"aggregating without: {list(results.failures)}")
```

A workflow that requires all sites just checks `results.failures` and raises — no extra call options needed in v1. This also gives the non-blocking iterator a defined terminal state for a dead site (its entry lands in `failures` when the call timeout expires), which resolves the unbounded `q.get(block=True)` hang as part of the same contract.

nvflare/collab/api/proxy.py
nvidianz: Addressed in 7fac7dfdf. Required calls now raise the public CollabCallError(site, func_name, cause), including remote exception type and secure traceback metadata; optional=True logs a warning and returns None. The per-call handle_exception/panic path is removed, so only an uncaught exception escaping @collab.main reaches the existing run_server fail-fast handling. Examples, docs, and regression tests were updated.

nvflare/collab/runtime/cell_dispatcher.py
nvidianz: Addressed in 7fac7dfdf. CellDispatcher now returns structured CollabCallError failures to the workflow and no longer invokes handle_exception for an individual call. Workflow code can catch the error; the job panics only if the exception escapes the main Collab function.

nvflare/collab/tracking/_transport/auto.py
nvidianz: Addressed in 7fac7dfdf by removing the entire unwired nvflare.collab.tracking package and its public-import references. CollabClientAPI.set_log_handler remains the supported metrics path; tracking compatibility can return in a separately wired and tested follow-up.

nvflare/collab/tracking/wandb.py
nvidianz: Addressed in 7fac7dfdf. The unwired tracking package, including the incomplete wandb compatibility module, was removed from this PR and deferred to a properly wired/tested follow-up.

nvflare/collab/api/group.py
nvidianz: Addressed in 7fac7dfdf. Every GroupCallContext now receives its own shallow copy of cb_kwargs, and result processing creates a callback-local copy before adding context. A regression test verifies the caller dict is unchanged and the callback receives the correct per-call context.

nvflare/collab/api/group_call_context.py
nvidianz: Addressed in 7fac7dfdf. Result delivery now restores the backend thread's prior call context, logs and wraps callback/filter failures as CollabCallError, ignores duplicate/unexpected sites, yields successes only, and exposes per-site errors through results.failures. Failure markers and local timeout handling give non-blocking iteration a terminal state. Regression tests cover context restoration, failure wakeup, deduplication, structured failures, and timeouts.

nvflare/collab/runtime/flare/controller.py
nvidianz: Addressed in 7fac7dfdf. The sync Task timeout now uses math.ceil, so fractional positive values cannot become zero/no-timeout, and the default was increased from 5 to 60 seconds in both CollabRecipe and CollabController. Tests cover the default and a 0.5-second value becoming 1 second.

nvflare/collab/api/group.py
```diff
@@ -0,0 +1,217 @@
...
+                            waiter=waiter,
+                        )
+
+                        # try to get permission to make next call
+                        gcc.set_send_complete_cb(self._request_sent, gcc=gcc, proxy=func_proxy)
+                        waiter.wait_for_call_permission(max_parallel, self._abort_signal)
+
+                        # make next call
+                        waiter.inc_call_count()
+                        func_proxy.backend.call_target_in_group(gcc, func_name, *call_args, **call_kwargs)
```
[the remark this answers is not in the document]
nvidianz: Addressed in 7fac7dfdf. The group fan-out now passes copy.copy(self._cb_kwargs) to each GroupCallContext, which also uses a callback-local kwargs copy before injecting context. The concurrency-sensitive behavior is covered by a regression test.

nvidianz:
@greptile-apps review

docs/design/client_api_execution_modes_plan.md
YuanTingHsieh, line 1: can you un-check this file from this PR?

examples/collab/async_aggregation/async_aggregation.py
```diff
@@ -0,0 +1,53 @@
...
+The server aggregates each client result as it arrives instead of waiting for
+the complete group call to finish.
+
+Run from the ``examples`` directory:
+
+    python -m collab.async_aggregation.async_aggregation
+"""
+
+from collab.async_aggregation.avg_intime import NPFedAvgInTime
+from collab.async_aggregation.runner import make_parser, run_recipe
```
YuanTingHsieh: i don't think we need this additional wrapper "runner", suggest use SimEnv or SimEnv / PocEnv directly

examples/collab/call_patterns/trainer.py
```diff
@@ -0,0 +1,53 @@
...
+        self.logger = get_obj_logger(self)
+
+    @collab.init
+    def init_trainer(self):
+        delta_config = collab.get_app_prop("client_delta", {})
+        self.delta = delta_config.get(collab.site_name, self.delta)
+        self.logger.info(f"init_trainer: client {collab.site_name}: delta={self.delta}")
+
+    @collab.init
+    def init_trainer2(self):
```
YuanTingHsieh: what happen when we have 2 init?

examples/collab/client_api/in_process_client.py
```diff
@@ -0,0 +1,170 @@
...
+            outputs = model(batch_inputs)
+            loss = criterion(outputs, batch_labels)
+            loss.backward()
+            optimizer.step()
+            final_loss = loss.item()
+
+    return model.state_dict(), final_loss
+
+
+def training_loop():
```
YuanTingHsieh: why do we need to wrap this into a method?

examples/collab/call_patterns/widgets.py
```diff
@@ -0,0 +1,35 @@
...
+    def __init__(self):
+        self.logger = get_obj_logger(self)
+
+    @collab.publish
+    def accept_metric(self, metrics: dict):
+        self.logger.info(f"[{collab.callee}] received metric report from {collab.caller}: {metrics}")
+
+    @collab.init
+    def init(self):
+        collab.register_event_handler("metrics", self._accept_metric)
```
YuanTingHsieh: Did we decide to have collab events?

nvidianz:
@greptile-apps review

nvflare/client/api.py
```diff
@@ -43,6 +67,9 @@ def get_context(ctx: Optional[APIContext] = None) -> APIContext:
     """
     if ctx:
         return ctx
+    local_context = _current_context.get()
+    if local_context:
+        return local_context
     elif default_context:
         return default_context
     else:
```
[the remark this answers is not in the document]
nvidianz: Addressed in f12f01102. get_context() now checks the scoped _api_override before the process-wide default_context. When use_api() has cleared the current context and wrapped code does not call init() again, it lazily creates and caches an APIContext for the override in that ContextVar scope. The regression test starts with an existing default context and verifies the override wins and is initialized once.

nvflare/collab/api/group_call_context.py
```diff
@@ -0,0 +1,331 @@
...
+                return item
+            if all_received:
+                raise StopIteration()
+
+            # More outcomes are expected. Do not hold update_lock while
+            # waiting, because append() needs it to enqueue the next item.
+            item = self.q.get(block=True)
+            if item is not _FAILURE_MARKER:
+                return item
+
```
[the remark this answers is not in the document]
nvidianz: Addressed in f12f01102 at the cancellation source. The future completion callback now distinguishes cancellation: it cancels the timer, records a CancelledError on the GroupCallContext, and calls send_completed(), so ResultWaiter reaches a terminal outcome instead of leaving ResultQueue blocked. The regression test saturates a one-worker executor, cancels queued futures through shutdown(cancel_futures=True), and verifies the site completes with CancelledError.

nvidianz:
@greptile-apps review

examples/collab/swarm/swarm_algo.py
```diff
@@ -0,0 +1,125 @@
...
+        collab.register_event_handler("final_model", self._save_final_model)
+
+    @collab.publish
+    def train(self, weights, current_round):
+        self.logger.info(f"[{collab.call_info}]: train asked by {collab.caller}: {current_round=}")
+        return weights + self.delta
+
+    def sag(self, model, current_round):
+        # results = collab.clients.train(model, current_round)
+        results = collab.other_clients.train(model, current_round)
```
YuanTingHsieh, line 63: we have "other_clients" as well?

examples/collab/call_patterns/cyclic.py
```diff
@@ -0,0 +1,69 @@
...
+        file_name = os.path.join(collab.workspace.get_work_dir(), "final_model.npy")
+        save_np_model(final_result, file_name)
+        self.logger.info(f"[{collab.call_info}]: saved final model {final_result} to {file_name}")
+
+    def _do_one_round(self, current_round, current_model):
+        clients = collab.clients
+        random.shuffle(clients)
+        for client in clients:
+            current_model = client.train(current_round, current_model)
+            if current_model is None:
```
holgerroth: **[P2] `None`-as-failure checks test a contract that no longer exists**

Under the new error semantics, a non-optional `client.train()` either returns the trained model or raises `CollabCallError` — it never returns `None` to signal failure. These checks (here, in `execute()` above, and duplicated in `workflow_composition/cyclic.py`) are dead code that teaches readers a stale failure idiom. Since the relay pattern is exactly where per-client failure handling matters, the loop should demonstrate the real contract:

```python
try:
    current_model = client.train(current_round, current_model)
except CollabCallError as e:
    self.logger.error(f"round {current_round} failed on {e.site}: {e.cause}")
    return None
```

— or drop the checks entirely if a mid-relay failure should simply abort the workflow.

examples/collab/async_aggregation/avg_intime.py
```diff
@@ -0,0 +1,90 @@
...
+
+        result = aggr_result.total / aggr_result.count
+        self.logger.info(
+            f"[{collab.call_info}] round {current_round}: " f"aggr result from {aggr_result.count} clients: {result}"
+        )
+        return result
+
+    def _accept_train_result(self, gcc, result, aggr_result: _AggrResult):
+        self.logger.info(f"[{collab.call_info}] got train result from {collab.caller} {result}")
+        with aggr_result.lock:
```
holgerroth, line 94: **[P2] The example silently requires users to know callbacks run on backend threads**

The lock is necessary because `process_resp_cb` fires concurrently on result-delivery threads — but neither this example nor the design doc states that threading contract, so users copying the pattern won't know why the lock is there (or that they need one). Suggest a sentence in the module docstring: response callbacks run concurrently as results arrive; shared state must be synchronized. Longer term, a `serialized=True` group-call option that delivers callbacks one at a time would let in-time aggregation drop the lock entirely — most users would happily trade a little latency for that.

examples/collab/hello_fedavg/hello_fedavg.py
```diff
@@ -0,0 +1,135 @@
...
+    def forward(self, x):
+        return self.fc(x)
+
+
+class Trainer:
+    @collab.publish
+    def train(self, weights=None):
+        # Per-site config: local epochs are configured per site on the recipe
+        # and read here at run time (default 5 if this site isn't configured).
+        local_epochs = collab.get_app_prop("local_epochs", {}).get(collab.site_name, 5)
```
holgerroth: **[P2] Per-site config idiom ships every site's values to all sites and puts resolution on the user**

`set_client_prop("local_epochs", {"site-1": 2, "site-2": 5})` sends the full dict to every client, and each trainer does the `.get(collab.site_name, default)` dance. Beyond ergonomics, sending site A's configuration to site B is a mild information leak that enterprise deployments will notice. Cleaner target: the recipe resolves per-site values so the trainer reads `collab.get_app_prop("local_epochs", 5)` and receives only its own — fine as a follow-up, but the docstring shouldn't present the dict-broadcast idiom as the blessed "per-site configuration" pattern in the meantime.

examples/collab/swarm_events/swarm_algo.py
```diff
@@ -0,0 +1,125 @@
...
+
+    @collab.init
+    def init(self):
+        # This example shows that there could be multiple listeners for the same event
+        collab.register_event_handler("final_model", self._accept_final_model)
+        collab.register_event_handler("final_model", self._save_final_model)
+
+    @collab.publish
+    def train(self, weights, current_round):
+        self.logger.info(f"[{collab.call_info}]: train asked by {collab.caller}: {current_round=}")
```
holgerroth: **[P3] `collab.call_info` in nearly every log line makes user code read like framework tracing**

The advanced examples prefix most log messages with `[{collab.call_info}]` (9 times in this file alone). Useful for debugging the framework, but as the copy-paste template it adds noise to exactly the examples meant to show how natural the API feels — and `call_info` isn't documented in the design doc's facade section. Suggest keeping it only where the call context genuinely matters and letting plain log lines carry the rest; if call tracing is broadly useful, it may belong in the runtime's debug logging rather than application code.

holgerroth (commented):
Follow-up review focused on the examples now that the error-semantics rework has landed (thanks for the fast and thorough turnaround on 7fac7dfdf/f12f0110 — the CollabCallError contract and results.failures came out very clean). These four are UX-level polish so the examples model the new contracts users will copy.

examples/collab/call_patterns/cyclic.py
nvidianz: Addressed in 0c23626f4. Both call_patterns/cyclic.py and workflow_composition/cyclic.py now handle the actual raise-by-default contract: execute() catches CollabCallError, logs the failing site and cause, and terminates the workflow. The stale None checks after client.train() were removed.

examples/collab/async_aggregation/avg_intime.py
nvidianz: Addressed in 0c23626f4. The async aggregation module docstring now states that response callbacks may run concurrently on backend result-delivery threads and that shared state must be synchronized. The same threading contract is now documented in the Collab API design.

examples/collab/hello_fedavg/hello_fedavg.py
nvidianz: Addressed in 0c23626f4 using the existing recipe-side isolation path. hello_fedavg now calls set_per_site_config({site: {local_epochs: value}}), generates entries for the requested client count, and each trainer reads its already-resolved scalar with collab.get_app_prop("local_epochs", 5). The full site map is no longer exposed as a client property.

examples/collab/swarm_events/swarm_algo.py
nvidianz: Addressed in 0c23626f4. Removed all nine collab.call_info prefixes from swarm_algo.py. Logs are now plain application messages, retaining collab.caller only where the remote caller or event source is meaningful.

nvflare/collab/runtime/flare/cell_dispatcher.py
```diff
@@ -0,0 +1,162 @@
...
+                topic=MSG_TOPIC,
+                targets=self.target_fqcn,
+                message=request,
+                secure=call_opt.secure,
+                optional=call_opt.optional,
+            )
+            return None
+
+    def call_target_in_group(self, gcc: GroupCallContext, func_name: str, *args, **kwargs):
+        self.thread_executor.submit(self._run_func, gcc, func_name, args, kwargs)
```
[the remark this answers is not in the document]
nvidianz: Addressed in 0c23626f4. CellDispatcher now captures the submitted future and attaches a cancellation-aware completion callback. A cancelled queued future records CancelledError and calls send_completed(), releasing the parallel slot and giving ResultQueue a terminal failure. The regression test saturates a one-worker executor, cancels queued futures during shutdown, and verifies completion.

nvidianz:
@greptile-apps review

nvflare/collab/api/filter.py
```diff
@@ -0,0 +1,182 @@
...
+    funcs = get_filter_f(obj)
+    if not funcs:
+        raise ValueError(f"filter impl object {obj.__class__.__name__} has no {filter_type} func")
+
+    if len(funcs) > 1:
+        raise ValueError(
+            f"filter object {obj.__class__.__name__} must have one {filter_type} func but got {len(funcs)}"
+        )
+    return funcs[0]
+
```
YuanTingHsieh: Please defer the public `CallFilter`/`ResultFilter` API until it lands with a site-enforcement design. Direct Collab calls bypass the normal `ClientRunner` task-data/result-filter path, while these chains are supplied by the application/recipe and therefore cannot serve as site-owned privacy policy. Publishing them now creates a parallel filter system with unclear ownership and security guarantees. A follow-up should define site configuration and enforcement for native direct calls, with policy parity for task-plane Executor clients. If hooks are retained internally for now, please describe them only as application middleware rather than security filters.

nvflare/collab/runtime/client_api.py
```diff
@@ -0,0 +1,389 @@
...
+
+from nvflare.apis.analytix import AnalyticsDataType
+from nvflare.app_common.abstract.fl_model import FLModel
+from nvflare.client.api import use_api
+from nvflare.client.api_spec import APISpec
+from nvflare.collab.api.decorators import publish
+from nvflare.fuel.utils.log_utils import get_obj_logger
+
+
+class CollabClientAPI(APISpec):
```
YuanTingHsieh: Please remove/defer `CollabClientAPI` and the associated `nvflare.client` override changes from this PR. This creates a third hybrid programming model, and the registered callback does not survive normal FedJob export: `set_training_func()` stores runtime state that is not a constructor argument, while `CollabRecipe` ships the source of `type(client_obj)`, not the callback module, so a reconstructed client reaches `execute()` with no training function. The follow-up bridge should instead be a generic task-plane Executor bridge: a server-side Collab facade schedules a normal NVFlare `Task`, and `ClientRunner` invokes the configured `Shareable -> Shareable` Executor. `ScriptRunner("hello.py")` is the first motivating use case, but this can then support any standard Executor and inherit site filters, lifecycle, abort handling, accounting, external-process execution, and DDP. Please do not implement that bridge by directly hosting an Executor behind a client-side Collab RPC, since that would bypass those `ClientRunner` guarantees.

nvflare/collab/runtime/local/runner.py
```diff
@@ -0,0 +1,119 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+from typing import Dict, List, Optional, Tuple, Union
+
+from nvflare.collab.api.app import ClientApp, ServerApp
+from nvflare.collab.api.module_wrapper import resolve_server_client
+from nvflare.collab.runtime.local.app_runner import AppRunner
+
+
```
YuanTingHsieh: Is standalone Collab execution an intended supported product mode? If not, please remove the runtime-portability structure from the shipped architecture. Collab should have one production runtime: a normal NVFlare `FedJob` executed through `SimEnv`/`PocEnv`/`ProdEnv`. The local runner/workspace/runtime should move out of the shipped runtime package; retain only a minimal direct dispatcher as a test utility. The proxy should keep a private `_CallDispatcher`/`_InvocationDispatcher` seam for focused tests, but that seam should not imply public backend selection or multiple runtimes. The package can then reflect the real boundary as well: move the NVFlare-specific recipe out of `core/`, remove the redundant `runtime/flare/` level, and use the normal NVFlare workspace/FLContext instead of parallel local/Collab workspace implementations.

YuanTingHsieh (commented):
Follow-up architectural review focused on keeping the initial Collab contract aligned with NVFlare's existing site-policy, task-execution, and runtime boundaries.

nvflare/collab/runtime/local/runner.py
YuanTingHsieh: @chesterxgchen @holgerroth @pcnudde

nvflare/collab/api/filter.py
YuanTingHsieh: @chesterxgchen @holgerroth @pcnudde

nvidianz:
@greptile-apps review

docs/design/client_api_execution_modes_plan.md
nvidianz: Addressed in 88b0b8793. Removed `docs/design/client_api_execution_modes_plan.md` from this PR.

examples/collab/async_aggregation/async_aggregation.py
nvidianz: Addressed in 88b0b8793. Removed the example-specific runner wrappers and changed the example entry points to execute `CollabRecipe` with `SimEnv` directly. The example README and migration guide now document that flow.

examples/collab/call_patterns/trainer.py
nvidianz: Addressed in 88b0b8793. Multiple `@collab.init` methods are supported: each initializer runs once in deterministic method-name order. This behavior is documented and covered by `test_app_runs_multiple_init_functions_in_name_order`.

examples/collab/client_api/in_process_client.py
nvidianz: Addressed in 88b0b8793 by removing the `examples/collab/client_api` example and deferring the `CollabClientAPI` bridge entirely. The wrapper method no longer exists.

examples/collab/call_patterns/widgets.py
nvidianz: Addressed in 88b0b8793. Collab-specific events are deferred, so the event decorators, facade APIs, and example usage were removed. Published Collab functions remain the notification mechanism for this PR.

examples/collab/swarm/swarm_algo.py
nvidianz: Yes. `collab.other_clients` is retained and used by the swarm example; it returns the client group with the current client removed. This is now stated explicitly in the design document.

nvflare/collab/api/filter.py
nvidianz: Addressed in 88b0b8793. Removed the public `CallFilter` and `ResultFilter` APIs plus their recipe/runtime wiring. The design document now defers filters until native calls and task-plane clients have site-owned policy enforcement parity.

nvflare/collab/runtime/client_api.py
nvidianz: Addressed in 88b0b8793. Removed `CollabClientAPI`, its tests and examples, and the associated `nvflare.client` override changes. The design document records a future generic task-plane bridge through normal `Task`, `Executor`, and `ClientRunner` semantics.

nvflare/collab/runtime/local/runner.py
nvidianz: Addressed in 88b0b8793. Removed the standalone local runner, local workspace, and alternate shipped runtime. The direct dispatcher is now test-only, the dispatcher seam is private, `CollabRecipe` and the production runtime are flattened, and examples use normal NVFlare environments and `Workspace`/`FLContext`.

nvflare/collab/runtime/executor.py
```diff
@@ -0,0 +1,213 @@
...
+        if task_name != SYNC_TASK_NAME:
+            self.log_error(fl_ctx, f"received unsupported task {task_name}")
+            return make_reply(ReturnCode.TASK_UNKNOWN)
+
+        server_collab_interface = shareable.get(SyncKey.COLLAB_INTERFACE)
+        server_fqcn = shareable.get(SyncKey.SERVER_FQCN)
+        if not isinstance(server_fqcn, str) or not server_fqcn:
+            self.log_error(fl_ctx, "missing server FQCN in sync task")
+            return make_reply(ReturnCode.BAD_TASK_DATA)
+        client_collab_interface = self.client_app.get_collab_interface()
```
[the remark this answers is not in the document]
nvidianz: Addressed in d7114f4a7. `CollabExecutor` now publishes `self.client_app` only after start-run configuration succeeds, and `execute()` returns `ReturnCode.ERROR` when initialization did not complete. Two regression tests cover the failed-configuration state and the guarded task path; the full Collab suite passes with 59 tests.

nvidianz:
@greptile-apps review

examples/collab/async_aggregation/avg_intime.py
holgerroth: Expanding on the "longer term" note above — we think this could actually be done now, and it's bigger than callbacks.

This lock is the visible symptom of a missing contract. The API's promise is that a trainer is an ordinary Python object — and ordinary Python code assumes one call runs at a time. But the runtime delivers calls on separate threads: published methods (two overlapping calls to one site), event handlers, and response callbacks like this one can all run concurrently against the same object. Two threads doing `total += result` at the same instant silently lose a client's contribution — no error, just an occasionally-wrong model. Nothing in the docs or examples tells users this; the lock here is the only hint, and someone copying the pattern won't know why it exists.

**Proposal: make the promise true — serialize delivery per object.** For a given collab object, run one invocation at a time (published methods, event handlers, response callbacks); queue the rest. This is what Ray actors do by default, and it's a big part of why actor code feels like normal Python. It doesn't slow FL down: different sites are different objects on different machines, so fan-out to N clients stays fully parallel — only calls into the *same* object queue. With that guarantee, this example drops the lock entirely, and the round-state race in `CollabClientAPI.execute()` disappears structurally instead of lock-by-lock. It's a runtime dispatch change, not an API change, so it seems feasible in this PR or as an immediate follow-up; an opt-in `concurrent=True` escape hatch can come later if a real use case appears.

**One caveat to design around:** re-entrancy. If an object blocks waiting for a call that must be delivered back into *itself*, strict serialization deadlocks. The swarm example hits this today: `swarm_learn` calls `collab.clients(expect_result=True).fire_event("final_model", ...)` — all clients *including itself* — while still running. Self-delivery would need to execute inline on the calling thread (or `fire_event` to self documented as inline). That's a small, well-understood carve-out (Ray has the same issue and handles it with opt-in concurrency).

If the serialization lands, the docstring note requested above shrinks to one line: "calls into your object run one at a time."

examples/collab/call_patterns/call_patterns.py
```diff
@@ -0,0 +1,78 @@
...
+
+from nvflare.collab import CollabRecipe, simple_logging
+from nvflare.recipe import SimEnv
+
+INITIAL_MODEL = [[1, 2, 3], [4, 5, 6], [7, 8, 9]]
+
+
+def make_strategy(pattern: str, num_rounds: int):
+    if pattern == "seq":
+        return NPFedAvgSequential(initial_model=INITIAL_MODEL, num_rounds=num_rounds)
```
YuanTingHsieh: Not sure this pattern "seq" is the best name though.
It is confusing when I see sequential I am thinking it is the cyclic algo.

examples/collab/call_patterns/widgets.py
```diff
@@ -0,0 +1,26 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from nvflare.collab import collab
+from nvflare.fuel.utils.log_utils import get_obj_logger
+
+
+class MetricReceiver:
```
YuanTingHsieh: where do we use this?

nvidianz:
@greptile-apps review

nvidianz:
@greptile-app review

nvidianz:
@greptile-apps review

nvidianz:
@greptile-apps review

nvidianz:
@greptile-apps review

nvflare/collab/recipe.py
```diff
@@ -0,0 +1,187 @@
...
+        job.to_clients(executor, id="executor", tasks=["*"])
+
+        # Ship the user-provided server/client code into each app's "custom" folder.
+        # A collab job runs user-defined objects (the server/client and any collab
+        # objects), so it is inherently bring-your-own-code (BYOC): the code must
+        # travel with the job, and the job is authorized via the BYOC right rather
+        # than the site class allow-list (which cannot enumerate arbitrary user
+        # classes). Adding a custom folder is what marks the job as BYOC.
+        for src in self._user_source_files([self.server_app.obj] + list((self.server_objects or {}).values())):
+            job.add_file_to_server(src, app_folder_type="custom")
```
YuanTingHsieh: **[P1] Preserve package topology for `ModuleWrapper` sources**

This adds only the module's single `.py` file to `custom/`, which flattens a package-qualified module such as `collab.hello_numpy_collab.hello_numpy_collab` into `custom/hello_numpy_collab.py`. Sibling imports and the original module name then cannot be resolved in a clean POC/production workspace. Class components get recursive source packaging through `FedJob`, but `ModuleWrapper` serializes only the module name, so that machinery does not cover this case. Please preserve the package path and dependencies, and add an exported-job test that imports the module in a clean process.

nvflare/collab/runtime/dispatch.py
```diff
@@ -0,0 +1,143 @@
...
+    if not m:
+        return _error_reply(f"no method named '{method_name}' or it is not collab", logger)
+    else:
+        logger.debug(f"found method for {method_name}")
+
+    # invoke this method
+    previous_ctx = get_call_context()
+    try:
+        _, method_kwargs = _preprocess(app, caller, obj_name, target_name, method_name, m, method_args, method_kwargs)
+        result = m(*method_args, **method_kwargs)
```
YuanTingHsieh, line 152: **[P1] Do not run user training on the shared Cell callback pool**

This callback directly invokes the published user method, so an hours-long training call occupies a Cell streaming callback thread for its entire duration. With enough concurrent or nested calls, user work can exhaust that shared pool and starve transport/reply callbacks, leading to timeouts or deadlock. The dispatcher's `max_call_threads_*` settings only bound outgoing calls; they do not isolate inbound execution here. Please hand published calls to a Collab-owned executor and send the reply asynchronously rather than running user code on the Cell callback thread.

nvflare/collab/api/group.py
```diff
@@ -0,0 +1,208 @@
...
+                            call_opt=self._call_opt,
+                            func_name=func_name,
+                            process_cb=self._process_resp_cb,
+                            cb_kwargs=copy.copy(self._cb_kwargs),
+                            context=ctx,
+                            waiter=waiter,
+                        )
+
+                        # try to get permission to make next call
+                        gcc.set_send_complete_cb(self._request_sent, gcc=gcc, proxy=func_proxy)
```
YuanTingHsieh: **[P1] `parallel` limits sends, not in-flight calls**

The slot is released by the send-complete callback, which fires when transmission finishes—not when the remote method finishes. Consequently, `parallel=1` can still start training on every target almost immediately, so this does not implement the documented maximum execution parallelism. If `parallel` is meant to bound in-flight calls, release the slot on terminal result/failure (which may also remove the need for the Cell API change). Otherwise this should be renamed and documented as a send-window limit.

nvflare/collab/runtime/executor.py
```diff
@@ -0,0 +1,220 @@
...
+                app.set_prop(name, value)
+
+        # Publish the app only after start-run initialization succeeds. This
+        # keeps execute() from using a partially configured app.
+        self.client_app = app
+
+    def _handle_end_run(self, event_type: str, fl_ctx: FLContext):
+        if self.client_ctx:
+            self.logger.info(f"finalizing client app {self.client_app.name}")
+            self.client_app.finalize(self.client_ctx)
```
YuanTingHsieh: **[P1] Install the Collab context during client finalization**

`END_RUN` may execute on a different thread from the earlier sync task, but this calls `finalize()` without installing `self.client_ctx` in the thread-local context. A `@collab.final` method that uses facade state such as `collab.site_name`, workspace, or `fl_ctx` can therefore fail even though the same access works during the main call. Please centralize lifecycle invocation under context activation/restoration (with `try/finally`) and test finalization on a different thread.

nvflare/collab/api/module_wrapper.py
```diff
@@ -0,0 +1,321 @@
...
+
+    def _create_main_method(self, name, original_func):
+        """Create a method wrapper for an already-decorated @collab.main function.
+
+        The original function is already decorated, so we just need to make it
+        callable as a bound method.
+        """
+
+        # Create a simple wrapper that forwards to the original decorated function
+        def method(self):
```
YuanTingHsieh: **[P1] Preserve context support for module-level `main`**

The generated method neither forwards `*args/**kwargs` nor copies `_FLAG_SUPPORT_CTX` from the decorated function. As a result, `@collab.main def run(context): ...` works for a class app but a module app is reported as not supporting context and raises `TypeError` when `context=` is supplied. Please make module- and class-based lifecycle semantics identical by forwarding arguments and preserving the context-support marker, with a regression test.

nvflare/collab/runtime/executor.py
```diff
@@ -0,0 +1,220 @@
...
+            target_fqcn=FQCN.join([client.get_fqcn(), job_id]),
+            abort_signal=abort_signal,
+            thread_executor=self.thread_executor,
+        )
+        proxy = Proxy(
+            app=self.client_app,
+            target_name=client.name,
+            target_fqn=client.get_fqsn(),
+            backend=backend,
+            target_interface=collab_interface.get(""),
```
YuanTingHsieh, line 172: **[P1] Do not reuse the local client's interface for every remote client**

`collab_interface` here is this site's own client interface, yet it is assigned to proxies for all other clients. The server already collects each client's interface, and `make_client_app(site_name)` permits site-specific apps, so heterogeneous sites will either reject a valid call locally or allow a call that the target cannot serve. Please either make client interfaces explicitly homogeneous (and remove the site-specific hook), or distribute and use a per-site interface map when constructing these proxies.

nvflare/collab/api/group.py
```diff
@@ -0,0 +1,208 @@
...
+
+                    assert isinstance(self._app, App)
+                    check_call_args(func_name, func_itf, adj_args, adj_kwargs)
+
+                    waiter = ResultWaiter([p.name for p in self._proxies])
+                    max_parallel = self._call_opt.parallel
+                    if max_parallel <= 0:
+                        max_parallel = len(self._proxies)
+
+                    for p in self._proxies:
```
YuanTingHsieh, line 130: **[P2] Validate the whole group before dispatching any call**

Validation and submission are interleaved in this loop. If a later proxy lacks the method or otherwise fails preparation, earlier sites may already be executing, leaving a partially dispatched operation even though the group call raises. Please resolve and validate every target first, then begin dispatch only after the full preflight succeeds.

examples/collab/workflow_composition/cyclic.py
```diff
@@ -0,0 +1,51 @@
...
+class NPCyclic:
+
+    def __init__(self, initial_model, num_rounds=2):
+        self.num_rounds = num_rounds
+        self.initial_model = initial_model
+        self._initial_model = parse_array_def(initial_model)
+        self.final_model = None
+        self.logger = get_obj_logger(self)
+
+    @collab.main
```
YuanTingHsieh: **[P2] Keep only the registered wrapper as `@collab.main`**

In this example, only `Controller.run` is discovered and registered as the server main; `NPCyclic.execute()` is just called as an ordinary local method. This decorator therefore has no runtime effect and makes the example appear to define multiple entry points despite the documented exactly-one-main model. The same applies to `NPFedAvgParallel.execute()` in `avg_para.py`. Please make these ordinary stage methods, or introduce and document an explicit reusable workflow-stage abstraction if standalone/composed dual use is intended.

nvflare/collab/api/proxy_utils.py
```diff
@@ -0,0 +1,78 @@
...
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Shared utilities for creating Proxy objects.
+
+This module provides common functionality for proxy preparation that is shared
+between AppRunner (simulation path) and CollabExecutor (FLARE path).
```
YuanTingHsieh: **[P2] Remove or actually use this stale simulation helper**

This helper has no callers, still describes an `AppRunner` simulation path that is no longer present, and the controller/executor now duplicate proxy-tree construction instead. Leaving it in the public `api` package preserves a misleading second boundary. Please either delete it with the retired simulation path or make it the single internal proxy builder used by both runtimes and update its contract.

YuanTingHsieh (commented):
Implementation review after tracing the current API/runtime paths. The inline comments cover source packaging, inbound execution isolation, `parallel` semantics, lifecycle context, module-level `main`, per-site interfaces, group preflight, composition decorator semantics, and stale boundary code. These paths need runtime/integration coverage beyond the current mocked unit tests.

nvflare/collab/api/proxy.py
```diff
@@ -0,0 +1,259 @@
...
+from .call_opt import CallOption
+
+
+class _ProxyCall:
+
+    def __init__(
+        self,
+        proxy,
+        expect_result: bool = True,
+        timeout: float = 5.0,
```
YuanTingHsieh: **[P2] Unify the default call timeout**

`CallOption` defaults to `timeout=60.0` (call_opt.py:20), but `_ProxyCall`, `Proxy.__call__`, and `ProxyList.__call__` all default to `5.0` (proxy.py:32, proxy.py:77, proxy_list.py:48). So `client.train(...)` waits 60s while `client(<any option>).train(...)` waits 5s — adding any unrelated call option silently cuts the timeout 12x, and a 5s default will spuriously time out real training rounds on the option-form paths. Please derive the proxy-level defaults from a single source (e.g. default to `CallOption()`); timeout defaults are hard to change once the API is public.

nvflare/collab/api/proxy.py
```diff
@@ -0,0 +1,259 @@
...
+            else:
+                call_error = CollabCallError(self.name, func_name, ex)
+            if call_opt.optional:
+                self.logger.warning(f"optional {call_error}")
+                return None
+            if call_error is ex:
+                raise
+            raise call_error from ex
+
+    def __getattr__(self, func_name):
```
YuanTingHsieh, line 245: **[P2] Guard the catch-all `__getattr__` against dunder and reserved names**

`Proxy.__getattr__` (and `_ProxyCall`/`ProxyList`) turns any attribute access into a remote call and never rejects `_`-prefixed names. So `copy.deepcopy(proxy)` or pickling a proxy issues a remote call named `__deepcopy__`/`__getstate__` instead of copying/pickling, and `hasattr(proxy, x)` is always True. Separately, `ProxyList(list)` shadows the list method names (`pop`, `count`, `clear`, `append`, `index`, ...): a published method with any of those names is unreachable, and `collab.clients.pop()` silently removes a proxy from the list instead of calling the sites. Please raise `AttributeError` for names starting with `_` and reserve (or underscore-prefix) the internal names, since this namespace is effectively frozen at release.

examples/collab/call_patterns/call_patterns.py
nvidianz: Obsolete in a6d6cd551: the `call_patterns` example was removed from this PR, and the Collab example catalog was updated accordingly.

examples/collab/call_patterns/widgets.py
nvidianz: Obsolete in a6d6cd551: the `call_patterns` example, including this unused widget, was removed from this PR.

examples/collab/workflow_composition/cyclic.py
nvidianz: Obsolete in a6d6cd551: the `workflow_composition` example was removed from this PR, so the redundant workflow-stage decorators are no longer present.

nvflare/collab/recipe.py
nvidianz: Addressed in c12301510. ModuleWrapper sources now retain their top-level package path and ship the package's Python sources, including sibling modules. Added an exported-job regression that starts an isolated (`-I`) process and imports the packaged module from the exported `custom/` directory.

nvflare/collab/runtime/dispatch.py
nvidianz: Addressed in c12301510. Inbound published calls are now submitted to the Collab-owned ThreadPoolExecutor, and the streaming Adapter sends Future-backed replies asynchronously so user training no longer occupies the shared Cell callback pool. Added executor-thread and async-reply regressions.

nvflare/collab/api/group.py
nvidianz: Addressed in c12301510. `parallel` now counts calls until terminal completion rather than transmission completion. The idempotent completion hook is released on result, failure, cancellation, or timeout, and the Cell send-complete API addition was removed. Added a `parallel=1` regression proving the second call waits for the first terminal result.

nvflare/collab/api/module_wrapper.py
nvidianz: Addressed in c12301510. Module-level main wrappers now forward `*args/**kwargs` and preserve the context-support marker. Added a regression for `@collab.main def run(context)` on a module app.

nvflare/collab/runtime/executor.py
nvidianz: Addressed in c12301510. App lifecycle invocation now centrally activates the supplied Collab context and restores the prior thread-local context in `finally`. Client lifecycle contexts are created without leaking onto the sync thread, and a regression finalizes on a different thread while reading `collab.site_name`.

nvflare/collab/runtime/executor.py
nvidianz: Addressed in c12301510. Client startup is now two-phase: clients first report their interfaces, then the controller distributes the full per-site interface map in the setup task. Each remote proxy is built from its target site's reported interface. Added heterogeneous-interface coverage.

nvflare/collab/api/group.py
nvidianz: Addressed in c12301510. Group calls now resolve, normalize, and validate every member into a work list before dispatch starts. Added a regression showing a later invalid member prevents dispatch to all earlier members.

nvflare/collab/api/proxy.py
nvidianz: Addressed in c12301510. A single `DEFAULT_CALL_TIMEOUT` now drives CallOption, Proxy, option-form Proxy calls, and ProxyList; all paths default to 60 seconds. Added regression coverage for the option variants.

nvflare/collab/api/proxy_utils.py
nvidianz: Addressed in c12301510 by deleting the unused `proxy_utils.py` simulation helper; it had no callers and no longer represents a supported runtime boundary.

nvflare/collab/api/proxy.py
nvidianz: Addressed in c12301510. Proxy, option proxies, and ProxyList now reject missing underscore-prefixed attributes. ProxyList is a read-only sequence wrapper rather than a list subclass, so published names such as `pop` dispatch remotely instead of mutating the local collection. Added regressions for both behaviors.

nvidianz:
@greptile-apps review

nvidianz:
@greptile-apps review

nvidianz:
@greptile-apps review

nvidianz:
@greptile-apps review

nvflare/collab/runtime/executor.py
```diff
@@ -0,0 +1,252 @@
...
+        client_name = fl_ctx.get_identity_name()
+        job_meta = fl_ctx.get_prop(FLContextKey.JOB_META)
+        job_clients = job_meta.get(JobMetaKey.JOB_CLIENTS)
+        all_clients = [from_dict(d) for d in job_clients]
+
+        prepare_for_remote_call(
+            cell=cell,
+            app=self.client_app,
+            logger=self.logger,
+            executor=self.thread_executor,
```
YuanTingHsieh: **[P1] Separate inbound and outbound Collab executors**

`self.thread_executor` is now used both for inbound published-method execution through `prepare_for_remote_call(..., executor=self.thread_executor)` and for outbound blocking calls through every `CellDispatcher(..., thread_executor=self.thread_executor)`. If outbound `_run_func` calls fill the pool while awaiting peer replies, processing those replies can depend on inbound `_call_app_method` work queued to the same exhausted pool; two peers can therefore mutually starve. The simpler reproduction is `max_call_threads_for_client=1`: an inbound `swarm_learn()` occupies the only worker, calls `collab.other_clients.train()`, and waits for an outbound job queued behind itself. This is a reachable deadlock in the documented client-to-client pattern, not merely a sizing concern. Please use separate inbound and outbound pools and configure/size their roles independently.

nvflare/collab/recipe.py
```diff
@@ -0,0 +1,223 @@
...
+
+    def _client_props_with_per_site_config(self) -> Dict[str, object]:
+        """Client app props for the executor, with per-site config attached.
+
+        The per-site map rides in the shared executor config under a reserved
+        key; each CollabExecutor resolves its own site's entries at start-run.
+        """
+        props = dict(self.client_app.get_props() or {})
+        if self._per_site_config:
+            props[PER_SITE_CONFIG_PROP] = self._per_site_config
```
YuanTingHsieh: **[P1] Materialize per-site properties per target instead of sharing the full map**

This puts the complete `set_per_site_config()` map into the common `CollabExecutor` constructor props, so the same map is serialized into every client's job config. At startup each client copies its own values into plain app props, but the reserved map remains present—and even removing it at runtime would not remove it from the shipped job. An earlier resolved thread states that the full map is no longer exposed as a client property, but the current implementation contradicts that claim. Please create site-targeted executor/app configuration (for example via `job.to(component, site_name)`) so each client receives only its own entry.

nvflare/collab/runtime/executor.py
```diff
@@ -0,0 +1,252 @@
...
+            app=self.client_app,
+            target_name=client.name,
+            target_fqn=client.get_fqsn(),
+            backend=backend,
+            target_interface=collab_interface.get(""),
+        )
+
+        collab_objs = self.client_app.get_collab_objects()
+        if collab_objs:
+            for name in collab_objs.keys():
```
YuanTingHsieh: **[P1] Build child proxies from the remote site's interface**

The two-phase exchange now supplies the correct per-site interface, but child construction still iterates this site's *local* Collab object names. If `make_client_app(site_name)` gives another site a named object that is absent locally, that remote object never gets a proxy even though it is present in `collab_interface`. I verified that a remote interface `{"": {}, "remote_only": {"foo": []}}` produced only the local `client` child. Please iterate the remote `collab_interface.items()` (skipping `""`), as `_prepare_server_proxy()` already does, and add heterogeneous named-object coverage.

nvflare/collab/api/group.py
```diff
@@ -0,0 +1,205 @@
...
+                            context=call_ctx,
+                            waiter=waiter,
+                        )
+                        gcc.set_completion_cb(self._call_completed, gcc=gcc, proxy=func_proxy)
+                        work_items.append((func_proxy, gcc, call_args, call_kwargs))
+
+                    for func_proxy, gcc, call_args, call_kwargs in work_items:
+                        # Limit calls that are still in flight, including remote
+                        # execution and response transfer.
+                        waiter.wait_for_call_permission(max_parallel, self._abort_signal)
```
YuanTingHsieh: **[P2] Preserve nonblocking semantics with bounded parallelism**

The caller waits here for terminal completion slots before execution reaches the `blocking=False` return below. With N targets and `parallel=1`, a nominally nonblocking call waits for the first N-1 remote calls to finish before returning; the new unit test's background thread actually codifies this behavior. That contradicts the API documentation that `blocking=False` returns immediately. Please move bounded dispatch scheduling to a background coordinator and return the `ResultQueue` once preflight succeeds, or change the public contract explicitly if this call is intentionally blocking while dispatching.

nvflare/collab/api/proxy.py
```diff
@@ -0,0 +1,265 @@
...
+            num_call_args = len(args) + len(kwargs)
+            if num_call_args > len(func_itf):
+                raise RuntimeError(
+                    f"there are {num_call_args} call args ({args=} {kwargs=}), "
+                    f"but function '{func_name}' only supports {len(func_itf)} args ({func_itf})"
+                )
+            call_kwargs = copy.copy(kwargs)
+            call_args = []
+            for i, arg_value in enumerate(args):
+                call_kwargs[func_itf[i]] = arg_value
```
YuanTingHsieh: **[P2] Preserve Python argument-binding semantics**

This positional-to-keyword conversion silently overwrites a keyword already supplied by the caller. For example, `proxy.train(1, x=2)` sends `x=1` instead of raising Python's “multiple values for argument 'x'” error. Because the exchanged interface stores names only, positional-only and keyword-only parameter rules are also lost even though the API promises normal-looking Python calls. At minimum reject a positional mapping whose name already exists in `kwargs`; the durable contract should serialize parameter kinds and bind against that signature. Please add regression coverage for duplicate and keyword-only binding.

YuanTingHsieh (commented):
Follow-up review on current head `973b29b`. The earlier architecture concerns are largely addressed, but the inline findings still affect deadlock safety, per-site isolation, heterogeneous interfaces, nonblocking behavior, and Python call semantics.

Please also update the PR description before merge. It still advertises the removed `CollabClientAPI`, `core/recipe.py`, `runtime/flare`, `runtime/local`, tracking writers, and removed examples, and it still reports 38 tests. That now contradicts both the shipped package and the current design document.

nvidianz:
@greptile-apps review

nvflare/collab/runtime/executor.py
nvidianz: Addressed in `4de721b0c`. `CollabController` and `CollabExecutor` now own distinct inbound and outbound thread pools, each independently configurable while preserving the existing `max_call_threads` fallback. Remote-call handlers use the inbound pool and every `CellDispatcher` uses the outbound pool. Added serialization and independent-pool regression coverage; all 86 Collab tests pass.

nvflare/collab/recipe.py
nvidianz: Addressed in `4de721b0c`. `CollabRecipe` now materializes a separate site-targeted client app/executor for every configured site and merges only that site's values into its props. The shared full-map property and reserved constant were removed. Coverage verifies `@ALL` is absent and each exported executor sees only shared values plus its own site entry.

nvflare/collab/runtime/executor.py
nvidianz: Addressed in `4de721b0c`. Client child proxies are now built by iterating the remote `collab_interface.items()` and skipping only the primary `""` entry, matching the server-side behavior. The new heterogeneous-interface test includes a `remote_only` object absent from the local app and verifies that its proxy is available.

nvflare/collab/api/group.py
nvidianz: Addressed in `4de721b0c`. A nonblocking group call now performs complete preflight, starts a daemon dispatch coordinator that enforces the terminal-completion parallel bound, and immediately returns its `ResultQueue`. The regression proves `blocking=False, parallel=1` returns before the first call completes while the second dispatch still waits for the first terminal result.

nvflare/collab/api/proxy.py
nvidianz: Addressed in `4de721b0c` with the durable contract. Published interfaces now serialize parameter names, kinds, and required state; proxy calls bind against that signature before transport; and receivers reconstruct positional-only invocation. Regression coverage rejects duplicate positional/keyword values and invalid positional-only/keyword-only calls while preserving legacy interface decoding.

nvidianz:
@YuanTingHsieh Addressed the follow-up review in `4de721b0c`: separated inbound/outbound executors, isolated per-site job configuration, built heterogeneous child proxies from each remote interface, restored immediate nonblocking bounded dispatch, and preserved Python signature binding. I replied with regression evidence and resolved all five inline threads. I also updated the PR description to remove the retired Client API/layout/examples and report the current 86-test Collab suite.

nvidianz:
@greptile-apps review

YuanTingHsieh (approved):

PR #4902, opened 2026-07-14
Implement per-site configuration for FedAvg and FedEval recipes

nvkevlu:
Enable `set_per_site_config()` as the canonical per-site configuration API for the built-in FedAvg, FedEval, and XGBoost recipes.

### Description

- Validates per-site configuration when `set_per_site_config()` is called and stores a stable snapshot of each site’s settings.
- Preserves referenced objects such as data loaders while preventing later caller mutations from changing deferred configuration.
- Defers client-app creation until client customization, export, or run, then materializes the topology exactly once.
- Creates named client apps when per-site configuration is supplied; otherwise creates the default `@ALL` client app.
- Keeps recipe-specific validation and construction within each concrete recipe, including XGBoost data loaders and client-rank resolution.
- Preserves the NumPy runner exchange format and applies later client additions, including CSE components, through the selected topology.
- Retains constructor-level `per_site_config` as a deprecated compatibility path that delegates to the canonical setter and emits a `FutureWarning`.
- Updates documentation and examples and adds focused coverage for configuration snapshots, deferred topology preparation, compatibility paths, and regressions.

### Validation

- Final focused tests: 127 passed.
- Broader affected recipe suite: 552 passed, 23 optional-dependency skips.
- `./runtest.sh -s`: Black, isort, flake8, and agent-skill lint passed.

### Types of changes

- [x] Documentation updated.


nvflare/app_opt/xgboost/recipes/vertical.py
```diff
@@ -258,6 +260,12 @@ def __init__(
         self.job = self.configure()
         Recipe.__init__(self, self.job)
 
+    def _apply_per_site_config(self, config: dict[str, dict]) -> None:
```
YuanTingHsieh: Why should XGBoost reject the API that is intended to replace constructor-level `per_site_config`? The current `configure()` method uses this mapping to derive client ranks, construct the server controller, and add each site's executor and data loader. Those operations can be moved into `_apply_per_site_config()`.

Please make the constructor argument a deprecated compatibility path that delegates to the helper, rather than making XGBoost permanently constructor-only. The same applies to the horizontal and bagging recipes.

nvflare/recipe/spec.py
```diff
@@ -309,13 +309,140 @@ def set_per_site_config(self, config: Dict[str, Dict]) -> None:
...
+            self._apply_per_site_config(dict(stored_config))
+        except Exception:
+            self._helper_per_site_config = previous_config
+            raise
 
     def _apply_per_site_config(self, config: Dict[str, Dict]) -> None:
         """Recipe-specific hook for helper-provided per-site configuration."""
         pass
 
+    def _record_client_runner(self, target: str, component_ids: Dict[str, str], script: str) -> None:
```
YuanTingHsieh: This logic appears too implementation-specific for the generic `Recipe` specification. It knows how `ScriptRunner` materializes component IDs and scripts, clones `ClientApp` internals, and directly modifies private `FedJob` fields such as `_deploy_map`, `_components`, and `_deployed`.

Could this be moved into a shared FedAvg/FedEval runner-topology helper or supported through a public `FedJob` API? `Recipe` should retain only the generic per-site configuration lifecycle.

nvflare/app_opt/pt/recipes/fedeval.py
```diff
@@ -193,34 +197,49 @@ def __init__(
...
+            server_expected_format=site_config.get("server_expected_format", self.server_expected_format),
+            memory_gc_rounds=self.client_memory_gc_rounds,
+            cuda_empty_cache=self.cuda_empty_cache,
+        )
+        return runner, script
+
+    def _add_client_runner(self, job: BaseFedJob, target: str, site_config: Dict) -> None:
+        runner, script = self._create_client_runner(site_config)
+        component_ids = job.to(runner, target)
+        self._record_client_runner(target, component_ids, script)
```
YuanTingHsieh: This calls base-class runner bookkeeping before `Recipe.__init__()` has run and maintains a separate constructor topology path. Could FedEval always construct its default `@ALL` runner, initialize `Recipe`, and then forward a legacy constructor `per_site_config` through `self.set_per_site_config()`?

That would make the helper the canonical implementation, avoid depending on uninitialized base-class state, and provide a straightforward constructor deprecation path.

nvflare/recipe/spec.py
```diff
@@ -309,13 +309,140 @@ def set_per_site_config(self, config: Dict[str, Dict]) -> None:
...
+        old_per_site_config = self.per_site_config
+
+        try:
+            for target in client_targets:
+                job._deploy_map.pop(target, None)
+            job.clients = [target for target in job.clients if target not in client_targets]
+            self._client_runner_component_ids = {}
+            self._client_runner_scripts = {}
+
+            for site_name, site_config in config.items():
```
YuanTingHsieh: `set_per_site_config(recipe, {})` passes validation, removes the existing `@ALL` client app, and adds no site apps, leaving a server-only job that can still be exported.

Please reject an empty mapping before modifying the topology. It may also be worth validating that the configured site count can satisfy `min_clients`.

YuanTingHsieh (changes requested):
The intended direction of `set_per_site_config()` is to make it the canonical per-site configuration API and eventually deprecate constructor-level `per_site_config`. This change currently adds a second configuration path for FedAvg/FedEval while keeping XGBoost permanently constructor-only.

Could we consolidate each recipe onto one implementation path? The legacy constructor argument can remain temporarily for compatibility, emit a deprecation warning, and delegate to `set_per_site_config()`. XGBoost can support this by moving its site validation, rank derivation, and site-dependent job construction into `_apply_per_site_config()`.

I also do not think ScriptRunner-specific removal/cloning logic belongs in the generic `Recipe` specification or should directly mutate private `FedJob` state. Details are included inline.

nvkevlu:
@YuanTingHsieh thanks for the guidance, this should now be reworked properly

nvflare/recipe/fedavg.py
```diff
@@ -358,75 +363,48 @@ def __init__(
...
+
+    def _create_client_runner(self, site_config: Dict) -> ScriptRunner:
+        return ScriptRunner(
+            script=self._site_value(site_config, "train_script", self.train_script),
+            script_args=self._site_value(site_config, "train_args", self.train_args),
+            launch_external_process=self._site_value(
+                site_config, "launch_external_process", self.launch_external_process
+            ),
+            command=self._site_value(site_config, "command", self.command),
+            framework=self._site_value(site_config, "framework", self.framework),
```
YuanTingHsieh: [P1] Preserve the NumPy runner format in the canonical setter path

`NumpyFedAvgRecipe` passes `FrameworkType.NUMPY` into this parent constructor, but after `super().__init__()` returns it changes `self.framework` to `FrameworkType.RAW` for CSE compatibility. A normal post-construction `set_per_site_config()` therefore rebuilds the site runners here with RAW, while the deprecated constructor path is applied before that override and builds them with NUMPY. `ScriptRunner` maps those values to different parameter exchange formats, so the replacement API is not equivalent and can break NumPy model exchange. Please preserve the client-runner framework separately (or override runner creation in the NumPy recipe) and add a test using the canonical setter.

nvflare/recipe/utils.py
```diff
@@ -291,6 +295,66 @@ def set_per_site_config(recipe: Recipe, config: Dict[str, Dict]) -> None:
...
+            "other client apps are added"
+        )
+
+    added_targets = []
+    try:
+        for site_name, site_config in config.items():
+            added_targets.append(site_name)
+            add_client_app(job, site_name, site_config)
+        if replace_all_clients:
+            job.remove_client_app(ALL_SITES)
```
YuanTingHsieh: [P1] Keep later client augmentation from recreating `@ALL`

After this removes `@ALL` and leaves named client apps, `add_cross_site_evaluation()` calls `NumpyFedAvgRecipe.add_cse_validator_if_needed()`, which still uses `job.to_clients()`. That recreates an `@ALL` app containing only the validator. During export, `FedJob._set_all_apps()` gives `@ALL` precedence, so the named apps—and their training runners—are silently omitted. Please add the validator through the topology-aware Recipe client API (for example `_add_to_client_apps()`) and add a per-site NumPy + CSE export regression test.

YuanTingHsieh (changes requested):
Thanks for the rework. The earlier XGBoost/FedEval concerns are substantially addressed.

I still have one broader lifecycle concern. Given the direction in #4903 that the generated FedJob is private Recipe state, do we need to preserve eager client-app construction here?

FedAvg/FedEval currently create an `@ALL` client app during construction, then `set_per_site_config()` creates named apps and removes `@ALL`. This create-then-replace lifecycle is why this PR needs exact `job.clients` assertions, rollback, `remove_client_app()`, and `_configure_per_site_clients()`; it also makes the topology fragile if a later operation calls `job.to_clients()`.

Could client topology instead be materialized once, internally, after the recipe configuration is known? `set_per_site_config()` could validate/store the configuration, and an internal one-time preparation hook could create either the named apps or the default `@ALL` app. The hook would need to run before client-targeted Recipe customization and from both `export()` and `run()`, since `run()` currently calls `env.deploy(_job)` directly. This would not add a public `configure()` step.

The two inline comments below are concrete correctness blockers in the current implementation.

YuanTingHsieh (commented):
The latest rework addresses the previous lifecycle and NumPy blockers. One remaining configuration-state issue is noted inline. Also, please update the PR description: it still says the helper converts an eagerly-created all-client app, while the implementation now defers client topology materialization.

nvflare/recipe/spec.py
```diff
@@ -310,11 +332,32 @@ def set_per_site_config(self, config: Dict[str, Dict]) -> None:
             )
         else:
             warn_on_unsupported_secret_refs(config, context="per_site_config")
-        self._helper_per_site_config = dict(config)
-        self._apply_per_site_config(dict(self._helper_per_site_config))
+        stored_config = dict(config)
```
YuanTingHsieh: [P2] Keep a single stable per-site configuration snapshot

What is the intended distinction between `_helper_per_site_config` and each concrete recipe's `self.per_site_config`? This mapping is used only as the applied-once marker and by `configured_sites()`, while `_prepare_client_apps()` reads the separate recipe copy. That leaves two sources of truth that can diverge.

Also, `dict(config)` copies only the top-level mapping. Because client materialization is deferred, mutating `config[site]` after this call changes what is eventually built and can bypass the validation already performed by `_apply_per_site_config()`. Please either keep one canonical configuration in `Recipe`, or store only the configured site names/applied marker here and let the concrete recipe own the configuration. In either case, snapshot each site dictionary (for example `{site: dict(site_config) for site, site_config in config.items()}`) while retaining referenced objects such as data loaders.

YuanTingHsieh (commented):
Two follow-up comments on the `spec.py`/`utils.py` boundary: one deferred target-validation issue and one non-blocking validation deduplication.

nvflare/recipe/utils.py
```diff
@@ -278,19 +280,50 @@ def _validate_per_site_config_shape(config: Any) -> Dict[str, Dict]:
...
+    Per-site values become part of the generated job definition and must never
+    contain actual secret values; see :mod:`nvflare.recipe.secrets`.
     """
     recipe.set_per_site_config(_validate_per_site_config_shape(config))
 
 
+def _validate_per_site_targets(config: Dict[str, Dict], min_clients: int) -> None:
+    """Validate site targets and the minimum runnable site count."""
+    reserved_targets = {SERVER_SITE_NAME, ALL_SITES}
+    for site_name in config:
```
YuanTingHsieh, line 302: [P2] Validate the complete FedJob target-name contract here

This rejects only the reserved targets, but `FedJob._validate_target()` also rejects an empty target and names containing characters such as `/` or `@`. With deferred client preparation, `set_per_site_config(recipe, {"site/1": ...})` succeeds and records the configuration; the failure occurs only later when `_prepare_client_apps()` calls `job.to()`, potentially after earlier site apps were already added. The recipe then cannot replace the recorded configuration without being reconstructed. Please validate all FedJob target-name rules before accepting the configuration—ideally through one shared/public target-name validator—and add coverage for empty and invalid-character site names.

nvflare/recipe/utils.py
```diff
@@ -278,19 +280,50 @@ def _validate_per_site_config_shape(config: Any) -> Dict[str, Dict]:
...
-    clients are present for a run. Per-site values become part of the generated
-    job definition and must never contain actual secret values; see
-    :mod:`nvflare.recipe.secrets`.
+    each site's dictionary. Supported recipes materialize client apps later,
+    before the first client customization or before export or execution. The
+    execution environment still controls which clients are present for a run.
+    Per-site values become part of the generated job definition and must never
+    contain actual secret values; see :mod:`nvflare.recipe.secrets`.
     """
     recipe.set_per_site_config(_validate_per_site_config_shape(config))
```
YuanTingHsieh: [non-blocking] Avoid validating the same shape twice

`_validate_per_site_config_shape()` checks the same dict/non-empty/key/value shape that `Recipe.set_per_site_config()` immediately checks again. The two paths already use different error wording, so these copies can drift as the contract evolves. Since the Recipe method must validate direct calls anyway, could this public helper simply delegate with `recipe.set_per_site_config(config)` and keep the generic shape contract in one place?

YuanTingHsieh (approved):

PR #4903, opened 2026-07-14
Make Recipe job storage private

nvkevlu:
Makes Recipe job storage private.

### Description

Follow-up to #4873, which migrated public examples and documentation away from direct `recipe.job` access.

This PR completes the cleanup by making the generated `FedJob` an internal Recipe implementation detail:

- Renames `Recipe.job` to `Recipe._job`.
- Updates Recipe internals, utilities, framework-specific recipes, and tests.
- Removes temporary public `self.job` assignments from XGBoost recipes.
- Updates integration paths to use public Recipe APIs where appropriate.
- Adds coverage confirming that Recipe no longer exposes a public `job` attribute.
- Rewords errors so they do not direct users toward generated-job internals.

No `recipe.job` references remain in the repository. No runtime behavior changes are intended; users should customize recipes through the supported Recipe APIs.

### Testing

- Recipe-focused unit tests: `510 passed, 23 skipped`
- Recipe utility tests: `96 passed`
- Multi-study ProdEnv integration: passed
- Per-site export regression test: passed
- Server- and client-side MLflow simulations: `2 passed`
- `./runtest.sh --skip-install -s`: passed

### Types of changes

- [ ] Documentation updated.


nvflare/app_common/np/recipes/fedavg.py
```diff
@@ -208,8 +208,8 @@ def add_cse_validator_if_needed(self):
 
         # Check if validation task is explicitly configured (not just via wildcard)
         has_explicit_validator = False
-        if hasattr(self.job, "_deploy_map"):
-            for target, app in self.job._deploy_map.items():
+        if hasattr(self._job, "_deploy_map"):
```
YuanTingHsieh: Is this introspection still necessary? It originated when the generic NumPy CSE path also had to handle standalone recipes that already configured a validator. The hook now lives specifically on `NumpyFedAvgRecipe`, which has no supported API for registering a dedicated validation executor. Reaching through `_job._deploy_map` also works against this PR's encapsulation goal. Could we add `NPValidator` unconditionally here—relying on `add_cross_site_evaluation()`'s idempotency—or make a custom CSE validator an explicit recipe option if that use case must be supported?

YuanTingHsieh (commented):
One non-blocking design question.

nvflare/app_common/np/recipes/fedavg.py
```diff
@@ -193,40 +193,17 @@ def _setup_model_and_persistor(self, job) -> str:
...
-                            except (TypeError, AttributeError):
-                                continue
-                if has_explicit_validator:
-                    break
-
-        if not has_explicit_validator:
-            # No explicit validator found - add NPValidator for cross-site evaluation
-            validator = NPValidator()
-            self.job.to_clients(validator, tasks=[AppConstants.TASK_VALIDATION])
+        self._job.to_clients(NPValidator(), tasks=[AppConstants.TASK_VALIDATION])
```
YuanTingHsieh: **[P1] Preserve per-site client topology when adding `NPValidator`**

For a `NumpyFedAvgRecipe` configured with `per_site_config`, the job contains separate site apps. Calling `_job.to_clients()` here creates an `@ALL` app. During export, `FedJob._set_all_apps()` prioritizes that app and omits the per-site apps, dropping their training executors; clients are left with the validator but no training executor. Please use the existing topology-aware helper instead:

```suggestion
        self._add_to_client_apps(NPValidator(), tasks=[AppConstants.TASK_VALIDATION])
```

A regression test should export a per-site NumPy FedAvg recipe after adding CSE and verify that each site retains both its training executor and `NPValidator`.

YuanTingHsieh (commented):
One blocking issue found on re-review.

YuanTingHsieh (approved):

PR #4907, opened 2026-07-14
Add comprehensive job statistics reporting

IsaacYangSLA:
## What changed

- add a `JobStatsReporter` that writes human-readable and structured JSON run summaries on the server
- report per-round participation, failures, final status, aggregation key matching, and a separate final aggregation row
- collect server/client task timing, logical communication sizes, transport totals, and communication-versus-computation timing
- collect optional process-scoped CPU, GPU, memory, and battery telemetry from clients
- publish aggregation statistics from the supported weighted aggregation and FedAvg paths
- preserve filtered task results for server-side reporting and handle multiple workflows without merging their rounds
- harden client telemetry validation, output path handling, duplicate submissions, and report generation against malformed or non-finite data

## Why

NVFlare jobs did not have one parseable run artifact covering participation, failures, timing, communication, aggregation consistency, and optional resource utilization. These changes provide both a readable summary and a JSON file containing the collected data for downstream analysis.

## User impact

Users can configure the reporter on the server and clients to obtain `job_stats_run_summary.log` and `job_stats_run_summary.json` in the server run directory. An optional separate error log can also be configured.

## Validation

- `./runtest.sh -s`
- `python3 -m pytest -q tests/unit_test/app_common/widgets/job_stats_reporter_test.py tests/unit_test/app_common/aggregators/in_time_accumulate_weighted_aggregator_test.py tests/unit_test/app_common/aggregators/weighted_aggregation_helper_test.py tests/unit_test/app_common/workflow/fedavg_test.py tests/unit_test/private/fed/server/server_runner_test.py`
  - 204 passed, 2 skipped


nvflare/app_common/workflows/base_model_controller.py
```diff
@@ -264,6 +264,7 @@ def _process_result(self, client_task: ClientTask, fl_ctx: FLContext) -> None:
         try:
             self.event(AppEventType.BEFORE_CONTRIBUTION_ACCEPT)
             accepted = self._accept_train_result(client_name=client_name, result=result, fl_ctx=fl_ctx)
+            self._set_ctx_prop_preserving_attrs(fl_ctx, AppConstants.AGGREGATION_ACCEPTED, accepted)
```
YuanTingHsieh, line 290: [P1] This flag only means the return code passed `_accept_train_result`; `FLModel` conversion and callback aggregation happen afterward and can fail or skip the update. FedAvg explicitly skips empty `params`, but `JobStatsReporter` retains this accepted-client signal and can report 100% participation even when final stats have `accepted_contributions=0`. Please publish a definitive signal only after conversion/enqueue succeeds, or make final aggregation contributors replace this provisional state.

nvflare/app_common/widgets/job_stats_reporter.py
```diff
@@ -0,0 +1,1940 @@
...
+        When the round saw per-client acceptance information (accepted or rejected names),
+        participation means the contribution was accepted. Otherwise — relay/cyclic flows that
+        never report acceptance, or aggregators that publish stats without contributor names —
+        participation is measured by tasks completed with an OK result instead. The condition
+        deliberately keys on the NAME sets, not on aggr_stats being present, so a stats dict
+        without contributor names cannot zero out participation for a successful round.
+        """
+        if rec.accepted_clients or rec.rejected_clients:
+            return rec.accepted_clients
+        return completed_by_round.get((rec.workflow, rec.round_num), set())
```
YuanTingHsieh, line 1224: [P2] When contributor names are absent, this promotes every OK-completed task to an accepted contribution even if `AGGREGATION_STATS` explicitly reports a smaller `accepted_contributions` count, including zero. The summary can therefore contradict its own aggregation stats. Please use the published count for numeric participation/status and only fall back to completed tasks when that count is also unavailable.

nvflare/app_common/workflows/base_fedavg.py
```diff
@@ -255,6 +257,19 @@ def aggregate(self, results: List[FLModel], aggregate_fn=None) -> FLModel:
         self._results = []
         self._set_metrics_aggregation_info(aggr_result)
 
+        # Reporting stats must never break a working aggregation: params is typed Any (custom
+        # aggregate_fn callers may use ndarrays etc.), so only dict params contribute key stats
+        # and any unexpected error is swallowed.
+        try:
+            contributions = {_get_client_name(r): list(r.params.keys()) for r in results if isinstance(r.params, dict)}
+            if contributions:
```
YuanTingHsieh, line 270: [P2] If `contributions` is empty or stats computation raises, no value is written and an `AGGREGATION_STATS` value from a previous round remains on the reused `FLContext`. `AFTER_AGGREGATION` can then expose stale stats, whose embedded round makes the reporter update the wrong round. Please clear this property at the start of every aggregation or publish an explicit current-round unavailable value.

nvflare/app_common/widgets/job_stats_reporter.py
```diff
@@ -0,0 +1,1940 @@
...
+
+    def _run(self):
+        while not self.stop_event.wait(self.interval):
+            self._sample(include_cpu=True)
+
+    def _sample(self, include_cpu: bool):
+        sample = {"time": time.time()}
+        if self.process:
+            try:
+                sample["memory_rss_mb"] = self.process.memory_info().rss / _MB
```
YuanTingHsieh: [P2] CPU percent and RSS are sampled only from the parent client process, while the GPU path correctly includes the process family. ScriptRunner and common Client API jobs execute training in child processes, so these CPU/memory figures omit the actual workload. Please aggregate CPU and RSS across the same process family.

nvflare/app_common/widgets/job_stats_reporter.py
```diff
@@ -0,0 +1,1940 @@
...
+            filename: human-readable report written to the server run directory.
+            json_filename: structured report filename, written by default. Set to None to disable JSON output.
+            error_filename: optional separate error report, created only when errors occurred.
+            collect_resources: whether client instances sample CPU/GPU/memory/battery during task execution.
+            resource_sample_interval: resource sampling interval in seconds.
+        """
+        super().__init__()
+        if resource_sample_interval <= 0:
+            raise ValueError("resource_sample_interval must be greater than 0")
+        _validate_report_filename(filename, "filename")
```
YuanTingHsieh, line 729: [P2] The three output names are individually valid even when they resolve to the same path. `END_RUN` writes text, JSON, and errors sequentially, so later artifacts silently overwrite earlier ones. Please resolve the configured paths and reject collisions during initialization.

YuanTingHsieh (dismissed):
The empty-round issue from the earlier review has been fixed in the latest force-push. I found five remaining correctness issues in acceptance/status reporting, aggregation-state handling, resource accounting, and report output handling.

nvflare/app_common/workflows/base_model_controller.py
IsaacYangSLA: Fixed in `ee8788dda`. `_process_result` now publishes `AGGREGATION_ACCEPTED` only after FLModel conversion and callback processing complete. Callbacks can return `False` for deliberately skipped results, and FedAvg now does that for empty parameters. Added regression tests for conversion failures and callback skips.

nvflare/app_common/workflows/base_fedavg.py
IsaacYangSLA: Fixed in `ee8788dda`. `BaseFedAvg.aggregate()` now clears `AGGREGATION_STATS` before each aggregation, so empty/unsupported contributions and stats-computation failures cannot expose the previous round's value. The regression test preloads stale stats and verifies they are cleared when no new stats can be produced.

nvflare/app_common/widgets/job_stats_reporter.py
IsaacYangSLA: Fixed in `ee8788dda`. A published `accepted_contributions` value is now authoritative for numeric participation and status, including zero; completed OK tasks are used only when that count is unavailable. Counts-only stats no longer invent accepted client names. Added regressions for both partial and zero accepted counts.

nvflare/app_common/widgets/job_stats_reporter.py
IsaacYangSLA: Fixed in `ee8788dda`. Report filenames are normalized during initialization and text/JSON/error path collisions are rejected before any output is written. Added coverage for exact collisions and equivalent paths containing `./` across all output combinations.

nvflare/app_common/widgets/job_stats_reporter.py
IsaacYangSLA: Fixed in `ee8788dda`. CPU utilization and RSS are now summed across the parent client process and all recursive children, using the same process-family scope as GPU accounting. The regression test verifies combined parent/child CPU and memory values.

nvidianz (approved):

PR #4914, opened 2026-07-16
Document custom type packaging across job apps

YuanTingHsieh:
## What changed

- clarify that Python import discovery during job export is best-effort and scoped to the target app
- document that modules defining FOBS types must be available in every app that serializes or deserializes those types
- document the Recipe `add_client_file()` and `add_server_file()` helpers for packaging shared modules in both generated app types
- document the existing NVFlare 2.8 `FedJob.add_file_to()`, `add_file_to_server()`, and `add_file_to_clients()` APIs for explicit placement
- show shared all-clients and per-site client app patterns, including exact `dest_dir` mappings for fully qualified module names
- warn that mixing all-clients and per-site placement calls creates a conflicting topology that export resolves silently by prioritizing the all-clients app and dropping named client apps

## Why

ScriptRunner may discover a custom type module imported by a client training script and package it in the client app. That discovery does not propagate the module to the server, so FOBS deserialization can fail there with `No module named ...`.

The existing Recipe and Job APIs already provide explicit file-placement methods. The missing piece is documentation that distinguishes best-effort import discovery from the deployment contract, shows users how to place shared type modules in every required app, and makes the client-app topology requirements explicit.

## Impact

Documentation only. Runtime behavior and existing Recipe file-helper semantics are unchanged.

## Validation

- cumulative diff against `upstream/main` contains only three documentation files
- `git diff --check`
- `./runtest.sh -s --skip-install` — black, isort, flake8, and agent-skill lint passed in an isolated environment


docs/programming_guide/execution_api_type/client_api.rst
```diff
@@ -442,5 +442,17 @@ For other custom classes, you will need to write a dedicated custom decomposer a
...
+app containing that entry point; it does not make the module available to the
+server. With a Recipe, add the defining module to both generated app types::
+
+    custom_type_file = "src/custom_types.py"
+    recipe.add_client_file(custom_type_file)
+    recipe.add_server_file(custom_type_file)
+
+Keep the source package layout consistent with the type's fully qualified
+module name. Add imported dependencies explicitly when they are required on
+both sides rather than relying on best-effort discovery.
```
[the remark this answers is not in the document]
YuanTingHsieh: Our intended user interface is Recipe

docs/programming_guide/execution_api_type/client_api.rst
```diff
@@ -442,5 +442,17 @@ For other custom classes, you will need to write a dedicated custom decomposer a
 using fobs.register on both the server side and client side, as well as in train.py.
 
 Please note that for the custom data class to work, it must be placed in a separate file from train.py.
+The module must be packaged in every app that serializes or deserializes the
+type. Import discovery from ``train.py`` is best-effort and scoped to the client
```
chesterxgchen: train.py is not shown in the folder, is this client.py ?

docs/programming_guide/execution_api_type/client_api.rst
```diff
@@ -442,5 +442,17 @@ For other custom classes, you will need to write a dedicated custom decomposer a
...
+app containing that entry point; it does not make the module available to the
+server. With a Recipe, add the defining module to both generated app types::
+
+    custom_type_file = "src/custom_types.py"
```
chesterxgchen: is "src" folder important here ?

docs/programming_guide/fed_job_api.rst
```diff
@@ -231,6 +247,58 @@ Example:
...
+      job.add_file_to(custom_type_file, site)
+
+By default, a file is copied by basename into the app's ``custom`` directory.
+Use ``dest_dir`` to place it in a relative subdirectory when its importable
+module path requires one. For example:
+
+.. code-block:: python
+
+  # custom_enum.CustomEnum -> custom/custom_enum.py
+  job.add_file_to_server("src/custom_enum.py")
```
chesterxgchen: "src"  -- why do we illustrate src
?

chesterxgchen (commented):
seems not clear to new user by reading these doc, what is the intended for. 

the example using train.py, "src" folder as illustration, but make the impression that they are required.

docs/programming_guide/execution_api_type/client_api.rst
YuanTingHsieh: Yes—this refers to the Client API entry script shown earlier as `client.py`. I updated this section to use `client.py` consistently and clarified that the custom type should be defined in a separate importable module.

docs/programming_guide/execution_api_type/client_api.rst
YuanTingHsieh: No, an `src` folder is not required. I replaced this with a top-level `custom_types.py` example; the requirement is only that the packaged path match the type's fully qualified module name.

docs/programming_guide/fed_job_api.rst
YuanTingHsieh: `src` is not required. I changed the primary example to a top-level `custom_enum.py` and kept a separately labeled optional package-layout example using `my_package/custom_enum.py` with `dest_dir="my_package"`.

holgerroth (approved):
Docs-only change, verified against the implementation on main: `add_file_to*` and the recipe file helpers exist with the documented semantics, explicit placement skips import discovery (`_copy_file_sources`), the all-clients precedence / silent drop of per-site apps matches `_set_all_apps`, the destination-collision `ValueError` and the discovery search order/scoping are all accurate. The topology warning documents a real silent footgun. LGTM.

nvkevlu (approved):

PR #4922, opened 2026-07-20
Add FedCE research recipe support

nvkevlu:
Adds dedicated PyTorch recipe support for FedCE.

### Description

FedCE uses the FedAvg global-model workflow but changes how client updates are evaluated and aggregated. Because it introduces its own client/server contract and aggregation behavior, it is provided as a separate `FedCERecipe` instead of a FedAvg configuration flag.

The PR includes:

- A FedCE client helper and contribution-aware model aggregator.
- A dedicated `FedCERecipe` using PyTorch weight differences.
- Recipe exports and CLI discovery support.
- Documentation describing how to use the recipe.
- Unit tests covering aggregation, validation, recipe construction, and export behavior.

FedSM and other personalized-learning algorithms are outside the scope of this PR.

### Validation

The original FedCE implementation and the new recipe were compared using the documented six-client prostate segmentation experiment.

Both implementations completed all 100 rounds successfully with no fatal validation errors. Their late-round mean global Dice scores were very similar: approximately `0.845` for the original implementation and `0.843` for the recipe. FedCE contribution coefficients also remained normalized throughout the run.

These results provide confidence that the recipe preserves the expected FedCE research behavior. 

### Types of changes

- [ ] Documentation updated.


holgerroth:
The contribution scoring and the actual aggregation currently operate on different parameter scopes:

- Contribution weights are computed on the **intersection** of all clients' returned keys, further filtered by `trainable_param_names` (`_get_cosine_param_names`).
- `aggregate_model()` then applies those weights to **all** params each client returned via `WeightedAggregationHelper` — including non-trainable buffers and any client-exclusive keys.

For buffers (`running_mean`, `num_batches_tracked`, ...) this seems fine (they get contribution-weighted instead of evenly averaged), but client-exclusive keys would be aggregated among only their contributors, entirely outside the contribution-scoring rationale (related to what Greptile flagged).

Proposal:
1. Log a warning in `aggregate_model()` when clients' returned key sets differ from each other (or from the scoring set), so silent divergence is visible.
2. Document the assumption in the `FedCEModelAggregator` docstring: FedCE expects homogeneous parameter sets across clients; contribution weights are derived from the common trainable params but applied to everything returned, including buffers.

holgerroth:
On the validation: the global Dice comparison (≈0.845 vs ≈0.843) is a good sanity check, but the signature result of FedCE — and what the research example documents — is the contribution-coefficient behavior, see the curves in [research/fed-ce](https://github.com/NVIDIA/NVFlare/tree/main/research/fed-ce#metrics-and-fedce-contribution-estimation-curves): `client_Promise12`'s coefficient rises over training due to its domain shift while the other clients' coefficients gradually decrease.

Could you attach the equivalent `fedce_coef` curves from the recipe run (and ideally per-client validation Dice) so we can confirm the recipe reproduces this behavior?

It would also be good to include the client training script / job config used for the prostate comparison — or better, add a recipe-based job under `research/fed-ce` — so the experiment is reproducible.

nvkevlu:
I enabled per-round contribution tracking in the recipe validation using NVFlare’s TensorBoard experiment-tracking path. The client emits FedCE_Coef each round, and the recipe installs the TensorBoard receiver.

I ran both the original implementation and FedCERecipe for 100 rounds using the documented six-client prostate configuration. Both runs completed successfully and produced complete per-client Dice and contribution series.

- original implementation: job `30891157`
- `FedCERecipe`: job `30918334`

image

Promise12 shows the expected behavior under the recipe. Its contribution coefficient rises from **0.169510 at round 1** to **0.273279 at round 99**—an absolute increase of **0.103769** (**61.2%**)—and peaks at **0.275921 at round 73**. It finishes with the largest coefficient. The original implementation similarly rises from **0.168981** to **0.266813**.

Across all clients and rounds 1–99, the original and recipe coefficient series have a correlation of **0.9976** and a mean absolute difference of **0.00281**. The recipe coefficients remain normalized: their per-round sums differ from 1 by at most **2.98e-8**.

The plot starts contribution curves at round 1 because round 0 has no prior-round contribution coefficient. These results reproduce the research example's qualitative behavior: Promise12 receives increasing weight while the other clients' weights generally decrease or remain near the uniform reference.

holgerroth:
Thanks @nvkevlu — this is convincing. The Promise12 trajectory (0.1695 → 0.2733, finishing largest while the others stay near or below the uniform reference) reproduces the signature behavior from the [research example](https://github.com/NVIDIA/NVFlare/tree/main/research/fed-ce#metrics-and-fedce-contribution-estimation-curves), and your rerun of the original implementation ending at 0.2668 matches the published figure, so the baseline itself reproduced. The 0.9976 correlation with ~0.0028 mean absolute difference is consistent with the minor numerical differences between the implementations (clip/normalize ordering, epsilon) plus run-to-run noise, and the per-round sums within 3e-8 of 1 confirm the recipe's by-construction normalization.

Two remaining items before approval:

1. **Reproducibility**: the job IDs are internal references. Could you include the client training script and job config used for this comparison — ideally as a recipe-based job under `research/fed-ce` — so the experiment is externally reproducible? The client script that emits `FedCE_Coef` and implements the minus-model contract is exactly the piece the PR doesn't show.
2. **The earlier scope comment**: the key-set mismatch warning and the docstring documenting the assumption (homogeneous parameter sets across clients; contribution weights applied to all returned params including buffers) haven't landed on the branch yet. Since the buffer weighting matches the original `FedCEAssembler` behavior, this is a documentation/logging fix, not an algorithm change.

holgerroth (approved):
All review items are resolved:

- **Contract guarded and documented**: homogeneous parameter names/shapes are now enforced with clear per-client errors, and the docstring documents that contribution scores derive from the trainable subset while the weights apply to all returned params, including buffers (matching the original `FedCEAssembler` behavior).
- **Validation evidence is convincing**: Promise12's coefficient trajectory reproduces the research example's signature behavior, with 0.9976 correlation to the original implementation across all clients and rounds.
- **Example migration verified**: the recipe-based `research/fed-ce` job preserves the original training contract exactly — minus-model formula, round-0 handling, historical-mean scoring, transforms, optimizer, and metrics all match the original learner.

Optional follow-up (non-blocking): `_get_param_shape` materializes lazy/disk-offloaded tensors just to read shapes, so with `enable_tensor_disk_offload=True` tensors are read from disk twice per round. Could be optimized to read shape metadata without materializing.

Nice work — approving.

PR #4930, opened 2026-07-20
Add Slurm job launcher

pcnudde:
## Summary

- Add client and server Slurm job launchers behind the existing `JobLauncherSpec` and `JobHandleSpec` interfaces.
- Support one-node bare, Apptainer, and Pyxis execution plus bare multi-node allocations with application-owned fan-out.
- Add Slurm-aware deployment preparation, study runtime policy, portable resource requests, BYOC image selection, accounting-backed monitoring, ownership-checked cancellation, and transient job artifacts.
- Add operator and design documentation plus focused launcher/deploy tests.

## User impact

Sites configure trusted Slurm policy in `slurm.yaml` and prepare the complete live workspace directly at a stable shared-filesystem path. They then start the parent from that workspace and submit NVFlare jobs normally. Slurm controls placement and resource enforcement while NVFlare reports the normal job lifecycle. Working `slurmdbd` accounting is required.

Preparing again to the same output replaces the complete workspace, so operators must stop the parent and preserve required runtime data first.

## Design notes

- Scheduler commands use structured argv and a scrubbed environment.
- Active ownership is verified by Slurm ID, user, deterministic name, and full NVFlare job comment before cancellation.
- Terminal state and exit attribution come from `sacct`.
- Partition, account, QOS, sandbox, setup, and routing remain trusted site or study policy.
- The internal worker-to-parent connection uses clear TCP on a trusted site network.
- Slurm worker launches consume the generic credential environment contract from #4929, keeping bootstrap credentials out of generated command arguments.


pcnudde:
@greptileai review

nvflare/app_opt/job_launcher/slurm/scheduler_client.py
```diff
@@ -0,0 +1,252 @@
...
+
+    def submit(self, argv: list[str], timeout: float) -> SubmissionResult:
+        command = self._run(argv, timeout)
+        lines = [line.strip() for line in command.stdout.splitlines() if line.strip()]
+        matched = _SUBMIT_ID_RE.fullmatch(lines[0]) if len(lines) == 1 else None
+        return SubmissionResult(
+            command=command,
+            job_id=matched.group(1) if matched else None,
+            cluster=matched.group(2) if matched else None,
+        )
```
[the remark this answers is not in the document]
pcnudde: This is intentional. A complete parsable job ID is authoritative even when a site wrapper or plugin exits nonzero: Slurm may already own the allocation, and discarding the ID would orphan accepted work. The scheduler-client regression test explicitly covers this contract. Launch fails when there is no valid ID or the command times out; a cluster-qualified ID follows the separate best-effort cancellation failure path.

nvflare/app_opt/job_launcher/slurm/batch.py
```diff
@@ -0,0 +1,195 @@
...
+    return f"{variable}={shlex.quote(configured or default)}"
+
+
+def _common_environment(plan: LaunchPlan, config: SlurmConfig) -> list[str]:
+    lines = ['[[ -r "${_nvfl_secret}" ]] || { echo "NVFlare Slurm secret file is unavailable" >&2; exit 100; }']
+    if plan.setup:
+        lines.extend(["export SLURM_EXPORT_ENV=ALL", plan.setup])
+    lines.extend(
+        [
+            "export SLURM_EXPORT_ENV=ALL",
```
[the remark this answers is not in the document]
pcnudde: The two exports have different scopes. The first makes the full allocation environment available to trusted setup commands, including setup-time srun. The second restores the required value after setup may have changed it, before the launcher starts the worker. Removing the first changes supported setup behavior, so both are retained intentionally.

examples/advanced/slurm/slurm_torchrun_node.py
```diff
@@ -0,0 +1,314 @@
...
+    torchrun_argv = _build_torchrun_argv(
+        sys.executable,
+        nnodes,
+        node_rank,
+        master_addr,
+        master_port,
+        nproc_per_node,
+        training_argv,
+    )
+    os.execv(sys.executable, torchrun_argv)
```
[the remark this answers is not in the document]
pcnudde: Superseded by the later scope reduction: the example helper and its tests were removed from this PR, so this thread no longer applies to the current diff.

pcnudde:
Greptile follow-up on the remaining summary-only observation:

- Parent directives remain unchanged. The public deploy flow calls `validate_config()` before backend `prepare()`. Validation rejects unsupported directive names, types, and whitespace; integer values are safely stringified during rendering. The backend `prepare()` function is not an independent public entry point.

The scheduler-ownership and setup-environment suggestions were answered in their inline threads.

nvflare/app_opt/job_launcher/slurm/manager.py
```diff
@@ -0,0 +1,476 @@
...
+            if record.state in _PENDING_STATES:
+                now = self._monotonic()
+                if handle.pending_started_at is None:
+                    handle.pending_started_at = now
+                if now - handle.pending_started_at >= handle.pending_timeout:
+                    self.logger.warning("Slurm job %s exceeded pending_timeout", handle.job_id)
+                    handle._request_cancel()
+            if handle.cancel_requested:
+                self.adapter.cancel(handle.job_id, timeout=_COMMAND_TIMEOUT)
+            return JobReturnCode.UNKNOWN
```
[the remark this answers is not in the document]
pcnudde: This is intentional. `UNAVAILABLE` means there is no authoritative scheduler observation: the allocation may still be pending, running, or terminal. Turning a prolonged scheduler outage into a terminal NVFlare result would synthesize state and could release the framework while Slurm work continues. The five-miss budget therefore counts only successful `sacct` queries that return `NOT_FOUND`; outages do not consume it. `wait()` runs in the framework's per-job waiter and remains pending until the scheduler recovers. Likewise, `pending_timeout` measures observed Slurm pending time and cannot safely advance from missing observations. `test_scheduler_outage_never_synthesizes_terminal_state` and `test_accounting_outage_does_not_consume_healthy_miss_budget` codify this contract. No code change is needed.

pcnudde:
Greptile follow-up on the second confidence-score observation:

- `terminate()` follows the `JobHandleSpec` request-style contract and does not report synchronous cancellation success. It first records `cancel_requested` and user-abort intent, then makes an immediate polling/cancellation attempt.
- A scheduler protocol error means ownership could not be safely verified, so issuing `scancel` would be unsafe. The exception is logged, the cancel request remains set, and the framework's subsequent waiter polls retry cancellation once a valid live scheduler record is available. An unavailable `scancel` similarly leaves the request pending for retry.

This is deliberate ownership-preserving behavior, not a lost cancellation request. No code change is needed. The inline scheduler-outage thread has also been answered and resolved. References in the generated summary to Slurm staging and the former prepared-kit layout are stale; staging was removed in the current head.

pcnudde:
@greptileai review

docs/user_guide/admin_guide/deployment/slurm_job_launcher.rst
```diff
@@ -0,0 +1,324 @@
...
+     - Read-only Pyxis/Enroot container
+     - Trusted container packaging
+   * - ``none``
+     - Python directly in the allocation
+     - Site-trusted code; required for multi-node jobs
+
+Prerequisites
+=============
+
+Before starting a parent, verify:
```
YuanTingHsieh, line 40: since this is user_guide suggest define what "parent" mean first

docs/user_guide/admin_guide/deployment/slurm_job_launcher.rst
```diff
@@ -0,0 +1,324 @@
...
+does not need Slurm commands. The submission host that runs the generated
+``submit_command`` needs ``sbatch`` on ``PATH``. The runtime parent host needs
+all four parent commands after its service environment or
+``parent.environment_setup`` has run.
+
+Configure the Site
+==================
+
+Create ``slurm.yaml`` beside the startup kit. This example runs the parent on a
+login or service host:
```
YuanTingHsieh, line 86: suggest we define these terms first

docs/user_guide/admin_guide/deployment/slurm_job_launcher.rst
```diff
@@ -0,0 +1,324 @@
...
+  at the same absolute paths on all participating nodes. The shared filesystem
+  supports ``O_EXCL`` and atomic rename.
+- Compute nodes can reach ``parent_host:internal_port``. Multi-node jobs also
+  require connectivity among their allocated nodes.
+- Slurm partition, association, QOS, reservation, cgroup, and device policies
+  enforce the site's resource limits.
+
+The prepare output is the runtime workspace. The parent and worker allocations
+must see it at the same absolute path. Every workspace must be private, owned
+by the runtime account, and dedicated to one NVFlare site or federation.
```
YuanTingHsieh, line 64: after prepare are those the generated nvflare startup kits? so the kit/workspace folder needs to be copied into the node where the nvflare parent process going to be run, and we assume the parent and the child/worker have a share filesystem.

docs/user_guide/nvflare_cli/deploy_command.rst
```diff
@@ -272,14 +275,23 @@ The command requires Kubernetes CLI access to the target cluster. It uses
...
+  same ownership check
 - patches ``helm_chart/values.yaml`` so the parent pod mounts the ConfigMap at
   ``workspace_mount_path/local`` and the Secret at
   ``workspace_mount_path/startup``
+- patches the prepared ``K8sJobLauncher`` to the effective stage namespace so
+  its child Pods use the same namespace as the parent and Helm RBAC
 - records the resolved namespace and object names so they can be removed by
   ``nvflare deploy k8s unstage``
 
+Each prepared chart has a random staging-owner token. Both objects carry it as
```
holgerroth: These K8s stage/unstage doc changes (here and in the `unstage` section around line 324) describe an owner-token mechanism — a random staging-owner token carried as label + annotation, ownership-verified replace/delete, and stage patching `K8sJobLauncher` to the effective namespace — that isn't implemented by this PR, and `nvflare/tool/deploy/k8s_stage.py` on `main` has no token logic either. This looks like content from a sibling branch that leaked into this diff. Suggest dropping the K8s doc edits from this PR (or sequencing it after the PR that implements the owner-token behavior), so the docs don't promise safety guarantees operators don't actually have.

holgerroth (commented):
One note on the PR description (no diff line to anchor it to): the summary bullet "Stop logging the client registration token and SSID" has no corresponding change in this diff — that change landed with #4929. Suggest removing the bullet so the merge commit message stays accurate.

docs/user_guide/admin_guide/deployment/slurm_job_launcher.rst
```diff
@@ -0,0 +1,324 @@
...
+Artifact removal prevents a pending allocation from starting unless the same
+job ID is relaunched before it starts.
+
+To change the workspace, stop the parent and prepare to a new ``--output``.
+Preparing to an existing output replaces it, so copy any data that must survive
+before running prepare.
+
+Server-job queue and startup time must fit within the client runner-sync timeout
+(60 seconds by default). Provide prompt server capacity or increase
+``max_runner_sync_timeout`` when the site cannot meet that bound.
```
YuanTingHsieh: this sentence is unclear to me as why these values need to be like that and how these values can be configured

docs/design/study_runtime_config.md
```diff
@@ -192,24 +216,29 @@ Merge order (later wins):
...
+
 ```text
 local/study_runtime.yaml present  → v2 strict parser (format_version: 2 required)
 local/study_data.yaml only        → v1 frozen parser, behavior unchanged
 both present (or v2 + legacy launcher args) → hard error
 ```
 
+Slurm requires the v2 file; a legacy-only file or coexistence of both files is a hard error.
+
 - `format_version: 2` required — the gate for future format evolution.
```
YuanTingHsieh, line 237: Is v1 the one we relelased in 2.8?

docs/user_guide/core_concepts/job.rst
```diff
@@ -149,9 +158,22 @@ launcher modes without choosing which mode a site uses.
...
+pod GPU request. Slurm translates it to a one-node GPU request unless the
+``slurm`` block supplies explicit topology.
+
+The Slurm job block accepts ``image``, ``nodes``, ``gpus_per_node``,
+``cpus_per_node``, ``mem_per_node`` (MiB), ``time``, and
+``pending_timeout``. A job cannot select a Slurm partition, account, QOS,
+sandbox, setup command, or raw scheduler flags; those are site-owned and may
+be overridden per study. A job image requires BYOC authorization and overrides
+study/site image defaults. It must be an absolute site-visible existing file.
+``pending_timeout`` may only reduce the site's configured value. Multi-node
```
YuanTingHsieh: so it is not overwrite, means it can't increase?

docs/design/job_launcher_and_job_handle.md
```diff
@@ -150,9 +164,14 @@ If a job lacks required configuration for the site's launcher (e.g. no image on
...
 1. Merge `launcher_spec["default"][mode]` with `launcher_spec[site_name][mode]` (site wins on conflict).
-2. Keep `resource_spec` separate from launcher settings; it is not a place for new Docker or K8s launcher configuration.
+2. Keep `resource_spec` separate from launcher settings; it is not a place for new backend-specific configuration.
 
-`resource_spec` (distinct from `launcher_spec`) is scheduler-facing: the scheduler reads it at job admission time to decide if the site has the required hardware. Docker and K8s use flat `resource_spec[site]["num_of_gpus"]` for GPU requests.
+`resource_spec` (distinct from `launcher_spec`) is scheduler-facing: the scheduler reads it at job admission time to
+decide if the site has the required hardware. Docker, K8s, and Slurm use flat
+`resource_spec[site]["num_of_gpus"]` as the portable GPU total. Slurm permits only `nodes`, `gpus_per_node`,
+`image`, `cpus_per_node`, `mem_per_node`, `time`, and reduce-only `pending_timeout` in its job block. A job image
+requires BYOC authorization and overrides study/site image defaults. Routing, billing, sandbox, setup, and raw
```
YuanTingHsieh: do we have "billing" flags?

nvflare/app_opt/job_launcher/slurm/config.py
```diff
@@ -0,0 +1,376 @@
...
+SACCT_FORMAT = "JobIDRaw%32,JobName%128,User%64,State%64,ExitCode%32"
+
+_MAX_STDOUT_BYTES = 16 * 1024 * 1024
+_MAX_STDERR_BYTES = 64 * 1024
+_SUBMIT_ID_RE = re.compile(r"^([0-9]+)(?:;([A-Za-z0-9._-]+))?$")
+_ENV_NAME_RE = re.compile(r"^[A-Za-z_][A-Za-z0-9_]*$")
+
+SLURM_SBATCH_DIRECTIVES = ("partition", "account", "qos", "time", "constraint", "reservation")
+SLURM_PARENT_EXECUTABLES = ("sbatch", "squeue", "sacct", "scancel")
+SLURM_COMPUTE_EXECUTABLES = ("apptainer", "srun")
```
YuanTingHsieh, line 50: srun is pyxis/enroot only? so it does not belong to SLURM?

nvflare/app_opt/job_launcher/slurm/batch.py
```diff
@@ -0,0 +1,198 @@
...
+def _submission_argv(plan: LaunchPlan, job_dir: str, job_name: str, marker: str, config: SlurmConfig) -> list[str]:
+    argv = [
+        config.executables["sbatch"],
+        "--parsable",
+        "--no-requeue",
+        "--export=NIL",
+        f"--chdir={plan.run_dir}",
+        f"--nodes={plan.resources.nodes}",
+        f"--ntasks={plan.resources.nodes}",
+        "--ntasks-per-node=1",
```
YuanTingHsieh, line 175: will people want to adjust tasks-per-node? what is it for?

nvflare/private/fed/utils/app_authz.py
```diff
@@ -46,7 +46,7 @@ def validate_app(app_path: str) -> (str, dict):
 
     @staticmethod
     def derive_local_app_info(app_info: dict, job_meta: dict, site_name: str) -> dict:
-        for mode in ("docker", "k8s"):
+        for mode in ("docker", "k8s", "slurm"):
```
YuanTingHsieh, line 49: for these strings do we need to use some CONSTANT in code?

nvflare/tool/deploy/templates/slurm/parent.slurm
```diff
@@ -0,0 +1,4 @@
+#!/usr/bin/env bash
+@@NVFLARE_SBATCH_DIRECTIVES@@set -euo pipefail
+@@NVFLARE_ENVIRONMENT_SETUP@@WORKSPACE_ROOT=@@NVFLARE_WORKSPACE_ROOT@@
+exec "$WORKSPACE_ROOT/startup/start_slurm.sh"
```
YuanTingHsieh, line 4: why do we need to have parent.slurm and start_slurm.sh separate?

docs/user_guide/nvflare_cli/deploy_command.rst
pcnudde: Agreed. Those K8s ownership and namespace-patching claims were sibling-branch carryover and are not implemented here. Commit 66c1cf0ce restores this K8s stage/unstage section to the behavior on main while retaining only the Slurm additions elsewhere in the shared CLI page. I also audited the rest of the PR for the removed Slurm staging/UUID/claim/attempt/helper designs and removed one unrelated Docker design-doc inventory edit. Documentation link checks, Sphinx, and style pass.

nvflare/app_opt/job_launcher/study_runtime.py
```diff
@@ -179,9 +241,9 @@ def _parse_pod_template(value, file_path: str, allow_pod_template: bool) -> dict
     return _validate_pod_template(value, "inline pod_template", file_path)
 
 
-def _parse_docker_kwargs(study: str, entry, file_path: str, allow_docker_kwargs: bool) -> dict:
+def _parse_docker_kwargs(study: str, entry, file_path: str, launcher_mode: str) -> dict:
     label = f"studies.{study}.docker_kwargs"
-    if not allow_docker_kwargs:
+    if launcher_mode != "docker":
```
YuanTingHsieh, line 246: if launcher_mode != "docker", maybe we should not even call this method?

nvflare/app_opt/job_launcher/study_runtime.py
```diff
@@ -169,8 +231,8 @@ def _validate_pod_template(pod_template, label: str, file_path: str) -> dict:
     return pod_template
 
 
-def _parse_pod_template(value, file_path: str, allow_pod_template: bool) -> dict:
-    if not allow_pod_template:
+def _parse_pod_template(value, file_path: str, launcher_mode: str) -> dict:
```
YuanTingHsieh, line 234: if launcher_mode != "k8s", maybe we should not even call this method?

nvflare/app_opt/job_launcher/study_runtime.py
```diff
@@ -140,6 +166,42 @@ def _require_str(value, label: str, file_path: str) -> str:
...
+        raise _error(file_path, f"{label} must not contain NUL.")
+    try:
+        value.encode("utf-8")
+    except UnicodeEncodeError as e:
+        raise _error(file_path, f"{label} must be valid Unicode text encodable as UTF-8.") from e
+    return value
+
+
+def _require_env_name(value, label: str, file_path: str, launcher_mode: str) -> str:
+    if launcher_mode == "slurm":
```
YuanTingHsieh, line 183: I wonder when this file grows larger for these different "ifs", does it make sense to split them out? I guess it depends on how common/different they are.

nvflare/tool/deploy/slurm_deploy.py
```diff
@@ -0,0 +1,253 @@
...
+            "image",
+            "internal_port",
+            "sbatch_directives",
+            "setup",
+            "python_path",
+            "executables",
+            "forward_env",
+            "parent_host",
+            "poll_interval",
+            "pending_timeout",
```
YuanTingHsieh, line 138: for these constants should we have a CONSTANT lists or something define them in like slurm/config.py? or slurm/def.py? so a centralized place to refer to? not sure where to put them though. just think it might be clear to have some lists so we know what each deploy is expecting for job_launcher

nvflare/app_opt/job_launcher/slurm/scheduler_client.py
```diff
@@ -0,0 +1,263 @@
...
+    def _rows(command: CommandResult, source: str, width: int):
+        for line in command.stdout.splitlines():
+            if not line.strip():
+                continue
+            fields = line.split("|", width - 1)
+            if len(fields) != width:
+                raise SlurmProtocolError(f"malformed {source} row")
+            yield fields
+
+    def submit(self, argv: list[str], timeout: float) -> SubmissionResult:
```
YuanTingHsieh, line 141: for these methods I am just thinking if SLURM has any official python library to do this?

nvflare/app_opt/job_launcher/slurm/manager.py
```diff
@@ -0,0 +1,476 @@
...
+
+def _job_name(job_key: str) -> str:
+    return f"nvfl-{job_key[:8]}"
+
+
+def _is_terminal(state: str) -> bool:
+    return state in _APPLICATION_TERMINAL_STATES or state in _INFRASTRUCTURE_TERMINAL_STATES or state == "CANCELLED"
+
+
+class SlurmJobManager:
```
YuanTingHsieh, line 109: do we need to separate SlurmJobManager vs SlurmJobLauncher?
if so should we have a separate K8sJobManager?

nvflare/app_opt/job_launcher/slurm/scheduler_client.py
```diff
@@ -0,0 +1,263 @@
...
+    result["LC_ALL"] = "C"
+    return result
+
+
+def _parse_state(value: str) -> str:
+    tokens = value.strip().split(maxsplit=1)
+    return tokens[0].removesuffix("+") if tokens else ""
+
+
+class _SlurmCliAdapter:
```
YuanTingHsieh, line 79: do we need to separate this Adapter out of the SlurmManager/SlurmJobLauncher?

YuanTingHsieh (commented):
design and implementatons LGTM, added some questions

docs/user_guide/admin_guide/deployment/slurm_job_launcher.rst
pcnudde: yes, this all needs to be on the lustre filesystem.

docs/design/study_runtime_config.md
pcnudde: v1 is 2.8. v2 is part of an earlier PR in 2.9

nvflare/app_opt/job_launcher/study_runtime.py
pcnudde: Agreed. at some point we need to split it. (like we did with the deploy commands. 
I think currently the file is still small enough.

nvflare/app_opt/job_launcher/slurm/scheduler_client.py
pcnudde: Sadly no. There is an REST api, but it is not always installed. 
The main way people interact with slurm in scripts is the CLI ...

docs/user_guide/admin_guide/deployment/slurm_job_launcher.rst
pcnudde: Agreed. Commit 9d7f6daf3 now defines the parent up front as the long-running client parent process (CP) or server parent process (SP), and distinguishes it from the per-job CJ/SJ processes launched through Slurm.

docs/user_guide/admin_guide/deployment/slurm_job_launcher.rst
pcnudde: Agreed. Commit 9d7f6daf3 adds concise definitions for the prepare host, optional submission host, runtime parent host, compute nodes, and the login/service-host case before those terms are used in the configuration instructions.

docs/user_guide/admin_guide/deployment/slurm_job_launcher.rst
pcnudde: Agreed. Commit 9d7f6daf3 now explains that clients wait for the Slurm-launched SJ, why its queue plus startup time must fit within the 60-second default, and that `max_runner_sync_timeout` is configured in the job's client `config_fed_client.json`. It also links to the timeout troubleshooting guide.

docs/design/job_launcher_and_job_handle.md
pcnudde: Good catch. There is no generic Slurm `billing` flag here. Commit 9d7f6daf3 replaces that wording with the concrete site/study-owned `account` and `QOS` settings.

docs/user_guide/core_concepts/job.rst
pcnudde: Yes. Commit 9d7f6daf3 makes this explicit: a job may override `pending_timeout` only with a smaller positive value than the site's configured timeout; it cannot increase the site limit.

nvflare/tool/deploy/templates/slurm/parent.slurm
pcnudde: start_slurm.sh starts the prepared parent directly on the current host. parent.slurm is an optional client-only sbatch wrapper that adds parent allocation directives and environment setup, then delegates to start_slurm.sh. that one is only used if you want to launch it in the cluster itself

nvflare/private/fed/utils/app_authz.py
pcnudde: I would keep it as it is. Only used here and it is pretty clear

nvflare/app_opt/job_launcher/slurm/scheduler_client.py
```diff
@@ -0,0 +1,263 @@
...
+            return self._runner(argv, timeout)
+        try:
+            completed = subprocess.run(
+                argv,
+                capture_output=True,
+                timeout=timeout,
+                env=_scheduler_env(),
+            )
+        except subprocess.TimeoutExpired:
+            return CommandResult(None, "", "", timed_out=True)
```
YuanTingHsieh, line 114: **[P2] Preserve an accepted job ID on timeout**

`subprocess.TimeoutExpired` can carry partial stdout. If `sbatch --parsable` prints a job ID after accepting the allocation and a site wrapper then hangs, this path discards the only ownership handle; `launch()` subsequently rejects the timed-out submission and removes the artifacts even though the allocation may already have sourced the secret and continued running. Please preserve and bound/decode `e.stdout`, then either adopt a complete parsable ID or ownership-check and cancel it before cleanup. A regression test should cover the allocation starting before artifact removal.

nvflare/app_opt/job_launcher/slurm/manager.py
```diff
@@ -0,0 +1,476 @@
...
+                BindMount(
+                    os.path.realpath(os.path.join(workspace, "local")),
+                    os.path.join(workspace, "local"),
+                    "ro",
+                ),
+            )
+            if plan.sandbox == "apptainer":
+                launcher_mounts += (
+                    BindMount(
+                        os.path.realpath(CONTAINER_RESOLV_CONF),
```
YuanTingHsieh: **[P2] Resolve `/etc/resolv.conf` on the compute node**

This canonicalizes the parent host's resolver symlink and embeds that source path in the later compute-node batch script. On heterogeneous clusters the parent may resolve `/etc/resolv.conf` to a target such as `/run/systemd/resolve/...` that does not exist on compute nodes, causing every Apptainer launch there to fail its bind. Please resolve the source inside the allocation, or otherwise bind the compute node's own `/etc/resolv.conf` rather than the parent host's canonical target.

nvflare/app_opt/job_launcher/slurm/config.py
```diff
@@ -0,0 +1,376 @@
...
+
+_JOB_SLURM_KEYS = {"image", "nodes", "gpus_per_node", "cpus_per_node", "mem_per_node", "time", "pending_timeout"}
+
+_PENDING_STATES = {"PENDING", "CONFIGURING", "RESV_DEL_HOLD"}
```
YuanTingHsieh: **[P2] Include held requeue states in the pending lifecycle**

Slurm can report `REQUEUE_HOLD` and `SPECIAL_EXIT` for jobs requeued into a held/pending state. `active_by_id()` continues to find those jobs, but because these states are neither terminal nor included here, `_poll_handle()` never starts or applies `pending_timeout`; `wait()` can remain blocked until an operator intervenes. Please classify the held requeue states as pending for timeout/cancellation purposes (and consider `REQUEUE_FED` if it is in scope), with a regression test.

YuanTingHsieh (commented):
Three P2 findings from an independent review of the Slurm submission and lifecycle paths.

nvflare/app_opt/job_launcher/slurm/manager.py
pcnudde: K8s is in simpler as it has a nicer API. The SlurmJobManager handles lifecycle of the job over the inferior cli API.

nvflare/app_opt/job_launcher/slurm/scheduler_client.py
pcnudde: Maybe it could be combined, but would not save code. In a way this is the python API on top of the cli that is missing in the official slurm code.

nvflare/app_opt/job_launcher/slurm/config.py
pcnudde: srun is a core Slurm command, not a Pyxis or Enroot command. Pyxis exposes its container options through Slurm srun integration. This launcher currently invokes srun only for the Pyxis backend and classifies it as compute-side because it must exist on the allocated compute node, unlike sbatch, squeue, sacct, and scancel on the runtime parent. No code change is needed.

nvflare/tool/deploy/slurm_deploy.py
pcnudde: I would keep this allowlist inline. It is the single-use schema for the public slurm.yaml job_launcher block. Shared Slurm value semantics are already centralized in the normalization functions and constants in slurm/config.py. Moving this one-use deployment schema into runtime configuration would add coupling without removing meaningful duplication. No code change is needed.

nvflare/app_opt/job_launcher/study_runtime.py
pcnudde: This parser is called only when docker_kwargs is present in the study entry. For a non-Docker launcher, calling it is the strict-validation path that rejects Docker-only configuration. Skipping the parser would silently accept and ignore an invalid key, so no code change is needed.

nvflare/app_opt/job_launcher/study_runtime.py
pcnudde: This parser is likewise called only when pod_template is present. The launcher-mode check intentionally rejects Kubernetes-only configuration under Docker or Slurm instead of silently ignoring it, so no code change is needed.

nvflare/app_opt/job_launcher/slurm/config.py
pcnudde: Fixed in 379f69215. REQUEUE_HOLD and SPECIAL_EXIT now use the existing pending-timeout lifecycle, with both states covered by the parameterized timeout regression test. REQUEUE_FED remains excluded because federation is explicitly outside the supported contract.

nvflare/app_opt/job_launcher/slurm/manager.py
pcnudde: Fixed in 379f69215. The built-in Apptainer bind now preserves /etc/resolv.conf as the source path so it is resolved on the compute node where apptainer exec runs, rather than embedding a runtime-parent-specific canonical symlink target. Shared workspace and study mounts retain their existing canonicalization. The new regression test simulates a parent-specific resolver target and verifies that it is not rendered into the batch script.

nvflare/app_opt/job_launcher/slurm/batch.py
pcnudde: --ntasks-per-node=1 tells Slurm to allocate one task slot on each requested node. Since --ntasks equals --nodes, this gives the job one task-sized CPU allocation per node, allowing --cpus-per-task to implement cpus_per_node.

nvflare/app_opt/job_launcher/slurm/scheduler_client.py
pcnudde: This is intentional under the adopted Docker/Kubernetes-style creation contract. A timeout is launch uncertainty and is reported as failure even if partial stdout happens to contain an ID; the launcher does not adopt or recover uncertain submissions. Artifacts are removed immediately so an allocation that has not started cannot source its secret. An allocation that already started before cleanup is an accepted, documented limitation bounded by its Slurm wall time. Preserving or acting on TimeoutExpired stdout would reintroduce the ambiguity-handling contract that was explicitly removed, so no code change is needed.

YuanTingHsieh (approved):

Issue #4858, opened 2026-07-02
Client AUTH_TOKEN is a replayable bearer credential exposed via multiple plaintext channels


YuanTingHsieh:
## Summary

The per-client authentication credential (`FLMetaKey.AUTH_TOKEN` + `AUTH_TOKEN_SIGNATURE`) is exposed in plaintext through several co-located channels and is a **replayable, non-expiring bearer credential**. Tightening any single leak (e.g. the config-file permissions in #4855) does not close the exposure, and the credential's design offers no replay resistance. This issue tracks the broader hardening.

Severity assessment: **medium**. Not remotely exploitable on its own (secure mode gates transport with mTLS), and on single-tenant hosts the marginal exposure is minimal. The risk is real in **multi-tenant / shared-filesystem / bind-mounted** deployments, where a local co-tenant can lift a valid credential.

## What the token is

The server mints a random UUID4 per client at registration (`nvflare/private/fed/server/client_manager.py:335`) and signs `client_name + token` with the **server SSL private key** (RSA-PSS). The pair rides in message headers and is verified on every message by a `channel=*/topic=*` incoming filter (`nvflare/private/fed/server/fed_server.py:1101-1107`).

### 1. No replay resistance
Verification (`nvflare/fuel/f3/cellnet/utils/identity_utils.py:149-155`) is a static RSA-PSS check over `client_name + token` against the server cert — **no nonce, no timestamp, no expiry, no TLS-connection binding**. The `(client_name, token, signature)` triple is a bearer credential valid until the client re-registers; anyone who obtains it and can reach the cell can replay it by setting the same headers (`authenticator.py:381`). Impact of a valid replay: abort/fail **any job by ID** via `REPORT_JOB_FAILURE` with no membership check (`fed_server.py:820-853`), spoofed heartbeats/quit, and attempted task-pull/submit impersonation (only weakly guarded by a header-based origin-FQCN check that is absent at relays, `relay.py:235-241`).

### 2. The token leaks through multiple plaintext channels
The identical value is simultaneously exposed via:
- **CJ command line** `-t <token>` — readable by any local user's `ps` (`nvflare/private/fed/client/client_executor.py:235`)
- **server INFO log** on every registration (`client_manager.py:183-186`; also `communicator.py:571`)
- **config file** `client_api_config.json` written world-readable on the `ScriptRunner(launch_external_process=True)` path (addressed by #4855)
[...]

PR #4932, opened 2026-07-20
Redact client auth token from log messages

pcnudde:
## Summary

- Remove the client auth token value from all log messages. The `(client_name, token, signature)` triple is a replayable bearer credential (#4858); logging it hands a working credential to anyone with log access (aggregators, support bundles).
- Redacted sites: registration (`client_manager.py` "Sent token"), disable-client token list (now a count), unknown-token warning, heartbeat debug (logs client name instead), client re-activation, relay auth success, unauthenticated-quit (`communicator.py`), and token-verification failure (`identity_utils.py`).
- Design-doc threat notes updated to record the redaction.

Addresses item 2 of #4858 (item 1 landed in #4929).

### DIFF
--- a/docs/design/job_process_credential_transport_design.md
+++ b/docs/design/job_process_credential_transport_design.md
@@ -56,7 +56,9 @@ design doc's step tables document the new step.
 
 ## Threat notes
 
-Protects against command-line inspection and log/telemetry capture. Does not protect against:
+Protects against command-line inspection and log/telemetry capture. As a ride-along, the
+token value is also redacted from server/client/relay log messages (registration,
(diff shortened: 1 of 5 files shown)


YuanTingHsieh (approved):

pcnudde:
Good catch on the `remove_client` success branch — that log call spans two lines, so the single-line grep used for the sweep missed it. Fixed in a follow-up together with two more multiline sites the same sweep pattern missed (`fed_server.py` dead-client removal, `fed_client_base.py` registration-success logging its own token+ssid); see the linked PR.

PR #4934, opened 2026-07-20
Convert Auto-FL skill scripts to SPDX license headers

pcnudde:
## Summary

- #4890 changed `ci/check_license_header.py` to require the compact SPDX header for files under `skills/` (they are loaded into LLM context, so the short header saves tokens), but the four `skills/nvflare-autofl/scripts/*.py` files from #4780 still carry the full Apache boilerplate.
- As a result `license-check` fails on every PR based on current main (e.g. #4933).
- Converts the four headers to the SPDX form; `python3 ci/check_license_header.py` passes repo-wide.


pcnudde:
created in error

PR #4935, opened 2026-07-20
Touch skills for NVSkills CI validation

chesterxgchen:
## Summary
- make a no-op inline YAML comment refresh across all NVFLARE SKILL.md entrypoints
- use this PR to exercise the NVSkills CI workflow after the license checker fix landed

## Validation
- python3 ci/check_license_header.py
- python3 -m pytest tests/unit_test/tool/agent_skill_checks/lints_test.py
- ./runtest.sh -s
- pre-push agent skill lint

The parsed skill metadata is unchanged; only YAML comments on the version lines changed.

PR #4931, opened 2026-07-20
Refresh skills for NVSkills CI bootstrap

chesterxgchen:
## Summary
- add no-op YAML comments to every NVFLARE SKILL.md entrypoint
- use this PR to route all current skills through the newly configured NVSkills CI process

## Validation
- python3 -m pytest tests/unit_test/tool/agent_skill_checks/lints_test.py
- ./runtest.sh -s
- pre-push agent skill lint

Note: python3 ci/check_license_header.py currently fails on existing main-branch Auto-FL script headers under skills/nvflare-autofl/scripts; this PR does not modify those files.

PR #4937, opened 2026-07-21
Remove NVSkills CI bootstrap comments from skill frontmatter

holgerroth:
## Summary

- Remove the one-time `# NVSkills CI bootstrap: no behavior change.` inline comments that #4931 added to the `version:` line of every skill's SKILL.md frontmatter.
- YAML parsers ignore inline comments, so the parsed frontmatter is byte-identical before and after; versions are intentionally unchanged.

## Why

The comments existed only to give every skill a diff so the PR would route the full catalog through the newly configured NVSkills CI. On #4931 the `/nvskills-ci` dispatch never reported back (no commit status, no signature commit), so the comments merged as permanent cruft without serving their purpose. Root cause of the silence: the NVSkills CI service rejects fork pull requests (`allow_fork_pull_requests: false` in `NVIDIA/nvskills-ci` `config/onboarded-repositories.json`), and rejections are only logged inside the internal service — the source PR gets no feedback. #4931 and #4935 are both fork PRs, so every `/nvskills-ci` attempt on them was silently rejected before validation ever ran.

Because this PR again touches all seven SKILL.md files, it satisfies the NVSkills CI watched-path gate — this PR is opened from a branch on NVIDIA/NVFlare directly, so a `/nvskills-ci` comment by a maintainer here should finally pass request validation and perform the intended full-catalog run against clean frontmatter. This PR therefore both removes the bootstrap hack and replaces #4931 as the bootstrap/test vehicle.

## Proposed upstream fix: `/nvskills-ci --all`

Catalog-wide validation should not require no-op file touches. A diff-independent scope flag would fix this permanently, and needs no changes in team repos — the trigger in `request-nvskills-ci.yml` uses `startsWith(comment.body, '/nvskills-ci')`, which already matches `/nvskills-ci --all`. Concretely:

1. **`NVIDIA/skills` `team-request.yml`** (reusable workflow): in the "Resolve request context" step, parse `--all` from the comment body into a `request_scope` value (`all` vs default `changed`); skip the `has_watched_change` paths gate when scope is `all`; add `request_scope` to the `workflow_dispatch` payload sent to `nvskills-ci`.
2. **`NVIDIA/nvskills-ci` `nvskills-ci.yml`**: accept a `request_scope` input, allowlist-validate it (`changed`/`all`) in `scripts/validate_request.py`, and — when scope is `all` — also skip that script's own server-side watched-files gate (it independently re-fetches the PR file list and raises `SkippedRequest` when no changes fall under watched paths), then expose the scope to the trigger step.
3. **`scripts/trigger_downstream.py`**: forward it as a GitLab pipeline variable in `build_variables()`, e.g. `NVSKILLS_VALIDATE_SCOPE` (default `changed`).
4. **Downstream GitLab pipeline**: when `NVSKILLS_VALIDATE_SCOPE=all`, enumerate every skill under the watched roots (`skills/`, `team-skills/`, `plugins/`, `rules/team-rules/`) at `SOURCE_HEAD_SHA` instead of deriving the skill set from the base..head diff, then attach signatures for the full catalog.

No new privilege surface: the existing maintain/admin requester check in `team-request.yml` already gates the command, and `--all` only widens validation coverage, not permissions.

## Validation

- `python3 -m pytest tests/unit_test/tool/agent_skill_checks/` — 134 passed
- `python3 ci/check_license_header.py` — all folders pass


holgerroth:
Closing in favor of #4940, which is already touching the skill frontmatter — folding the bootstrap-comment cleanup into that PR avoids yet another PR colliding on the same frontmatter lines (see the version-bump discussion in https://github.com/NVIDIA/NVFlare/pull/4940#discussion_r3626882071). Will reopen if a cleanup gap remains for the skills #4940 doesn't touch after it lands.

PR #4940, opened 2026-07-21
Refine NVSkills validation findings

chesterxgchen:
## Summary

- Harden `nvflare-autofl` simulator execution by replacing full host environment forwarding with a sanitized allowlist for simulator runs and capability probes.
- Keep Auto-FL behavior/instruction changes out of this PR; `skills/nvflare-autofl/SKILL.md` is touched only to remove the inline `version:` comment as part of the skill metadata cleanup.
- Keep optional `environment.simulator_env_passthrough` support for job-specific environment variable names; absent config defaults to `[]`.
- Expand `nvflare-diagnose-job` failure patterns for resource requests that exceed host capacity, server/site config-file validation errors, and infrastructure/deployment failures.
- Explicitly define the no-match fallback as `matched_pattern=UNKNOWN` and `recovery_category=UNKNOWN`.
- Align PyTorch conversion guidance so `FedAvgRecipe.key_metric` matches `FLModel.metrics`, with negated metrics for lower-is-better values.
- Keep all skill package versions at `0.1.0` until the next release version bump and remove inline version comments.

## Compatibility

- Existing Auto-FL configs remain valid when `environment.simulator_env_passthrough` is absent.
- Existing simulator jobs continue to receive the default runtime allowlist.
- Jobs that previously depended on custom, non-allowlisted env vars must now list those variable names explicitly in `environment.simulator_env_passthrough`.

## Testing

- `python -m pytest tests/unit_test/tool/agent_skill_checks/seed_skills_test.py tests/unit_test/tool/autofl_skill_job_importer_test.py tests/unit_test/tool/autofl_skill_runner_test.py -q` (`174 passed`)
- `./runtest.sh -s`

### DIFF
--- a/skills/nvflare-autofl/SKILL.md
+++ b/skills/nvflare-autofl/SKILL.md
@@ -2,7 +2,7 @@
 name: nvflare-autofl
 description: "Optimize an existing NVFLARE job.py through an agent-assisted Auto-FL campaign that preserves FLARE execution, policy, artifacts, and reproducibility."
 license: Apache-2.0
-version: "0.1.0" # NVSkills CI bootstrap: no behavior change.
+version: "0.1.0"
 compatibility: "Requires NVFLARE 2.8.0+, Python, and permission to run NVFLARE jobs in the selected environment."
 metadata:
   author: "NVIDIA FLARE Team <federatedlearning@nvidia.com>"
--- a/skills/nvflare-autofl/SKILL.md
+++ b/skills/nvflare-autofl/SKILL.md
@@ -2,7 +2,7 @@
 name: nvflare-autofl
 description: "Optimize an existing NVFLARE job.py through an agent-assisted Auto-FL campaign that preserves FLARE execution, policy, artifacts, and reproducibility."
 license: Apache-2.0
-version: "0.1.1"
+version: "0.1.0" # NVSkills CI bootstrap: no behavior change.
 compatibility: "Requires NVFLARE 2.8.0+, Python, and permission to run NVFLARE jobs in the selected environment."
 metadata:
   author: "NVIDIA FLARE Team <federatedlearning@nvidia.com>"
@@ -14,18 +14,22 @@ metadata:
 # NVFLARE Auto-FL
 
 ## Use When
-Use when optimizing an existing NVFLARE `job.py` for accuracy, AUC, loss, runtime, robustness, or another metric.
+Use this skill when the user asks to optimize an existing NVFLARE `job.py` for accuracy,
+AUC, loss, runtime, robustness, or another metric in simulation, POC, or production.
 
 ## Do Not Use When
-Do not use for converting non-FL training code, diagnosing jobs without an optimization goal, production setup,
-evaluation/statistics-only recipes, or generic tuning outside an NVFLARE job.
+Do not use for converting non-FL training code into NVFLARE, diagnosing failed jobs without an optimization goal,
+production deployment setup, evaluation/statistics-only recipes, or generic tuning outside an NVFLARE job.
 
 ## Workflow
-Use this skill to optimize an existing NVFLARE `job.py` without a new Auto-FL command tree. The user provides a job,
-objective, environment, and optional budget; NVFLARE provides deterministic import/execution, policy boundaries,
-artifacts, and contracts. The coding agent owns hypotheses, source edits, new algorithms, and candidate choice.
+Use this skill to optimize an existing NVFLARE `job.py` without asking the user to learn
+a new Auto-FL command tree. The user selects this skill, points to a job, and states the
+objective, environment, and optional budget. NVFLARE provides the deterministic campaign
+import, execution substrate, policy boundaries, artifacts, and machine-readable
+contracts. The coding agent owns hypotheses, source edits, new algorithm
+implementations, and candidate choice.
 
-Resolve [run_job_campaign.py](scripts/run_job_campaign.py) as `RUNNER`, then initialize:
+Resolve [run_job_campaign.py](scripts/run_job_campaign.py) relative to this `SKILL.md`, store its absolute path as `RUNNER`, and initialize the campaign:
 
 ```bash
 python "$RUNNER" initialize ./job.py [--metric <metric>] --mode <max|min> --env <sim|poc|prod> [--max-candidates <n>]
@@ -110,9 +114,6 @@ candidate comparability, or production submission before running candidates.
   import, validation, execution, metric extraction, plotting, and reporting.
   Do not search for alternate interpreters or install dependencies unless the
   user explicitly asks you to prepare the environment.
-- Simulator child processes receive only the default runtime environment
-  allowlist. For job-specific variables such as `DATASET_DIR`, add names, not
-  values, to `environment.simulator_env_passthrough`.
 - Treat generated `autofl.yaml`, task-local `mutation_schema.yaml`, and
   existing NVFLARE job/runtime configuration as authoritative; the default
   simulation flow needs no prose profiles, branch setup, or harness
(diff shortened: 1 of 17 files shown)


skills/nvflare-autofl/scripts/run_job_campaign.py
```diff
@@ -2064,15 +2088,17 @@ def probe_simulator_workspace_override_support(
...
-        )
-        payload = json.loads(completed.stdout.strip().splitlines()[-1])
+        with tempfile.TemporaryDirectory(prefix="nvflare-autofl-probe-") as probe_dir:
+            rc, stdout, _runtime = run(
+                [python, "-c", script],
+                cwd,
+                timeout,
+                Path(probe_dir) / "probe.log",
+                simulator_no_progress_timeout=0,
+            )
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in f9b58c49e. The probe runner now passes env=simulator_child_env(Path(probe_dir)), and the new regression test asserts probe subprocesses preserve allowlisted runtime variables while excluding credential-like environment variables.

chesterxgchen:
Addressed the diagnose-job QA finding in a926ac983. The failure-pattern catalog now states that recovery_category must be copied exactly from the matched row, rewords ROUND_TIMEOUT so timeout-limit changes are only a temporary mitigation, and keeps PARTIAL_LOG_VISIBILITY pinned to UNKNOWN until missing logs are collected. Added a seed-skill regression test for those catalog guarantees.

skills/nvflare-autofl/SKILL.md
```diff
@@ -14,22 +14,18 @@ metadata:
...
-objective, environment, and optional budget. NVFLARE provides the deterministic campaign
-import, execution substrate, policy boundaries, artifacts, and machine-readable
-contracts. The coding agent owns hypotheses, source edits, new algorithm
-implementations, and candidate choice.
+Use this skill to optimize an existing NVFLARE `job.py` without a new Auto-FL command tree. The user provides a job,
+objective, environment, and optional budget; NVFLARE provides deterministic import/execution, policy boundaries,
+artifacts, and contracts. The coding agent owns hypotheses, source edits, new algorithms, and candidate choice.
 
-Resolve [run_job_campaign.py](scripts/run_job_campaign.py) relative to this `SKILL.md`, store its absolute path as `RUNNER`, and initialize the campaign:
+Resolve [run_job_campaign.py](scripts/run_job_campaign.py) as `RUNNER`, then initialize:
```
holgerroth: Could we strip the cosmetic prose reflows in this file and keep only the two SKILL.md changes the env hardening actually needs — the `version` bump and the new `simulator_env_passthrough` bullet in the constraints section (~L110–113)?

The "Use When / Do Not Use When / Workflow" rewording collides head-on with two scheduled Auto-FL PRs that rewrite exactly these lines:

- #4925 rewrites the trigger sections as its core deliverable, with new trigger evals (`autofl_trigger_evals_test.py`) pinning that wording.
- #4927 reflows the same lines again.

Whichever PR lands second inherits a manual merge conflict for zero behavioral benefit, and if this condensation lands first it reverts the deliberate trigger wording #4925 introduces. Scoped down to the runner hardening + version bump + passthrough bullet, this PR merges cleanly alongside #4925/#4926/#4927 — the new bullet at ~L110 sits outside all of their hunks.

chesterxgchen: I can remove all auto-fl related changes, the changes are made because the skills formatter requires the 200 lines limits, auto-fl > 200, the agent trying to recorded to fit the 200 lines limt

holgerroth: Follow-up on the version-bump half of my comment above: do we actually need to keep bumping skill versions during development, before the skills have shipped in any release? It seems like unnecessary hassle:

- Nothing in-repo enforces it: the admission lints (`dev_tools/agent/skills/checks/lints.py`) validate `name`, `blast_radius`, `description`, `min_flare_version`, and `category` but never look at `version`, and `skills/README.md` documents no bump policy.
- No consumer has 0.1.0 yet, so 0.1.1 informs nobody.
- With five skill PRs in flight (#4925, #4926, #4927, #4928, this one), per-PR bumps mean every PR conflicts on that one line. Manual bumping is also error-prone — this PR bumps 3 of the 5 skills it touches (orient and nvflare-shared change content without a bump).

The one thing that would justify per-change bumps is the NVSkills signing chain: if signatures are keyed by `(skill, version)` or the chain refuses to re-sign changed content under an unchanged version, we have to bump (and then this PR should bump all five skills, not three). Could the NVSkills CI owners confirm?

If the chain doesn't care, I'd propose: freeze versions while the skills are unreleased, drop the bumps from this PR too, and set versions once, deliberately, in the release-cut PR.

holgerroth: Thanks — please don't remove *all* the auto-fl changes! The runner env hardening (allowlist + `simulator_env_passthrough` + probe rewrite in `run_job_campaign.py`/`job_importer.py`) is the valuable part of this PR and I'm ready to approve it. Only the SKILL.md trigger-section reflows are the problem.

The 200-line limit context helps — since this file sits at exactly 200 lines, adding the 3-line passthrough bullet does force freeing 3 lines somewhere. But it doesn't have to come from the "Use When / Workflow" sections that #4925/#4927 rewrite. Lines ~101–136 (the Requirements section) are untouched by every scheduled Auto-FL PR, and the two bullets right next to the insertion point compress by exactly 3 lines without losing meaning, e.g.:

```markdown
- If the environment provides `PYTHON`, `VIRTUAL_ENV`, or a venv on `PATH`, treat
  that prepared runtime as authoritative for all campaign steps; do not search for
  alternate interpreters or install dependencies unless the user explicitly asks.
- Treat generated `autofl.yaml`, task-local `mutation_schema.yaml`, and existing
  NVFLARE job/runtime configuration as authoritative; the default simulation flow
  needs no prose profiles, branch setup, or harness initialization.
```

(5+4 lines → 3+3 lines.)

So the suggested scope for this file: keep the new passthrough bullet, restore the "Use When / Do Not Use When / Workflow" text to what's on `main`, and free the 3 lines via the compression above. That stays ≤200 and merges cleanly with #4925/#4926/#4927. Version line per the open question above — happy to approve once that's pushed.

chesterxgchen: I am reverting the changes, get distracted for another PR (but end up close that PR after all)

chesterxgchen: you can add the changes you like in your PR.

chesterxgchen: I will keep the security pieces for auto-fl

holgerroth (approved):
Approving — thanks for the quick rescope. Verified against the previous head:

- The Auto-FL runner env hardening (`run_job_campaign.py`, `job_importer.py`, and both runner/importer test files) is byte-identical to what I reviewed: allowlist + `simulator_env_passthrough` + bounded probe, all solid and well-tested.
- `nvflare-autofl/SKILL.md` is now a comment-only change, so this merges cleanly with #4925/#4926/#4927, and the bootstrap-comment cleanup across all 7 skills fully absorbs #4937 (closed).
- The three new failure-pattern rows (`RESOURCE_EXCEEDS_HOST_CAPACITY`, `CONFIG_FILE_VALIDATION_ERROR`, `INFRASTRUCTURE_DEPLOYMENT_FAILURE`) and the explicit no-match → `UNKNOWN` fallback all have sensible category assignments and matching test pins.
- The `key_metric`/`FLModel.metrics` guidance checks out against the code: `FedAvgRecipe.key_metric` defaults to `"accuracy"`, and since the recipes deliberately don't expose `negate_key_metric`, the client-side `neg_loss` pattern is the right recipe-level advice.

Non-blocking note:

**Follow-up:** with the SKILL.md bullet dropped for the 200-line cap, `simulator_env_passthrough` is now undocumented outside the generated `autofl.yaml` and error messages. Worth a short note in an autofl `references/` file (no line cap pressure there) in a follow-up.

PR #4947, opened 2026-07-22
Add shared-file transport driver for cell communication

pcnudde:
## Summary

- Adds a new `file` scheme transport driver (`nvflare/fuel/f3/drivers/file_driver.py`) that exchanges SFM frames through a shared filesystem such as Lustre, for deployments where two cells share storage but have no network path between them (e.g. an HPC compute node running the client job cell and a gateway node running the client parent).
- Each connection is a directory with one append-only log per direction. The 16-byte SFM prefix provides framing; logs rotate at a size cap and are sealed with their final size so readers can distinguish a complete log from appends that are not yet visible. Consumed logs are deleted to bound disk usage.
- Liveness uses lease files judged by observed mtime changes (immune to cross-node clock skew) plus adaptive polling with backoff. SFM heartbeats ride the log, so existing PING/PONG and stall detection work unchanged. The closed marker is written only after all data, and readers drain pending frames on orderly close.
- Robustness: torn appends and rotation failures close the connection for a clean reconnect; garbage collection is gated on a driver ownership marker (never touches unrelated directories) and timed with local monotonic clocks; the active side removes its own connection dir when a listener is dead; reads are chunked with single-copy frame extraction.
- Security: no transport encryption; directory permissions are the trust boundary. Directories are created 0o770 and files 0o660 regardless of umask, overridable via `dir_mode`/`file_mode` resources. The driver reports `SUPPORT_SSL: False` so secure-required connectors are rejected.
- `parse_file_url` is exported as the single source of truth for the `file://0/<absolute-dir>` URL form; malformed URLs (non-'0' authority, relative paths, reserved characters in paths) are rejected.
- Slurm integration: the Slurm launcher preserves valid file parent URLs (query parameters intact) and keeps tcp rewriting unchanged; pyxis and apptainer launch plans bind-mount the listener directory read-write at the same absolute path, rejecting overlap with study mounts; `deploy prepare` preserves an existing `internal.scheme: file` comm config (validating absolute `root_dir` and clear connection security) instead of overwriting it with tcp; the Docker launcher passes file URLs through unchanged.
- Configuration: `internal: {scheme: "file", resources: {root_dir: ...}}` plus `backbone: {connect_generation: 1}` in the client comm_config routes job-cell traffic through the client parent. Tuning resources propagate to the child cell via the connect URL query; a cell can override them locally (except `max_log_size`) through a `file` section in its own comm_config.

Validated end-to-end on real Lustre with Slurm (bare and Pyxis) workers communicating with their client parent, aggregating with a cloud server. Local cross-process benchmark: large-payload throughput on par with tcp/grpc (~1.7 GB/s at 100MB payloads); small-message round-trip is poll-bound (~2 poll intervals). An idle connection costs roughly 1.4 client-side metadata syscalls/sec at defaults (max_poll_interval=2s), shrinking proportionally as max_poll_interval grows; the Docker launcher bind-mounts the listener directory into job containers.

### DIFF
--- a/nvflare/app_opt/job_launcher/docker_launcher.py
+++ b/nvflare/app_opt/job_launcher/docker_launcher.py
@@ -45,6 +45,8 @@ from nvflare.app_opt.job_launcher.study_runtime import (
     load_study_runtime_file,
     resolve_study_runtime,
 )
+from nvflare.fuel.f3.comm_error import CommError
+from nvflare.fuel.f3.drivers.file_driver import parse_file_url
 from nvflare.utils.job_launcher_utils import (
     get_client_job_args,
     get_credential_env,
@@ -549,12 +551,20 @@ class DockerJobLauncher(JobLauncherSpec):
         # Derive parent_url at runtime: site name (= container name on Docker DNS) + port
         # from the original PARENT_URL in job_args. This avoids baking parent_url into
         # resources.json at provision time.
+        file_parent_dir = None
         if JobProcessArgs.PARENT_URL in job_args:
             flag, original_url = job_args[JobProcessArgs.PARENT_URL]
             if str(original_url).startswith("file:"):
-                # Shared-file transport URLs are location-independent; pass through unchanged
-                # (the listener directory must be volume-mounted into the container)
-                pass
+                # Shared-file transport URLs are location-independent; pass through unchanged and
+                # bind-mount the listener directory at the same path inside the container
+                try:
+                    file_parent_dir = parse_file_url(str(original_url))
+                except CommError as e:
+                    raise ValueError(f"invalid shared-file parent URL {original_url!r}: {e}")
+                if file_parent_dir.startswith(self.WORKSPACE_MOUNT):
+                    raise ValueError(
+                        f"shared-file parent directory {file_parent_dir} overlaps the container workspace mount"
+                    )
             else:
                 port = original_url.rsplit(":", 1)[-1]
                 parent_url = f"tcp://{site_name}:{port}"
@@ -716,6 +726,11 @@ class DockerJobLauncher(JobLauncherSpec):
                         type="bind",
                         read_only=dataset_mount.read_only,
                     )
+                )
+            if file_parent_dir:
+                # The job cell must update leases and append its log, so read-write
+                mounts.append(
+                    docker.types.Mount(target=file_parent_dir, source=file_parent_dir, type="bind", read_only=False)
                 )
             if study_runtime is not None:
                 for secret_mount in study_runtime.secret_mounts:
--- a/nvflare/app_opt/job_launcher/docker_launcher.py
+++ b/nvflare/app_opt/job_launcher/docker_launcher.py
@@ -18,6 +18,7 @@ import re
 import threading
 import time
 from abc import abstractmethod
+from urllib.parse import urlsplit
 
 try:
     import docker.errors
@@ -46,6 +47,7 @@ from nvflare.app_opt.job_launcher.study_runtime import (
     resolve_study_runtime,
 )
 from nvflare.fuel.f3.comm_error import CommError
+from nvflare.fuel.f3.drivers.file_driver import SCHEME as SHARED_FILE_SCHEME
 from nvflare.fuel.f3.drivers.file_driver import parse_file_url
 from nvflare.utils.job_launcher_utils import (
     get_client_job_args,
@@ -554,7 +556,7 @@ class DockerJobLauncher(JobLauncherSpec):
         file_parent_dir = None
         if JobProcessArgs.PARENT_URL in job_args:
             flag, original_url = job_args[JobProcessArgs.PARENT_URL]
-            if str(original_url).startswith("file:"):
+            if urlsplit(str(original_url)).scheme == SHARED_FILE_SCHEME:
                 # Shared-file transport URLs are location-independent; pass through unchanged and
                 # bind-mount the listener directory at the same path inside the container
                 try:
--- a/nvflare/app_opt/job_launcher/docker_launcher.py
+++ b/nvflare/app_opt/job_launcher/docker_launcher.py
@@ -551,10 +551,15 @@ class DockerJobLauncher(JobLauncherSpec):
         # resources.json at provision time.
         if JobProcessArgs.PARENT_URL in job_args:
             flag, original_url = job_args[JobProcessArgs.PARENT_URL]
-            port = original_url.rsplit(":", 1)[-1]
-            parent_url = f"tcp://{site_name}:{port}"
-            job_args = dict(job_args)
-            job_args[JobProcessArgs.PARENT_URL] = (flag, parent_url)
+            if str(original_url).startswith("file:"):
+                # Shared-file transport URLs are location-independent; pass through unchanged
+                # (the listener directory must be volume-mounted into the container)
+                pass
+            else:
+                port = original_url.rsplit(":", 1)[-1]
+                parent_url = f"tcp://{site_name}:{port}"
+                job_args = dict(job_args)
+                job_args[JobProcessArgs.PARENT_URL] = (flag, parent_url)
 
         module_args = self.get_module_args(job_args)
         module_args_list = []
--- a/nvflare/app_opt/job_launcher/slurm/launcher.py
+++ b/nvflare/app_opt/job_launcher/slurm/launcher.py
@@ -55,6 +55,8 @@ from nvflare.app_opt.job_launcher.study_runtime import (
     resolve_study_runtime,
     study_runtime_file_path,
 )
+from nvflare.fuel.f3.comm_error import CommError
+from nvflare.fuel.f3.drivers.file_driver import parse_file_url
 from nvflare.utils.job_launcher_utils import (
     get_client_job_args,
     get_credential_env,
@@ -160,13 +162,26 @@ def _rewrite_parent_url(job_args: dict, parent_host: Optional[str], internal_por
     flag, raw_url = entry
     try:
         parsed = urlsplit(str(raw_url))
+    except ValueError as e:
+        raise SlurmLauncherError("malformed parent URL in JOB_PROCESS_ARGS") from e
+
+    if parsed.scheme == "file":
+        # Shared-file transport: the directory path is location-independent, so the URL
+        # (including its tuning query parameters) is preserved without any host/port rewrite
+        try:
+            parse_file_url(str(raw_url))
+        except CommError as e:
+            raise SlurmLauncherError(f"malformed shared-file parent URL in JOB_PROCESS_ARGS: {raw_url!r}") from e
+        return copied
+
+    try:
         port = parsed.port
         host = parsed.hostname
     except ValueError as e:
         raise SlurmLauncherError("malformed parent URL in JOB_PROCESS_ARGS") from e
     if parsed.scheme != "tcp" or port != internal_port or not host:
         raise SlurmLauncherError(
-            f"parent URL must use tcp and configured internal_port {internal_port}, got {raw_url!r}"
+            f"parent URL must use file or tcp with configured internal_port {internal_port}, got {raw_url!r}"
         )
     host = _resolve_parent_host(parent_host)
     rendered_host = host if host.startswith("[") and host.endswith("]") else f"[{host}]" if ":" in host else host
@@ -174,6 +189,17 @@ def _rewrite_parent_url(job_args: dict, parent_host: Optional[str], internal_por
     rewritten = urlunsplit(("tcp", netloc, parsed.path, parsed.query, parsed.fragment))
     copied[JobProcessArgs.PARENT_URL] = (flag, rewritten)
     return copied
+
+
+def _file_parent_mount(parent_url: str, workspace_path: str) -> Optional[BindMount]:
+    """Bind mount for a shared-file transport parent URL so the CJ container can reach the
+    listener directory at the same absolute path as the CP."""
+    if urlsplit(str(parent_url)).scheme != "file":
+        return None
+    listen_dir = parse_file_url(str(parent_url))
+    destination = _validate_mount_destination(listen_dir, "shared-file parent directory")
+    source = _validate_mount_source(listen_dir, workspace_path, "shared-file parent directory")
+    return BindMount(source, destination, "rw")
 
 
 class SlurmJobLauncher(JobLauncherSpec):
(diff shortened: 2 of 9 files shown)


nvflare/fuel/f3/drivers/file_driver.py
```diff
@@ -0,0 +1,756 @@
...
+                entries = sorted(os.listdir(conns_dir))
+            except OSError as ex:
+                raise CommError(CommError.ERROR, f"Cannot list {conns_dir}: {secure_format_exception(ex)}")
+
+            now = time.monotonic()
+            if now - last_housekeeping >= cfg.lease_interval:
+                last_housekeeping = now
+                _touch(lease, cfg.file_mode)
+                self._gc_conn_dirs(conns_dir, cfg, entries)
+
```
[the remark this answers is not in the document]
pcnudde: Fixed in 0180cd279: transient listdir errors now log and retry inside the accept loop; only a genuinely missing listener dir raises. Note the connector retry loop in conn_manager already restarted listen() on exception, so the failure was not permanent, but inline tolerance avoids the restart churn and error logs.

nvflare/fuel/f3/drivers/file_driver.py
```diff
@@ -0,0 +1,756 @@
...
+        conn = FileConnection(conn_dir, connector, cfg)
+        self.add_connection(conn)
+        try:
+            conn.read_loop(self.stop_event)
+        finally:
+            conn.close()
+            self.close_connection(conn)
+            if not conn.received_anything:
+                # Nothing ever arrived (e.g. the listener is dead); remove our own dir so
+                # reconnect attempts don't accumulate directories on the shared filesystem
```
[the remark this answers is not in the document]
pcnudde: Fixed in 0180cd279: read_loop now also exits when connector.stopped is set. In-tree, remove_connector always calls shutdown() on the per-connector driver instance so both events fire together, but the explicit check removes that coupling.

nvflare/fuel/f3/drivers/file_driver.py
```diff
@@ -0,0 +1,756 @@
...
+        os.utime(path, None)
+    except FileNotFoundError:
+        try:
+            open(path, "ab").close()
+            if mode is not None:
+                os.chmod(path, mode)
+        except OSError as ex:
+            log.debug(f"Cannot create {path}: {secure_format_exception(ex)}")
+    except OSError as ex:
+        log.debug(f"Cannot touch {path}: {secure_format_exception(ex)}")
```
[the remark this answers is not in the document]
pcnudde: Done in 0180cd279 (both _touch and _create_file).

nvflare/tool/deploy/slurm_deploy.py
```diff
@@ -201,6 +205,25 @@ def _patch_comm_config(kit_dir: Path, port: int) -> None:
...
+    resources = _internal_resources(comm_config)
+    root_dir = resources.get("root_dir")
+    if not isinstance(root_dir, str) or not Path(root_dir).is_absolute():
+        _fail(
+            "INVALID_KIT",
+            "file transport requires an absolute internal.resources.root_dir.",
+            "Fix the startup kit comm config.",
+        )
+    if resources.setdefault("connection_security", "clear") != "clear":
+        _fail(
```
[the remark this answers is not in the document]
pcnudde: Docstring updated in 0180cd279 to state that connection_security is defaulted to clear when absent.

nvflare/fuel/f3/drivers/file_driver.py
```diff
@@ -0,0 +1,764 @@
...
+                self.file = open(path, "rb")
+            self.file.seek(self.offset)
+            data = self.file.read(min(size - self.offset, READ_CHUNK))
+        except OSError:
+            # Not created yet (writer still opening the next log after rotation), or transient
+            # FS error; either way retry on the next poll and rely on lease timeout for liveness
+            return None
+        if data:
+            self.offset += len(data)
+        return data
```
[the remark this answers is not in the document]
pcnudde: Fixed in 714bed00a: the OSError handler now drops the open handle so the next poll reopens by path (offset/buffer preserved). Added a regression test simulating a dead descriptor. Good catch — with the peer lease still fresh this could wedge silently without even hitting the lease timeout.

nvflare/fuel/f3/drivers/file_driver.py
```diff
@@ -0,0 +1,766 @@
...
+        if data:
+            self.buffer.extend(data)
+            self.progressed = True
+
+        frames = []
+        pos = 0
+        buf = self.buffer
+        total = len(buf)
+        while total - pos >= FRAME_LEN_STRUCT.size:
+            length = FRAME_LEN_STRUCT.unpack_from(buf, pos)[0]
```
nvidianz: The length of frame can be retrieved like this. This relieves driver from internal knowledge of SFM framing,

        prefix = Prefix.from_bytes(prefix_buf)
        if prefix.length > MAX_FRAME_SIZE:

nvflare/fuel/f3/drivers/file_driver.py
```diff
@@ -0,0 +1,766 @@
...
+from nvflare.fuel.f3.drivers.driver_params import DriverCap, DriverParams
+from nvflare.fuel.f3.drivers.net_utils import MAX_FRAME_SIZE
+from nvflare.fuel.f3.sfm.prefix import PREFIX_LEN
+from nvflare.fuel.utils.argument_utils import str2bool
+from nvflare.fuel.utils.log_utils import get_obj_logger
+from nvflare.security.logging import secure_format_exception
+
+log = logging.getLogger(__name__)
+
+FRAME_LEN_STRUCT = struct.Struct(">I")
```
nvidianz: Prefix class defines this already.

nvflare/fuel/f3/drivers/file_driver.py
```diff
@@ -0,0 +1,766 @@
...
+]
+
+# Directory layout
+CONNS_DIR = "conns"
+LISTENER_PREFIX = "lst_"
+LISTENER_LEASE_FILE = "lease"
+OWNER_MARKER = ".nvf_file_transport"
+CLOSED_FILE = "closed"
+TMP_PREFIX = "."
+A2P = "a2p"  # active-to-passive log prefix
```
nvidianz: The convention is to use writer's name, active or passive, like in CellPipe.

nvflare/fuel/f3/drivers/file_driver.py
```diff
@@ -0,0 +1,766 @@
...
+        super().__init__()
+        self.stop_event = threading.Event()
+        self.dir_lock = threading.Lock()
+        self.handled_dirs = set()
+        self.done_dirs: Dict[str, float] = {}
+        self.logger = get_obj_logger(self)
+
+    @staticmethod
+    def supported_transports() -> List[str]:
+        return ["file"]
```
nvidianz: file is a common URL scheme already taken and it's also misleading here. We can use something like,

nas: Network-attached storage
shared_file
dfs: Distributed FS

Or lustre if it doesn't work on other shared fs like NFS.

nvflare/fuel/f3/drivers/file_driver.py
```diff
@@ -0,0 +1,766 @@
...
+
+    @staticmethod
+    def capabilities() -> Dict[str, Any]:
+        return {DriverCap.SEND_HEARTBEAT.value: True, DriverCap.SUPPORT_SSL.value: False}
```
nvidianz, line 553: We can turn this on if apps require SSL. Because all filesystem traffic is encrypted, this is equivalent to SSL. No MITM can read the traffic.

nvflare/fuel/f3/drivers/file_driver.py
```diff
@@ -0,0 +1,766 @@
...
+        if pos:
+            del buf[:pos]
+
+        # A seal is only ever written at or past max_log_size, so skip the probe until then.
+        # This requires max_log_size to be identical on both sides of a connection.
+        if self.offset >= self.cfg.max_log_size:
+            self._advance_if_sealed()
+        return frames
+
+    def _read_new_data(self) -> Optional[bytes]:
```
nvidianz, line 309: Is there a way to truncate the file at certain intervals? Otherwise, all data sent through the connection will accumulate on the filesystem indefinitely.

nvflare/app_opt/job_launcher/slurm/launcher.py
```diff
@@ -160,13 +162,26 @@ def _rewrite_parent_url(job_args: dict, parent_host: Optional[str], internal_por
     flag, raw_url = entry
     try:
         parsed = urlsplit(str(raw_url))
+    except ValueError as e:
```
nvidianz, line 166: The launncher's current configuration of host:port is TCP centric. Should change it to use an URL.

nvflare/app_opt/job_launcher/docker_launcher.py
```diff
@@ -549,12 +551,25 @@ def launch_job(self, job_meta: dict, fl_ctx: FLContext) -> JobHandleSpec:
...
         # from the original PARENT_URL in job_args. This avoids baking parent_url into
         # resources.json at provision time.
+        file_parent_dir = None
         if JobProcessArgs.PARENT_URL in job_args:
             flag, original_url = job_args[JobProcessArgs.PARENT_URL]
-            port = original_url.rsplit(":", 1)[-1]
-            parent_url = f"tcp://{site_name}:{port}"
-            job_args = dict(job_args)
-            job_args[JobProcessArgs.PARENT_URL] = (flag, parent_url)
+            if str(original_url).startswith("file:"):
```
nvidianz: The launcher shouldn't care about driver details. This check should be part of config validation.

nvflare/fuel/f3/drivers/file_driver.py
```diff
@@ -0,0 +1,766 @@
...
+        while total - pos >= FRAME_LEN_STRUCT.size:
+            length = FRAME_LEN_STRUCT.unpack_from(buf, pos)[0]
+            if length < PREFIX_LEN or length > MAX_FRAME_SIZE:
+                raise CommError(
+                    CommError.BAD_DATA,
+                    f"Corrupt frame length {length} in {_log_path(self.conn_dir, self.prefix, self.index)}",
+                )
+            if total - pos < length:
+                break
+            frames.append(bytes(memoryview(buf)[pos : pos + length]))
```
nvidianz, line 298: This wrap is expensive. Can we just use memoryview?

nvflare/fuel/f3/drivers/file_driver.py
pcnudde: Don't understand that. This is on a shared filesystem so anyone with the read credentials to the folder can read the communication. Why would filesystem traffic be encrypted?

nvflare/fuel/f3/drivers/file_driver.py
pcnudde: it is capped in segments of 1GB (configurable)

nvflare/fuel/f3/drivers/file_driver.py
pcnudde: Done in 19d8090b0: the reader now uses Prefix.from_bytes(...).length (waiting for PREFIX_LEN buffered bytes, which is the minimum frame size anyway), so the driver no longer carries private knowledge of the prefix layout.

nvflare/fuel/f3/drivers/file_driver.py
pcnudde: Removed in 19d8090b0 together with the Prefix.from_bytes change.

nvflare/fuel/f3/drivers/file_driver.py
pcnudde: Renamed in 19d8090b0: logs are now active.<n>.log / passive.<n>.log (and matching lease file names), named after the writer per the CellPipe convention.

nvflare/fuel/f3/drivers/file_driver.py
pcnudde: Renamed in 19d8090b0 to 'shared-file' — it works on any coherent shared filesystem so 'lustre' felt too narrow. Note it is hyphenated rather than shared_file: underscores are invalid in URL schemes (RFC 3986), and urllib.parse.urlsplit returns an empty scheme for shared_file://..., which would break URL parsing throughout. The scheme string is exported as file_driver.SCHEME and the launchers/deploy tool use that constant.

nvflare/app_opt/job_launcher/slurm/launcher.py
pcnudde: Agreed — the parent_host/internal_port config surface predates this PR and generalizing it to a URL is a launcher-config change worth doing on its own. Filed FLARE-3050 to track it.

nvflare/app_opt/job_launcher/docker_launcher.py
pcnudde: Partially addressed in 19d8090b0: the scheme check now uses urlsplit against the driver-exported SCHEME constant, and URL validity itself is delegated to parse_file_url (the driver-owned single source of truth). The remaining check guards the launcher's own container mount table (overlap with WORKSPACE_MOUNT), which deploy-time config validation cannot see — same placement as the Slurm launcher's study-mount overlap check.

nvflare/fuel/f3/drivers/file_driver.py
pcnudde: The bytes() copy is deliberate and is the only copy per frame (the memoryview exists precisely to avoid the bytearray-slice double copy). A zero-copy memoryview cannot be handed out: frames are queued to the frame-manager executor and outlive the read buffer, which is compacted (del buf[:pos]) right after extraction — that would either raise BufferError (exported buffers cannot resize) or deliver bytes that mutate under the consumer.

nvflare/fuel/f3/drivers/file_driver.py
pcnudde: One more consideration beyond the read-credentials point: in F3, SUPPORT_SSL also implies peer authentication — the mTLS CN-binding check in ConnManager.update_endpoint — which this driver cannot provide. Advertising it would let secure-required connectors run with unverified peer identity, a silent downgrade rather than an equivalence. Lustre/NFS wire encryption is site configuration, not a property of the transport. If needed later, app-layer signing/encryption could be added as an explicit feature.

nvflare/fuel/f3/drivers/file_driver.py
pcnudde: To complete the picture: consumed segments are deleted by the reader once their seal is verified, so steady-state disk usage per connection is bounded by roughly one in-flight segment per direction — total traffic never accumulates. Closed connection directories are also garbage-collected.

nvidianz (approved):

PR #4948, opened 2026-07-23
Add HuggingFace Trainer Client API

chesterxgchen:
## Summary

- Add `nvflare.client.hf.patch()` for HuggingFace `Trainer` / TRL `SFTTrainer` federated rounds.
- Support train, evaluate, and submit-model tasks with rank-0 Client API ownership, DDP coordination, checkpoint-backed Trainer state restore, budgeted round execution, metrics capture, PEFT adapter scope, and bf16-to-numpy-safe output conversion.
- Add a clean `examples/advanced/hf_client_api` Qwen SFT/PEFT example using the default `flare.patch(trainer)` API and a standard `prepare_data.py` data-prep step, while leaving the legacy `examples/advanced/llm_hf` example unchanged.
- Add the HF Client API design doc as implemented PR documentation and keep only one `3rdParty/packaging.LICENSE.txt` file for the optional `packaging` dependency.
- Add focused fake and real dependency contract tests, including transformers min/latest and TRL 0.18/latest CI jobs.

## Testing

- `python -m pytest tests/unit_test/app_opt/hf -q`
- `python -m pytest tests/unit_test/recipe/fedavg_recipe_test.py -q`
- `python -m py_compile examples/advanced/hf_client_api/client.py examples/advanced/hf_client_api/model.py examples/advanced/hf_client_api/prepare_data.py examples/advanced/hf_client_api/job.py`
- `python examples/advanced/hf_client_api/client.py --help`
- `python examples/advanced/hf_client_api/prepare_data.py --help`
- `python examples/advanced/hf_client_api/prepare_data.py --n_clients 2 --data_root /private/tmp/nvflare_hf_client_api_qwen_data_prep`
- `python examples/advanced/hf_client_api/job.py --help`
- `python examples/advanced/hf_client_api/job.py --export_config --job_dir /private/tmp/nvflare_hf_client_api_qwen_job_prepared --data_root /private/tmp/nvflare_hf_client_api_qwen_data_prep`
- Verified exported launcher config uses `python3 -u custom/client.py ...` rather than an absolute source path under `custom/`
- `./runtest.sh -s`
- Earlier in this PR: `./runtest.sh -l`

## Review Notes

- The new Qwen example is validated locally through compile/help/export checks. The README now follows the standard hello-world example layout. Full simulation requires downloading the Qwen checkpoint and installing HF/TRL example dependencies.
- FedAvg recipe `negate_key_metric` pass-through was split into separate PR #4953. This HF PR keeps the Qwen example independent by disabling recipe best-model selection with `key_metric=""`.
- The TRL contract test constructs real CPU `SFTTrainer` instances and applies `flare.patch()`, but does not run a real TRL train/send round. The train/send path is covered by real `transformers.Trainer` and fake-contract tests.
- `train_with_evaluation` hard-fail policy currently depends on existing Client API config. The new HF example explicitly calls `trainer.evaluate()` before `trainer.train()` each round; recipe-level pass-through can be added later if we want every recipe user to get a hard failure when pre-train eval is omitted.


nvflare/app_opt/hf/utils.py
```diff
@@ -0,0 +1,619 @@
...
+        return
+
+    torch = _import_torch()
+    state_dict = extract_params(trainer, params_scope)
+    try:
+        from safetensors.torch import save_file
+
+        save_file(state_dict, os.path.join(checkpoint_dir, "model.safetensors"))
+    except Exception:
+        torch.save(state_dict, os.path.join(checkpoint_dir, "pytorch_model.bin"))
```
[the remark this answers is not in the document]
chesterxgchen: Updated `write_params_to_checkpoint()` docstring to make the in-memory model mutation explicit. Also added a short comment on the legacy `torch.load` fallback path to document the trusted local-checkpoint assumption.

nvflare/app_opt/hf/api.py
```diff
@@ -0,0 +1,1154 @@
...
+# latest job passes; pin it back below the first failing release if HF internals drift.
+VERIFIED_TRANSFORMERS_VERSION_MAX_EXCLUSIVE = "6.0.0"
+PARAMS_FILE_EXCHANGE_MIN_BYTES_ENV_VAR = "NVFLARE_HF_PARAMS_FILE_EXCHANGE_MIN_BYTES"
+DEFAULT_PARAMS_FILE_EXCHANGE_MIN_BYTES = 64 * 1024 * 1024
+PARAMS_EXCHANGE_STRATEGY_ENV_VAR = "NVFLARE_HF_PARAMS_EXCHANGE_STRATEGY"
+PARAMS_EXCHANGE_STRATEGY_AUTO = "auto"
+PARAMS_EXCHANGE_STRATEGY_OBJECT = "object"
+PARAMS_EXCHANGE_STRATEGY_FILE = "file"
+
+_ACTIVE_STATE = None
```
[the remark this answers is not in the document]
chesterxgchen: Kept `_ACTIVE_STATE` as the process-level invariant for one patched Trainer per process, but removed the test-isolation hazard: `tests/unit_test/app_opt/hf/conftest.py` now restores `transformers`, `peft`, `nvflare.app_opt.hf*`, and `nvflare.client.hf` module-cache entries around every HF unit test. The full HF directory passes both in the base env (`75 passed, 2 skipped`) and the real transformers/TRL env (`80 passed`).

nvflare/app_opt/hf/api.py
```diff
@@ -0,0 +1,1166 @@
...
+            elif self.weight_override_strategy == STRATEGY_CHECKPOINT_INJECTION:
+                if self.rank == 0:
+                    utils.write_params_to_checkpoint(
+                        self.trainer,
+                        checkpoint_path,
+                        self.received_params,
+                        params_scope=self.params_scope,
+                        strict=self.load_state_dict_strict,
+                    )
+                _barrier()
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in `2792c1e91`. Rank-0-only checkpoint writes now go through a guard that broadcasts success/failure before entering the success barrier. If checkpoint injection or checkpoint state persistence fails on rank 0, all ranks raise a `RuntimeError` with the rank-0 error instead of non-zero ranks waiting in `dist.barrier()`. Added focused tests for rank 0 broadcasting the failure and non-zero ranks raising before the barrier for both checkpoint injection and state persistence paths.

setup.cfg
```diff
@@ -68,6 +68,7 @@ app_opt =
     %(TRACKING)s
     %(MONITORING)s
     %(K8S)s
+    packaging
```
YuanTingHsieh, line 71: what is this for?

nvflare/app_opt/hf/api.py
```diff
@@ -0,0 +1,1194 @@
...
+        if flare_api.is_submit_model():
+            return TASK_SUBMIT_MODEL
+        raise RuntimeError("Received an unsupported Client API task for the HuggingFace Trainer integration")
+
+    def _prepare_train_call(self, train_kwargs: dict):
+        self._capture_budget_if_needed()
+        self.train_start_global_step = max(
+            int(getattr(getattr(self.trainer, "state", None), "global_step", 0) or 0),
+            int(self.last_completed_global_step or 0),
+        )
```
holgerroth: With `restore_state=False`, this clamp looks like it makes `NUM_STEPS_CURRENT_ROUND` come out as `0` for every round after the first:

- When not resuming from a checkpoint, real HF `Trainer.train()` resets `self.state` to a fresh `TrainerState`, so round 1 runs steps `0 → N` again. But `max(trainer.state.global_step, last_completed_global_step)` (here, and again in `on_train_begin`) keeps `train_start_global_step` clamped at the previous round's end step `N`.
- `on_train_end` then computes `step_delta = max(0, N - N) = 0`, so the client sends an explicit `meta[NUM_STEPS_CURRENT_ROUND] = 0`. `base_fedavg.py` uses `meta.get(NUM_STEPS_CURRENT_ROUND, 1.0)` as the aggregation weight, so the explicit `0` bypasses the `1.0` default — and if all clients report `0`, the weighted average divides by zero (or effectively drops every update).
- `metric_step_offset` also stops advancing (`+= 0`), so streamed metrics for rounds >= 2 overwrite round 1's steps in TensorBoard.

The fake trainer in the contract tests keeps a persistent `state.global_step` that increments across rounds (it never resets), so the fake tests don't exercise this, and the real-transformers CI tests don't run a two-round `restore_state=False` sequence.

Suggestion: when `restore_state` is `False`, take the round's start step from the fresh `TrainerState` in `on_train_begin` (i.e., don't clamp with `last_completed_global_step`), and add a two-round `restore_state=False` real-transformers test asserting a positive `NUM_STEPS_CURRENT_ROUND` in the second round's sent `FLModel`.

YuanTingHsieh:
The persistent checkpoint provenance appears to support exactly one additional lifecycle: `restore_state=True` with `launch_once=False`.

With the default `launch_once=True`, the same `_HFTaskState` survives all rounds and already remembers `last_checkpoint_path` in memory. The example uses this default, so `fl_state.json` is written every round but provides no value on the normal path.

This also does not currently provide crash recovery. If the Trainer subprocess crashes, `LauncherExecutor` marks the task failed; NVFlare does not transparently launch another Trainer and reconnect it to the in-flight task. A new job would have a different job ID, so its persisted state is intentionally ignored.

Could we narrow Phase 1 to the existing Lightning contract?

- Keep standard HF checkpoints for optimizer/scheduler restoration.
- Keep the checkpoint path in memory.
- Support `restore_state=True` only for the single-process lifecycle.
- Remove persistent `fl_state.json` state and the crash/rejoin guarantee.
- Treat cross-process continuity under `launch_once=False` as a separate feature requiring an end-to-end launcher test and, ideally, consistent Lightning support.

`launch_once` is serialized into the external Client API configuration, so `patch()` can reject an explicitly configured `launch_once=False`; absence of that field can represent the inherently single-process in-process executor.

This retains the useful HF checkpoint mechanism without introducing an untested recovery contract.

YuanTingHsieh:
Two additional findings:

### Blocking: stateless rounds report zero training steps after round 0

At `nvflare/app_opt/hf/api.py:594`, `on_train_begin()` clamps the current HF `global_step` against the previously recorded `train_start_global_step`.

With `restore_state=False`, HF resets `state.global_step` to zero at the beginning of each fresh `train()` call. The `max()` retains the previous round's stale value instead. For example, after round 0 ends at step 2, round 1 trains from HF step 0 to 2 while the recorded baseline remains 2, so `on_train_end` calculates `step_delta = 2 - 2 = 0`.

This silently causes:

- `NUM_STEPS_CURRENT_ROUND=0`; FedAvg treats a non-positive value as weight 1.0, producing incorrect aggregation for heterogeneous clients.
- `metric_step_offset` does not advance, so later rounds reuse the same metric steps.

For stateless mode, `on_train_begin` should replace the baseline with the fresh HF `state.global_step` rather than clamping it against the previous value. Please add a real multi-round `restore_state=False` test asserting positive round weights and monotonically increasing metric steps.

### Documentation: weight-override escape hatch is undiscoverable

`NVFLARE_HF_WEIGHT_OVERRIDE_STRATEGY` is not documented, although the design says users can select the in-memory strategy after validating a newer Transformers version. Users outside the verified range otherwise have no discoverable way to avoid the automatic checkpoint-injection fallback.

Please either document `auto`, `in_memory`, and `checkpoint_injection` (including their risks), or keep this override private and remove the user-facing promise from the design.

holgerroth:
Suggestion on example placement: let's move `examples/advanced/hf_client_api` to `examples/hello-world/hello-huggingface`.

Rationale:

- This example is exactly the minimal entry point for the new API, and it already matches the `hello-<framework>` convention — the same file set as `hello-lightning` (`client.py`, `model.py`, `job.py`, data-prep script, `requirements.txt`), and the README already follows the hello-world layout. With the synthetic 4-row JSONL data and Qwen2.5-0.5B, it is hello-world scale.
- It slots naturally next to `hello-pt` / `hello-lightning` / `hello-tf` as the HuggingFace entry point.
- `examples/advanced/llm_hf` then stays the advanced HF tier (large-model streaming, quantization filters, multi-node SLURM). As a tracked follow-up, we should migrate `llm_hf` to `flare.patch()` so all HF examples use the HF Client API — that migration also doubles as at-scale validation of the new adapter (large payloads, filters, multi-GPU).

Mechanical changes this implies:

- `.github/workflows/premerge.yml`: the two TRL matrix legs reference `examples/advanced/hf_client_api/...` paths.
- Docs: in `client_api.rst` and `client_api_usage.rst`, list `hello-huggingface` in the per-framework example list (next to `hello-pt`/`hello-lightning`/`hello-tf`) and keep the `llm_hf` link under "Advanced Examples" until it is migrated, instead of replacing it; update the "Complete Example" section in `hf_client_api.rst` and the link in `llm_fine_tuning.rst`.
- `examples/hello-world/README.md`: add a "Hello HuggingFace" entry.
- Inside the example, rename the `hf_client_api_qwen` default workspace/data paths in `job.py`/`prepare_data.py`/README (e.g., `/tmp/nvflare/hello-huggingface/...`) and the directory references in the README.

nvflare/app_opt/hf/api.py
```diff
@@ -0,0 +1,1194 @@
...
+            )
+
+        grad_accum = max(1, int(getattr(getattr(self.trainer, "args"), "gradient_accumulation_steps", 1) or 1))
+        steps_per_epoch = math.ceil(dataloader_len / grad_accum)
+        return int(math.ceil(local_epochs * steps_per_epoch))
+
+    def _apply_cumulative_max_steps(self):
+        args = getattr(self.trainer, "args")
+        if not self.restore_state:
+            setattr(args, "max_steps", int(self.per_round_budget_steps))
```
pcnudde, line 1121: `restore_state=False` only rewrites `args.max_steps`; it does not reset/recreate the `Trainer` optimizer, scheduler, or round state. In a real two-round Trainer reproduction, optimizer steps continued 2→4, the scheduler object was reused, and the second result reported `NUM_STEPS_CURRENT_ROUND=0`. A separate two-site/two-round FedAvg run reached optimizer step 8 where a fresh second-round optimizer should have ended at 4. This makes the advertised stateless mode stateful and can also give FedAvg the wrong aggregation weight. Please reset the optimizer/scheduler and round bookkeeping for every stateless task, or reject/remove this mode until that lifecycle is supported.

nvflare/app_opt/hf/api.py
```diff
@@ -0,0 +1,1194 @@
...
+            )
+            params = utils.apply_server_key_prefix(params, self.server_key_prefix)
+            output_model = FLModel(
+                params=params,
+                metrics=self.pre_train_metrics,
+                current_round=self.current_round,
+                total_rounds=self.total_rounds,
+                meta=meta,
+            )
+            flare_api.send(output_model)
```
pcnudde: Durable round/checkpoint provenance is updated only after this blocking `send()` returns. In both real two-site/two-round PEFT and full-model runs, the server accepted and aggregated the final result, then the launcher terminated the trainer subprocess while it was still waiting for tensor-download completion here. The final checkpoint existed, but `_fl_exchange/fl_state.json` still named the prior round/checkpoint. The already-aggregated server model was correct, but restart and audit provenance was stale. Please make final-round persistence part of the lifecycle guaranteed before teardown—either persist the completed checkpoint/round before this wait with appropriate failure semantics, or add an explicit acknowledgement/teardown contract that guarantees the post-send state write runs.

tests/unit_test/app_opt/hf/real_transformers_contract_test.py
```diff
@@ -0,0 +1,202 @@
...
+def _has_model_weights(checkpoint_dir):
+    return (checkpoint_dir / "model.safetensors").exists() or (checkpoint_dir / "pytorch_model.bin").exists()
+
+
+def _remove_rng_state_files(checkpoint_dir):
+    # This test checks model-weight restore ordering, not RNG restore. Newer PyTorch
+    # defaults can reject Transformers' numpy-backed rng_state*.pth files under
+    # weights_only loading, so keep the checkpoint focused on the contract under test.
+    for path in checkpoint_dir.glob("rng_state*.pth"):
+        path.unlink()
```
pcnudde: This deletes a normal part of the resumable checkpoint and masks a supported-version failure. With Transformers 4.40.2 (inside the code's advertised `>=4.40.0` range) and torch 2.13.0, a real second `restore_state=True` round fails loading `rng_state.pth` because that Transformers version calls `torch.load()` without overriding torch's safe-loading default. The green test therefore does not validate complete optimizer/scheduler/RNG continuity. Please either add compatibility for this combination, raise/narrow the supported Transformers range, or explicitly test the incompatibility; the compatibility test should not remove checkpoint state required by the public resume contract.

pcnudde (commented):
Behavioral findings from pinned-head QA using real Transformers/TRL trainers, real two-process Gloo, and two-site/two-round federated runs.

setup.cfg
chesterxgchen: packaging here means the Python package packaging, mainly for version parsing/
  comparison.

  It is used for things like:

  - HF Client API: checks whether installed transformers is within the verified range
    before using the in-memory restore path. If packaging is missing, HF falls back to
    checkpoint injection instead of failing. See nvflare/app_opt/hf/api.py:223.

  - Flower integration: validates flwr>=1.26; this is a hard import. See nvflare/
    app_opt/flower/recipe.py:19.

  - XGBoost secure histogram path: compares XGBoost versions. See nvflare/app_opt/
    xgboost/histogram_based_v2/sec/client_handler.py:18.

  It is declared under optional extras like app_opt, app_opt_mac, and app_opt_cpu in
  setup.cfg:63, not as a core dependency.

  For the HF Client API specifically, packaging is not strictly required for basic
  functionality because the code catches ImportError; it just enables correct version
  gating for the optimized restore strategy. Since other app_opt integrations import
  it directly, keeping it in the app-opt extras is justified.

nvflare/app_opt/hf/api.py
chesterxgchen: Fixed in `3768ebfd7`. For `restore_state=False`, `_prepare_train_call()` and `on_train_begin()` now take the round start step from the current/fresh HF `TrainerState` instead of clamping against `last_completed_global_step`. The cumulative clamp remains only for `restore_state=True`. Added a fake Trainer regression that resets state between rounds and a real `transformers.Trainer` two-round regression asserting the second round reports a positive `NUM_STEPS_CURRENT_ROUND`.

nvflare/app_opt/hf/api.py
chesterxgchen: Fixed in b1a2f6fcf. on_train_end now records last_completed_round, checkpoint_path, and last_completed_global_step and persists them before the blocking send(), with result_upload_pending=True. After send returns it clears result_upload_pending and persists again. Same-round retry with a pending upload ignores that checkpoint to avoid overtraining from a possibly already-submitted round.

nvflare/app_opt/hf/api.py
chesterxgchen: Fixed in b1a2f6fcf. For restore_state=False, each stateless train task now clears Trainer optimizer, scheduler, _created_lr_scheduler, TrainerState, and TrainerControl before recomputing the round budget. I also added fake-contract and real-Trainer two-round assertions that the second round gets fresh optimizer/scheduler objects and reports a positive NUM_STEPS_CURRENT_ROUND.

tests/unit_test/app_opt/hf/real_transformers_contract_test.py
chesterxgchen: Fixed in b1a2f6fcf. The test no longer deletes rng_state*.pth. The HF API now allow-lists the NumPy RNG checkpoint globals before checkpoint resume, including user-supplied resume_from_checkpoint. The real Transformers checkpoint-restore contract calls the same compatibility helper before resuming.

chesterxgchen:
Addressed the remaining PR conversation comments in `69b087595`:

- Moved the minimal Qwen HF Client API example from `examples/advanced/hf_client_api` to `examples/hello-world/hello-huggingface` and updated CI paths, README paths, default `/tmp/nvflare/hello-huggingface/...` locations, quickstart/docs links, and the design doc.
- Added `docs/hello-world/hello-huggingface/index.rst` and listed Hello HuggingFace with the other hello-world framework examples.
- Documented `NVFLARE_HF_WEIGHT_OVERRIDE_STRATEGY` with `auto`, `in_memory`, and `checkpoint_injection` behavior in the HF Client API guide.
- Kept persisted `fl_state.json` provenance because it is needed for launcher teardown/relaunch audit and `launch_once=False` continuity, but narrowed the user-guide/design wording: it resumes from the last completed round for the same job and does not promise transparent recovery of an in-flight task already reported failed by the launcher.

Validation run locally:

- `python -m py_compile examples/hello-world/hello-huggingface/client.py examples/hello-world/hello-huggingface/model.py examples/hello-world/hello-huggingface/prepare_data.py examples/hello-world/hello-huggingface/job.py`
- `python examples/hello-world/hello-huggingface/client.py --help`
- `python examples/hello-world/hello-huggingface/prepare_data.py --help`
- `python examples/hello-world/hello-huggingface/job.py --help`
- `python -m pytest tests/unit_test/app_opt/hf -q`
- `./runtest.sh -s`

chesterxgchen:
Addressed the remaining top-level review comments in `45afeaf90` after rebasing onto current `upstream/main`:\n\n- DDP round-budget capture now runs on rank 0 and broadcasts the resolved budget/source to all ranks, so multi-rank runs do not derive budget independently from possibly different dataloader views. Added a nonzero-rank contract test.\n- Removed the redundant `TrainingArguments.max_steps` setter inside `_extend_cumulative_max_steps_if_needed()`; the caller remains the single place that applies the cumulative target.\n- The real two-round `restore_state=False` Transformers regression now also asserts streamed metric steps advance monotonically across rounds, in addition to positive round weights and fresh optimizer/scheduler state.\n- Fake Transformers contract tests now use a version inside the verified range, so the default fake-test path exercises the normal in-memory strategy; explicit tests still cover checkpoint-injection fallback.\n\nValidation run locally:\n\n- `python -m pytest tests/unit_test/app_opt/hf/task_dispatch_contract_test.py -q`\n- `uv run --active python -m pytest tests/unit_test/app_opt/hf/real_transformers_contract_test.py -q`\n- `python -m pytest tests/unit_test/app_opt/hf -q`\n- `./runtest.sh -s`

chesterxgchen:
Addressed the remaining top-level lifecycle concern in `7941d6abc`:\n\n- Narrowed `restore_state=True` to the single trainer-process lifecycle for Phase 1. `patch()` now rejects explicitly configured `launch_once=False` when `restore_state=True`; `restore_state=False` remains allowed for per-task launches.\n- Removed persisted HF FL checkpoint provenance: no `_fl_exchange/fl_state.json`, no startup state reload, no same-job relaunch/crash-rejoin contract. The checkpoint path is kept in memory only on the current `_HFTaskState`.\n- Removed the old state helper functions/tests and added contract tests for the new boundary plus a negative assertion that normal train does not write `fl_state.json`.\n- Updated the HF user guide and design doc to match the narrower Lightning-style contract.\n\nThe other top-level comments were already addressed in prior commits:\n\n- `restore_state=False` uses fresh HF `global_step` and has a real two-round Transformers regression asserting positive steps and monotonic metric steps.\n- `NVFLARE_HF_WEIGHT_OVERRIDE_STRATEGY` is documented with `auto`, `in_memory`, and `checkpoint_injection`.\n- The clean Qwen example is under `examples/hello-world/hello-huggingface`, CI/docs point to that path, and `examples/advanced/llm_hf` remains the advanced example.\n\nValidation run locally:\n\n- `python -m pytest tests/unit_test/app_opt/hf/api_contract_test.py tests/unit_test/app_opt/hf/task_dispatch_contract_test.py tests/unit_test/app_opt/hf/utils_test.py -q`\n- `python -m pytest tests/unit_test/app_opt/hf -q`\n- `uv run --active python -m pytest tests/unit_test/app_opt/hf/real_transformers_contract_test.py -q`\n- `./runtest.sh -s`

chesterxgchen:
Addressed the latest review findings in `d9d026879`:\n\n- DDP checkpoint-resume branch is now rank-0-owned: rank 0 broadcasts the selected restore checkpoint path, and non-zero ranks use that decision instead of branching on their local filesystem view. Added a rank-1 regression that joins checkpoint-injection collectives using the rank-0 path.\n- Updated distributed docs/design: distributed must initialize before `flare.patch(trainer)`; multi-node `restore_state=True` also requires shared `output_dir` checkpoint visibility.\n- Fixed bf16/fp16 NumPy conversion edge case: outgoing params now upcast half/bfloat16 whenever `exchange_format=NUMPY`, even if `server_expected_format` is not NumPy. Added a regression test.\n- Added an inline comment where `world_size` is snapshotted to document the distributed-before-patch contract.\n\nValidation run locally:\n\n- `python -m pytest tests/unit_test/app_opt/hf/task_dispatch_contract_test.py -q`\n- `python -m pytest tests/unit_test/app_opt/hf/utils_test.py tests/unit_test/app_opt/hf/utils_contract_test.py -q`\n- `python -m pytest tests/unit_test/app_opt/hf -q`\n- `uv run --active python -m pytest tests/unit_test/app_opt/hf/real_transformers_contract_test.py -q`\n- `./runtest.sh -s`

chesterxgchen:
Addressed the latest example/docs review findings in `276b8577f`:\n\n- Hello HuggingFace now gates BF16 on `torch.cuda.is_bf16_supported()`. CUDA GPUs without BF16 use FP16; CPU uses FP32. The SFTConfig precision flags and model `torch_dtype` use the same helper.\n- Fixed the user-guide snippet from `--local_epoch` to `--local_epochs`.\n- Fixed the README recipe name from `qwen-hf-client-api-*` to `hello-huggingface-*`.\n\nValidation run locally:\n\n- `python -m py_compile examples/hello-world/hello-huggingface/client.py examples/hello-world/hello-huggingface/model.py examples/hello-world/hello-huggingface/prepare_data.py examples/hello-world/hello-huggingface/job.py`\n- `python examples/hello-world/hello-huggingface/client.py --help`\n- precision helper smoke check for BF16 CUDA / FP16 CUDA / CPU\n- `python -m pytest tests/unit_test/app_opt/hf/utils_test.py tests/unit_test/app_opt/hf/utils_contract_test.py -q`\n- `./runtest.sh -s`

chesterxgchen:
Addressed the latest P2/P3 review findings in `49feed5c2`:\n\n- Large-param file exchange now converts NumPy arrays to CPU tensors before writing exchange files, avoiding unsafe/unsupported NumPy pickles under `torch.load(weights_only=True)`. Added a NumPy round-trip regression.\n- `patch()` now rejects any `WORLD_SIZE`/`LOCAL_WORLD_SIZE > 1` environment when `torch.distributed` is not initialized, including rank 0, preventing split-brain single-process behavior. Added rank-0 env regression.\n- Rank-zero train/eval/submit sends now use the same status-broadcast hardening as checkpoint injection. If rank 0 `send()` fails, non-zero ranks raise before marking the task complete. Added train rank-0/rank-1 and eval/submit rank-1 regressions.\n- Cleaned the stale design-doc bullet that still referenced state-file persistence tests.\n\nValidation run locally:\n\n- `python -m pytest tests/unit_test/app_opt/hf/api_contract_test.py tests/unit_test/app_opt/hf/task_dispatch_contract_test.py tests/unit_test/app_opt/hf/utils_test.py -q`\n- `python -m pytest tests/unit_test/app_opt/hf -q`\n- `uv run --active python -m pytest tests/unit_test/app_opt/hf/real_transformers_contract_test.py -q`\n- `./runtest.sh -s`

pcnudde:
### Remaining DDP failure-propagation gap: rank-zero result materialization is outside the guarded operation

The new status broadcast correctly handles a failure in `flare_api.send()`, but rank-zero result construction can still fail *before* rank 0 reaches that broadcast.

Current train-end flow:

```python
if self.rank == 0:
    params = utils.extract_params(...)
    params = utils.prepare_out_params(...)
    params = utils.apply_server_key_prefix(...)
    output_model = FLModel(...)

self._send_fl_model("train result send", output_model)
```

Nonzero ranks skip the extraction block and enter `_send_fl_model()`, where they wait for the rank-zero status broadcast. If rank 0 raises during extraction, conversion, prefixing, or `FLModel` construction, it never enters `_send_fl_model()`, so the other ranks remain blocked in the collective. The same gap exists in `_submit_model()` around checkpoint/current-model extraction and result construction.

I reproduced this against exact head `998274b76d7610ab9ff40aee61c9af9e89d390d7` with two real distributed processes:

- injected rank-zero `flare_api.send()` failure: both ranks received `HuggingFace distributed train result send failed on rank 0`, and rank 1 exited — **fixed**;
- injected rank-zero `utils.extract_params()` failure: after a five-second bound, both processes were still alive, only rank 0 had reported the injected exception, and rank 1 had produced no result — **still stranded**.

Gloo was only the locally available test transport; this is not the dismissed CPU/`cpu:0` issue. NCCL follows the same control flow: a nonzero rank enters a collective that rank 0 never reaches. `torchrun` may eventually terminate peers after noticing rank-zero death, but the integration does not promptly propagate the actual error.

Please make the protected rank-zero operation cover the complete result-materialization path — extraction, conversion, prefixing, `FLModel` construction, and send — and add a two-rank regression injecting an extraction failure and asserting that both ranks promptly receive the rank-zero error.

nvflare/app_opt/hf/api.py
```diff
@@ -0,0 +1,1292 @@
...
+def patch(
+    trainer,
+    restore_state=True,
+    load_state_dict_strict=True,
+    params_scope="auto",
+    server_key_prefix=None,
+    local_epochs=None,
+    local_steps=None,
+    stream_metrics=False,
+):
```
holgerroth, line 73: `patch()` is the primary public entry point of the new API, and the docs reference it via ``:func:`patch<nvflare.app_opt.hf.api.patch>``` (in `client_api.rst` and `client_api_usage.rst`), but it has no docstring — so the Sphinx API links render an empty description.

Please add a Google-style docstring like the Lightning counterpart (`nvflare/app_opt/lightning/api.py::patch`), covering:

- what patching does (wraps `trainer.train()`/`trainer.evaluate()` to drive FL train/evaluate/submit_model tasks, rank-0 Client API ownership),
- each argument (`restore_state`, `load_state_dict_strict`, `params_scope`, `server_key_prefix`, `local_epochs`/`local_steps` mutual exclusivity, `stream_metrics`),
- the return value, and
- the errors raised for unsupported configs (DeepSpeed/FSDP, `load_best_model_at_end`, `save_only_model` with `restore_state=True`, `launch_once=False` with `restore_state=True`, second trainer in the same process).

examples/hello-world/hello-huggingface/requirements.txt
```diff
@@ -0,0 +1,8 @@
+nvflare~=2.9.0rc1
```
holgerroth:
```suggestion
nvflare~=2.9.0rc
```

Please don't anchor the pin on `rc1` specifically — if the published pre-release ends up being rc2 (or rc1 is yanked), this exact pin gets stale. `nvflare~=2.9.0rc` matches the convention used by the other examples (`nvflare~=2.7.2rc`) and resolves to any 2.9.0 rc as well as later 2.9.x releases.

Please also update the README, which currently says "`requirements.txt` uses `nvflare~=2.9.0rc1`".

chesterxgchen:
Addressed the latest PR comments in 8e90a09bd: added the public patch() docstring, changed the hello-huggingface NVFLARE requirement/docs from nvflare~=2.9.0rc1 to nvflare~=2.9.0rc, and moved train/eval/submit result materialization plus send under the rank-zero DDP status-broadcast operation. This covers extraction, conversion, prefixing, FLModel construction, and send before peers can block. Added fake distributed regressions and a real two-process Gloo extraction-failure regression. Validation: python -m pytest tests/unit_test/app_opt/hf/task_dispatch_contract_test.py -q; python -m pytest tests/unit_test/app_opt/hf/distributed_contract_test.py -q; python -m pytest tests/unit_test/app_opt/hf -q; ./runtest.sh -s.

chesterxgchen:
Addressed the new DDP file-exchange hang comment in 7b0ac3a78: file-backed param exchange now catches read errors on every rank, gathers each rank status with torch.distributed.all_gather_object, and raises on all ranks if any rank failed before proceeding into training. The cleanup barrier still runs and rank 0 cleans up the staged file. Added fake-distributed coverage for rank 0 seeing a nonzero read failure and local nonzero read failure, plus a real two-process Gloo regression for a rank-1 file-read failure. Validation: python -m pytest tests/unit_test/app_opt/hf/task_dispatch_contract_test.py -q; python -m pytest tests/unit_test/app_opt/hf/distributed_contract_test.py -q; python -m pytest tests/unit_test/app_opt/hf -q; ./runtest.sh -s.

chesterxgchen:
Addressed the P1 DDP receive-path hang in b3bcd9200: rank 0 now wraps the full task receive/classification/server-prefix strip path, and for distributed runs broadcasts a task-dispatch failure payload when receive(), _read_task_kind(), or strip_server_key_prefix() fails before normal task broadcast. Nonzero ranks now receive the rank-0 error and raise instead of blocking in _broadcast_object. Single-process behavior is preserved: non-AgentClosed receive protocol errors still raise the original exception. Added fake-distributed rank-0/rank-1 regressions plus a real two-process Gloo rank-0 receive-failure regression. Validation: python -m pytest tests/unit_test/app_opt/hf/task_dispatch_contract_test.py -q; python -m pytest tests/unit_test/app_opt/hf/distributed_contract_test.py -q; python -m pytest tests/unit_test/app_opt/hf -q; ./runtest.sh -s.

chesterxgchen:
Addressed the P1 DDP is_running hang in 908abd700: distributed is_running() now catches unexpected rank-0 Client API errors, logs the failure, broadcasts a falsey value, and returns False on all ranks instead of leaving nonzero ranks blocked in broadcast_object_list. Single-process behavior is preserved by re-raising the original exception when world_size <= 1. Added fake-distributed rank-0/rank-1 regressions plus a real two-process Gloo rank-0 is_running-failure regression. Validation: python -m pytest tests/unit_test/app_opt/hf/task_dispatch_contract_test.py -q; python -m pytest tests/unit_test/app_opt/hf/distributed_contract_test.py -q; python -m pytest tests/unit_test/app_opt/hf -q; ./runtest.sh -s.

chesterxgchen:
Addressed the latest review findings in `f14fee664`:

- Distributed `is_running()` now broadcasts a structured rank-zero failure and raises the same actionable `RuntimeError` on every rank; unexpected Client API failures are no longer converted into normal loop termination.
- Positional `resume_from_checkpoint` arguments are recognized and preserved, preventing NVFlare from also injecting the keyword and causing duplicate-argument failures.
- Strict PEFT loading now rejects incomplete adapter payloads before calling `set_peft_model_state_dict()`.
- Main-branch Hello HuggingFace installation docs now install NVFlare from the repository and install the remaining dependencies explicitly; `requirements.txt` remains pinned to the first compatible unreleased version for post-release use.
- The design now states that all multi-node ranks must share `output_dir` and that missing staged files fail loudly.
- The real TRL contract now executes a complete CPU train/send round for both full-model and PEFT `SFTTrainer` paths.

Validation:

- `python -m pytest tests/unit_test/app_opt/hf/task_dispatch_contract_test.py tests/unit_test/app_opt/hf/utils_test.py -q` (63 passed)
- `python -m pytest tests/unit_test/app_opt/hf/distributed_contract_test.py -q` (6 passed)
- `python -m pytest tests/unit_test/app_opt/hf -q` (102 passed, 2 optional-dependency skips)
- `/Users/chesterc/nvflare-env13/bin/python -m pytest tests/unit_test/app_opt/hf/real_transformers_contract_test.py tests/unit_test/app_opt/hf/real_trl_contract_test.py -q` (6 passed)
- `./runtest.sh -s` (passed)

examples/hello-world/hello-huggingface/job.py
holgerroth, line 1: Now that this example lives in hello-world, `job.py` should match the hello-world shape: it is currently 166 lines with 15 CLI arguments, versus 84 lines / 6 args for `hello-pt` and 64 lines for `hello-lightning`. Most of the extra weight is avoidable:

- **Per-site config machinery** (`join_args`/`shlex.quote`, the per-site `train_args` loop, `set_per_site_config`, `SimEnv(clients=..., num_threads=...)`): this exists only because each site gets its own `--train_data`/`--eval_data`/`--output_dir`. Suggest passing one uniform `--data_root` in `train_args` and letting `client.py` resolve `{data_root}/{site_name}/` via `flare.get_site_name()` (it can call `flare.init()` before `flare.patch()`; `patch()` already tolerates a pre-initialized context). That deletes ~50 lines and lets `SimEnv(num_clients=...)` be used like the other hello examples. Per-site script args have precedent in `llm_hf`, but that is the advanced tier.
- **Hyperparameter pass-through args** (`max_length`, `learning_rate`, `gradient_accumulation_steps`, `per_device_train_batch_size`, `lora_r/alpha/dropout`): `client.py` already defines identical defaults; forwarding them through `job.py` doubles the surface without hello-world benefit. Drop them here.
- **`os.chdir(SCRIPT_DIR)` try/finally**: the other hello examples assume you run from the example directory, per the README.
- **`--export_config` + unconditional `recipe.export()`**: no other hello-world `job.py` exports, and CI only runs `job.py --help`, so the flag is not load-bearing.
- **`validate_site_data`**: a two-line existence check with "run `python prepare_data.py` first" gives the same UX as the 20-line quoted-path error assembly.

Debatable but worth considering: keep only the default PEFT mode here (removing `--train_mode`, `model_config()` branching, and half of `model.py`) and let full-model SFT be a README note — or something the migrated `llm_hf` demonstrates.

Please keep `server_expected_format=ExchangeFormat.PYTORCH`, `key_metric=""`, and `enable_tensor_disk_offload=True` — those are the genuinely LLM-specific settings and are good teaching content — but give each a one-line comment explaining why (bf16 must not round-trip through NumPy, model selection disabled, large tensor payloads), in the style of `hello-pt`'s inline comments.

With these changes `job.py` should land around 60–70 lines and read like `hello-pt`: parse a few args → `FedAvgRecipe(...)` → `SimEnv` → `execute`.

chesterxgchen: Simplified in `d8dde4fc2`.

- Reduced `job.py` from 166 to 74 lines and from 15 arguments to 4 (`n_clients`, `num_rounds`, `model_name_or_path`, `data_root`).
- Removed per-site config mutation, quoting helpers, working-directory switching, export mode, workspace/job-dir controls, and duplicated training/LoRA hyperparameter pass-through.
- `job.py` now passes one `data_root`; `client.py` initializes the Client API, gets `site_name`, and resolves `<data_root>/<site_name>/{train.jsonl,valid.jsonl}` plus a site-local output directory.
- Made the hello-world example PEFT-only and removed the full-model server wrapper. Full-model and multi-node variants remain in `advanced/llm_hf`.
- Kept `server_expected_format=PYTORCH`, `key_metric=""`, and tensor disk offload with one-line rationale comments.
- Updated the README, rendered guide, and design document to match the reduced workflow.

Validation: example compile/help checks passed; generated two-site data and recipe-construction smoke passed; full HF suite passed; `./runtest.sh -s` passed.

holgerroth (dismissed):
Nice support for HF!

YuanTingHsieh:
I took another pass focused on what is actually blocking versus what is architectural simplification.

## Blocking correctness issue

`_reset_stateless_trainer_task_state()` unconditionally sets `trainer.optimizer` and `trainer.lr_scheduler` to `None` for `restore_state=False`, including before the first FL train round. Hugging Face supports constructing `Trainer` with prebuilt instances via `optimizers=(optimizer, scheduler)`. The current reset silently discards that user configuration and lets Trainer create its default optimizer/scheduler instead, changing training semantics from round 1.

Please either:

- preserve/recreate the supplied optimizer and scheduler in their initial state each round, or
- reject `restore_state=False` when the Trainer was constructed with prebuilt optimizer/scheduler instances.

A real regression should construct Trainer with a distinctive optimizer such as SGD plus a custom scheduler and verify that the configured behavior is not silently replaced. HF documents prebuilt optimizer/scheduler support here: https://huggingface.co/docs/transformers/main/optimizers

## Non-blocking simplification concerns

- `_params_to_exchange_tensors()`, `write/read/cleanup_params_exchange_file()`, the two exchange environment variables, and several barriers/error paths exist only for the shared-file parameter transport. Since Phase 1 supports replicated DDP rather than FSDP/DeepSpeed, consider broadcasting only task metadata as an object and synchronizing the selected model tensors/buffers directly with `dist.broadcast()`. This removes the shared-`output_dir` requirement for parameter delivery. PyTorch also recommends tensor collectives rather than object collectives for tensor data: https://docs.pytorch.org/docs/stable/distributed.html
- If file staging is retained, `cleanup_params_exchange_file()` is necessary; without file staging, it and `_params_to_exchange_tensors()` can both disappear. I would prefer file staging to be justified by a measured workload and kept as an explicit optimization rather than becoming the automatic path above 64 MiB.
- The rank-aware `nvflare.client.hf.is_running()` behavior is necessary for DDP, but the separate `_running.py` forwarding module is not; it can be collapsed into the public namespace. The `pending`/`aborted` exceptions are reasonable fail-fast guards because the core Client API's `is_running()` prefetches and caches a task. A short code comment explaining that would make the otherwise surprising behavior much clearer.

So my merge position is: fix the prebuilt optimizer/scheduler behavior; the transport/module cleanup can be follow-up work if we intentionally accept the present complexity and shared-storage contract.

chesterxgchen:
Addressed this review in `2ae0d731b`:

- `patch(..., restore_state=False)` now rejects a Trainer that already has a prebuilt optimizer or LR scheduler. The rejection happens before Client API initialization or Trainer mutation and tells users to use `restore_state=True` or let Trainer create these objects.
- Added fake-contract coverage for optimizer-only and scheduler-only cases.
- Added a real Transformers regression using SGD (`momentum=0.75`) plus a custom `LambdaLR`; it verifies the patch is rejected and the exact configured optimizer/scheduler instances remain attached.
- Changed distributed parameter delivery to in-memory object broadcast by default. Shared-file staging is now explicit via `NVFLARE_HF_PARAMS_EXCHANGE_STRATEGY=auto` (threshold-based) or `file` (always). Tests cover all three policies and docs identify the shared-storage requirement only for the opt-in file modes.
- Removed `nvflare/client/hf/_running.py`; the public namespace now aliases the rank-aware implementation directly.
- Added the requested comment explaining that the pending/aborted guards prevent a second core `is_running()` task prefetch while the cached task is unfinished.

I retained file staging as an explicit optimization rather than replacing it with direct tensor collectives in this PR. A tensor-collective implementation changes dtype/device/key metadata synchronization and should be evaluated separately against large-model workloads.

Validation:

- `python -m pytest tests/unit_test/app_opt/hf/api_contract_test.py tests/unit_test/app_opt/hf/client_hf_contract_test.py tests/unit_test/app_opt/hf/task_dispatch_contract_test.py -q` (71 passed)
- `/Users/chesterc/nvflare-env13/bin/python -m pytest tests/unit_test/app_opt/hf/real_transformers_contract_test.py -q` (5 passed)
- `python -m pytest tests/unit_test/app_opt/hf -q` (105 passed, 2 skipped)
- `./runtest.sh -s` (passed)

chesterxgchen:
Addressed the generated training-argument quoting finding in `b76a289c9`. `job.py` now applies `shlex.quote()` to both `model_name_or_path` and the resolved `data_root` before constructing `train_args`, so `SubprocessLauncher` preserves local paths containing whitespace as single arguments. Validation: whitespace-path argument-tokenization smoke passed; `python -m py_compile examples/hello-world/hello-huggingface/job.py` passed; `./runtest.sh -s` passed.

chesterxgchen:
Addressed the two remaining quality observations in `52dcd2714`:

- Round-result metric extraction now uses the same finite-scalar normalization as streamed metrics. Tensor-like scalars are converted to Python floats; NaN, infinity, strings, and non-scalar tensors are omitted.
- Added a focused regression covering numeric, tensor-like, non-finite, and non-scalar metric values.
- Added an explanatory comment to the post-checkpoint-injection barrier: it ensures all ranks observe the completed checkpoint update before Hugging Face resumes from it.

Validation: `python -m pytest tests/unit_test/app_opt/hf/api_contract_test.py tests/unit_test/app_opt/hf/callbacks_contract_test.py -q` (26 passed); `python -m pytest tests/unit_test/app_opt/hf -q` (106 passed, 2 skipped); `./runtest.sh -s` (passed).

YuanTingHsieh (approved):

PR #4955, opened 2026-07-24
Fix agent inspection and NVFLARE skill routing

chesterxgchen:
## Summary

This PR addresses FLARE-3051 and the conversion-skill failures found while
validating the fix:

1. Scope Agent Inspector evidence to the authoritative project instead of the
   entire recursively scanned directory.
2. Improve `nvflare-diagnose-job` selection without attracting healthy
   NVFLARE lifecycle requests.
3. Make PyTorch-family recipe construction capability-based, including tensor
   transport, server disk offload, decomposer registration, and process mode.
4. Add narrow dependency-install and conversion-preflight retry guardrails.
5. Preserve Lightning validation metrics through custom aggregation so server
   metric artifacts and best-model selection work on the first run.
6. Reuse FedStats recipe site assignments in `SimEnv` so simulation does not
   specify the same clients twice.

The changes stay at the inspector, skill, recipe, and generated-template
boundaries. They do not add PyTorch- or Lightning-specific behavior to
framework-neutral `app_common` runtime code.

### Problem

Inspector previously stored recursively discovered `job.py` files and NVFLARE
signals in directory-wide state. A historical or fixture source job such as
`tests/fixture/job.py` could therefore promote an unrelated root PyTorch
repository to:

- conversion state: `flare_job`
- target type: `flare_job_source`
- recommended skill: `nvflare-autofl`

The same false authority could survive when the fixture imported production
code, because import connectivity merged the fixture and production files.

### Design

Inspector now preserves provenance for `job.py`, `SimEnv`, NVFLARE
imports/calls, integration signals, and export support. It builds one cached
directed local-import graph and derives authority in three layers:

1. **Local import graph:** resolve local modules once and cache reachable source
   dependencies.
2. **Project components:** use entry points and framework-bearing files as
   anchors, group connected anchors, and select components nearest the
   inspected root.
3. **Source-job candidates:** correlate each `job.py` or `SimEnv` candidate only
   with NVFLARE evidence belonging to its authoritative project component.

This allows a legitimate source job to delegate construction through
`job.py -> job_utils.py -> nvflare` without accepting unrelated NVFLARE evidence
elsewhere in the monorepo.

### Authority and ambiguity rules

- A root-level source job or an active nested source job remains authoritative.
- A deeper project becomes authoritative when it is reached from an active root
  anchor.
- Equally near independent projects produce
  `conversion_state: ambiguous` / `target_type: mixed_workspace` and route only
  to `nvflare-orient`.
- Entry points under test, fixture, archive, and vendored trees cannot suppress
  an independent active framework project.
- Secondary-tree FLARE evidence remains visible but is non-authoritative at
  repository scope, even when the fixture imports production code and joins its
  import component.
- Inspecting that secondary subtree directly remains a positive control because
  its entry point is root-relative in the requested scope.

When a secondary entry point participates in an equal-depth tie, Inspector
intentionally keeps all tied groups. This biases an uncertain workspace toward
`mixed_workspace` rather than risking a false-positive FLARE classification.

### Result

For the reported root PyTorch repository with a nested fixture job, Inspector
now returns:

- framework: `pytorch`
- conversion state: `not_converted`
- target type: `training_repository`
- recommended skill: `nvflare-convert-pytorch`

Positive controls with a root `job.py`, an authoritative nested
`jobs/fedavg/job.py`, or a root launcher importing a nested job still return
`flare_job`, `flare_job_source`, and `nvflare-autofl`. Export guidance for an
authoritative nested job uses its relative path, for example
`python jobs/fedavg/job.py --export ...`.

### Problem

The original `nvflare-diagnose-job` description under-triggered weaker models,
while stronger imperative wording attracted NVFLARE near-misses such as job
creation, code conversion, healthy monitoring, and successful-result download.

### Implementation

- Keep frontmatter concise and positive because it is the selector-visible
  trigger surface.
- Define the reported failure-signal set once in frontmatter.
- Use a post-load proceed/stop check in the body for defense-in-depth.
- Explicitly exclude healthy lifecycle and successful-result requests.
- Restrict terminal-job artifact download to bounded failure-evidence
  collection.
- Add near-miss evals for PyTorch conversion, healthy monitoring, job creation,
  and completed-result download.

The post-load check runs only after the `Skill` tool activates; it is not
counted as selector precision.

### Historical candidate comparison

The initial candidate sweep ran outside the sandbox with Claude Haiku 4.5 and
Claude Opus 4.8. It used four diagnosis prompts, four NVFLARE near-misses, three
attempts per prompt and model, and prompts that did not name the skill.

| Wording | Haiku positive | Haiku false trigger | Opus positive | Opus false trigger |
| --- | ---: | ---: | ---: | ---: |
| Original | 10/12 | 0/12 | 12/12 | 0/12 |
| Long imperative frontmatter | 8/12 | 5/12 | 12/12 | 1/12 |
| Short imperative plus exclusions | 10/12 | 2/12 | 10/12 | 2/12 |
| Concise positive frontmatter at `e86a009d2` | 12/12 | 1/12 | 12/12 | 0/12 |

This is a historical comparison of candidate wording. Review cleanup later
changed the exact selected text in `bafb0f965`.

### Exact-current revalidation

The exact skill at `7626a342a` was revalidated outside the sandbox on
2026-07-25:

- `SKILL.md` SHA-256:
  `54b441074634bfa63179df6204c3f907fb12b5f7764ccf26f1c5a7a0ddea85f8`
- Claude Code 2.1.220 with `--tools Skill`
- the same eight prompts and three attempts per prompt/model
- all 48 initialization events reported exactly `["Skill"]` as the tool set and
  the requested canonical model

| Current exact wording | Haiku positive | Haiku false trigger | Opus positive | Opus false trigger |
| --- | ---: | ---: | ---: | ---: |
| `7626a342a` | 12/12 | 0/12 | 12/12 | 0/12 |

Installed skill metadata remained visible, making this a competitive selection
environment. Some near-misses selected unrelated installed skills, but none
selected `nvflare-diagnose-job`. The historical Haiku 1/12 versus current 0/12
is sampling variation, not a claim that the underlying false-trigger
probability is exactly zero.

### Capability-based policy

A PyTorch conversion evaluation generated a tensor-native FedAvg job whose
generic in-process executor could not deserialize DOT 6 (`TENSOR_DOWNLOAD`).
The fix belongs at the generated job/recipe boundary, not in
framework-neutral runtime code.

Both plain PyTorch and Lightning now consume one canonical PyTorch-family
construction reference. After `recipe show`, the skill independently applies
only capabilities exposed by the selected recipe:

- pass `server_expected_format=ExchangeFormat.PYTORCH` when exposed;
- enable `enable_tensor_disk_offload=True` only when tensor-native transport is
  selected and the recipe exposes the optimization;
- register `TensorDecomposer` after per-site configuration whenever
  tensor-native exchange is selected;
- select `params_transfer_type` only when exposed;
- customize checkpoint naming with `best_model_filename` only, without the
  deprecated `save_filename` alias;
- configure a source-derived `key_metric` only when the selected execution path
  delivers that exact metric to server-side `FLModel.metrics`.

Cyclic and FedEval expose tensor format but not disk offload; Swarm exposes
neither parameter. Tests inspect the real recipe catalog and construct a real
Cyclic recipe without passing an unsupported offload keyword.

### Disk offload is a server optimization

Tensor disk offload is intentionally separate from the model-exchange contract.
It downloads incoming streamed tensors to temporary server files and
materializes them lazily during aggregation, reducing peak server memory and
OOM risk. It is not a client payload format and is effective only with
tensor-native transport.

Generated clients send PyTorch tensors rather than relying on NumPy conversion.
They do not register FOBS handlers themselves, and no framework-specific
registration was added to `app_common`.

### Execution mode

`launch_external_process=True` is selected from process-spawning or distributed
evidence such as DDP, `torchrun`, or `torch.distributed`, not merely from GPU
count. Single-process `torch.nn.DataParallel` remains in-process even when it
uses multiple GPUs.

### Dependency installation

A dependency prewarm failed with `No space left on device`, after which the
conversion attempted multiple installers, package-level retries, cache
deletion, uninstall, and direct `site-packages` repair.

The shared conversion contract now requires:

- load dependency guidance before Python commands import user, framework,
  product, or declared dependency modules;
- create one combined canonical install command containing every applicable
  requirements/constraints input and NVFLARE when it is missing;
- treat multiple planned `-r`/`-c` inputs in that command as one install, not as
  retries;
- redact credentials in the reported command and product error;
- on a nonzero install exit, stop validation for the current conversion run and
  preserve generated source only as an unvalidated draft;
- do not switch installer, index, backend, version, or package granularity, and
  do not purge caches, uninstall packages, or mutate `site-packages` directly.

A later user-directed run may retry after the environmental blocker is
resolved. Successful-install behavior and dependency selection are unchanged.

### Conversion preflight

A conversion preflight recovered from three avoidable generated-code failures:

1. conflicting `best_model_filename` and deprecated `save_filename`;
2. in-place shuffle of a read-only Pandas/NumPy index view;
3. hand-guessed exact partition sizes instead of invariant-based validation.

The skills now:

- use `best_model_filename` without the deprecated alias;
- shuffle writable positional index copies and select rows with
  `DataFrame.iloc`;
- validate partition completeness, non-overlap, determinism, and algorithm
  guarantees;
- assert exact site sizes only when supplied by the user/source or calculated
  programmatically.

These rules prevent known first-pass errors without prohibiting a scoped code
fix and validation rerun for a genuine generated-code defect.

### Problem

A generated Lightning job completed all FL rounds but did not create
`metrics_summary.json`. The generated client called
`trainer.validate(...)` and discarded its result. With
`train_with_evaluation=False`, the generic executor returned parameters and step
metadata without `FLModel.metrics`; the custom aggregator then produced a
parameters-only aggregate, leaving `MetricsArtifactWriter` nothing to write.

### Conversion contract

- Capture finite scalar pre-fit `trainer.validate(...)` results.
- Preserve them under
  `model.__fl_meta__[MetaKey.INITIAL_METRICS]`, the supported metadata channel
  used by the patched Lightning exchange.
- Keep the metrics associated with the received global model evaluated before
  local training.
- Require custom `ModelAggregator` implementations to aggregate supported
  client metrics with the same contribution weights and return them in the
  aggregated `FLModel.metrics`.
- Configure a source-derived recipe `key_metric` only when that exact metric is
  delivered to the server. Default DDP paths that do not provide the metric
  must omit it.
- Treat terminal `Finished` evidence without the requested server metric as
  incomplete validation.

The packaged Lightning client and shared aggregator templates implement this
single exchange path. No second manual `flare.send` path or runtime-specific
framework behavior is introduced.

### Problem

`FedStatsRecipe(sites=...)` creates a `StatsJob` and assigns every named site
with `to()`. A generated job that then used `SimEnv(num_clients=...)` supplied
client ownership a second time and failed before execution with:

`ValueError: You already specified clients using to(). Don't use n_clients in simulator_run.`

### Preferred construction

The public FedStats recipe schema requires `sites`, and the maintained tabular
and image examples reuse the same list for simulation:

```python
from nvflare.recipe import SimEnv
from nvflare.recipe.fedstats import FedStatsRecipe

sites = ["site-1", "site-2"]
recipe = FedStatsRecipe(..., sites=sites, ...)
env = SimEnv(clients=sites, workspace_root=workspace_root)
recipe.execute(env)
```

`SimEnv` is exported from `nvflare.recipe`; `FedStatsRecipe` is exposed by
`nvflare.recipe.fedstats`, not the package root. The exact paths apply to both
import preflights and generated `job.py`.

`SimEnv` derives the thread count from `clients`; generated jobs may set
`num_threads=len(sites)` when explicit control is needed. They must not use
`SimEnv(num_clients=...)` for `FedStatsRecipe` or provide both client forms.

The FedStats validation ladder now checks the public imports and site-ownership
invariant before the first simulator run. The representative eval requires the
canonical import paths and `sites`/`clients` pairing, and prohibits both the
package-root `FedStatsRecipe` import and `num_clients`. This is deliberately a
FedStats-specific skill fix: generic recipes that have not assigned named
clients may continue to use `SimEnv(num_clients=...)`, and no runtime code
changes are needed.

## Validation

- Inspector:
  - `121 passed` in `agent_inspector_test.py`
  - `145 passed` across Inspector, CLI, and JSON-contract tests
- Skill checks:
  - `151 passed` in `tests/unit_test/tool/agent_skill_checks`
  - real non-FedAvg recipe construction and FedEval capability coverage
  - plain-PyTorch `DataParallel` process-mode coverage
  - Lightning pre-fit metadata and weighted metric aggregation coverage
  - FedStats public-import and named-site ownership coverage against docs and
    both maintained examples
- PyTorch simulations:
  - in-process and external-process one-client FedAvg jobs completed round 0
  - tensor-native exchange, server disk offload, and recipe-level decomposer
    registration enabled
  - 1/1 updates aggregated with zero-byte server and client error logs
- Trigger evaluation:
  - 48 exact-current Haiku/Opus trials described above
- Performance:
  - a 241-file transitive source-job component scanned and classified in about
    0.21 seconds locally
- Quality gates:
  - `./runtest.sh -s` passed
  - repository and pre-push skill lint: 0 errors, 0 warnings, 0 informational
    findings

## Release constraints

- All shipped skill versions remain `0.1.0` for this release.
- Tensor disk offload remains documented as a server memory optimization, not
  model exchange.
- Framework-specific decomposer and metric behavior stays in recipes, skills,
  and generated templates rather than `app_common`.
- FedStats site targeting remains skill-owned and reuses explicit named clients;
  generic recipe simulation semantics are unchanged.
- Ambiguous Inspector results fail closed to `nvflare-orient` instead of
  selecting a converter or `nvflare-autofl`.

Addresses FLARE-3051.

### DIFF
--- a/nvflare/tool/agent/inspector.py
+++ b/nvflare/tool/agent/inspector.py
@@ -153,7 +153,8 @@ def inspect_path(
     ranked_frameworks = _rank_frameworks(state)
     detected_framework = _detect_primary_framework(state, ranked_frameworks)
     ranked_frameworks = _order_frameworks_for_display(ranked_frameworks, detected_framework)
-    conversion_state = _conversion_state(state, detected_framework, exported_job_info)
+    source_job_file = _authoritative_source_job_file(state)
+    conversion_state = _conversion_state(state, detected_framework, exported_job_info, source_job_file)
     target_type = _target_type(target, state, detected_framework, conversion_state)
 
     # Data-target classification runs when code classification found nothing,
@@ -217,7 +218,9 @@ def inspect_path(
         "findings": state.findings[:MAX_EVIDENCE_PER_BUCKET],
         "dataset": dataset,
         "skill_selection": _skill_selection(detected_framework, conversion_state, state, dataset),
-        "recommended_next_commands": _recommended_next_commands(detected_framework, conversion_state, state),
+        "recommended_next_commands": _recommended_next_commands(
+            detected_framework, conversion_state, state, source_job_file
+        ),
         "installed_skills": _installed_skills(target),
     }
 
@@ -1018,7 +1021,12 @@ def _exported_job_info(state: InspectState) -> dict:
     return {"submit_ready_candidates": submit_ready, "nested_candidates": nested}
 
 
-def _conversion_state(state: InspectState, detected_framework: Optional[str], exported_job_info: dict) -> str:
+def _conversion_state(
+    state: InspectState,
+    detected_framework: Optional[str],
+    exported_job_info: dict,
+    source_job_file: Optional[str],
+) -> str:
     if exported_job_info["submit_ready_candidates"]:
         return "exported_job"
     # job.py is a common filename (SLURM launchers) and SimEnv is a natural class
@@ -1026,7 +1034,7 @@ def _conversion_state(state: InspectState, detected_framework: Optional[str], ex
     # root-level source candidate with corroborating nvflare evidence from that
     # same file. This prevents historical jobs, fixtures, and vendored projects
     # from classifying the recursively inspected repository root as a FLARE job.
-    if _authoritative_source_job_file(state):
+    if source_job_file:
         return "flare_job"
     if _has_conversion_integration(state):
         return "client_api_converted"
@@ -1329,16 +1337,15 @@ def _has_problematic_skips(state: InspectState) -> bool:
 
 
 def _recommended_next_commands(
-    detected_framework: Optional[str], conversion_state: str, state: InspectState
+    detected_framework: Optional[str],
+    conversion_state: str,
+    state: InspectState,
+    source_job_file: Optional[str],
 ) -> list[str]:
     commands = []
     if conversion_state == "exported_job":
         commands.append("nvflare job submit <job-folder> --format json")
-    elif (
-        (source_job_file := _authoritative_source_job_file(state))
-        and source_job_file == "job.py"
-        and source_job_file in state.export_support_files
-    ):
+    elif source_job_file == "job.py" and source_job_file in state.export_support_files:
         # Only suggest `job.py --export` for a genuine FLARE job.py: `.export`
         # calls (torch.onnx.export, YOLO model.export, ...) over-match, so without
         # corroborating nvflare evidence this would ship a command that fails with
--- a/nvflare/tool/agent/inspector.py
+++ b/nvflare/tool/agent/inspector.py
@@ -440,7 +440,7 @@ class _PythonInspector(ast.NodeVisitor):
         self._detector_states = {detector.name: detector.new_file_state() for detector in self._detectors}
         self._ctx = DetectContext(
             self._emit_framework_evidence,
-            self.state.flare_calls.add,
+            self._add_flare_call,
             self._add_integration_signal,
         )
 
@@ -1051,16 +1051,18 @@ def _conversion_state(
 
 def _authoritative_source_job_file(state: InspectState) -> Optional[str]:
     flare_import_files = {item["file"] for item in state.flare_imports}
-    root_job_files = {path for path in state.job_py_paths if _is_root_level_file(path) and path in flare_import_files}
-    if root_job_files:
-        return sorted(root_job_files)[0]
-
-    # Preserve support for a root-level recipe script that constructs SimEnv
-    # under a filename other than job.py, while retaining file provenance.
-    root_sim_env_files = {
-        path for path in state.sim_env_files if _is_root_level_file(path) and path in flare_import_files
-    }
-    return sorted(root_sim_env_files)[0] if root_sim_env_files else None
+    candidate_files = set(state.job_py_paths) | state.sim_env_files
+    authoritative_candidates = [
+        path
+        for path in candidate_files
+        if path in flare_import_files and _file_belongs_to_active_root_project(state, path)
+    ]
+    if not authoritative_candidates:
+        return None
+    # A root-level candidate remains the strongest signal when both root and
+    # nested source jobs exist; otherwise accept a nested candidate that belongs
+    # to the active project (for example jobs/fedavg/job.py).
+    return min(authoritative_candidates, key=lambda path: (not _is_root_level_file(path), path))
 
 
 def _is_root_level_file(path: str) -> bool:
@@ -1085,25 +1087,41 @@ def _file_belongs_to_active_root_project(state: InspectState, file_path: str) ->
     if _is_root_level_file(file_path):
         return True
 
-    # Root-level Python files are the strongest available project anchors for a
-    # recursive repository inspection. Nested evidence belongs to that active
-    # project only when one of those files imports it (directly or transitively).
-    # If there are no root anchors, retain the historical whole-directory
-    # behavior for package-only layouts whose entry points live below the root.
-    root_files = {path for path in state.file_imports if _is_root_level_file(path)}
-    if not root_files:
+    anchor_files = _active_project_anchor_files(state)
+    if not anchor_files:
+        return True
+    if file_path in anchor_files:
         return True
     local_files_by_module = _local_files_by_module(state)
     return any(
         _imports_reach_file(
             state,
-            root_file,
-            state.file_imports.get(root_file, set()),
+            anchor_file,
+            state.file_imports.get(anchor_file, set()),
             file_path,
             local_files_by_module,
         )
-        for root_file in root_files
+        for anchor_file in anchor_files
     )
+
+
+def _active_project_anchor_files(state: InspectState) -> set[str]:
+    entry_point_files = {entry["path"] for entry in state.entry_points}
+    root_entry_points = {path for path in entry_point_files if _is_root_level_file(path)}
+    root_framework_files = {
+        item["file"]
+        for evidence in state.framework_evidence.values()
+        for item in evidence
+        if _is_root_level_file(item["file"])
+    }
+    root_anchors = root_entry_points | root_framework_files
+    if root_anchors:
+        return root_anchors
+
+    # A packaging-only root (for example setup.py plus src/mypkg/train.py) has
+    # no root training anchor. In that layout the nested entry points define the
+    # active project instead of the unrelated root scaffolding.
+    return entry_point_files
 
 
 def _has_conversion_integration(state: InspectState) -> bool:
--- a/nvflare/tool/agent/inspector.py
+++ b/nvflare/tool/agent/inspector.py
@@ -101,14 +101,19 @@ class InspectState:
     framework_evidence: dict[str, list[dict]] = field(default_factory=dict)
     flare_imports: list[dict] = field(default_factory=list)
     flare_calls: set[str] = field(default_factory=set)
+    flare_calls_by_file: dict[str, set[str]] = field(default_factory=dict)
     # framework name -> FLARE conversion-integration call names (e.g. Lightning
     # flare.patch). Populated by framework detectors; used by _conversion_state.
     integration_signals: dict[str, set[str]] = field(default_factory=dict)
+    integration_signal_files: set[str] = field(default_factory=set)
     file_imports: dict[str, set[str]] = field(default_factory=dict)
     entry_points: list[dict] = field(default_factory=list)
     job_py: Optional[str] = None
+    job_py_paths: list[str] = field(default_factory=list)
     sim_env_used: bool = False
+    sim_env_files: set[str] = field(default_factory=set)
     export_support: bool = False
+    export_support_files: set[str] = field(default_factory=set)
     exported_job_markers: list[str] = field(default_factory=list)
     exported_job_marker_paths: list[Path] = field(default_factory=list)
     distributed_patterns: list[dict] = field(default_factory=list)
@@ -398,7 +403,11 @@ def _inspect_file(path: Path, state: InspectState, max_file_bytes: int) -> None:
     state.files_scanned += 1
     state.bytes_scanned += size
     if path.name == "job.py":
-        state.job_py = rel_path
+        state.job_py_paths.append(rel_path)
+        # Keep the public summary pointed at the authoritative root candidate
+        # when one exists; a later nested fixture must not overwrite it.
+        if state.job_py is None or _is_root_level_file(rel_path):
+            state.job_py = rel_path
 
     try:
         tree = ast.parse(text, filename=str(path))
@@ -437,6 +446,7 @@ class _PythonInspector(ast.NodeVisitor):
 
     def _add_integration_signal(self, framework: str, name: str) -> None:
         self.state.integration_signals.setdefault(framework, set()).add(name)
+        self.state.integration_signal_files.add(self.rel_path)
 
     def visit_Import(self, node: ast.Import) -> None:
         for alias in node.names:
@@ -525,19 +535,25 @@ class _PythonInspector(ast.NodeVisitor):
         # framework activity (pytorch_call, lightning_trainer) and conversion
         # signals (flare.patch) are recorded by framework detectors via on_call.
         if call_name.startswith("flare.") or call_name.startswith("nvflare."):
-            self.state.flare_calls.add(call_name)
+            self._add_flare_call(call_name)
         if call_name in {"FedJob", "FLModel", "SimEnv"}:
-            self.state.flare_calls.add(call_name)
+            self._add_flare_call(call_name)
         if call_name == "SimEnv" or call_name.endswith(".SimEnv"):
             self.state.sim_env_used = True
+            self.state.sim_env_files.add(self.rel_path)
         if call_name.endswith(".export"):
             self.state.export_support = True
+            self.state.export_support_files.add(self.rel_path)
         if call_name in {"importlib.import_module", "__import__", "getattr"}:
             self.state.dynamic_patterns.append(_evidence(self.rel_path, lineno, "dynamic_dispatch", call_name))
         if call_name == "torch.compile":
             self.state.dynamic_patterns.append(_evidence(self.rel_path, lineno, "torch_compile", call_name))
         if call_name.endswith(("DataParallel", "FSDP", "Accelerator")):
             self.state.distributed_patterns.append(_evidence(self.rel_path, lineno, "distributed_call", call_name))
+
+    def _add_flare_call(self, call_name: str) -> None:
+        self.state.flare_calls.add(call_name)
+        self.state.flare_calls_by_file.setdefault(self.rel_path, set()).add(call_name)
 
     def _inspect_secret_assignment(self, targets: list[ast.AST], value: ast.AST, lineno: Optional[int]) -> None:
         if not isinstance(value, ast.Constant) or not isinstance(value.value, str):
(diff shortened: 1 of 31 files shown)


nvflare/tool/agent/inspector.py
```diff
@@ -1254,7 +1334,11 @@ def _recommended_next_commands(
     commands = []
     if conversion_state == "exported_job":
         commands.append("nvflare job submit <job-folder> --format json")
-    elif state.job_py and state.export_support and state.flare_imports:
+    elif (
+        (source_job_file := _authoritative_source_job_file(state))
+        and source_job_file == "job.py"
+        and source_job_file in state.export_support_files
+    ):
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in 7ad600cfa. `inspect_path` now computes `source_job_file` once and threads it into both `_conversion_state` and `_recommended_next_commands`, eliminating the duplicate provenance scan and making the shared dependency explicit. Validation: 108 inspector tests passed, `./runtest.sh -s` passed, and the pre-push agent skill lint was clean.

holgerroth (commented):
Two behavioral concerns on the new authority model, inline below.

nvflare/tool/agent/inspector.py
```diff
@@ -1002,33 +1021,102 @@ def _exported_job_info(state: InspectState) -> dict:
...
         return "client_api_converted"
-    if state.flare_imports or state.flare_calls:
+    if _has_authoritative_flare_evidence(state):
         return "partial_client_api"
     if detected_framework:
         return "not_converted"
     return "unknown"
 
 
+def _authoritative_source_job_file(state: InspectState) -> Optional[str]:
```
holgerroth: This hard-requires a *root-level* candidate, which means repos whose job scripts legitimately live in subdirectories (`jobs/fedavg/job.py`, examples-style layouts) can no longer classify as `flare_job` — even when there is no competing root project. The no-root-anchor fallback in `_file_belongs_to_active_root_project` keeps the *evidence* authoritative, but the best such a repo can now get is `partial_client_api`, which loses the `nvflare-autofl` routing entirely (`_skill_selection` only routes `flare_job` → autofl).

If "inspect the job's own directory, not the parent" is the intended scoping decision, could you add a test asserting the downgrade so it's documented behavior rather than an untested side effect?

nvflare/tool/agent/inspector.py
```diff
@@ -1002,33 +1021,102 @@ def _exported_job_info(state: InspectState) -> dict:
...
+
+
+def _has_authoritative_flare_call(state: InspectState, call_name: str) -> bool:
+    return any(
+        call_name in calls and _file_belongs_to_active_root_project(state, path)
+        for path, calls in state.flare_calls_by_file.items()
+    )
+
+
+def _file_belongs_to_active_root_project(state: InspectState, file_path: str) -> bool:
```
holgerroth: Strict mode activates when *any* root-level `.py` has imports, so packaging scaffolding becomes a false anchor. Concrete regression scenario: a src-layout, already-converted project — `setup.py` at root, `src/mypkg/train.py` with `flare.receive`/`flare.send`. `setup.py` imports only `setuptools` and reaches nothing under `src/`, so **all** flare evidence is discarded → `not_converted` → the inspector recommends `nvflare-convert-pytorch` for already-converted code.

Suggestion: also anchor on `state.entry_points` — framework evidence already does exactly this via `_framework_evidence_tied_to_inspected_file_or_entry_point` / `_entry_point_imports_file`, so the FLARE-authority check using a different, weaker anchor set is an avoidable inconsistency. Alternatively, fall back to whole-directory behavior when no root anchor reaches *any* nested file (a sign the anchors are packaging scaffolding, not the project).

nvflare/tool/agent/inspector.py
chesterxgchen: Fixed in 42a8d03ae. Nested `job.py`/`SimEnv` candidates are now accepted when they belong to the active project, while root-level candidates remain preferred when both exist. Added `test_nested_flare_job_source_is_authoritative_without_competing_root_project`, which verifies `jobs/fedavg/job.py` still classifies as `flare_job_source` and routes to `nvflare-autofl`; the original nested-fixture negative control still passes.

nvflare/tool/agent/inspector.py
chesterxgchen: Fixed in 42a8d03ae. Authority now prefers root entry points and root framework-bearing files; when those are absent (for example a packaging-only `setup.py` root), nested entry points become the active anchors. Added `test_src_layout_converted_project_is_not_hidden_by_root_packaging_scaffold`, which verifies `src/mypkg/train.py` with `flare.receive`/`flare.send` remains `client_api_converted` and is not routed back to conversion. Detector-originated FLARE calls now also flow through the per-file provenance callback. Full inspector suite: 110 passed; `./runtest.sh -s` passed.

chesterxgchen:
Addressed the Codecov patch-coverage comment in ac0e7cb78. Added focused end-to-end tests for the two previously uncovered changed branches: authoritative `FLModel` conversion and the no-project-anchor whole-directory fallback for nested FLARE evidence. Local focused coverage no longer reports changed lines 1044 or 1092 as missing; inspector suite is now 112 passed, and `./runtest.sh -s` passes.

chesterxgchen:
Follow-up on the source-job authority review:

- Audited all 74 tracked `job.py` files in this repository; all 74 import `nvflare` directly, so the previous same-file gate did not downgrade an in-tree example.
- The concern is valid for external projects, so commit `73d059bd6` now accepts NVFLARE evidence in the candidate file **or its reachable local import closure**. This supports `job.py -> job_utils.py -> nvflare` while an unimported NVFLARE helper or unrelated nested fixture cannot promote the candidate.
- Added positive delegated-helper and negative unreachable-helper regressions.
- Nested authoritative jobs with export support now receive a path-aware command such as `python jobs/fedavg/job.py --export --export-dir <job-dir>`.
- Rebased onto current `upstream/main` (`2c63764cc`).

Validation after rebase: 114 inspector tests passed; `./runtest.sh -s` passed, including agent-skill lint with 0 findings. All existing inline review threads are resolved; the new CI run is in progress.

chesterxgchen:
Implemented the component-based alternative in `acf8f5b27`.

The inspector now has three explicit internal layers:

1. `_LocalImportGraph` resolves local imports once and caches one cycle-safe traversal per source file.
2. `_ProjectComponent` / `_ProjectScope` select the nearest entry-point or framework-anchored components. Deeper components need directed reachability from those anchors; equally near independent components make the root ambiguous rather than allowing cross-project promotion. Components that merely share a helper stay independent.
3. `_SourceJobCandidate` correlates each `job.py`/`SimEnv` candidate only with NVFLARE imports in its own component and records the authorizing project component.

New coverage verifies a deeper fixture beside a nested PyTorch project, same-depth independent projects sharing a helper, same-depth components connected by an import, and a root launcher importing a nested source job. Validation: 118 inspector tests pass; `./runtest.sh -s` passes; pre-push skill lint is clean. A bounded 241-file transitive component classified in about 0.21 seconds locally, with no additional model calls, tokens, or conversion passes.

chesterxgchen:
Addressed the PyTorch-family skill ownership/capability review in `c93a6dc54`:

- added one canonical `pytorch-family-recipe-construction.md` policy consumed by both plain PyTorch and Lightning
- recipe settings are now capability-gated after `recipe show`: tensor format when exposed, disk offload when exposed, and `TensorDecomposer` when tensor-native exchange is selected
- external execution now follows distributed process-spawning evidence; plain `torch.nn.DataParallel` and Lightning `dp` stay in-process
- moved writable Pandas positional-index guidance to the framework-neutral conversion workflow
- moved `best_model_filename` / deprecated `save_filename` ownership to the shared PyTorch-family recipe layer
- replaced duplicated-wording admission assertions with canonical-policy, real catalog-capability, and real non-FedAvg construction checks
- added FedEval-without-offload and plain-PyTorch DataParallel eval coverage

Validation: `146 passed` in `tests/unit_test/tool/agent_skill_checks`; `./runtest.sh -s` passes; pre-push agent-skill lint reports 0 findings. The branch is current with `upstream/main`, and all inline review threads are resolved.

chesterxgchen:
Addressed the latest skill-ownership review in b597ead81.

- Kept all skill versions at `0.1.0`; this is the required version for the current release, so the proposed forward bump does not apply.
- Made `pytorch-family-recipe-construction.md` the canonical owner of recipe capability parsing, tensor transport/decomposer setup, transfer mode, best-model filename/metric, and process-based execution.
- Separated `enable_tensor_disk_offload` from model exchange: it is documented as a server memory optimization that downloads incoming streamed tensors to temporary files and materializes them lazily during aggregation, reducing peak memory/OOM risk. It is enabled only when tensor-native transport is selected and the recipe exposes the capability.
- Removed duplicated tensor-send, metric-direction, execution-mode, and FedAvg optional-setting policy from consumer references; consumers now delegate to the canonical owners.
- Restored the Lightning construction/runtime-output load map, moved framework-neutral HE wording out of the PyTorch-only reference path, and limited diagnosis artifact download to terminal jobs with a reported failure signal.
- Added ownership, capability, server-offload semantics, and release-version regression assertions.

Validation: `148 passed` in `tests/unit_test/tool/agent_skill_checks`; `./runtest.sh -s` passes; pre-push skill lint reports 0 findings. The branch was fetched/rebased against `upstream/main` (already current) before push.

chesterxgchen:
Addressed the additional review findings in `ebb34b46e`:

- **Nested converted entry point:** project anchors are now grouped by directed import connectivity before root-depth selection. A nested `jobs/train.py` that imports root `model.py` retains authority, while a separate deeper fixture remains excluded.
- **Ambiguous workspaces:** equally near independent entry-point projects now return `conversion_state: ambiguous`, `target_type: mixed_workspace`, and only `nvflare-orient`; no converter command is emitted.
- **Dependency plan:** “one canonical command” now explicitly means one combined installer invocation containing every applicable `-r`/`-c` input, direct declarations, and `nvflare` when missing/not otherwise supplied. Failed command/error reporting is explicitly redacted.
- **Lightning DDP metric boundary:** a source-derived `key_metric` now requires verified server metric delivery. The default Lightning DDP path omits that customization and reports the best-model/round-metric limitation unless an advanced `train_with_evaluation=True` path is selected.

The earlier single-sourcing fixes remain in `b597ead81`; all skill versions intentionally remain `0.1.0` for this release.

Validation: 119 inspector tests passed; 143 inspector/CLI/JSON contract tests passed; 148 skill checks passed; `./runtest.sh -s` passed; pre-push skill lint reports 0 findings. The branch was fetched/rebased against `upstream/main` (already current) before push.

chesterxgchen:
Addressed the Lightning metrics regression from benchmark run `745846` in `50b87e07d`.

- The generated Lightning client now captures finite scalar pre-fit `trainer.validate(...)` results and preserves them under `model.__fl_meta__[MetaKey.INITIAL_METRICS]`, so the existing patched exchange carries them even when `train_with_evaluation` is disabled.
- The shared custom `ModelAggregator` template now step-weights supported client metrics and returns them in aggregated `FLModel.metrics`; parameters-only aggregates are prohibited when metrics are requested.
- Validation now requires the requested server metric evidence and does not accept terminal `Finished` alone.
- Added deterministic template tests plus a benchmark-shaped Lightning custom-aggregation eval.

Validation: `150 passed` in `tests/unit_test/tool/agent_skill_checks`; `./runtest.sh -s` passed; pre-push skill lint reports 0 findings. All skill versions remain `0.1.0`. The branch is current with `upstream/main`.

chesterxgchen:
Resolved the remaining equal-depth fixture authority finding in `a09da0761`.

The inspector still prefers a genuine entry-point group over an unrelated framework-only helper at the same depth, but it no longer gives that preference to entry points under secondary trees (`tests`, `test`, `fixtures`, `archive`, or vendored directories). Therefore independent `src/pkg/model.py` and `tests/fixture/job.py` components at equal depth remain ambiguous and route only to `nvflare-orient`; the fixture cannot classify the repository as `flare_job`.

Added the exact regression and retained the existing context-resolved package/entry-point controls.

Validation: focused authority set `9 passed`; full inspector suite `120 passed`; inspector/CLI/JSON contracts `144 passed`; `./runtest.sh -s` passed; pre-push skill lint is clean. The branch was fetched/rebased against current `upstream/main` before push. The three residual documentation notes were explicitly optional and pre-existing, so I left them out of this scoped fix.

chesterxgchen:
Addressed the connected-fixture authority finding in `7626a342a`.

- Secondary-tree FLARE evidence remains visible but cannot establish repository-level conversion authority, even when `tests/fixture/job.py` imports production code and joins its import component.
- Direct inspection of the fixture subtree remains the positive control and still routes to `nvflare-autofl`.
- Added the connected fixture regression; inspector/CLI/JSON coverage now passes 145 tests.
- Kept the equal-depth all-or-nothing tie behavior intentionally and documented its conservative bias toward `mixed_workspace` ambiguity rather than false FLARE authority. The broad `test`/`tests` boundary remains root-relative, so directly inspected projects under those directories are unaffected.
- `./runtest.sh -s` passed; repository skill lint and the pre-push lint both report 0 findings.

chesterxgchen:
Revalidated the diagnose-skill trigger comparison against the exact current `SKILL.md`.

The audit found that the original table was a valid historical candidate sweep, but its “Final concise” row corresponded to the pre-review description at `e86a009d2`; the exact frontmatter/body changed later in `bafb0f965`. I corrected the PR description so it no longer presents that old row as a current-text measurement.

Exact-current rerun (`7626a342a`, SKILL.md SHA-256 `54b441074634bfa63179df6204c3f907fb12b5f7764ccf26f1c5a7a0ddea85f8`):

| Model | Positive activation | Diagnosis false-trigger on near-miss |
| --- | ---: | ---: |
| Claude Haiku 4.5 | 12/12 | 0/12 |
| Claude Opus 4.8 | 12/12 | 0/12 |

Method: same four positive + four NVFLARE near-miss prompts, three attempts each, prompts did not name the skill, Claude Code 2.1.220 outside the sandbox, and `--tools Skill`. All 48 init events reported exactly `["Skill"]` and the requested canonical model. Installed skill metadata remained visible, so this was a competitive catalog; some near-misses chose unrelated installed skills, but none chose `nvflare-diagnose-job`. The historical Haiku 1/12 versus current 0/12 should be treated as sampling variation.

Deterministic scope test passes and the repository skill lint reports 0 findings.

holgerroth (approved):
Approving — I verified the `_ProjectScope` authority model end-to-end against the nested-fixture, nested-only-job, delegated-job.py, and src-layout scenarios, and confirmed the Lightning `__fl_meta__` → `MetaKey.INITIAL_METRICS` bridge against the product code (`app_opt/lightning/api.py`, `fl_model_utils`, `IntimeModelSelector`). Nice work on the redesign.

Two non-blocking questions before/shortly after merge:

1. **`validate_global_model` hard-fail tightening** (`skills/nvflare-convert-lightning/assets/lightning_client.py`): it changed from permissively returning `trainer.callback_metrics` to raising on empty/non-scalar/non-finite validation results, and the generic round-loop snippets in `lightning-conversion.md` and the DDP reference now call it unconditionally. A converted `LightningModule` that logs nothing in `validation_step` now crashes at runtime where it previously proceeded metric-less. Is that intended for conversions that never requested server metrics, or should the metrics-mandatory path be reserved for the server-metrics case?

2. **Diagnose frontmatter re-benchmark**: the PR description's A/B table justified the earlier "concise positive" wording, but the final description ("Use when the user asks why a reported NVFLARE job failure signal occurred: …") differs from all four benchmarked variants. Given the table showed small wording changes move Haiku trigger rates, was the final wording re-run through the same trigger harness? If not, worth one more pass.

PR #4956, opened 2026-07-24
Add client-local require_signed_jobs policy

pcnudde:
## Summary

- add a client-local `require_signed_jobs` policy for received job deployment bytes
- preserve exact-byte signature verification and reject invalid signatures even when unsigned jobs are allowed
- let `StaticFileBuilder` render the same optional policy into server and client startup configs
- document the site-local opt-out and update the CLI guidance

## Root cause

PR #4817 correctly changed clients to verify the exact app ZIP received from the server, but it also made every non-hub client with `rootCA.pem` reject unsigned bytes unconditionally. The server continued to honor `require_signed_jobs: false`, so an unsigned job could be accepted by the server and then rejected by every client, ending as `FINISHED:FAILED_TO_RUN`.

This change resolves FLARE-3054 by allowing a client to opt out only through its own startup configuration. The policy is not accepted from the deploy request, so a server cannot disable client-side signature enforcement while substituting job bytes.

## Behavior

- missing policy: preserve the existing default (`true` when `rootCA.pem` exists)
- `require_signed_jobs: true`: reject unsigned job bytes
- `require_signed_jobs: false`: allow bytes with no signature
- signature present but invalid: always reject
- malformed or non-boolean policy: fail closed

For centralized provisioning, `StaticFileBuilder(require_signed_jobs=False)` renders the setting into both server and client startup configs. Distributed sites can independently include the same builder argument in their participant definition.

## Validation

- `74 passed` across client/server signature policy, deploy, hub, provisioning, and distributed-package tests
- `./runtest.sh -s`
- `git diff --check`


chesterxgchen:
1. P2: require_signed_jobs=true is ineffective without rootCA.pem.
     Client verification is gated by the presence of the root CA. An explicitly strict policy can therefore accept unsigned jobs when the CA is
     absent. The client should fail closed with a configuration error when signing is required but verification material is unavailable.

  2. P3: Legacy custom-template compatibility.
     This is narrower than my original wording:
      - The running client does not copy or process templates.
      - Standard distributed provisioning uses the updated bundled template and is unaffected.
      - The issue applies only when a site administrator explicitly packages with a custom master_template.yml copied from before this PR.
      - Such a template lacks the new placeholder, so require_signed_jobs=False may be silently omitted from the generated fed_client.json. The
        resulting client remains strict by default.

     This fails securely, so it is a compatibility/usability issue, not a security vulnerability. A rendered-output validation or an end-to-end
     legacy-template test would address it.

pcnudde:
Thanks — addressed both points in bc28ae472.

1. Missing trust root: the client now evaluates its local require_signed_jobs policy independently of rootCA.pem. When signing is required, unsigned payloads are rejected, and signed payloads are rejected with a clear verification-material error if rootCA.pem is unavailable. The no-CA compatibility path remains unchanged when signing is not required.
2. Legacy custom templates: StaticFileBuilder now validates the rendered fed_server.json and fed_client.json whenever require_signed_jobs is explicitly configured. A stale custom template that omits the placeholder now fails provisioning with a clear update instruction instead of silently dropping the setting.

Regression coverage includes both signed/unsigned missing-CA cases and legacy server/client templates. Validation:
- 280 related tests passed
- ./runtest.sh -s passed
- git diff --check passed

chesterxgchen (approved):

PR #4959, opened 2026-07-24
Fix FedJob subdirectory sibling imports

nvkevlu:
Fixes the FedJob export regression introduced by #4885 for unqualified sibling imports in registered scripts under flat subdirectories.

## Final import-discovery rules

- A top-level registered script such as `client.py`, or a script in a flat directory without `__init__.py` such as `src/client.py`, searches its own directory before the project source root for unqualified absolute imports.
- Discovered top-level modules continue the same sibling discovery recursively.
- A package-qualified script such as `pkg/client.py`, where `pkg/__init__.py` exists, does not interpret `from helper import Helper` as `pkg/helper.py`. Use a package-qualified import such as `from pkg.helper import Helper` to export that file under `custom/pkg/`.
- If different discovered files map to the same export destination, export raises `ValueError` instead of silently overwriting one file.

## Compatibility

| Scenario | NVFlare 2.8 | Main after #4885 | This PR |
| --- | --- | --- | --- |
| `src/client.py` imports sibling `net.py`; `src/` has no `__init__.py` | Exports sibling | Misses sibling | Exports sibling |
| Discovered `net.py` imports sibling `helper.py` | Exports both recursively | Stops before `net.py` | Exports both recursively |
| Top-level `client.py` is registered from a directory containing `__init__.py` | Exports sibling | Exports sibling | Exports sibling |
| Both `src/net.py` and project-root `net.py` exist | Source-directory file wins | Project-root file wins | Source-directory file wins |
| `pkg/client.py` imports `pkg.helper` | Exports `custom/pkg/helper.py` | Exports it | Exports it |
| `pkg/client.py` imports unqualified `helper`, with `pkg/__init__.py` present | Exports the package sibling as a top-level module, allowing shadows | Does not export package sibling | Deliberately does not export package sibling |
| Two flat script directories discover different files that both map to `custom/net.py` | First file silently wins | Neither sibling is discovered | Export fails with `ValueError` |

## Import terminology

- **Unqualified absolute import:** `import helper` or `from helper import Helper`; this names the top-level module `helper`.
- **Package-qualified absolute import:** `import pkg.helper` or `from pkg.helper import Helper`; this explicitly names `pkg/helper.py`.
- **Explicit relative import:** `from .helper import Helper`; this requires package context and is separate from the script-style import case fixed here.

## Validation

- Added regression coverage for flat subdirectory sibling discovery and the ScriptRunner simulator path.
- Added compatibility-boundary coverage for top-level scripts authored from package directories, package-qualified imports, and unqualified package collisions.
- FedJob and ScriptRunner tests: 82 passed.
- Previously affected recipe export tests: 4 passed.
- `./runtest.sh -s` passed.


nvflare/job_config/fed_job_config.py
```diff
@@ -458,13 +465,16 @@ def _copy_source_file(self, custom_dir, module, source_file, dest_file, source_r
         shutil.copyfile(source_file, dest_file)
 
         source_dir = os.path.dirname(source_file)
+        is_flat_external_script = is_external_script and not os.path.isfile(os.path.join(source_dir, "__init__.py"))
+        search_source_dir = is_flat_external_script or (not is_external_script and "." not in module)
```
YuanTingHsieh: [P2] This condition differs from both `2.8` and current `main` for a top-level registered script. If a job is authored from inside `pkg/` (where `pkg/__init__.py` exists) and registers `client.py`, then `module == "client"`. The `2.8` implementation always checks `source_dir`, and current `main` checks it because `"." not in module`; here, however, `is_external_script` is true and `is_flat_external_script` is false, so `search_source_dir` becomes false and an imported sibling such as `helper.py` is silently omitted. Please preserve the top-level-module arm independently, for example with `is_flat_external_script or "." not in module`.

tests/unit_test/job_config/fed_job_config_test.py
```diff
@@ -63,6 +63,21 @@ def test_copy_ext_script_finds_top_level_import_in_same_directory(self, tmp_path
...
+        (script_dir / "client.py").write_text("from net import Net\n", encoding="utf-8")
+        monkeypatch.chdir(tmp_path)
+
+        custom_dir = tmp_path / "exported" / "custom"
+        job_config = FedJobConfig(job_name="job_name", min_clients=1)
+        job_config._copy_ext_scripts(str(custom_dir), ["src/client.py"])
+
+        assert (custom_dir / "src" / "client.py").is_file()
+        assert (custom_dir / "net.py").is_file()
+        assert not (custom_dir / "src" / "net.py").exists()
```
YuanTingHsieh, line 94: [P2] Please add tests that make the `2.8` compatibility boundary explicit. At minimum cover: (1) `cwd=pkg/`, register top-level `client.py`, with `pkg/__init__.py` and sibling `helper.py`—the helper should still be exported; (2) register package-qualified `pkg/client.py` with an unqualified colliding import—the package sibling should not be exported; and (3) register `pkg/client.py` with an explicitly qualified `pkg.helper` import—the helper should be exported under `custom/pkg/`. The first case catches the regression above, while the latter two document the deliberate difference from `2.8`.

YuanTingHsieh (commented):
Compared this import-export path with the current `2.8` branch. I left two comments: one compatibility issue in the new condition and one request for tests that make the intentional package boundary explicit.

YuanTingHsieh (commented):
The latest implementation looks good; the earlier condition and boundary-test concerns are addressed. One non-blocking clarification would make the compatibility decision much easier to understand: this restores the useful NVFlare 2.8 behavior for top-level and flat script directories, but deliberately does not restore all 2.8 behavior.

| Case | 2.8 | Pre-4959 main / #4885 behavior | 4959 |
|---|---|---|---|
| `src/client.py` uses `from net import Net`, with sibling `src/net.py` and no `__init__.py` | exports sibling | misses sibling | exports sibling |
| `net.py` then imports sibling `helper.py` | exports both recursively | discovery stops before `net.py` | exports both recursively |
| top-level `client.py`, authored from a directory containing `__init__.py` | exports sibling | exports sibling | exports sibling |
| both `src/net.py` and project-root `net.py` exist | source-directory file wins | project-root file wins | source-directory file wins |
| `pkg/client.py` uses `from pkg.helper import Helper` | exports `custom/pkg/helper.py` | exports it | exports it |
| `pkg/client.py` uses `from helper import Helper`, with `pkg/__init__.py` present | exports the package sibling as top-level, including possible stdlib shadows | does not export it | does not export it deliberately |
| two flat script directories resolve different sibling `net.py` files to `custom/net.py` | first silently wins | neither sibling is discovered | export fails with `ValueError` deliberately |

It would also help to define the import terminology:

- **Unqualified absolute import:** `import helper` or `from helper import Helper`; this names top-level module `helper`.
- **Package-qualified absolute import:** `import pkg.helper` or `from pkg.helper import Helper`; this explicitly names `pkg/helper.py`.
- `from .helper import Helper` is an explicit relative import and should be discussed separately, since scripts executed by path may not have a package context.

Could the PR description state this distinction? The user-facing documentation in #4914 should describe the final rule without the historical columns. As optional hardening, tests for chained discovery, source-directory precedence when a root module has the same name, and conflicting discovered dependencies would lock down the remaining rows.

YuanTingHsieh (approved):

PR #4961, opened 2026-07-24
Make Slurm CLI timeouts configurable

pcnudde:
## Summary

- add site-owned `submit_timeout`, `query_timeout`, and `cancel_timeout` settings to the Slurm launcher configuration
- use those settings for `sbatch`, `squeue`/`sacct`, and `scancel`
- document the defaults and the remaining accepted-but-untracked submission limitation

## Motivation

The Slurm launcher applied one hardcoded 10-second timeout to submission, query, and cancellation commands. In a real campaign, an `sbatch` call exceeded that bound and caused an otherwise valid NVFlare job to fail before the launcher received its Slurm job ID.

This change makes the bounds site-configurable and raises the default submission timeout to 30 seconds while preserving the existing 10-second defaults for queries and cancellation. It intentionally does not add submission reconciliation; an accepted job whose `sbatch` response is lost remains a documented limitation.

## Validation

- `173 passed` across the focused Slurm launcher and deployment tests
- `UV_CACHE_DIR=/private/tmp/uv-cache ./runtest.sh -s`


holgerroth (approved):
LGTM. Verified the deleted `_COMMAND_TIMEOUT` has no dangling references, all three `normalize_slurm_launcher_settings` callers are updated, and the validation tests are fully backed by `_require_positive_number`. Two optional follow-ups: consider having `accounting_probe` honor `query_timeout` instead of its hardcoded 2.0s (same slow-scheduler conditions apply to sacct), and the `cancel_suffix` compensation path is the one wired-up timeout site without a test assertion.

nvflare/app_opt/job_launcher/slurm/config.py
```diff
@@ -334,6 +340,9 @@ def normalize_slurm_launcher_settings(
     internal_port = _require_int(internal_port, "internal_port")
     if internal_port > 65535:
         raise SlurmLauncherError("internal_port must be at most 65535")
+    submit_timeout = _require_positive_number(submit_timeout, "submit_timeout")
```
chesterxgchen, line 343: should we use pydyntic to do the validation

pcnudde: Good suggestion. We'll consider using Pydantic in a follow-up PR if evaluating the full Slurm configuration and validation boundary shows that it simplifies the overall implementation. For this PR, we'll keep the existing validation approach to stay scoped.

PR #4962, opened 2026-07-24
Add launcher-owned multi-node Slurm training

pcnudde:
## Summary

- Add a scheduler-neutral node-group environment contract and a PyTorch `torchrun_node` helper.
- Let the Slurm launcher start one task per allocated node under bare, Apptainer, or Pyxis execution.
- Keep rank 0 as the normal NVFlare client-job process and run `additional_node_command` on every other node.
- Reuse a managed external-process `ScriptRunner` command when an explicit multi-node launcher block omits `additional_node_command`, while retaining explicit override and application-owned fan-out.
- Document topology, rendezvous port policy, lifecycle, sandbox behavior, and the rendezvous-first constraint for hand-written commands.

## Why

Before this change, NVFlare could request multiple Slurm nodes, but the application had to own process fan-out and containerized multi-node jobs were unsupported. The launcher can now own the node group while preserving the existing client-job lifecycle and scheduler identity.

## User impact

PyTorch jobs can use the same `torchrun_node` command for single-node and multi-node execution. Recipe users author one external training command and set `nodes` and `gpus_per_node` for the multi-node site; export copies the fully assembled command into `additional_node_command`.

An explicit `additional_node_command` always wins, and explicit `null` preserves application-owned fan-out. Default/inherited launcher blocks and custom launchers still provide the full command explicitly. Neither generated nor explicit additional-node commands support secret references.

A higher-level Recipe convenience API for topology and GPU resources is intentionally left for a follow-up PR.

## Validation

- 207 focused ScriptRunner, Recipe, Slurm launcher/batch, and `torchrun_node` tests passed.
- `./runtest.sh -s` passed (Black, isort, flake8, and agent-skill lint).
- Full `./build_doc.sh --html` completed successfully.
- `git diff --check` passed.

## Remaining validation

- A fresh live Slurm QA cycle is running against exact head `a14a9853874cfa1672d90a33289edab237b76d1f`.


nvflare/client/ex_process/api.py
```diff
@@ -144,7 +166,10 @@ def _configure_subprocess_logging(self, client_config: ClientConfig) -> None:
                 return
 
             dict_config = conf.to_dict()
-            _downgrade_rotating_handlers(dict_config)
+            if rank == "0":
```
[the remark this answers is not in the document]
pcnudde: Thanks for flagging this. The supported `flare.init()` path already normalizes every integer rank to a string in `nvflare/client/api.py` before constructing `APIContext`, so `ExProcessClientAPI.init()` and `_configure_subprocess_logging()` receive `"0"`, not `0`. `TestPublicAPIInit::test_explicit_then_implicit_effective_rank_reuses_context` covers integer normalization; I reran it and both parameterized cases pass. Direct `ExProcessClientAPI.init(rank=0)` is outside its `Optional[str]` internal contract, so no product change is needed here.

pcnudde:
@greptileai review again

nvflare/app_opt/job_launcher/slurm/batch.py
```diff
@@ -89,23 +149,105 @@ def _apptainer_parts(plan: LaunchPlan, config: SlurmConfig, worker_words: list[s
...
+def _render_node_script(plan: LaunchPlan, config: SlurmConfig) -> str:
+    """Render the per-node dispatch script for a launcher-owned node group.
+
+    For bare mode it runs on the host; for Pyxis it runs inside the per-task
+    container that srun already created; for Apptainer it runs on the host and
+    starts the per-node container itself.
+    """
+    worker_words = _build_worker_words(plan)
+    node_words = [shlex.quote(word) for word in plan.additional_node_command]
+    nonzero_setup = [
```
YuanTingHsieh, line 209: The CJ bootstrap credentials currently become part of the common `srun` environment. `plan.study_secret_env` also contains `get_credential_env(job_args)`, so `NVFLARE_JOB_AUTH_TOKEN`, `NVFLARE_JOB_TOKEN_SIGNATURE`, and `NVFLARE_JOB_SSID` reach every `node.sh` (and are explicitly forwarded by the container backends). Node 0 is safe because the CJ parser consumes/removes them before `ClientAPIExecutor` launches trainer rank 0, but ranks 1..N-1 execute `additional_node_command` directly and never clear them.

The intended boundary seems to be: CJ credentials only to the node-0 CJ; rendezvous variables to all trainers. Could we either deliver/source the CJ credentials only in the rank-0 branch, or at minimum unset the three `JobProcessEnv` variables (plus Apptainer mirrors) before the nonzero command?

nvflare/job_config/script_runner.py
```diff
@@ -59,6 +62,32 @@ def _to_external_process_argv(value: _CommandArg, arg_name: str) -> list[str]:
...
+def _fill_additional_node_command(job: FedJob, target: str, command: list[str], launch_once: bool) -> None:
+    """Copy a managed external-process command into explicit multi-node launcher blocks."""
+    meta_props = job.job.meta_props
+    launcher_spec = meta_props.get(JobMetaKey.JOB_LAUNCHER_SPEC.value) if isinstance(meta_props, dict) else None
+    if not isinstance(launcher_spec, dict):
+        return
+
+    command_text = None
+    for site_name, site_spec in launcher_spec.items():
+        if site_name == "default" or (target != ALL_SITES and site_name != target) or not isinstance(site_spec, dict):
```
YuanTingHsieh: When `target == ALL_SITES` (the normal `job.to_clients()` path), this loop visits every top-level launcher entry except `default`, including `launcher_spec["server"]`. A valid application-owned multi-node server block such as `{"slurm": {"nodes": 2}}` would therefore receive the client's generated `additional_node_command`, and `ServerSlurmJobLauncher` will reject it because that field is client-only. Could we exclude `SERVER_SITE_NAME` or otherwise restrict this to client site blocks?

YuanTingHsieh (commented):
Two issues found around multi-node command generation and credential scope.

nvflare/app_opt/job_launcher/slurm/batch.py
pcnudde: Fixed in a0fe74c56. The non-zero node branch now removes all three JobProcessEnv bootstrap credentials before executing additional_node_command; Apptainer environment mirrors are removed there as well. Rank 0 still retains the credentials for the CJ parser. The node-script execution test now proves that rank 0 receives the credentials while rank 1 receives none, and the Apptainer renderer test covers the mirrored variables. The affected test set passes 272/272 and the full style gate passes.

nvflare/job_config/script_runner.py
pcnudde: Fixed in a0fe74c56. Managed client command generation now skips SERVER_SITE_NAME, so job.to_clients() cannot add additional_node_command to a server launcher block. The regression coverage includes a multi-node server block alongside the client blocks and verifies that only the eligible client block is populated. The affected test set passes 272/272 and the full style gate passes.

YuanTingHsieh (approved):
Approving — both review findings are addressed: the server-block exclusion in export, and the credential unset in node.sh (verified across bare/Apptainer/Pyxis, with the executable test now asserting unset-ness at rank 1).

One scoping note for the record: with this fix, the CJ bootstrap credentials still briefly reach the trusted generated node.sh on ranks 1..N-1 through the common srun environment, where they are removed before any user training code executes. So the enforced boundary is "trainers never hold CJ credentials" — the same guarantee pop_credential_env provides on node 0 — but "credentials never leave node 0 at all" is not implemented. The transit is confined to launcher-generated, exclusive-create 0700 scripts, so this is acceptable as-is; if we ever want the stronger node-0-only property, it would require rank-scoped secret delivery rather than the shared srun environment. Follow-up material, not a blocker.

PR #4966, opened 2026-07-25
Optimize F3 tensor transfer performance and benchmarks

nvidianz:
## Summary

- add a CellNet benchmark with raw F3 streaming and optimized Python/TCP modes for high-speed networks
- add YAML-based F3 tuning, including binary size units such as `128K`, `16M`, and `2G`
- add an end-to-end TensorDecomposer and Download Service benchmark for memory and disk-offload modes
- eliminate repeated Download Service transactions for aliased tensors
- avoid synchronous look-ahead serialization and unnecessary full-size FOBS buffer copies
- tune and document the F3 streaming defaults used by the benchmark

## Why

Raw F3 streaming sustained approximately 2.6 GiB/s on the 25 Gbit/s test network, while the original end-to-end tensor path reached only about 250–260 MiB/s. Profiling found higher-layer bottlenecks in alias handling, tensor production, FOBS reply staging, and multi-buffer section assembly.

The optimized path keeps raw F3 fan-out unchanged while reducing redundant downloads, serialized work, staging allocations, and buffer copies.

## Impact

- provides reproducible transport and tensor-transfer benchmarks under `dev_tools/f3`
- improves full-checkpoint throughput to 399.6 MiB/s in memory mode and 452.9 MiB/s in disk mode in the measured environment
- improves the alias-heavy subset to 456.5–468.7 MiB/s by eliminating repeated downloads
- adds `BufferList.read_bytes()` as an additive one-copy assembly path while preserving the existing `BufferList.read()` behavior

## Validation

- `475 passed in 26.67s` across the changed tensor downloader, F3 streaming, FOBS, benchmark, and buffer-list tests
- Black, isort, and Flake8 passed for all 18 changed Python files
- `git diff --check upstream/main...HEAD` passed
- remote memory and disk-offload benchmark runs passed tensor-count, key, shape, dtype, size, and sample SHA-256 validation

The repository-wide `./runtest.sh -s` wrapper was attempted. It was blocked only by three unrelated vendored `node_modules/.../flatted.py` files that Black reports would be reformatted; the scoped checks for every Python file changed by this PR pass.


nvidianz:
Superseded by #4973 — closing.

#4973 contains every code change in this PR: 22 of the 25 changed files have byte-identical diffs, and `nvflare/fuel/f3/streaming/cacheable.py` is a strict superset there (same `prefetch_item`/`get_item_size` additions and `produce()` changes, plus prefetch depth raised from 1 to 2).

#4973 additionally adds the pipelined download loop in `download_service.py` with its unit tests, and consolidates the benchmark docs into `dev_tools/f3/README.md`.

The only files unique to this PR are `cellnet_bench_RESUME.md` and `tensor_download_bench.md`. The former is working notes from a superseded run (including a stale host address); the latter is replaced by the new README. Neither is carried forward intentionally.

Please review #4973 instead.

PR #4968, opened 2026-07-27
Redact headers from unknown Cell reply warnings

pcnudde:
## What

- Stop including the complete reply-header dictionary in the warning for an unknown Cell request ID.
- Add a regression test with sentinel header values to verify the warning remains useful without exposing header contents.

## Why

A late reply can arrive after its request waiter has already been removed. The unknown-reply warning serialized every header in that reply, including values that may be sensitive.

## Impact

The warning still identifies the unknown request ID and records that the reply was discarded. Only the raw header dump is removed.

This is a pre-existing CellNet logging issue discovered while validating PR #4962; that PR does not modify the affected CellNet path.

## Checks

- `python -m pytest tests/unit_test/fuel/f3/cellnet/cell_progress_wait_test.py -q` — 10 passed
- `python -m pytest tests/unit_test/fuel/f3/cellnet -q` — 210 passed
- `./runtest.sh -s` — passed


holgerroth (approved):
makes sense!

IsaacYangSLA (approved):
Thanks.  LGTM.

PR #4972, opened 2026-07-27
Reject shared-file transport in container deploy prepare

pcnudde:
## Summary

- reject an explicit `internal.scheme: shared-file` configuration during Docker and Kubernetes deployment preparation
- report an `INVALID_KIT` error with guidance to use TCP, the original process-mode kit, or the Slurm runtime
- keep default and explicit TCP preparation behavior unchanged
- add parameterized regression coverage for both container runtimes and verify failed preparation leaves no output

## Why

Docker and Kubernetes deployment preparation only supports the TCP internal transport. It previously rewrote an explicitly selected shared-file transport to TCP and completed successfully, silently discarding the user's configuration.

Failing fast makes the supported runtime boundary clear and avoids producing a prepared kit with different communication semantics than requested.

## Validation

- `python -m pytest tests/unit_test/tool/deploy/deploy_commands_test.py -q` (`102 passed`)
- `./runtest.sh -s`
- `git diff --check`


IsaacYangSLA (approved):
Thanks.  LGTM.

PR #4973, opened 2026-07-28
Optimize F3 tensor download throughput: pipelining, prefetch, and TCP_NODELAY

nvidianz:
## Summary

This PR optimizes the complete F3 tensor-download path from `main`, not just the final pipelining change. On the same 25 Gbit/s network and full 22.55 GiB GPT-J checkpoint, end-to-end throughput improves from roughly 250 MiB/s to 471–552 MiB/s:

| Mode | `main` | This PR | Improvement |
|---|---:|---:|---:|
| Memory (receiver reconstructs tensors in RAM) | 250.6 MiB/s | **471.5 MiB/s** | **1.88×** (+88.1%) |
| Disk offload (receiver writes safetensors) | 259.8 MiB/s | **551.6 MiB/s** | **2.12×** (+112.3%) |

Results are medians of three repetitions. The alias-heavy 455.29 MiB subset improves from 8.3 MiB/s to 441–554 MiB/s (**53–64×**) by eliminating duplicate downloads.

Raw F3 streaming already sustains about 2.6 GiB/s on this network (96.5% of the optimized raw-Python/TCP control), so the bottlenecks were in tensor decomposition, serialization, buffering, and Download Service request sequencing rather than the transport itself.

### 1. Eliminate duplicate downloads for aliased tensors

`ViaDownloaderDecomposer` previously used the current unique-item count to decide whether to register its post-processing callback. Multiple aliases of the first tensor could therefore register the callback repeatedly and create multiple full Download Service transactions for the same tensor set. The callback is now registered exactly once using an explicit first-item signal.

- `nvflare/fuel/utils/fobs/decomposers/via_downloader.py`

### 2. Overlap tensor serialization with transfer

`CacheableObject` now supports cheap item-size estimates and asynchronous item prefetch. `TensorDownloadable` uses raw tensor size to avoid synchronously serializing a look-ahead tensor only to discover that it does not fit in the current chunk. It prefetches up to two small `safetensors.save()` operations on the streaming worker pool, but never queues two oversized tensors together. Locking, duplicate suppression, cancellation, and release handling are preserved.

- `nvflare/fuel/f3/streaming/cacheable.py`
- `nvflare/app_opt/pt/tensor_downloader.py`

### 3. Pipeline Download Service requests and receiver consumption

`download_object()` previously ran stop-and-wait: request the next chunk, wait for the reply, consume it, then issue the next request. Consumers explicitly declaring value-stable state submit the next request immediately after a DATA reply and overlap it with `consumer.consume()` on a shared bounded request pool. State-transforming consumers retain stop-and-wait behavior, preserving the public `Consumer.consume()` contract without duplicate requests.

The pipeline snapshots state before submission to prevent in-place mutation races and retains the existing timeout retry, abort, progress, and receiver-confirmation behavior.

- `nvflare/fuel/f3/streaming/download_service.py`

### 4. Remove large FOBS buffer copies

FOBS-encoded streamed blobs now preserve their incoming network chunks instead of first copying the complete reply into a preallocated staging buffer. `BufferList.read_bytes()` joins multi-buffer ranges directly into the required immutable `bytes` result, avoiding the previous `bytearray` assembly followed by another full `bytes` copy. Consumed owned network chunks are released as forward-only FOBS parsing advances, and both declared-size and configured-size limits are enforced before excess data can accumulate. FOBS keeps its public section type stable for consumers such as safetensors.

- `nvflare/fuel/f3/streaming/blob_streamer.py`
- `nvflare/fuel/utils/buffer_list.py`
- `nvflare/fuel/utils/fobs/buf_list_stream.py`
- `nvflare/fuel/utils/fobs/lobs.py`

### 5. Reduce TCP request/reply latency and provide high-bandwidth tuning

TCP and STCP connections now enable `TCP_NODELAY` by default on both connected and accepted IPv4/IPv6 sockets, avoiding Nagle/delayed-ACK stalls for small request, final-chunk, and ACK frames. The behavior is configurable with `tcp_no_delay: false`.

Built-in flow-control defaults remain unchanged for mixed-version compatibility and bounded per-stream memory. The checked-in benchmark configuration provides opt-in 25 Gbit/s tuning that must be applied to both endpoints:

| Setting | Built-in default | Benchmark tuning |
|---|---:|---:|
| Streaming chunk | 1 MiB | 1 MiB |
| Streaming window | 16 MiB | 64 MiB |
| ACK interval | 4 MiB | 16 MiB |
| Retry pending-byte limit | 32 MiB | 128 MiB |

- `nvflare/fuel/f3/comm_config.py`
- `nvflare/fuel/f3/drivers/socket_conn.py`
- `nvflare/fuel/f3/streaming/byte_streamer.py`
- `nvflare/fuel/f3/streaming/byte_receiver.py`
- `docs/user_guide/admin_guide/configurations/communication_configuration.rst`

### 6. Add reproducible two-machine benchmark tooling

`dev_tools/f3` now contains:

- `tensor_download_bench.py`: end-to-end FOBS/TensorDecomposer/Download Service benchmark in memory and disk-offload modes, including alias-aware byte accounting, validation, repeated runs, RSS and disk reporting, cleanup checks, size-limited smoke tests, and F3 config loading.
- `cellnet_bench.py`: raw F3 reliable/unreliable streaming benchmark plus an optimized raw-TCP control, warm-up support, receiver reuse, throughput and line-utilization reporting, and IPv4/IPv6 handling.
- `comm_config.yml`: annotated 25 Gbit/s tuning example with binary-size parsing.
- `README.md`: setup, usage, controls, and configuration guidance.

### 7. Expand regression coverage

Tests cover:

- duplicate alias callback registration;
- tensor size estimation, depth-2 prefetch, future reuse, release, and cancellation;
- pipelined stable-state downloads, stop-and-wait state-transforming consumers, in-place mutation detection, retries, aborts, progress, EOF, producer errors, and consumer failures;
- chunk-preserving blob handling, declared-size enforcement, consumed-buffer release, and direct `BufferList.read_bytes()` behavior;
- TCP_NODELAY enable/disable and non-IP sockets;
- benchmark byte-size parsing, configuration validation, defaults, socket controls, and tensor-selection/validation behavior.

Changed and added tests span:

- `tests/unit_test/app_opt/pt/test_tensor_downloader.py`
- `tests/unit_test/fuel/f3/{cellnet_bench_test.py,tensor_download_bench_test.py}`
- `tests/unit_test/fuel/f3/drivers/socket_conn_nodelay_test.py`
- `tests/unit_test/fuel/f3/streaming/{blob_streamer_test.py,download_object_test.py}`
- `tests/unit_test/fuel/utils/buffer_list_test.py`
- `tests/unit_test/fuel/utils/fobs/{fobs_test.py,decomposers/via_downloader_test.py}`

## Validation

- Full 22.55 GiB GPT-J checkpoint: 3 repetitions in each of memory and disk-offload modes.
- Every benchmark run passed tensor count, unique count, key digest, shape, dtype, byte-size, and sampled SHA-256 validation.
- Raw transport controls: iperf3, optimized Python/TCP, and raw F3 reliable/unreliable streaming.
- 477-unit-test validation for the optimization set; the review-fix commit adds targeted regressions and passes 93 focused tests across Download Service, blob streaming, FOBS, tensor downloading, and benchmark tooling.
- Receiver cleanup verified: no residual per-run offload data and configured parent directories preserved.

## Memory results

Peak RSS deltas from the full 22.55 GiB checkpoint runs are shown alongside throughput. Main memory/disk values below are clean control runs; PR values are medians of three repetitions.

| Mode | Version | Throughput | Sender RSS delta | Receiver RSS delta |
|---|---|---:|---:|---:|
| Memory | `main` | 250.6 MiB/s | 1.88 GiB | 24.28 GiB |
| Memory | This PR | 471.5 MiB/s | 1.81 GiB | 23.24 GiB |
| Disk offload | `main` | 259.8 MiB/s | 1.87 GiB | 1.81 GiB |
| Disk offload | This PR | 551.6 MiB/s | 1.25 GiB | 0.00 MiB sampled delta |

The review fix also releases consumed FOBS chunks and prevents two oversized serialized tensors from being prefetched together, addressing the reported transient-memory amplification.

## Remaining limit

After these changes the end-to-end path is serialization-bound rather than network-bound. Sender-side `safetensors.save()` alone measures about 22.8 seconds for the full checkpoint, implying an approximately 1,011 MiB/s ceiling. Safetensors remains required for security; further work must retain the safetensors format and focus on reducing surrounding copies and improving overlap.


nvflare/fuel/f3/streaming/download_service.py
```diff
@@ -1855,130 +1856,177 @@ def _emit_progress(state: str, force: bool = False):
...
+                consumer.download_failed(ref_id, "download aborted")
+                _emit_progress("aborted", force=True)
+                return
 
-        # Update state for next request
-        current_state = new_state
+            # If consume returned a different state object, cancel the speculative request
+            # and relaunch with the correct state.  This is the slow path; for all current
+            # ItemConsumer subclasses `new_state is state` is always true.
+            if new_state is not state:
```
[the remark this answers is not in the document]
nvidianz: Fixed in 58d94dcae. The pipeline now compares returned state dictionaries by value, so a fresh but value-equivalent file-download state keeps the existing speculative request instead of launching a duplicate. Added a regression test that asserts the download completes with exactly two requests. Validation: 20 download-object tests and 67 adjacent streaming tests passed.

pcnudde (dismissed):

nvflare/fuel/f3/streaming/byte_receiver.py
```diff
@@ -39,7 +39,7 @@
 
 MAX_OUT_SEQ_CHUNKS = 16
 # 1/4 of the window size
-ACK_INTERVAL = 1024 * 1024 * 4
+ACK_INTERVAL = 1024 * 1024 * 16
```
YuanTingHsieh: [P1] **Derive the ACK default from the configured window**

Raising the unconditional default ACK interval to 16 MiB breaks existing configurations that set only `streaming_window_size` below 16 MiB. For example, with an 8 MiB window the sender reaches roughly 9 MiB and waits because its unacknowledged data exceeds the window, while the receiver will not ACK until 16 MiB; a non-final stream then times out. This configuration was valid with the previous 4 MiB default. Please derive or cap the default ACK interval from the configured window, or validate this relationship in the core configuration path.

nvflare/fuel/f3/streaming/download_service.py
```diff
@@ -1855,130 +1856,177 @@ def _emit_progress(state: str, force: bool = False):
...
+                    total_items = (total_items or 0) + len(data)
+            state = payload.get(_PropKey.STATE)
+
+            # Launch the next request BEFORE consuming current data so that the
+            # ~20 % consume cost overlaps with the following send_request.
+            # For ItemConsumer, consume() returns `state` unchanged, so launching
+            # with `state` is always correct.  In the rare case a consumer changes
+            # the state values, detect that and relaunch below.
+            current_state = state
+            _pending_future = _pipeline.submit(_do_request, current_state)
```
YuanTingHsieh: [P2] **Do not issue a speculative request before the consumer's state is known**

The public `Consumer.consume()` contract allows the consumer to return a modified state that is sent back to the producer. Here the next request is launched with the producer-returned state before `consume()` supplies that value. If the state changes, `_pending_future.cancel()` will normally return `False` because `send_request()` is already running, so the stale request still executes and the corrected request queues behind it. That can duplicate producer-side work or invoke a stateful `Downloadable` with the wrong state. Consider pipelining only consumers that explicitly guarantee state passthrough, or otherwise avoid issuing the request until consumption completes.

dev_tools/f3/cellnet_bench.py
```diff
@@ -0,0 +1,764 @@
...
+    try:
+        bytes_sent = future.result()
+        elapsed = time.perf_counter() - start
+        mem.stop()
+        print(
+            f"[send] RESULT reliable={effective_reliable}: sent {bytes_sent:,} bytes "
+            f"in {elapsed:,.2f} seconds ({rate(bytes_sent, elapsed)})"
+        )
+        print(f"[send] MEMORY reliable={effective_reliable}: {mem.summary()}")
+    finally:
```
YuanTingHsieh, line 619: [P2] **Always stop the sampler and cell when the sender fails**

If `send_stream()` or `future.result()` raises, this `finally` only stops progress reporting. `mem.stop()`, the settling delay, and `cell.stop()` are skipped because they occur outside the protected block, potentially leaving sampler/F3 worker threads and the cell running. Please put the sampler and cell cleanup in an outer `finally` that covers the complete sender lifecycle.

YuanTingHsieh (commented):
I found one backward-compatibility issue in the new flow-control defaults and two error/state-handling issues in the pipeline and benchmark tooling. Details are inline.

nvflare/fuel/f3/streaming/download_service.py
YuanTingHsieh: [P2] **Copy the state before handing it to the speculative request**

Related to, but separate from, the speculative-request ordering issue on this line: `_pipeline.submit(_do_request, current_state)` and the subsequent `consumer.consume(ref_id, state, data)` share the same dict object. A consumer that mutates `state` in place (a) races the worker thread, which serializes that dict inside `send_request()` — a torn read or `dictionary changed size during iteration` — and (b) silently defeats the `new_state != state` guard, because the returned object always compares equal to itself, so the corrective relaunch never fires and the producer sees whatever happened to be serialized. All in-tree consumers return the state unchanged (`ItemConsumer`) or build a fresh dict (`_ChunkConsumer`), but `Consumer` is a public ABC. Submitting `dict(state)` here and documenting that `consume()` must not mutate `state` in place closes this cheaply.

nvflare/fuel/f3/streaming/byte_receiver.py
YuanTingHsieh: Adding a second scenario that constrains the fix: version skew with pure defaults. An un-upgraded sender uses the previous 16 MiB default window while an upgraded receiver ACKs only at a 16 MiB lag, so the sender exhausts its entire window and stalls one RTT per 16 MiB — no explicit configuration involved, and unlike the small-window case it degrades silently rather than timing out. Deriving the default ACK interval from the receiver's own window does not cover this (the upgraded receiver's window is 64 MiB; the sender's window is not visible to it). That argues for keeping the receiver-side default at 4 MiB — ACK frames are small and now benefit from TCP_NODELAY, and the gains in this PR come from the window increase, pipelining, and prefetch rather than from fewer ACKs — or for documenting that mixed-version federations should pin `streaming_ack_interval: 4M` on upgraded receivers.

chesterxgchen:
Review: PR #4973 — Optimize F3 tensor download throughput

Scope: 26 files, +2567/−143. Six independent optimizations plus benchmark tooling. Reviewed at aa54c43dd.

Overview

Attacks the tensor-download path end to end: (1) fix duplicate downloads for aliased tensors, (2) sender-side prefetch of safetensors.save(), (3) pipeline Download Service requests against consumption, (4) preserve network chunks instead of copying into a staging buffer, (5) TCP_NODELAY + 4× larger streaming window, (6) dev_tools/f3 benchmarks.

I verified locally: all 73 tests in the changed test files pass, plus 30 bench tests; black/flake8 clean.

The via_downloader fix (#1) is the clear win — correct, minimal, and the 53–64× alias-subset number is credible. The other changes trade memory for throughput in ways the PR description doesn't quantify. Since you asked me to focus on memory, that's where most of this review sits.

---
High — verified defects

1. download_object pipeline issues a duplicate, undiscardable reges state

download_service.py:1999 submits the next request before consumeg_future.cancel() (lines 2010/2017/2024/2027) to undo it. Withmax_workers=1 and an idle worker, the task is picked up immediately, so cancel() returns False and the request is already on the wire.

Reproduced with a consumer returning a value-different state (5/5 trials, all on dl_pipe_0):

[(None), ({'start': 0, 'count': 1}), ({'start': 99, 'count': 1})]

Three requests where main sends two. The middle one transfers a full chunk that is silently discarded. Consequences:

- Bandwidth/CPU: a consumer that changes state on every chunk doubles both transfer and sender-side safetensors.save() work — the exact pathology fix #1  removes elsewhere.
- Cache accounting: the discarded request still runs CacheableObject._adjust_cache(start, count), double-incrementing the per-item receiver count. In a   multi-receiver transaction that evicts chunks other receivers stction.
- Contract: Consumer.consume's docstring says "Returns: new state to be sent back to the data owner." Both in-tree consumers happen to return value-equal state, so this is latent — but it's a public extension point.
                                                                                                                                                          The comment at line 1985 ("detect that and relaunch below") descan't deliver. Suggested fix: gate pipelining on an explicitopt-in, e.g. a Consumer.state_is_stable = False class attribute set to True on ItemConsumer and _ChunkConsumer, and fall back to stop-and-wait otherwise. The current new_state != state check is fine as a cheap assertiohanism.

Also note test_retry_resends_same_state was rewritten to drop itmer, so this case is no longer covered.
                                                                                                                                                        2. preserve_chunks removes the declared-size overrun guard on th
                                                                                                                                                        blob_streamer.py:85 — pre_allocated = self.size > 0 and not selfeserve_chunks keys on PAYLOAD_ENCODING == Encoding.FOBS, which isthe encoding for every non-bytes cellnet payload, essentially all object messages now take _append_dynamic_chunk, which only checks max_blob_size. The  declared size is checked once, at the end.
                                                                                                                                                        Verified — sender declares 4 bytes, streams 4 MiB:
                                                                                                                                                        preserve_chunks=False: buffered=4 bytes,         error=Buffer ov
preserve_chunks=True:  buffered=4,194,304 bytes, error=Size mismatch: declared 4 but received 4194304                                                   
The effective bound goes from declared size to streaming_max_blob_size (default 2,144,337,904 ≈ 2 GiB). A buggy or hostile peer can now make a receiver buffer ~2 GiB before the mismatch fires. One-line fix in _append

limit = blob_task.max_size
if blob_task.size > 0:
    limit = blob_task.size if limit <= 0 else min(limit, blob_ta

3. ~2× receiver peak RSS for large single-section FOBS payloads

Measured on a 512 MiB section (BufListStream.read of the whole t

┌──────────────────────────────────────────────────┬───────────┐
│                                                  │ peak RSS  │
├──────────────────────────────────────────────────┼───────────┤
│ main (pre-allocated blob → zero-copy memoryview) │   565 MiB │
├──────────────────────────────────────────────────┼───────────┤
│ this PR (chunk list → read_bytes join → bytes)   │ 1,087 MiB │
└──────────────────────────────────────────────────┴───────────┘

Two compounding causes: read_bytes joins into a new allocation wfully alive (BufListStream never releases consumed chunks), andlobs.py:201 now forces bytes(data).

For the tensor path each tensor is its own datum section, so the overshoot is bounded by one tensor and the PR's numbers still win. But this is a global
change to the FOBS receive path — a multi-GiB numpy/bytes datum sage pays full 2× peak. The mitigation is to have BufListStreamdrop buffers behind pos as sections are extracted (the stream is forward-only, so it's safe), which brings peak back to blob + largest section.

---
Medium

4. Prefetch depth 2 is unbounded by max_chunk_size (sender memor

cacheable.py:213 hard-codes depth 2; TensorDownloadable.prefetchtensors() per index. max_chunk_size defaults to 2 MiB, but asingle item always bypasses the size check (if not result), so a 1 GiB embedding tensor becomes one chunk — and two of them can be in flight in          _prefetch_futures on top of the current chunk. That's GiB-scale se_obj tensors are typically mmapped; the serialized copies arenot).                                                                                                                                                    
Suggest making the depth configurable and/or capping total prefetched bytes via get_item_size(), which you already have. The # Depth 1 ... left the largetensors partially un-hidden comment says the largest tensors areivated depth 2 — which is also exactly the case where the memorycost is worst.                                                                                                                                           
Minor related point: future.cancel() in release() is a no-op for an already-running save, so the serialized bytes survive until the pool task finishes.  
5. Quadrupling the global streaming window multiplies per-stream sender buffering                                                                        
STREAM_WINDOW_SIZE 16 → 64 MiB, and retry_max_pending_bytes defaults to 2 * window_size (byte_streamer.py:249) — so per reliable stream,                 buffered-for-retransmit memory goes 32 → 128 MiB. That's per conanning out to 100 clients goes from ~3.2 GB to ~12.8 GB ofretransmit buffer in the worst case.                                                                                                                     
The docs line ("the memory usage will be higher") is accurate but understated for a default change, and the PR description's table presents it as pure   upside. This was tuned on a 25 Gbit/s two-machine benchmark; I'dthe recommended comm_config.yml value (asdev_tools/f3/comm_config.yml already does) rather than the built-in default, or at minimum stating the concurrency math in the PR body and release notes.
6. Per-download thread and shutdown behavior                                                                                                             
_pipeline = ThreadPoolExecutor(max_workers=1, ...) is created per download_object call → one extra thread per concurrent download. The finally does      shutdown(wait=False, cancel_futures=True), so on an error/abort l be blocked in send_request for up to per_request_timeout;ThreadPoolExecutor threads are joined at interpreter exit, so a long timeout delays shutdown. Main kept that call on the caller's thread, where the callecontrolled it. Consider reusing stream_thread_pool (as tensor_doof a fresh executor.
                                                                                                                                                         Also: the first _pipeline.submit (line 1894) is outside the try,submit raises.
                                                                                                                                                         ---
Low / nits                                                                                                                                               
- read_bytes truncates silently when end exceeds total size — read() returned a zero-padded full-length buffer. _get_one_section's length check catches itoday; a ValueError would be safer for a new public method.
- read_bytes single-part path copies (parts[0].tobytes()), losing main's zero-copy fast path. Combined with bytes(data) in lobs this is consistent, but ia new copy on the previously-free path.
- New test-time dependencies. tests/unit_test/fuel/f3/tensor_download_bench_test.py imports dev_tools/f3/tensor_download_bench.py, which imports psutil, torch, and safetensors at module scope. psutil is not in setup.pct.toml — ./runtest.sh -u will fail collection on a minimal env.Wrap with pytest.importorskip. Also, tests for dev_tools/ code living under tests/unit_test/fuel/f3/ is misleading placement.                            - cellnet_bench.rss_bytes() reads /proc/self/statm — Linux only.the platform requirement.
- tensor_download_bench.py imports private _ENABLE_TENSOR_DISK_OFFLOAD / _TENSOR_DISK_OFFLOAD_ROOT_DIR from app_common. Fine for a dev tool, but it will silently break on rename — worth a comment or a public accessor.
- TCP_NODELAY default flip is a real behavior change for existing deployments, but it's guarded, documented, and defensible. The AF_INET/AF_INET6 check and OSError swallow are the right shape. No objection.
- BYTE_SIZE_PATTERN uses ^...$ anchors with .fullmatch() — redundant.


Requests before merge

1. Fix or gate the speculative-request path (#1) and restore a state-transforming-consumer test.
2. Restore the declared-size cap in _append_dynamic_chunk (#2).
3. Publish the RSS numbers. tensor_download_bench.py already collects sender and receiver peak RSS deltas, and the summary table in the PR body reports only throughput. Given #3, #4, and #5, the memory column is the missing half of the story — please add main vs. PR peak RSS for both modes on the 22.55 GiB checkpoint.
4. Decide whether the window/ACK increases ship as defaults or as recommended tuning (#5), and note the change in release notes either way.

chesterxgchen (changes requested):
very nice feature, but there are several bugs 

in particular, double the RSS memory

nvflare/fuel/f3/streaming/byte_receiver.py
nvidianz: Fixed in `69f807968`. The built-in flow-control defaults are restored to the compatible 16 MiB window / 4 MiB ACK interval. The 64 MiB / 16 MiB settings now remain opt-in benchmark tuning that must be applied to both endpoints. The admin guide and 2.9 release notes document the compatibility and memory decision, and the default-value regression test was updated. Validation: 93 focused tests passed; black, isort, flake8, and `git diff --check` passed.

nvflare/fuel/f3/streaming/download_service.py
nvidianz: Fixed in `69f807968`. `Consumer` now defaults to `supports_pipelining = False`; only value-stable `ItemConsumer` opts in. State-transforming consumers use stop-and-wait and send the state returned by `consume()` without a speculative stale request. Added a regression that verifies the transformed state is used in exactly the next request. Pipelined requests also use a shared bounded pool instead of one executor per download. Validation: 93 focused tests passed; scoped style checks passed.

dev_tools/f3/cellnet_bench.py
nvidianz: Fixed in `69f807968`. `run_sender()` now wraps the complete cell/sampler lifecycle in an outer `try/finally`: progress reporting is stopped, the sampler is stopped exactly once, the ACK settling delay runs, and the cell is stopped even when `send_stream()` or `future.result()` raises. Added a failure-path regression asserting sampler and cell cleanup. Validation: 93 focused tests passed; scoped style checks passed.

nvflare/fuel/f3/streaming/download_service.py
nvidianz: Fixed in `69f807968`. Pipelined consumers now submit a `dict(state)` snapshot, and the `Consumer.consume()` contract explicitly forbids in-place mutation. A defensive value check catches an opted-in consumer that still mutates or transforms state without launching a second corrected request. Added an in-place mutation regression alongside the stop-and-wait transforming-consumer test. Validation: 93 focused tests passed; scoped style checks passed.

nvidianz:
Addressed the review in `69f807968`:

- Gated speculative download requests behind explicit stable-state opt-in; state-transforming consumers now use stop-and-wait. Added transforming-state and in-place-mutation regressions.
- Snapshots state before pipelined submission and documents that `Consumer.consume()` must not mutate state in place.
- Restored the preserve-chunks declared-size cap before excess buffering.
- Added forward-only release of consumed owned FOBS network chunks while preserving ordinary caller-owned lists.
- Bounded prefetch so two oversized serialized tensors are never queued together.
- Restored built-in 16 MiB window / 4 MiB ACK defaults. The 64 MiB / 16 MiB settings remain opt-in 25 Gbit/s tuning; admin docs and 2.9 release notes now explain this.
- Replaced per-download executors with a shared bounded request pool.
- `BufferList.read_bytes()` now rejects ranges past available data.
- Fixed benchmark sampler/cell cleanup on failure and documented Linux `/proc` RSS reporting.
- Removed redundant regex anchors and documented the intentional repository-internal disk-offload imports. `psutil` is already a declared dependency in `setup.cfg`.
- Updated the PR description with main-versus-PR RSS deltas and clarified that safetensors remains required for security.

Validation with `~/venv/3.12/bin/python`: 93 focused tests passed. Black, isort, flake8, license checks, and `git diff --check` passed.

pcnudde:
How much memory are we talking about. Can we really not have 1 setting that workx for most network links ?

nvidianz:
@pcnudde Yes—we can keep one setting that works for most network links. The PR has restored the built-in defaults to a **16 MiB streaming window / 4 MiB ACK interval**. The **64 MiB / 16 MiB** values are now only opt-in benchmark tuning for high-bandwidth/high-BDP links (and must be applied to both endpoints).

On memory: the changes do **not** double memory usage for tensor downloads overall. For the full 22.55 GiB GPT-J checkpoint, measured receiver peak RSS deltas were:

- Memory mode: `main` **24.28 GiB** → this PR **23.24 GiB**
- Disk-offload mode: `main` **1.81 GiB** → this PR **0.00 MiB sampled delta**

The ~2× result in the review applies to a specific edge case: one very large, single-section FOBS payload. In that path, the receiver temporarily holds both the original network chunks and the contiguous `bytes` allocation created by joining them. A 512 MiB section can therefore transiently approach 1 GiB plus process overhead. For tensor downloads, each tensor is a separate FOBS datum section, consumed chunks are released as parsing advances, and the transient overhead is bounded by the largest individual serialized tensor—not the whole model.

The window setting has separate bounded memory implications: the retry pending-byte limit defaults to 2× the window, so it is **32 MiB per active reliable stream** with the general default versus **128 MiB** with the 25 Gbit/s tuning. That is why the larger value should remain opt-in rather than becoming the universal default.

chesterxgchen (approved):

PR #4979, opened 2026-07-28
[FLARE-3030] Prevent secret leaks in CLI errors

pcnudde:
## What changed

- Redact complete single-quoted and double-quoted sensitive values, including values containing whitespace.
- Recognize sensitive keys embedded in JSON-shaped text.
- Apply the same quoted-value handling to sensitive CLI options and Authorization values.
- Stop echoing rejected `KUBECTL` environment values in validation errors.

## Why

The text redactor previously stopped at the first whitespace and expected `:` or `=` immediately after a sensitive key. Multi-word credentials were therefore only partially redacted, while quoted JSON keys such as `"client_secret"` were not matched at all. The Kubernetes staging validation path exposed the problem by including the rejected environment value in its error message.

## Impact

Credential-like values are fully removed from CLI messages before human, JSON, or JSONL output is emitted. Invalid Kubernetes command values are no longer reflected in error output, including opaque values that cannot be recognized as secrets by a pattern-based redactor.


nvflare/tool/cli_output.py
```diff
@@ -70,22 +70,27 @@
...
     r"access[-_]?token|refresh[-_]?token|id[-_]?token|session[-_]?token|auth[-_]?token|"
     r"api[-_]?key|private[-_]?key|secret[-_]?key|client[-_]?secret"
-    r")(?:\s+|=))([\"']?)([^\"'\s,;]+)([\"']?)"
+    r")(?:\s+|=))"
+    rf"(?P<value>{_SENSITIVE_TEXT_VALUE_PATTERN})"
 )
 _BEARER_TOKEN_PATTERN = re.compile(r"(?i)\b(authorization\s*:\s*bearer\s+)([A-Za-z0-9._~+/=-]+)")
-_AUTH_VALUE_PATTERN = re.compile(r"(?i)\b(authorization\s*[:=]\s*)(?!bearer\s+)([\"']?)([^\"'\s,;]+)([\"']?)")
+_AUTH_VALUE_PATTERN = re.compile(
+    rf"(?i)(?P<prefix>\bauthorization\s*[:=]\s*)(?!bearer\s+)" rf"(?P<value>{_SENSITIVE_TEXT_VALUE_PATTERN})"
```
YuanTingHsieh: Could we handle standard unquoted non-Bearer Authorization headers as well?

```python
sanitize_cli_output("Authorization: Basic dXNlcjpwYXNz")
# => "Authorization: <redacted> dXNlcjpwYXNz"
```

Only the scheme is redacted; the Base64 credential remains visible in CLI output. The same pattern also leaves Digest credentials behind. Since this PR is intended to prevent secret leakage, please redact the full non-Bearer Authorization value (or preserve the scheme while redacting all credential material) and add a regression test for this form.

YuanTingHsieh (commented):
One Authorization redaction case remains; concrete reproduction inline.

nvflare/tool/cli_output.py
YuanTingHsieh: Same pattern has a second gap: JSON/dict-shaped `Authorization` headers are not redacted at all — on either layer.

```python
sanitize_cli_output('{"Authorization": "Bearer sk-live-abc123"}')
# => '{"Authorization": "Bearer sk-live-abc123"}'

sanitize_cli_output({"headers": {"Authorization": "Bearer sk-live-abc123"}})
# => {"headers": {"Authorization": "Bearer sk-live-abc123"}}
```

Text path: `_AUTH_VALUE_PATTERN` and `_BEARER_TOKEN_PATTERN` expect `[:=]` right after the key, so the closing quote of a JSON key blocks the match — this PR added the `["']?` key handling to `_SENSITIVE_ASSIGNMENT_PATTERN` only. Dict path: `authorization` is not in `_SENSITIVE_KEY_NAMES`, and the bare value has no `Authorization:` prefix for the text patterns to anchor on. A request-headers dict attached to an HTTP error message hits both shapes, so there is no defense-in-depth for this key.

Minimal fix (verified locally), and it composes with the non-Bearer scheme fix above:

- `_AUTH_VALUE_PATTERN`: allow an optional quote after the key — `\bauthorization["']?\s*[:=]\s*`
- add `"authorization"` to `_SENSITIVE_KEY_NAMES`

Please add regression tests for both the string and dict forms.

pcnudde: Addressed both cases in 395f8ce99. Unquoted non-Bearer Authorization values are now redacted through the end of the header line (covering Basic and Digest), quoted JSON Authorization keys are recognized by both Authorization patterns, and structured `authorization` fields are redacted recursively. Added regressions for the Basic string, JSON string, and nested dict forms. Validation: 173 relevant tests passed and `./runtest.sh -s` passed.

YuanTingHsieh (approved):

PR #4988, opened 2026-07-29
Fix Collab call identity contracts

nvidianz:
## What changed

- Normalize async callback caller identity to the logical remote site while retaining the fully qualified target for routing.
- Use the fully qualified target name as the callee for named-object dispatch.
- Preserve the fully qualified target on `CollabCallError` and in its diagnostic message while keeping the site-only compatibility field.
- Fail secure Collab calls immediately when the local Cell lacks certificate-backed secure-message support.
- Add regression coverage for all four behaviors.

## Why

Collab call contexts and errors were inconsistently collapsing or expanding target identities. In addition, secure calls from a non-secure simulation environment attempted certificate exchange and surfaced as a timeout instead of reporting the invalid precondition immediately.

## Impact

Callbacks, named-object handlers, and errors now expose consistent caller/callee/target identities. Invalid secure-call configurations fail locally with an actionable error rather than waiting for the remote-call timeout.

## Validation

- `PYTEST_DISABLE_PLUGIN_AUTOLOAD=1 ~/venv/3.12/bin/python -m pytest -q tests/unit_test/collab tests/unit_test/fuel/f3/cellnet/core_cell_test.py` — 118 passed
- Black, isort, and flake8 checks on all changed files
- License-header validation for `nvflare` and `tests`
- `git diff --check`

### DIFF
--- a/nvflare/collab/api/exceptions.py
+++ b/nvflare/collab/api/exceptions.py
@@ -30,9 +30,11 @@ class CollabCallError(Exception):
         cause_type: str = None,
         remote_traceback: str = None,
     ):
+        self.target = site
+        self.target_name = site
         self.site = site.split(".", 1)[0]
         self.func_name = func_name
         self.cause = cause
         self.cause_type = cause_type or type(cause).__name__
         self.remote_traceback = remote_traceback
-        super().__init__(f"call to {self.site}.{func_name} failed: {cause}")
+        super().__init__(f"call to {self.target_name}.{func_name} failed: {cause}")
--- a/nvflare/collab/api/exceptions.py
+++ b/nvflare/collab/api/exceptions.py
@@ -30,7 +30,12 @@ class CollabCallError(Exception):
         cause_type: str = None,
         remote_traceback: str = None,
     ):
-        self.target = site
+        """Create an error for a failed Collab call.
+
+        Args:
+            site: The logical site name or fully qualified ``<site>.<object>`` target.
+                The legacy ``site`` attribute always contains only the site prefix.
+        """
         self.target_name = site
         self.site = site.split(".", 1)[0]
         self.func_name = func_name
--- a/nvflare/collab/api/group_call_context.py
+++ b/nvflare/collab/api/group_call_context.py
@@ -285,7 +285,10 @@ class GroupCallContext:
 
             # swap caller/callee
             original_caller = ctx.caller
-            ctx.caller = ctx.callee
+            # The callback observes the remote logical site as its caller.
+            # Keep self.target_name fully qualified for routing and diagnostics,
+            # but do not expose the target object suffix as caller identity.
+            ctx.caller = ResultWaiter._get_site_name(self.target_name)
             ctx.callee = original_caller
 
             if not isinstance(result, Exception):
--- a/nvflare/collab/api/group_call_context.py
+++ b/nvflare/collab/api/group_call_context.py
@@ -28,6 +28,11 @@ _SHORT_WAIT = 1.0
 _FAILURE_MARKER = object()
 
 
+def _get_site_name(target_name: str):
+    # target_name is either <site_name> or <site_name>.<obj_name>
+    return target_name.split(".", 1)[0]
+
+
 class ResultQueue:
     def __init__(self, limit: int):
         if limit <= 0:
@@ -119,7 +124,7 @@ class ResultWaiter(threading.Event):
     def __init__(self, sites: list[str]):
         super().__init__()
         self.sites = sites
-        self._expected_sites = {self._get_site_name(site) for site in sites}
+        self._expected_sites = {_get_site_name(site) for site in sites}
         if len(self._expected_sites) != len(sites):
             raise ValueError(f"sites must be unique but got {sites}")
         self._received_sites = set()
@@ -178,12 +183,6 @@ class ResultWaiter(threading.Event):
             if done:
                 break
 
-    @staticmethod
-    def _get_site_name(target_name: str):
-        # target_name is either <site_name> or <site_name>.<obj_name>
-        parts = target_name.split(".")
-        return parts[0]
-
     def set_result(self, target_name: str, result):
         return self._set_outcome(target_name, result=result)
 
@@ -191,7 +190,7 @@ class ResultWaiter(threading.Event):
         return self._set_outcome(target_name, error=error)
 
     def _set_outcome(self, target_name: str, result=None, error: CollabCallError = None):
-        site_name = self._get_site_name(target_name)
+        site_name = _get_site_name(target_name)
         with self._result_lock:
             if site_name not in self._expected_sites:
                 self.logger.warning(f"ignored result from unexpected site '{site_name}'")
@@ -210,7 +209,7 @@ class ResultWaiter(threading.Event):
             return True
 
     def add_partial_result(self, target_name: str, partial_result):
-        site_name = self._get_site_name(target_name)
+        site_name = _get_site_name(target_name)
         self.results.append((site_name, partial_result), is_whole=False)
 
 
@@ -288,7 +287,7 @@ class GroupCallContext:
             # The callback observes the remote logical site as its caller.
             # Keep self.target_name fully qualified for routing and diagnostics,
             # but do not expose the target object suffix as caller identity.
-            ctx.caller = ResultWaiter._get_site_name(self.target_name)
+            ctx.caller = _get_site_name(self.target_name)
             ctx.callee = original_caller
 
             if not isinstance(result, Exception):
--- a/nvflare/collab/runtime/cell_dispatcher.py
+++ b/nvflare/collab/runtime/cell_dispatcher.py
@@ -47,6 +47,9 @@ class CellDispatcher(_InvocationDispatcher):
     ):
         set_call_context(context)
 
+        if call_opt.secure and not self.cell.core_cell.supports_secure_messages():
+            raise RuntimeError("secure Collab call requires a Cell configured with certificates")
+
         payload = {
             ObjectCallKey.CALLER: self.caller,
             ObjectCallKey.TARGET_NAME: target_name,
--- a/nvflare/collab/runtime/cell_dispatcher.py
+++ b/nvflare/collab/runtime/cell_dispatcher.py
@@ -36,6 +36,10 @@ class CellDispatcher(_InvocationDispatcher):
         self.target_fqcn = target_fqcn
         self.thread_executor = thread_executor
 
+    def _check_secure_call_supported(self, call_opt: CallOption):
+        if call_opt.secure and not self.cell.core_cell.supports_secure_messages():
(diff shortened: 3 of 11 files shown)


pcnudde:
The callback caller normalization, the `CollabCallError` fields, `supports_secure_messages`, and the tests all look good. Three issues found; I'd like the first two fixed in this PR:

1. **The new `callee` identity is wire-controlled** (`nvflare/collab/runtime/dispatch.py:68`). `_preprocess` now copies the payload `TARGET_NAME` into `ctx.callee` verbatim, while `_call_app_method` only checks it is a str and uses `parts[1]` for object lookup. A participant routed to `site-1` can send `TARGET_NAME="server.trainer.extra"`, still invoke the local `trainer`, and the handler sees the forged callee. Before this change callee was locally derived from `app.name`, so this is a regression in trust properties, and the facade documents callee as the fully qualified name of the invoked object. Please derive it locally from the resolved object (`app.name` or `f"{app.name}.{obj_name}"`) and consider rejecting a site prefix that does not match `app.name`.

2. **Secure fire-and-forget calls still report success** (`nvflare/collab/runtime/cell_dispatcher.py:50`). The new guard runs inside the submitted worker. With `expect_result=False`, `call_target()` returns `None` before the worker raises; the `RuntimeError` goes to `gcc.set_exception` -> `waiter.set_failure` on a waiter nobody reads, with no logging. The PR description says invalid secure configs fail locally with an actionable error, but this path is silent. Please hoist the check so it runs synchronously before dispatch (before the executor submit or the fire-and-forget early return).

3. **Local pre-dispatch failures keep a site-only target** (`nvflare/collab/api/proxy.py:237`). Binding and lookup failures are wrapped with `self.name` (the parent proxy), so for `proxy(target="trainer").method(...)` the new `error.target_name` is `site-1` rather than `site-1.trainer`, even though the child was already resolved as `p`. Remote failures are unaffected. Minor, but since this area is being touched: use `p.target_name` in the wrapper and add a local-failure regression test.

holgerroth (commented):
Ran the touched suites at the PR head (`tests/unit_test/collab` + `tests/unit_test/fuel/f3/cellnet/core_cell_test.py`): 118 passed. The identity fixes line up end to end — child proxies are built as `{site}.{obj}`, `ResultWaiter` keys outcomes by site name, and the new secure guard matches exactly the condition that makes `CredentialManager.encrypt/decrypt` fail.

A few non-blocking suggestions inline.

nvflare/collab/api/exceptions.py
```diff
@@ -30,9 +30,11 @@ def __init__(
         cause_type: str = None,
         remote_traceback: str = None,
     ):
+        self.target = site
+        self.target_name = site
```
holgerroth: `target` and `target_name` are set to the same value, and nothing in production code consumes either yet — two synonymous public fields on an exception invite drift. Suggest keeping just `target_name` (matches `GroupCallContext.target_name` / `ObjectCallKey.TARGET_NAME`), unless there's a known consumer of `.target`.

Also, the `site` parameter now semantically receives a fully qualified target name (`<site>.<obj>`); renaming it would break keyword callers, but a docstring note on the constructor would help.

nvflare/collab/runtime/cell_dispatcher.py
```diff
@@ -47,6 +47,9 @@ def _call_target(
     ):
         set_call_context(context)
 
+        if call_opt.secure and not self.cell.core_cell.supports_secure_messages():
```
holgerroth: This fail-fast only surfaces for `expect_result=True` calls. With `expect_result=False`, `call_target()` returns `None` before the executor runs `_run_func`; the `RuntimeError` raised here lands in `gcc.set_exception()` → `waiter.results.failures`, which nobody reads and which logs nothing — so a fire-and-forget secure call on a non-secure cell still fails silently.

That's the pre-existing pattern for all fire-and-forget errors, but this particular error is a purely local precondition, so it could be checked synchronously (e.g., in `call_target_in_group()` before submitting to the executor) to make fire-and-forget misconfigurations fail loudly too. At minimum, a log line in `_run_func`'s exception path would help.

nvflare/collab/runtime/dispatch.py
```diff
@@ -65,7 +65,7 @@ def _error_reply(error: str, logger, error_type: str = None, traceback_text: str
 
 
 def _preprocess(app: App, caller, target_obj_name, target_name, func_name, func, args, kwargs):
-    ctx = app.new_context(caller=caller, callee=app.name)
+    ctx = app.new_context(caller=caller, callee=target_name)
```
holgerroth: `target_name` comes from the remote payload, and `_call_app_method` never validates that its site prefix matches `app.name` — so a buggy or misbehaving peer could label this context with an arbitrary callee (e.g. `"other-site.trainer"`) while executing on this app. Previously `callee=app.name` was locally authoritative.

Deriving the callee locally keeps the new fully-qualified behavior for well-formed calls while staying authoritative:

```suggestion
    callee = f"{app.name}.{target_obj_name}" if target_obj_name else app.name
    ctx = app.new_context(caller=caller, callee=callee)
```

Consistent with the existing trust model (payload `caller` is already trusted as-is), and these fields aren't used for authorization today — so this is hardening, not a bug.

nvflare/collab/api/group_call_context.py
```diff
@@ -285,7 +285,10 @@ def set_result(self, result):
 
             # swap caller/callee
             original_caller = ctx.caller
-            ctx.caller = ctx.callee
+            # The callback observes the remote logical site as its caller.
+            # Keep self.target_name fully qualified for routing and diagnostics,
+            # but do not expose the target object suffix as caller identity.
+            ctx.caller = ResultWaiter._get_site_name(self.target_name)
```
holgerroth: Nit: `_get_site_name` is private-by-convention on `ResultWaiter`, and this is now a cross-class call. Same module so it works, but promoting it to a module-level helper (used by both `ResultWaiter` and `GroupCallContext`) would be cleaner.

nvflare/collab/api/exceptions.py
nvidianz: Fixed in `56c2cf1b6`. Removed the redundant `target` alias and retained `target_name` as the fully qualified diagnostic field. Added constructor documentation clarifying that the legacy `site` argument may receive `<site>.<object>` while `.site` remains the site prefix. The related assertion now verifies `site` and `target_name` only.

nvflare/collab/runtime/cell_dispatcher.py
nvidianz: Fixed in `56c2cf1b6`. The secure-message capability check now runs synchronously in `call_target_in_group()` before executor submission, while `_call_target()` retains the defensive check. Added a regression using the public fire-and-forget path (`expect_result=False`) that verifies the `RuntimeError` is raised and neither the executor nor Cell send APIs are invoked.

nvflare/collab/runtime/dispatch.py
nvidianz: Fixed in `56c2cf1b6`. `_preprocess()` now derives the callee from the local `app.name` and resolved object name. The regression deliberately supplies `other-site.trainer` on the wire and verifies the invoked handler still observes the locally authoritative `site-1.trainer` callee.

nvflare/collab/api/group_call_context.py
nvidianz: Fixed in `56c2cf1b6`. Promoted `_get_site_name()` to a module-level helper and now use it consistently from `ResultWaiter` initialization/outcome handling and `GroupCallContext` callback context normalization.

nvidianz:
@pcnudde Addressed all three findings:

1. `56c2cf1b6` derives the callee locally from `app.name` plus the resolved object name. A regression supplies a mismatched wire target (`other-site.trainer`) and verifies the handler sees the authoritative local callee (`site-1.trainer`).
2. `56c2cf1b6` moves the secure-message capability check before executor submission. Fire-and-forget secure calls on non-secure Cells now raise synchronously; the regression verifies no executor or Cell send API is invoked.
3. `e1ff3f942` preserves the resolved child target for local pre-dispatch failures. A binding failure through `proxy(target="trainer")` now reports `target_name == "site-1.trainer"` while retaining `site == "site-1"`.

Validation with `~/venv/3.12/bin/python`: 119 tests passed across `tests/unit_test/collab` and `core_cell_test.py`. Scoped Black, isort, flake8, and `git diff --check` passed.

nvflare/collab/api/exceptions.py
holgerroth: Verified in 56c2cf1b6 — `target` alias removed, `target_name` retained as the fully qualified field, and the constructor docstring now covers the legacy `site` semantics. Resolved from my side.

nvflare/collab/runtime/cell_dispatcher.py
holgerroth: Verified in 56c2cf1b6 — the synchronous check in `call_target_in_group()` covers fire-and-forget before executor submission, and the upgraded test (public `call_target()` with `expect_result=False`, asserting no `submit`/`send_request`/`fire_and_forget`) pins it well.

One residual, non-blocking: in the *group* path, `_dispatch_work_items()` catches the synchronous raise per item and records it in the waiter, so a group fire-and-forget secure call on a non-secure cell is still silent (`expect_result=False` callers never read `results.failures`). Fixing that would mean validating call options at the group layer before dispatch — fine as a follow-up.

nvflare/collab/runtime/dispatch.py
holgerroth: Verified in 56c2cf1b6 — callee is derived locally now, and the regression test putting `other-site.trainer` on the wire while asserting `site-1.trainer` proves local authority rather than just exercising the happy path. Dropping the now-unused `target_name` parameter is a nice cleanup.

nvflare/collab/api/group_call_context.py
holgerroth: Verified in 56c2cf1b6 — module-level `_get_site_name()` used consistently from both `ResultWaiter` and `GroupCallContext`.

holgerroth (approved):
All review feedback addressed in 56c2cf1b6 (verified each fix against the diff), and e1ff3f942's child-target fix is a good catch beyond the original scope: local failures on `proxy(target=...)` calls now carry the fully qualified target. Re-ran `tests/unit_test/collab` + `tests/unit_test/fuel/f3/cellnet/core_cell_test.py` at e1ff3f942: 119 passed.

Two minor residuals for the record, neither blocking and neither worse than pre-PR behavior:
- Group fire-and-forget secure calls are still silent — `_dispatch_work_items()` records the per-item failure in a waiter that `expect_result=False` callers never read (details in the `cell_dispatcher.py` thread).
- A binding failure during *implicit* child resolution (function found on a child without explicit `target=`) still reports the outer proxy name, since `p` is only rebound after `adjust_func_args()` returns.

PR #4989, opened 2026-07-29
Add Client API attach execution mode

YuanTingHsieh:
## What changed

- Implement `ClientAPIExecutor(execution_mode="attach")` for externally owned trainers using the standard `flare.init()` / `receive()` / `send()` API.
- Add a non-owning `AttachBackend` and extract shared Cell protocol behavior into `CellBackendBase` while preserving external-process lifecycle behavior.
- Add IPCAgent-style rendezvous using a pre-provisioned `attach_id`, a job-child trainer FQCN, and CJ-initiated `SESSION_OPEN` handshake.
- Add task/result attempt tracking, status recovery, deduplication, heartbeat/reconnect handling, and receiver-confirmed lazy-result lifetime management.
- Enforce transport policy: production network Attach requires mTLS; bare-CA TLS is rejected; every clear network listener, including loopback, requires `allow_insecure_attach`, and the trainer profile must explicitly declare `connection_security="clear"`; shared-file Attach uses its protected filesystem boundary.
- Add the attach connection-profile schema, design document, NumPy example, and focused unit coverage.

## Why

Externally started trainers currently need the legacy `IPCAgent` / `IPCExchanger` programming model. Attach mode lets the same trainer code use the ordinary Client API regardless of whether NVFlare launches the process, while reusing the Cell/FOBS data path introduced by #4906.

This implementation deliberately keeps process ownership outside NVFlare and makes no changes under `nvflare/fuel/f3/**`.

## Impact

Operators configure a dedicated CJ-owned listener in site-local communication settings and give the trainer the same non-secret `attach_id` configured in the job. A direct network profile also contains the job-specific CJ FQCN and URL, so it is normally generated after job submission. A shared-file profile instead uses a stable rendezvous directory and permits either the trainer or job to start first. Job shutdown stops only the Attach session—not the externally managed process.

## Validation

- 379 targeted attach, external-process, in-process, bootstrap, and executor tests passed after rebasing onto latest `main`.
- `./runtest.sh -s` passed (Black, isort, flake8, and agent-skill lint).
- `git diff --check` passed.


nvflare/client/cell/api.py
```diff
@@ -597,10 +636,16 @@ def clear(self):
         self._fl_model = None
         self._receive_called = False
         self._current_task = None
+        if self._attach:
+            self._attach.clear_result()
         self._result_receiver_ids = None
 
     def shutdown(self):
         """Stop this one-session trainer and its process-global F3 runtime."""
```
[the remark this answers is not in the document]
YuanTingHsieh: Fixed in d14d88a40. `shutdown()` now documents the mode-dependent ownership: external-process mode shuts down its dedicated process-global F3 runtime, while Attach stops only its session Cell and preserves shared process-global services.

nvflare/client/cell/api.py
```diff
@@ -628,16 +674,23 @@ def _handle_task_ready(self, request):
         if reject_reason:
             return self._reply(Topic.TASK_FAILED, **{MsgKey.REASON: reject_reason})
         task_id = payload.get(MsgKey.TASK_ID)
+        attempt_id = payload.get(MsgKey.ATTEMPT_ID)
+        if self._attach:
+            duplicate_reply = self._attach.reserve_task(task_id, attempt_id)
+            if duplicate_reply is not None:
+                return duplicate_reply
```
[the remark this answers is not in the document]
YuanTingHsieh: Fixed in d14d88a40. `_note_cj_activity()` now runs immediately after authenticated CJ-control validation and before Attach deduplication can return an idempotent reply. The duplicate-task regression clears `_last_cj_activity`, delivers a duplicate `TASK_READY`, and verifies that activity is refreshed.

nvflare/client/cell/bootstrap.py
```diff
@@ -97,21 +110,66 @@ def get_bootstrap_client_api_type(config: dict, path: str = "<bootstrap config>"
...
+                f"{BootstrapKey.CA_CERT!r} must be a non-empty string"
+            )
+        if connection_security != ConnectionSecurity.CLEAR and not ca_cert:
+            raise ValueError(
+                f"invalid Client API bootstrap config {path}: "
+                f"{connection_security!r} attach requires field {BootstrapKey.CA_CERT!r}"
+            )
+        if BootstrapKey.SECURE_MODE in config:
+            expected_secure = connection_security != ConnectionSecurity.CLEAR
+            if config[BootstrapKey.SECURE_MODE] is not expected_secure:
```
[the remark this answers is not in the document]
YuanTingHsieh: Fixed in d14d88a40. `secure_mode` is now type-validated before the Attach security-consistency check, and the validation test covers both external-process and Attach profiles.

nvflare/app_common/executors/client_api/attach_backend.py
```diff
@@ -0,0 +1,504 @@
...
+            deadline = (
+                None
+                if self._context.result_wait_timeout is None
+                else time.monotonic() + self._context.result_wait_timeout
+            )
+            while not task.result_ready.wait(_RESULT_POLL_INTERVAL):
+                if abort_signal.triggered:
+                    self._send_abort(session, f"task {task_name!r} aborted")
+                    return make_reply(ReturnCode.TASK_ABORTED)
+                if not self._session_matches(session.session_id):
```
holgerroth, line 507: `finalize()` sets `_closed` and stops the session-monitor thread — the only thing that can set `session.error` on heartbeat loss — and once `_closed` is set, `_handle_result_ready` rejects any late result, so `result_ready` can never fire. That leaves this wait loop with no exit besides the task abort signal when `result_wait_timeout=None` (a configuration the initializer explicitly allows while `heartbeat_timeout > 0`). It's a narrow window since the framework normally aborts tasks before END_RUN, but `_wait_for_session` already checks `self._closed` on every poll; doing the same here closes the gap:

```suggestion
                if self._closed or not self._session_matches(session.session_id):
```

A regression test along these lines hangs on the current code and passes with the change (note the overrides — the test file's `_context()` defaults use `heartbeat_timeout=0.0`, which `initialize()` rejects when `result_wait_timeout` is None):

```python
def test_finalize_unblocks_pending_result_wait():
    cell = FakeCell()
    cell.deliver_result = False
    backend = AttachBackend()
    fl_ctx = _fl_ctx(cell)
    backend.initialize(_context(result_wait_timeout=None, heartbeat_timeout=30.0), fl_ctx)
    _wait_ready(backend)

    result_box = {}
    execution = threading.Thread(
        target=lambda: result_box.setdefault("result", backend.execute("train", Shareable(), fl_ctx, Signal()))
    )
    execution.start()
    assert _wait_until(lambda: cell.task_ready_count == 1)

    backend.finalize(fl_ctx)
    execution.join(timeout=2.0)

    assert not execution.is_alive()
    assert result_box["result"].get_return_code() == ReturnCode.EXECUTION_EXCEPTION
```

nvflare/app_common/executors/client_api/cell_backend.py
```diff
@@ -0,0 +1,362 @@
...
+
+    @staticmethod
+    def _protocol_reply(reply_topic: str, **fields):
+        body = {MsgKey.REPLY_TOPIC: reply_topic}
+        body.update(fields)
+        return make_cell_reply(CellReturnCode.OK, body=body)
+
+    # Attach originally used ``_reply``; retain the short internal spelling.
+    _reply = _protocol_reply
+
```
holgerroth: Nothing on the CJ side calls `self._reply` — the trainer-side `CellClientAPI._reply` is a separate method — so this alias is dead code, and the comment records development history rather than something the next reader needs. (Unlike `_TaskContext = CellTask` in `external_process_backend.py`, which the existing tests do reference.)

```suggestion
```

holgerroth (commented):
Two small suggestions on the attach backend and the shared Cell base — details inline.

nvflare/client/cell/attach_session.py
```diff
@@ -0,0 +1,456 @@
...
+                    )
+                except Exception as e:
+                    return api._reply(
+                        Topic.SESSION_REJECTED,
+                        **{
+                            MsgKey.SESSION_ID: session_id,
+                            MsgKey.REASON: f"failed to configure trainer runtime: {e}",
+                        },
+                    )
+                api._cj_fqcn = origin
```
pcnudde, line 551: [P1] Bind the opening origin to the configured site/job before accepting it. `_validate_open` only requires a nonempty origin and checks attacker-supplied payload fields, then this assignment makes that first origin authoritative. A custom job component on another federated site that can route to this stable trainer FQCN and knows the non-secret `attach_id` can send the expected site/trainer values, win `SESSION_OPEN`, and subsequently inject tasks or receive the trainer's results under `_validate_cj_control`. I reproduced a site-B job Cell receiving `SESSION_ACCEPTED` from a site-A trainer. Validate that the authenticated origin is the expected CJ under `api._site_name` and that its job component agrees with a valid `JOB_ID` before setting `_cj_fqcn`; add a two-site negative integration test.

nvflare/app_common/executors/client_api/attach_backend.py
```diff
@@ -0,0 +1,504 @@
...
+            return
+        topic = body.get(MsgKey.REPLY_TOPIC)
+        if topic == Topic.SESSION_REJECTED:
+            session.error = str(body.get(MsgKey.REASON) or "trainer rejected SESSION_OPEN")
+            return
+        if topic != Topic.SESSION_ACCEPTED or body.get(MsgKey.SESSION_ID) != session.session_id:
+            session.error = f"invalid SESSION_OPEN reply topic/session: {topic!r}"
+            return
+        try:
+            validate_attach_transport(
```
pcnudde: [P1] Do not enforce transport policy from the peer's self-reported reply. `SESSION_OPEN` (including the job/session configuration) has already gone to the target, and both values passed here come from that target's body rather than the actual passive connection. On a clear listener reachable beyond loopback, a rogue trainer can report `grpc://127.0.0.1`/`clear` and bypass `allow_insecure_attach=False`, losing the claimed confidentiality and peer-authentication boundary; F3 network listener allocation binds to all interfaces even when its published URL names localhost (`net_utils.py:291-293`). Enforce the policy on authenticated, immutable connection/listener metadata before sending `SESSION_OPEN`, and test that a spoofed `SESSION_ACCEPTED` over a non-loopback clear route is rejected before metadata disclosure.

nvflare/client/cell/api.py
```diff
@@ -224,16 +239,23 @@ def init(self, rank: Optional[str] = None):
...
                 raise
             self._initialized = True
+            if self._attach:
+                self._attach.register_cleanup()
             self.logger.info(f"trainer session established: fqcn={self._trainer_fqcn} session_id={self._session_id}")
 
     def _register_control_cbs(self, cell: Cell) -> None:
+        if self._attach:
+            self._attach.register_callbacks(cell)
         cell.register_request_cb(channel=CHANNEL, topic=Topic.TASK_READY, cb=self._handle_task_ready)
```
pcnudde, line 263: [P2] Authorize attach control traffic before decoding the request body. These callbacks are routable while the only origin/session checks are inside their handlers, but both Cell paths call `decode_payload` before invoking the registered callback (`cell.py:113-135`, `core_cell.py:1731-1755`). A foreign Cell that knows the stable attach FQCN can therefore send a bogus `TASK_READY` and force allocation/FOBS work or an attacker-selected lazy download before the message is rejected, even with an invalid or absent session ID. Add a header-level/pre-decode origin/topic gate (site-scoped before binding and exact-origin after binding), plus a negative test whose rejected canary/lazy payload must never be materialized or fetched.

nvflare/client/cell/attach_session.py
```diff
@@ -0,0 +1,456 @@
...
+                last_error = e
+                accepted_attempt_id = self._accepted_result_attempt_from_status(task_id, result_id)
+                if not accepted_attempt_id:
+                    session_end = api._result_publication_end_reason()
+                    if session_end:
+                        raise TrainerSessionError(session_end) from e
+            if accepted_attempt_id:
+                self._keep_canonical_attempt(accepted_attempt_id, attempt_waiters)
+                return
+        raise TrainerSessionError(f"result publication failed after status recovery: {last_error}") from last_error
```
pcnudde, line 408: [P2] Preserve result sources while acceptance is still ambiguous. `_handle_result_ready` commits `task.result` and its accepted attempt before returning the ACK (`cell_backend.py:221-232`), so losing both ACKs and the status replies does not mean the result was rejected. Raising here enters `CellClientAPI.send()`'s unconditional exception cleanup at `api.py:477-490`, which deletes every attempt's transfer and clears the handles; a CJ/server that already holds the accepted lazy envelope can then no longer download it. Keep all candidate sources until authority is confirmed or the receiver/session reaches a terminal outcome, and test a first-attempt commit with every ACK/status reply dropped and a delayed downstream pull.

pcnudde (commented):
Four findings from the focused attach-mode security and correctness review are inline.

nvflare/client/cell/attach_session.py
YuanTingHsieh: Fixed in `15b4244c`. The trainer is now a child of the job Cell. Direct profiles require and pre-bind the job-specific `cj_fqcn`; shared-file profiles validate a leased endpoint record and pre-bind its published CJ. `_validate_open()` requires a valid one-segment job ID, exact `<site>.<job>` origin, and equality with the pre-bound CJ before committing session state. Negative tests cover cross-site and same-site/different-job opens.

nvflare/client/cell/attach_session.py
YuanTingHsieh: Fixed in `15b4244c`. Result publication now retains all candidate lazy sources while acceptance is ambiguous, probes `RESULT_STATUS` before honoring routine shutdown, accepts receiver-confirmed terminal completion as authority, and deletes only non-canonical attempts. `test_all_control_replies_lost_keeps_lazy_source_until_receiver_confirmation` covers dropped ACK/status replies with delayed downstream completion; the routine-SHUTDOWN recovery path is covered separately.

nvflare/app_common/executors/client_api/attach_backend.py
YuanTingHsieh: Fixed in `15b4244c`. `_run_task()` now exits when the backend closes or the session changes, and `test_finalize_unblocks_pending_result_wait` covers `result_wait_timeout=None`. Dedicated-listener teardown also keeps an already accepted lazy-result route alive until the trainer source disconnects.

nvflare/client/cell/api.py
YuanTingHsieh: Fixed in `15b4244c`. Attach installs a Cell message interceptor before Cell start. It unwraps streaming headers and rejects unauthorized Attach traffic before FOBS decoding: only the expected CJ is admitted after profile/rendezvous resolution. The canary regression verifies rejected cross-site and premature task payloads are never touched.

nvflare/app_common/executors/client_api/attach_backend.py
YuanTingHsieh: Fixed in `15b4244c`. `AttachBackend` now starts a dedicated CJ-owned listener and classifies the concrete parameters returned by that listener before starting the session loop or sending `SESSION_OPEN`. Clear network is rejected unless explicitly opted in, mTLS is checked from actual listener state, and shared-file permissions are validated. Peer reply fields are no longer transport-policy inputs.

chesterxgchen:
• Although the patch enables externally launched trainers and generally provides pass-through transport, it introduces a Swarm memory regression and several connection/task-state failure modes
  that can cause OOMs, interrupted lazy transfers, or prolonged hangs.

  Full review comments:

  - [P1] Preserve lazy refs for attach Swarm learners — /Users/chesterc/projects/NVFlare/tests/unit_test/app_common/ccwf/test_swarm_forward_memory_path.py:188-188
    On a non-aggregation Swarm client using Attach, _learn_executor_accepts_lazy_refs() still recognizes only EXTERNAL_PROCESS, and this test preserves the in-process expectation. Consequently
    the CJ resolves the full model before AttachBackend serializes it again for the trainer, defeating Cell pass-through and potentially causing an OOM for large models. Treat ATTACH as
    another lazy-ref-capable Cell hop.

  - [P1] Keep live result sources bound during reconnect — /Users/chesterc/projects/NVFlare/nvflare/app_common/executors/client_api/attach_backend.py:341-342
    When allow_reconnect=True and heartbeats lapse after RESULT_READY is accepted but while lazy result transfers are still being served, _run_task() has already cleared _current_task, so this
    code replaces the session despite session.result_source_live. The replacement loses the source-liveness barrier, and finalize() then checks only the new session and can remove the Attach
    connector before the downstream download completes. Defer session rotation while the accepted result source remains live.

  - [P2] Publish QUEUED only after task conversion succeeds — /Users/chesterc/projects/NVFlare/nvflare/client/cell/attach_session.py:265-265
    When task conversion or lazy downloading outlasts the CJ's TASK_READY request timeout, the CJ can query status while this reservation is already QUEUED and interpret it as accepted. If
    conversion subsequently fails, forget_reserved_task() removes the reservation without enqueueing anything, leaving the CJ waiting for a result from a task the trainer never received. Keep
    the reservation in a non-accepted state until conversion and queue insertion succeed.

  - [P2] Fail fast on local task serialization errors — /Users/chesterc/projects/NVFlare/nvflare/app_common/executors/client_api/attach_backend.py:570-571
    When Cell encoding fails locally before any request is transmitted, such as for an unsupported payload object or decomposer failure, this broad catch treats the error as ambiguous delivery
    and repeatedly probes/retries until task_wait_timeout—600 seconds by default. The trainer can only return UNKNOWN, so a deterministic local error becomes a ten-minute stall instead of
    immediate error propagation.

chesterxgchen (commented):
add few comments

pcnudde:
1. trainer accepts implicit cleartext to any host (attach.py:66, bootstrap.py)
A profile with connect_url: "grpc://10.0.0.5:9000" and no connection_security silently defaults to CLEAR — no opt-in, no warning. On clear transport, SESSION_OPEN is authenticated only by the spoofable ORIGIN header, so a rogue listener can bind the trainer and collect model updates. mTLS/bare-CA-TLS enforcement is correct; only the cleartext leg is trust-the-profile. Suggest requiring explicit connection_security: "clear" for network schemes (symmetric with the CJ's allow_insecure_attach).

2. PR body misstates the transport policy. "Cleartext is limited to loopback unless explicitly enabled" is not implemented anywhere: the CJ gate is allow_insecure_attach regardless of loopback (stricter — even 127.0.0.1 cleartext needs the flag), and the trainer side has no restriction at all (finding 1). Fix the body, and decide whether the trainer leg should enforce anything.

pcnudde (approved):

PR #4990, opened 2026-07-29
Split Client API tests across deployment modes

YuanTingHsieh:
## Summary

- rename the existing Client API integration configuration to `client_api_poc.yml` for `POCSiteLauncher` coverage
- add `client_api_prod.yml` for secure `ProvisionSiteLauncher` coverage
- partition all 15 existing Client API cases by deployment mode instead of running the full suite twice
- run the basic external-process and in-process smoke tests in both environments
- preserve the original `/tmp/nvflare/data` cleanup after the final provisioned CIFAR-dependent case

## Why

Client API behavior needs coverage in both the lightweight POC environment and a securely provisioned deployment, but duplicating every test would double the serial Jenkins workload. This split preserves complete example coverage while exercising both launcher paths with a small representative overlap.

## Impact

- 10 cases run with `POCSiteLauncher`
- 7 cases run with `ProvisionSiteLauncher`
- 2 core smoke cases run in both environments
- 17 total executions instead of 30 with full duplication

General examples remain in POC. Torchrun, launch-once, native-tensor, and quantized-streaming cases run in the provisioned environment, alongside the existing provisioned external-process streaming and Swarm relay configurations. The provisioned suite removes its prepared CIFAR data before subsequent configurations or builds run.

## Validation

- `git diff --check`
- parsed both YAML configurations successfully
- verified the union contains all 15 original test cases
- verified the overlap contains exactly the external-process and in-process smoke tests
- verified the POC configuration omits `project_yaml` and the provisioned configuration uses `./data/projects/client_api.yml`
- verified the CIFAR teardown follows the final provisioned case that prepares the dataset and no later case depends on it

Full pytest collection and `./runtest.sh -s` were attempted locally but could not run because the available system Python lacks the development dependencies and blocks package installation under PEP 668.


tests/integration_test/data/test_configs/standalone_job/client_api_prod.yml
```diff
@@ -0,0 +1,146 @@
...
+          "type": "job_submit_success"
+      - "trigger":
+          "type": "run_state"
+          "data": { "run_finished": True }
+        "actions": [ "ensure_current_job_done" ]
+        "result":
+          "type": "run_state"
+          "data": { "run_finished": True }
+    setup:
+      - python tools/prepare_cifar10.py --root /tmp/nvflare/data
```
nvkevlu, line 74: Could we preserve the original suite’s /tmp/nvflare/data cleanup here? client_api_poc.yml removes it, but that configuration runs first; client_api_prod.yml then recreates the dataset in several setup steps and never removes it. Before this split, the original configuration removed it after the last CIFAR-dependent case. Adding the same teardown after pt-client-api-in-process—the final provisioned case that needs CIFAR—would avoid leaving state for subsequent configurations or builds.

YuanTingHsieh: Fixed in a73b5a2a0. I restored the teardown after pt-client-api-in-process, the final provisioned case that prepares CIFAR data, and verified that no later provisioned case prepares or depends on it. Thanks!

nvkevlu (approved):
Thanks, LGTM

PR #4991, opened 2026-07-30
Add MetricsArtifactWriter to default component allowlist

YuanTingHsieh:
## What changed

- Add `nvflare.app_common.widgets.metrics_artifact_writer.MetricsArtifactWriter` to the curated default component allowlist.
- Add a regression test that authorizes `MetricsArtifactWriter` through the built-in fallback policy.

## Why

`BaseFedJob` and multiple built-in recipes configure `MetricsArtifactWriter` by default. The class was missing from the enforced default allowlist, causing otherwise-default non-BYOC recipe jobs to fail component authorization.

The change uses an exact class path, so it does not broaden authorization to the surrounding package.

## Validation

- `104 passed`: component authorizer, static provisioning policy, and simulator environment tests
- `74 passed`: FedAvg and XGBoost recipe tests
- `./runtest.sh -s`
- `git diff --check`


nvflare/app_common/default_component_policy.py
```diff
@@ -63,6 +63,7 @@
     "nvflare.app_common.widgets.event_recorder.ClientEventRecorder",
     "nvflare.app_common.widgets.event_recorder.ServerEventRecorder",
     "nvflare.app_common.widgets.intime_model_selector.IntimeModelSelector",
+    "nvflare.app_common.widgets.metrics_artifact_writer.MetricsArtifactWriter",
```
pcnudde, line 66: [P2] Keep the documented default allowlist synchronized

This adds `MetricsArtifactWriter` to the policy emitted into provisioned and simulated resources, but `docs/user_guide/admin_guide/security/unsafe_component_detection.rst` lines 117–160 says its JSON block is the provisioned/built-in default and still omits this class. Since an explicit `class_allow_list` replaces the built-in default, operators who copy that documented policy will continue to reject the default `BaseFedJob` configuration. Please add this exact path to the guide (or make the documented snapshot generated/checked against `DEFAULT_CLASS_ALLOW_LIST`).

pcnudde (commented):
One documentation-policy consistency issue remains.

nvflare/app_common/default_component_policy.py
YuanTingHsieh: Addressed in `1e921a18`: added the exact `nvflare.app_common.widgets.metrics_artifact_writer.MetricsArtifactWriter` path to the documented `class_allow_list` in policy order. I also verified that the documented JSON parses and contains the path exactly once; the 71 component-authorizer tests continue to pass.

pcnudde (approved):

PR #4992, opened 2026-07-30
Stabilize external process backend shutdown test

YuanTingHsieh:
## What changed

- Let the mocked natural process wait block until the test explicitly releases it, without a competing two-second timeout.
- Always release the mocked wait in a `finally` block so assertion failures cannot strand the finalize thread.

## Why

Under a heavily loaded CI runner, the test thread could be descheduled for more than two seconds after the mocked process wait began. The mock would then raise, causing the backend to correctly fall back to `SIGTERM`; the test subsequently interpreted that signal as a product failure. This made the test sensitive to runner scheduling rather than backend behavior.

## Impact

This changes test synchronization only. Runtime external-process shutdown behavior is unchanged.

## Validation

- `python -m pytest tests/unit_test/app_common/executors/external_process_backend_test.py::TestInitializeAndFinalize -q` — 25 passed
- `./runtest.sh -s` — passed


nvkevlu (dismissed):
LGTM, just need to resolve the conflicts

PR #4994, opened 2026-07-30
Fix external trainer cleanup on abort

YuanTingHsieh:
## Summary

- skip the orderly natural-exit wait once the external-process backend has latched an abort
- send trainer SHUTDOWN without blocking, then immediately enter the existing process-group SIGTERM / grace / SIGKILL teardown
- preserve the configured `shutdown_timeout` behavior for normal, non-abort shutdown
- add focused coverage for both per-task cleanup and finalize cleanup

## Root cause

`ExternalProcessBackend._stop_trainer` performs orderly SHUTDOWN, waits for natural process exit, and only then terminates the owned process group. The abort path used the normal shutdown budget, which defaults to 30 seconds.

The client job process is torn down much sooner during job abort. If the trainer is blocked in user computation and cannot poll `flare.is_running()`, the client job process exits before process-group termination is reached. The trainer survives, is reparented to PID 1, and continues holding resources after the job is marked aborted.

## Impact

External trainers can remain alive after an aborted job and continue holding GPU and host memory. The legacy launcher path does not exhibit this behavior because it signals the process group immediately.

The abort path now matches that behavior while retaining the existing configurable `stop_grace_period` between SIGTERM and SIGKILL.

## Validation

- `97 passed`: `tests/unit_test/app_common/executors/external_process_backend_test.py`
- `./runtest.sh -s`
- Black, isort, Flake8, and `git diff --check`
- regression tests were confirmed red with the old 30-second abort wait and green with this change

The downstream container case `L0_ml2fl_client_api_ext_process_abort_cleanup_converted` can serve as the end-to-end regression guard once the corresponding QA change is enabled.


pcnudde (commented):
Independent review findings:

nvflare/app_common/executors/client_api/external_process_backend.py
```diff
@@ -370,7 +373,7 @@ def finalize(self, fl_ctx: FLContext) -> None:
                     self._request_trainer_shutdown(trainer, wait_timeout=_LIVE_RESULT_SHUTDOWN_ACK_TIMEOUT)
                     self._reap_trainer_after_result(trainer)
                 else:
-                    self._stop_trainer(trainer, natural_exit_wait=self._shutdown_wait_bound())
+                    self._stop_trainer(trainer, natural_exit_wait=self._stop_wait_bound())
```
pcnudde, line 376: [P1] Bound the SIGTERM grace on abort

`_stop_wait_bound()` removes the natural-exit wait, but `_stop_trainer()` still passes the unchanged `stop_grace_period` to `_terminate_process_tree()`. Its default is 30 seconds, and SIGKILL is sent only after that grace. FLARE-3056 records that the client-job process exits about seven seconds after abort. Therefore a trainer/group member that ignores or delays SIGTERM can outlive the client before SIGKILL is reached, recreating the orphan/resource leak. Please apply an abort-specific bound to the TERM-to-KILL grace too and cover a SIGTERM-resistant group in the abort tests.

nvflare/app_common/executors/client_api/external_process_backend.py
```diff
@@ -359,8 +359,11 @@ def finalize(self, fl_ctx: FLContext) -> None:
         # Serialize close with RESULT_READY's acceptance commit.
         with self._task_lock:
             self._closed = True
-        # The same gate orders END_RUN against the launch-install-to-Popen window.
+        # The same gate orders END_RUN against the launch-install-to-Popen window. Keep
+        # this ordering bound on abort: a process handle may not have been installed yet.
         admitted = self._execute_gate.acquire(timeout=self._shutdown_wait_bound())
+        # Cleanup is unconditional if the gate times out. It marks the launch cleaned so
```
pcnudde: [P2] Do not return from finalize with a spawned process still alive

When gate acquisition times out, `_stop_trainer()` can see `trainer.process is None`, mark the session cleaned, and let `finalize()` return. The late cleanup at `_launch_trainer()` runs only if `Popen` later returns while the client process is still alive. The PR's own `test_finalize_stops_process_returned_after_launch_cleanup[abort]` asserts that the process remains alive immediately after `finalize()` returns. With valid `shutdown_timeout=0`, client teardown can therefore occur before that late kill. Please make finalize/launch coordination guarantee that no already-created child is alive when finalize returns, and assert that invariant in the regression test.

nvflare/app_common/executors/client_api/external_process_backend.py
YuanTingHsieh: Addressed in 4b26b648a. `_stop_trainer()` now passes `_termination_grace()` to `_terminate_process_tree()`; it returns `0.0` after abort while preserving the configured grace for normal shutdown. Added `test_abort_escalates_sigterm_to_sigkill_without_configured_grace`, which models a SIGTERM-resistant group and verifies SIGTERM followed by SIGKILL with zero abort grace. The full backend suite passes (101 tests).

nvflare/app_common/executors/client_api/external_process_backend.py
YuanTingHsieh: Addressed in 4b26b648a, with the lock-release ordering made explicit in 4c8ce4602. `_launch_trainer()` now holds `trainer._stop_lock` across `Popen` and process-handle/PGID installation. If `finalize()` times out on the execute gate, `_stop_trainer()` waits for that same lock and terminates the installed process before returning. The updated `test_finalize_stops_process_returned_after_launch_cleanup` forces this in-flight `Popen` window for both normal and abort paths and asserts that the process is dead when `finalize()` returns. The full backend suite passes (101 tests).

pcnudde (approved):

PR #4995, opened 2026-07-30
Synchronize F3 streaming parameters

nvidianz:
## What changed

- Advertise the sender's effective streaming chunk size, window size, ACK interval, and retry pending-byte limit in the first stream message.
- Validate and adopt those values on the receiver for the lifetime of the stream.
- Preserve backward compatibility by retaining the receiver's local configuration when a sender does not provide a parameter or provides an invalid value.
- Update the defaults to the benchmark-selected 25 Gbit/s settings:
  - chunk size: 1 MiB
  - window size: 64 MiB
  - ACK interval: 16 MiB
  - retry pending-byte limit: 128 MiB
- Update F3 benchmark defaults, documentation, and regression coverage.

## Why

Previously, the sender used its local streaming window while the receiver used its own local ACK interval. If the endpoint configurations differed, the sender could exhaust its window before the receiver reached its ACK threshold, causing the stream to stall and eventually time out.

Making the sender's effective parameters authoritative for each stream keeps flow control consistent without requiring identical site configuration files. Older senders remain supported because missing headers fall back to the receiver's existing local settings.

## Validation

- `~/venv/3.12/bin/python -m pytest -q tests/unit_test/fuel/f3/streaming tests/unit_test/fuel/f3/cellnet_bench_test.py` — 385 passed
- Focused streaming/default tests after formatting — 96 passed
- Scoped Black, isort, flake8, and `git diff --check` — passed

The repository-wide style scan also traverses three unrelated ignored `node_modules` Python files that require Black formatting; those generated dependencies were left untouched.


chesterxgchen (commented):
From UX perspective, how many of the configuration is really data scientist-tuned vs. FLARE org admin tuned ? 

How many of these parameters need to be tuned from Job recipes ?

nvidianz:
@chesterxgchen For normal networks, none of these parameters need to be tuned by data scientists or from job recipes. We tested the defaults across networks ranging from 100 Mbit/s to 25 Gbit/s, and the current defaults were optimal throughout that range. These are transport-level settings intended for a FLARE org admin to adjust only for unusual network environments, such as satellite communications; they are tunable as an escape hatch for those cases.

nvflare/fuel/f3/streaming/byte_receiver.py
```diff
@@ -38,10 +42,8 @@
 log = logging.getLogger(__name__)
 
 MAX_OUT_SEQ_CHUNKS = 16
```
pcnudde, line 43: [P1] Preserve the legacy ACK fallback for headerless senders.

Raising the receiver fallback ACK interval from 4 MiB to 16 MiB deadlocks a headerless legacy sender configured with (for example) an 8 MiB window: the sender blocks after roughly `window + one chunk` in flight, while this receiver consumes those bytes but does not ACK until 16 MiB. Neither side can make progress and the sender eventually times out.

Please keep the 4 MiB compatibility fallback specifically when `ACK_INTERVAL` is absent; new senders can still advertise the 16 MiB default. Add a transfer-level regression test with a headerless 8 MiB-window sender and a payload larger than 16 MiB.

pcnudde (changes requested):
Request changes: the receiver’s new fallback ACK interval breaks the stated rolling-upgrade compatibility for headerless legacy senders.

chesterxgchen:
> @chesterxgchen For normal networks, none of these parameters need to be tuned by data scientists or from job recipes. We tested the defaults across networks ranging from 100 Mbit/s to 25 Gbit/s, and the current defaults were optimal throughout that range. These are transport-level settings intended for a FLARE org admin to adjust only for unusual network environments, such as satellite communications; they are tunable as an escape hatch for those cases.

Is it possible to allow the system self-adapt the configuration rather than user have to manually edit and change, which is usually out of data scientist's domain. 

The system knows the speed of transfer and network speed; instead hard-coded default, we can adapt with the network, model size etc. to dynamic changes to make sure the best result

nvflare/fuel/f3/streaming/byte_receiver.py
nvidianz: Good catch. Fixed.

nvidianz:
@chesterxgchen In principle this is possible, but reliable auto-tuning is difficult and expensive to calibrate and validate. Selecting these parameters requires large end-to-end transfers across many combinations; our benchmark sweep took hours to complete.

The resulting defaults worked well across every network we tested, from 100 Mbit/s through 25 Gbit/s. Regular customers should therefore not need to tune these settings. Adding dynamic adaptation would introduce substantial complexity without a demonstrated benefit for normal deployments, so we prefer stable, validated defaults while retaining manual overrides as an escape hatch for unusual network environments. We can revisit auto-tuning if we encounter a concrete customer or network case that the defaults do not handle well.

pcnudde (approved):

nvflare/fuel/f3/streaming/byte_streamer.py
```diff
@@ -239,14 +241,22 @@ def __init__(
         config = CommConfigurator()
         self.reliable = config.get_streaming_reliable(False) if reliable is None else reliable
         self.window_size = config.get_streaming_window_size(STREAM_WINDOW_SIZE)
+        self.ack_interval = config.get_streaming_ack_interval(STREAM_ACK_INTERVAL)
+        if self.ack_interval > self.window_size:
+            log.warning(
+                f"{self} streaming_ack_interval {self.ack_interval} exceeds streaming_window_size "
+                f"{self.window_size}; using {self.window_size}"
+            )
+            self.ack_interval = self.window_size
```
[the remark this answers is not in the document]
nvidianz: Thanks. The validated defaults already preserve ample flow-control margin: the ACK interval is 16 MiB and the window is 64 MiB. The clamp is intended as a basic safety bound when an override sets the ACK interval above the window; it is not intended to make every arbitrary parameter combination optimal or completely foolproof.

These are advanced transport-level settings exposed as an escape hatch for unusual environments. Anyone overriding them is expected to understand the flow-control tradeoffs and validate the resulting configuration. Regular customers should use the defaults and should not need to tune these parameters, so we do not plan to add further automatic adjustment here.

PR #5003, opened 2026-07-31
Fix Collab execution in POC and simulation

nvidianz:
## What changed

- initialize the server job process `ServerEngine` with its newly created communication Cell
- fail Collab setup cleanly if a server Cell is unavailable
- propagate simulator `FATAL_SYSTEM_ERROR` state through the simulator exit code
- add focused regression tests for missing Cells and simulator fatal status

## Why

Collab jobs submitted through `PocEnv` synchronized their clients and then crashed before setup because `engine.get_cell()` returned `None`. The server job process created a Cell but did not initialize its engine with it, so `CollabController` failed while calling `get_fqcn()`.

Separately, exceptions raised by `@collab.main` triggered a fatal server error but the simulator still returned exit code 0. As a result, `recipe.execute()` returned normally and `Run.get_status()` remained unavailable in `SimEnv`. The simulator now returns failure when the server context records `FATAL_SYSTEM_ERROR`, allowing the existing `SimEnv` failure handling to surface the error to the caller.

## Impact

- Collab jobs can complete in POC deployments.
- Failed Collab workflows in simulation no longer look successful to the calling script.

## Validation

- `python -m pytest -q tests/unit_test/collab/runtime/controller_test.py tests/unit_test/private/fed/app/simulator/simulator_runner_test.py` — 30 passed
- `python -m pytest -q tests/unit_test/collab tests/unit_test/recipe/sim_env_test.py` — 107 passed
- Black, isort, and flake8 pass for all changed files
- shipped `examples/hello-world/hello-collab/job.py` with `PocEnv(num_clients=2)` — `FINISHED:COMPLETED`
- sentinel failing `@collab.main` with `SimEnv` — surfaced as `RuntimeError` with simulator return code 2

The full-tree style command also scans generated files under an existing `node_modules` directory and reported three unrelated `flatted.py` formatting failures; all files changed by this PR pass the scoped checks.


nvflare/private/fed/app/server/runner_process.py
```diff
@@ -102,6 +102,7 @@ def main(args):
             server.cell = server.create_job_cell(
                 args.job_id, args.root_url, args.parent_url, secure_train, server_config
             )
+            server.engine.initialize_comm(server.cell)
```
holgerroth: `initialize_comm()` does more than set the engine's cell — it also registers `_handle_aux_message` on `CellChannel.AUX_COMMUNICATION, topic="*"` (`server_engine.py:508`). At this point, `create_job_cell()` (just above) has already started the `ServerCommandAgent`, which registered `aux_communicate` on the same channel/topic (`fed_server.py:573`, `server_command_agent.py:45`). `Registry.set()` silently overwrites (`fuel/f3/cellnet/registry.py`), so this call replaces the SJ's aux handler with the engine's.

The two handlers aren't equivalent:

1. **Auth/state checks are lost.** `aux_communicate` runs the `server_state.aux_communicate()` state check plus `authentication_check()` (SSID validation — `fed_server.py:1045-1056`); `_handle_aux_message` dispatches without either. All incoming SJ aux traffic (runner sync, fed events, aux requests) in POC/production would now skip those checks.
2. **The documented CJ-before-SJ race loses its graceful path.** `aux_communicate` returns `SERVICE_UNAVAILABLE` while the engine isn't ready so the CJ retries cleanly (`server_command_agent.py:123-128`); `_handle_aux_message` instead raises `AttributeError` on `self.run_manager.aux_runner` while `run_manager` is still `None`. The client sync loop retries on any non-OK reply (`client_runner.py:704-732`), so startup still converges, but with error stack traces instead of the intended handling.

Collab doesn't actually need the aux registration: `CollabController` registers its own `MSG_CHANNEL` callback and only needs `engine.get_cell()` to return the cell. The aux *send* path already works without this call, since `start_run()` sets `run_manager.cell = self.cell` (`fed_server.py:986`).

Suggestion: use a narrow setter instead — e.g. a `ServerEngine.set_cell(cell)` that assigns `self.cell` (and `run_manager.cell` if present) without re-registering the aux callback. If converging the SJ on `_handle_aux_message` is actually intended, that seems worth doing explicitly with the auth/state checks ported over, plus a test pinning which callback is registered on the SJ cell's `AUX_COMMUNICATION` channel.

chesterxgchen (dismissed):
LGTM. you will need to test POC end-to-end before merge

nvflare/private/fed/app/server/runner_process.py
nvidianz: Thanks — you are right that `initialize_comm()` was too broad here and could overwrite the SJ AUX handler. I replaced it with a narrow `ServerEngine.set_cell()` that only updates `engine.cell` and `run_manager.cell` when present, without registering any callback. I also added a parameterized regression test that asserts `register_request_cb()` is not called, with and without a run manager. Validation: 59 focused tests pass, scoped Black/isort/flake8 pass, and the shipped two-client Collab POC example still finishes `FINISHED:COMPLETED`.

chesterxgchen (commented):
Change my original Approval, I think Holger's comment is correct. 

  I would also add tests that:

  - set_cell() updates engine.cell and run_manager.cell without registering a callback.
  - After server-job Cell creation and set_cell(), the AUX wildcard callback remains ServerCommandAgent.aux_communicate.
  - Invalid SSID/state still rejects AUX traffic through the server-job Cell.

  A suitable response to the reviewer:

  > Good catch—Holger is right. I verified that Cell._register_request_cb() replaces both the CoreCell and streaming callbacks, so initialize_comm()
  > overwrites ServerCommandAgent.aux_communicate. That bypasses its state/SSID checks and its pre-run SERVICE_UNAVAILABLE path. Collab only needs
  > the engine’s Cell reference, so we’ll switch this to a narrow set_cell() method and add a regression test pinning the SJ AUX callback.

nvidianz:
@chesterxgchen Addressed in `adae133ef`. In addition to the existing `set_cell()` unit test, I added server-job Cell tests that:

- capture the AUX wildcard callback after `create_job_cell()` followed by `set_cell()` and assert it remains `ServerCommandAgent.aux_communicate`;
- exercise that captured callback with a wrong SSID and a not-in-service server state, asserting both return `AUTHENTICATION_ERROR` and never dispatch AUX traffic.

Validation: `fed_server_test.py` plus `server_engine_test.py` pass (`55 passed`), and the changed test passes Black, isort, and flake8. The two-client Collab POC run from the prior follow-up remains `FINISHED:COMPLETED`.

holgerroth (approved):
The aux-handler concern is fully addressed: `set_cell()` keeps `ServerCommandAgent.aux_communicate` as the server job's AUX handler (auth/state checks and the pre-run `SERVICE_UNAVAILABLE` path intact), the new fed_server test pins the registered callback identity, and the SSID/state parametrization cleanly isolates both rejection paths. Also verified the refactored `initialize_comm()` is behavior-identical for its parent-server and simulator callers. LGTM.

YuanTingHsieh (approved):
LGTM

PR #5005, opened 2026-07-31
Add federated LLM SFT example for Collab API

ZiyueXu77:
## Description

This PR adds a full-parameter federated supervised fine-tuning example for Hugging Face causal language models using the Collab API.

- Adds a `prepare_data.py`-first workflow with lightweight synthetic data and a Dolly mode that deterministically partitions `databricks/databricks-dolly-15k` across clients.
- Runs stateful SFT clients with five federated synchronizations per local epoch by default.
- Exchanges native `dict[str, torch.Tensor]` model states through Collab calls, without application-level conversion to NumPy, `FLModel`, or `Shareable`.
- Supports configurable checkpoints, precision, client count, synchronization frequency, evaluation, and checkpoint retention.
- Keeps Hugging Face remote checkpoint code disabled by default, with an explicit trusted-checkpoint opt-in.
- Exposes simulator GPU placement and supports the standard Recipe export arguments.
- Documents the Collab lifecycle, resource considerations, and training-curve alignment.

## Motivation

Collab workflows can invoke published Python methods and carry ordinary Python objects directly. This example shows how to express full-parameter federated SFT with native tensor dictionaries while keeping the client training and server aggregation responsibilities visible.

The example intentionally keeps the training path focused on the Collab programming model. Multi-process launch, experiment tracking, and deployment-oriented configuration remain in the existing full Hugging Face LLM example.

## Validation

- `conda run -n nvflare_llm env PYTHONPATH=. python -m pytest tests/unit_test/collab -q` — 97 passed.
- `conda run -n nvflare_llm ./runtest.sh -s` — Black, isort, Flake8, and agent-skill lint passed.
- Recipe export smoke test completed without prepared local data or model loading.
- `python -m compileall -q examples/advanced/collab/pt_llm_sft`
- `python examples/advanced/collab/pt_llm_sft/prepare_data.py --help`
- `PYTHONPATH=examples/advanced python -m collab.pt_llm_sft.pt_llm_sft --help`


examples/advanced/collab/pt_llm_sft/pt_llm_sft.py
```diff
@@ -0,0 +1,434 @@
...
+        if self.tokenizer.pad_token is None:
+            self.tokenizer.pad_token = self.tokenizer.eos_token
+
+        _, _, dtype = precision_config(self.precision)
+        self.model = AutoModelForCausalLM.from_pretrained(
+            self.model_name_or_path,
+            revision=self.model_revision,
+            trust_remote_code=True,
+            dtype=dtype,
+        )
```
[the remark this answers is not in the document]
ZiyueXu77: fixed

examples/advanced/collab/pt_llm_sft/pt_llm_sft.py
```diff
@@ -0,0 +1,442 @@
...
+    simple_logging()
+    print(
+        "Starting Collab full-parameter SFT simulation\n"
+        f"  model: {args.model_name_or_path}\n"
+        f"  clients: {args.num_clients}\n"
+        f"  epochs: {args.num_epochs}\n"
+        f"  syncs per epoch: {args.syncs_per_epoch}\n"
+        f"  data: {args.data_root}"
+    )
+    run = make_recipe(args).execute(SimEnv(num_clients=args.num_clients, workspace_root=args.workspace_root))
```
pcnudde: [P2] Expose simulator GPU placement

The entry point never forwards a `gpu_config` to `SimEnv`. Without one, every simulator client worker inherits the same host-visible GPU set, and this Trainer setup selects the default CUDA device, so a four-client run stacks all four full-model allocations on GPU 0. That makes the documented Qwen3-8B “one A100 per client” placement impossible to reproduce and can OOM while the other GPUs remain idle. Please expose a `--gpu-config` option and pass it through as `SimEnv(gpu_config=...)`.

examples/advanced/collab/pt_llm_sft/README.md
```diff
@@ -0,0 +1,170 @@
...
+- one epoch divided into five federated synchronizations;
+- 10 local optimizer steps per client and synchronization;
+- batch size 1, sequence length 64, and learning rate `2e-5`;
+- evaluation disabled so the measurement focused on training and exchange.
+
+For reference, run the regular NVFlare simulator from
+`examples/advanced/llm_hf` with:
+
+```bash
+python job.py \
```
pcnudde: [P2] Make the benchmark command reproduce the table

This command does not run the matched workload described above. It supplies one `dolly` client and otherwise keeps the regular example’s defaults: three rounds, GPT-Neo-1.3B, GPU 0, batch size 2 with gradient accumulation 20, learning rate `5e-4`, and sequence length 1024. The table instead claims four clients, five synchronizations, TinyLlama/Qwen3, batch size 1, learning rate `2e-5`, and sequence length 64. The matched adapter/config used earlier in this PR was removed, so the performance result is no longer reproducible from the final branch. Please publish the exact harness/config or remove or clearly qualify this comparison.

pcnudde (commented):
Two actionable findings:

examples/advanced/collab/pt_llm_sft/pt_llm_sft.py
ZiyueXu77: added

examples/advanced/collab/pt_llm_sft/README.md
ZiyueXu77: Tested again, and it appears that the data format conversion may only be a very small portion of the total cost, and the actual time is dependent on many factors especially on a cluster env where the memory / connection condition are dynamic. Remove the time comparisons and keep only a narrow correctness note: the two paths followed similar loss trajectories.

pcnudde (approved):

PR #5011, opened 2026-08-03
Atomically persist PyTorch checkpoints

pcnudde:
## What changed

- Write current and best PyTorch checkpoints to unique temporary files in the destination directory.
- Flush and fsync the completed temporary file before atomically installing it with `os.replace()`.
- Fsync the parent directory where supported, clean only the owned temporary path on failure, and propagate persistence errors.
- Add focused regression coverage for partial writes, ENOSPC-style failure, interruption, cleanup, successful replacement, and both current/best paths.

## Why

`PTFileModelPersistor.save_model_file()` previously passed the authoritative checkpoint path directly to `torch.save()`. A partial serialization caused by ENOSPC, an I/O error, or interruption could therefore truncate the previously valid current or best checkpoint. Both save paths share this boundary, so the fix is contained in the core persistor rather than individual workflows or examples.

The persistence dictionary, authoritative filenames, and PyTorch load format remain unchanged. The completed file becomes visible only at the atomic replacement point.


chesterxgchen:
### Review finding

  NamedTemporaryFile creates a file readable/writable only by its owner. After os.replace(), that temporary inode becomes the checkpoint, so an existing 0644/0640 checkpoint may
  silently become 0600 and lose existing ACLs or other inode metadata. Python documents these restrictive tempfile permissions explicitly. Python tempfile docs
  (https://docs.python.org/3/library/tempfile.html#tempfile.mkstemp)

  That could break another user or group reading exported checkpoints. I’d suggest preserving at least the existing destination’s mode with os.stat() and os.fchmod(), or
  explicitly documenting 0600 as intentional. Replacing the inode also changes symlink/hard-link behavior, which is worth confirming.

  A useful additional test would make the temporary-file os.fsync() raise ENOSPC or EIO, then verify that the old checkpoint remains and the temporary file is removed. The
  existing test raises ENOSPC from torch.save(), but delayed-allocation filesystems may report it only during fsync().

pcnudde:
Addressed in 4ea6f50d1.

- Replacements now preserve the existing checkpoint mode, while newly created checkpoints use normal exclusive file creation and therefore honor the process umask instead of inheriting tempfile's 0600 mode.
- Added an fsync fault-injection test that raises ENOSPC before replacement and verifies the previous checkpoint remains byte-identical/loadable and the owned temporary file is removed.
- Confirmed the atomic install intentionally replaces the destination directory entry/inode; symlink or hard-link identity and arbitrary inode metadata such as ACLs are not preserved by os.replace. Preserving those identities would conflict with the requested atomic path replacement and is outside this focused fix.

Validation: 7 focused tests passed; ./runtest.sh -s passed (Black, isort, flake8, and agent-skill lint).

nvidianz (approved):

PR #4983, opened 2026-07-28
Finish concrete Recipe entrypoint and documentation cleanup

nvkevlu:
Finishs concrete Recipe entrypoint and documentation cleanup.

### Description

Recipe-facing application code should work through documented concrete recipe constructors without depending on the generated backing job. This keeps internal `FedJob` assembly encapsulated while preserving the lower-level Job API for advanced use cases.

- Removed the optional backing `CCWFJob` parameter from `BaseSwarmLearningRecipe` and exposed `min_clients` directly for job scheduling.
- Added dedicated `KaplanMeierRecipe` implementation.
- Simplified the Kaplan–Meier and FedBPT `job.py` entrypoints to instantiate concrete recipes without exposing `Recipe` or `FedJob`.
- Removed stale Recipe API technical-preview wording.
- Reinforced concrete recipes such as `FedAvgRecipe` and `CyclicRecipe` as the primary public entrypoints.
- Clarified that the lower-level FedJob API remains intentionally documented separately for advanced component placement and custom job construction.

## Validation

- Python syntax compilation passed for the changed Python files.
- Black, isort, flake8, license-header, and whitespace checks passed for the changed files.
- Confirmed that Recipe-facing documentation and entrypoints no longer access `recipe.job` or instantiate the base `Recipe`.

### Types of changes

- [x] Non-breaking change (fix or new feature that would not break existing functionality).
- [x] Breaking change (swarm constructor change).
- [ ] New tests added to cover the changes.
- [ ] Quick tests passed locally by running `./runtest.sh`.
- [ ] In-line docstrings updated.
- [ ] Documentation updated.

PR #5013, opened 2026-08-03
Add concrete FedBPT recipe entrypoint

nvkevlu:
Extracts FedBPT job assembly into a dedicated concrete recipe while preserving the existing CLI and generated job behavior.

### Description

This keeps the research-specific change separate from the broader Recipe API and documentation cleanup in #4983.

- Added a concrete `FedBPTRecipe` that encapsulates the existing FedBPT server, client, workflow, and component configuration.
- Simplified `research/fed-bpt/job.py` to parse the existing CLI arguments and instantiate the concrete recipe.
- Preserved the existing export and execution flags.
- Updated the FedBPT README to demonstrate the concrete recipe entrypoint.
- Added tests covering early export-flag parsing and exported recipe configuration.
- Ensured the entrypoint works both when invoked directly and when loaded as a Python module.

## Compatibility

There are no intended behavioral or CLI changes. This refactors where the FedBPT job is assembled without changing its generated configuration.

A job exported from the original implementation and one exported through `FedBPTRecipe`, using the same arguments, were compared recursively and were byte-for-byte identical.

## Validation

Validation was performed with Python 3.11 and `cma==3.4.0`.

- The two refactor-specific FedBPT tests passed:
  - CLI export flags are parsed before importing the recipe and optional dependencies.
  - The recipe exports the expected FedBPT job configuration.
- The original and refactored CLI entrypoints both successfully exported a job using the same minimal configuration.
- `diff -ru` reported no differences between the original and refactored exported job directories.
- Black, isort, flake8, license-header, and whitespace checks passed for the changed files.

The remaining test in the full FedBPT test file encounters an existing CMA state-serialization failure. The same failure was reproduced against the unmodified baseline, so it was not introduced by this refactor.

### Types of changes

- [ ] Documentation updated.


research/fed-bpt/fedbpt_recipe.py
```diff
@@ -0,0 +1,173 @@
...
+            launcher=launcher,
+        )
+        job.to_clients(runner, tasks=["train"])
+        job.to_clients(RegisterDecomposer(), id="register_decomposer")
+        super().__init__(job)
+
+    def export(self, *args, **kwargs):
+        with _temporary_sys_path(SRC_DIR):
+            return super().export(*args, **kwargs)
+
```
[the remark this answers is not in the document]
nvkevlu: Updated both wrappers to mirror the explicit Recipe.export() and Recipe.run() signatures and forward each argument by name while preserving the temporary source-path context.

research/fed-bpt/fedbpt_recipe.py
```diff
@@ -0,0 +1,193 @@
...
+            eval_clients,
+            "--llama_causal",
+            llama_causal,
+        ]
+        if batch_size is not None:
+            client_args.extend(["--batch_size", batch_size])
+        if train_args:
+            client_args.extend(shlex.split(train_args))
+        if extra_train_args:
+            client_args.extend(extra_train_args)
```
YuanTingHsieh, line 127: `train_args` and `extra_train_args` are transformed into the launcher script, but neither is retained on the recipe nor explicitly passed to the secret scanner. Since `Recipe._warn_potential_secrets_in_params()` checks conventional attributes such as `self.train_args`, `run()` can deploy secret-like values without the expected `PotentialSecretWarning`. Please retain or explicitly scan the combined user-supplied arguments before use.

YuanTingHsieh (commented):
One security-contract issue in the new concrete recipe.

research/fed-bpt/fedbpt_recipe.py
nvkevlu: Thank you, addressed in commit

YuanTingHsieh (approved):

PR #5014, opened 2026-08-03
Handle FedCE client participation edge cases

ZiyueXu77:
## What changed

- Reject FedCE recipes configured with fewer than two clients before a run starts.
- Use a uniform prior when a participating client has no prior contribution weight.
- Carry contribution weights for absent clients across rounds and warn when the client set changes.
- Document the full-participation assumption and add regression coverage for both edge cases.

## Why

FedCE requires at least two clients for its leave-one-out calculation and assumes full participation. Those constraints were implicit when the implementation was promoted into the recipe API in #4922.

With a single client, the contribution weight becomes 1.0 and the next round fails while constructing the minus model. With partial participation, a returning client previously received a 0.0 client-side fallback while the server used an incompatible 1.0 prior, silently corrupting the contribution estimate.

This change keeps full participation as the recommended configuration while making changing client sets explicit and deterministic.

## Validation

- Reproduced the pristine single-client recipe accepting min_clients=1; it now raises at construction.
- Reproduced rotating two-of-three participation; the aggregator now warns, assigns a uniform prior to the new client, and retains the absent client weight.
- 34 focused FedCE unit tests passed.
- 5 related FedCE research and recipe CLI tests passed.
- ./runtest.sh -s passed.

### DIFF
--- a/docs/user_guide/data_scientist_guide/available_recipes.rst
+++ b/docs/user_guide/data_scientist_guide/available_recipes.rst
@@ -196,6 +196,11 @@ FedCE
 contribution from gradient-direction novelty and a client-computed leave-one-out (minus-model) score,
 then uses those estimates as aggregation weights.
 
+FedCE requires at least two clients and assumes full participation in every round. Configure
+``min_clients`` to include all participating clients whenever possible. If the client set changes,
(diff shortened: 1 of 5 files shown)


nvkevlu (approved):
Thanks, LGTM

PR #5015, opened 2026-08-03
Raise default streaming blob size to 4 GiB

pcnudde:
## What changed

- Raise the default `streaming_max_blob_size` from the F3 message payload limit (2,144,337,904 bytes) to 4 GiB (4,294,967,296 bytes).
- Keep `streaming_max_blob_size` and `NVFLARE_STREAMING_MAX_BLOB_SIZE` overrides unchanged.
- Add one focused regression test for the new default.

## Why

A Qwen2.5-72B BF16 tensor was observed with a declared streamed size of 2,491,416,937 bytes. The receiver rejected it under the old default even though streaming transports the blob incrementally; a 3 GiB override allowed the transfer to proceed.

The streaming blob ceiling is independent of F3 message framing, so retaining the message payload limit as its default unnecessarily blocks large streamed tensors. A 4 GiB default covers the observed 72B tensor while preserving a finite receiver safety boundary.

## User impact

The observed 72B tensor now fits without additional configuration. This change does not reserve 4 GiB for small models, change frame/chunk/window sizes, or silently remove the runtime receive limit. Objects above 4 GiB still require an explicit override or future tensor-layout/chunking work.

## Validation

- `python -m pytest -q tests/unit_test/fuel/f3/comm_config_test.py tests/unit_test/fuel/f3/streaming/blob_streamer_test.py` — 21 passed
- `./runtest.sh -s` — passed (Black, isort, flake8, and agent-skill lint)

End-to-end 72B or multinode training was not run locally; this removes the observed default-limit blocker only.

Jira: [FLARE-3076](https://jirasw.nvidia.com/browse/FLARE-3076)

### DIFF
--- a/nvflare/fuel/f3/comm_config.py
+++ b/nvflare/fuel/f3/comm_config.py
@@ -19,6 +19,7 @@ _comm_config_files = ["comm_config.json", "comm_config.json.default"]
 
 
 DEFAULT_MAX_MSG_SIZE = MAX_PAYLOAD_SIZE
+DEFAULT_STREAMING_MAX_BLOB_SIZE = 4 * 1024 * 1024 * 1024
 
 
 class VarName:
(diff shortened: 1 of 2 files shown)


chesterxgchen (approved):

nvidianz (approved):
This is an artificial limit. The purpose is to discourage the use of BLOB streaming APIs for very large data. It's more efficient to use byte streamer.

PR #5017, opened 2026-08-03
Fix HCI session expiry during active result downloads

pcnudde:
## What changed

- Bind each authorized result-download transaction to the authenticated HCI session that created it.
- Refresh only that session when the server observes advancing transfer bytes.
- Cancel and detach bound downloads when the session ends, the transfer completes, or the transfer fails.
- Abort client-side download work when `SESSION_EXPIRED` arrives and raise the typed `SessionExpired` outcome without retrying a stopped messenger.
- Add deterministic coverage for progressing, stalled, expired-certificate, logout, concurrent-session, cleanup, and post-expiry behavior.

## Why

Result downloads use the Cell file-transfer path after their initial HCI command. Transfer progress therefore did not update `Session.last_active_time`. A healthy download lasting longer than the default 1,800-second idle timeout was treated as inactivity, causing the server to expire the session and the client to stop its messenger mid-transfer.

This is a core NVFlare defect because applications using the standard admin API cannot reliably retrieve sufficiently large result artifacts without issuing unrelated keepalive commands.

## Validation

- `71 passed`: focused session-download and FLARE API regression tests
- `313 passed`: broader HCI, FLARE API, and job-command tests
- Black, isort, and flake8 pass on all eight changed files
- `git diff --check upstream/main...HEAD` passes

## Boundaries

This change does not redesign artifact delivery or add cross-process durable resume. Full provisioned TLS/multinode behavior was not reproduced locally; the regression uses deterministic in-process transfer coverage with final size and SHA-256 validation.

Jira: FLARE-3079


nvflare/fuel/flare_api/flare_api.py
```diff
@@ -197,6 +198,9 @@ def close(self):
         self.api.logout()
 
     def try_connect(self, timeout):
+        session_expired_reason = getattr(self.api, "session_expired_reason", None)
```
chesterxgchen, line 201: the same code repeated in multiple places, considering refactoring this into one function

nvflare/fuel/hci/server/binary_transfer.py
```diff
@@ -40,17 +41,31 @@ def download_folder(self, conn: Connection, tx_id: str, folder_name: str):
...
+                session_mgr.mark_download_active(session_id, tx_id)
 
         cell = engine.get_cell()
         source_fqcn = cell.get_fqcn()
         downloader = ObjectDownloader(
             num_receivers=1,
-            cell=engine.get_cell(),
-            timeout=timeout,
+            cell=cell,
+            timeout=admin.timeout,
```
chesterxgchen, line 57: is admin.timeout always defined ? the old code seems to suggest this is not always defined

chesterxgchen (commented):
### Findings

  - P2 — Session teardown can orphan active downloads. In end_session_by_id()
    (https://github.com/NVIDIA/NVFlare/blob/1a4e823424434757cc0d021dc8a64d5282bb8c06/nvflare/fuel/hci/server/sess.py#L238-L251), the
    session and download registry are removed before the expiry notification and cancellation callbacks run. If fire_and_forget()
    raises, no downloads are canceled; if one cancellation raises, later downloads are skipped. Because the registry has already been
    discarded, they cannot be canceled by a subsequent teardown. Wrap notification/cancellation independently and catch/log each
    callback failure.

  - P2 — Aborted downloads leave partial client files behind. The new abort signal is passed at api.py:498
    (https://github.com/NVIDIA/NVFlare/blob/1a4e823424434757cc0d021dc8a64d5282bb8c06/nvflare/fuel/hci/client/api.py#L492-L508), but the
    downloader returns the partial temporary path on failure and only closes it—it does not remove it. Downloader implementation

    (https://github.com/NVIDIA/NVFlare/blob/1a4e823424434757cc0d021dc8a64d5282bb8c06/nvflare/fuel/f3/streaming/file_downloader.py#L181-L217).
    AdminAPI.download_file() discards that path, while pull_folder() also returns without deleting previously downloaded files. A
    certificate expiry or logout during a large download can therefore leave a large file in /tmp plus a partial transaction directory.
    Delete file_path on error and clean the local transaction directory when pull_folder() fails.

nvflare/fuel/flare_api/flare_api.py
pcnudde: Refactored in `a13a6738f`: `_raise_if_session_expired()` now owns the check, with callers before connect/command and immediately after command execution. This removes the repetition without changing the typed-expiry behavior. Validation after rebasing onto current `main`: 71 focused tests and 313 broader related tests passed; Black, isort, flake8, and `git diff --check` also passed.

nvflare/fuel/hci/server/binary_transfer.py
pcnudde: Yes, on this path `timeout` is defined by `FedAdminServer.__init__`: it has a `10.0` default and assigns `self.timeout`, while `create_admin_server()` always passes `server_conf.get("admin_timeout", 10.0)`. The old conditional guarded a missing `admin` object, not a missing `timeout` attribute. This download path now also requires that same authenticated admin object for `sess_mgr`, so an absent admin is not a supported case.

pcnudde:
@chesterxgchen Addressed both P2 findings in `57222bd02`:

- `SessionManager.end_session_by_id()` now isolates the expiry notification and every cancellation callback, logs each failure safely, and continues attempting the remaining cleanup.
- `AdminAPI.download_file()` removes its temporary partial file on error, and `FileTransferModule.pull_folder()` removes the local transaction directory when a file transfer fails after earlier files were downloaded.

Added focused regressions for notification/cancellation exceptions, temporary-file cleanup, and transaction-directory cleanup. Validation: 74 focused tests and 316 broader HCI/FLARE API/job-command tests passed; Black, isort, flake8, and `git diff --check` passed.

chesterxgchen (approved):

PR #5018, opened 2026-08-03
Fix external trainer liveness during streamed task ingestion

pcnudde:
## Problem

Before an external trainer can return `TASK_ACCEPTED`, it must finish the complete `TASK_READY` ingestion path: download the model payload, decode/recompose FOBS data, convert lazy values into framework tensors, and queue the materialized task.

The CJ previously suppressed heartbeat expiry only while it could see a CJ-owned download transaction. That signal is incomplete:

- In the normal pass-through path restored by #5042, the trainer downloads lazy tensor references directly from their original source using the delegated site credentials, so the CJ has no local transfer waiter.
- If task encoding does create a CJ-owned transfer for concrete or locally modified values, that waiter becomes terminal when raw bytes finish, before trainer-side decode and tensor materialization finish.

Consequently, a healthy trainer could be declared dead after a successful large-model transfer while it was still making forward progress. Raising `heartbeat_timeout` only widens the race and weakens ordinary process-failure detection.

## Fix

This PR gives the pending `TASK_READY` operation its own scoped progress contract:

- The trainer reports authenticated task-ingestion progress during payload download, per-parameter NumPy-to-PyTorch materialization, and final queue handoff.
- The CJ accepts those reports only while that trainer has an active `TASK_READY` request. Stale or out-of-scope progress does not refresh the ordinary heartbeat lease.
- The Cell wait uses the existing 600-second streaming idle timeout. Each new trainer progress report starts a fresh idle window, independent of the payload source.
- `task_wait_timeout`, when configured, remains an absolute maximum for the complete ingestion operation; `None` disables only that maximum, not the idle-stall bound.
- Transfers created by the CJ remain tracked only for deterministic cleanup on cancellation, timeout, or rejection; they are not a liveness signal.

Abort, backend closure, and real trainer process death still cancel immediately. A stalled ingestion fails with its last reported phase and idle timeout, an absolute-deadline failure is identified separately, and trainer decode/materialization exceptions continue to surface as `TASK_FAILED`. The progress scope closes in `finally` on every reply, cancellation, timeout, and exception path.

## Compatibility

- The default heartbeat interval and timeout are unchanged.
- Ordinary heartbeat behavior resumes after a real trainer reply and cannot be extended by stale task-progress messages.
- Existing explicit `task_wait_timeout` values, including `0`, remain absolute limits.
- Attach mode remains CJ-materialized and does not emit the managed-trainer progress protocol.
- No benchmark/example workaround is included.

## Validation

- Focused external-process, executor, trainer Cell API, protocol-surface, and converter tests on merged main: `323 passed`.
- Real external-process/Cell/PyTorch regression with direct pass-through download, heartbeats withheld, and delayed materialization: `1 passed`.
- Full repository style/static check: `./runtest.sh -s` passed (black, isort, flake8, agent-skill lint).

The local end-to-end test is a single-client simulation. Production-scale multinode/FSDP rank coordination, secure multi-host direct download, and 72B memory/timing behavior remain unproved locally.


nvflare/app_common/executors/client_api/external_process_backend.py
```diff
@@ -856,14 +908,21 @@ def _cancel_cause():
...
+            if not ingestion_live:
                 liveness_error = self._trainer_liveness_error(trainer)
                 if liveness_error:
                     return _SEND_SESSION_DEAD, liveness_error
             return None
 
         cancel = _TaskReadyCancelSignal(_cancel_cause)
+
+        def _has_live_task_ingestion():
+            return not cancel.triggered
```
YuanTingHsieh: **[P2] Preserve `task_wait_timeout` for inline `TASK_READY`**

`Cell._send_one_request()` treats `progress_wait_cb() == True` as permission to continue waiting after `timeout`. For an inline payload, `_ingestion_liveness()` returns false, but while the heartbeat remains healthy `_cancel_cause()` returns `None`, so this callback still returns `True`. A configured `task_wait_timeout` therefore becomes a polling interval instead of a bound, and a wedged `TASK_READY` handler can wait indefinitely while its heartbeat thread continues running. Please return `True` only while a streamed-ingestion grace is actually active, and add a regression test with an inline payload, finite `task_wait_timeout`, and continuing heartbeats.

YuanTingHsieh (commented):
The fixed post-download grace is a reasonable short-term approach, but the progress callback also changes timeout semantics for inline tasks.

nvflare/app_common/executors/client_api/external_process_backend.py
```diff
@@ -839,8 +840,10 @@ def _run_task(
                     self._current_task = None
 
     def _send_task_ready(self, trainer: _TrainerSession, task_message: dict, abort_signal: Signal) -> Tuple[str, Any]:
-        """Send TASK_READY with native cancellation for abort and liveness failures."""
+        """Send TASK_READY with native cancellation for abort, backend closure, and process death."""
         timeout = self._context.task_wait_timeout
+        if timeout is None:
+            timeout = DEFAULT_STREAMING_IDLE_TIMEOUT
```
YuanTingHsieh: **[P1] Do not base task-ingestion liveness on a CJ-local relay transaction**
This progress callback only works when encoding TASK_READY creates a CJ-owned DownloadService transaction, as with relay=True. With relay=False, the task contains pass-through LazyDownloadRefs, _on_transaction_created() is not called, and transfer_waiters remains empty. The request therefore has a fixed 600-second deadline; the reported healthy 1,432–1,639-second downloads would still time out.
Even with relay=True, the waiter becomes terminal when the raw transfer finishes and does not cover the remaining decode/recomposition/materialization before TASK_ACCEPTED. Because the callback is sampled at timeout boundaries, that phase receives anywhere from almost zero to 600 additional seconds depending on timing.
Task-ingestion liveness should be relay-independent: the trainer should report scoped progress through download/decode/materialization, while the shared Cell backend maintains an idle timeout plus an absolute maximum deadline. CJ-local transfer waiters and external-process death checks can be supplemental signals.

YuanTingHsieh (commented):
See inline comment.

nvflare/app_common/executors/client_api/external_process_backend.py
pcnudde: Confirmed. The previous deadline-only version was topology-dependent: a direct pass-through download creates no CJ waiter, and a relayed waiter ends before trainer-side materialization. Fixed in c7439eae8 (current head d7bb2e3d4): the trainer now reports authenticated TASK_PROGRESS during FOBS download, per-parameter materialization, and handoff; the backend accepts it only within the active TASK_READY scope, applies a 600s idle bound plus the existing optional task_wait_timeout absolute maximum, and keeps CJ-local waiters only as a supplemental signal. Stale progress does not refresh the heartbeat lease. Validation after merging the concurrent remote head: 223 focused tests passed, the real external-process/Cell/PyTorch regression passed, and ./runtest.sh -s passed.

pcnudde:
Superseded by #5067, rebuilt from current main as the deliberately minimal deadline-only fix. The replacement removes the progress protocol, converter hooks, phase tracking, and associated integration surface from this PR.

PR #5020, opened 2026-08-03
Fix Agent Inspector denied job authority routing

chesterxgchen:
### Description

Fix Agent Inspector existing-job routing when source-job evidence is present but cannot establish unique project authority.

- Preserve an authoritative nested generated job beside model-only root source.
- Route an independent non-secondary nested job beside an active training owner to `nvflare-orient` instead of recommending another conversion.
- Route a project that exists entirely under a secondary-tree name to `nvflare-orient` instead of treating its job evidence as absent.
- Keep historical secondary fixtures from suppressing a real root conversion target, preserving the PR #4955 regression boundary.

The change is limited to the final existing-job authority fallback. It does not restore directory-global evidence scoring or broaden static framework detection.

### Types of changes

- [ ] Documentation updated.

### Validation

- `python -m pytest -q tests/unit_test/tool/agent`: 287 passed, 1 skipped
- `./runtest.sh -s`: passed
- `./runtest.sh -l`: passed
- `git diff --check`: passed

### DIFF
--- a/nvflare/tool/agent/inspection/existing_job.py
+++ b/nvflare/tool/agent/inspection/existing_job.py
@@ -68,7 +68,13 @@ def _directory_source_state(scan: SourceScan, ownership: dict, graph: LocalImpor
             return "credible"
         if len(roots) > 1 or len(authoritative) > 1:
             return "possible"
-        return "credible" if authoritative else "none"
+        if authoritative:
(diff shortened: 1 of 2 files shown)


IsaacYangSLA (approved):
LGTM.

PR #5022, opened 2026-08-03
Fix stale global failure count in FedStats

chesterxgchen:
## Summary

This PR fixes two independent FedStats defects.

### Issue 1: stale Global failure_count

- move `failure_count` from first-round to second-round aggregation so Global uses each client final cumulative value
- evaluate `failure_count` after histogram and other requested metrics so failures discovered during round two are included
- verify the reported case produces Global count 8, histogram samples 7, and Global failure count 1

### Issue 2: redundant failure_count calls

- compare `StatisticConfig.name` values when checking whether `failure_count` is already configured
- avoid requesting `failure_count` twice in round one while still adding it when missing

## Testing

- `python -m pytest tests/unit_test/app_common/statistics tests/unit_test/app_common/workflow/statistics_controller_test.py tests/unit_test/app_common/executors/statistics/statistics_executor_test.py -q` (84 passed, 12 skipped)
- `./runtest.sh -s`


nvflare/app_common/executors/statistics/statistics_task_handler.py
```diff
@@ -62,7 +62,8 @@ def execute_task(self, task_name: str, shareable: Shareable, fl_ctx: FLContext,
             ds_features = self.get_numeric_features()
             statistics_task = shareable.get(StC.STATISTICS_TASK_KEY)
             target_statistics: List[StatisticConfig] = fobs.loads(shareable.get(StC.STATS_TARGET_STATISTICS))
-            if StC.STATS_FAILURE_COUNT not in target_statistics:
+            target_statistics.sort(key=lambda t: t.name == StC.STATS_FAILURE_COUNT)
```
YuanTingHsieh: [P2] Keep the implicit count before the final failure snapshot

`target_statistics` only contains metrics assigned to this round. Since `count` is not a second-round target, this sort puts `failure_count` after the histogram but the automatic `count()` block below still runs afterward. If a `Statistics` implementation discovers or records failures while counting, the returned and globally aggregated `failure_count` remains stale. Please insert the implicit count into `target_statistics` before sorting, or sample `failure_count` after the automatic-count block, and add a regression test for this ordering.

YuanTingHsieh (commented):
One P2 ordering issue remains in the failure-count fix.

nvflare/app_common/executors/statistics/statistics_task_handler.py
chesterxgchen: Fixed in 659ae2ea1. Implicit count is now inserted into target_statistics before failure_count is sorted last, so failures recorded by count() are included in the final snapshot. Added a second-round regression test covering that side effect; focused FedStats tests, style, and license checks pass.

IsaacYangSLA (approved):
LGTM.

PR #5027, opened 2026-08-04
Fix persistent trainer cleanup on abort

YuanTingHsieh:
## Summary

- stop a persistent `launch_once=true` external trainer before aborted task execution returns
- preserve the configured `stop_grace_period` between SIGTERM and SIGKILL during abort teardown
- add deterministic regression coverage for abort-after-result, entry abort, and grace propagation

## Root cause

Persistent trainers were only retired during `END_RUN`. Job-abort teardown can remove the Client Job process before that event reaches the executor, leaving the trainer reparented to PID 1. The abort path also reduced both the orderly shutdown wait and the separately configured SIGTERM grace to zero, causing immediate SIGKILL escalation.

The fix keeps the intended zero orderly-wait behavior on abort, but retires the owned persistent trainer before `execute()` returns and honors `stop_grace_period`. No F3 code or public configuration surface changes.

## Validation

- `103 passed`: `tests/unit_test/app_common/executors/external_process_backend_test.py`
- `114 passed`: `tests/unit_test/app_common/executors/client_api_executor_test.py`
- external-process abort matrix: 3/3 passed
  - `launch_once=true` abort-after-result: trainers terminated without orphans
  - configured grace: SIGTERM cleanup completed without force-kill escalation
  - `launch_once=false` abort-before-result control: passed
- final trainer leak audit was empty

### DIFF
--- a/nvflare/app_common/executors/client_api/external_process_backend.py
+++ b/nvflare/app_common/executors/client_api/external_process_backend.py
@@ -233,6 +233,7 @@ class ExternalProcessBackend(CellBackendBase):
                 self._send_abort(trainer, f"'{task_name}' is aborted, abort_signal_triggered")
                 if trainer is not None:
                     self._latch_abort(f"'{task_name}' aborted at entry, abort_signal_triggered")
+                self._finish_task_trainer(trainer, launch_once=True)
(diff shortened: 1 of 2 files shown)


nvkevlu (approved):
LGTM

PR #5034, opened 2026-08-04
Reject empty CollabRecipe job names

YuanTingHsieh:
## Summary

- reject empty and whitespace-only `CollabRecipe.job_name` values during construction
- use the existing `check_non_empty_str` validation helper
- add unit coverage for empty and whitespace-only names

## Motivation

An invalid job name currently propagates into `FedJob`. Exporting an empty name writes `meta.json` and `app/` directly into the caller's export directory instead of a job-specific folder, while a whitespace-only name creates an awkward whitespace-named directory.

The root cause is that `CollabRecipe` uses `check_str`, which validates only the value's type. The existing `check_non_empty_str` helper also rejects values that are empty after stripping whitespace.

After this change, invalid values raise `ValueError` naming `job_name`, consistent with the constructor's other argument validation.

## Validation

- `python -m pytest tests/unit_test/collab -q` — 96 passed
- `./runtest.sh -s --skip-install` — passed (Black, isort, flake8, agent-skill lint)
- `git diff --check` — passed


chesterxgchen (approved):
we probably should standarize on pydantic validation

PR #5036, opened 2026-08-05
Add synchronous CIFAR-10 Collab examples

ZiyueXu77:
### Description

Adds a synchronous PyTorch CIFAR-10 Collab API example with separate FedAvg,
FedProx, and SCAFFOLD implementations. The examples exchange native tensor
dictionaries through direct Collab method arguments and return values, keeping
the algorithm-specific data flow in ordinary Python and PyTorch code.

The example uses one shared data-preparation workflow and follows the standard
client/server/model/job structure. Built-in FedAvg, FedProx, and SCAFFOLD
recipes remain the recommended production path; these implementations show how
to express a custom algorithm through the Collab API.

The corrected workflows were rerun with the submitted defaults: two clients,
three rounds, one local epoch, batch size 64, learning rate 0.01, and the shared
Dirichlet split (alpha 0.5, seed 0).

| Algorithm | Round 1 | Round 2 | Round 3 |
|---|---:|---:|---:|
| FedAvg | 32.62% | 38.62% | 41.99% |
| FedProx | 25.43% | 36.05% | 42.54% |
| SCAFFOLD | 27.36% | 39.06% | 45.59% |

These replace the earlier round-5 Collab figures (34.74%, 34.65%, and 36.49%,
respectively), which are not directly comparable: the earlier benchmark used
four clients, five rounds, `ModerateCNN`, model seed 42, and the old SCAFFOLD
aggregation. The current example intentionally keeps only the lean submitted
configuration, so the table contains its complete three-point curves and no
comparison plot is included.

### Types of changes

- [ ] Quick tests passed locally by running ./runtest.sh.
- [x] Documentation updated.

### Validation

- Passed the 17 focused Collab replayable-result tests from #5183.
- Verified a replayable blocking result can be aggregated twice: model weights
  remain step-weighted and SCAFFOLD control deltas use a uniform client mean.
- Passed Python compilation, `git diff --check`, and the full repository style
  check (`./runtest.sh -s`) in the `nvflare_llm` Conda environment.
- Refreshed the Conda environment to one editable NVFlare 2.9 installation
  pointing at this worktree.
- Completed the FedAvg, FedProx, and SCAFFOLD simulator jobs with the submitted
  defaults using that exact Conda environment in a disposable GPU container.


examples/advanced/collab/pt_sync_cifar10/fedavg.py
```diff
@@ -0,0 +1,361 @@
...
+                loss.backward()
+                self.optimizer.step()
+                self._after_optimizer_step(self.optimizer)
+                batch_size = labels.size(0)
+                loss_sum += loss.item() * batch_size
+                examples_seen += batch_size
+                local_steps += 1
+            self.scheduler.step()
+
+        return loss_sum / examples_seen, local_steps
```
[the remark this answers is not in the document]
ZiyueXu77: addressed

pcnudde (commented):
Detailed read-only review of the current head. I found four medium-priority correctness, product-documentation, data-integrity, and test-coverage issues; see the inline comments. I did not find a concrete security vulnerability or an unrelated core regression. The reported CI checks are green, but I recommend addressing these findings before approval.

examples/advanced/collab/pt_sync_cifar10/scaffold.py
```diff
@@ -0,0 +1,176 @@
...
+    averaged = {}
+    for name in sorted(control_names):
+        reference = next(iter(updates.values()))["control_delta"][name]
+        accumulator = torch.zeros_like(reference, device="cpu")
+        for site_name, update in updates.items():
+            delta = update["control_delta"]
+            if set(delta) != control_names or delta[name].shape != reference.shape:
+                raise ValueError(f"{site_name} returned incompatible SCAFFOLD controls")
+            accumulator.add_(delta[name].cpu(), alpha=update["num_steps"])
+        averaged[name] = accumulator / total_weight
```
pcnudde: [P2] Clarify or correct the step-weighted SCAFFOLD variant

This computes `sum(num_steps_i * control_delta_i) / sum(num_steps_i)`. With the intentionally heterogeneous Dirichlet partitions, client step counts can differ. Canonical SCAFFOLD Algorithm 1 / equation (5) uses a client-average delta and updates `c` by the participating-client fraction: https://proceedings.mlr.press/v119/karimireddy20a.html

NVFlare's existing framework aggregator also weights controls by steps, so product parity may be intentional. If so, please explicitly document and justify this as the NVFlare step-weighted variant; otherwise use the paper's client-average update. As written, the README presents this as canonical SCAFFOLD and says the equations are implemented correctly.

examples/advanced/collab/README.md
```diff
@@ -44,14 +45,17 @@ abstraction.
 The NumPy examples run in a base installation; `hello_fedavg` needs PyTorch.
 `simple_split_learning` needs PyTorch and torchvision and downloads MNIST on
 its first run.
+`pt_sync_cifar10` needs PyTorch and torchvision; follow its
```
pcnudde: [P2] Document the distributed data topology before advertising `ProdEnv`

The catalog says above that the same recipe can be executed with `ProdEnv`, but this workflow prepares and validates `data_root` only on the submitter, embeds that absolute path in the server/client objects, and ships only Python files. In a distributed deployment, the server and clients will not automatically have the submitter's CIFAR cache and split files; the default model/metrics output is also server-local under `/tmp`.

Please document the required server and per-site data staging plus output retrieval, or scope this example's deployment guidance to the local simulator/POC modes where the filesystem assumptions hold.

examples/advanced/collab/pt_sync_cifar10/prepare_data.py
```diff
@@ -0,0 +1,157 @@
...
+        "dataset": "CIFAR10",
+        "num_clients": args.num_clients,
+        "dirichlet_alpha": args.alpha,
+        "split_seed": args.seed,
+        "train_size": len(train_dataset),
+        "test_size": len(test_dataset),
+        "site_counts": site_counts,
+        "class_counts": class_counts,
+    }
+    with manifest_path.open("w", encoding="utf-8") as stream:
```
pcnudde: [P2] Publish overwritten partitions atomically

When an old manifest exists, `--overwrite` replaces each `site-*.npy` before replacing this manifest. If regeneration stops midway—for example while changing the seed with the same client count—the old manifest plus a mixture of old/new split files still passes `validate_prepared_data`, which checks only client count and file presence. Training can then silently use overlapping or missing data despite the documented disjoint-partition guarantee.

Please stage the full split set and manifest and swap them into place atomically, or at least invalidate the old manifest before rewriting any split and validate the recorded per-site counts.

examples/advanced/collab/pt_sync_cifar10/fedavg.py
```diff
@@ -0,0 +1,365 @@
...
+    print(f"  Clients: {args.num_clients}")
+    print(f"  Rounds: {args.num_rounds}")
+    print(f"  Local epochs: {args.local_epochs}")
+    print(f"  Dirichlet alpha: {manifest['dirichlet_alpha']}")
+    for label, value in extra_summary:
+        print(f"  {label}: {value}")
+    print(f"  Data root: {args.data_root}")
+    print("=" * 80)
+
+    run = recipe_factory(args).execute(SimEnv(num_clients=args.num_clients, workspace_root=args.workspace_root))
```
pcnudde: [P2] Add recipe and algorithm smoke coverage

This PR adds three executable entry points and substantial aggregation/data-preparation logic but no tests. `tests/unit_test/collab/examples_test.py` explicitly says every catalog example must import and finalize its recipe, yet these FedAvg, FedProx, and SCAFFOLD recipes are not registered there. CI can therefore remain green if their imports, source packaging, or recipe finalization drift.

Please add catalog smoke coverage for all three recipes and focused unit tests for the aggregation and overwrite behavior.

examples/advanced/collab/pt_sync_cifar10/scaffold.py
ZiyueXu77: Explicitly documented SCAFFOLD as NVFlare’s step-weighted variant rather than canonical client averaging.

examples/advanced/collab/README.md
ZiyueXu77: Documented distributed data staging and server-local output retrieval.

examples/advanced/collab/pt_sync_cifar10/prepare_data.py
ZiyueXu77: Made partition overwrites safe by staging files, invalidating the old manifest first, and publishing the new manifest last.

ZiyueXu77: Made partition overwrites safe by staging files, invalidating the old manifest first, and publishing the new manifest last.

examples/advanced/collab/pt_sync_cifar10/fedavg.py
ZiyueXu77: Added recipe smoke coverage and focused aggregation/overwrite tests.

nvidianz (dismissed):
I think examples shouldn't have unit tests. I would remove the tests.

examples/advanced/collab/README.md
```diff
@@ -25,6 +25,7 @@ python -m collab.swarm.swarm --num-clients 3
 | `simple_split_learning` | Split learning on MNIST with client-side images and bottom model, server-side labels and top model, and direct activation/gradient exchange |
```
chesterxgchen, line 26: the readme should follow the newly established standards reference hello-pt/readme.md

examples/advanced/collab/pt_sync_cifar10/data.py
```diff
@@ -0,0 +1,129 @@
...
+        Subset(dataset, indices),
+        batch_size=batch_size,
+        shuffle=True,
+        num_workers=num_workers,
+        pin_memory=True,
+        persistent_workers=num_workers > 0,
+    )
+
+
+def make_test_loader(data_root: str | Path, batch_size: int, num_workers: int) -> DataLoader:
```
chesterxgchen: are we over engineering this data loadings. Should this be the same as any other CIFAR10 examples. Why this example adding so much complexity for data.py and preapare_data.py ?

tests/unit_test/collab/examples_test.py
```diff
@@ -30,6 +30,29 @@
 _ADVANCED_EXAMPLES_ROOT = _EXAMPLES_ROOT / "advanced"
 _HELLO_COLLAB_ROOT = _EXAMPLES_ROOT / "hello-world" / "hello-collab"
 
+_PT_SYNC_CIFAR10_ROOT = _ADVANCED_EXAMPLES_ROOT / "collab" / "pt_sync_cifar10"
+
```
chesterxgchen: not sure we need unit tests for the example

chesterxgchen (commented):
The examples should be designed for researcher to learn something, not pile of code. 

I don't like this code structure, by pile all the examples into one folder. 
why not split to them as they should be 

collab
 -- hello-fedavg
 -- hello-proxy
 -- hello-scaffold

I felt you can split the code into 
hello-fedavg/
    client.py
    server.py 
    model.py 
    job.py 
to show where the code responsibilities -- and align our existing recommended code structurs, you can one example, such as 
     fedavg_onefile.py
to demonstrate they can be all in one file. 

we don't need to do this in every example, by can add to readme to reference fedavg if the prefer one file format. 

The data engineering side seems over-engineered. This should no difference with other CIFAR10 examples in hello-pt etc. Why we have so much code

The main() also shows much more complicated argument lists. do we really need that. Our focus is demonstrate the Collab API, not complete options

chesterxgchen (changes requested):

chesterxgchen (commented):
the example's purpose to educate user how to use Collab API

this example doesn't servers that purpose. 

At any of the key line of code, there is single line of comments or explanation. 

User can run the code, but has no idea why it works and whats the key interaction points. 

Suggest to additional comments on all key API  to explain what it means what it does

-- @collab.publish  -- publish the collaboration function
-- @collab.main  -- entry point
-- collab.clients.train -- how is this related to the collab.publish its function, what it dos

examples/advanced/collab/pt_sync_cifar10/data.py
ZiyueXu77: agree, reuse the existing data prepare script from cifar10

tests/unit_test/collab/examples_test.py
ZiyueXu77: remove excessive unit tests - some AI review request those, will let coding agent ignore and only add when really necessary

examples/advanced/collab/README.md
ZiyueXu77: updated, and updated this to a hardened skill that the coding agent will use to code towards flare example

ZiyueXu77:
> The examples should be designed for researcher to learn something, not pile of code.
> 
> I don't like this code structure, by pile all the examples into one folder. why not split to them as they should be
> 
> collab -- hello-fedavg -- hello-proxy -- hello-scaffold
> 
> I felt you can split the code into hello-fedavg/ client.py server.py model.py job.py to show where the code responsibilities -- and align our existing recommended code structurs, you can one example, such as fedavg_onefile.py to demonstrate they can be all in one file.
> 
> we don't need to do this in every example, by can add to readme to reference fedavg if the prefer one file format.
> 
> The data engineering side seems over-engineered. This should no difference with other CIFAR10 examples in hello-pt etc. Why we have so much code
> 
> The main() also shows much more complicated argument lists. do we really need that. Our focus is demonstrate the Collab API, not complete options

Thanks for the input! Reorganized to pt_cifar10/ with subfolders, one thing is that we currently have several other "simple" or "hello" or single name examples which are mostly numpy - shall we organize them to a single "hello_world" folder? Specifically:
"async_aggregation" - numpy
"hello_fedavg" - random torch with single layer net
"simple_split_learning" - mnist
"swarm" - numpy

examples/advanced/collab/pt_cifar10/fedavg/server.py
```diff
@@ -0,0 +1,61 @@
...
+    def run(self) -> dict[str, torch.Tensor]:
+        device = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
+        model = SimpleNetwork().to(device)
+        test_loader = make_data_loader(train=False, batch_size=BATCH_SIZE)
+        global_weights = get_model_state(model)
+
+        for round_number in range(self.num_rounds):
+            # This calls every client's @collab.publish train method and collects results by site.
+            client_results = collab.clients.train(global_weights)
+            global_weights = aggregate_result(client_results, "weights", round_number)
```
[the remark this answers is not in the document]
ZiyueXu77: applied fix, check the update

examples/advanced/collab/pt_cifar10/README.md
```diff
@@ -0,0 +1,170 @@
+# PyTorch CIFAR-10 with the Collab API
+
+This example demonstrates how to use NVIDIA FLARE's Collab API with PyTorch to
+train a CIFAR-10 image classifier. It includes FedAvg, FedProx, and SCAFFOLD
+workflows that use direct Python calls between the server and clients.
```
holgerroth, line 5: Could we add a sentence here making the positioning explicit? FedAvg, FedProx, and SCAFFOLD already ship as built-in recipes, and those remain the recommended way to run these particular algorithms. This example re-implements them deliberately — as a reference for how to build an algorithm with the Collab API when no built-in recipe exists, with the standard recipes serving as the accuracy baseline. Without that note, the example can read as an alternative implementation path for algorithms the framework already provides.

Suggested wording:

> FedAvg, FedProx, and SCAFFOLD are available as built-in recipes, which remain the recommended way to run these algorithms in production. This example re-implements them with the Collab API as a reference for building *custom* algorithms that have no built-in recipe, using the standard recipes as a correctness and accuracy baseline.

examples/advanced/collab/pt_cifar10/scaffold/server.py
```diff
@@ -0,0 +1,52 @@
...
+        global_weights = get_model_state(model)
+        global_controls = {
+            name: torch.zeros_like(parameter, device="cpu") for name, parameter in model.named_parameters()
+        }
+
+        for round_number in range(self.num_rounds):
+            # The call maps to ScaffoldClient.train and sends both model and control state.
+            client_results = collab.clients.train(global_weights, global_controls)
+            global_weights = aggregate_result(client_results, "weights", round_number)
+            control_delta = aggregate_result(client_results, "control_delta", round_number)
```
holgerroth: The blocking group call returns a `ResultQueue`, which is a single-pass iterator (`__next__` pops from an internal queue and raises `StopIteration` once drained). The first `aggregate_result(client_results, "weights", ...)` consumes it via `dict(client_results)`, so this second call sees an empty dict, `WeightedAggregationHelper.get_result()` silently returns `{}`, and the `global_controls[name].add_(delta)` loop below is a no-op every round — the global control variate never updates and the example runs a degenerate SCAFFOLD (only the local variates act) while printing plausible accuracies.

Fix: materialize the results once and aggregate both keys from that dict, e.g.

```python
client_results = collab.clients.train(global_weights, global_controls)
results = dict(client_results)  # ResultQueue is single-pass
global_weights = aggregate_result(results, client_results.failures, "weights", round_number)
control_delta = aggregate_result(results, client_results.failures, "control_delta", round_number)
```

(or have `aggregate_result` accept an already-materialized mapping). The reported 36.49% round-5 accuracy was measured with the control variates stuck at zero, so the SCAFFOLD comparison numbers should be re-run after the fix.

examples/advanced/collab/pt_cifar10/fedavg/server.py
```diff
@@ -0,0 +1,61 @@
...
+    @collab.main
+    def run(self) -> dict[str, torch.Tensor]:
+        device = torch.device("cuda:0" if torch.cuda.is_available() else "cpu")
+        model = SimpleNetwork().to(device)
+        test_loader = make_data_loader(train=False, batch_size=BATCH_SIZE)
+        global_weights = get_model_state(model)
+
+        for round_number in range(self.num_rounds):
+            # This calls every client's @collab.publish train method and collects results by site.
+            client_results = collab.clients.train(global_weights)
```
holgerroth: This group call uses the default call timeout (`DEFAULT_CALL_TIMEOUT = 60.0` in `nvflare/collab/api/call_opt.py`), but a full local epoch over ~25k CIFAR-10 images at batch 64 can easily exceed 60s on a CPU-only machine. A timed-out site lands in `client_results.failures`, and `aggregate_result` re-raises it, so the documented `python -m collab.pt_cifar10.fedavg.job` run dies mid-round for CPU users. The other heavy-training Collab examples set this explicitly (`pt_async_cifar10` and `pt_llm_sft` both use `collab.clients(timeout=3600).train(...)`).

Suggest passing an explicit generous timeout here and in `scaffold/server.py` (FedProx reuses this server), e.g. `collab.clients(timeout=3600).train(global_weights)`.

examples/advanced/collab/pt_cifar10/prepare_data.py
```diff
@@ -0,0 +1,275 @@
...
+
+import numpy as np
+from data.cifar10_data_utils import get_site_class_summary, load_cifar10_data
+
+from nvflare.apis.event_type import EventType
+from nvflare.apis.fl_component import FLComponent
+from nvflare.apis.fl_context import FLContext
+
+
+class Cifar10DataSplitter(FLComponent):
```
holgerroth: `Cifar10DataSplitter` is dead code in this example: nothing registers it as a component, so its `START_RUN` handler never fires — the documented workflow only uses the `split_and_save` CLI path below. It also pulls `nvflare.apis` imports into an otherwise standalone data script and duplicates the summary-writing logic with a subtle inconsistency (`split()` writes and checks a `Seed:` line that `split_and_save` never writes, so a non-zero-seed CLI split would later be mistaken for seed 0 by `_summary_matches_config`).

Suggest deleting the class (and the `nvflare.apis` imports); behavior is unchanged and the inconsistency goes away with it.

examples/advanced/collab/pt_cifar10/data/cifar10_data_utils.py
```diff
@@ -0,0 +1,162 @@
...
+
+    for site, data_idx in site_idx.items():
+        unq, unq_cnt = np.unique(train_label[data_idx], return_counts=True)
+        tmp = {int(unq[i]): int(unq_cnt[i]) for i in range(len(unq))}
+        class_sum[site] = tmp
+    return class_sum
+
+
+def create_datasets(site_name, train_idx_root, central=False):
+    """To be called only after cifar10_data_split.split_and_save() downloaded the data and computed splits"""
```
holgerroth: `create_datasets` and `create_data_loaders` are never called in this example, and `data/cifar10_dataset.py` (`CIFAR10_Idx`) is only referenced by the unused `create_datasets` — the live loading path is `loader.py`, which builds its own pipeline. `prepare_data.py` only needs `load_cifar10_data` and `get_site_class_summary` from this module.

Beyond the dead weight, this is confusing: the unused code carries the cifar10/pt training transforms (per-channel stats, crop/flip augmentation) while `loader.py` actually uses `Normalize(0.5)` with no augmentation, so a reader following the README's "copied from cifar10/pt" note will assume the wrong preprocessing.

Suggest trimming this file to the two functions actually used and dropping `cifar10_dataset.py` entirely.

examples/advanced/collab/pt_cifar10/loader.py
```diff
@@ -0,0 +1,46 @@
...
+"""CIFAR-10 loading shared by the synchronous Collab examples."""
+
+from pathlib import Path
+
+import numpy as np
+from torch.utils.data import DataLoader, Subset
+from torchvision import datasets, transforms
+
+DATA_ROOT = "/tmp/cifar10"
+SPLIT_ROOT = "/tmp/cifar10_splits/pt_cifar10_2sites_alpha0.50_seed0"
```
holgerroth: `SPLIT_ROOT` hardcodes the `2sites_alpha0.50_seed0` directory name, while `prepare_data.py` derives its output dir from `--num_sites`/`--alpha`/`--seed` and the README tells users to keep `--num_sites` in sync with `NUM_CLIENTS` and to tune `--alpha`. If a user changes those (e.g. NUM_CLIENTS=4 and prepares a 4-site split), the loader still reads the old 2-site directory: site-3/site-4 fail with `FileNotFoundError`, and — worse — if the stale dir exists, site-1/site-2 silently train on partitions from the wrong split.

Suggest deriving `SPLIT_ROOT` from the same parameters as `prepare_data.py` (or passing it per site via `recipe.set_per_site_config` / `collab.get_app_prop`, as `hello_fedavg` demonstrates).

examples/advanced/collab/pt_cifar10/scaffold/server.py
```diff
@@ -0,0 +1,52 @@
...
+            for name, delta in control_delta.items():
```
holgerroth, line 46: Separate from the double-iteration issue above: `aggregate_result` weights every contribution by `num_steps`, which is right for model weights but not for the control variates. SCAFFOLD's server update is c ← c + (|S|/N)·mean(Δc_i) — a *uniform* mean over participating clients. With the alpha=0.5 Dirichlet split the two sites have unequal sample counts and therefore unequal `num_steps`, so a step-weighted average biases the control update toward the larger site. (Today this is masked by the empty second aggregation; once that's fixed, this becomes the live behavior.)

Suggest aggregating `control_delta` with weight 1 per client (uniform mean), or noting the deviation in the README if the weighted variant is intentional.

examples/advanced/collab/pt_cifar10/README.md
ZiyueXu77: Addressed in 568016ea. The introduction now states that the built-in FedAvg, FedProx, and SCAFFOLD recipes remain the recommended production path, and positions these implementations as Collab API references for custom algorithms, with the standard recipes as correctness and accuracy baselines.

examples/advanced/collab/pt_cifar10/scaffold/server.py
ZiyueXu77: Addressed in 568016ea. Added materialize_results(), which propagates any group-call failure and then consumes the single-pass ResultQueue exactly once into a dictionary. SCAFFOLD now reuses that mapping for both model-weight and control-delta aggregation, so the global control variate is updated. No comparison accuracy is committed in this PR; future measurements will use the corrected workflow.

examples/advanced/collab/pt_cifar10/fedavg/server.py
ZiyueXu77: Addressed in 568016ea. Added a shared TRAIN_TIMEOUT of 3600 seconds. The FedAvg server, which is also reused by FedProx, and the SCAFFOLD server now call collab.clients(timeout=TRAIN_TIMEOUT).train(...).

examples/advanced/collab/pt_cifar10/prepare_data.py
ZiyueXu77: Addressed in 568016ea. Removed Cifar10DataSplitter and the nvflare.apis event/component imports. prepare_data.py now contains only the standalone CLI partitioning and persistence path used by the documented workflow.

examples/advanced/collab/pt_cifar10/data/cifar10_data_utils.py
ZiyueXu77: Addressed in 568016ea. Trimmed cifar10_data_utils.py to the two live helper functions plus the shared data paths, deleted the unused cifar10_dataset.py module, and updated the README to describe the partitioning code as adapted from the existing CIFAR-10 example rather than implying that its unused training pipeline is copied.

examples/advanced/collab/pt_cifar10/loader.py
ZiyueXu77: Addressed in 568016ea. Data preparation and the client loader now use the same SPLIT_ROOT constant from cifar10_data_utils.py. Prepared site files are always written to /tmp/cifar10_splits/pt_cifar10, so changing num_sites, alpha, or seed no longer changes the directory selected by clients.

examples/advanced/collab/pt_cifar10/scaffold/server.py
ZiyueXu77: Addressed in 568016ea. aggregate_result keeps local-step weighting as the default for model weights, while the SCAFFOLD control-delta call sets weight_by_steps=False for a uniform client mean. The README now documents this distinction.

holgerroth (approved):
All review findings have been addressed, several at the framework level rather than in the example:

- The single-pass result iteration issue was fixed properly in #5183 (blocking group calls now return a replayable snapshot), and this example was updated to aggregate model weights and SCAFFOLD control deltas from the same result.
- Explicit 3600s call timeout, shared `SPLIT_ROOT` constant between splitter and loader, dead code removed, and SCAFFOLD control deltas now use a uniform client mean, documented in the README.
- The benchmark table was re-measured on the corrected workflow with the submitted defaults, and the earlier incomparable figures were clearly retired.
- The README positions the built-in recipes as the recommended production path, with this example as a Collab API reference for custom algorithms.

Nice example of the Collab contract — SCAFFOLD passing its control variate as a plain second argument makes the case well. LGTM.

examples/advanced/collab/pt_cifar10/data/cifar10_data_utils.py
```diff
@@ -0,0 +1,70 @@
+# Copyright (c) 2023, NVIDIA CORPORATION.  All rights reserved.
```
chesterxgchen, line 1: this is already cifar10 folder, can we just called data_utis.py

chesterxgchen (approved):
add minor suggestions, LGTM

examples/advanced/collab/pt_cifar10/data/cifar10_data_utils.py
ZiyueXu77: sounds good! let's merge it first, will address in another PR

PR #5039, opened 2026-08-05
Fix F3 reassembly buffer sizing under load

nvidianz:
Jira: [FLARE-3093](https://jirasw.nvidia.com/browse/FLARE-3093)

## What changed

- derive the receiver's out-of-sequence chunk allowance from the stream window and chunk size instead of using a fixed default of 16
- cap peer-derived buffer sizing while preserving larger operator-configured limits
- repeat only `CHUNK_SIZE` and `WINDOW_SIZE` on data frames so a receiver can size its buffer even when sequence 0 is delayed by frame-pool scheduling
- adopt those sender parameters before buffering pre-sequence-0 chunks
- stop normal sends when the flow-control window reaches its limit
- remove the receiver's unused retry-pending field and stop transmitting that receiver-irrelevant parameter

## Why

FLARE-3093 reports healthy F3 streams aborting under CPU load with `Too many out-of-sequence chunks: 16`. `ConnManager` processes inbound frames in a thread pool, so a sender's in-order frames can reach `RxTask` out of order. The default 64 MiB window allows substantially more than the fixed 16-chunk receiver tolerance.

Sequence 0 carries stream negotiation parameters, but it can itself be delayed behind later frames. Repeating the two parameters needed for buffer sizing lets the receiver establish the correct limit before admitting those later chunks without changing header semantics for older receivers.

## Impact

Large F3 transfers no longer fail solely because scheduling jitter reorders more than 16 chunks. Explicit sender window sizes are honored before sequence 0 is processed, and anomalous peer-derived sizing remains bounded.

## Validation

- `python -m pytest tests/unit_test/fuel/f3/streaming -q` — 374 passed
- scoped Black, isort, and flake8 checks on the four changed files
- deterministic regression coverage for 66 later chunks processed before sequence 0 with a 128 MiB sender and default 64 MiB receiver

### DIFF
--- a/nvflare/fuel/f3/streaming/byte_receiver.py
+++ b/nvflare/fuel/f3/streaming/byte_receiver.py
@@ -45,7 +45,7 @@ MIN_OUT_SEQ_CHUNKS = 16
 # ratio would let a peer size this receiver's reassembly buffer at will. A site
 # that legitimately needs more can raise streaming_max_out_seq_chunks in its own
 # config, which is not capped.
-MAX_DERIVED_OUT_SEQ_CHUNKS = 1024
+MAX_PEER_DERIVED_OUT_SEQ_CHUNKS = 1024
 # Headerless legacy senders can use windows smaller than the new sender
 # default. Keep their receiver-side fallback at the historical 4 MiB so an
 # ACK is sent before those senders exhaust their flow-control window.
@@ -88,13 +88,13 @@ def required_out_seq_chunks(window_size: int, chunk_size: int) -> int:
     the FLARE-3093 regression, where a 64 MiB default window put far more than 16
     chunks in flight.
 
-    The result is capped at MAX_DERIVED_OUT_SEQ_CHUNKS because window_size and
+    The result is capped at MAX_PEER_DERIVED_OUT_SEQ_CHUNKS because window_size and
     chunk_size may be adopted from peer-supplied headers.
     """
     if chunk_size <= 0:
         return MIN_OUT_SEQ_CHUNKS
     window_chunks = (window_size + chunk_size - 1) // chunk_size
-    return max(MIN_OUT_SEQ_CHUNKS, min(window_chunks + 1, MAX_DERIVED_OUT_SEQ_CHUNKS))
+    return max(MIN_OUT_SEQ_CHUNKS, min(window_chunks + 1, MAX_PEER_DERIVED_OUT_SEQ_CHUNKS))
 
 
 class RxTask:
@@ -319,11 +319,11 @@ class RxTask:
         )
         if self.chunk_size > 0:
             window_chunks = (self.window_size + self.chunk_size - 1) // self.chunk_size
-            if window_chunks + 1 > MAX_DERIVED_OUT_SEQ_CHUNKS:
+            if window_chunks + 1 > MAX_PEER_DERIVED_OUT_SEQ_CHUNKS:
                 log.warning(
                     f"{self} streaming_window_size {self.window_size} from {self.origin} needs "
                     f"{window_chunks + 1} out-of-sequence chunks, above the "
-                    f"{MAX_DERIVED_OUT_SEQ_CHUNKS} cap; raise streaming_max_out_seq_chunks on this "
+                    f"{MAX_PEER_DERIVED_OUT_SEQ_CHUNKS} cap; raise streaming_max_out_seq_chunks on this "
                     f"site if this peer's window is legitimate"
                 )
         required_max_out_seq = required_out_seq_chunks(self.window_size, self.chunk_size)
--- a/nvflare/fuel/f3/streaming/byte_receiver.py
+++ b/nvflare/fuel/f3/streaming/byte_receiver.py
@@ -41,6 +41,12 @@ from nvflare.fuel.f3.streaming.stream_utils import ONE_MB, stream_stats_category
 log = logging.getLogger(__name__)
 
 MAX_OUT_SEQ_CHUNKS = 16
+# Upper bound on the out-of-sequence tolerance derived from a peer's
+# WINDOW_SIZE/CHUNK_SIZE headers. Those headers are hints, and an unbounded
+# ratio would let a peer size this receiver's reassembly buffer at will. A site
+# that legitimately needs more can raise streaming_max_out_seq_chunks in its own
+# config, which is not capped.
+MAX_DERIVED_OUT_SEQ_CHUNKS = 1024
 # Headerless legacy senders can use windows smaller than the new sender
 # default. Keep their receiver-side fallback at the historical 4 MiB so an
 # ACK is sent before those senders exhaust their flow-control window.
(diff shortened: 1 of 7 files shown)


nvflare/fuel/f3/streaming/byte_receiver.py
```diff
@@ -68,6 +74,34 @@ def _abbrev(value, limit: int = 64) -> str:
...
+    could therefore send one additional full chunk. Both cases matter during a
+    rolling upgrade.
+
+    The result is capped at MAX_DERIVED_OUT_SEQ_CHUNKS because window_size and
+    chunk_size may be adopted from peer-supplied headers.
+    """
+    if chunk_size <= 0:
+        return MAX_OUT_SEQ_CHUNKS
+    window_chunks = (window_size + chunk_size - 1) // chunk_size
+    return max(MAX_OUT_SEQ_CHUNKS, min(window_chunks + 1, MAX_DERIVED_OUT_SEQ_CHUNKS))
```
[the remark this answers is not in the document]
nvidianz: Thanks. The buffer counts only out-of-sequence chunks, not every in-flight frame. Whenever reassembly is blocked, the expected sequence itself is absent from out_seq_chunks. For a legacy sender with a non-multiple window, at most ceil(window/chunk) normal frames plus one final frame can be in flight, so at most ceil(window/chunk) entries can be buffered; the current ceil(window/chunk) + 1 allowance is already conservative by one. For a divisible window, the legacy sender can add one extra normal frame and one final frame, producing window/chunk + 2 in-flight frames, but only window/chunk + 1 of them can be out of sequence, exactly matching the current allowance. No code change is needed.

nvidianz:
@greptileai

nvflare/fuel/f3/streaming/byte_receiver.py
```diff
@@ -68,6 +74,34 @@ def _abbrev(value, limit: int = 64) -> str:
...
 
 
+def required_out_seq_chunks(window_size: int, chunk_size: int) -> int:
+    """Return the out-of-sequence tolerance needed for a sender's flow-control window.
+
+    Inbound frames are dispatched to a thread pool (ConnManager.process_frame), so
+    chunks that arrive in order can still be *processed* out of order, with a depth
+    that grows under scheduling jitter. The depth is bounded by how many chunks the
+    sender may have unacked, which is what its flow-control window governs:
+    TxTask.send_loop blocks while ``window >= window_size`` and every non-final chunk
```
YuanTingHsieh: [P1] **Account for short reads when deriving the frame-count limit**

This derivation assumes every non-final frame is `chunk_size` bytes, but `Stream.read(size)` explicitly permits shorter reads. In `TxTask.send_loop`, when the next read would overflow the pending buffer, the current partial buffer is sent as a non-final frame. Alternating reads of `chunk_size - 1` and `2` can therefore produce nearly twice as many frames within the same byte window (about 127 normal frames for the default 64 MiB/1 MiB settings), while this calculation permits only 65. A conforming stream can still hit `Too many out-of-sequence chunks` when frame processing is delayed. Please either coalesce/split reads so non-final frames are full-sized or make reassembly protection account for actual frame/byte usage, and add a short-read regression test.

nvflare/fuel/f3/streaming/byte_receiver.py
```diff
@@ -284,6 +305,48 @@ def _handle_new_stream(self, message: Message):
...
+        if self.chunk_size > 0:
+            window_chunks = (self.window_size + self.chunk_size - 1) // self.chunk_size
+            if window_chunks + 1 > MAX_DERIVED_OUT_SEQ_CHUNKS:
+                log.warning(
+                    f"{self} streaming_window_size {self.window_size} from {self.origin} needs "
+                    f"{window_chunks + 1} out-of-sequence chunks, above the "
+                    f"{MAX_DERIVED_OUT_SEQ_CHUNKS} cap; raise streaming_max_out_seq_chunks on this "
+                    f"site if this peer's window is legitimate"
+                )
+        self.max_out_seq = max(self.max_out_seq, required_out_seq_chunks(self.window_size, self.chunk_size))
```
YuanTingHsieh: [P2] **Preserve an explicit receiver-side maximum**

`self.max_out_seq` may contain an explicitly configured `streaming_max_out_seq_chunks`, but this `max()` lets peer-controlled headers raise that value (the new test codifies configured `16` becoming `129`). That contradicts the documented maximum and removes an operator's memory-control setting. Please distinguish an absent setting from an explicit cap. If overriding smaller values is intentional, rename and document the setting as a minimum; similarly, `MAX_OUT_SEQ_CHUNKS` now behaves as `MIN_OUT_SEQ_CHUNKS`, while `MAX_PEER_DERIVED_OUT_SEQ_CHUNKS` would describe the 1024 ceiling more clearly.

nvflare/fuel/f3/streaming/byte_streamer.py
```diff
@@ -285,12 +285,16 @@ def send_loop(self):
...
             window = self.offset - self.offset_ack
-            # It may take several ACKs to clear up the window
-            while window > self.window_size:
-                log.debug(f"{self} window size {window} exceeds limit: {self.window_size}")
+            # It may take several ACKs to clear up the window.
+            # The comparison is >=, not >, so a normal data chunk does not add
+            # another full chunk once the window reaches its limit. The EOS path
+            # above can still send one final frame; RxTask includes that slot when
+            # sizing its reassembly buffer.
+            while window >= self.window_size:
```
YuanTingHsieh: [P2] **Avoid deadlocking a zero-sized window before the first send**

`CommConfigurator` currently accepts `streaming_window_size=0`. Previously the `>` comparison allowed the first frame and then provided stop-and-wait behavior. With `>=`, the initial `window == 0` waits before any frame has been sent, so no ACK can ever open the window and a multi-read stream times out. Changing the unit test from zero to one masks this compatibility regression. Please either reject non-positive windows in every production configuration path or special-case zero to preserve the prior behavior.

YuanTingHsieh (commented):
Review identified three remaining correctness and configuration issues; details are inline.

nvflare/fuel/f3/streaming/byte_receiver.py
nvidianz: Fixed in 816f01995. TxTask now requests only the remaining pending-buffer capacity and keeps a full pending buffer until a non-empty lookahead read, so every non-final frame is chunk-sized. Added a conforming short-read regression that alternates 7-byte and 2-byte reads with an 8-byte chunk size and verifies frame sizes [8, 8, 8, 1].

nvflare/fuel/f3/streaming/byte_receiver.py
nvidianz: Fixed in 816f01995 and 012fba593. An absent setting now derives the required tolerance, while an explicitly configured streaming_max_out_seq_chunks remains authoritative and logs a warning when smaller than the peer-derived requirement. Renamed the bounds to MIN_OUT_SEQ_CHUNKS and MAX_PEER_DERIVED_OUT_SEQ_CHUNKS. Tests cover absent, explicit-small, explicit-large, and derived-cap cases.

nvflare/fuel/f3/streaming/byte_streamer.py
nvidianz: Fixed in 816f01995. Restored the strict window > window_size guard, preserving the existing zero-window stop-and-wait behavior. The watchdog regression now uses streaming_window_size=0 again and verifies that ACK progress allows the stream to complete.

YuanTingHsieh:
[P1] **Update the shipped configurations so they do not pin the old limit**

The receiver now correctly treats an explicitly configured `streaming_max_out_seq_chunks` as authoritative. However, the checked-in [25-Gbit benchmark configuration](https://github.com/NVIDIA/NVFlare/blob/012fba5937a44b17e4aa08ad9f9cca00beed7340/dev_tools/f3/comm_config.yml#L18-L22) still combines a 64 MiB window and 1 MiB chunks with `streaming_max_out_seq_chunks: 16`. That makes `max_out_seq_is_configured` true, so the receiver remains capped at 16 instead of deriving 65—the exact mismatch behind FLARE-3093. The [benchmark README](https://github.com/NVIDIA/NVFlare/blob/012fba5937a44b17e4aa08ad9f9cca00beed7340/dev_tools/f3/README.md#L152-L158) and [communication guide](https://github.com/NVIDIA/NVFlare/blob/012fba5937a44b17e4aa08ad9f9cca00beed7340/docs/user_guide/admin_guide/configurations/communication_configuration.rst#L245-L264) also still advertise 16 as the default.

Please remove this setting from the benchmark/default examples so the value is derived automatically (preferred), or set it to at least 65 for the current 64 MiB/1 MiB defaults. The documentation should explain that 16 is the minimum fallback and that an unset limit is derived from the effective window and chunk sizes.

nvidianz:
@YuanTingHsieh Addressed in 3a887a5a1. Removed streaming_max_out_seq_chunks from the shipped 25-Gbit benchmark configuration so the 64 MiB window / 1 MiB chunk settings derive the required 65-chunk tolerance. Updated the benchmark README and communication guide to document the derived limit, 16-chunk minimum fallback, 1024 peer-derived ceiling, and explicit-value semantics. I also added coverage for the oversized Stream.read result validation; the full F3 streaming suite passes (378 tests).

YuanTingHsieh (approved):

PR #5040, opened 2026-08-05
Fix Lightning DDP result sending from nonzero ranks

YuanTingHsieh:
## What changed

- Check the active Lightning trainer's runtime global rank at the training and evaluation hook boundaries.
- Call the model-only federated result-send helper only when that runtime global rank is 0.
- Add hook-level regression coverage for both rank 0 and rank 1 across training and evaluation result paths.

## Why

With PyTorch Lightning `strategy="ddp"` and multiple devices, callback hooks execute on every DDP process. The callback guarded model receive, but its result-send path was unguarded. A nonzero rank could therefore send a result without owning the matching Client API receive/result transaction, causing `TASK_ABORTED`, an invalid DXO at the server, and FedAvg failure.

Checking `trainer.global_rank` at the callback boundary ensures only the process that owns the federated transaction calls the Client API send helper. The helper remains concerned only with sending the supplied model, while nonzero ranks continue the remaining callback cleanup and Lightning/DDP participation.

## Impact

PyTorch Lightning applications using multi-GPU DDP can now complete federated rounds without nonzero ranks issuing unmatched result sends. The fix covers both simulator API and CLI execution paths.

## Validation

- `tests/unit_test/app_opt/lightning/api_test.py`: 87 passed.
- Scoped flake8, Black, isort, and `git diff --check`: passed.
- Manual end-to-end validation with two simulated clients and two-GPU NCCL DDP per client:
  - Both simulator API and CLI paths passed.
  - All federated result sends came from runtime global rank 0.
  - Nonzero ranks issued no Client API result sends.
  - FedAvg completed without task-abort, invalid-result, or minimum-response failures.


nvflare/app_opt/lightning/api.py
```diff
@@ -271,7 +271,7 @@ def on_validation_end(self, trainer, pl_module):
         if pl_module and self.metrics is None:
             self.metrics = _extract_metrics(trainer.callback_metrics)
             if self._is_evaluation:
-                self._send_model(FLModel(metrics=self.metrics))
+                self._send_model(FLModel(metrics=self.metrics), trainer)
```
chesterxgchen: trainer has wide scope than rank, but _send_model() is really only need rank. 

In general case, if the function should ask more than what it needs, unless this is intended for future expansion. 

For symantic point of view, send model requires trainer doesn't sounds right

chesterxgchen (commented):
add some comments

nvflare/app_opt/lightning/api.py
YuanTingHsieh: good suggestion, fixed.

chesterxgchen (approved):

PR #5042, opened 2026-08-05
Align Client API pass-through and Attach trust boundaries for 2.9

YuanTingHsieh:
## Summary

- remove the unreleased Client API lazy-download relay code from the 2.9 development branch; released 2.8 had no such Client API relay, and 2.9 will ship without one
- keep managed `external_process` results as ordinary Cell pass-through, preserving the trainer's original source FQCN, ref ID, and DownloadService receiver accounting
- make network `attach` use the site's existing Client Parent (CP) listener, matching the physical connection and server-facing materialization boundary of the 2.8 `IPCExchanger`/`IPCAgent` design
- keep the Client Job (CJ) as the Attach task/result materialization boundary; an attached trainer never receives the site's server `AUTH_TOKEN` or `AUTH_TOKEN_SIGNATURE`
- protect secure network Attach messages end to end through the CP with the provisioned site Cell identity, with shared pre-decode enforcement for streamed and direct Client API/DownloadService routes
- propagate the protected FOBS context to legitimate `ViaDownloader` chunk requests without forcing a second key exchange against a managed trainer on a forwarded cross-site pull
- keep protected shared-file Attach as a job-owned CJ-to-trainer route for network-isolated trainers while CP-to-CJ remains normally networked
- keep Attach externally owned: NVFlare closes only the session and never signals, kills, or waits on the trainer process
- require Attach heartbeat liveness and retire a dead accepted-result source only after a bounded confirmed-disconnect grace, preserving transient reconnects
- update focused security, provenance, ownership, lifecycle, Attach-topology, and secure nested-Swarm coverage

## Why

During 2.9 development, the Client API gained code that could materialize a trainer lazy reference and create another CJ-owned DownloadService transaction while rewriting the source reference. That behavior was not present in released 2.8. This PR removes it before the 2.9 release.

The two out-of-process modes intentionally use different server-facing boundaries:

- `external_process` is launched and owned by the CJ. It preserves ordinary pass-through so the real trainer remains the DownloadService source for the workflow's receiver.
- `attach` is independently started and externally owned. It follows the safer 2.8 containment boundary: the CJ receives and fully materializes the trainer result. Sending that concrete result onward can create an ordinary new CJ-owned DownloadService transaction, but the trainer reference never crosses the CJ boundary.

For Attach, this intentionally makes CJ peak memory scale with the complete task/result object plus serialization buffers; the configured DownloadService chunk size is a transport chunk and is not a hard peak-RSS bound for one large tensor. This is the accepted 2.9 tradeoff relative to the unreleased lazy/relay implementation on `main`, which used a different trust model. The 2.10 scoped trainer identity, DownloadService ACL, and revocation work is the prerequisite for restoring lazy Attach transfer without delegating broad site authority.

### Managed external process

This reuses the NVFlare 2.8 CellPipe authentication model. After successful launch-token `HELLO` authentication, `HELLO_ACCEPTED` delivers the site's `AUTH_TOKEN` and `AUTH_TOKEN_SIGNATURE`; they are never written to the bootstrap file. The trainer installs the normal outgoing site auth-header filters, and its descendant FQCN satisfies the existing server origin binding.

The managed trainer's same-host CJ-local Cell hop has Cell transport security disabled. The launch token authenticates `HELLO_ACCEPTED`, but the delegated bearer credential is not TLS-encrypted on that hop; confidentiality relies on the host/OS trust boundary and the owner-only bootstrap file. This improves on 2.8's on-disk credential export, while the 2.10 scoped-identity work will remove the full bearer-token handoff.

### Attach

Network Attach follows the 2.8 `IPCExchanger`/`IPCAgent` topology and materialization boundary:

- the trainer connects to the site's existing stable CP listener, not to a dynamic CJ-owned network listener
- the trainer is a stable CP child, `<cp_fqcn>.-client_api_<attach_id>`
- the CP physically routes Cell messages; the dynamic CJ remains the application-level session, task, result, and materialization endpoint
- secure network Attach reuses the provisioned site CA/client certificate for Cell peer identity, including custom site certificate CNs
- a shared CJ CoreCell pre-decode guard covers streamed envelopes and direct Client API/DownloadService routes, rejects unclaimed CP-child origins, and requires both Cell `SECURE` and `ENCRYPTED` protection from the bound trainer before FOBS decoding
- concurrent Attach sessions hold independent guard claims; releasing one claim does not disable another session's protection
- `ViaDownloader` chunk requests inherit the protected decode context, so large tensors remain protected across the complete trainer-to-CJ transfer
- `SESSION_OPEN` authenticates and binds the trainer to one dynamic CJ session, but never carries the site's server bearer token or signature
- the trainer installs no server auth-header filters and cannot issue authenticated server-facing requests
- no CJ listener certificate, key, port, or `listening_host` provisioning is required
- protected shared-file Attach remains CJ-owned and supports a trainer with network access denied
- `allow_insecure_attach` remains a deprecated compatibility argument with no effect; unsafe shared-file routes are rejected
- `heartbeat_timeout` must be positive: heartbeat expiry is the terminal fallback when the externally owned trainer misses protocol `SHUTDOWN`
- with `allow_reconnect=True`, a live accepted-result source survives transient route loss but is retired after a continuous confirmed-disconnect grace so a fresh session can attach

Delegating the site's client certificate remains broad site-level Cell authority. For 2.10, replace it with a short-lived scoped trainer identity, explicit DownloadService ACLs, and revocation before enabling direct Attach trainer-to-server pass-through.

The launch token and Attach ID are session-binding/rendezvous values, not FL authentication credentials.

## Validation

Definitive signed head `47df37977ac6e9624105d21131292bf8bb3b7045`, based on `upstream/main` `509ce700fb9b1cc3d0e29bb1bd6216949303e77b`:

- fresh exact-SHA Colossus qualification: **20/20 PASS**, all `FINISHED:COMPLETED`
  - 1 server, 3 clients, 2 rounds, 120/120 trainer-round records
  - FedAvg and Swarm; mixes A/B/C; gRPC, HTTP, and protected shared-file Attach; tensor streaming Off/On
  - every checkpoint correct
  - all 10 Off cases created zero lazy tensor transactions
  - all 10 On cases exercised 160 real FOBS/`ViaDownloader` transactions; every pull and terminal outcome completed, with zero failures
  - 32 managed trainer-origin result transactions retained the original trainer FQCN/ref and settled directly at the ultimate receiver; zero intermediate CJ result transactions
  - 372 accepted managed cross-root authentication events
  - Attach created 20 trainer-to-CJ transactions materialized at the CJ and 20 ordinary CJ onward transactions
  - 20 Attach `SESSION_OPEN` messages carried no site bearer/signature; trainers installed no server auth filters and issued no server requests
  - 465 protected Attach network chunks and zero unprotected chunks
  - all 20 Attach trainers remained externally owned, received zero NVFlare OS signals, closed cleanly with rc=0, and left zero leaks
  - all four shared-file cases completed with trainer network access denied while CP-to-CJ remained networked
- supplemental exact-SHA transport-hardening qualification: **16/16 PASS**
  - FedAvg and Swarm; mixes A/B; gRPC and HTTP; ViaDownloader Off/On; 96/96 trainer-round records and correct checkpoints
  - Attach used actual transport mTLS plus secure Cell mode: `grpcs://127.0.0.1:18422` and `https://127.0.0.1:18423`
  - CA-signed CP certificate chain/key/SAN verified; bilateral TLS 1.3 with AES-256-GCM succeeded
  - fail-closed probes without a client certificate were rejected by both gRPC and HTTP before `SESSION_OPEN`
  - all 8 Off cases created zero lazy transactions; all 8 On cases exercised 125 real ViaDownloader transactions with complete outcomes
  - 16 bearer-free `SESSION_OPEN` bindings, 32 certificate identity matches, 465 SECURE+ENCRYPTED chunks, and zero unprotected chunks
  - managed original-source settlement, Attach CJ materialization/onward serialization, no trainer server authority, and all ownership/cleanup gates passed
- case 12's initial external-audit failure was a classifier false positive: Swarm selected the managed trainer's same-site CJ as receiver, so no cross-root authentication event applied; the two distinct CJ transactions were normal aggregation broadcasts, not intermediate copies. The preserved run was reclassified without rerun or product changes.
- #5027 overlap passed: 3 focused unit tests and 3/3 runtime abort/lifecycle jobs, including grace handling and persistent `launch_once=true`; all terminated their process groups with no survivors
- delayed audit after more than 300 seconds found no trainer, CJ, job, PPID-1 orphan, tested listener, deleted-CWD process, or SIGKILL escalation
- no product or F3 changes were made by the qualification harness

Current-head local validation:

- 493 targeted Client API, Attach, Swarm, FOBS, and overlapping persistent-trainer regression tests passed (5 skipped)
- all 10 real Cell/subprocess Client API integrations passed
- `./runtest.sh -s` passed: Black (2,611 files unchanged), isort, flake8 (0 findings), and agent-skill lint
- `git diff --check` passed
- all eight PR commits have verified GPG signatures


nvflare/client/cell/api.py
```diff
@@ -301,22 +302,53 @@ def _hello(self) -> None:
...
 
+    def _install_site_auth_headers(self, secure_mode, auth_token=None, token_signature=None) -> None:
+        if type(secure_mode) is not bool:
+            raise TrainerSessionError(f"session secure_mode must be a bool, got {secure_mode!r}")
+        if self._session_security_configured:
+            if secure_mode != self._secure_mode:
+                raise TrainerSessionError("session secure_mode changed after authentication")
+            return
+        if not self._is_attach and secure_mode != self._secure_mode:
+            raise TrainerSessionError("HELLO_ACCEPTED secure_mode disagrees with the launch bootstrap")
```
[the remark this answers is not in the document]
YuanTingHsieh: Addressed in `4edda63a025b04cd14e118bf3388308afcc248a5`. The external-process `HELLO` path now uses `body.get(MsgKey.SECURE_MODE, False)`, preserving protocol-v1 compatibility when a non-secure CJ omits the field. A secure bootstrap still sees the missing field as `False` and fails the existing secure-mode mismatch check; present invalid values remain rejected. Attach does not use this credential-installation path and receives no site credential. Added `test_non_secure_init_accepts_protocol_v1_reply_without_secure_mode` and `test_secure_init_rejects_protocol_v1_reply_without_secure_mode`; the focused suite passes.

chesterxgchen (commented):
add review comments

chesterxgchen:
## Focused review: CJ memory usage and Swarm behavior

### P1: Attach + Swarm can bypass the CJ materialization boundary

[`_learn_executor_accepts_lazy_refs()`](https://github.com/NVIDIA/NVFlare/blob/7e855d73cb97d25dea069c0634c90ace7db09392/nvflare/app_common/ccwf/swarm_client_ctl.py#L792-L822) treats `ExecutionMode.ATTACH` like managed `EXTERNAL_PROCESS`.

With `enable_tensor_disk_offload=True`, [`send_learn_task()` marks the task for pass-through](https://github.com/NVIDIA/NVFlare/blob/7e855d73cb97d25dea069c0634c90ace7db09392/nvflare/app_common/ccwf/swarm_client_ctl.py#L609-L614). On a non-aggregation client, `_prepare_learn_task_data()` then leaves the `LazyDownloadRef` unchanged. When that task is serialized to the attached trainer, [`LazyDownloadRef` re-emits the original source FQCN/ref ID](https://github.com/NVIDIA/NVFlare/blob/7e855d73cb97d25dea069c0634c90ace7db09392/nvflare/fuel/utils/fobs/decomposers/via_downloader.py#L335-L364).

This contradicts the documented Attach contract: CJ must fully materialize the payload, and an attached trainer must not consume an original remote reference. In a secure Swarm topology it may also fail authorization because the attached trainer does not receive the server authentication token.

Recommended fix:

- Exclude `ExecutionMode.ATTACH` from the lazy-reference handoff, or enforce equivalent materialization in `AttachBackend`.
- Add a secure Attach + Swarm + `enable_tensor_disk_offload=True` test asserting that the original source/ref does not reach the trainer and that CJ owns the onward transaction.

### P1: Attach changes CJ memory from relay/lazy-reference usage to model-sized materialization

The transfer behavior compared with the PR base is:

| Path | PR #5042 CJ behavior | Current `main` CJ behavior |
|---|---|---|
| Managed external, server → trainer | Lazy pass-through; trainer downloads from server | Lazy/direct or secure CJ relay |
| Managed external, trainer → server/peer | Lazy pass-through; receiver downloads from trainer | Lazy/direct or secure CJ relay |
| Attach, server/peer → trainer | CJ fully materializes, then creates an ordinary CJ source | Lazy/direct or secure CJ relay |
| Attach, trainer → server/peer | CJ fully materializes, then may create a new CJ source | Lazy/direct or secure CJ relay |

Therefore, managed `external_process` does **not** materialize the tensor payload in CJ in either direction. Attach **does** materialize it in CJ in both directions under this PR. That is required by the PR's selected trust boundary, not by `DownloadService` itself.

This is a substantial memory change from current `main`: CJ now holds at least the complete task/result object, plus serialization buffers while publishing it onward. `ArrayDownloadable` and `TensorDownloadable` serialize whole array/tensor items, so the configured chunk size is not a hard peak-RSS bound when one tensor is large.

The added 4 MiB integration case validates transaction ownership but cannot detect multi-GiB CJ OOM behavior. For Attach Swarm, every training CJ becomes a full task/result materialization point. Aggregator disk offload can reduce memory at the destination aggregator, but it does not eliminate the source-side trainer-to-CJ materialization.

Recommended product/implementation gate:

- Provide CJ-owned disk staging or another bounded-memory re-ownership path; **or**
- Explicitly accept and document the model-sized CJ memory requirement and add a realistic peak-memory regression test, including a model with one large tensor.

I would not approve the PR until the Swarm trust-boundary defect is fixed and the Attach memory regression is explicitly resolved or accepted.

nvflare/app_common/executors/client_api/attach_backend.py
```diff
@@ -162,32 +151,70 @@ def initialize(self, context: ClientAPIBackendContext, fl_ctx: FLContext) -> Non
...
             self._unwind()
             raise
 
-    def _start_attach_listener(self) -> str:
+    def _install_secure_protocol_guard(self) -> None:
+        core_cell = getattr(self._cell, "core_cell", None)
+        add_filter = getattr(core_cell, "add_incoming_filter", None)
+        if not callable(add_filter):
+            raise RuntimeError("secure network Attach requires a pre-decode Cell incoming filter")
+        add_filter(channel=STREAM_CHANNEL, topic=STREAM_DATA_TOPIC, cb=self._secure_protocol_guard)
```
pcnudde: **[P1] Secure Attach's claimed pre-decode guard is bypassable**

This installs the guard only on the outer streaming route. `_secure_protocol_guard()` then returns `None` for every origin except this backend's expected trainer. A foreign CP child targeting the CJ can therefore reach `Adapter.call()` and FOBS decoding before `_validate_session_msg()` rejects the origin. Separately, `Cell._register_request_cb()` registers the same application callback directly on `CoreCell`; a bound trainer can send a direct clear `client_api` request that never crosses this stream-only filter, and `CoreCell._process_request()` decodes it before dispatch.

That leaves bypasses around the documented requirement that Client API and DownloadService messages be origin-authorized and carry both `SECURE` and `ENCRYPTED` before FOBS decoding. Please enforce the policy at a shared pre-decode point that covers direct and streamed routes and rejects unclaimed origins without breaking concurrent Attach sessions. Add adversarial coverage for a foreign CP child and a direct clear CoreCell request.

nvflare/app_common/executors/client_api/attach_backend.py
```diff
@@ -338,11 +349,9 @@ def _session_loop(self) -> None:
...
             reason = liveness_error
-            # RESULT_READY acceptance can outlive execute(): the downstream
-            # consumer may still be pulling trainer-owned lazy sources after
-            # _current_task is cleared. Keep this session authoritative until
-            # the trainer reports release; otherwise finalize() could observe a
-            # replacement session and tear down the live source route.
+            # Keep the session authoritative until the trainer reports that its
+            # result publication to the CJ has settled. Otherwise finalize()
+            # could replace the session while the CJ is still materializing it.
             if self._context.allow_reconnect and session.result_source_live.is_set():
```
pcnudde, line 438: **[P1] `allow_reconnect=True` cannot recover a dead accepted-result source**

`CellBackendBase._handle_result_ready()` unconditionally sets `session.result_source_live` when it accepts a result. Once that happens, this branch ignores every later liveness error while the latch remains set, but the monitor never checks whether `session.trainer_fqcn` is still connected. If the trainer crashes or disconnects after result acceptance, it cannot send the heartbeat that clears the latch, so the stale session remains authoritative forever and later rounds keep targeting it despite reconnect being enabled.

Please retire the source after a bounded, confirmed-disconnect grace period (while preserving transient reconnects), then install a fresh session. Add a test where the trainer disappears after acceptance without sending `RESULT_SOURCE_LIVE=False`.

nvflare/app_common/executors/client_api/attach_backend.py
```diff
@@ -693,6 +702,7 @@ def _request_shutdown(self, session: _AttachedTrainerSession) -> None:
                 ),
                 timeout=_CONTROL_ATTEMPT_TIMEOUT,
                 optional=True,
+                secure=self._protocol_secure,
```
pcnudde, line 814: **[P2] Heartbeat-disabled Attach can strand finalization and the external trainer after a lost `SHUTDOWN`**

`initialize()` accepts `heartbeat_timeout=0` whenever `result_wait_timeout` is finite, and the trainer then starts no heartbeat thread. After an accepted result sets `result_source_live`, finalization sends this `SHUTDOWN` request only once. If it is lost, no heartbeat can clear the source-live latch: the CJ can wait for the full streaming idle timeout plus grace, while the externally owned trainer remains blocked in `receive()` indefinitely.

Please either disallow heartbeat-disabled Attach, retry/acknowledge shutdown through a terminal session-close protocol, or add another connection-loss mechanism that unblocks both sides. Cover a lost `SHUTDOWN` with `heartbeat_timeout=0`.

pcnudde (commented):
Additional findings on exact head `7e855d73cb97d25dea069c0634c90ace7db09392`. The two Swarm/CJ-memory findings are already filed in the existing focused review, so this review adds the three remaining trust-boundary and lifecycle findings.

nvflare/app_common/executors/client_api/attach_backend.py
YuanTingHsieh: Addressed in b106106f2ed1299e4ba5c429df96ba568cb6738c. Secure network Attach now installs one shared _AttachProtocolGuard on the CJ CoreCell with independent claims for concurrent Attach backends. The pre-decode filter covers the streamed envelope plus direct client_api and DownloadService CoreCell routes. It rejects unclaimed CP-child origins and requires both SECURE and ENCRYPTED for the bound trainer before application FOBS decoding. Added adversarial tests for a protected foreign CP child, a bound trainer direct clear request, direct clear DownloadService traffic, the real CoreCell pre-decode path, and concurrent claim release. The exact-head 10/10 Cell integrations pass, and Colossus recorded 462 protected Attach CP chunks with zero unprotected chunks across the definitive 20/20 matrix.

nvflare/app_common/executors/client_api/attach_backend.py
YuanTingHsieh: Addressed in b106106f2ed1299e4ba5c429df96ba568cb6738c. An accepted result source remains authoritative while connected, but each Attach session now tracks continuous disconnection. After liveness failure, the backend requires is_cell_connected(trainer_fqcn) to remain false through a bounded grace capped at five seconds; reconnection resets that grace. A confirmed dead source has its live latch and retained authority retired before a fresh session is installed. Tests cover disappearance after acceptance without RESULT_SOURCE_LIVE=False and a transient disconnect that preserves the original session. The exact-head Colossus lifecycle probes passed both cases, and the full matrix passed 20/20.

nvflare/app_common/executors/client_api/attach_backend.py
YuanTingHsieh: Addressed in b106106f2ed1299e4ba5c429df96ba568cb6738c by taking the fail-closed option suggested here: Attach now requires heartbeat_timeout > 0 even when result_wait_timeout is finite. ClientAPIExecutor rejects the configuration at construction, and AttachBackend repeats the validation defensively before creating a Cell/session, so a heartbeat-disabled trainer can never enter the lost-SHUTDOWN state. heartbeat_timeout=0 remains supported for managed external_process mode. Constructor/backend tests verify rejection before SESSION_OPEN or SHUTDOWN, and the exact-head Colossus fault probe passed this case together with the 20/20 normal matrix.

YuanTingHsieh:
@chesterxgchen Thanks for the focused review. Both points are now addressed at signed head `b106106f2ed1299e4ba5c429df96ba568cb6738c`.

The Attach + Swarm trust-boundary defect is fixed: `ExecutionMode.ATTACH` is excluded from lazy-reference handoff, so the CJ materializes the original remote reference and owns any onward transaction. Focused tests assert that the original source/ref never reaches the attached trainer, and the definitive Swarm Attach cases passed.

The CJ memory behavior is intentional for 2.9. Attach preserves the released 2.8-style CJ trust and materialization boundary, so the site credential is never delegated to the externally owned trainer. The memory difference is relative to the unreleased lazy/relay implementation on `main`, which uses a different security model. We accept this peak-memory tradeoff temporarily; 2.10 will address lazy Attach transfer through scoped trainer identity, DownloadService ACLs, and revocation.

chesterxgchen (approved):

PR #5043, opened 2026-08-05
Restore Docker DinD compatibility

YuanTingHsieh:
Follow-up to #5029.

### Description

Restore compatibility with Docker-in-Docker and other remote Docker daemon setups while retaining the local socket validation and permission handling added in #5029.

The generated `start_docker.sh` previously treated `NVFL_DOCKER_SOCK` as both the Docker CLI endpoint and the bind-mount source. That fails in the Docker E2E Jenkins pod: the CLI runs in the `gpu` container and reaches a DinD sidecar through `DOCKER_HOST`, while `/var/run/docker.sock` exists only on the daemon side.

This change:

- pins outer Docker commands to `unix://$DOCKER_SOCK` for local Unix endpoints, keeping the parent, network, and jobs on one daemon;
- retains the existing `DOCKER_HOST` or Docker context for TCP/SSH/DinD endpoints, where `NVFL_DOCKER_SOCK` is a daemon-host bind source;
- safely probes remote sockets through the configured parent image to validate the file type and discover its numeric GID;
- adds the discovered socket GID to the non-root parent container;
- supports `NVFL_DOCKER_SOCK_GID` as an explicit administrator/CI override when the remote probe must be bypassed;
- uses `--mount` so a missing daemon-host bind source is rejected rather than created as a directory;
- rejects relative, missing, and non-socket local paths; and
- documents the local-versus-remote behavior.

### Types of changes

- [x] Documentation updated.

### Testing

- `python -m pytest tests/unit_test/tool/deploy/deploy_commands_test.py -v` (109 passed)
- `./runtest.sh -s`
- `bash -n nvflare/tool/deploy/templates/docker/start_docker.sh`
- `git diff --check`

The behavioral tests cover both sides of the boundary:

- local socket overrides pin every outer Docker command to the selected Unix daemon;
- TCP/DinD paths are validated on the daemon host and their GID is added to the parent;
- a verified `NVFL_DOCKER_SOCK_GID` bypasses the probe without changing the remote CLI endpoint;
- failed remote probes stop before launching the parent; and
- local Unix endpoints reject missing/non-socket paths.

### DIFF
--- a/docs/design/docker_job_launcher_design.md
+++ b/docs/design/docker_job_launcher_design.md
@@ -175,8 +175,8 @@ docker run ... -e NVFL_DOCKER_WORKSPACE="$HOST_WORKSPACE" ...
 SP/CP container (site admin grants via start_docker.sh)
   ├── /var/run/docker.sock mounted            ← can create job containers
   ├── --user $(id -u):$(id -g)               ← runs as calling user (workspace files not root-owned)
-  ├── --group-add <docker-socket-gid>         ← grants access when the socket is locally visible
-  ├── --group-add 0 when needed               ← handles locally visible sockets that appear root-owned
+  ├── --group-add <docker-socket-gid>         ← grants access using the local stat or daemon-host probe
+  ├── --group-add 0 when needed               ← handles sockets that appear root-owned
   ├── workspace bind mount at /var/tmp/nvflare/workspace
   ├── nvflare-network                         ← intra-site: SP↔SJ / CP↔CJ (PARENT_URL, Docker DNS)
   └── host network (-p fed_learn_port)        ← cross-site: CP→SP over HTTPS, same as process mode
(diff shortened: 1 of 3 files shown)


chesterxgchen:
• Two key findings:

  1. P1 — Local socket override can use two different Docker daemons

     Setting NVFL_DOCKER_SOCK=/custom/docker.sock no longer makes the outer Docker commands use that
     socket. The parent container and network may be created on the default daemon, while jobs use the
     custom daemon mounted inside the parent. Those containers cannot share the required network/
     topology.

  2. P1 — Remote DinD socket may be inaccessible

     When the socket exists only on the remote daemon host, the script adds no supplementary socket
     group. The parent runs as a non-root user, while standard Docker sockets are typically
     root:<docker-gid> with mode 0660. The outer launch succeeds, but job launch inside the parent can
     fail with PermissionError(13).

  The new tests use a fake Docker command, so they do not test either the two-daemon topology or actual
  socket access inside the non-root parent.

pcnudde (approved):

PR #5045, opened 2026-08-05
Warn that recipe export arguments are reserved

chesterxgchen:
## Summary

- warn whenever NVFlare consumes the reserved --export or --export-dir recipe arguments
- explain that these system-level names are unavailable to the script argument parser
- retain the additional diagnostic when --export-dir is supplied without --export
- preserve transactional argument handling when warnings are promoted to errors
- document the import-time reservation in the Recipe API guide

## Behavior

NVFlare cannot distinguish a script-defined collision from its documented recipe export syntax because both use the same command line. The warning is therefore emitted for every consumed recipe export argument, including valid NVFlare exports.

## Tests

- python -m pytest tests/unit_test/recipe/spec_test.py -q (70 passed)
- python -m pytest tests/unit_test/recipe -q (525 passed, 22 skipped)
- ./runtest.sh -s
- ./runtest.sh -l

### DIFF
--- a/docs/user_guide/data_scientist_guide/recipe_api.rst
+++ b/docs/user_guide/data_scientist_guide/recipe_api.rst
@@ -73,6 +73,10 @@ All concrete recipes support the common execution surface:
 
       python job.py --export --export-dir /tmp/nvflare/job
 
+   Recipe scripts reserve ``--export`` and ``--export-dir``. NVFlare consumes
+   these arguments during module import, before the script's argument parser
+   runs. Use different option names for script-specific behavior.
+
 ``recipe.run(env, server_exec_params=None, client_exec_params=None)``
    Submit directly through the environment and return a ``Run`` handle. Most
    user examples should use ``execute`` so export flags continue to work.
--- a/docs/user_guide/data_scientist_guide/recipe_api.rst
+++ b/docs/user_guide/data_scientist_guide/recipe_api.rst
@@ -75,7 +75,8 @@ All concrete recipes support the common execution surface:
 
    Recipe scripts reserve ``--export`` and ``--export-dir``. NVFlare consumes
    these arguments during module import, before the script's argument parser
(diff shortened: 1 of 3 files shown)


pcnudde (approved):

YuanTingHsieh (approved):

YuanTingHsieh:
This PR formalizes `--export` and `--export-dir` as reserved Recipe API arguments, but many Recipe-based examples still define and handle their own `--export_config` or `--export_dir` options (for example hello-numpy, hello-collab, multi-GPU, LLM HF, MLflow/W&B, PSI, and swarm-learning).

Can we first confirm the canonical export interface? If the framework-owned options are retained, these Recipe examples and their documentation should be migrated—either in this PR or in a linked follow-up issue. Otherwise, users will continue seeing multiple competing export conventions.

This should cover Recipe-based examples only; `FedJob`-only scripts may legitimately own their export arguments. I also think we should resolve whether import-time `sys.argv` consumption is the desired long-term design before expanding its use throughout the examples.

chesterxgchen:
> This PR formalizes `--export` and `--export-dir` as reserved Recipe API arguments, but many Recipe-based examples still define and handle their own `--export_config` or `--export_dir` options (for example hello-numpy, hello-collab, multi-GPU, LLM HF, MLflow/W&B, PSI, and swarm-learning).
> 
> Can we first confirm the canonical export interface? If the framework-owned options are retained, these Recipe examples and their documentation should be migrated—either in this PR or in a linked follow-up issue. Otherwise, users will continue seeing multiple competing export conventions.
> 
> This should cover Recipe-based examples only; `FedJob`-only scripts may legitimately own their export arguments. I also think we should resolve whether import-time `sys.argv` consumption is the desired long-term design before expanding its use throughout the examples.

We should have consistency.  Remove the customized --export-config or export-dir etc from another examples will be in another PR

PR #5048, opened 2026-08-06
Use standard Recipe export CLI in examples

chesterxgchen:
## Summary
- remove custom export-mode and export-directory arguments from 16 Recipe examples
- route ordinary Recipe execution and export through `recipe.execute(env)` and the system-level `--export --export-dir` interface
- stop three examples from exporting unconditionally before normal execution
- remove redundant Swarm workspace deletion and verify export short-circuiting preserves existing results
- update affected READMEs, LLM launch commands, and the Recipe API user guide
- add regression tests preventing migrated examples from reintroducing custom export branches or export-only setup side effects

## Compatibility
Normal simulation behavior and environment construction are unchanged. Export remains available through `python job.py --export --export-dir <path>`.

The two Edge production examples intentionally retain their existing `--export_job` path. Their `ProdEnv` validates a deployment startup kit during construction, while job export does not require one; constructing it before `recipe.execute()` would break standalone export. FedJob/Job API examples remain out of scope.

## Validation
- `python -m pytest -q tests/unit_test/recipe tests/unit_test/examples/recipe_export_arguments_test.py` (527 passed, 22 skipped)
- focused Recipe export tests (72 passed)
- Hello NumPy and log-streaming export smoke tests against this branch
- all 16 migrated scripts compile
- 11/16 example `--help` checks passed; 5 require optional example dependencies unavailable in the development environment
- `./runtest.sh -s`
- `./runtest.sh -l`
- `git diff --check`


chesterxgchen:
Addressed both review findings in 88bb81f54. The two Edge production examples now retain their explicit export path because ProdEnv requires a provisioned startup kit before execute() can dispatch export. Swarm no longer recursively deletes its workspace before execute(), and a regression test verifies export short-circuiting preserves existing results. Local validation: 527 passed, 22 skipped; style and license checks passed.

YuanTingHsieh (approved):
good to be consistent, a question this mean we don't support export and run in the same script, can just choose one, i guess that's fine.

PR #5050, opened 2026-08-06
Upgrade CIFAR-10 SplitNN example to the Collab API

ZiyueXu77:
### Description

Upgrades the existing two-party PyTorch SplitNN CIFAR-10 example in place to use the Collab API. This replaces the component-based notebook and static job configuration instead of adding a parallel implementation.

- Keeps the established `examples/advanced/vertical_federated_learning/cifar10-splitnn` entry point.
- Preserves the prepare-data-first workflow: vertically split CIFAR-10, run PSI, then train on aligned sample IDs.
- Organizes responsibilities into `client.py`, `server.py`, `model.py`, and `job.py`, with focused data and PSI helpers.
- Keeps images and the bottom model on `site-1`, and labels and the top model on `site-2`.
- Expresses activation, gradient, metric, and model-state exchange as ordinary Python arguments and return values without application-level `Shareable`, `DXO`, FOBS, task, auxiliary-handler, or JSON component wiring.
- Links the single upgraded implementation from both the vertical-FL and Collab example catalogs.

### Types of changes

- [x] Non-breaking example update.
- [x] Documentation updated.

### Validation

- `./runtest.sh -s` passed: Black, isort, Flake8, and agent-skill lint.
- All relocated Python files compile.
- `job.py --help` passed and the two-site `CollabRecipe` constructed successfully with prepared PSI placeholders.
- The CI relative-link checker passed for every changed README.
- The full 15,625-step baseline and Collab simulator runs completed on the same prepared data and one-GPU environment, as documented below.

<details>
<summary>Before/after runtime and numerical-equivalence validation</summary>

The component-based baseline used the existing example at [`c467e40d`](https://github.com/NVIDIA/NVFlare/commit/c467e40d40e3391482ae3b8db6cb7c933d3c5373). The Collab run used the training logic at [`2a2ff170`](https://github.com/NVIDIA/NVFlare/commit/2a2ff170d770c5c09e5596fd5e7b1c30d8df4ebf), which is relocated to the established example path by this PR.

Both runs reused `/tmp/cifar10` and the exact same site-specific PSI artifacts. They also held model initialization, random batches, 15,625 steps, batch size 64, SGD settings, validation cadence, and float16 cut-layer exchange constant.

#### Environment

| Component | Configuration |
| --- | --- |
| GPU | One NVIDIA RTX 6000 Ada Generation, 49,140 MiB; both sites shared `cuda:0` |
| NVIDIA driver | 580.173.02 |
| CUDA and cuDNN | PyTorch CUDA 12.6; cuDNN 9.5.1 |
| CPU | Intel Core i7-6800K; 6 cores, 12 threads |
| Host memory | 109 GiB |
| OS | Ubuntu 24.04.4 LTS; Linux kernel 6.8.0-136-generic |
| Python | 3.10.0 |
| PyTorch | 2.6.0+cu126 |
| torchvision | 0.21.0+cu126 |
| NVFlare | Source checkout based on `c467e40d`, plus the Collab example changes |

#### Commands

Component-based baseline:

```bash
cd examples/advanced/vertical_federated_learning/cifar10-splitnn
export PYTHONPATH=${PWD}/src:${PWD}/../../cifar10:${PWD}/../../cifar10/pt/src
mkdir -p /tmp/nvflare/cifar10_splitnn/local
printf "%s\n" "{\"class_allow_list\": [\"nvflare.\", \"splitnn.\", \"pt.src.model.ModerateCNN\"]}" > /tmp/nvflare/cifar10_splitnn/local/resources.json

CUDA_VISIBLE_DEVICES=0 python - <<PY
from nvflare import SimulatorRunner

simulator = SimulatorRunner(
    job_folder="jobs/cifar10_splitnn",
    workspace="/tmp/nvflare/cifar10_splitnn",
    n_clients=2,
    threads=2,
    log_config="ERROR",
)
simulator.run()
PY
```

Collab API implementation:

```bash
cd examples/advanced/collab/pt_splitnn
CUDA_VISIBLE_DEVICES=0 python job.py
```

#### Results

| Implementation | Elapsed time |
| --- | ---: |
| Component-based SplitNN | 50m 04.4s |
| Collab API SplitNN | 41m 20.4s |

Elapsed time covers the complete simulator command, including startup, validation, model export, and shutdown. Both runs used ERROR-level framework logging. These are single measurements rather than averages over repeated runs.

The Collab API run was about **17.4% faster**. All 15,625 recorded training-loss and training-accuracy values were identical at matching steps. Across the 16 unsmoothed validation points, the largest loss difference was `4.23e-6`, and the largest accuracy difference was one prediction among 10,000 validation images.

Component-based and Collab API SplitNN curve alignment

The training panels use a 200-step moving average for readability; validation panels show every recorded value without smoothing.

</details>


chesterxgchen (commented):
code file names and readme structure are not following standard layout and conventions

chesterxgchen (commented):
general comments, many are identical to comments to https://github.com/NVIDIA/NVFlare/pull/5050

don't treat as pure research example, treat this as education example

examples/advanced/collab/pt_splitnn/README.md
```diff
@@ -0,0 +1,384 @@
...
+This example consumes the data and private-set-intersection artifacts produced
+by the existing
+[CIFAR-10 SplitNN example](../../vertical_federated_learning/cifar10-splitnn/README.md).
+It does not define a separate Collab-specific preparation format.
+
+From the repository root, run the existing vertical splitter:
+
+```bash
+cd examples/advanced/vertical_federated_learning/cifar10-splitnn
+export PYTHONPATH=${PWD}/src:${PWD}/../../cifar10:${PWD}/../../cifar10/pt/src
```
chesterxgchen: This line is very strange to put the python code on some other code python. Very hard to understand. 
If you are going to run this, why not make a copy. i.e. give clear instructions to say you are going to copy the code from another CIFAR10 example. 

so you have the complete the example here before you start.

This line seems to save code, but adding confusion to user

examples/advanced/collab/pt_splitnn/README.md
```diff
@@ -0,0 +1,384 @@
...
+python cifar10_split_data_vertical.py \
+    --split_dir /tmp/cifar10_vert_splits \
+    --overlap 10000
+```
+
+This downloads CIFAR-10 under `/tmp/cifar10` and creates the same per-site
+vertical splits used by the existing example. From the same directory, authorize
+its custom PSI adapter in the local simulator workspace and run the existing PSI
```
chesterxgchen: did not explain what is PSI and why it is needed

chesterxgchen (commented):
add few comments

examples/advanced/collab/pt_splitnn/README.md
ZiyueXu77: This is mainly for apple-to-apple comparison with our existing split-NN example, and same for the PSI part. But this particular line is for "Set PYTHONPATH to include custom files of this example and some reused files from the [CIFAR-10](https://github.com/NVIDIA/NVFlare/blob/main/examples/advanced/cifar10) examples" - which is also needed in the existing splitNN example. So it would need to be there anyways. 
I will add more explanation

examples/advanced/collab/pt_splitnn/README.md
ZiyueXu77: added description

examples/advanced/collab/pt_splitnn/README.md
chesterxgchen: for new users, they will not refers to old example, by keep referring what we used to do, is adding historical burden for user to understand. 

They proabably don't care what was done before, just show me the way I need to do

examples/advanced/vertical_federated_learning/cifar10-splitnn/src/splitnn/cifar10_learner_splitnn.py
```diff
@@ -307,7 +307,7 @@ def _val_step_label_side(self, batch_indices, activations, fl_ctx: FLContext):
             pred = self.model.forward(activations)
 
             loss = self.criterion(pred, labels)
-            self.val_loss.append(loss.unsqueeze(0))  # unsqueeze needed for later concatenation
+            self.val_loss.append((loss * len(labels)).unsqueeze(0))
```
chesterxgchen: do we still keep the learner pattern ?

examples/advanced/collab/pt_splitnn/README.md
ZiyueXu77: The original intention is to showcase the easy to use and reproducible features of Collab, and I feel both benefit from a reference to compare to - so I use the exsiting example as that reference to show the new API is easier to work with, and they produce the identical results. 
But I also agree for new users, they do not need to know what the old way is. However since we indeed have two examples both present in the repo for the exact same task, I feel we still need to explain the reason to avoid confusion.

examples/advanced/vertical_federated_learning/cifar10-splitnn/src/splitnn/cifar10_learner_splitnn.py
ZiyueXu77: we may choose to update our previous example, this is just an incidental finding by Greptile over our old example, we can choose to update it later in a sepatate PR.

examples/advanced/collab/pt_splitnn/README.md
chesterxgchen: for new user, explain who old way to do thing is not needed. You can change the old example to explain there is new approach.  But the otherway around

examples/advanced/vertical_federated_learning/cifar10-splitnn/src/splitnn/cifar10_learner_splitnn.py
chesterxgchen: ok

examples/advanced/collab/pt_splitnn/README.md
ZiyueXu77: agreed, yeah that makes sense, will update this

ZiyueXu77: updated

examples/advanced/vertical_federated_learning/cifar10-splitnn/src/splitnn/cifar10_learner_splitnn.py
```diff
@@ -318,10 +318,9 @@ def _val_step_label_side(self, batch_indices, activations, fl_ctx: FLContext):
 
     def _log_validation(self, fl_ctx: FLContext):
         if len(self.val_loss) > 0:
-            loss = torch.mean(torch.cat(self.val_loss))
-
             _val_pred_labels = torch.cat(self.val_pred_labels)
             _val_labels = torch.cat(self.val_labels)
+            loss = torch.sum(torch.cat(self.val_loss)) / len(_val_labels)
```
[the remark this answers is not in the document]
ZiyueXu77: torch.cat(self.val_labels) is one-dimensional because batches are added using extend(labels.unsqueeze(0)). Therefore:

  len(_val_labels) == _val_labels.numel()

  The denominator is already the number of validation examples, not batches. A shape test confirmed this, and the aligned validation curves provide additional evidence.

  No code changes are needed.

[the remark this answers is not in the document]
ZiyueXu77: Please update your Greptile Summary

examples/advanced/vertical_federated_learning/cifar10-splitnn/data.py
```diff
@@ -0,0 +1,73 @@
...
+    """Expose either images or labels for the same prepared sample indices."""
+
+    def __init__(self, dataset_root: str, intersection_file: str, role: str, train: bool):
+        if role not in ("image", "label"):
+            raise ValueError(f"role must be 'image' or 'label', got {role!r}")
+
+        dataset = datasets.CIFAR10(root=dataset_root, train=train, download=False)
+        indices = np.arange(len(dataset))
+        if train:
+            indices = np.loadtxt(intersection_file)
```
holgerroth: `np.loadtxt` returns a 0-d array for a one-line intersection file, so the `np.sort` on the next line raises `AxisError: axis -1 is out of bounds for array of dimension 0` during `@collab.init` (reproducible: `prepare_data.py --overlap 1` is accepted, and `FilePSIWriter` writes one ID per line). An empty intersection yields `train_size=0` and later crashes in `rng.randint(0, 0)`. The old learner guarded this with an explicit `Expected train dataset size to be larger zero` error.

Suggest `np.loadtxt(intersection_file, ndmin=1)` plus a `size > 0` check with a clear message.

examples/advanced/vertical_federated_learning/cifar10-splitnn/client.py
```diff
@@ -0,0 +1,300 @@
...
+            if collab.is_aborted:
+                return {
+                    "steps_completed": steps_completed,
+                    "model_path": None,
+                    "last_metrics": last_metrics,
+                }
+
+            models = {
+                "bottom_model": self.get_model(),
+                "top_model": label_client(timeout=call_timeout).get_model(),
```
holgerroth: This transfers the trained label-side `TopModel` state dict to site-1 and persists it there, which directly contradicts `model.py`'s `TopModel` docstring ("labels and classifier parameters remain local"). In the old component-based example, trained label-side weights never left site-2.

Since this is the reference vertical-FL example, users will copy the pattern into real two-party deployments where the label owner's trained classifier (which encodes label information) shouldn't be exported to the feature owner. Either persist each half at its owning site, or — if collecting both halves at site-1 is an intentional simplification — update the `TopModel` docstring and add a note in the README's privacy paragraph stating that final model collection crosses the site boundary.

examples/advanced/vertical_federated_learning/cifar10-splitnn/server.py
```diff
@@ -0,0 +1,47 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+"""Server-side SplitNN workflow expressed as direct Collab API calls."""
+
+
+from nvflare.collab import collab
+from nvflare.fuel.utils.log_utils import get_obj_logger
+
+RUN_TIMEOUT = 7_200.0
```
holgerroth: `RUN_TIMEOUT = 7_200.0` caps the entire 15,625-step run in one collab call. The old job config used `task_timeout: 0` (unlimited). The README says PyTorch falls back to CPU automatically, but the GPU run already took ~41 min — a CPU run will very likely exceed 2 h, at which point the dispatcher raises `CollabCallError(TimeoutError)` out of `@collab.main` and the run ends with no result while the clients are still training, with no hint to raise `RUN_TIMEOUT`.

Suggest a much larger default (or documenting that CPU users must raise this constant).

examples/advanced/vertical_federated_learning/cifar10-splitnn/client.py
```diff
@@ -0,0 +1,300 @@
...
+
+    def _require_role(self, expected: str) -> None:
+        if self.role != expected:
+            raise RuntimeError(f"[{collab.site_name}] operation requires role {expected!r}, got {self.role!r}")
+
+    def _get_label_client(self):
+        # collab.clients contains callable proxies for the participating sites.
+        # Calling one invokes a published method on that remote client.
+        clients = {client.name: client for client in collab.clients}
+        label_client = clients.get("site-2")
```
holgerroth: The role-to-site mapping is now hardcoded in three files: here (`"site-2"`), `server.py` (`"site-1"`/`"site-2"`, with site-1 assumed to be the image side), and `job.py`'s `set_per_site_config` — which the README presents as the configuration point. Swapping roles or renaming sites in job.py deploys fine and then fails at runtime with `operation requires role 'image', got 'label'` from a different file than the one edited.

Suggest deriving the peer from the configured role (e.g. a `label_site` app prop set by job.py) and using the existing `collab.get_clients([...])` facade helper instead of the hand-rolled `{client.name: client}` lookups here and in `server.py` — it already raises a clear error naming any missing site.

examples/advanced/vertical_federated_learning/cifar10-splitnn/job.py
```diff
@@ -0,0 +1,113 @@
...
+        default=DEFAULT_PSI_WORKSPACE,
+        help="Workspace produced by prepare_data.py",
+    )
+    parser.add_argument("--workspace-root", default="/tmp/nvflare")
+    return parser
+
+
+def _intersection_file(psi_workspace: str, site_name: str) -> str:
+    path = (
+        Path(psi_workspace).expanduser().resolve() / site_name / "simulate_job" / site_name / "psi" / "intersection.txt"
```
holgerroth: This reconstructs the PSI output location by hardcoding the simulator-internal `simulate_job` layout plus `DhPSIRecipe`'s default `output_path` (`psi/intersection.txt`), which `prepare_data.py` never states explicitly; the README repeats the same path. If either default changes, or the PSI step runs outside SimEnv, this raises a misleading "Run prepare_data.py first" `FileNotFoundError` even though PSI succeeded.

Suggest passing `output_path` explicitly in `prepare_data.py`'s `DhPSIRecipe` and sharing one constant — or having `prepare_data.py` copy the intersections to a stable, documented location next to the split dir.

examples/advanced/vertical_federated_learning/cifar10-splitnn/client.py
```diff
@@ -0,0 +1,300 @@
...
+        """Run the image-side forward pass and retain its autograd graph."""
+        self._require_role("image")
+        if collab.is_aborted:
+            return None
+
+        self.model.train()
+        inputs = self.train_dataset.get_batch(batch_indices).to(self.device)
+        self._train_activations = self.model(inputs)
+        cut_activations = self._train_activations.detach().flatten(start_dim=1)
+        cut_activations = cut_activations.to(dtype=torch.float16)
```
holgerroth: The float16 wire encoding is hand-copied at three send sites (lines 232, 257, 279: `.to(dtype=torch.float16)` + `.cpu()`) and three receive sites (lines 247, 268, 291: `.to(device, float32)`). Since the README advertises the float16 cut-layer exchange as a feature, changing the precision means consistently editing six coordinated sites across two roles — missing one silently changes numerics with no error.

Two tiny helpers (e.g. `_to_wire(t)` / `_from_wire(t, device)`) would give the protocol a single owner.

examples/advanced/vertical_federated_learning/cifar10-splitnn/README.md
```diff
@@ -1,34 +1,276 @@
...
+
+```bash
+cd examples/advanced/vertical_federated_learning/cifar10-splitnn
+python prepare_data.py
+```
+
+`prepare_data.py` downloads CIFAR-10 to `/tmp/cifar10`, writes the per-site
+sample IDs to `/tmp/cifar10_vert_splits`, and uses NVFlare's `DhPSIRecipe` to
+produce the aligned intersection files. The generated `overlap.npy` records the
+ground-truth overlap for checking the PSI result.
```
holgerroth: This says `overlap.npy` records the ground-truth overlap "for checking the PSI result", but the deleted notebook was the only thing that performed that check — nothing in the remaining workflow reads `overlap.npy`. If PSI ever produced a wrong intersection, misaligned image/label pairs would only surface after a full 15,625-step run.

Suggest either adding a few-line check at the end of `prepare_data.py` (compare `intersection.txt` against `overlap.npy`) or dropping the claim.

holgerroth (commented):
Nice upgrade overall — the Collab version reads much more clearly than the component-based implementation, and the numerical-equivalence validation is appreciated. A few issues found during review, added inline. The two I'd consider blocking are the intersection-file loading edge case (dropped size guard) and the top-model transfer contradicting the code's own privacy docstring.

examples/advanced/vertical_federated_learning/cifar10-splitnn/client.py
```diff
@@ -0,0 +1,300 @@
...
+        writer.add_scalar("val_accuracy", result["accuracy"], step)
+        return result
+
+    # @collab.publish exposes this method to remote Collab callers. The
+    # server invokes it on site-1, which then owns the complete training loop.
+    @collab.publish
+    def run_splitnn(self):
+        train_size = len(self.train_dataset)
+        test_size = len(self.valid_dataset)
+        num_steps = NUM_STEPS
```
holgerroth: Minor: `run_splitnn()` opens with seven pure-alias locals (`num_steps = NUM_STEPS`, `batch_size = BATCH_SIZE`, …) that are never varied, while the rest of the file uses the module constants directly. The `validation_frequency > 0` guard on a constant 1000 is likewise dead. Using the constants directly (as `initialize()` does) would keep the focus on the Collab call flow this example exists to showcase.

holgerroth (approved):
All review comments were addressed — thanks for the thorough follow-up (site-local model persistence, PSI verification against overlap.npy, stable intersection paths, and the wire/lookup cleanups).

Beyond the fixes, this is a clear win for users: the example drops from ~1,525 lines across 18 files (Python + JSON configs + notebook) to ~822 lines of plain Python in 7 files, with the algorithm-and-wiring layer shrinking ~70% since no dedicated controller/executor, task/aux-topic mapping, or per-step DXO/FOBS conversion is needed. The activation/gradient exchange as a direct call with an ordinary return value makes split learning far easier to read, adapt, and debug. LGTM.

PR #5063, opened 2026-08-07
Reject unsafe FedJob names

nvidianz:
## Summary

- validate job names at the shared `FedJob` entry point so every recipe rejects empty and path-bearing names
- reject POSIX and Windows separators, traversal components, absolute paths, and drive-qualified paths
- revalidate inside `FedJobConfig.generate_job_config()` so direct or mutated low-level configurations cannot collapse onto the caller's export directory
- add regression coverage proving existing user files are preserved even when the export root contains `meta.json` and would otherwise look safe to delete

## Root cause

`FedJob` previously performed only a type check on `name`. `FedJobConfig.generate_job_config()` then passed the value directly to `os.path.join(job_root, job_name)`. An empty or `.` name resolved to `job_root` itself, while traversal and absolute names escaped it. If the resolved directory contained `meta.json`, export treated it as a replaceable job folder and silently removed it with `shutil.rmtree(..., ignore_errors=True)`.

Validating at `FedJob` covers direct users and all recipe families through their shared construction path. Revalidation at the filesystem boundary provides defense in depth.

NVBug: [6567126](https://nvbugspro.nvidia.com/bug/6567126)

## Validation

- `171 passed`: validation utilities, `FedJob`, `FedJobConfig`, and Collab recipe tests
- destructive export-root regression: `7 passed`
- representative `NumpyFedAvgRecipe` construction rejects empty, traversal, absolute, and Windows path-bearing names
- Black, isort, flake8, and agent-skill lint passed on all changed files
- broader `tests/unit_test/job_config` run: `132 passed`, with 3 unrelated ScriptRunner failures because PyTorch is not installed in the focused test environment


nvflare/fuel/utils/validation_utils.py
```diff
@@ -90,6 +91,14 @@ def check_non_empty_str(name, value):
         raise ValueError(f"{name} must not be empty")
 
 
+def check_job_name(name, value):
+    """Validate a job name that will be used as one directory component."""
+    check_non_empty_str(name, value)
+    drive, _ = ntpath.splitdrive(value)
+    if value in (".", "..") or "/" in value or "\\" in value or drive:
```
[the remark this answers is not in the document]
nvidianz: Fixed in ebcb22ab4. `check_job_name()` now rejects trailing spaces/dots before filesystem access, preventing Win32 aliases such as `job.` and `job ` from targeting `job`. It also rejects Windows reserved device basenames, including extension forms. Coverage was added at the validator, `FedJob`, and `generate_job_config()` boundary levels; 192 focused tests and scoped Black/isort/flake8/agent-skill lint pass.

YuanTingHsieh (commented):
Two Windows filename-normalization gaps remain in `check_job_name()`.

nvflare/fuel/utils/validation_utils.py
```diff
@@ -90,6 +100,18 @@ def check_non_empty_str(name, value):
...
+        raise ValueError(f"{name} must be a single directory name without path components")
+    if value.endswith((" ", ".")):
```
YuanTingHsieh: **[P1] Reject leading ASCII spaces too**

`check_non_empty_str()` only checks the stripped value, while this condition rejects spaces only at the end. The Windows Object Manager removes ASCII spaces from both the beginning and end of a path component, so a name such as ` job` can resolve to an existing `job` directory. If that directory contains `meta.json`, `generate_job_config()` can treat it as replaceable and recursively delete it. Please reject leading ASCII spaces and add the corresponding preservation regression.

Reference: https://learn.microsoft.com/en-us/troubleshoot/windows-client/shell-experience/file-folder-name-whitespace-characters

nvflare/fuel/utils/validation_utils.py
```diff
@@ -90,6 +100,18 @@ def check_non_empty_str(name, value):
...
+        raise ValueError(f"{name} must not end with a space or dot")
+    if value.split(".", maxsplit=1)[0].upper() in _WINDOWS_RESERVED_FILE_NAMES:
```
YuanTingHsieh: **[P2] Include the superscript Windows device aliases**

Windows also reserves `COM¹`, `COM²`, `COM³`, `LPT¹`, `LPT²`, and `LPT³`, including extension forms such as `COM¹.txt`. These currently pass this check and then fail during export on Windows. Please include them in the reserved-name set and tests.

Reference: https://learn.microsoft.com/en-us/windows/win32/fileio/naming-a-file

nvflare/fuel/utils/validation_utils.py
```diff
@@ -12,11 +12,21 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import ntpath
 import random
 
 SYMBOL_ALL = "@all"
 SYMBOL_NONE = "@none"
 
+_WINDOWS_RESERVED_FILE_NAMES = {
```
pcnudde: Why Windows checks. We don't support it?

nvflare/fuel/utils/validation_utils.py
YuanTingHsieh: windows not supported.

nvflare/fuel/utils/validation_utils.py
YuanTingHsieh: windows not supported.

nvflare/fuel/utils/validation_utils.py
nvidianz: Removed in 0e7d8e173. Windows is not supported, so the Windows drive, reserved-device-name, trailing-dot/space, and backslash checks—and their tests—have been removed. The validation now retains only the cross-platform safeguards against empty names, `.`/`..`, and POSIX path separators.

nvflare/fuel/utils/validation_utils.py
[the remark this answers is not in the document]
nvidianz: Addressed in 60e40db1e. Replacement is now guarded by the exported metadata identity: `generate_job_config()` calls `rmtree()` only when `meta.json` records exactly the requested `job_name`. An aliased existing folder therefore has a different recorded name and raises without deletion. This is filesystem-agnostic and does not reintroduce Windows-specific validation. The regression test preserves an existing folder whose recorded name differs.

nvidianz:
@greptileai review

nvidianz:
@greptileai review

YuanTingHsieh (commented):
One ownership-marker invariant remains to make explicit.

nvflare/job_config/fed_job_config.py
```diff
@@ -149,12 +149,20 @@ def generate_job_config(self, job_root):
...
+            if self._is_valid_job_folder(job_dir, self.job_name):
                 shutil.rmtree(job_dir, ignore_errors=True)
             else:
-                raise RuntimeError(f"Job folder {job_dir} already exists and is not a valid job folder.")
+                raise RuntimeError(f"Job folder {job_dir} already exists and does not belong to job {self.job_name}.")
+
+        # Write the ownership marker before generating app files. If a later
+        # export step fails, a retry can safely identify and replace this folder.
+        os.makedirs(job_dir, exist_ok=True)
+        self._generate_meta(job_dir)
```
YuanTingHsieh: **[P2] Keep the canonical job name non-overridable**

`_generate_meta()` initializes `meta_json["name"]` from `self.job_name`, but then applies `meta_props` afterward. Consequently, `FedJobConfig("job", 1, meta_props={"name": "custom-name"})` writes `custom-name` as this ownership marker; its first export succeeds, while an identical second export fails because `_is_valid_job_folder()` compares `custom-name` with `job` and reports that the folder does not belong to the job.

Since `name` is now the canonical folder identity, please make it non-overridable—preferably by rejecting the reserved `name` key in `meta_props` with a clear `ValueError`—and add an override-rejection/repeated-export regression test. This keeps the fix scoped without requiring a broader read-only-property API change.

pcnudde: **[P2] Keep failed ownership-marker writes retryable**

`_generate_meta()` opens/truncates `meta.json` before serialization and writing complete. If serialization or the write fails—for example because `meta_props` contains a non-serializable value, the disk fills, or the process is interrupted—it can leave an empty or truncated marker. The next `generate_job_config()` call then has `_is_valid_job_folder()` classify the folder as foreign because the marker cannot be parsed, so even a corrected retry fails with `RuntimeError`.

This retry failure is introduced by this PR: on the exact head, I reproduced an initial `TypeError`, a zero-byte `meta.json`, and then a `RuntimeError` on the corrected retry; the same corrected retry succeeds on the base commit. Please make marker creation failure-safe (for example, serialize before creating the folder and use an atomic write with cleanup on failure, or use a separate ownership marker) and add a failed-write-then-retry regression test.

nvidianz: Addressed in 3435a1b63. `meta_props` now rejects the reserved `name` key, keeping the export ownership marker canonical. Added regression coverage for the rejection and for repeated exports with other metadata properties.

nvidianz: Addressed in 3435a1b63. Metadata is serialized before the marker is opened, written to a temporary file, and atomically replaced. If marker creation fails, the newly created export directory is removed so a corrected retry succeeds. Added a serialization-failure/retry regression test.

nvidianz:
@greptileai review

nvflare/job_config/fed_job_config.py
```diff
@@ -139,22 +145,44 @@ def _generate_meta(self, job_dir):
...
                 shutil.rmtree(job_dir, ignore_errors=True)
             else:
-                raise RuntimeError(f"Job folder {job_dir} already exists and is not a valid job folder.")
+                raise RuntimeError(f"Job folder {job_dir} already exists and does not belong to job {self.job_name}.")
+
+        # Write the ownership marker before generating app files. If a later
+        # export step fails, a retry can safely identify and replace this folder.
+        os.makedirs(job_dir, exist_ok=True)
+        try:
+            self._generate_meta(job_dir)
```
YuanTingHsieh: **[P2] Validate replacement metadata before deleting the previous export**

The post-construction `meta_props` check runs here, after lines 172–174 have already removed an existing owned job folder. I reproduced this by exporting once, setting `job_config.meta_props = {"name": "other-job"}`, and exporting again: the intended `ValueError` is raised, but the previous valid export (including a marker file) has already been deleted.

Please perform the metadata validation—ideally preparation/serialization of the complete ownership marker—before `shutil.rmtree()`, and add a regression test that starts with an existing valid export and verifies it remains intact when replacement metadata is rejected.

YuanTingHsieh (commented):
One destructive-ordering edge case remains in re-export validation.

nvflare/job_config/fed_job_config.py
```diff
@@ -139,22 +145,44 @@ def _generate_meta(self, job_dir):
...
+        # constructed directly or job_name was changed after construction.
+        check_job_name("job_name", self.job_name)
         job_dir = os.path.join(job_root, self.job_name)
         if os.path.exists(job_dir):
-            if self._is_valid_job_folder(job_dir) or self._is_partial_export_folder(job_dir):
+            if self._is_valid_job_folder(job_dir, self.job_name):
                 shutil.rmtree(job_dir, ignore_errors=True)
             else:
-                raise RuntimeError(f"Job folder {job_dir} already exists and is not a valid job folder.")
+                raise RuntimeError(f"Job folder {job_dir} already exists and does not belong to job {self.job_name}.")
```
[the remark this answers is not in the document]
nvidianz: Addressed in e3df1d990. Exports now build in a temporary sibling directory and publish with `os.replace()` only after the metadata marker and app files are complete. An interrupted export leaves no target job directory, so the next attempt can proceed. Added an interruption/retry regression test.

nvflare/job_config/fed_job_config.py
nvidianz: Addressed in e3df1d990. `generate_job_config()` now validates and serializes the complete metadata before it checks or removes an existing export directory. Added a regression test confirming an existing owned export remains intact when post-construction metadata attempts to override the reserved name.

nvidianz:
@greptileai review

nvidianz:
@greptileai review

nvflare/job_config/fed_job_config.py
```diff
@@ -139,38 +141,64 @@ def _generate_meta(self, job_dir):
...
+        # Revalidate at the filesystem boundary in case this low-level object was
+        # constructed directly or job_name was changed after construction.
+        check_job_name("job_name", self.job_name)
+        json_dump = self._prepare_meta()
+        os.makedirs(job_root, exist_ok=True)
         job_dir = os.path.join(job_root, self.job_name)
         if os.path.exists(job_dir):
-            if self._is_valid_job_folder(job_dir) or self._is_partial_export_folder(job_dir):
+            if self._is_valid_job_folder(job_dir, self.job_name):
                 shutil.rmtree(job_dir, ignore_errors=True)
```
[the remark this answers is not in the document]
nvidianz: Addressed in c8a8eb6b7. The existing owned export remains in place while the replacement is staged. Only after successful generation is the old directory moved to a unique backup and the staged directory published; if publication fails, the backup is restored. Added regression coverage for generation failure during replacement.

nvidianz:
@greptileai review

nvidianz:
@greptileai review

nvidianz:
@greptileai review

nvidianz:
@greptileai review

nvidianz:
@greptileai review

nvidianz:
@greptileai review

nvflare/job_config/fed_job_config.py
```diff
@@ -139,38 +141,75 @@ def _generate_meta(self, job_dir):
...
+                    self._get_server_app(config_dir, custom_dir, fed_app)
+
+                if fed_app.client_app:
+                    self._get_client_app(config_dir, custom_dir, fed_app)
+
+            if replace_existing:
+                backup_job_dir = f"{temp_job_dir}.previous"
+                os.replace(job_dir, backup_job_dir)
+                try:
+                    os.replace(temp_job_dir, job_dir)
```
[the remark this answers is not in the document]
nvidianz: Addressed in 2ed29a7de through 6c0e92992. The `<job_name>.previous` sibling path is gone. A hard interruption between the two renames now leaves the prior export preserved under `.nvflare_job_backups/`, the retry succeeds because nothing blocks the canonical path, and subsequent exports log a warning naming the backup root so the preserved content is discoverable for manual recovery.

nvflare/job_config/fed_job_config.py
```diff
@@ -139,38 +141,89 @@ def _generate_meta(self, job_dir):
...
         Returns:
 
         """
+        # Revalidate at the filesystem boundary in case this low-level object was
+        # constructed directly or job_name was changed after construction.
+        check_job_name("job_name", self.job_name)
+        os.makedirs(job_root, exist_ok=True)
         job_dir = os.path.join(job_root, self.job_name)
+        backup_job_dir = f"{job_dir}.previous"
+        self._recover_previous_export(job_dir, backup_job_dir)
```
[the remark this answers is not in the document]
nvidianz: Addressed in c37e22134. Backups no longer use job-name-derived sibling paths; they are created under the reserved hidden directory `.nvflare_job_backups/` with unique names, and a job_name equal to that reserved name is rejected. A regression test exports a job while a valid `<job_name>.previous` job exists and confirms it is untouched.

nvflare/job_config/fed_job_config.py
```diff
@@ -139,38 +142,101 @@ def _generate_meta(self, job_dir):
...
+            return
 
-            if fed_app.client_app:
-                self._get_client_app(config_dir, custom_dir, fed_app)
+        if not self._is_valid_job_folder(backup_job_dir, self.job_name):
+            raise RuntimeError(f"Backup job folder {backup_job_dir} does not belong to job {self.job_name}.")
 
-        self._generate_meta(job_dir)
+        if os.path.exists(job_dir):
+            self._remove_backup_export(backup_job_dir)
```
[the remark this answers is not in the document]
nvidianz: Addressed in 49b71a499. Automatic recovery was removed entirely. The exporter never deletes or moves a backup it did not create in the same invocation, so a pre-existing backup is preserved regardless of the canonical directory's state. Regression: an unowned canonical folder fails the ownership check while the backup remains intact.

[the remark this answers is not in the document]
nvidianz: Addressed in 49b71a499. Nothing under `.nvflare_job_backups/` is interpreted, relocated, or deleted based on its contents. The only directory the exporter removes is the backup it created earlier in the same call, addressed by the in-memory path. Caller content in the namespace is preserved (regression test included).

nvflare/job_config/fed_job_config.py
```diff
@@ -139,38 +145,136 @@ def _generate_meta(self, job_dir):
...
+            os.replace(backup_job_dir, job_dir)
+            self._remove_backup_marker(backup_job_dir)
+
+    def _is_backup_export(self, backup_job_dir):
+        marker_file = self._backup_marker_file(backup_job_dir)
+        try:
+            with open(marker_file) as f:
+                return f.read() == self.job_name and self._is_valid_job_folder(backup_job_dir, self.job_name)
+        except OSError:
+            return False
```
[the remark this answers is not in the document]
nvidianz: Addressed in 49b71a499. The provenance marker and `_is_backup_export()` were removed together with automatic recovery, so no on-disk marker—predictable or otherwise—can cause deletion or relocation of caller-owned directories.

nvflare/job_config/fed_job_config.py
```diff
@@ -139,38 +145,136 @@ def _generate_meta(self, job_dir):
...
-        self._generate_meta(job_dir)
+    def _remove_backup_marker(self, backup_job_dir):
+        try:
+            os.remove(self._backup_marker_file(backup_job_dir))
+        except FileNotFoundError:
+            pass
+
+    def _remove_backup_export(self, backup_job_dir):
+        shutil.rmtree(backup_job_dir, ignore_errors=True)
+        self._remove_backup_marker(backup_job_dir)
```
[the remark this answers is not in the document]
nvidianz: Addressed in 49b71a499. Later exports no longer read or validate backup contents, so a half-deleted backup cannot block a retry. Cleanup first moves the backup out of the backup root and then deletes best-effort; any leftover is only reported by the stranded-backup warning (6c0e92992), never treated as an error.

nvflare/job_config/fed_job_config.py
```diff
@@ -139,38 +143,126 @@ def _generate_meta(self, job_dir):
...
+        else:
+            os.replace(backup_job_dir, job_dir)
+            self._remove_backup_root_if_empty(backup_root)
 
-            if fed_app.server_app:
-                self._get_server_app(config_dir, custom_dir, fed_app)
+    def _is_backup_export(self, backup_job_dir):
+        try:
+            uuid.UUID(os.path.basename(backup_job_dir))
+            return self._is_valid_job_folder(backup_job_dir, self.job_name)
```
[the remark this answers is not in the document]
nvidianz: Addressed in 49b71a499. Recovery was removed, and with it any trust derived from on-disk state under the caller-controlled export root. The exporter only deletes or moves the backup directory it created moments earlier in the same invocation, using the path held in memory; pre-existing directories in the backup namespace are never touched, however they are named or populated. Stranded backups are surfaced by a warning (6c0e92992) instead of being auto-recovered.

nvidianz:
@greptileai review

nvflare/job_config/fed_job_config.py
```diff
@@ -139,38 +143,111 @@ def _generate_meta(self, job_dir):
...
 
         Returns:
 
         """
+        # Revalidate at the filesystem boundary in case this low-level object was
+        # constructed directly or job_name was changed after construction.
+        check_job_name("job_name", self.job_name)
+        if self.job_name == BACKUP_ROOT:
+            raise ValueError(f"job_name must not use the reserved name {BACKUP_ROOT}")
+        os.makedirs(job_root, exist_ok=True)
```
nvkevlu, line 176: A relative job_root can redirect deletion after validation.

`job_dir` and `backup_root` remain relative throughout the potentially long staging operation. On Python 3.12+, mkdtemp() returns an absolute path. If the process CWD changes before the swap, the staged job from directory A can replace—and delete—an unvalidated exports/job in directory B. I reproduced exactly that. Normalize job_root to an absolute path once at entry, preferably also using directory-relative operations.

nvidianz: Addressed in 48c4b3b40. `job_root` is normalized with `os.path.abspath()` once at entry, so `job_dir`, `backup_root`, and the staged/cleanup directories all resolve against the tree that was validated, regardless of later CWD changes. Added a regression test that changes the CWD mid-generation with a relative `job_root` and verifies the export lands in the entry directory while the other tree stays untouched (it fails on the previous head). Kept `abspath` rather than dir_fd-relative operations since `os.replace`/`shutil` dir_fd support is not uniformly available.

nvflare/job_config/fed_job_config.py
```diff
@@ -139,38 +143,111 @@ def _generate_meta(self, job_dir):
...
+            if replace_existing:
+                os.makedirs(backup_root, exist_ok=True)
+                backup_job_dir = os.path.join(backup_root, uuid.uuid4().hex)
+                os.replace(job_dir, backup_job_dir)
+                try:
+                    os.replace(temp_job_dir, job_dir)
+                except BaseException:
+                    os.replace(backup_job_dir, job_dir)
+                    raise
+                self._remove_backup_export(job_root, backup_job_dir)
```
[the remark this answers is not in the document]
nvidianz: Addressed in 8af6b6de1. Backup disposal after the staged directory is published is now best-effort: an OSError logs a warning naming the leftover backup instead of propagating, so callers see the committed export as successful. The leftover is also surfaced by the stranded-backup warning on later exports. Added a regression test where cleanup fails and the export still succeeds with the backup preserved.

nvidianz:
@greptileai review

nvflare/fuel/utils/validation_utils.py
```diff
@@ -90,6 +90,17 @@ def check_non_empty_str(name, value):
...
 
 
+def check_job_name(name, value):
+    """Validate a job name that will be used as one POSIX directory component.
+
+    NVFlare does not support Windows. Names are therefore evaluated with POSIX
+    path semantics and passed to the filesystem unchanged.
+    """
+    check_non_empty_str(name, value)
+    if value in (".", "..") or "/" in value:
```
nvkevlu: **[P1] Align the accepted name set with every downstream consumer**

Backslash is an ordinary character under this POSIX-only check, but NVFlare later reinterprets it as a separator. `normpath_for_zip()` replaces `\` with `/` before checking the source path, so `r"nested\job"` cannot be archived and `r"..\evil"` becomes `../evil`; the latter can cause the ZIP path to read a sibling directory outside the export root.

Other names accepted here are also unusable: simulator command construction interprets spaces, tabs, newlines, quotes, `#`, and backslashes through `shlex.split()`, while production submission only accepts `[A-Za-z0-9][A-Za-z0-9._-]*`. Embedded NUL also passes this check and fails later with `ValueError: embedded null byte`.

Please define one canonical end-to-end job-name grammar and use it at every boundary. Given the existing submission contract, the simplest option is a real full match such as `re.fullmatch(r"[A-Za-z0-9][A-Za-z0-9._-]*", value)`. Do not reuse the current `$`-based `name_check()` unchanged because it mistakenly accepts a final newline.

nvidianz: Addressed in aea000d27. `check_job_name()` now enforces the canonical grammar with `re.fullmatch(r"^[A-Za-z0-9][A-Za-z0-9._-]*$", ...)`, matching the production submission contract. This rejects backslashes (including the `..\evil` ZIP traversal), shell metacharacters, whitespace, control characters, embedded NUL, and a trailing newline that a `$`-anchored match would accept. The pattern is defined once in `validation_utils.py` and `format_check.py` now references it, so the authoring and submission boundaries cannot drift. The explicit `BACKUP_ROOT` reserved-name check was removed since a leading dot can no longer pass validation. Rejection tests were extended accordingly.

nvidianz:
@greptileai review

nvkevlu (approved):
This PR is good for now, these are not blockers but can be addressed in the future:

[fed_job_config.py (line 191)]: exported roots now receive 0700 rather than the previous typical 0755. This only materially affects undocumented cross-user/shared-filesystem handoffs.
[format_check.py (line 40)]: the older production helper can still accept one trailing newline because it uses re.match with $. This predates the PR; the authoring/export paths fixed here use fullmatch.

PR #5064, opened 2026-08-07
Fix stale Recipe CLI documentation commands

ZiyueXu77:
## What changed

- Update the FLARE API tutorial notebook to export hello-numpy through the system-level `--export --export-dir` Recipe interface.
- Remove the obsolete `--job_dir` argument from the FedOBD LLM training command.

## Why

PR #5048 migrated Recipe examples away from per-example export flags, but these two callers live outside the migrated example directories and retained removed arguments.

The notebook command now exports to `/tmp/nvflare/jobs/job_config`, preserving the following cell path at `/tmp/nvflare/jobs/job_config/hello-numpy`. The FedOBD command only runs training, so it does not need an export directory.

## Validation

- Reproduced both removed-argument parser failures on current main.
- Corrected hello-numpy export command completed successfully and produced the expected named job folder.
- Corrected llm_hf command arguments were accepted by its parser.
- Notebook JSON validation passed.
- Recipe export migration regression tests passed: 2 tests.
- `./runtest.sh -s` passed.

### DIFF
--- a/examples/tutorials/flare_api.ipynb
+++ b/examples/tutorials/flare_api.ipynb
@@ -105,7 +105,7 @@
     "import sys\n",
     "\n",
     "# Run export\n",
-    "os.system(\"cd ../hello-world/hello-numpy && python job.py --export_config\")\n",
+    "os.system(\"cd ../hello-world/hello-numpy && python job.py --export --export-dir /tmp/nvflare/jobs/job_config\")\n",
     "\n",
     "# Copy exported job to transfer directory\n",
     "transfer_dir = \"/tmp/nvflare/poc/example_project/prod_00/admin@nvidia.com/transfer/\"\n",
--- a/examples/tutorials/flare_api.ipynb
+++ b/examples/tutorials/flare_api.ipynb
@@ -105,7 +105,7 @@
     "import sys\n",
     "\n",
     "# Run export\n",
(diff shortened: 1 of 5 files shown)


examples/tutorials/flare_api.ipynb
```diff
@@ -105,7 +105,7 @@
     "import sys\n",
     "\n",
     "# Run export\n",
-    "os.system(\"cd ../hello-world/hello-numpy && python job.py --export_config\")\n",
+    "os.system(f\"cd ../hello-world/hello-numpy && \\\"{sys.executable}\\\" job.py --export --export-dir /tmp/nvflare/jobs/job_config\")\n",
```
YuanTingHsieh, line 108: can you do a scan through all the examples and notebooks and documentation to remove all --export_config, --export_dir

YuanTingHsieh (commented):
the change of this pr is good, there might be more examples suggest do a complete scan through the repo

examples/tutorials/flare_api.ipynb
ZiyueXu77: Good point! Let me do that.

YuanTingHsieh (approved):

chesterxgchen (approved):

PR #5067, opened 2026-08-07
Fix TASK_READY heartbeat expiry during materialization

pcnudde:
## Problem

During external-process task ingestion, `TASK_READY` does not return until the trainer has downloaded, decoded, materialized, and queued the task. The previous code ignored heartbeat expiry only while a download transfer waiter was active. Once the raw download completed, long model materialization could exceed `heartbeat_timeout` and make the controller falsely declare a healthy trainer dead.

## Fix

Treat the entire pending `TASK_READY` exchange as the liveness boundary:

- do not apply heartbeat expiry between sending `TASK_READY` and receiving the trainer reply
- continue to terminate immediately on user abort, backend close, or trainer process death
- enforce `task_wait_timeout`, when configured, as an absolute wall-clock deadline
- refresh peer activity when the trainer successfully replies, so result waiting starts with a fresh lease
- retain transfer waiters only for existing transaction cleanup

This deliberately adds no progress protocol, materialization state machine, converter callbacks, or phase-specific timeout handling.

## Accepted tradeoff

A live but wedged trainer is no longer detected by heartbeat while `TASK_READY` is pending. A finite `task_wait_timeout` bounds that case; `task_wait_timeout=None` is intentionally unbounded. The timeout reports a single `TASK_READY timed out` error rather than a materialization-phase stall diagnosis.

## Validation

- `.venv/bin/python -m pytest tests/unit_test/app_common/executors/external_process_backend_test.py -q` — 104 passed
- `./runtest.sh -s` — passed (black, isort, flake8, agent-skill-lint)
- `git diff --check` — passed

The regression uses an already-completed transfer waiter, waits longer than `heartbeat_timeout`, returns `TASK_ACCEPTED`, and delivers the result afterward.

### DIFF
--- a/nvflare/app_common/executors/client_api/external_process_backend.py
+++ b/nvflare/app_common/executors/client_api/external_process_backend.py
@@ -854,15 +854,13 @@ class ExternalProcessBackend(CellBackendBase):
                     self._current_task = None
 
     def _send_task_ready(self, trainer: _TrainerSession, task_message: dict, abort_signal: Signal) -> Tuple[str, Any]:
-        """Send TASK_READY with native cancellation for abort and liveness failures."""
-        timeout = self._context.task_wait_timeout
+        """Send TASK_READY until reply, cancelling on abort, closure, process death, or deadline."""
+        max_timeout = self._context.task_wait_timeout
         transfer_waiters = []
+        started = time.monotonic()
 
         def _on_transaction_created(transaction):
             transfer_waiters.append(DownloadService.get_transfer_waiter(transaction.tx_id))
-
-        def _has_live_task_download():
-            return any(not waiter.done() for waiter in tuple(transfer_waiters))
 
         def _cancel_cause():
             if abort_signal.triggered or (self._context.launch_once and self._abort):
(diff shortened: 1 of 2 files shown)


nvflare/app_common/executors/client_api/external_process_backend.py
```diff
@@ -910,6 +904,8 @@ def _cancel_cause():
         if cause is not None:
             self._delete_task_transfers(transfer_waiters)
             return cause
+        if reply is not None and reply.get_header(MessageHeaderKey.RETURN_CODE) == CellReturnCode.OK:
```
YuanTingHsieh, line 907: **[P2] Re-evaluate the deadline after the reply**

`cancel.value` only returns a previously latched cause; it does not run `_cancel_cause()`. Because the Cell waiter returns `IS_SET` before polling its abort signal, a `TASK_ACCEPTED` arriving just after `task_wait_timeout` can reach this point with `cancel.value is None` and be accepted, violating the new absolute deadline. Please force a final cancellation/deadline evaluation after `send_request()` returns and add a delayed-reply test whose handler returns without polling the signal.

YuanTingHsieh (commented):
One timeout correctness issue.

nvflare/app_common/executors/client_api/external_process_backend.py
pcnudde: Fixed in b3635ba3b4efed6bf4ad684681020254aa2d8a84.

After `send_request()` returns, `_send_task_ready()` now evaluates `cancel.triggered` before reading its value. This forces the final abort/process/deadline evaluation even when Cell observes the reply event first.

I also replaced the existing deadline test with the requested race: the handler sleeps past `task_wait_timeout` and returns `TASK_ACCEPTED` without polling the signal. The late reply is rejected with the deadline error.

Validation:
- exact regression: 1 passed
- full external-process backend suite: 104 passed
- `./runtest.sh -s`: passed (black, isort, flake8, agent-skill-lint)
- `git diff --check`: passed

YuanTingHsieh (approved):

PR #5068, opened 2026-08-07
Fail fast when streamed blobs exceed size limit

nvidianz:
## What changed

- reject blobs that exceed `streaming_max_blob_size` on the sender before any frames are transmitted
- report an actionable `BlobSizeError` with configuration, sharding, and object-downloader guidance
- terminate an affected server job process when an oversized streamed response cannot be sent, allowing the job to fail instead of leaving clients polling indefinitely
- add boundary and server-job regression tests

## Root cause and impact

The existing size guard ran only on the receiver. When a single response exceeded the configured limit, the receiver stopped consuming it while the sender and surrounding job machinery did not receive a synchronous, job-level failure. The server job could therefore remain running indefinitely.

Validating the known blob size before streaming makes the error immediate and actionable, and the server-job response path now converts that fatal transport error into a job-process failure.

Jira: [FLARE-3037](https://jirasw.nvidia.com/browse/FLARE-3037)

## Validation

- `python -m pytest -q tests/unit_test/fuel/f3/streaming/blob_streamer_test.py tests/unit_test/fuel/f3/cellnet/cell_blob_size_test.py` (24 passed)
- streaming unit suite (380 passed)
- cellnet unit suite (214 passed)
- scoped Black, isort, flake8, and `git diff --check`

The repository-wide `./runtest.sh -s` reaches Black but is blocked by three unrelated generated `node_modules/.../flatted.py` files under the medical-devices example dashboard. All files changed by this PR pass the equivalent scoped checks.

### DIFF
--- a/nvflare/fuel/f3/streaming/blob_streamer.py
+++ b/nvflare/fuel/f3/streaming/blob_streamer.py
@@ -110,6 +110,12 @@ class BlobHandler:
         self.chunk_size = config.get_streaming_chunk_size(STREAM_CHUNK_SIZE)
         self.max_blob_size = config.get_streaming_max_blob_size()
 
+    def validate_size(self, headers: dict) -> Optional[BlobSizeError]:
+        size = (headers or {}).get(StreamHeaderKey.SIZE, 0)
+        if self.max_blob_size > 0 and isinstance(size, int) and size > self.max_blob_size:
+            return _make_blob_size_error(size, self.max_blob_size)
+        return None
+
     @staticmethod
     def _fail(stream: Stream, future: StreamFuture, error: StreamError):
         if hasattr(stream, "task"):
@@ -297,4 +303,11 @@ class BlobStreamer:
 
     def register_blob_callback(self, channel, topic, blob_cb: Callable, *args, **kwargs):
         handler = BlobHandler(blob_cb)
-        self.byte_receiver.register_callback(channel, topic, handler.handle_blob_cb, *args, **kwargs)
+        self.byte_receiver.register_callback(
+            channel,
+            topic,
+            handler.handle_blob_cb,
+            *args,
+            preflight_cb=handler.validate_size,
+            **kwargs,
+        )
--- a/nvflare/fuel/f3/streaming/blob_streamer.py
+++ b/nvflare/fuel/f3/streaming/blob_streamer.py
@@ -22,12 +22,20 @@ from nvflare.fuel.f3.message import Message
 from nvflare.fuel.f3.streaming.byte_receiver import ByteReceiver
 from nvflare.fuel.f3.streaming.byte_streamer import STREAM_CHUNK_SIZE, STREAM_TYPE_BLOB, ByteStreamer
 from nvflare.fuel.f3.streaming.stream_const import EOS, StreamHeaderKey
-from nvflare.fuel.f3.streaming.stream_types import Stream, StreamError, StreamFuture
+from nvflare.fuel.f3.streaming.stream_types import BlobSizeError, Stream, StreamError, StreamFuture
 from nvflare.fuel.f3.streaming.stream_utils import FastBuffer, callback_thread_pool, stream_thread_pool, wrap_view
 from nvflare.fuel.utils.buffer_list import BufferList, ConsumableBufferList
 from nvflare.security.logging import secure_format_traceback
 
 log = logging.getLogger(__name__)
+
+
+def _make_blob_size_error(size: int, limit: int) -> BlobSizeError:
+    return BlobSizeError(
+        f"Blob size {size} exceeds configured limit {limit} (streaming_max_blob_size). "
+        "Increase NVFLARE_STREAMING_MAX_BLOB_SIZE (or streaming_max_blob_size), "
+        "shard the object, or use the object downloader."
+    )
 
 
 class BlobStream(Stream):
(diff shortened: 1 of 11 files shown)


nvflare/fuel/f3/streaming/blob_streamer.py
```diff
@@ -264,6 +273,8 @@ def send(
             raise StreamError(f"BLOB is invalid type: {type(message.payload)}")
 
         blob_stream = BlobStream(message.payload, message.headers)
+        if self.max_blob_size > 0 and blob_stream.get_size() > self.max_blob_size:
```
YuanTingHsieh: **Asymmetric site limits can leave the original hang unresolved.** `streaming_max_blob_size` is configured independently at each site and is documented as a received-blob limit. If the server sender allows 8 GiB and the client receiver allows 4 GiB, a 5 GiB response passes this local check. The client then rejects it in `BlobTask` before the registered blob callback is scheduled; that rejection is returned asynchronously through `reply_future`, but `Adapter._send_response()` does not observe the future. The server job therefore is not exited, while the client waiter can time out and retry—the failure mode this PR is intended to eliminate. The reverse mismatch also makes a sender with a lower local limit reject a blob that the receiver could accept.

Could receiver-side rejection always notify the callback and propagate a structured `BlobSizeError`, with `_send_response()` also observing asynchronous `reply_future` failures? This sender-side check can remain as an optimization, but it should not be the sole correctness path. An asymmetric-limit regression test would cover both cases.

YuanTingHsieh (commented):
One correctness gap remains for asymmetric per-site blob limits.

nvflare/fuel/f3/streaming/blob_streamer.py
pcnudde: Confirmed independently on current head `235b0da8e`. I reproduced the sender path with a returned `StreamFuture` already failed by receiver rejection: `Adapter._send_response()` still returns without invoking the server-job exit path because it never observes the future. This leaves the high-sender/low-receiver asymmetric case unresolved, while the reverse configuration introduces a false rejection. I consider this blocking until the preflight is receiver-aware or the asynchronous failure is observed, with regression coverage for both directions.

nvidianz: Addressed in f601b7ca7. Blob size limits are now enforced only by the receiver, so a sender with a lower local receive limit no longer rejects a blob that the destination can accept. Early receiver-side rejection always schedules the registered blob callback, the stream ACK preserves BlobSizeError as a structured error, and Adapter._send_response() observes the asynchronous reply future so a rejected oversized server-job response exits the job process. Added regression coverage for both asymmetric directions and async failure handling. Validation: 116 focused tests, 381 streaming tests, and 216 cellnet tests passed.

nvflare/fuel/f3/streaming/blob_streamer.py
```diff
@@ -167,12 +186,12 @@ def handle_blob_cb(self, future: StreamFuture, stream: Stream, resume: bool, *ar
             else:
                 blob_task = BlobTask(future, stream, self.max_blob_size)
         except StreamError as ex:
-            self._fail(stream, future, ex)
+            self._fail_and_notify(stream, future, ex, args, kwargs)
```
YuanTingHsieh, line 195: **The final frame can win the race and discard this failure.** `ByteReceiver.process_chunk()` marks the `RxTask` completed when `FINAL` arrives before it schedules `BlobHandler`. If the blob fits in one frame (or the handler is delayed until after the final frame), this call reaches `_fail()` with `stream.task.completed` already true. `RxTask.stop(error)` then returns without setting the future error or sending a `BlobSizeError` ACK, while the sender may already have completed and removed its transmit task. I reproduced this with declared size 5 and receiver limit 4 in a single `FINAL` frame: `completed=True`, `failed=False`, `future.done()=False`, and only the normal ACK was emitted, so the original polling hang remains possible.

Could the declared-size check happen before receive completion, or could terminal success be deferred until `BlobHandler` accepts the stream? Please add an end-to-end regression test that sends a single-frame oversized blob through `ByteReceiver`/`BlobHandler` and asserts that the receiver future completes with `BlobSizeError`, an ACK carrying `ERROR_TYPE=BlobSizeError` is emitted, and the server-side reply future/job-exit path observes it.

YuanTingHsieh (commented):
One remaining terminal-state race can preserve the original oversized-response hang; see the inline comment. I reproduced it deterministically and included a suggested end-to-end regression test.

nvflare/fuel/f3/streaming/blob_streamer.py
nvidianz: Addressed in 9999e5897. BlobHandler now supplies a declared-size preflight callback to ByteReceiver, so an oversized blob is rejected before payload buffering, receive completion, or a normal final ACK. Streamed CellNet replies are sent reliably, which keeps the sender task/future alive until the receiver ACKs acceptance or rejection. The new single-frame regression drives the complete path and verifies the receiver future fails with BlobSizeError, completed remains false, the error ACK carries ERROR_TYPE=BlobSizeError, the sender reply future preserves BlobSizeError, and Adapter triggers the server-job exit path. Validation: 117 focused tests, 381 streaming tests, and 219 cellnet tests passed.

YuanTingHsieh (approved):
The latest implementation addresses the terminal-state race with receiver-side declared-size preflight, typed error propagation, reliable streamed replies, and the new single-frame end-to-end regression test. Approving.

Minor non-blocking follow-up: the PR description is now outdated. It still says oversized blobs are rejected on the sender before frames are transmitted and describes the failure as synchronous, while the final design validates the declared size on the receiver before payload admission and propagates the rejection asynchronously through the reliable reply future. The validation totals are stale as well. Please update the description to match the final implementation and current test counts.

PR #5072, opened 2026-08-07
Wait for client terminal outcomes before finalizing jobs

pcnudde:
## Summary

- have each client report its terminal child-process outcome, including success, through the existing server topic
- wait for active client outcomes before publishing the server terminal job status
- apply reported failures before releasing the reporting client from the completion barrier
- bound that wait with a configurable grace, then warn about unresolved clients and finalize from the server known outcome
- resolve missing client outcomes conservatively through existing heartbeat and dead-client handling

## Root cause

Server-side process exit was treated as sufficient evidence that the distributed job had completed. A client could detect result-publication failure only afterward, when the server had already published and cached `FINISHED:COMPLETED`. The existing fire-and-forget path reported only selected failure exit codes and did not participate in terminal-status ordering.

## Repair

This reuses the existing terminal-outcome message and adds one pending-client set per job. Each client makes one acknowledged terminal-outcome report. `JobRunner` defers finalization until the set is empty, and the server applies failure or abort classification before removing the reporting client. Admin abort bypasses the barrier, missing or dead clients are resolved through existing reconciliation paths, and authenticated late or duplicate reports receive an idempotent `OK`.

When the server process has exited but outcomes remain pending, `JobRunner` starts an in-memory per-job deadline. `client_outcome_wait_timeout` is configurable in the application configuration and defaults to 900 seconds. On expiry, the server logs the unresolved client names, releases the barrier, and runs the existing server-side status calculation; expiry does not synthesize a failure.

The change adds no client retry policy, persisted deadline, or new protocol abstraction.

## Validation

- 103 focused client/server lifecycle tests passed
- full `./runtest.sh -s` style gate passed
- no fresh live Kubernetes E2E was run; QA MR !185 remains the deployment-shaped validation boundary


nvflare/private/fed/server/job_runner.py
```diff
@@ -412,6 +431,9 @@ def _job_complete_process(self, engine):
                 if job_id not in engine.run_processes.keys():
                     job = self.running_jobs.get(job_id)
                     if job:
+                        with self.lock:
+                            if self._pending_client_outcomes.get(job_id) and not job.run_aborted:
```
YuanTingHsieh: **[P1] Do not let every active client hold terminal publication indefinitely.** `_pending_client_outcomes` contains every client that acknowledged `START_JOB`, regardless of `min_sites` or `required_sites`. A connected straggler whose child process never exits or never reports its outcome therefore prevents the job from reaching a terminal status indefinitely. Heartbeat reconciliation only helps after the client stops listing the job, and dead-client handling only helps after disconnection; this change deliberately has no bounded wait. Please add a timeout/grace period or narrow the set of authoritative clients, then finalize from the server's known outcome while recording any unresolved client outcomes separately.

YuanTingHsieh (commented):
One blocking design concern: terminal publication should not be held indefinitely by every active client.

nvflare/private/fed/server/job_runner.py
pcnudde: Fixed in 205361645. Once the server process exits with client outcomes still pending, JobRunner now starts an in-memory per-job grace deadline. The default is 900 seconds and is configurable with `client_outcome_wait_timeout`. Before expiry the existing barrier is unchanged. At expiry it logs the unresolved client names, releases the barrier, and runs the existing server-side status calculation, so a successful server outcome remains `FINISHED:COMPLETED` rather than becoming a synthetic failure. The deadline is cleaned up with the job; HA restore simply starts a fresh grace window. Added a focused before/after-deadline lifecycle test; 103 focused tests and the full style gate pass.

nvflare/private/fed/server/fed_server.py
```diff
@@ -852,11 +862,13 @@ def process_job_failure(self, request: Message):
             with self.engine.new_context() as fl_ctx:
                 self.logger.info(f"Failing job {job_id} due to reported failure from {client}: {reason}")
                 failure_code = code if code == ProcessExitCode.INFRASTRUCTURE_ERROR else ProcessExitCode.EXCEPTION
-                self.engine.job_runner.fail_run(job_id, failure_code, fl_ctx)
+                job_runner.fail_run(job_id, failure_code, fl_ctx)
         elif code in (ProcessExitCode.UNSAFE_COMPONENT, JobReturnCode.ABORTED):
             with self.engine.new_context() as fl_ctx:
                 self.logger.info(f"Aborting job {job_id} due to reported failure from {client}: {reason}")
-                self.engine.job_runner.stop_run(job_id, fl_ctx)
+                job_runner.stop_run(job_id, fl_ctx)
```
YuanTingHsieh, line 870: **[P1] Preserve a recorded failure when another client reports `ABORTED`.** Consider client A reporting `ProcessExitCode.EXCEPTION`: `fail_run()` records the execution failure and aborts the remaining clients. When client B exits because of that abort, it reports `JobReturnCode.ABORTED`, and this call to `stop_run()` sets `job.run_aborted = True`. `_job_complete_process()` then both bypasses any remaining pending outcomes and selects `FINISHED_ABORTED` before consulting the recorded exception, so the original failure is overwritten and additional client outcomes may be dropped. Please distinguish an explicit admin abort from a client terminal abort, or otherwise preserve failure precedence and allow only the explicit admin-abort path to bypass the barrier.

YuanTingHsieh (commented):
One remaining correctness concern in client terminal-outcome classification.

nvflare/private/fed/server/fed_server.py
pcnudde: Fixed in b13f216c4. Client-reported `ABORTED` now goes through `fail_run()` instead of `stop_run()`, so it does not set `job.run_aborted` or bypass outcomes still pending from other clients. `fail_run()` records `ABORTED` only when no outcome is already recorded: an existing execution/infrastructure failure remains authoritative, while a later real failure can replace an earlier `ABORTED` (and infrastructure retains highest precedence). Explicit admin abort still uses `stop_run()` and bypasses the barrier; existing `UNSAFE_COMPONENT` behavior is unchanged. The existing server classification and precedence tests now cover both orderings. 105 focused lifecycle tests and the full style gate pass.

YuanTingHsieh (approved):

PR #5075, opened 2026-08-10
Fix pass-through tensor download retention

YuanTingHsieh:
## Summary

- use count-based tensor-download completion for broadcasts explicitly marked `PASS_THROUGH`
- preserve receiver-identity-aware completion for direct broadcasts
- add regression coverage for both paths

Follow-up to #4965. Tracks the observed regression in FLARE-3104; the longer-term delegated logical-receiver design is tracked in FLARE-3107.

## Root cause

Broadcast encoding records the first-hop Cell targets as `receiver_ids`. For a direct broadcast, those targets are also the tensor consumers and identity-aware completion is correct.

With `PASS_THROUGH`, however, a target client-job Cell can forward the lazy tensor reference to an external trainer Cell. The trainer is allowed to download the valid reference, but the transaction records the trainer FQCN while waiting for the original client-job FQCN. Completion is therefore never reached, and the source transaction retains its model-sized `TensorDownloadable.base_obj` until timeout or run shutdown.

This change keeps `num_receivers` for pass-through broadcasts but omits first-hop `receiver_ids`, so the transaction completes after the expected number of downstream consumers download the object.

## Impact

In a fixed-aggregator Swarm workload, the stale source transactions retained roughly one 5 GiB model per aggregation round in the aggregator client-job process. The fix releases each broadcast transaction promptly and keeps post-round RSS bounded.

## Validation

Large-model qualification used four clients, external-process trainers, a fixed site-1 aggregator, PyTorch aggregation, and an exact 5 GiB synthetic model:

- unmodified offload path: cleanup `RssAnon` grew by approximately 4.943 GiB/round and reached about 190.12 GiB by round 29
- count-based pass-through fix: 6/6 rounds completed, all 7 broadcast transactions finished, zero transactions remained outstanding, and steady-round cleanup `RssAnon` stayed within a 0.84 GiB range
- patched run peak site-1 RSS: approximately 23.95 GiB

Local checks on current `main`:

- `214 passed` — Cell unit-test suite
- `61 passed` — streaming download transaction/object/starvation suites
- `1 passed, 4 deselected` — secure Swarm tensor disk-offload integration case
- `./runtest.sh -s` — passed

### DIFF
--- a/nvflare/fuel/f3/cellnet/cell.py
+++ b/nvflare/fuel/f3/cellnet/cell.py
@@ -255,8 +255,21 @@ class Cell(StreamCell):
         results = dict()
         future_to_target = {}
 
-        # encode the request now so each target thread won't need to do it again.
-        self._encode_message(request, abort_signal, num_receivers=len(targets), receiver_ids=targets)
+        # Encode the request now so each target thread won't need to do it again.
+        # For a direct broadcast, the routing targets are also the tensor download
+        # consumers, so the transaction can track their exact identities. With
+        # PASS_THROUGH, each target may forward the refs to another Cell (for
+        # example, an external trainer), and only the final consumer count is known.
+        # Pinning the transaction to the first-hop identities would prevent it from
+        # completing after those downstream consumers finish downloading.
+        pass_through = bool(request.get_header(MessageHeaderKey.PASS_THROUGH, False))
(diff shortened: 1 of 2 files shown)


nvkevlu (approved):
LGTM

PR #5079, opened 2026-08-10
Bind Collab calls to authenticated job participants

nvidianz:
## Summary

- bind Collab caller identity to the authenticated Cell origin and an exact per-job participant allowlist
- validate the versioned target/method envelope before buffering or decoding request arguments
- add generic ByteStreamer preflight rejection reporting through a reserved stream-error topic
- return correlated stream errors as `COMM_ERROR` replies for request/reply calls while logging late non-request errors

## Root cause

Collab dispatch trusted the caller and routing fields serialized in the request payload. It did not bind the claimed caller to the authenticated Cell origin, verify that the origin belonged to the receiving job, or reject an invalid target before receiving and decoding the request body.

## Impact

Cross-job and caller-spoofed Collab calls are rejected before application code or argument decoding runs. Valid server/client and same-job peer calls continue to use the existing public Collab API. Accepted streams do not incur an additional handshake or round trip.

Jira: [FLARE-3062](https://jirasw.nvidia.com/browse/FLARE-3062)

## Validation

- `python -m pytest -q tests/unit_test/fuel/f3/streaming` — 384 passed
- `python -m pytest -q tests/unit_test/fuel/f3/cellnet` — 212 passed
- `python -m pytest -q tests/unit_test/collab` — 109 passed
- scoped Black, isort, flake8, and `git diff --check`


nvflare/fuel/f3/streaming/byte_streamer.py
```diff
@@ -738,3 +753,43 @@ def _ack_handler(message: Message):
...
+            if tx_task and tx_task.cell is not self.cell:
+                tx_task = None
+
+        if not tx_task:
+            log_func = log.warning if message.get_header(StreamHeaderKey.STREAM_REQ_ID) else log.error
+            log_func(
+                f"Late stream error: stream_id={sid} channel={channel} topic={topic} "
+                f"sender={sender} receiver={origin}: {error}"
+            )
+            self._notify_error_callbacks(message)
```
YuanTingHsieh, line 834: **Blocking:** This forwards uncorrelated/late errors to callbacks even when there is no `tx_task`, so there is no trusted target/origin binding. A routable Cell can send `STREAM_ERROR_TOPIC` with an arbitrary stream ID and `ERROR_TYPE=BlobSizeError`; `Cell._process_stream_error()` finds no waiter and calls `os._exit(1)` in a server job cell. Please retain trusted correlation for late errors or otherwise validate the expected origin/context before invoking fatal callbacks. A wire-provided error type alone cannot authorize process termination.

YuanTingHsieh (commented):
Found two security blockers and one mixed-version behavior issue. Details are inline. I do not think this should merge as-is; the underlying security fix should still target 2.9 if these can be addressed and validated in time.

nvflare/collab/runtime/dispatch.py
```diff
@@ -18,20 +18,122 @@
...
+def prepare_for_remote_call(cell, app, logger, executor, participants: dict[str, str]):
     logger.info(f"register cb for cell {cell.get_fqcn()}: {type(cell)}")
+    authorizer = CollabCallAuthorizer(app, cell.get_fqcn(), participants, logger)
     adapter = Adapter(_submit_app_method, cell.core_cell.my_info, cell)
+    handler = _AuthorizedCallHandler(authorizer, adapter)
     cell.register_blob_cb(
         channel=MSG_CHANNEL,
         topic=MSG_TOPIC,
-        blob_cb=adapter.call,
+        blob_cb=handler,
```
YuanTingHsieh: **Blocking:** The authorizer is installed inside the blob callback. Before this handler runs, `ByteReceiver` has accepted the first chunk and `BlobHandler.handle_blob_cb()` constructs `BlobTask`; for a non-FOBS raw stream that can preallocate the peer-declared size (up to the default 4 GiB), and the reader may begin buffering. A cross-job sender can therefore force substantial resource use before its origin/target is rejected. The participant/envelope check needs a header-only admission point ahead of `BlobTask` construction and stream buffering.

docs/release_notes/flare_290.rst
```diff
@@ -7,6 +7,12 @@ What's New in FLARE v2.9.0
 Compatibility and Migration Notes
 =================================
 
+- Collab calls now carry a versioned authorization envelope and are accepted
+  only from authenticated participants in the same job. All sites that run a
+  Collab job must use NVFlare 2.9 or newer; a 2.8 or older Collab sender lacks
+  this envelope and its calls are rejected with ``COMM_ERROR``. The receiving
```
YuanTingHsieh: This `COMM_ERROR` guarantee is not reliable for a 2.8 sender. Authorization rejection happens asynchronously in the blob callback, while the default non-reliable 2.8 `TxTask` removes itself once local sending completes. The legacy ACK-topic error can then arrive after the task is gone, and 2.8 has no new ERROR-topic correlation callback, so the caller waits for a timeout instead. Please either make the rejection synchronously/correlatably observable to the old sender or document the timeout behavior.

nvflare/fuel/f3/streaming/byte_streamer.py
nvidianz: Addressed in a7362773a. ByteStreamer now retains a bounded trusted context for completed sends and invokes error callbacks only when origin, Cell instance, channel, topic, stream ID, and request ID match that context. Unknown or mismatched late errors are logged and ignored. Added regressions proving an uncorrelated forged BlobSizeError cannot reach the fatal Cell callback, while a legitimate correlated late receiver error still does.

nvflare/collab/runtime/dispatch.py
nvidianz: Addressed in a7362773a. Collab now installs a Collab-owned CoreCell incoming filter on stream data. It authorizes the authenticated routing/envelope headers before ByteReceiver._data_handler runs, so rejected cross-job frames cannot create RxTask or BlobTask state or start buffering. ByteReceiver only supplies the generic error-reporting helper. The end-to-end regression replaces BlobTask with a failure sentinel and verifies the unauthorized request returns COMM_ERROR without allocation.

docs/release_notes/flare_290.rst
nvidianz: Addressed in a7362773a. The 2.9 release note now distinguishes the behaviors: a 2.9 sender receives immediate COMM_ERROR through the correlated stream-error path, while a 2.8-or-older sender may observe a request timeout because it lacks that dedicated correlation mechanism.

pcnudde (dismissed):

YuanTingHsieh (approved):

PR #5092, opened 2026-08-11
Add portable CPU and memory job resources

### LINKED ISSUE #5083: Add portable CPU and memory requirements with resource defaults
## Problem

`resource_spec` has a portable GPU count, but CPU and host-memory requirements are launcher-specific. Jobs must also repeat identical requirements for every target site.

## Proposal

Expose only these portable fields:

- `num_of_gpus`: integer greater than or equal to 0; whole GPU devices per job worker.
- `num_of_cpus`: integer greater than or equal to 1; whole schedulable CPU units per job worker.
- `memory`: positive integer followed by `Mi`, `Gi`, or `Ti`; host memory per job worker.

Add `resource_spec["@default"]`, shallow-merged with each site; site values override defaults:

```json
{
  "resource_spec": {
    "@default": {
      "num_of_gpus": 1,
      "num_of_cpus": 4,
      "memory": "8Gi"
    },
    "site-2": {
      "num_of_gpus": 8,
      "num_of_cpus": 16
    }
  }
}
```

`@default` avoids conflicting with a valid site named `default`.

## Backend translation

| Field | Docker | Kubernetes | Slurm |
|---|---|---|---|
| `num_of_gpus` | GPU device requests | `nvidia.com/gpu` request and limit | `--gres` |
| `num_of_cpus` | `nano_cpus` | CPU request and limit | `--cpus-per-task` |
| `memory` | `mem_limit` in bytes | Memory request and limit | `--mem` in MiB |

The scheduler and launcher must use the same resolved default-plus-site requirements.

## Compatibility and scope

- Existing per-site and launcher-native jobs remain valid.
- Reject portable CPU or memory combined with the equivalent launcher-native field.
- Keep topology, accelerator type, GPU memory, storage, wall time, and other backend-specific options in `launcher_spec` or site policy.
- GPU memory is deliberately excluded because Docker, Kubernetes, and Slurm do not share a portable, enforceable allocation by GPU-memory amount.

## Acceptance criteria

- Validate the three portable fields exactly as described above.
- Resolve `@default` with shallow site overrides and without mutating job metadata.
- Apply defaults to explicitly targeted sites that do not have a site entry.
- Translate CPU and memory consistently in Docker, Kubernetes, and Slurm.
- Preserve legacy flat and nested resource specifications.
- Cover default resolution, site overrides, target expansion, malformed values, conflicts, translations, and compatibility with focused tests.

pcnudde:
## Summary

- add portable `num_of_cpus` and host `memory` requirements alongside `num_of_gpus`
- add `resource_spec["@default"]` with shallow per-site overrides
- use the same resolved requirements for scheduling, dispatch, cancellation, and launcher enforcement
- translate the portable fields to Docker, Kubernetes, and Slurm while preserving legacy resource specifications

## Backend mappings

| Field | Docker | Kubernetes | Slurm |
|---|---|---|---|
| `num_of_gpus` | GPU device requests | `nvidia.com/gpu` request and limit | `--gres` |
| `num_of_cpus` | `nano_cpus` | CPU request and limit | `--cpus-per-task` |
| `memory` | `mem_limit` in bytes | memory request and limit | `--mem` in MiB |

GPU memory and other backend-specific topology or policy remain outside the portable contract.

## Validation

- `557 passed` across the focused scheduler, validator, utility, Docker, Kubernetes, and Slurm tests
- `./runtest.sh -s`
- `git diff --check`

Closes #5083


nvflare/utils/job_launcher_utils.py
```diff
@@ -127,6 +128,16 @@ def generate_server_command(fl_ctx) -> str:
...
 
+PORTABLE_RESOURCE_DEFAULT_KEY = "@default"
+PORTABLE_RESOURCE_KEYS = ("num_of_gpus", "num_of_cpus", "memory")
+_PORTABLE_MEMORY_PATTERN = re.compile(r"^([1-9][0-9]*)(Mi|Gi|Ti)$")
+_MEMORY_UNIT_TO_MIB = {"Mi": 1, "Gi": 1024, "Ti": 1024 * 1024}
+_PORTABLE_NATIVE_RESOURCE_KEYS = {
+    "docker": {"num_of_cpus": {"nano_cpus"}, "memory": {"mem_limit"}},
+    "k8s": {"num_of_cpus": {"cpu", "cpu_request"}, "memory": {"memory", "memory_request"}},
+    "slurm": {"num_of_cpus": {"cpus_per_node"}, "memory": {"mem_per_node"}},
+}
```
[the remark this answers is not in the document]
pcnudde: Fixed in `048c8bd31`. Job admission now rejects portable `num_of_gpus` combined with Docker `num_of_gpus` or `device_requests`, and with Kubernetes `num_of_gpus`, so launcher allocation cannot exceed the scheduler-reserved portable count. Slurm topology remains allowed because its existing launcher validation requires `num_of_gpus == nodes * gpus_per_node`. Added helper coverage for default conflicts and matching Slurm topology, plus an end-to-end `JobMetaValidator` test for the site-level K8s conflict. Validation: 562 focused tests passed and `./runtest.sh -s` passed.

pcnudde:
@greptileai Please re-evaluate the remaining 4/5 summary against the current head.

The summary treats a flat `num_of_gpus` beside a mode key such as `docker` as a portable request. Existing NVFlare semantics intentionally treat any site block containing a mode key as the legacy nested format; flat sibling fields are ignored. This predates this PR and is explicitly covered by `test_launch_num_of_gpus_mixed_with_mode_keys_ignored`.

Therefore, in the cited mixed example the NVFlare resource manager receives no GPU requirement and Docker receives the nested launcher-native GPU value. That is legacy nested launcher behavior, not a portable/native conflict introduced by this PR. Rejecting mixed shapes would be a separate backward-incompatible validation change, contrary to this PR's explicit preservation of nested resource specifications without `@default`.

Both Greptile inline threads are resolved and the Greptile check passes. Please refresh the summary and confidence score with this compatibility boundary in mind.

nvflare/utils/job_launcher_utils.py
```diff
@@ -147,6 +166,126 @@ def get_launcher_resource_spec(job_meta, site_name, mode):
     return get_site_launcher_spec(resource_spec.get(site_name), mode)
 
 
+def _validate_portable_values(resource_spec: dict, label: str) -> None:
+    num_gpus = resource_spec.get("num_of_gpus")
+    if num_gpus is not None and (isinstance(num_gpus, bool) or not isinstance(num_gpus, int) or num_gpus < 0):
```
YuanTingHsieh: Could we distinguish an absent field from an explicitly supplied `null` here? The current `.get()` / `is not None` checks accept `null` for all three portable fields even though their types are part of the contract. The resolved spec then retains the key: Docker can fail while multiplying/converting CPU or memory, and Kubernetes can emit invalid resource quantities. Please validate whenever the key is present and add `null` cases for both `@default` and site overrides.

docs/user_guide/core_concepts/job.rst
```diff
@@ -102,10 +102,37 @@ The system also keeps additional information about the job such as:
...
+                "num_of_gpus": 8,
+                "num_of_cpus": 16
+            }
+        }
+    }
+
+Docker enforces these as GPU device requests, CPU quota, and a memory limit. Kubernetes sets equal requests and limits
+for GPUs, CPU, and memory. Slurm maps them to ``--gres``, ``--cpus-per-task``, and ``--mem``. GPU memory is not portable;
+MIG, device-plugin sharing, and typed GRES remain site-specific. Site-specific resource managers may continue to use
+additional fields in per-site blocks, but ``@default`` accepts only the three portable fields.
```
YuanTingHsieh: Should we explicitly name the existing `mem_per_gpu_in_GiB` field and show it in a per-site block alongside `@default`? It remains a valid `GPUResourceManager` VRAM requirement and the resolver preserves it, but readers may confuse portable host `memory` with GPU memory or try putting `mem_per_gpu_in_GiB` under `@default`, where validation rejects it. A small combined example—and ideally a compatibility test using this real field rather than only a generic custom resource—would make the boundary clear.

But that is only useful if GPUResourceManager is configured

YuanTingHsieh (commented):
Left two comments on portable-value validation and GPU-memory compatibility documentation.

nvflare/utils/job_launcher_utils.py
pcnudde: Fixed in 7a5e34cc6. Validation now checks field presence instead of treating None as absence, so an explicit null is rejected for all three portable fields. Added null coverage for each field in both @default and site overrides. Validation: 573 affected tests and the full style gate passed.

docs/user_guide/core_concepts/job.rst
pcnudde: Fixed in 7a5e34cc6. The resource example now shows mem_per_gpu_in_GiB in a per-site block and explains that memory is portable host memory, while mem_per_gpu_in_GiB is a site-specific VRAM requirement used only when GPUResourceManager is configured and is not allowed under @default. The compatibility test now uses mem_per_gpu_in_GiB and verifies it reaches resource-manager requirements alongside the portable GPU default. Validation: 573 affected tests and the full style gate passed.

nvflare/utils/job_launcher_utils.py
```diff
@@ -147,6 +166,132 @@ def get_launcher_resource_spec(job_meta, site_name, mode):
...
+    """Reject simultaneous portable and equivalent launcher-native resource fields."""
+    resource_spec = job_meta.get(JobMetaKey.RESOURCE_SPEC.value, {}) or {}
+    launcher_spec = job_meta.get(JobMetaKey.JOB_LAUNCHER_SPEC.value, {}) or {}
+    site_names = (set(resource_spec) - {PORTABLE_RESOURCE_DEFAULT_KEY}) | (set(launcher_spec) - {"default"})
+    for site_name in site_names:
+        portable = get_portable_resource_spec(job_meta, site_name)
+        for mode, portable_to_native in _PORTABLE_NATIVE_RESOURCE_KEYS.items():
+            native = get_job_launcher_spec(job_meta, site_name, mode)
+            for portable_key, native_keys in portable_to_native.items():
+                conflicts = sorted(native_keys & set(native)) if portable_key in portable else []
```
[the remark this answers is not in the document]
pcnudde: Fixed in `877c21881`. Legacy nested specs now require the GPU count reserved through `process.num_of_gpus` to agree with every effective launcher request: Docker and Kubernetes `num_of_gpus` must match, Slurm `nodes * gpus_per_node` must match, and Docker `device_requests` is rejected because its GPU count cannot be safely inferred for admission. Launcher-only legacy specs and the established behavior for flat siblings beside mode blocks remain unchanged. Coverage includes matching and mismatched counts for Docker, Kubernetes, and Slurm; Docker `device_requests`; a `launcher_spec` override; backward-compatibility cases; and end-to-end job metadata validation. Validation after merging current `main`: 583 affected tests passed and the full style gate passed.

YuanTingHsieh (commented):
Found one remaining backward-compatibility issue in zero-GPU resource-manager projection.

nvflare/utils/job_launcher_utils.py
```diff
@@ -147,6 +166,166 @@ def get_launcher_resource_spec(job_meta, site_name, mode):
...
+
+def get_resource_manager_spec(job_meta: dict, site_name: str) -> dict:
+    """Return resources enforced by the site's NVIDIA FLARE resource manager."""
+    resource_spec = job_meta.get(JobMetaKey.RESOURCE_SPEC.value, {}) or {}
+    site_spec = resource_spec.get(site_name) or {}
+    if PORTABLE_RESOURCE_DEFAULT_KEY not in resource_spec and any(key in site_spec for key in _LAUNCHER_MODE_KEYS):
+        return get_site_launcher_spec(site_spec, "process")
+    resolved = resolve_site_resource_spec(job_meta, site_name)
+    result = {key: value for key, value in resolved.items() if key not in ("num_of_cpus", "memory")}
+    if result.get("num_of_gpus") == 0:
```
YuanTingHsieh, line 225: **[P1] Normalize the legacy zero-GPU pair atomically**

Could we avoid removing only `num_of_gpus` here? Existing CPU-only jobs—including the feature-election generator—emit `{"num_of_gpus": 0, "mem_per_gpu_in_GiB": 0}`. This projection leaves `{"mem_per_gpu_in_GiB": 0}`, and `GPUResourceManager` rejects any non-empty requirement missing its GPU-count key, so those jobs can no longer be scheduled.

For this PR, the least-risk fix seems to be normalizing the standard zero-count/zero-VRAM pair to `{}` together (while preserving other custom-resource shapes) and adding a scheduler regression test using that legacy pair with `GPUResourceManager`. `GPUResourceManager`'s broader zero-count behavior—currently not reserving `{}` consistently—can be cleaned up separately without expanding this PR's allocation/consumer behavior.

pcnudde: Fixed in `07c7d8f32`. `get_resource_manager_spec` now removes `mem_per_gpu_in_GiB` together with `num_of_gpus` when both resolve to zero, while preserving unrelated custom resource requirements. Added a projection test that retains a custom `license` requirement and an end-to-end scheduler regression using the legacy zero-GPU/zero-VRAM pair with a zero-capacity `GPUResourceManager`. The broader `GPUResourceManager` zero-count behavior remains unchanged and out of scope, as suggested. Validation: 585 affected tests passed, the projection matrix passed, the full style gate passed, and `git diff --check` passed.

YuanTingHsieh (approved):

PR #5094, opened 2026-08-11
Preserve Slurm worker failure reasons

### LINKED ISSUE #5093: Slurm-launched jobs can fail without reporting the underlying worker error
## Problem

A client job launched through Slurm can terminate without its failure reason reaching the controller. From the user's perspective, the job ends or the worker disappears, but no actionable infrastructure or process failure is reported.

This occurs in two cases:

1. **A non-primary job step fails.** With `--kill-on-bad-exit`, the primary task and allocation can end with a signal or generic status while Slurm records the actual highest step exit code in `DerivedExitCode`. The launcher currently reads only the allocation's `ExitCode`, so a reportable worker failure can be lost.
2. **The worker fails before checking in.** If container or process startup fails before the worker reaches the started state, the scheduler returns a generic execution error. Because the worker never produced its own return code, that generic result is not surfaced as an infrastructure failure.

## Expected behavior

- Preserve a recognized, reportable process exit code from Slurm's `DerivedExitCode`.
- Treat a generic execution error as an infrastructure/startup failure only when the worker is still in the starting state.
- Leave ordinary application failures unchanged after the worker has checked in.

## Suggested scope

- Extend the existing allocation-only `sacct -X` query to include `DerivedExitCode`; do not add a second step-row query.
- Fall back to the existing behavior for unrecognized derived exit codes.
- Add regression coverage for both the non-primary step failure and pre-start failure paths.

Slurm documents `DerivedExitCode` as the highest exit code returned by a job's steps: https://slurm.schedmd.com/job_exit_code.html

pcnudde:
## What changed

- Include Slurm's `DerivedExitCode` in the existing allocation-only accounting query and preserve it when it is a recognized, reportable worker exit code.
- Classify a generic launcher execution error as an infrastructure failure only when the worker never advanced beyond the starting state.
- Add regression coverage for failures from non-primary Slurm tasks and failures before worker check-in.

## Why

For multi-task jobs, the allocation's `ExitCode` can describe the signal that terminated the primary task while `DerivedExitCode` retains the original failing step's exit code. Separately, a container or process that fails before worker check-in cannot produce its own return-code file. Both paths could therefore end without an actionable failure reaching the controller.

This keeps ordinary application failures unchanged after the worker has started and avoids adding another Slurm accounting query.

## Validation

- `pytest -q tests/unit_test/app_opt/job_launcher/slurm_scheduler_client_test.py tests/unit_test/app_opt/job_launcher/slurm_manager_test.py tests/unit_test/private/fed/client/client_executor_test.py` — 112 passed
- `./runtest.sh -s`
- `git diff --check`

Live Slurm execution was not repeated for this change.

Fixes #5093


nvflare/app_opt/job_launcher/slurm/config.py
```diff
@@ -44,7 +44,7 @@
 DEFAULT_MULTINODE_PORT_RANGE = (29400, 30399)
 
 SQUEUE_FORMAT = "%i|%T|%U|%k|%j"
-SACCT_FORMAT = "JobIDRaw%32,JobName%128,User%64,State%64,ExitCode%32"
+SACCT_FORMAT = "JobIDRaw%32,JobName%128,User%64,State%64,ExitCode%32,DerivedExitCode%32"
```
chesterxgchen, line 47: why the format needs to be in this way ? %32 etc.

nvflare/private/fed/client/client_executor.py
```diff
@@ -614,6 +614,10 @@ def _wait_child_process_finish(
 
             return_code = get_return_code(job_handle, job_id, workspace, self.logger)
 
+            process_status = self.run_processes.get(job_id, {}).get(RunProcessKey.STATUS)
+            if return_code == JobReturnCode.EXECUTION_ERROR and process_status == ClientStatus.STARTING:
+                return_code = ProcessExitCode.INFRASTRUCTURE_ERROR
```
chesterxgchen, line 619: why this is infra error ?  ( job failed with error, process is starting )

chesterxgchen (commented):
few questions

nvflare/private/fed/client/client_executor.py
pcnudde: `STARTING` is the CP's "worker has not checked in" state. The worker notifies the CP with `STARTED` before application execution begins. Therefore, a generic launcher execution error while it remains `STARTING` means the job ended before a live worker established itself. Explicit worker codes such as 101-103 are preserved, and generic failures after `STARTED` remain unchanged.

nvflare/app_opt/job_launcher/slurm/config.py
pcnudde: `%32` is `sacct`'s output-width modifier, not part of the field value. The existing query already uses explicit widths to avoid truncation; this adds `DerivedExitCode` with the same width as `ExitCode`. With `-P`, the fields remain `|`-delimited, and the parser strips the field before parsing its normal `status:signal` value.

chesterxgchen (approved):

PR #5103, opened 2026-08-12
Correct documented F3 streaming defaults

nvkevlu:
Correct documented F3 streaming defaults.

### Description

Align the communication guide with the effective runtime defaults: a 4 GiB maximum streamed blob size and 300-second read and ACK timeouts.

### Types of changes

- [ ] Documentation updated.

### DIFF
--- a/docs/user_guide/admin_guide/configurations/communication_configuration.rst
+++ b/docs/user_guide/admin_guide/configurations/communication_configuration.rst
@@ -331,6 +331,7 @@ are intentionally omitted.
     "streaming_window_size": 67108864,
     "streaming_ack_interval": 16777216,
     "streaming_ack_wait": 300,
+    "streaming_ack_progress_timeout": 60.0,
     "streaming_reliable": false,
     "streaming_retry_wait": 5.0,
     "streaming_retry_timeout": 60.0,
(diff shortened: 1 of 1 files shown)


docs/user_guide/admin_guide/configurations/communication_configuration.rst
```diff
@@ -453,7 +453,7 @@ streaming_ack_wait
 ------------------
 
 The number of seconds that the sender waits for the next ACK.
-The default value is 10 seconds. 
+The default value is 300 seconds.
```
YuanTingHsieh, line 457: The numeric default is correct, but this reads as if the sender will tolerate 300 seconds with no ACK. `TxTask` also defaults `streaming_ack_progress_timeout` to 60.0 seconds and checks that guard before `streaming_ack_wait`, so a stalled stream normally fails after 60 seconds without ACK progress. Could we document both controls (and add `streaming_ack_progress_timeout` to the fixed-default example) so operators know the effective timeout is whichever expires first?

YuanTingHsieh (commented):
One documentation clarification:

docs/user_guide/admin_guide/configurations/communication_configuration.rst
nvkevlu: Thanks, this is now clarified

YuanTingHsieh:
The documentation issue is addressed and the current CI checks are green. One remaining readiness concern: this PR now also includes a substantial runtime change that disables inbound admin/HCI services in Simulator, while the title and description still present it as a documentation-only correction. Could you either split the Simulator change into a separate PR, or update this PR’s title and description to cover that behavior change (and mark the documentation/tests checkboxes accordingly)?

YuanTingHsieh (approved):

PR #5104, opened 2026-08-12
Require BYOC for Slurm node commands

pcnudde:
## Summary

- classify a non-null effective Slurm `additional_node_command` as BYOC during local app deployment
- enforce the existing BYOC authorization policy on both server and client deployment receivers
- cover default, site-specific, and legacy launcher metadata resolution, including null and absent values

## Root cause

Local BYOC derivation considered job-selected launcher images and Python paths, but not the executable command supplied for additional Slurm nodes. A submitter allowed to submit jobs but denied BYOC could therefore provide a command that the Slurm launcher would execute on nonzero nodes of a multi-node client job.

## Impact

Jobs with an effective non-null `additional_node_command` now require the submitter to have the BYOC right. Explicit `null` continues to disable an inherited default command, and the existing Slurm client-only execution restriction is unchanged.

## Validation

- `python -m pytest -q tests/unit_test/private/fed/utils` (`81 passed`)
- `./runtest.sh -s`
- `git diff --check`


nvflare/private/fed/utils/app_authz.py
```diff
@@ -48,7 +48,11 @@ def validate_app(app_path: str) -> (str, dict):
     def derive_local_app_info(app_info: dict, job_meta: dict, site_name: str) -> dict:
         for mode in ("docker", "k8s", "slurm"):
             spec = get_job_launcher_spec(job_meta, site_name, mode)
-            if "image" in spec or "python_path" in spec:
+            if (
+                "image" in spec
+                or "python_path" in spec
+                or (mode == "slurm" and spec.get("additional_node_command") is not None)
```
chesterxgchen, line 54: if we have branch for mode = slurm, then should we consider if k8s, if dockers , if other env ? 

seems that we need more general mechanism than if else

chesterxgchen (commented):
approved with some comments for future improvements

chesterxgchen (approved):

nvflare/private/fed/utils/app_authz.py
pcnudde: image and pyhtonpath are for docker and k8s. the additional_node command is a slurm only thing, but yes, if we this becomes more complicated we need something a bit less ad hoc

PR #5107, opened 2026-08-12
Fix external-process authentication race

YuanTingHsieh:
## Summary

- add a mandatory `SESSION_READY` acknowledgement after a managed trainer processes `HELLO_ACCEPTED` and installs the delegated site authentication headers
- keep `ExternalProcessBackend` from exposing the trainer to task execution until that acknowledgement is validated
- add deterministic secure external-process streaming coverage that widens the original race window

## Problem

A secure `external_process` Client API job could intermittently hang when a streamed `TASK_READY` payload was sent immediately after the trainer's `HELLO`.

The backend set `trainer.ready` inside `_handle_hello()` before the `HELLO_ACCEPTED` reply reached the trainer. The execution thread could therefore start `TASK_READY` while the trainer was still processing `HELLO_ACCEPTED` and had not installed the delegated site authentication filters.

If the task contained lazy or streamed model parameters, the trainer could contact the server without `client_name`, `AUTH_TOKEN`, or `AUTH_TOKEN_SIGNATURE`. The server rejected the first streaming message, and the job could remain running.

## Root cause

The protocol treated these two events as equivalent:

1. the backend accepted the trainer's `HELLO`
2. the trainer finished processing `HELLO_ACCEPTED` and installed the delegated site credentials

They must be ordered separately before task dispatch is allowed.

## Fix

The trainer now sends `SESSION_READY(session_id)` only after it:

1. validates `HELLO_ACCEPTED`
2. installs the delegated authentication-header filters
3. binds the accepted session and heartbeat policy

The backend validates the active session, trainer origin, session ID, and—when secure—the exact delegated client name, token, and signature before setting `trainer.ready`.

The resulting ordering is:

`HELLO_ACCEPTED -> install delegated auth headers -> SESSION_READY -> trainer.ready -> TASK_READY`

Invalid, stale, wrong-origin, or incorrectly authenticated readiness messages do not release task execution.

`SESSION_READY` is mandatory in the initial public Client API Cell protocol v1 contract. This protocol exists on the development branch and was not part of the 2.8 release.

## Scope

This PR does not change CoreCell filtering, ByteStreamer/F3 delivery, DownloadService transport behavior, reliable-stream semantics, or `StreamFuture` lifecycle handling. It uses the existing Cell request/reply mechanism solely to establish the missing handshake ordering.

## Validation

Exact signed head: `7cc1c145efd1491c44c702d7a32cdf11795c6d0d`.

- 235 affected Client API/backend tests passed, 2 skipped, on the exact head
- review regressions cover a slow HELLO followed by a fresh SESSION_READY budget, accurate SESSION_READY timeout diagnostics, non-ASCII delegated credentials, and teardown/readiness lock ordering
- full style validation passed: Black, isort, flake8, and agent-skill lint
- maintained secure two-client external-process integration passed with a built-in two-second authentication-install delay
- both inline and greater-than-8-MiB streamed task/result cases completed successfully
- a no-delay control also passed
- authentication installation completed before `SESSION_READY`, backend readiness, and `TASK_READY` for both trainers
- no `unauthenticated msg`, `missing client name`, hang, or orphan SJ/CJ/trainer process was observed


nvflare/fuel/f3/cellnet/core_cell.py
```diff
@@ -2211,6 +2228,55 @@ def _process_received_msg(self, endpoint: Endpoint, connection: Connection, mess
             # the message is either a reply or a return for a previous request: handle replies
             self._process_reply(origin, message, msg_type)
 
+    def _send_stream_filter_rejection(self, request: Message, filter_rejection: object) -> None:
+        """Fail a ByteStreamer sender immediately when an incoming filter rejects a data frame.
+
+        Stream data frames are fire-and-forget Cell requests, so the ordinary filter
+        reply has no waiter and would otherwise be dropped. ByteStreamer already has
+        an error-ACK path; use it to terminate the sender's stream future.
+        """
```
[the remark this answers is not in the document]
YuanTingHsieh: Good catch. This race remains possible for arbitrary raw non-reliable streams. For the paths in this PR, I addressed it without changing global non-reliable semantics: Cell request streaming now supports an explicit reliability setting, and managed Client API requests plus DownloadService chunk requests use reliable=True (Cell replies were already reliable). A reliable TxTask remains registered until it receives a sequence ACK or ERROR, so a delayed incoming-filter error terminates the stream future instead of being dropped. test_reliable_task_stays_registered_for_delayed_filter_error covers that timing. Raw non-reliable send_blob behavior remains unchanged, and this PR does not claim a generic guarantee for those streams.

nvflare/client/cell/defs.py
```diff
@@ -31,6 +31,9 @@ class Topic:
     HELLO_ACCEPTED = "client_api.hello_accepted"
     # Semantic handshake rejection, distinct from transport/protocol ERROR.
     HELLO_REJECTED = "client_api.hello_rejected"
+    # Trainer acknowledgement sent only after HELLO_ACCEPTED has been fully
+    # processed, including installation of delegated site authentication.
+    SESSION_READY = "client_api.session_ready"
```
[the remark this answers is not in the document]
YuanTingHsieh: Thanks for flagging this. SESSION_READY was added before the first public release of the Client API Cell protocol. This external-process protocol is not present in the released 2.8.x line, and no 2.9 distribution containing the earlier implementation has been published, so compatibility between arbitrary unreleased main revisions is not part of the supported wire contract. Protocol v1 is therefore the initial public contract and includes mandatory SESSION_READY. I updated the protocol comment and design documentation to make that boundary explicit; future incompatible changes after v1 ships should increment the version.

chesterxgchen (commented):
- [P1] Rejection ACK trusts identity from an unauthenticated frame. _send_stream_filter_rejection()
    (https://github.com/NVIDIA/NVFlare/blob/195f11dafa1c7042e67e241149b89649d080f7fc/nvflare/fuel/f3/cellnet/core_cell.py#L2243-L2270) takes ORIGIN and STREAM_ID from the message that just
    failed an incoming filter—which may be the authentication filter—and sends a server-originated error ACK to that claimed origin. On receipt, _ack_handler()
    (https://github.com/NVIDIA/NVFlare/blob/195f11dafa1c7042e67e241149b89649d080f7fc/nvflare/fuel/f3/streaming/byte_streamer.py#L726-L740) selects a transfer solely by stream ID, and
    handle_ack() (https://github.com/NVIDIA/NVFlare/blob/195f11dafa1c7042e67e241149b89649d080f7fc/nvflare/fuel/f3/streaming/byte_streamer.py#L512-L523) does not verify the ACK origin against
    TxTask.target.

    A modified client can therefore submit a deliberately unauthenticated stream frame claiming another client as ORIGIN. If its stream ID matches an active transfer, the server sends a
    trusted error ACK to the victim and terminates that transfer. Stream IDs are time-derived and incrementing, which makes guessing more realistic.

    The ACK needs a correlation value the rejected sender cannot forge—such as a cryptographically random per-stream nonce copied into every DATA/ACK frame—and the sender should validate both
    that nonce and the ACK origin. Add a regression test showing that a rejected frame with a spoofed origin cannot stop another peer’s TxTask.

  Everything else looks sound:

  - SESSION_READY correctly closes the authentication-installation race.
  - Backend readiness is withheld until the delegated headers are confirmed.
  - Reliable streaming on the affected Client API and download paths closes the delayed error-ACK race.
  - Keeping protocol v1 is reasonable because this is still its initial public contract.
  - The newly added lifecycle integration coverage is useful.

YuanTingHsieh:
Addressed the P1 stream-rejection ACK correlation issue in signed commit `9608c07185b848e7d7b52eab11ef41e142b0db19`.

Each `TxTask` now creates a cryptographically random 128-bit stream token and carries it on every DATA/error frame. `RxTask` records and echoes that token on ACK/error frames, and CoreCell copies it into incoming-filter rejection ACKs. Before mutating a `TxTask`, the sender now requires both the expected ACK origin and a constant-time token match; receiver-side frames with a mismatched token are also ignored.

The regression explicitly verifies that neither a correct peer with a forged token nor a different origin with the correct token can terminate the active stream, while the fully correlated rejection still fails it promptly.

Validation: 131 focused streaming/CoreCell security tests passed, 555 broader streaming/CellNet tests passed, and the full project style gate passed.

chesterxgchen (commented):
[P2] Routed filter-rejection ACKs are ignored

  Incoming filters execute before the destination/forwarding check
  (https://github.com/NVIDIA/NVFlare/blob/9608c07185b848e7d7b52eab11ef41e142b0db19/nvflare/fuel/f3/cellnet/core_cell.py#L2083-L2138). Both the secure server
  (https://github.com/NVIDIA/NVFlare/blob/9608c07185b848e7d7b52eab11ef41e142b0db19/nvflare/private/fed/server/fed_server.py#L1145-L1156) and relay
  (https://github.com/NVIDIA/NVFlare/blob/9608c07185b848e7d7b52eab11ef41e142b0db19/nvflare/private/fed/app/relay/relay.py#L212-L226) register wildcard
  filters, so they can reject transit frames destined for another client.

  The rejection ACK is then generated by that intermediate cell, but TxTask.handle_ack() requires origin == self.target
  (https://github.com/NVIDIA/NVFlare/blob/9608c07185b848e7d7b52eab11ef41e142b0db19/nvflare/fuel/f3/streaming/byte_streamer.py#L516-L525). Therefore:

  trainer A → server/relay → aggregation client B
                   rejects
  trainer A ← ACK origin=server/relay
                   ignored because target=B

  The stream eventually fails through retry/timeout instead of the intended immediate rejection. This matters for Swarm, peer-result, and hierarchical/relay
  deployments, without any mixed-version deployment.

  Suggested fix: after validating the unpredictable stream token, allow an ERROR ACK from a routing hop while continuing to require target origin for
  ordinary progress ACKs—or explicitly validate the rejector against the route. Add a three-cell transit-rejection test; the existing test only covers a
  direct sender-to-rejector connection.

  Otherwise, I found no happy-path progress-aware streaming regression. progress_wait_cb remains intact, reliable delivery is propagated correctly, the
  SESSION_READY handshake closes the authentication race, and the trainer-lifecycle/shutdown changes look consistent. All PR CI checks and 547 targeted
  local tests passed. So my conclusion is: normal homogeneous deployments should stream correctly, but routed rejection handling still has this P2 gap.

YuanTingHsieh:
Addressed the P2 routed filter-rejection ACK issue in signed commit ea3d513eab23f73b45898c141d2b21ef910e5a7e.

TxTask now validates the unpredictable stream token before applying origin policy. Ordinary progress ACKs remain restricted to the final target, while only a token-correlated ERROR frame with an error detail may terminate the stream from a non-target routing cell. This lets a server or relay that actually observed and rejected the transit DATA frame fail the sender promptly without reopening wrong-origin progress updates or token-forgery termination.

The unit regressions prove that an invalid-token routing error and a correct-token progress ACK from the wrong origin are ignored, while a correlated routing-hop ERROR is accepted. A real three-process Cell test sends site-a through server toward site-b, has the server reject the transit DATA frame, and verifies site-a receives an immediate StreamError from server.

Validation: 131 affected streaming/CoreCell tests passed; 663 broader available F3 tests passed (the local lightweight environment excluded only unrelated Torch and OmegaConf benchmark modules); and the full project style gate passed.

pcnudde:
I found one lifecycle ownership issue on the current head (`9162f710`) that I think should be fixed before merge.

`Cell._send_one_request()` returns when its timeout or abort condition fires and removes the request waiter, but it does not terminalize the underlying `StreamFuture`. This PR now opts `TASK_READY`, `RESULT_READY`, heartbeat, HELLO/SESSION_READY, SHUTDOWN, and DownloadService requests into `reliable=True`, so their `TxTask` can keep retrying after the higher-level request has returned.

I reproduced this on the exact PR head with a disconnected reliable `RESULT_READY` request:

```
caller_return_code: timeout
live_tx_tasks_after_return: 1
wire_sends_at_return: 1
wire_sends_five_seconds_later: 2
request_waiters_after_return: 0
```

That creates ambiguous late delivery: for example, a `TASK_READY` retry can remain live after the backend has handled abort, or a `RESULT_READY`/download request can be delivered after its caller has already observed failure and released lifecycle state.

The obvious `future.cancel()` repair currently deadlocks for a real `TxTask`: `StreamFuture.cancel()` holds the future lock while `TxTask.cancel()` calls back into `StreamFuture.set_exception()` and waits for that same lock. I reproduced that cancellation call remaining blocked.

Suggested direction:

1. Fix the `StreamFuture.cancel()` lock ordering so task cancellation occurs outside the future lock.
2. On every `Cell._send_one_request()` return path before send completion, stop/join the transmit task before removing the waiter and returning.
3. Add a regression that forces a missing ACK plus timeout/abort and asserts both that the `TxTask` is gone and that no later wire send occurs.

The lifecycle invariant I would enforce is: `Cell request returned => transmit task terminal`. The new `SESSION_READY` state and stream-token correlation otherwise look sound, and 456 focused tests including the real routed socket cases passed locally.

nvflare/fuel/f3/streaming/byte_streamer.py
```diff
@@ -230,6 +232,7 @@ def __init__(
         self.cell = cell
         self.chunk_size = chunk_size
         self.sid = gen_stream_id()
+        self.stream_token = secrets.token_hex(16)
```
nvidianz: What's the token for? sid is already an unique identifier. Carrying secret in clear header doesn't make it any more secure. If we really want some security here, should put a hmac(sid+secret) as token.

YuanTingHsieh: good suggestion, removed all unrelated codes

nvflare/client/cell/api.py
```diff
@@ -330,8 +330,38 @@ def _hello(self) -> None:
...
         self._note_cj_activity()
 
+    def _confirm_session_ready(self, deadline: float) -> None:
+        """Tell the CJ that HELLO_ACCEPTED processing and auth-filter setup are complete."""
+        attempt = 0
+        while True:
+            attempt += 1
+            remaining = deadline - time.monotonic()
+            if remaining <= 0:
+                raise TrainerSessionError(f"no SESSION_READY confirmation from the CJ after {_HELLO_TIMEOUT}s")
```
pcnudde: `_confirm_session_ready` reuses the same 30s `_HELLO_TIMEOUT` deadline the HELLO retry loop may have already consumed. A slow listener connect where HELLO succeeds at ~29.9s leaves `remaining <= 0` here, so this raises without making a single SESSION_READY attempt — even though the CJ just accepted the session — and the message falsely implies 30s of SESSION_READY retries. Pre-PR that launch succeeded; the CJ side is then left half-open (session_id set, ready never set) until launch_timeout.

Suggest guaranteeing at least one attempt (or a small fresh budget) after HELLO_ACCEPTED, and including the actual elapsed/attempt info in the error.

nvflare/app_common/executors/client_api/external_process_backend.py
```diff
@@ -1403,9 +1415,52 @@ def _handle_hello(self, request):
...
+
+    def _delegated_auth_headers_match(self, request) -> bool:
+        expected = (
+            (CellMessageAuthHeaderKey.CLIENT_NAME, self._site_name),
+            (CellMessageAuthHeaderKey.TOKEN, self._site_auth_token),
+            (CellMessageAuthHeaderKey.TOKEN_SIGNATURE, self._site_auth_token_signature),
+        )
+        for key, value in expected:
+            actual = request.get_header(key)
+            if not isinstance(actual, str) or not isinstance(value, str) or not secrets.compare_digest(actual, value):
```
pcnudde: `secrets.compare_digest` raises `TypeError` for str arguments containing non-ASCII characters (`isinstance(actual, str)` doesn't exclude those). A broken or malicious same-host process sending SESSION_READY with a non-ASCII auth header value crashes the handler; cellnet converts it to PROCESS_EXCEPTION, so the trainer retries for 30s instead of getting the intended clean `Topic.ERROR` rejection.

`_handle_hello` already guards exactly this for the launch-token proof ("Validate type before compare_digest so non-ASCII or forged proofs reject cleanly") by comparing `.encode("utf-8")` values — same fix applies here.

nvflare/app_common/executors/client_api/external_process_backend.py
```diff
@@ -1403,9 +1415,52 @@ def _handle_hello(self, request):
...
+                reason = None
+                trainer.touch_peer_activity()
+                if not trainer.ready.is_set():
+                    trainer.ready.set()
+                    self.logger.info(f"trainer readiness confirmed from {origin} (session_id={trainer.session_id})")
+
+        if reason:
+            self.logger.warning(f"rejecting SESSION_READY: {reason}")
+            return self._protocol_reply(Topic.ERROR, **{MsgKey.REASON: reason})
+        return self._protocol_reply(Topic.SESSION_READY, **{MsgKey.SESSION_ID: trainer.session_id})
```
pcnudde: This reads `trainer.session_id` after `_launch_lock` is released, and `_cleanup_trainer` nulls `trainer.session_id` without holding the lock (its callers in `_stop_trainer` and the reaper thread don't hold it either). A SESSION_READY racing teardown can pass all in-lock checks, then return a success reply with `SESSION_ID: None`; the trainer's session-id match fails, it retries, hits "no accepted trainer session", and exits with `TrainerSessionError` instead of an orderly shutdown. The lock-free null also races the in-lock comparisons above.

Cheap fix: capture `session_id` (or build the reply) inside the `with` block, and null `session_id` under `_launch_lock` in cleanup.

pcnudde (commented):
Three correctness issues found in review; inline comments below.

YuanTingHsieh:
Addressed the three latest readiness-handshake review findings in signed commit `7cc1c145efd1491c44c702d7a32cdf11795c6d0d`:

- `SESSION_READY` now receives a fresh 30-second retry budget after `HELLO_ACCEPTED`; diagnostics report the actual send-attempt count and elapsed time. Added slow-HELLO and timeout-diagnostic regressions.
- Delegated credential values are type-checked and compared as UTF-8 bytes, so non-ASCII input returns a clean `Topic.ERROR` instead of raising `PROCESS_EXCEPTION`. Added coverage for client name, token, and signature.
- Accepted session IDs are snapshotted under `_launch_lock`, and cleanup clears the session ID under the same lock. Added regressions for both the readiness reply snapshot and cleanup serialization.

Exact-head validation: 235 affected tests passed, 2 skipped; Black, isort, flake8, and agent-skill lint all passed.

pcnudde (approved):

PR #5109, opened 2026-08-12
Harden internal CellNet trust boundaries

nvidianz:
## Summary

- bind clear internal CellNet listeners to loopback and require mTLS for non-loopback Docker, Kubernetes, and Slurm internal hops
- validate privileged CP administrative and server-job parent commands against both their logical headers and authenticated physical route
- avoid adding reusable bearer credentials to Challenge, Register, and CellNet Bye pre-auth replies
- restrict `configure_job_log` to log levels and built-in modes instead of accepting arbitrary `dictConfig` objects

## Root cause

Internal TCP listeners could silently bind to all interfaces, while distributed launchers used clear transport for cross-host internal links. Privileged receivers trusted caller-selected logical headers without binding them to the physical connection. Pre-auth replies could also receive reusable server bearer headers, and `configure_job_log` accepted arbitrary logging configurations.

## Validation

- focused CellNet security regression suite: 32 passed
- broader affected unit-test suite: 944 passed
- client attach end-to-end scenarios validated; one timing-sensitive gRPC observation assertion passed on isolated rerun
- scoped Black, isort, Flake8, and `git diff --check`: passed
- repository-wide `./runtest.sh -s`: Black reached three pre-existing vendored `node_modules/flatted.py` files that would be reformatted; no files in this PR were implicated

## Tracking

- [FLARE-3123](https://jirasw.nvidia.com/browse/FLARE-3123)

### DIFF
--- a/nvflare/app_opt/job_launcher/slurm/launcher.py
+++ b/nvflare/app_opt/job_launcher/slurm/launcher.py
@@ -455,8 +455,8 @@ class SlurmJobLauncher(JobLauncherSpec):
         if not isinstance(connection_entry, (tuple, list)) or len(connection_entry) != 2:
             raise SlurmLauncherError(f"malformed {JobProcessArgs.PARENT_CONN_SEC} in JOB_PROCESS_ARGS")
         process_connection_security = _require_string(connection_entry[1], f"{JobProcessArgs.PARENT_CONN_SEC} value")
-        if process_connection_security != "clear":
-            raise SlurmLauncherError("Slurm job launch requires clear parent connection security")
+        if process_connection_security not in ("clear", "mtls"):
+            raise SlurmLauncherError("Slurm job launch requires clear or mTLS parent connection security")
         job_args = _rewrite_parent_url(
             raw_job_args,
             parent_host=self.config.parent_host,
--- a/nvflare/app_opt/job_launcher/slurm/launcher.py
+++ b/nvflare/app_opt/job_launcher/slurm/launcher.py
@@ -15,6 +15,7 @@
 
 from __future__ import annotations
 
+import ipaddress
 import os
 import shlex
 from abc import abstractmethod
@@ -221,15 +222,21 @@ def _rewrite_parent_url(job_args: dict, parent_host: Optional[str], internal_por
         host = parsed.hostname
     except ValueError as e:
         raise SlurmLauncherError("malformed parent URL in JOB_PROCESS_ARGS") from e
-    if parsed.scheme != "tcp" or port != internal_port or not host:
+    if parsed.scheme not in ("tcp", "stcp") or port != internal_port or not host:
         raise SlurmLauncherError(
-            f"parent URL must use {SHARED_FILE_SCHEME} or tcp with configured internal_port "
+            f"parent URL must use {SHARED_FILE_SCHEME}, tcp, or stcp with configured internal_port "
             f"{internal_port}, got {raw_url!r}"
         )
     host = _resolve_parent_host(parent_host)
-    rendered_host = host if host.startswith("[") and host.endswith("]") else f"[{host}]" if ":" in host else host
-    netloc = f"{rendered_host}:{internal_port}"
-    rewritten = urlunsplit(("tcp", netloc, parsed.path, parsed.query, parsed.fragment))
+    unbracketed_host = host[1:-1] if host.startswith("[") and host.endswith("]") else host
+    try:
+        address = ipaddress.ip_address(unbracketed_host)
+    except ValueError:
+        address = None
+    if isinstance(address, ipaddress.IPv6Address) or ":" in host:
+        raise SlurmLauncherError("IPv6 parent_host is not supported by the F3 TCP transport")
+    netloc = f"{host}:{internal_port}"
+    rewritten = urlunsplit((parsed.scheme, netloc, parsed.path, parsed.query, parsed.fragment))
     copied[JobProcessArgs.PARENT_URL] = (flag, rewritten)
     return copied
 
@@ -462,6 +469,14 @@ class SlurmJobLauncher(JobLauncherSpec):
             parent_host=self.config.parent_host,
             internal_port=self.config.internal_port,
         )
+        parent_scheme = urlsplit(str(job_args[JobProcessArgs.PARENT_URL][1])).scheme
+        if parent_scheme != SHARED_FILE_SCHEME:
+            expected_scheme = "stcp" if process_connection_security == "mtls" else "tcp"
+            if parent_scheme != expected_scheme:
+                raise SlurmLauncherError(
+                    f"parent URL scheme '{parent_scheme}' does not match "
+                    f"parent connection security '{process_connection_security}' (expected '{expected_scheme}')"
+                )
 
         study = job_meta.get(JobMetaKey.STUDY.value)
         if study is not None:
--- a/nvflare/fuel/f3/cellnet/connector_manager.py
+++ b/nvflare/fuel/f3/cellnet/connector_manager.py
@@ -99,7 +99,7 @@ class ConnectorManager:
         # set up default drivers
         self.int_scheme = comm_configurator.get_internal_connection_scheme(_Defaults.SCHEME_FOR_INTERNAL_CONNECTIONS)
         self.int_resources = {
-            _KEY_HOST: "localhost",
+            _KEY_HOST: "127.0.0.1",
         }
         self.adhoc_allowed = comm_configurator.allow_adhoc_connections(_Defaults.ALLOW_ADHOC_CONNECTIONS)
         self.adhoc_scheme = comm_configurator.get_adhoc_connection_scheme(_Defaults.SCHEME_FOR_ADHOC_CONNECTIONS)
@@ -131,7 +131,7 @@ class ConnectorManager:
 
     def _validate_internal_listener_security(self):
         conn_sec = self.int_resources.get(DriverParams.CONNECTION_SECURITY, ConnectionSecurity.CLEAR)
-        host = self.int_resources.get(DriverParams.LISTEN_HOST, self.int_resources.get(_KEY_HOST, "localhost"))
+        host = self.int_resources.get(DriverParams.LISTEN_HOST, self.int_resources.get(_KEY_HOST, "127.0.0.1"))
         if _is_loopback_host(host):
             return
         if conn_sec == ConnectionSecurity.MTLS:
--- a/nvflare/fuel/f3/cellnet/connector_manager.py
+++ b/nvflare/fuel/f3/cellnet/connector_manager.py
@@ -34,6 +34,33 @@ _KEY_HOST = "host"
 _KEY_PORTS = "ports"
 
 
+def _is_loopback_host(host) -> bool:
+    if not isinstance(host, str):
+        return False
+
+    host = host.strip()
+    if host.rstrip(".").lower() == "localhost":
+        return True
+
+    # Accept the bracketed form commonly used for IPv6 URL hosts as well as
+    # bare address literals used by listener resources.
+    if host.startswith("[") and host.endswith("]"):
+        host = host[1:-1]
+
+    try:
+        address = ipaddress.ip_address(host)
+    except ValueError:
+        return False
+
+    if address.is_loopback:
+        return True
+
+    # On older supported Python versions, IPv4-mapped IPv6 addresses do not
+    # inherit the mapped IPv4 address's is_loopback value.
+    mapped_address = getattr(address, "ipv4_mapped", None)
+    return mapped_address is not None and mapped_address.is_loopback
+
+
 class _Defaults:
 
     ALLOW_ADHOC_CONNECTIONS = False
@@ -105,13 +132,8 @@ class ConnectorManager:
     def _validate_internal_listener_security(self):
         conn_sec = self.int_resources.get(DriverParams.CONNECTION_SECURITY, ConnectionSecurity.CLEAR)
         host = self.int_resources.get(DriverParams.LISTEN_HOST, self.int_resources.get(_KEY_HOST, "localhost"))
-        if host == "localhost":
+        if _is_loopback_host(host):
             return
-        try:
-            if ipaddress.ip_address(host).is_loopback:
-                return
-        except ValueError:
-            pass
         if conn_sec == ConnectionSecurity.MTLS:
             return
         raise ConfigError(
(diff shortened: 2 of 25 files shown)


nvflare/tool/deploy/slurm_deploy.py
```diff
@@ -211,8 +211,9 @@ def _patch_comm_config(kit_dir: Path, port: int) -> None:
     resources.update(
         {
             "host": "0.0.0.0",
+            "listen_host": "0.0.0.0",
             "port": port,
-            "connection_security": "clear",
+            "connection_security": "mtls",
```
YuanTingHsieh, line 216: **[P1] Slurm rejects the generated mTLS configuration.** This value is propagated as `PARENT_CONN_SEC`, but the unchanged Slurm launcher rejects every value other than `clear` in `nvflare/app_opt/job_launcher/slurm/launcher.py:458-459`. Consequently every TCP Slurm job prepared by this path fails before `sbatch`. Please update the launcher and an integration-level launch-plan test to support mTLS before emitting this setting.

nvflare/fuel/f3/drivers/net_utils.py
```diff
@@ -339,13 +343,23 @@ def enhance_credential_info(params: dict):
         server_cert_path = os.path.join(cred_folder, SSL_SERVER_CERT)
         if os.path.exists(server_cert_path):
             params[DriverParams.SERVER_CERT.value] = server_cert_path
+        elif client_cert_path:
+            # Internal CellNet peers can act as both connector and listener.
+            params[DriverParams.SERVER_CERT.value] = client_cert_path
```
YuanTingHsieh: **[P1] A role-limited client certificate cannot be reused as the listener certificate.** The distributed certificate workflow issues client certificates with only the `CLIENT_AUTH` EKU (`nvflare/tool/cert/cert_commands.py:1205-1210`), while a connecting peer validates the listener certificate for `SERVER_AUTH`. Since deploy accepts client kits containing only `client.crt`/`client.key`, Docker/Kubernetes parent-listener handshakes fail with an unsuitable-certificate-purpose error. Please require a real server certificate or provision an explicit dual-use internal certificate.

nvflare/fuel/f3/drivers/net_utils.py
```diff
@@ -339,13 +343,23 @@ def enhance_credential_info(params: dict):
...
     server_key_path = params.get(DriverParams.SERVER_KEY.value)
     if not server_key_path:
         # see whether the server key file exists
         server_key_path = os.path.join(cred_folder, SSL_SERVER_PRIVATE_KEY)
         if os.path.exists(server_key_path):
             params[DriverParams.SERVER_KEY.value] = server_key_path
+        elif client_key_path:
+            params[DriverParams.SERVER_KEY.value] = client_key_path
+
+    if not client_cert_path and server_cert_path:
```
YuanTingHsieh: **[P1] Server-only credentials never gain client fields here.** When `CLIENT_CERT` or `CLIENT_KEY` is absent, lines 329 and 336 first assign a nonempty candidate pathname even when that file does not exist. These conditions are therefore false, so standard server credentials containing only `server.crt`/`server.key` remain without client credentials. A server job's active mTLS connector then sends no client certificate and the parent's `CERT_REQUIRED` handshake fails. Please track candidate existence separately and supply a client-auth-capable credential.

YuanTingHsieh (commented):
I found three runtime blockers in the internal mTLS rollout. Details are inline.

nvflare/fuel/f3/drivers/net_utils.py
nvidianz: Fixed in 28898320f. Distributed participant certificates are now explicitly dual-use, with both CLIENT_AUTH and SERVER_AUTH EKUs, so the same identity pair is valid for internal CellNet connector and listener roles. Admin-role certificates remain CLIENT_AUTH-only. The certificate-signing regression test covers both client and server participant certificate types.

nvflare/tool/deploy/slurm_deploy.py
nvidianz: Fixed in 28898320f. The Slurm launcher now accepts clear or mtls parent connection security, while continuing to reject unsupported values such as tls. The launch-plan regression test sets PARENT_CONN_SEC to mtls and verifies that --parent_conn_sec mtls is propagated into the child module arguments.

nvflare/fuel/f3/drivers/net_utils.py
nvidianz: Fixed in 28898320f. Candidate paths are now assigned only after existence checks succeed. A complete server cert/key pair fills the client role, and a complete client pair fills the server role; incomplete pairs are never mixed. Regression tests cover server-only fallback and mismatched incomplete pairs.

nvflare/app_opt/job_launcher/slurm/launcher.py
```diff
@@ -455,8 +455,8 @@ def _build_launch_plan(self, job_meta: dict, fl_ctx: FLContext) -> LaunchPlan:
         if not isinstance(connection_entry, (tuple, list)) or len(connection_entry) != 2:
             raise SlurmLauncherError(f"malformed {JobProcessArgs.PARENT_CONN_SEC} in JOB_PROCESS_ARGS")
         process_connection_security = _require_string(connection_entry[1], f"{JobProcessArgs.PARENT_CONN_SEC} value")
-        if process_connection_security != "clear":
-            raise SlurmLauncherError("Slurm job launch requires clear parent connection security")
+        if process_connection_security not in ("clear", "mtls"):
```
nvidianz, line 465: One-way TLS is intentionally not accepted for remote internal CellNet links because it authenticates the listener but not the connecting job peer, so it does not close this trust-boundary issue. For Slurm, the network parent transport is TCP with mTLS; shared-file remains clear because it has no TLS layer. HTTP is not a supported Slurm parent transport: the launcher accepts only tcp and shared-file parent URLs.

chesterxgchen (commented):
- P2 — IPv6 internal listeners cannot start.
    net_utils.py:296 (https://github.com/NVIDIA/NVFlare/blob/28898320f/nvflare/fuel/f3/drivers/net_utils.py#L296) builds URLs by interpolating hosts
    as tcp://{host}:{port}. For the IPv6 loopback values that this PR explicitly accepts—::1 and [::1]—this produces tcp://::1:9000 or
    tcp://[::1]:9000. The existing parser splits the netloc on :, yielding an empty/invalid host and port, so the listener/connector cannot
    initialize. The new loopback tests validate only the security check, not URL construction or an actual connection. Format IPv6 literals with
    brackets when building URLs and update parse_url() to use urlparse().hostname / .port (or otherwise parse bracketed IPv6 correctly); add a
    round-trip test for ::1.

nvflare/fuel/f3/cellnet/connector_manager.py
```diff
@@ -33,6 +34,33 @@
...
+    # bare address literals used by listener resources.
+    if host.startswith("[") and host.endswith("]"):
+        host = host[1:-1]
+
+    try:
+        address = ipaddress.ip_address(host)
+    except ValueError:
+        return False
+
+    if address.is_loopback:
```
YuanTingHsieh: **[P1] Do not allow clear IPv6 loopback listeners until CellNet can parse their URLs safely.** `_is_loopback_host()` now accepts values such as `::1`, `[::1]`, and `0:0:0:0:0:0:0:1`, after which `get_tcp_urls()` renders them directly as `tcp://<host>:<port>`. However, `net_utils.parse_url()` parses the authority with `parsed_url.netloc.split(":")`. The expanded form is interpreted as host `""` and port `0`; `Communicator.start_listener()` then overwrites the original resources with those parsed fields, and `TcpStreamServer` binds `("", 0)`—an IPv4 wildcard ephemeral listener using clear transport. Other IPv6 forms fail startup instead. Please either make URL formatting/parsing IPv6-safe (including brackets) throughout the TCP drivers, or reject IPv6 here for now, and add a regression test through actual listener creation rather than only `ConnectorManager` construction.

YuanTingHsieh (commented):
One remaining runtime trust-boundary issue is inline.

nvflare/fuel/f3/cellnet/connector_manager.py
nvidianz: Fixed in 1f719a48f using the safe interim option. Literal IPv6 hosts, including native, expanded, bracketed, and IPv4-mapped forms, are rejected by get_tcp_urls before URL generation or driver binding. _is_loopback_host now approves only IPv4 loopback literals (plus localhost), so clear IPv6 loopback cannot bypass the mTLS boundary. The new Communicator.start_listener regression exercises the actual listener-creation path and verifies that each IPv6 form raises before any wildcard bind; 33 focused tests pass.

nvidianz:
@chesterxgchen Addressed your IPv6 listener review in commit `1f719a48f`. I chose the safe interim option of rejecting literal IPv6 hosts rather than implementing only URL bracketing/parsing, because the raw F3 TCP driver still creates IPv4 sockets and partial URL support would remain unsafe. Native, expanded, bracketed, and IPv4-mapped IPv6 literals are now rejected by `get_tcp_urls()` before URL generation or driver binding, and `_is_loopback_host()` no longer approves IPv6 as a clear loopback listener. The regression test goes through `Communicator.start_listener()` and verifies that these forms raise before they can degrade into a wildcard IPv4 bind. The default same-host listener is now deterministic IPv4 loopback (`127.0.0.1`). All 33 focused listener/security tests pass.

chesterxgchen (commented):
### [P1] Loopback does not make the parent route authenticated

  ConnectorManager permits clear transport whenever the listener is on loopback
  (https://github.com/NVIDIA/NVFlare/blob/8b120ad8988dd85378ad865f0eda2e33f6672f39/nvflare/fuel/f3/cellnet/connector_manager.py#L129-L135), including secure-mode local launches.

  On a clear connection, CellNet accepts the endpoint name supplied during the peer handshake; certificate-to-endpoint binding only happens for mTLS connections (connection handling
  (https://github.com/NVIDIA/NVFlare/blob/8b120ad8988dd85378ad865f0eda2e33f6672f39/nvflare/fuel/f3/sfm/conn_manager.py#L420-L441)). Consequently, any process on the same host can connect
  to the client parent’s loopback listener and claim the endpoint name server.

  That forged endpoint then satisfies the new administrative check because it compares:

  - Caller-controlled ORIGIN == "server"
  - Caller-controlled destination
  - The unauthenticated handshake endpoint name server

  See FedAdminAgent._dispatch_request (https://github.com/NVIDIA/NVFlare/blob/8b120ad8988dd85378ad865f0eda2e33f6672f39/nvflare/private/fed/client/admin.py#L108-L134).

  This exposes privileged processors including shutdown, restart, deploy, shell/file inspection, and job control. A forged envelope can additionally use REQUIRE_AUTHZ=false; choosing an
  internal command name such as _cert_login also bypasses SiteSecurity because the command header is only checked for presence, not bound to the dispatched topic.

  Loopback prevents remote connections but does not isolate other users or processes on a shared host.

  Recommended fix:

  - In secure mode, default internal listeners to mTLS even on loopback.
  - Permit clear loopback only for explicitly insecure deployments, or use an authenticated Unix socket/process-specific secret.
  - Add an integration test proving that a clear peer claiming server cannot invoke CLIENT_MAIN administrative processors.

  The distributed mTLS configuration, pre-auth bearer suppression, logging-config restriction, IPv6 rejection, and dual-use credential handling otherwise look sound. All current CI checks
  pass, but they do not cover the clear-loopback impersonation case.

pcnudde:
For this PR we need to do some end to end testing in various modes. (k8s, docker) just to make sure things still work.

pcnudde (commented):
Additional exact-head review at `8b120ad8988dd85378ad865f0eda2e33f6672f39` found the two inline blockers below and one further deployment defect that cannot be anchored inline because the relevant line is outside the PR diff:

**[P2] Reject unsupported IPv6 Slurm parent hosts.** `nvflare/app_opt/job_launcher/slurm/launcher.py:229-233` accepts a `parent_host` containing `:` and emits a bracketed URL such as `tcp://[2001:db8::2]:8102`. F3's `parse_url()` splits the authority on `:`, producing host `"[2001"` and port `"db8"`, and `TcpDriver.connect()` uses `AF_INET` in any case. I reproduced that exact parse result. Please reject IPv6 `parent_host` values until parsing and transport support are implemented end-to-end, and extend the regression beyond URL formatting into F3 parsing/connector setup.

Focused validation remains green (`206 passed`), but those tests do not exercise these production-shaped paths.

nvflare/private/fed/server/fed_server.py
```diff
@@ -411,6 +411,20 @@ def _add_auth_headers(self, message: Message):
...
         topic = message.get_header(MessageHeaderKey.TOPIC)
+        # Endpoints that are intentionally reachable before client authentication
+        # (see authenticator.validate_auth_headers, which exempts these same
+        # channel/topic pairs from the incoming auth check). Their replies must
+        # NOT disclose the server's reusable bearer token and signature, or an
+        # uncredentialed peer could elicit a reply, harvest the headers, and
+        # replay them to authenticate as the server. This must stay in sync with
+        # the incoming exemptions: Challenge/Register on SERVER_MAIN and the
+        # direct-neighbor CELLNET/Bye teardown ack.
+        pre_auth_reply = (
```
pcnudde, line 422: **[P0] Bind the reusable server bearer to a server-owned ingress.** This suppression covers only pre-authentication replies, while normal replies still deliberately expose the static server token/signature. A registered client can harvest that tuple, replay it with `ORIGIN=server`, and send a parent-only command to `server.<job>`. At the root, `_resolve_client_fqcn_for_auth()` returns `None` for the server token, so `validate_auth_headers()` skips origin binding; after forwarding, `ServerCommandAgent` sees the root server as the immediate endpoint and accepts the command. A focused reproduction showed `root_filter_accepted=True`, `job_command_rc=ok`, and the mocked `ABORT` processor executed. Please reject server credentials arriving from client-family physical endpoints (or replace the replayable bearer with message/connection-bound authentication) and add an end-to-end client-to-server-job replay test.

nvflare/app_opt/job_launcher/slurm/launcher.py
pcnudde, line 465: **[P1] The Slurm mTLS fix still rejects the production URL.** Accepting `PARENT_CONN_SEC=mtls` here is not sufficient: an mTLS TCP listener is advertised as `stcp://...`, but the immediately following `_rewrite_parent_url()` accepts only `parsed.scheme == "tcp"` and hardcodes `tcp` in its output. Directly passing `stcp://0.0.0.0:8102` produces `SlurmLauncherError: parent URL must use shared-file or tcp...`. The new test changes only the connection-security argument while leaving its fixture URL as `tcp://`, so it misses the real path. Please accept and preserve `stcp` and make the regression pair `mtls` with a production-shaped `stcp://` parent URL.

nvidianz:
Closing this combined PR because the changes have been split into focused PRs:

- #5120 — remote job log configuration controls
- #5122 — SimEnv loopback listener binding

Superseded by #5120 and #5122.

PR #5112, opened 2026-08-13
FLARE-3123/3125: harden CellNet job command trust boundaries

nvidianz:
## Summary

- harden internal CellNet listener defaults and require mTLS for non-loopback job-process hops
- authorize server-job and client-job administrative command senders against the expected control-plane route
- prevent pre-authentication replies from carrying reusable bearer material
- restrict runtime log reconfiguration to safe log levels and built-in modes instead of executable `dictConfig` payloads
- add focused security regression coverage, release notes, and a checklist for remaining defense-in-depth work

## Root cause

Internal job command receivers relied on federation enrollment and routing without consistently enforcing that the sender was the job's trusted parent control plane. An enrolled site could therefore address administrative handlers on another job cell. The log-configuration path also accepted attacker-controlled Python logging dictionaries, creating a code-execution sink without requiring BYOC permission.

## Impact

Enrolled sites can no longer use these paths to control peer/server jobs or execute code through log reconfiguration. Legitimate commands relayed by the root server or a client job's own parent continue to dispatch normally.

## Validation

- `python -m pytest -q tests/unit_test/fuel/f3/cellnet/connector_manager_security_test.py tests/unit_test/fuel/f3/drivers/net_utils_test.py tests/unit_test/private/fed/security_boundary_test.py tests/unit_test/tool/deploy/deploy_commands_test.py tests/unit_test/tool/deploy/slurm_deploy_test.py` (`151 passed`)
- scoped `black --check`, `isort --check-only`, and `flake8` on all 18 changed Python files

The repository-wide `./runtest.sh -s` reaches Black but is currently blocked by three pre-existing Python files under an example `node_modules` tree; the complete changed-file style scope passes.


nvflare/private/fed/client/admin.py
```diff
@@ -103,10 +105,54 @@ def _dispatch_request(
...
+            self.cell.logger.warning(
+                f"rejected {request_type} from {origin} via "
+                f"{getattr(peer_endpoint, 'name', '<unknown>')} to {destination}"
+            )
+            return new_cell_message(
+                {MessageHeaderKey.RETURN_CODE: ReturnCode.UNAUTHENTICATED},
+                error_reply(f"unauthenticated {request_type}"),
+            )
+
+        if not is_job_status and (outer_topic != topic or not cmd or authz_flag not in ("true", "false")):
```
YuanTingHsieh, line 151: **[P1] Preserve the established server→client-parent admin envelope.** `send_requests()` sends the outer CellNet topic as the literal `"admin"`, while `req.topic` is the actual operation (`train.deploy`, `scheduler.check_resource`, etc.), so `outer_topic != topic` is true for every normal admin request. The `not cmd` / `authz_flag` clauses independently reject internal senders such as start, cancel-resource, abort, and delete-run that do not populate those headers. These requests return `INVALID_REQUEST` before dispatch, preventing jobs from being scheduled or controlled. Please either update all senders atomically or validate the existing wire format, with end-to-end deploy/start/abort coverage.

nvflare/private/fed/server/server_command_agent.py
```diff
@@ -68,6 +70,21 @@ def execute_command(self, request: CellMessage) -> CellMessage:
...
         command = ServerCommands.get_command(command_name)
         if command:
+            if command_name not in ServerCommands.client_request_commands_names:
+                origin = request.get_header(MessageHeaderKey.ORIGIN)
+                destination = request.get_header(MessageHeaderKey.DESTINATION)
+                peer_endpoint = request.get_prop(MessagePropKey.ENDPOINT)
+                if (
+                    origin != FQCN.ROOT_SERVER
+                    or destination != self.cell.get_fqcn()
+                    or not peer_endpoint
```
YuanTingHsieh, line 80: **[P1] `ENDPOINT` is unavailable on the streamed `SERVER_COMMAND` delivery path.** `SERVER_COMMAND` is not in `CHANNELS_TO_EXCLUDE`, so production requests use `send_blob`; `Adapter.call()` reconstructs the callback request as `Message(future.headers, result)` and does not preserve message props. Consequently `request.get_prop(MessagePropKey.ENDPOINT)` is `None` here and every parent command other than the client request commands is rejected. In-process simulator delivery can mask this. Please authenticate using metadata preserved by the stream layer (or propagate a verified endpoint identity) and add an interprocess regression test.

nvflare/fuel/f3/drivers/net_utils.py
```diff
@@ -339,13 +343,23 @@ def enhance_credential_info(params: dict):
...
     server_key_path = params.get(DriverParams.SERVER_KEY.value)
     if not server_key_path:
         # see whether the server key file exists
         server_key_path = os.path.join(cred_folder, SSL_SERVER_PRIVATE_KEY)
         if os.path.exists(server_key_path):
             params[DriverParams.SERVER_KEY.value] = server_key_path
+        elif client_key_path:
+            params[DriverParams.SERVER_KEY.value] = client_key_path
+
+    if not client_cert_path and server_cert_path:
```
YuanTingHsieh, line 359: **[P1] This server→client credential fallback is dead when the client files are absent.** Earlier, `client_cert_path` and `client_key_path` are assigned non-empty `os.path.join(...)` strings even when `os.path.exists()` is false, so these `if not ...` conditions never run. A server-only kit therefore never receives `CLIENT_CERT` / `CLIENT_KEY`, and its server-job connector cannot authenticate to the newly mandatory mTLS internal listener. Keep the variables unset when the candidate files do not exist (or test existence explicitly), and cover the server-only direction.

nvflare/fuel/f3/drivers/net_utils.py
```diff
@@ -339,13 +343,23 @@ def enhance_credential_info(params: dict):
         server_cert_path = os.path.join(cred_folder, SSL_SERVER_CERT)
         if os.path.exists(server_cert_path):
             params[DriverParams.SERVER_CERT.value] = server_cert_path
+        elif client_cert_path:
```
YuanTingHsieh, line 346: **[P1] Do not blindly reuse a role-scoped certificate for the opposite TLS role.** The distributed `nvflare cert` flow issues client certificates with `CLIENT_AUTH` EKU and server certificates with `SERVER_AUTH` EKU. Presenting `client.crt` from a listener (and the reverse fallback below from an active connector) fails certificate-purpose validation. Internal peers need separate credentials for both directions or a deliberately provisioned dual-purpose certificate.

FLARE-3125-remaining-tasks.md
```diff
@@ -0,0 +1,74 @@
...
+- **FLARE-3125 SJ route** enrolled-site -> `server.<job_id>/configure_job_log`
+  is rejected: `configure_job_log` is a parent command, gated to `origin==ROOT_SERVER`;
+  `ORIGIN` is framework-stamped and unspoofable.
+- **FLARE-3125 CJ route** enrolled-site -> `<victim>.<job_id>/CLIENT_COMMAND/*`
+  is rejected: `CommandAgent.execute_command` now allows only `origin in {ROOT_SERVER,
+  own-parent CP}` addressed to this cell. (commit "authorize CLIENT_COMMAND sender")
+
+## Remaining tasks
+
+### 1. [P0] FLARE-3125 fix 4 — isolate job processes from startup keys
```
YuanTingHsieh, line 28: **[P1] Please do not ship this internal security checklist in the public repository.** It enumerates live, unresolved P0/P1 attack surfaces with implementation pointers, including startup-key exfiltration and the fail-open identity binding. Keep the working checklist and exploit-relevant details in Jira/security tracking; the user-facing release note already documents the completed behavior.

YuanTingHsieh (commented):
Blocking issues are called out inline. One additional architectural concern cannot be attached to a changed line: these new gates treat the framework-stamped `ORIGIN` as authenticated identity, but `validate_auth_headers()` only enforces origin binding when `client_fqcn_resolver()` returns a value; `None` currently fails open, and client cells do not install an equivalent incoming authentication filter. Please enforce origin↔authenticated-connection binding below the individual command receivers, or narrow the security guarantee. Stamping `ORIGIN` prevents an ordinary sender from overriding its own header, but is not by itself proof of peer identity.

pcnudde:
Additional findings on current head `d9a7ed929f26c4e55d27824e5207d604ebd2040a`:

1. **[P1] Generated Slurm deployments cannot launch TCP jobs.** `nvflare/tool/deploy/slurm_deploy.py:211-217` writes `connection_security: mtls`, which makes the parent listener URL `stcp://...`. However, `nvflare/app_opt/job_launcher/slurm/launcher.py:454-459` rejects every connection-security value except `clear`, and `:224-232` only accepts and emits `tcp://`. A deployment produced by this PR therefore fails before `sbatch`. Please make the launcher support and preserve the secure scheme and add a regression test that feeds prepared Slurm configuration into the launcher path.

2. **[P1] Docker downgrades the new mTLS parent URL to plaintext.** `nvflare/tool/deploy/docker_deploy.py:129-134` enables mTLS, but `nvflare/app_opt/job_launcher/docker_launcher.py:556-577` discards the original `stcp` scheme and hardcodes `tcp://<site>:<port>`. `get_ssl_context()` enables TLS from the secure scheme/`secure` flag, not merely the `connection_security` field, so the job attempts plaintext against a TLS listener and cannot start. Preserve the parsed original scheme and cover `stcp://... -> stcp://<site>:...`.

3. **[P2] IPv6 loopback is accepted but cannot be used.** `ConnectorManager._validate_internal_listener_security()` accepts `::1` as loopback, while `nvflare/fuel/f3/drivers/net_utils.py:295-297` emits an unbracketed URL such as `tcp://::1:9000`. The custom parser then derives empty host/port fields, and `TcpDriver` uses `AF_INET`. Either reject IPv6 listener values or implement bracketed parsing and IPv6 binding end-to-end.

These are in addition to the existing unresolved review threads. The reported CI is green, but it does not exercise the distributed Docker/Slurm mTLS launch paths above.

PR #5117, opened 2026-08-13
Clean up clients after server job failure

### LINKED ISSUE #5115: [BUG] Client jobs remain active after an abnormal server job exit
**Describe the bug**

When a launcher-managed server job process terminates abnormally, a participating client job can remain active even though there is no longer a server job process to drive it.

The client parent continues reporting the client job in its heartbeat. Server-side synchronization treats every job awaiting a client terminal outcome as if it were still running on the server, so the heartbeat response does not request cleanup. With an external scheduler such as Slurm, this can leave the client allocation reserved until delayed terminal-outcome cleanup or external scheduler termination.

**To Reproduce**

1. Configure the server and at least one client to launch job processes through an external job launcher.
2. Submit a job and wait for both the server and client job processes to become active.
3. Externally terminate only the server job process while leaving the root server and client parent processes active.
4. Observe that the client job process remains active instead of being cleaned up by heartbeat reconciliation.

**Expected behavior**

Once the root server records that the server job process has failed, subsequent client heartbeat reconciliation should request cleanup of surviving client jobs.

A normally completed server job that is still waiting for acknowledged client terminal outcomes must remain protected from premature cleanup.

**Environment**

- NVFlare `main`
- Launcher-managed server and client job processes
- Reproducible with an external batch scheduler

**Additional context**

The server tracks active server job processes separately from jobs awaiting client terminal outcomes. Heartbeat reconciliation currently combines both sets without excluding jobs that already have a recorded server-side terminal failure.

Acceptance criteria:

- A client job is selected for heartbeat cleanup after its server job has exited with a recorded failure.
- Active server jobs are not selected for cleanup.
- Normally completed server jobs awaiting client outcomes retain the terminal-outcome barrier.
- Regression tests cover all three lifecycle states.

pcnudde:
Fixes #5115.

### Description

Client heartbeat reconciliation now distinguishes jobs awaiting normal client terminal outcomes from jobs whose server process has already failed.

Pending client outcomes continue to protect normally completed server jobs from premature client cleanup. Once the server records a terminal server-process failure and removes that process from the active set, surviving client jobs are returned in the heartbeat cleanup list.

### Root cause

Heartbeat reconciliation treated every job awaiting a client terminal outcome as if it were still active on the server. That outcome barrier is required for normal completion, but it also masked orphaned client jobs after an abnormal server-process exit.

### Impact

Launcher-managed client jobs are cleaned up after a known server-process failure instead of remaining active until delayed timeout or external scheduler termination. Active server jobs and normally completed jobs awaiting client outcomes retain their existing behavior.

### Types of changes

- [x] Non-breaking bug fix
- [x] New regression tests

### Validation

- 190 focused server, client, Kubernetes, and Slurm lifecycle tests passed.
- `./runtest.sh -s` passed.
- `git diff --check` passed.
- Live external-scheduler validation was not performed.


nvflare/private/fed/server/fed_server.py
```diff
@@ -927,7 +927,11 @@ def _sync_client_jobs(self, request, client_token):
 
         client_jobs = set(client_jobs)
         outcome_jobs = self.engine.job_runner.get_client_outcome_jobs()
-        server_jobs = set(self.engine.run_processes.keys()).union(outcome_jobs)
+        # Keep normally completed jobs alive until clients report their outcomes, but do not
+        # protect client jobs after the server job has already failed.
+        server_jobs = set(self.engine.run_processes.keys()).union(
+            outcome_jobs.difference(self.engine.exception_run_processes)
```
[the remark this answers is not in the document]
pcnudde: Fixed in 74171bab4. Reconciliation now iterates the stable outcome_jobs snapshot and performs only per-job membership checks against exception_run_processes, so it no longer iterates the mutable exception dictionary. The three server-running, awaiting-client-outcome, and server-failed regression cases pass, as do the full style checks.

nvflare/private/fed/server/fed_server.py
```diff
@@ -927,7 +927,11 @@ def _sync_client_jobs(self, request, client_token):
 
         client_jobs = set(client_jobs)
         outcome_jobs = self.engine.job_runner.get_client_outcome_jobs()
-        server_jobs = set(self.engine.run_processes.keys()).union(outcome_jobs)
+        # Keep normally completed jobs alive until clients report their outcomes, but do not
+        # protect client jobs after the server job has already failed.
+        server_jobs = set(self.engine.run_processes.keys()).union(
+            {job_id for job_id in outcome_jobs if job_id not in self.engine.exception_run_processes}
```
YuanTingHsieh, line 933: **[P2] Preserve missing-outcome reconciliation for failed jobs**

`server_jobs` is used both as the client cleanup allowlist and as the set examined below for missing client outcomes. Filtering failed jobs here correctly triggers client cleanup, but it also excludes them from `jobs_on_server_but_not_on_client`. If the client terminates the job but its one-shot terminal-outcome report fails, subsequent heartbeats omit the job and `_resolve_missing_client_outcome()` is never called, so finalization waits for the default 900-second timeout.

Please use separate sets for cleanup eligibility and pending-outcome reconciliation. For an already-failed job, the missing-outcome path should resolve the pending client without overwriting the recorded server failure.

YuanTingHsieh (commented):
Found one remaining lifecycle issue in heartbeat reconciliation.

nvflare/private/fed/server/fed_server.py
pcnudde: Fixed in 215f0568f. Missing-outcome reconciliation now considers all pending outcome jobs, while the resolver leaves an existing server-side failure unchanged. The lifecycle state-matrix test now covers the failed-job/missing-client-report path. Validation: 422 server unit tests passed, ./runtest.sh -s passed, and git diff --check passed.

YuanTingHsieh (approved):

PR #5120, opened 2026-08-13
Align remote job log configuration controls

pcnudde:
## Summary

- Reuse existing site log validation for remote job-log updates.
- Apply the same controls to server and client job processes.
- Align the API documentation with supported inputs and add regression coverage.

## Validation

- 21 focused logging and command-routing tests
- `./runtest.sh -s`
- `git diff --check`


IsaacYangSLA (approved):
Use the implemented util to ensure proper config.  LGTM.

PR #5121, opened 2026-08-13
[2.8] Harden remote job logging and localhost listeners

chesterxgchen:
## Summary
- focused 2.8 security backport referencing #5109
- allow remote `configure_job_log` only for log levels and built-in modes; reject dictConfig documents and file paths that can invoke arbitrary Python code
- require the real root-server CellNet endpoint before a server-job process handles an internal command
- document the remote job-log input restriction in the FLARE API

## Scope

## Validation
- `python3 -m pytest -q tests/unit_test/private/fed/configure_job_log_security_test.py tests/unit_test/private/fed/server/server_command_agent_test.py tests/unit_test/fuel/f3/drivers/net_utils_test.py` (22 passed)


pcnudde (changes requested):
I would not bring this big of a change into past branches. Would do absolutle minimal only. 
#5120 for log config
and #5122 or simpler for SimEnv. 

We need a really good review and test on #5109 before that one is ready to merge.

chesterxgchen:
> I would not bring this big of a change into past branches. Would do absolutle minimal only. #5120 for log config and #5122 or simpler for SimEnv.
> 
> We need a really good review and test on #5109 before that one is ready to merge.
 
 I remove most of the broad changes to only few change.

chesterxgchen:
Superseded by fork-based PR #5128.

PR #5122, opened 2026-08-13
Use loopback for SimEnv/PocEnv listeners

pcnudde:
## Summary

Use explicit IPv4 loopback binding for local listeners configured with `localhost`. Non-localhost configurations retain their existing defaults.

## Testing

- Focused listener tests
- Local SimEnv and POC end-to-end tests
- Style checks


nvidianz:
A simpler fix is to make localhost listen on 127.0.0.1, instead of 0. Then, it addresses all localhost issues and no need to change simulator code.

nvidianz (dismissed):

YuanTingHsieh (commented):
Found one test-isolation issue that is currently blocking pre-merge CI.

tests/unit_test/fuel/f3/cellnet/core_cell_test.py
```diff
@@ -197,6 +198,22 @@ def callback():
...
+        secure=False,
+        credentials={},
+        internal_listener_host="127.0.0.1",
+    )
+
+    try:
+        assert cell.connector_manager.int_resources[DriverParams.HOST.value] == "127.0.0.1"
+        assert cell.connector_manager.int_resources[DriverParams.LISTEN_HOST.value] == "127.0.0.1"
+    finally:
+        CoreCell.ALL_CELLS.pop("server", None)
```
YuanTingHsieh: **[P1] Restore the global stats-pool state**

Constructing `CoreCell` registers globally named `StatsPoolManager` pools, but this cleanup removes only the cell registry entry. The pools remain registered, so a later test that creates another `server` cell raises `ValueError: pool 'request_response@server' is already defined`; this currently fails two pre-merge CI jobs. Please test `ConnectorManager` directly or save and restore the stats-pool state during cleanup.

pcnudde: Fixed in 96137308a by testing ConnectorManager directly instead of constructing CoreCell, so the test no longer registers global StatsPoolManager pools. I reproduced the former test order locally (2 passed), reran the focused listener suite (63 passed), and the replacement CI unit-test matrix is green, including Ubuntu 24.04/Python 3.10 and Ubuntu 22.04/Python 3.11.

YuanTingHsieh (approved):

PR #5123, opened 2026-08-13
[2.7] Harden remote job logging and localhost listeners

chesterxgchen:
## Summary
- focused 2.7 security backport referencing #5109
- allow remote `configure_job_log` only for log levels and built-in modes; reject dictConfig documents and file paths that can invoke arbitrary Python code
- require the real root-server CellNet endpoint before a server-job process handles an internal command

## Scope

## Validation
- `python3 -m pytest -q tests/unit_test/private/fed/configure_job_log_security_test.py tests/unit_test/private/fed/server/server_command_agent_test.py tests/unit_test/fuel/f3/drivers/net_utils_test.py` (21 passed)


chesterxgchen:
Superseded by fork-based PR #5127.

PR #5125, opened 2026-08-13
Disable inbound admin services in Simulator

nvkevlu:
Disables unneeded inbound admin services in Simulator.

### Description

Disables unneeded inbound admin services in Simulator.

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh:
Simulator still allocates an admin port, and `BaseServer.deploy()` still creates an `admin_listener=true` endpoint. This change prevents HCI callbacks from being registered, but it does not remove the inbound admin listener itself. Should Simulator also disable creation of the admin listener and add a test confirming that no admin endpoint is opened?

YuanTingHsieh (approved):

PR #5127, opened 2026-08-13
[2.7] Harden remote job logging and localhost listeners

chesterxgchen:
## Summary
- focused 2.7 security backport referencing #5109
- restrict remote `configure_job_log` and `configure_site_log` to log levels and built-in modes; reject dictConfig documents and file paths that can invoke arbitrary Python code
- require the root-server CellNet endpoint before a server-job process handles an internal command
- backport main #5026’s `xgboost<3.4` constraint: XGBoost 3.4 removed NVFLARE’s required `DataSplitMode` API

## Scope

## Validation
- targeted remote-log and server-command security tests


chesterxgchen:
Addressed: both remote `configure_site_log` handlers now validate their inputs with `validate_site_log_config`, and regression tests cover allowed log levels/modes plus rejection of dictConfig documents and config file paths.

chesterxgchen:
Addressed the remote log-level filename ambiguity: remote command paths now call `dynamic_log_config(..., allow_file_config=False)`, so an accepted level such as `DEBUG` cannot be reinterpreted as a local dictConfig file. Trusted local file/dict configuration remains supported.

chesterxgchen:
Addressed the `RELOAD` design case: remote callers may request the built-in `RELOAD` mode, but cannot supply a path. The implementation now permits file loading only when it selects the trusted workspace `reload_path` internally; arbitrary remote file paths remain disabled.

chesterxgchen:
Closing: this maintenance backport is not sufficient to resolve the full internal CellNet trust-boundary issue. The complete hardening will be prepared separately for main.

PR #5129, opened 2026-08-13
Add Client API attach acceptance coverage and docs

YuanTingHsieh:
## What changed

- add a secure two-site Client API Attach + CCWF acceptance job covering three Swarm rounds and live cross-site `submit_model`/`validate` tasks
- extend the integration harness with test-owned background processes, bounded natural-exit waiting, and process-group cleanup while preserving existing profiles that omit the field
- document the external-process rank-0 contract, CPU/Gloo torchrun coverage, multi-node Slurm ownership through the LLM-HF example, trainer-side parameter conversion, and Attach usage
- correct stale Flower metrics examples and remaining references to removed Client API/Pipe components

## Why

Attach is the supported mode when an external scheduler or application owns the trainer process. This PR verifies that two externally owned trainers can attach securely, complete Swarm training, and participate in cross-site evaluation without relying on the removed legacy execution stack.

It also makes the supported execution model explicit: rank 0 owns the Client API session for distributed training, Slurm owns multi-node allocation, and parameter representation conversion occurs at the trainer boundary rather than in the executor, CJ, or Cell transport.

Destructive process-death and payload-lifecycle testing remains in a separate test-only change so this PR stays focused on normal Attach + CCWF acceptance coverage and documentation.

## User impact

Users get an explicit multi-process/multi-node execution contract, an existing Slurm example to follow, accurate parameter-conversion boundaries, and correct Flower metric-tracking examples. The integration suite now verifies that externally owned attached trainers work through both Swarm training and cross-site evaluation.

## Validation

Initial validation used an immutable Colossus snapshot with canonical binary diff SHA256 `fad269d452a7d412f2e37c2218f2fa8db2f29e8d4e1f4c466251ff9b0627b806`:

- 244 focused tests passed
- fast Attach E2E: 10/10 passed
- compatibility canary without `background_processes`: passed, including maintained hello-CCWF
- scoped Black, isort, flake8, license, documentation-link, YAML/JSON/AST/import, and stale Flower-argument checks passed
- immediate, +45s, and +75s audits found no tested processes, listeners, or deleted working directories

On the current head, the supported consolidated selector ran the Attach + CCWF case with strict post-kill cleanup enabled:

- `NVFLARE_TEST_FRAMEWORK=client_api pytest -q system_test.py -k client_api_attach_ccwf`: 1 passed, 6 deselected in 50.44s; both trainers completed rounds 0/1/2, two `submit_model` calls, and four `validate` calls


tests/integration_test/system_test.py
```diff
@@ -78,6 +80,44 @@ def _resolve_test_config_path(config_dir: str, path: str) -> str:
...
+    deadline = time.monotonic() + 10.0
+    for process in running:
+        try:
+            process.wait(timeout=max(0.0, deadline - time.monotonic()))
+        except subprocess.TimeoutExpired:
+            try:
+                os.killpg(os.getpgid(process.pid), signal.SIGKILL)
+            except (ProcessLookupError, PermissionError):
+                pass
+            process.wait(timeout=5.0)
```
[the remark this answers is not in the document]
YuanTingHsieh: Fixed in `facab1f25`. `_stop_background_processes()` now catches the final post-SIGKILL `TimeoutExpired`, so cleanup remains best-effort and cannot mask the original test failure or prevent teardown/reset. I also ran a direct regression that forces both cleanup waits to time out, plus scoped Black, isort, and flake8.

tests/integration_test/system_test.py
```diff
@@ -78,6 +80,47 @@ def _resolve_test_config_path(config_dir: str, path: str) -> str:
...
+            running.append(process)
+            try:
+                os.killpg(os.getpgid(process.pid), signal.SIGTERM)
+            except (ProcessLookupError, PermissionError):
+                pass
+
+    deadline = time.monotonic() + 10.0
+    for process in running:
+        try:
+            process.wait(timeout=max(0.0, deadline - time.monotonic()))
```
nvkevlu: Cleanup waits only for the process-group leader. If the leader exits after SIGTERM while a descendant ignores it, this wait succeeds and the SIGKILL path is skipped; groups whose leader already exited are also omitted above. Please retain the PGID independently, check whole-group liveness after the grace period, and add a regression test with a surviving child process.

tests/integration_test/data/test_configs/standalone_job/client_api_attach_ccwf.yml
```diff
@@ -0,0 +1,37 @@
...
+          type: run_state
+          data: {run_finished: true}
+    validators:
+      - path: tests.integration_test.src.validators.LogPatternValidator
+        args:
+          required_patterns:
+            - "attached trainer session established"
+            - "Round 2 started."
+            - "to do submit_model"
+            - "to do validate"
```
nvkevlu, line 33: These patterns prove only that each invocation was logged at least once. They do not prove the advertised two successful submit_model calls and four successful validations; the validation message is emitted before execution, and a failed CSE result need not fail the job. Please validate the generated results, for example by requiring all four site-{1,2}.site-{1,2}.val_mean paths with JsonStatsResultValidator.

docs/hello-world/hello-flower/index.rst
```diff
@@ -122,13 +125,16 @@ The Job Recipe contains the Flower app configuration and deploys it within NVFla
...
 .. code-block:: python
 
+    from nvflare.app_opt.flower.recipe import FlowerRecipe
+    from nvflare.recipe import SimEnv, add_experiment_tracking
+
     recipe = FlowerRecipe(
         name="hello-flower",
         min_clients=n_clients,
-        num_rounds=num_rounds,
         flower_app_path="local/custom/flwr-pt", # local/custom is the mandatory location for flower apps.
```
nvkevlu, line 134: The documented pre-deployed command cannot run with the accompanying job.py: its parser requires --content_dir and does not define --flower_app_path. Please add mutually exclusive CLI options and pass the selected value to FlowerRecipe, or remove the pre-deployed command. The requirements/version guidance should also be aligned with the release that supports this argument.

docs/user_guide/data_scientist_guide/client_api_usage.rst
```diff
@@ -196,6 +196,8 @@ For complete working examples across different frameworks:
 * PyTorch Lightning: :github_nvflare_link:`hello-lightning <examples/hello-world/hello-lightning>`
 * TensorFlow: :github_nvflare_link:`hello-tf <examples/hello-world/hello-tf>`
 * HuggingFace Trainer: :github_nvflare_link:`hello-huggingface <examples/hello-world/hello-huggingface>`
+* Multi-GPU/multi-node Slurm: :github_nvflare_link:`LLM-HF <examples/advanced/llm_hf/MULTINODE.md>`
+* Independently managed trainer: :github_nvflare_link:`Client API Attach <examples/advanced/client-api-attach>`
```
nvkevlu, line 200: Adding Attach makes the following statement inaccurate: this example uses trainer.py, not client.py. Consider saying that each example includes a training entry point and job definition.

tests/integration_test/system_test.py
YuanTingHsieh: Fixed in 2e570e51c. The runner records each PGID at launch, tests whole-group liveness after SIGTERM, then SIGKILLs any remaining group even if its leader has already exited. Added a regression that leaves a child ignoring SIGTERM after its leader exits; it passes.

tests/integration_test/data/test_configs/standalone_job/client_api_attach_ccwf.yml
YuanTingHsieh: Fixed in 2e570e51c. The Attach+CCWF profile now uses JsonStatsResultValidator to require site-1.site-1, site-1.site-2, site-2.site-1, and site-2.site-2 val_mean results. The exact system profile passes with this validator (1 passed, 50.80s).

docs/hello-world/hello-flower/index.rst
YuanTingHsieh: Fixed in 2e570e51c. Removed the invalid simulation command and document pre-deployed Flower apps with ProdEnv, where the administrator installs the app under local/custom. Updated the example requirements and Flower documentation to the 2.9 release-candidate line, which contains flower_app_path support.

docs/user_guide/data_scientist_guide/client_api_usage.rst
YuanTingHsieh: Fixed in 2e570e51c. The guide now describes each example as including a training entry point, such as client.py or an externally managed trainer.py, plus its job definition.

tests/integration_test/system_test.py
```diff
@@ -78,6 +80,73 @@ def _resolve_test_config_path(config_dir: str, path: str) -> str:
...
+
+
+def _wait_for_process_groups(processes, timeout: float) -> list:
+    """Wait for process groups to exit and return any that remain alive."""
+
+    deadline = time.monotonic() + timeout
+    remaining = [item for item in processes if _is_process_group_alive(item[2])]
+    while remaining and time.monotonic() < deadline:
+        time.sleep(min(0.1, max(0.0, deadline - time.monotonic())))
+        remaining = [item for item in remaining if _is_process_group_alive(item[2])]
```
nvkevlu, line 107: Reap process leaders while waiting for groups

A leader terminated by SIGTERM remains a zombie until Popen.poll() or wait() reaps it. Because _wait_for_process_groups() only calls killpg(pgid, 0), it continues reporting the group as alive through the full 10-second graceful timeout and 5-second kill timeout, even when the process exited immediately. Please poll/reap each direct leader during the liveness loop; descendants will remain discoverable through the retained PGID.

tests/unit_test/integration_test/system_test_background_process_test.py
```diff
@@ -0,0 +1,79 @@
...
+    program = textwrap.dedent(
+        f"""
+        import os
+        import signal
+        import time
+        from pathlib import Path
+
+        child_pid = os.fork()
+        if child_pid:
+            Path({str(child_ready)!r}).write_text(str(child_pid), encoding="utf-8")
```
nvkevlu: Signal readiness after installing the handler

The parent writes child_ready immediately after fork(), before the child installs its SIGTERM-ignore handler. Cleanup can therefore send SIGTERM first, allowing this test to pass without exercising the surviving-descendant/SIGKILL path. Please have the child install the handler and then write the readiness marker itself.

tests/unit_test/integration_test/system_test_background_process_test.py
```diff
@@ -0,0 +1,79 @@
...
+
+        signal.signal(signal.SIGTERM, signal.SIG_IGN)
+        while True:
+            time.sleep(1)
+        """
+    )
+    command = f"{sys.executable} -c {shlex.quote(program)}"
+    process = system_test.run_command_in_subprocess(command)
+    background_process = (command, process, process.pid)
+
```
nvkevlu: Arm cleanup immediately after process launch

The readiness wait and assertion occur before the try/finally. If startup is delayed or the assertion fails, the infinite process group is leaked. Please enter the try immediately after launching the process so every setup failure reaches the cleanup block.

tests/unit_test/integration_test/system_test_background_process_test.py
```diff
@@ -0,0 +1,79 @@
...
+import pytest
+
+_REPO_ROOT = Path(__file__).resolve().parents[3]
+_INTEGRATION_ROOT = _REPO_ROOT / "tests" / "integration_test"
+
+
+def _load_system_test_module(monkeypatch):
+    monkeypatch.chdir(_INTEGRATION_ROOT)
+    monkeypatch.setenv("NVFLARE_TEST_FRAMEWORK", "client_api_attach_ccwf")
```
nvkevlu: This selector now silently runs the NumPy configuration

client_api_attach_ccwf was removed from test_configs.yml, so system_test.py silently falls back to NumPy. The PR body’s acceptance command uses the same removed selector and therefore no longer reproduces the Attach test on this head. Please restore the dedicated alias or use a supported selector, update the validation command, and rerun the actual Attach case on the final commit.

nvkevlu (changes requested):

tests/integration_test/system_test.py
YuanTingHsieh: Fixed in 77482555d. `_wait_for_process_groups()` now polls every directly owned `Popen` before probing its retained PGID, so a SIGTERM’d leader is reaped while a surviving descendant remains discoverable by `killpg`.

tests/unit_test/integration_test/system_test_background_process_test.py
YuanTingHsieh: Removed in 77482555d. I agree the fork/readiness regression was fragile test-harness plumbing. The actual Attach+CCWF system profile remains as the acceptance test and now runs through the consolidated `client_api` selector.

tests/unit_test/integration_test/system_test_background_process_test.py
YuanTingHsieh: Removed in 77482555d with the entire bespoke process test, so it cannot leak during setup. Acceptance coverage remains the actual Attach+CCWF profile, not a synthetic infinite-child test.

tests/unit_test/integration_test/system_test_background_process_test.py
YuanTingHsieh: Fixed without restoring a separate stage. The supported command is `NVFLARE_TEST_FRAMEWORK=client_api pytest -q system_test.py -k client_api_attach_ccwf`; it passed on the current head with 1 passed, 6 deselected in 50.44s. The PR validation section now uses this command.

tests/integration_test/system_test.py
```diff
@@ -78,6 +80,81 @@ def _resolve_test_config_path(config_dir: str, path: str) -> str:
...
+        except (ProcessLookupError, PermissionError):
+            pass
+
+    remaining_groups = _wait_for_process_groups(running_groups, graceful_timeout)
+    for _, _, pgid in remaining_groups:
+        try:
+            os.killpg(pgid, signal.SIGKILL)
+        except (ProcessLookupError, PermissionError):
+            pass
+    _wait_for_process_groups(remaining_groups, kill_timeout)
```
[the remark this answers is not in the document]
YuanTingHsieh: Fixed in 0c56846fa. Process-group cleanup is now wrapped so configured teardown commands and `reset_test_info()` always run before a survivor error propagates. The supported Attach+CCWF integration case passes on this commit with strict post-kill cleanup enabled.

tests/integration_test/system_test.py
```diff
@@ -192,51 +276,73 @@ def test_run_job_complete(self, setup_and_teardown_system):
...
+                for command in background_commands:
+                    print(f"Starting background process: {command}")
+                    process = run_command_in_subprocess(command)
+                    # run_command_in_subprocess starts a new session, making its
+                    # PID the process-group ID even if the leader exits later.
+                    background_processes.append((command, process, process.pid))
+
+                test_driver.run_event_sequence(event_sequence)
+                _wait_for_background_processes(background_processes)
+
```
nvkevlu, line 304: Retire PGIDs after successful natural exit

_wait_for_background_processes() returns only after every leader has been reaped and every process group has disappeared. Validators then run before the finally block passes the old numeric PGIDs to _stop_background_processes(). Those IDs can be reused during validation, allowing cleanup to signal an unrelated process group. Please clear or mark background_processes immediately after a successful wait, retaining the entries only when waiting raises and descendants may still exist.

tests/integration_test/test_configs.yml
```diff
@@ -29,9 +29,7 @@ test_configs:
     - ./data/test_configs/standalone_job/ext_process_streaming.yml
     - ./data/test_configs/standalone_job/swarm_external_process_pass_through.yml
     - ./data/test_configs/standalone_job/swarm_tensor_disk_offload.yml
+    - ./data/test_configs/standalone_job/client_api_attach_ccwf.yml
+    - ./data/test_configs/standalone_job/model_controller_api.yml
   client_api_qa:
     - ./data/test_configs/standalone_job/client_api_qa.yml
```
nvkevlu, line 35: Preserve the existing standalone test selectors

This consolidation removes the existing model_controller_api and ext_process_streaming selectors. Consequently, the former ext_process_streaming command silently runs NumPy, while ci/run_integration.sh model_controller_api now exits as an unknown mode. Please retain compatibility aliases while still including these configurations in client_api, or explicitly coordinate and document the breaking change.

YuanTingHsieh: those are added by accident, this is intended

tests/integration_test/system_test.py
YuanTingHsieh: Fixed in 4e04083aa. `background_processes.clear()` now runs immediately after the natural-exit wait succeeds and before validators. If that wait raises, the list remains intact for the `finally` cleanup path.

nvkevlu (approved):
LGTM

PR #5131, opened 2026-08-14
Harden confidential computing authorizers

IsaacYangSLA:
## Summary

- migrate GPU confidential-computing attestation from the retired Python SDK to the native NVAT `nvattest` CLI, with local or remote appraisal and Rego relying-party policy support
- harden SNP and TDX evidence generation and verification with stricter parsing, bounded subprocess execution, nonce replay protection, and safer certificate/token handling
- document the updated NVAT configuration and add focused unit coverage for GPU, SNP, TDX, and nonce-history behavior

## Validation

- `pytest -q tests/unit_test/app_opt/confidential_computing` (61 passed)
- `PYENV_VERSION=3.11.10 ./runtest.sh -s`


IsaacYangSLA:
This PR handles authorizers and is the prerequisite of later CoCo-related PR.

nvflare/app_opt/confidential_computing/gpu_authorizer.py
```diff
@@ -12,122 +12,280 @@
...
+            raise ValueError("verifier must be 'local' or 'remote'")
+        if cmd_timeout <= 0:
+            raise ValueError("cmd_timeout must be positive")
+        if max_token_size <= 0:
+            raise ValueError("max_token_size must be positive")
+
+        if verifier_url:
+            nras_url = verifier_url.removesuffix("/v4/attest/gpu")
+        if verifier == "remote" and (not isinstance(nras_url, str) or not nras_url.startswith("https://")):
+            raise ValueError("nras_url must be an HTTPS URL for remote verification")
```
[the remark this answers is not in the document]
IsaacYangSLA: Addressed in `3efc3b1bf`: `verifier_url` now rejects non-string and empty values, trims surrounding whitespace and trailing slashes, and correctly removes the legacy endpoint suffix. Added coverage for both trailing-slash forms and invalid values. The confidential-computing suite passes 66 tests and the repository style check passes.

nvflare/app_opt/confidential_computing/gpu_authorizer.py
```diff
@@ -12,122 +12,280 @@
...
+                self.relying_party_policy = policy_stream.read()
+            if not self.relying_party_policy.strip():
+                raise ValueError("NVAT relying-party policy must not be empty")
+            if "package " not in self.relying_party_policy or "nv_match" not in self.relying_party_policy:
+                raise ValueError("NVAT relying-party policy must be Rego defining nv_match")
+            self._uses_default_policy = False
+
+        if service_key is None:
+            service_key = os.environ.get(NVAT_SERVICE_KEY_ENV) or os.environ.get(LEGACY_SERVICE_KEY_ENV)
+        self.service_key = service_key
```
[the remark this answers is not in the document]
IsaacYangSLA: Addressed in `3efc3b1bf`: explicit and environment-derived `service_key` values are now required to be non-empty strings and are normalized before use. Added regression coverage for valid whitespace normalization plus empty, whitespace-only, and non-string values. The confidential-computing suite passes 66 tests and the repository style check passes.

chesterxgchen (commented):
## Findings

  1. [P1] SNP freshness is not verifier-bound. The SNP issuer chooses its own nonce, while the verifier only checks a bounded, in-memory history. The tests explicitly allow an old
     nonce again after eviction; restarting the verifier clears the history entirely. A compromised participant can therefore replay a previously valid report after eviction or
     restart. The protocol should supply a verifier challenge bound into report_data, or durably retain evidence identifiers for their entire validity period.
     snp_authorizer.py (https://github.com/NVIDIA/NVFlare/blob/3efc3b1bf0840a5c5cad99ff2667329a02f0be47/nvflare/app_opt/confidential_computing/snp_authorizer.py#L183-L248) · utils.py
     (https://github.com/NVIDIA/NVFlare/blob/3efc3b1bf0840a5c5cad99ff2667329a02f0be47/nvflare/app_opt/confidential_computing/utils.py#L19-L34) · eviction test
     (https://github.com/NVIDIA/NVFlare/blob/3efc3b1bf0840a5c5cad99ff2667329a02f0be47/tests/unit_test/app_opt/confidential_computing/utils_test.py#L23-L35)

  2. [P1] SNP evidence can trigger unbounded work before authenticity is established. verify() accepts arbitrarily large base64 input, writes it to disk, extracts attacker-controlled
     chip/TCB fields, and performs retried VCEK retrieval before signature verification. An authenticated malicious site can consume verifier memory, disk, KDS traffic, cache space,
     and retry time. Enforce a strict report-size bound—preferably the pinned ABI size—and bound/rate-limit endorsement retrieval and cache growth. Deterministic KDS failures should
     not receive exponential retries.
     snp_authorizer.py (https://github.com/NVIDIA/NVFlare/blob/3efc3b1bf0840a5c5cad99ff2667329a02f0be47/nvflare/app_opt/confidential_computing/snp_authorizer.py#L212-L228) · retry and
     VCEK handling (https://github.com/NVIDIA/NVFlare/blob/3efc3b1bf0840a5c5cad99ff2667329a02f0be47/nvflare/app_opt/confidential_computing/snp_authorizer.py#L153-L175)

  3. [P2] TDX verification inherits generation’s automatic sudo. When running as non-root and sudo exists, _command() prefixes both generation and verification with sudo -n.
     Verification processes peer-controlled tokens and does not need generation privileges. This may fail on systems without passwordless sudo and unnecessarily increases the privilege
     of token parsing. Use separate privilege settings, with verification unprivileged by default.
     tdx_authorizer.py (https://github.com/NVIDIA/NVFlare/blob/3efc3b1bf0840a5c5cad99ff2667329a02f0be47/nvflare/app_opt/confidential_computing/tdx_authorizer.py#L85-L100)

  4. [P2] The documented GPU args are outside the issuer entry. In both examples, args is shown at the root after the cc_issuers item has ended. Copying the example will not configure
     that authorizer. Nest args beneath the GPU issuer and recommend environment/secret injection for service_key.
     deployment guide
     (https://github.com/NVIDIA/NVFlare/blob/3efc3b1bf0840a5c5cad99ff2667329a02f0be47/docs/user_guide/confidential_computing/on_premises/cc_deployment_guide.rst#L671-L685) · example
     README (https://github.com/NVIDIA/NVFlare/blob/3efc3b1bf0840a5c5cad99ff2667329a02f0be47/examples/advanced/cc_provision/README.md#L134-L150)

IsaacYangSLA:
@chesterxgchen Addressed all four findings in `badf744de`:

1. SNP replay protection now durably records verified `report_data` nonce hashes in an atomic SQLite store (mode `0600`), so replay rejection survives bounded-history eviction and verifier restart. The compatibility `max_nonce_history` setting no longer evicts SNP evidence identifiers. Added a restart/eviction regression test.
2. SNP evidence is rejected unless it has the exact pinned ABI size (`0x4A0` / 1184 bytes), including a pre-decode base64 length bound. VCEK retrieval is serialized, rate-limited, cache-bounded, and performed once without exponential retry for report-derived failures. Added size, no-retry, and cache-limit tests.
3. TDX generation and verification now have separate privilege controls. `use_sudo` applies only to generation; verification is unprivileged by default, with explicit `verify_use_sudo=False` configuration and regression coverage.
4. The GPU `args` mappings are nested under the GPU issuer in both YAML examples. The existing guidance recommends `NV_ATTESTATION_SERVICE_KEY`/secret injection and keeping credentials outside images and startup kits.

Validation: all 74 confidential-computing unit tests pass, the repository style check passes, and `git diff --check` passes.

These findings are in a review summary rather than inline review threads, so GitHub does not provide individual thread-resolution controls for this review.

chesterxgchen (commented):
Latest head badf744d closes the original size-bound, retry, TDX-sudo, and documentation findings. Two significant SNP issues remain.

  ## Findings

  1. [P1] VCEK rate limiting can reject legitimate sites during cold start. After fetching one previously unseen VCEK, another distinct site arriving within one second is rejected
     instead of delayed or queued. Cross-site verification is sequential, so a multi-site deployment can fail depending on KDS response speed. I reproduced this: the first distinct
     SNP site passed and the second returned VCEK fetch rate limit exceeded. Use bounded waiting or a request queue, and test consecutive reports with distinct chip IDs.
     snp_authorizer.py (https://github.com/NVIDIA/NVFlare/blob/badf744de7249f08a680bc0b3ffbfa88813ded91/nvflare/app_opt/confidential_computing/snp_authorizer.py#L307-L340)

  2. [P1] Durable deduplication still does not establish freshness. The requester supplies no challenge; the issuer independently generates report_data. SQLite prevents a report from
     being accepted twice, but a previously captured report that has never reached this verifier remains acceptable indefinitely. Proper freshness requires a request-scoped verifier
     challenge passed through generate(challenge) and checked by verify(token, challenge).
     request creation (https://github.com/NVIDIA/NVFlare/blob/badf744de7249f08a680bc0b3ffbfa88813ded91/nvflare/app_opt/confidential_computing/cc_manager.py#L506-L520) · nonce
     generation (https://github.com/NVIDIA/NVFlare/blob/badf744de7249f08a680bc0b3ffbfa88813ded91/nvflare/app_opt/confidential_computing/snp_authorizer.py#L214-L231)

  3. [P2] The replay database grows without a lifecycle bound. Every accepted validation permanently inserts a row; there is no expiration, pruning, or maximum size. The now-unused
     max_nonce_history parameter suggests bounded storage but no longer provides it. Challenge-based verification would remove the need for permanent history; otherwise define
     retention and cleanup.
     replay persistence (https://github.com/NVIDIA/NVFlare/blob/badf744de7249f08a680bc0b3ffbfa88813ded91/nvflare/app_opt/confidential_computing/snp_authorizer.py#L277-L289)

nvflare/app_opt/confidential_computing/snp_authorizer.py
```diff
@@ -96,164 +132,222 @@ def __init__(
...
+                vcek_path = self._ensure_amd_vcek(vcek_cache_key, report_path)
+                for filename in (AMD_ARK, AMD_ASK):
+                    shutil.copy2(os.path.join(self.amd_certs_dir, filename), os.path.join(work_dir, filename))
+                shutil.copy2(vcek_path, os.path.join(work_dir, AMD_VCEK))
+
+                # Signature verification is local and deterministic. Retrying a
+                # bad report only amplifies an attacker's resource consumption;
+                # bounded retries are reserved for the AMD CA certificate fetch.
+                self._run_once(
+                    [self.snpguest_binary, "verify", "attestation", work_dir, report_path],
```
YuanTingHsieh, line 268: [P1] Authenticate the cached VEK before trusting it

`_ensure_amd_ca_certs()` fetches ARK/ASK, but this path only runs `snpguest verify attestation`. In upstream `snpguest`, that subcommand checks the report's TCB/signature against the supplied VEK; AMD chain authentication is the separate `verify certs` subcommand. Therefore a pre-seeded or replaced cached `vcek-*.pem` plus a matching report can pass without proving that the VEK was signed by the AMD ASK. Please run `snpguest verify certs` on the assembled work directory before accepting the report and add a regression test for a chain failure.

YuanTingHsieh (commented):
The latest commit addresses the SNP size bound, TDX sudo scope, and NVAT documentation indentation findings. One security issue remains: SNP verification validates the report with the cached VEK but does not authenticate that VEK through the AMD ASK/ARK chain.

nvflare/app_opt/confidential_computing/snp_authorizer.py
IsaacYangSLA: Addressed in the latest update. Cached and newly fetched VCEKs are now authenticated with `snpguest verify certs` before `verify attestation`. If certificate-chain verification fails, attestation verification is not attempted. A regression test covers that failure path.

IsaacYangSLA:
@chesterxgchen Addressed the follow-up findings:

- Added request-scoped verifier challenges and propagated them through `CCManager`, `generate(challenge)`, and `verify(token, challenge)`. Each remote site receives a distinct locally generated challenge, and returned evidence is bound to that local value before verification.
- Bound SNP `report_data` and GPU evidence nonces to the verifier challenge. Mismatched or missing challenges are rejected before external verification commands run.
- Removed the SQLite replay database; freshness is now request-scoped, so there is no unbounded replay table.
- Changed VCEK rate limiting to bounded serialized waiting, allowing consecutive requests for distinct chip IDs instead of immediately rejecting the second request.
- Authenticated cached and fetched VCEKs with `snpguest verify certs` before attestation verification.

Validation: confidential-computing tests pass (80 passed), full unit suite passes (9767 passed, 87 skipped), and `./runtest.sh -s` passes.

nvflare/app_opt/confidential_computing/cc_manager.py
```diff
@@ -528,8 +544,21 @@ def _collect_all_site_tokens(self, fl_ctx: FLContext) -> dict[str, list[dict[str
...
+                                    if not isinstance(token_info, dict):
+                                        break
+                                    token_info = dict(token_info)
+                                    token_info[CC_CHALLENGE] = challenge
+                                    bound_cc_info.append(token_info)
+                                if len(bound_cc_info) != len(cc_info):
+                                    self.logger.warning(f"Invalid CC token data from {target_site}: {cc_info}")
+                                    failed_sites.append(target_site)
+                                    continue
+                                all_tokens[site_name] = bound_cc_info
```
YuanTingHsieh, line 669: [P1] Bind the response to the requested site identity

`site_name` is controlled by the responding peer, but it becomes the key passed to `_verify_participants_tokens()`, which automatically skips keys absent from `cc_enabled_sites`. A compromised enabled target can therefore return `site_name="bogus"`; the expected site disappears, the bogus identity is marked valid without calling a verifier, and cross-site validation succeeds. I reproduced exactly that result. Preserve the FQCN-to-name mapping from `all_sites`, reject a response whose identity does not match its target, and assert that the collected identities exactly equal the requested enabled sites.

nvflare/app_opt/confidential_computing/cc_manager.py
```diff
@@ -209,7 +222,7 @@ def _setup_cc_authorizers(self, fl_ctx):
 
     def _generate_and_attach_tokens(self, fl_ctx: FLContext):
         """Generate and attach CC tokens for sending to peer."""
-        cc_infos = self._generate_fresh_tokens_for_validation()
+        cc_infos = self._generate_fresh_tokens_for_validation(_new_challenge())
```
YuanTingHsieh, line 251: [P1] Have the verifier issue the registration challenge

This registration path creates the challenge on the same participant that generates the evidence, then sends both token and challenge to the verifier. Replaying an old complete `CC_INFO` therefore still passes: SNP only checks that the report contains the peer-supplied value, and its replay store was removed in this commit. The later CC-channel handshake is verifier-bound, but registration is not. Add a verifier-to-attester challenge round trip for registration, or retain replay protection until every validation path is verifier-bound.

nvflare/app_opt/confidential_computing/tdx_authorizer.py
```diff
@@ -12,76 +12,181 @@
...
+            raise RuntimeError(f"TDX {action} timed out after {self.cmd_timeout} seconds") from e
+        except OSError as e:
+            raise RuntimeError(f"TDX {action} could not start: {e}") from e
+
+    def generate(self, challenge: str | None = None) -> str:
         try:
-            with open(error_file, "r") as f:
-                if "Error:" in f.read():
+            result = self._run(
+                self._command("-c", self.config_file, "token", *self.token_options, use_sudo=self.use_sudo),
```
YuanTingHsieh, line 143: [P1] Bind the TDX token to the request challenge

The new `challenge` argument is ignored by both `generate()` and `verify()`: it is not passed to the CLI or checked in signed claims. I reproduced the same compact JWT being accepted by fresh `TDXAuthorizer` instances under two unrelated challenges. The bounded fingerprint history does not protect another verifier or a restart, and the CCManager test masks this by replacing `TDXAuthorizer.verify` with a challenge-aware mock. Bind the challenge into attested user data and verify the corresponding signed claim, or explicitly declare this authorizer challenge-unsupported so the manager cannot report verifier-bound freshness. `ACIAuthorizer` and `AZCVMAuthorizer` currently have the same ignored-argument behavior.

YuanTingHsieh (commented):
Re-review of `63649c6`: the VEK-chain check, bounded VCEK waiting, and SNP/GPU challenge binding are implemented correctly. Three security issues remain: remote responses are keyed by a peer-supplied identity that can skip verification, registration evidence uses an attester-generated rather than verifier-issued challenge, and TDX (as well as the Azure authorizers) accepts the challenge parameter without binding it into signed evidence. The focused confidential-computing suite (80 tests), repository style check, and current CI matrix pass, but those tests do not cover these bypasses.

IsaacYangSLA:
@YuanTingHsieh Addressed the follow-up findings in `c7348d4cf`:

- Cross-site token collection now checks the responding site's self-reported `site_name` against the identity expected for the target FQCN (from the server-provided site list), rejecting any mismatch instead of trusting the peer. It also asserts the final collected identity set exactly equals the requested enabled sites. Added a regression test for a site claiming a different identity.
- Registration evidence uses an attester-chosen challenge (not verifier-issued), so a captured report+challenge pair would otherwise remain valid forever. `CCManager` now applies a bounded in-memory replay cache to registration tokens specifically, since a verifier-issued round trip isn't possible before a site has registered. Added a regression test confirming a replayed registration token is rejected on the second use.
- Added `CCAuthorizer.supports_challenge_binding()` (default `False`) so an authorizer can declare whether it cryptographically binds `challenge` into its evidence. SNP and GPU override it to `True`. TDX explicitly overrides to `False`, logs a warning at init, and `CCManager` now applies the same bounded replay cache to any authorizer reporting `False` (covering TDX, and ACI/AZCVM which are updated the same way), even outside registration.

Validation: confidential-computing suite passes (84 passed), repository style check passes.

nvflare/app_opt/confidential_computing/cc_manager.py
IsaacYangSLA: Addressed in `c7348d4cf`: cross-site token collection now checks the responding site's self-reported `site_name` against the identity expected for that FQCN (from the server-provided site list, never from the peer itself), rejecting any mismatch instead of accepting it. It also asserts the final collected identity set exactly equals the requested enabled sites, so a spoofed or dropped identity fails loudly instead of silently passing. Added `test_collect_tokens_rejects_site_identity_spoofing` covering a site claiming a different identity. The confidential-computing suite passes (84 tests) and the style check passes.

nvflare/app_opt/confidential_computing/cc_manager.py
IsaacYangSLA: Addressed in `c7348d4cf`: registration evidence uses an attester-chosen challenge (not verifier-issued), and a verifier-issued round trip isn't possible before a site has registered, so `CCManager` now applies a bounded in-memory replay cache to registration tokens specifically. A captured report+challenge pair can therefore only be accepted once per process lifetime. Added `test_verify_participants_tokens_rejects_replayed_registration_token` covering the second use of the same token being rejected. The confidential-computing suite passes (84 tests) and the style check passes.

nvflare/app_opt/confidential_computing/tdx_authorizer.py
IsaacYangSLA: Addressed in `c7348d4cf`: added `CCAuthorizer.supports_challenge_binding()` (default `False`) so an authorizer can declare whether `challenge` is cryptographically bound into its evidence. `TDXAuthorizer` explicitly overrides it to `False` and logs a warning at init that it cannot provide verifier-bound freshness. `CCManager` now applies the same bounded replay cache used for registration to any authorizer reporting `False`, including outside registration, so a captured TDX token can only be accepted once. `ACIAuthorizer`/`AZCVMAuthorizer` are updated the same way for consistency. Added `test_verify_participants_tokens_rejects_replay_for_non_challenge_binding_authorizer`. The confidential-computing suite passes (84 tests) and the style check passes.

nvflare/app_opt/confidential_computing/cc_manager.py
```diff
@@ -239,17 +264,23 @@ def _validate_client_tokens(self, fl_ctx: FLContext):
             self._shutdown_system(msg, fl_ctx)
             return
 
-        self._validate_cc_infos(peer_cc_info, fl_ctx)
+        self._validate_cc_infos(peer_cc_info, fl_ctx, is_registration=True)
```
YuanTingHsieh, line 306: [P1] Bind registration evidence to the authenticated peer identity

This passes the participant name from peer-controlled `CC_INFO` without requiring it to equal `peer_ctx.get_identity_name()`. `_verify_participants_tokens()` then treats every key outside `cc_enabled_sites` as valid and skips its verifier. I reproduced this with an authenticated `client1` supplying `CC_INFO={"bogus": [...]}`: the verifier was never called and shutdown was not triggered. Please reject registration evidence unless its participant identity exactly matches the authenticated peer (and apply the corresponding expected-server identity check on the client side).

nvflare/app_opt/confidential_computing/cc_manager.py
```diff
@@ -258,22 +289,43 @@ def _validate_cc_infos(self, participants_cc_info: dict[str, list[dict[str, str]
...
+        Returns:
+            True if this (namespace, token) pair was already accepted.
+        """
+        token_fingerprint = hashlib.sha256(token.encode("utf-8")).hexdigest()
+        key = (namespace, token_fingerprint)
+        with self._seen_tokens_lock:
+            if key in self._seen_tokens:
+                return True
+            self._seen_tokens[key] = None
+            if len(self._seen_tokens) > MAX_SEEN_TOKENS:
```
YuanTingHsieh, line 358: [P1] Do not rely on a process-local bounded cache for freshness

For authorizers without challenge binding, this cache only rejects duplicates within one `CCManager` lifetime and until 4,096 newer entries evict the fingerprint. I reproduced the same otherwise-valid TDX token being accepted by two fresh manager/authorizer instances; restart clears both replay histories, and SNP registration evidence has no expiry in this implementation. A captured token can therefore be accepted as a first-use replay at another verifier or after restart/eviction. Please provide verifier-bound or durable replay protection, or explicitly narrow the guarantee and ensure no security decision relies on freshness here.

nvflare/app_opt/confidential_computing/cc_manager.py
```diff
@@ -258,22 +289,43 @@ def _validate_cc_infos(self, participants_cc_info: dict[str, list[dict[str, str]
...
         else:
             return ""
 
+    def _token_already_used(self, namespace: str, token: str) -> bool:
+        """Bounded in-memory replay check for evidence without verifier-bound freshness.
+
+        Returns:
+            True if this (namespace, token) pair was already accepted.
+        """
+        token_fingerprint = hashlib.sha256(token.encode("utf-8")).hexdigest()
```
YuanTingHsieh: [P3] Normalize accepted bytes tokens before hashing

`SNPAuthorizer.verify()` explicitly accepts both `str` and `bytes`, but after it returns `True` this line unconditionally calls `token.encode()`. I reproduced an actual bytes-form SNP token passing `SNPAuthorizer.verify()` and then raising `AttributeError: 'bytes' object has no attribute 'encode'` here. Please normalize the accepted serialization before hashing, or consistently reject non-string tokens before verification.

YuanTingHsieh (commented):
Re-reviewed the current head. The registration identity binding issue is the strongest blocker; the replay comment is scoped to the freshness guarantee for non-challenge-binding evidence; the SNP bytes issue is lower-severity robustness feedback.

nvflare/app_opt/confidential_computing/cc_manager.py
IsaacYangSLA: Addressed in `d9911baab`: `_validate_client_tokens` now requires `peer_cc_info.keys()` to exactly equal `{peer_ctx.get_identity_name()}`, rejecting registration otherwise. Previously an authenticated client could submit `CC_INFO` under an unlisted/bogus key, which `_verify_participants_tokens` auto-trusts (since it's outside `cc_enabled_sites`), letting the client's real identity go unverified entirely -- this reproduces and closes exactly that. Applied the corresponding check on the client side in `_validate_server_tokens`: the server's CC info must claim a participant identity the client actually has CC configuration for (`cc_enabled_sites` membership). I noted in the client-side comment why this is the strongest check available there: unlike the server, the client has no independent pre-registration channel to learn the server's asserted identity to compare against. Added `test_validate_client_tokens_rejects_identity_mismatch` and `test_validate_server_tokens_rejects_unlisted_identity`. Suite passes (88 tests), style check passes.

nvflare/app_opt/confidential_computing/cc_manager.py
IsaacYangSLA: Acknowledged in `d9911baab` -- narrowing the guarantee rather than claiming durability the cache doesn't have, per your suggested fallback. Added an explicit LIMITATION block above `MAX_SEEN_TOKENS` in cc_manager.py: this cache is process-local and non-durable, clears on restart, evicts after 4096 newer entries, and does not provide verifier-bound freshness. A captured token can still be accepted as a first use by a different verifier instance or after restart/eviction.

A fully durable/verifier-bound fix would need a verifier-issued challenge round trip for registration specifically, which the current registration protocol can't support without deeper redesign: the client has no connection to request a challenge from until *after* it authenticates via the existing mTLS-based `ClientManager.authenticate()` flow, which itself runs after this CC check in the registration handler. I called that out explicitly in the new comment so it's not accidentally relied on elsewhere, and noted that the actual identity/authorization guarantee for registration comes from that separate mTLS-based authentication, not from this cache. Happy to open a follow-up issue to track the deeper protocol redesign if you'd like it tracked separately from this PR.

nvflare/app_opt/confidential_computing/cc_manager.py
IsaacYangSLA: Addressed in `d9911baab`: `_token_already_used` now normalizes the token to bytes before hashing (`token if isinstance(token, bytes) else str(token).encode("utf-8")`) instead of unconditionally calling `.encode()`, so a bytes-form token from `SNPAuthorizer.verify()` no longer raises `AttributeError`. Added `test_token_already_used_normalizes_bytes`, which also checks the bytes and str forms of the same token dedup consistently. Suite passes (88 tests), style check passes.

YuanTingHsieh (approved):

PR #5143, opened 2026-08-17
Surface resource manager admission errors

pcnudde:
> **Post-merge privacy follow-up:** #5148 keeps formatted resource-manager exception details site-local and returns a fixed, bounded diagnostic across the site boundary. The original implementation described below returned the formatted detail in the resource-check response.

## Description

Resource-manager exceptions during job admission were converted into a generic `not enough sites have enough resources` result. This made incompatible or malformed resource requirements look like temporary capacity shortages and left operators without the underlying site-specific reason.

This change:

- records resource-manager exceptions in the client log and resource-check response
- includes non-empty site-specific rejection reasons in job scheduling history
- clarifies that `resource_spec["@default"]` applies to every targeted client and the server, while named client or `server` blocks shallow-override individual fields
- keeps resource normalization and admission behavior unchanged

Jobs that do not match their configured resource manager remain rejected, but the responsible site and reason are now visible instead of being silently reduced to a generic capacity message.

## Validation

- 22 focused scheduler and resource-check tests passed
- `./runtest.sh -s` passed
- `./build_doc.sh --html` passed
- `git diff --check` passed


chesterxgchen (approved):

nvflare/app_common/job_schedulers/job_scheduler.py
```diff
@@ -206,17 +207,18 @@ def _try_job(self, job: Job, fl_ctx: FLContext) -> (int, Optional[Dict[str, Disp
             else:
                 if site_name in required_sites:
                     no_resource_message += site_name + ":" + token + ";"
```
holgerroth, line 209: The new `if token:` on the next line tolerates `None`/empty tokens, but this pre-existing line still concatenates `token` unguarded: a resource manager returning `(False, None)` for a **required** site raises `TypeError: can only concatenate str (not "NoneType") to str` here (the `DummyResourceManager` in `job_scheduler_test.py` returns exactly that). The TypeError is then swallowed by the bare `except:` in `schedule_job`, so `_cancel_resources` never runs (reservations already granted by healthy sites leak), schedule count/history never update (no backoff, `max_schedule_count` never trips), and no reason is surfaced — the exact failure mode this PR is fixing. Reproduced at this PR's head; note `test_require_sites_not_enough_resource` currently passes *via* the swallowed TypeError rather than the intended return path.

```suggestion
                    no_resource_message += site_name + ":" + (token or "") + ";"
```

Worth adding a test for the required-site + `None`-token path, since the existing test masks it.

nvflare/private/fed/client/scheduler_cmds.py
```diff
@@ -80,7 +80,10 @@ def process(self, req: Message, app_ctx) -> Message:
                     is_resource_enough, token = resource_manager.check_resources(
                         resource_requirement=resource_spec, fl_ctx=fl_ctx
                     )
-            except Exception:
+            except Exception as e:
+                reason = f"resource check failed: {secure_format_exception(e)}"
+                engine.logger.error(f"Job {job_id}: {reason}")
```
holgerroth, line 85: `logger` isn't part of the `ClientEngineInternalSpec` contract this processor validates `engine` against — only the concrete `ClientEngine` happens to set `self.logger` (which is why the new test has to graft `engine.logger = MagicMock()` onto the spec'd mock). If an engine without that attribute ever reaches this handler, the `AttributeError` fires *before* `token = reason`, so the one path meant to surface the error would lose it (FedAdminAgent would reply with a generic `exception_occurred: AttributeError...` instead). Suggest a module-level logger, e.g. `get_module_logger`/`get_obj_logger` from `nvflare.fuel.utils.log_utils` as used elsewhere in this package — or at minimum assign `token = reason` before the log call.

holgerroth (commented):
A few comments on the error-surfacing path (details inline) — all should be small fixes.

tests/unit_test/private/fed/client/scheduler_cmds_test.py
```diff
@@ -0,0 +1,45 @@
...
+    engine.get_component.return_value = resource_manager
+    engine.new_context.return_value.__enter__.return_value = FLContext()
+    engine.logger = MagicMock()
+
+    request = Message(topic=TrainingTopic.CHECK_RESOURCE, body={"license": 2})
+    request.set_header(RequestHeader.JOB_ID, "job-1")
+
+    reply = CheckResourceProcessor().process(request, engine)
+
+    reason = "resource check failed: ValueError: resource_requirement is missing num_gpu_key num_of_gpus"
```
holgerroth, line 40: This exact string bakes in `secure_format_exception`'s non-secure format. If `NVFLARE_SECURE_LOGGING` is set in the environment, it returns `"<class 'ValueError'>"` instead of `"ValueError: ..."` and the assertions below fail even though the code behaved correctly (reproduced locally). Suggest making the test hermetic with `monkeypatch.delenv("NVFLARE_SECURE_LOGGING", raising=False)`, or asserting on a stable substring such as `"resource check failed:"`.

tests/unit_test/private/fed/client/scheduler_cmds_test.py
```diff
@@ -0,0 +1,45 @@
...
+    assert reply.body.get_return_code() == ReturnCode.EXECUTION_EXCEPTION
```
holgerroth, line 41: This pins a return code the server never reads: `ServerEngine.check_client_resources` only checks `MsgHeader.RETURN_CODE` on the admin `Message` (set by `error_reply`, never by this processor) and never calls `get_return_code()` on the body `Shareable` — the error actually travels via the `RESOURCE_RESERVE_TOKEN` header alone. Asserting it here makes the dead channel look load-bearing and invites a later "cleanup" that relies on it (re-hiding these errors). Either drop this assertion, or make the return code real — e.g. don't catch here at all and let `FedAdminAgent`'s `error_reply` wrapper carry the message, which the server's existing non-OK branch already converts to `(False, message)`.

PR #5145, opened 2026-08-17
Fix Lightning best-model metric reporting

chesterxgchen:
## Summary

- Forward metrics captured by an explicit PyTorch Lightning `trainer.validate()` call before `trainer.fit()`, even when `train_with_evaluation=False`.
- Keep `train_with_evaluation=True` as a requirement that pre-training validation metrics exist; when it is `False`, metrics are optional rather than suppressed.
- Exclude Lightning sanity-check and in-fit validation metrics from global-model selection.
- Keep the hello-lightning example's federated metric contract framework-neutral by reporting epoch-level validation accuracy as `accuracy`, matching the recipe default.
- Add regression coverage and document the corrected behavior in the 2.9 release notes.

## Behavior

| Client workflow | Training result |
| --- | --- |
| Explicit `validate()` before `fit()`, flag `False` | Captured metrics are returned |
| Explicit `validate()` before `fit()`, flag `True` | Captured metrics are returned |
| `fit()` without explicit pre-fit validation, flag `False` | Metrics may be absent |
| `fit()` without explicit pre-fit validation, flag `True` | Fails because required metrics are missing |
| Lightning sanity or mid-fit validation | Never treated as the received global model's score |

## Implementation

- `FLCallback` accepts validation metrics for model selection only while Lightning is in the standalone `TrainerFn.VALIDATING` state and before the training round starts. During sanity checks and in-fit validation, Lightning remains in `TrainerFn.FITTING`, so those results are ignored.
- `on_train_end()` now attaches captured metrics whenever they exist. The `train_with_evaluation` flag only determines whether their absence is an error.
- Metric names are not implicitly remapped by NVFlare. Applications may use custom metric names, but their recipe `key_metric` must match. The hello-lightning example emits the semantic key `accuracy` directly.

This allows `IntimeModelSelector` to receive the intended pre-training global-model score and persist the best global model without confusing it with local-model validation performed during training.

## Validation

- `python3 -m pytest tests/unit_test/tool/autofl_skill_job_importer_test.py -q` (46 passed)
- `git diff --check`
- PR CI triggered with `/build`

### DIFF
--- a/examples/hello-world/hello-lightning/model.py
+++ b/examples/hello-world/hello-lightning/model.py
@@ -76,7 +76,8 @@ class LitNet(LightningModule):
 
         if stage:
             self.log(f"{stage}_loss", loss)
-            self.log(f"{stage}_acc", self.valid_acc, on_step=True, on_epoch=True)
+            # Keep the federated selection metric stable and framework-neutral.
+            self.log("accuracy", self.valid_acc, on_step=False, on_epoch=True)
         return outputs
 
     def validation_step(self, batch, batch_idx):
--- a/examples/hello-world/hello-lightning/model.py
+++ b/examples/hello-world/hello-lightning/model.py
@@ -76,8 +76,11 @@ class LitNet(LightningModule):
 
         if stage:
             self.log(f"{stage}_loss", loss)
-            # Keep the federated selection metric stable and framework-neutral.
-            self.log("accuracy", self.valid_acc, on_step=False, on_epoch=True)
+            if stage == "val":
+                # Match the recipes' default key_metric for federated model selection.
+                self.log("accuracy", self.valid_acc, on_step=False, on_epoch=True)
+            else:
+                self.log(f"{stage}_acc", self.valid_acc, on_step=True, on_epoch=True)
         return outputs
 
     def validation_step(self, batch, batch_idx):
--- a/examples/hello-world/hello-lightning/model.py
+++ b/examples/hello-world/hello-lightning/model.py
@@ -68,19 +68,19 @@ class LitNet(LightningModule):
         self.log("train_acc", self.train_acc, on_step=True, on_epoch=False)
         return loss
 
-    def evaluate(self, batch, stage=None):
+    def evaluate(self, batch, evaluation_mode=None):
         x, labels = batch
         outputs = self(x)
         loss = criterion(outputs, labels)
         self.valid_acc(outputs, labels)
 
-        if stage:
-            self.log(f"{stage}_loss", loss)
-            if stage == "val":
+        if evaluation_mode:
+            self.log(f"{evaluation_mode}_loss", loss)
+            if evaluation_mode == "val":
                 # Match the recipes' default key_metric for federated model selection.
                 self.log("accuracy", self.valid_acc, on_step=False, on_epoch=True)
             else:
-                self.log(f"{stage}_acc", self.valid_acc, on_step=True, on_epoch=True)
+                self.log(f"{evaluation_mode}_acc", self.valid_acc, on_step=True, on_epoch=True)
         return outputs
 
     def validation_step(self, batch, batch_idx):
--- a/docs/release_notes/flare_290.rst
+++ b/docs/release_notes/flare_290.rst
@@ -55,6 +55,17 @@ Compatibility and Migration Notes
   multi-optimizer FedAvg client reports their combined step count unless it
   supplies ``NUM_STEPS_CURRENT_ROUND`` explicitly; explicit client metadata is
   still preserved.
+- Patched PyTorch Lightning clients now transmit validation metrics captured by
+  an explicit ``trainer.validate()`` call before ``trainer.fit()``, regardless
+  of ``train_with_evaluation``. Sanity-check and in-fit validation metrics are
+  not transmitted as global-model scores. Setting
+  ``train_with_evaluation=True`` continues to require validation metrics;
+  otherwise metrics remain optional, but ``False`` no longer suppresses metrics
+  from an explicit pre-fit validation. This enables ``IntimeModelSelector`` and
+  best-global-model persistence for recipe-based Lightning jobs. Applications
+  that must keep such metrics local should omit the explicit pre-fit validation;
+  if they still require it locally, a custom task-result filter must remove
+  ``MetaKey.INITIAL_METRICS`` before the result reaches the server.
 - Recipe discovery now exposes the concrete PyTorch ``FedProxRecipe`` as
   ``fedprox-pt`` and no longer advertises the ``fedprox-tf`` manual pattern as
   a concrete recipe. TensorFlow clients can continue to combine a FedAvg
--- a/examples/hello-world/hello-lightning/job.py
+++ b/examples/hello-world/hello-lightning/job.py
@@ -71,6 +71,7 @@ def main():
             # For pre-trained weights: initial_ckpt="/server/path/to/pretrained.pt",
             train_script="client.py",
             train_args=train_args,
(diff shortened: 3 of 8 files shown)


YuanTingHsieh (dismissed):

YuanTingHsieh (dismissed):

holgerroth:
Did a deeper pass on the end-to-end metric flow this change enables (client capture → `to_shareable` → server consumers). The direction is right — it aligns Lightning with the HF integration and unblocks `IntimeModelSelector` for the validate-then-fit pattern. Three things are worth addressing before merge; the first is the substantive one.

### 1. Gate what gets *captured*, not only what gets sent

With `train_with_evaluation=False` — the effective default for recipe/ScriptRunner jobs, since ScriptRunner never sets it and `ClientAPIExecutor` defaults to `False` (`client_api_executor.py:85`) — a fit-only script now transmits metrics that don't represent the received global model, and the server treats them as exactly that: `FLModelUtils.to_shareable` labels train-result metrics as `MetaKey.INITIAL_METRICS` ("treated as evaluation on the global model", `fl_model_utils.py:68-69`), which `IntimeModelSelector` and best-model persistence consume.

Both cases reproduce at this PR's head (PL 2.6.x):

- **Fit-only script with `val_dataloaders`**: the first in-fit validation runs *after* local training, and first-capture-wins in `on_validation_end` pins it. Repro sent `val_loss=0.00041` (post-training local model) while the received global model scored `9.99` on the same val set — the selector persists the wrong "best" checkpoint every round.
- **Round-1 sanity check**: `on_validation_end` fires during `sanity_checking` with 2-batch metrics and pins them (`reset_state`'s `num_sanity_val_steps = 0` at `api.py:177` only helps a reused trainer, so with `launch_once=False` this recurs every round). Repro: transmitted `val_loss=1.286` (2-batch sanity value) vs `0.739` from the real validation in the same fit.

The HF integration already solved this: it also transmits unconditionally, but restricts *capture* to pre-train evaluation of the received model (`_inside_train` guard, `hf/api.py:776-785`). Suggest mirroring that here — skip capture when `trainer.sanity_checking`, and only attach metrics captured before the training round started (i.e. the validate-before-fit pattern this PR's release note assumes). That keeps the intended fix fully working while cutting off the mis-attribution. It also matters beyond the selector: pre-PR, `result.metrics is None` kept FedAvg's `stop_cond`/patience inert (`fedavg.py:410`); with mis-attributed metrics flowing, a configured `stop_cond` can now stop a job on sanity noise or post-training local values.

### 2. Release-note wording (`flare_290.rst:58-63`)

- *"This restores `IntimeModelSelector` and best-global-model persistence for recipe-based Lightning jobs"* — nothing is restored: `train_with_evaluation` defaulted to `False` in 2.7/2.8 and ScriptRunner never plumbed it, so recipe Lightning jobs have never transmitted train-round metrics. This bullet enables the behavior for the first time; "restores" will make 2.8 operators misclassify it as a return to known behavior and skip the review a new-behavior note would trigger.
- *"No application changes are required."* — for sites that relied on `train_with_evaluation=False` keeping validation metrics on-site (the `ClientAPIExecutor` docstring still documents the flag as controlling exactly that), this is new egress with no opt-out knob. Note the metrics travel in DXO **meta**, which the built-in output filters don't touch, so restoring the old behavior requires a custom task-result filter. The bullet should state the transmission change plainly and how to opt out.

### 3. Error message (`api.py:237`)

The raise says "missing **training** metrics" but what's missing is **validation** metrics (captured via `trainer.validate()` / `on_validation_end`) — the release note itself correctly says "validation metrics". A user hitting this will go debug `training_step` logging instead of adding `validate()`. Since the line is rewritten here anyway and the new test pins the string via `match=`, cheapest to fix both together. (Same phrasing in `hf/api.py:791` can be a follow-up.)

---

**Non-blocking, if you're touching these files anyway:** the `train_with_evaluation` docstring in `client_api_executor.py:143` still describes transmission gating and could be updated to the new semantics; and the new test could assert the `to_shareable` round-trip (`INITIAL_METRICS`) like the neighboring test does, so the delivery mechanism the release note references is what's actually verified.

A few pre-existing issues surfaced in the process (`__fl_meta__[MetaKey.INITIAL_METRICS]` silently overriding freshly captured metrics via `dxo.meta.update` ordering in `to_shareable`; `SimpleIntimeModelSelector` doing `dict * float` → `TypeError` on dict metrics; empty-`{}` metrics satisfying the `is None` guard; DDP shard-only metrics without `sync_dist`) — those can be tracked separately and shouldn't hold this PR.

chesterxgchen:
Thanks for the deeper end-to-end review. Addressed in `96e5aeb58`:

1. Metric capture is now limited to explicit pre-fit validation. Lightning exposes the same boundary as HF through trainer lifecycle state: a direct `trainer.validate()` runs under `TrainerFn.VALIDATING`, while both fit sanity checks and in-fit validation remain `TrainerFn.FITTING`. The callback now captures only the former and only before the training round starts. Unit coverage rejects sanity/in-fit metrics, and a real-Lightning fit-only test with sanity plus epoch validation asserts that both returned models have `metrics is None`.
2. The release note now says this *enables* recipe-based selection, states that `train_with_evaluation=False` no longer suppresses an explicitly captured pre-fit metric, and documents how applications can keep such metrics local.
3. The Lightning error now says validation metrics are required and tells users to call `validate` before `fit`.

Also addressed the non-blocking suggestions: the executor docstring now defines the flag as required-vs-optional rather than a transmission gate, and the regression test verifies the `to_shareable`/`from_shareable` round trip seen by the server.

Validation after the final lifecycle guard: `tests/unit_test/app_opt/lightning/api_test.py` (91 passed), `./runtest.sh -s`, and `git diff --check`.

holgerroth (dismissed):
The updated revision addresses the review concerns: metric capture is now gated to explicit pre-fit validation via trainer.state.fn (sanity-check and in-fit validation metrics no longer transmit as global-model scores), the release note accurately describes the new behavior including the opt-out path, and the error message/docstring now match the actual semantics. Verified the new regression tests cover both previously reproduced failure modes; full Lightning API suite passes locally (91 passed, PL 2.6.1). Remaining minor items (HF message wording, stale example comments, SimpleIntimeModelSelector dict handling) are tracked as follow-ups.

examples/hello-world/hello-lightning/model.py
```diff
@@ -76,7 +76,11 @@ def evaluate(self, batch, stage=None):
 
         if stage:
             self.log(f"{stage}_loss", loss)
-            self.log(f"{stage}_acc", self.valid_acc, on_step=True, on_epoch=True)
+            if stage == "val":
+                # Match the recipes' default key_metric for federated model selection.
```
YuanTingHsieh, line 80: Non-blocking: `stage` predates this PR, but this new special case makes the name easier to confuse with Lightning's own `trainer.state.stage`/`TrainerFn`. Here it is only a caller-supplied selector (`"val"`, `"test"`, or `None`) that controls metric logging. Consider renaming it to something like `evaluation_mode`, or moving the stage-specific logging into `validation_step()` and `test_step()`, in a follow-up. This does not need to hold up this PR.

YuanTingHsieh (commented):
Non-blocking readability suggestion.

examples/hello-world/hello-lightning/model.py
chesterxgchen: Addressed in b9eee411f: renamed the caller-supplied selector from stage to evaluation_mode throughout LitNet.evaluate(), avoiding confusion with Lightning lifecycle state. The real-example regression still verifies validation accuracy and test_acc_epoch.

YuanTingHsieh (approved):

PR #5146, opened 2026-08-17
Bind localhost TCP listeners to loopback

YuanTingHsieh:
## Summary

- bind TCP listeners to explicit IPv4 loopback when the advertised host defaults to or is configured as `localhost`
- preserve explicit `listen_host` values
- preserve the wildcard listener fallback for non-local advertised hosts

## Root cause

`get_tcp_urls()` advertised `localhost` to local child cells but independently defaulted the listening URL to `0`, which bound the socket on all interfaces. As a result, generic process-mode parent listeners could expose a clear internal CellNet port beyond loopback even though children were instructed to connect locally.

## Impact

Default local process listeners now bind to `127.0.0.1`. Kubernetes, Slurm, Docker, and other non-local configurations retain their existing explicit or wildcard bind behavior.

## Validation

- `python -m pytest -q tests/unit_test/fuel/f3/drivers/net_utils_test.py tests/unit_test/fuel/f3/cellnet/core_cell_test.py tests/unit_test/private/fed/server/fed_server_test.py tests/unit_test/private/fed/app/deployer/simulator_deployer_test.py` — 76 passed
- `git diff --check`

### DIFF
--- a/nvflare/fuel/f3/drivers/net_utils.py
+++ b/nvflare/fuel/f3/drivers/net_utils.py
@@ -292,7 +292,7 @@ def get_tcp_urls(scheme: str, resources: dict) -> (str, str):
 
     listening_host = resources.get(DriverParams.LISTEN_HOST.value) if resources else None
     if not listening_host:
-        listening_host = "0"
+        listening_host = "127.0.0.1" if host.rstrip(".").lower() == "localhost" else "0"
 
     port = get_open_tcp_port(resources)
     if not port:
--- a/nvflare/fuel/f3/drivers/net_utils.py
+++ b/nvflare/fuel/f3/drivers/net_utils.py
@@ -11,6 +11,7 @@
 # WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
 # See the License for the specific language governing permissions and
 # limitations under the License.
+import ipaddress
 import logging
 import os
 import random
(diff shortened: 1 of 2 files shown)


nvflare/fuel/f3/drivers/net_utils.py
```diff
@@ -292,7 +292,7 @@ def get_tcp_urls(scheme: str, resources: dict) -> (str, str):
 
     listening_host = resources.get(DriverParams.LISTEN_HOST.value) if resources else None
     if not listening_host:
-        listening_host = "0"
+        listening_host = "127.0.0.1" if host.rstrip(".").lower() == "localhost" else "0"
```
pcnudde: **[P1] Treat IPv4 loopback literals as local too.** This condition only recognizes the normalized hostname `localhost`; `get_tcp_urls("tcp", {"host": "127.0.0.1"})` still advertises `127.0.0.1` but emits a listening URL on `0`, exposing the clear internal port on every interface despite the loopback connect address. Please recognize supported IPv4 loopback literals (at minimum `127.0.0.1`), bind the literal address, and add a regression test without an explicit `listen_host`. Do not silently include `::1` until F3 URL parsing and the TCP driver support IPv6 end to end.

pcnudde (commented):
One blocking listener-containment finding.

nvidianz:
Does this PR have the same change? https://github.com/NVIDIA/NVFlare/pull/5122

nvflare/fuel/f3/drivers/net_utils.py
nvidianz: There is also IPv6 loopback addresses. If we want to do this right for all cases, we should check

`ipaddress.ip_address(host).is_loopback`

YuanTingHsieh: Addressed in 74943651d. get_tcp_urls now parses non-localhost hosts with ipaddress.ip_address() and binds the original literal only when it is an IPv4Address with is_loopback=true, covering the full 127.0.0.0/8 range. The IPv4 type gate deliberately excludes ::1 until F3 URL parsing and the TCP driver support IPv6 end to end. Regression coverage includes 127.0.0.1 and 127.42.0.9 without listen_host, plus an explicit ::1 non-support case. The focused suite passes 13/13 and the full project style check passes.

nvidianz (approved):

PR #5148, opened 2026-08-17
Keep resource manager exceptions site-local

pcnudde:
## Description

Follow-up to #5143.

The resource-check exception path returned `secure_format_exception(e)` to the server. Unless secure logging was enabled at the client, that included `str(e)`, allowing site-local exception details to enter scheduling history and job metadata.

This change keeps the formatted exception in the site-local log and returns the fixed, bounded diagnostic `resource manager raised an exception; see site log` across the site boundary. It also uses a module logger instead of relying on an undeclared `engine.logger` attribute.

Resource admission and scheduling behavior are otherwise unchanged.

## Validation

- 22 focused scheduler and resource-check tests passed
- `./runtest.sh -s` passed
- `git diff --check` passed

### DIFF
--- a/nvflare/private/fed/client/scheduler_cmds.py
+++ b/nvflare/private/fed/client/scheduler_cmds.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import json
+import logging
 from typing import List
 
 from nvflare.apis.event_type import EventType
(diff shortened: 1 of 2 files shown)


holgerroth (approved):
LGTM. Verified the exception detail now stays site-local (fixed bounded token across the boundary) and the updated test is hermetic — passes with and without NVFLARE_SECURE_LOGGING set. Remaining items from the #5143 review (the None-token guard in job_scheduler.py and the dead EXECUTION_EXCEPTION assertion) can be handled in a separate follow-up.

PR #5149, opened 2026-08-17
Harden resource admission failure handling

pcnudde:
## Description

- tolerate resource managers returning no reservation token for a required-site rejection
- keep the normal cancellation, scheduling-history, retry-backoff, and maximum-attempt path intact
- remove the unused return code from the resource-check response body
- strengthen the existing required-site test so it cannot pass through the scheduler's broad exception handler

## Root cause

The required-site error message concatenated the reservation token without handling `None`. A resource manager returning `(False, None)` raised `TypeError` before cancellation and scheduling bookkeeping could run. Separately, the client set a return code inside the response `Shareable`, but the server reads only the outer admin message return code and the resource token.

## Impact

Required-site resource rejection now follows the intended no-resource path even when no token is returned. Resource-manager exception details remain site-local, and the cross-site diagnostic behavior introduced in #5148 is unchanged.

## Validation

- `22 passed`: focused scheduler and client resource-check unit tests
- `./runtest.sh -s`
- `git diff --check`

### DIFF
--- a/nvflare/app_common/job_schedulers/job_scheduler.py
+++ b/nvflare/app_common/job_schedulers/job_scheduler.py
@@ -206,7 +206,7 @@ class DefaultJobScheduler(JobSchedulerSpec, FLComponent):
                     required_sites_not_enough_resource.remove(site_name)
             else:
                 if site_name in required_sites:
-                    no_resource_message += site_name + ":" + token + ";"
+                    no_resource_message += site_name + ":" + (token or "") + ";"
                 if token:
                     resource_failure_details.append(f"{site_name}: {token}")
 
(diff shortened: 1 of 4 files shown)


YuanTingHsieh (approved):
Verified the fix red/green: the strengthened test reproduces the swallowed `TypeError` against the old code, and all 22 focused tests pass at the PR head. I also traced the removed inner `Shareable` return code through the request path; the server consumes the outer admin-message status and resource headers, so the removal is behavior-neutral. The fix restores cancellation and scheduling bookkeeping for tokenless required-site rejection. No blocking findings.

PR #5150, opened 2026-08-17
Handle dict metrics in SimpleIntimeModelSelector

holgerroth:
## Summary

`SimpleIntimeModelSelector` (`nvflare/app_common/ccwf/comps/simple_intime_model_selector.py`) computes `validation_metric * weight` on the raw `INITIAL_METRICS` meta value with no type handling. When the metrics payload is a dict — which is what framework integrations attach (`FLModelUtils.to_shareable` places the whole `FLModel.metrics` dict under `MetaKey.INITIAL_METRICS`, and the patched Lightning/HF clients always produce dicts) — every `BEFORE_CONTRIBUTION_ACCEPT` raises `TypeError: unsupported operand type(s) for *: 'dict' and 'float'`. `fire_event` swallows and logs the exception per contribution, so best-model selection is silently disabled with per-round exception spam.

Surfaced while reviewing #5145, which widens how often dict-valued metrics reach the server; the sibling `app_common.widgets.IntimeModelSelector` already handles this case.

## Changes

- Mirror the widget selector's dict handling: add a `key_metric` argument (default `"val_accuracy"`, matching `IntimeModelSelector`) used to select the metric when the payload is a dict, and warn-skip contributions whose dict lacks the key.
- Add unit tests covering scalar payloads (unchanged behavior), dict payloads with the default and a custom `key_metric`, and the missing-key skip path.

## Validation

- `python3 -m pytest tests/unit_test/app_common/ccwf/comps/simple_intime_model_selector_test.py -v` (4 passed)
- `python3 -m pytest tests/unit_test/app_common/ccwf/ tests/unit_test/app_common/widgets/in_time_model_selector_test.py` (158 passed; 4 pre-existing environment-dependent errors in `test_swarm_self_message_deadlock.py` from sandboxed socket binding, unrelated)
- `black`/`isort`/`flake8` clean on both files


nvflare/app_common/ccwf/comps/simple_intime_model_selector.py
```diff
@@ -98,8 +105,19 @@ def _before_accept(self, fl_ctx: FLContext):
         if validation_metric is None:
             self.log_debug(fl_ctx, f"validation metric not existing in {client_name}")
             return False
-        else:
-            self.log_info(fl_ctx, f"validation metric {validation_metric} from client {client_name}")
+
+        # select key metric if a dictionary of metrics is provided
+        if isinstance(validation_metric, dict):
+            if self.key_metric in validation_metric:
+                validation_metric = validation_metric[self.key_metric]
```
YuanTingHsieh, line 129: The selected value isn't validated as numeric, so a payload like `{"val_accuracy": None}` (or a str/list/nested-dict value) reproduces the same swallowed TypeError this PR fixes, just below at `validation_metric * weight`. A numpy array is worse: the multiply succeeds and poisons the accumulator, then `val_metric > best_val_metric` in `_before_aggregate` raises, which skips `_reset_stats()` and leaves stale sums until the next `ROUND_STARTED`.

Suggest coercing after the dict branch (covers scalar payloads too):

```python
try:
    validation_metric = float(validation_metric)
except (TypeError, ValueError):
    self.log_warning(
        fl_ctx, f"validation metric {validation_metric!r} from {client_name} is not a number; skipping"
    )
    return False
```

nvflare/app_common/ccwf/comps/simple_intime_model_selector.py
```diff
@@ -25,20 +25,27 @@
 
 class SimpleIntimeModelSelector(Widget):
     def __init__(
-        self, weigh_by_local_iter=False, aggregation_weights=None, validation_metric_name=MetaKey.INITIAL_METRICS
+        self,
+        weigh_by_local_iter=False,
+        aggregation_weights=None,
+        validation_metric_name=MetaKey.INITIAL_METRICS,
+        key_metric: str = "val_accuracy",
```
YuanTingHsieh, line 33: `widgets.IntimeModelSelector` pairs this with `negate_key_metric` plus a constructor warning when the key name looks lower-is-better (loss/err heuristic). Without either, `key_metric="val_loss"` here silently persists and distributes the *worst* model — the `>` comparison assumes higher-is-better unconditionally. Please port at least the constructor warning; porting `negate_key_metric` too would make the two selectors actually equivalent.

nvflare/app_common/ccwf/comps/simple_intime_model_selector.py
```diff
@@ -25,20 +25,27 @@
...
     ):
         """Handler to determine if the model is globally best.
 
         Args:
             weigh_by_local_iter (bool, optional): whether the metrics should be weighted by trainer's iteration number.
             aggregation_weights (dict, optional): a mapping of client name to float for aggregation. Defaults to None.
             validation_metric_name (str, optional): key used to save initial validation metric in the DXO meta properties (defaults to MetaKey.INITIAL_METRICS).
+            key_metric: if metrics are a `dict`, `key_metric` can select the metric used for global model selection.
+                Defaults to "val_accuracy". Higher values are treated as better.
```
YuanTingHsieh: Worth calling this out more loudly: unlike the widget under `BaseFedJob`, nothing in the CCWF recipe path plumbs `key_metric`, so this default is load-bearing — and common Client API payloads report `"accuracy"`, not `"val_accuracy"`. For those jobs the new dict support warn-skips every contribution, which looks a lot like today's failure mode. Suggest documenting that the key usually needs to be configured explicitly, and adding a release note for the behavior flip: jobs whose metric dicts *do* contain `val_accuracy` go from selection-dead to BEST-model persistence/distribution active with zero config change.

tests/unit_test/app_common/ccwf/comps/simple_intime_model_selector_test.py
```diff
@@ -0,0 +1,102 @@
...
+    dxo = DXO(
+        DataKind.WEIGHTS,
+        data=dict(),
+        meta={MetaKey.INITIAL_METRICS: metric_value},
+    )
+    shareable = dxo.to_shareable()
+    shareable.add_cookie(AppConstants.CONTRIBUTION_ROUND, current_round)
+
+    peer_ctx = FLContext()
+    peer_ctx.set_prop(FLContextKey.CLIENT_NAME, client_name, private=False)
```
YuanTingHsieh: The selector reads `peer_ctx.get_identity_name(default="?")`, not `CLIENT_NAME`, so `client_name` is always `"?"` in these tests and per-client `aggregation_weights` can't be exercised with this fixture. Use `peer_ctx.set_prop(FLContextKey.IDENTITY_NAME, client_name, private=False)` instead.

tests/unit_test/app_common/ccwf/comps/simple_intime_model_selector_test.py
```diff
@@ -0,0 +1,102 @@
...
+
+    def test_dict_metrics_missing_key_metric_skips_contribution(self):
+        handler = SimpleIntimeModelSelector()
+        engine = MockSimpleEngine()
+
+        fl_ctx = _make_contribution_ctx(engine, {"val_loss": 0.1})
+        handler.handle_event(AppEventType.BEFORE_CONTRIBUTION_ACCEPT, fl_ctx)
+        handler.handle_event(AppEventType.BEFORE_AGGREGATION, fl_ctx)
+
+        assert handler.validation_metric_sum_of_weights == 0
```
YuanTingHsieh, line 116: `_before_aggregate` ends with `_reset_stats()` on the accepted path too, so this holds after `BEFORE_AGGREGATION` regardless of whether the contribution was skipped — the assertion is vacuous as written. Assert it right after `BEFORE_CONTRIBUTION_ACCEPT` (before firing `BEFORE_AGGREGATION`) to actually pin the skip.

YuanTingHsieh (commented):
Direction is right — consumer-side selection mirroring the widget — and this strictly improves on main: dict payloads carrying the default key go from selection-dead TypeError spam to working best-model selection. Three asks below before merge (numeric coercion, lower-is-better handling, default-key documentation), plus two test notes.

FYI, two pre-existing limitations worth knowing while touching this component (not asks for this PR): the selector is a total no-op under `ModelController`/FedAvg because `BEFORE_CONTRIBUTION_ACCEPT` fires before `TRAINING_RESULT` is set (only SAG and swarm set the prop first), and a "skipped" contribution is still aggregated by the Gatherer — the skip only excludes it from the advertised best metric, so that metric is computed over a subset of what's actually in the model.

nvflare/app_common/ccwf/comps/simple_intime_model_selector.py
holgerroth: Done in 3359dbc11 — adopted the coercion as suggested, placed after the dict branch so scalar payloads are covered too. Verified the failure modes first: `float(None)` / `float(str)` / `float(multi-element ndarray)` raise TypeError/ValueError and now warn-skip; previously a multi-element array poisoned the accumulator and the ambiguous-truth ValueError in `_before_aggregate` indeed skipped `_reset_stats()`. Added parametrized tests for `None` / `"not-a-number"` / `[0.1, 0.2]`.

nvflare/app_common/ccwf/comps/simple_intime_model_selector.py
holgerroth: Done in 3359dbc11 — ported both `negate_key_metric` and the constructor warning. I reused `_looks_lower_is_better` from `widgets.intime_model_selector` rather than duplicating the heuristic (single source of truth); if importing a private helper across modules bothers you, I can promote it to a shared util in a follow-up. Negation mirrors the widget: applied after extraction/coercion, before accumulation. Tests cover the negate path and the warning matrix (`val_loss`/False warns; `val_loss`/True and `val_accuracy`/False stay quiet).

nvflare/app_common/ccwf/comps/simple_intime_model_selector.py
holgerroth: Done in 3359dbc11 — the `key_metric` docstring now says the CCWF job path does not plumb it, that clients commonly report other names (e.g. `accuracy`), and that non-matching dict contributions are warn-skipped. Kept the default at `val_accuracy` for parity with `widgets.IntimeModelSelector`. Also added the `flare_290.rst` bullet for the behavior flip: dicts containing the configured key go from selection-dead (logged TypeError) to active best-global-model tracking with no config change; missing keys and non-numeric values warn-skip; `negate_key_metric` covers losses.

tests/unit_test/app_common/ccwf/comps/simple_intime_model_selector_test.py
holgerroth: Done in 3359dbc11, with one correction to the suggestion: `FLContextKey` has no `IDENTITY_NAME` — `get_identity_name` reads `ReservedKey.IDENTITY_NAME` (fl_context.py:236), which is also what production sets (fl_context.py:349) — so the fixture now uses `peer_ctx.set_prop(ReservedKey.IDENTITY_NAME, client_name, private=False, sticky=False)`. Added `test_aggregation_weights_use_client_identity` to prove per-client weights are now actually exercised (3:1 weights over metrics 1.0/0.0 → 0.75 weighted average; the old fixture resolved every client to "?" and weight 1.0). The widget selector's test fixture has the same `CLIENT_NAME` flaw — leaving that for a separate cleanup.

tests/unit_test/app_common/ccwf/comps/simple_intime_model_selector_test.py
holgerroth: Done in 3359dbc11 — you're right, `_before_aggregate` resets the sums on the accepted path too, so the assertion held either way (the `last_event != GLOBAL_BEST_MODEL_AVAILABLE` check was doing the real work). The sum assertion now sits directly after `BEFORE_CONTRIBUTION_ACCEPT` with a comment explaining why, and the new non-numeric skip tests follow the same structure.

holgerroth:
All five inline comments addressed in 3359dbc11 — numeric coercion, `negate_key_metric` + lower-is-better warning ported from the widget selector, docstring + release-note bullet for the default-key caveat and behavior flip, and both test fixes. Validation: 12/12 tests in the new file pass, ccwf + widget-selector suites pass (166 passed; the 4 errors in `test_swarm_self_message_deadlock.py` are pre-existing sandbox socket failures), black/isort/flake8 clean.

On the two FYIs: agreed both are pre-existing and out of scope here. The `ModelController`/FedAvg no-op matches what we found while reviewing #5145 (only SAG and swarm set `TRAINING_RESULT` before firing `BEFORE_CONTRIBUTION_ACCEPT`), and the Gatherer-still-aggregates-skipped-contributions point is a good candidate for a follow-up issue on the ccwf selector semantics.

YuanTingHsieh (approved):

PR #5151, opened 2026-08-17
Clarify train_with_evaluation example comments

chesterxgchen:
## Summary
- update the duplicated Client API task comments in 19 examples/tutorials and 2 integration fixtures
- clarify that supplied evaluation metrics are transmitted regardless of train_with_evaluation
- clarify that train_with_evaluation=True makes evaluation metrics required rather than enabling transmission

This is a comments-only follow-up to #5145. It intentionally excludes the Hugging Face diagnostic wording and SimpleIntimeModelSelector behavior.

## Validation
- ./runtest.sh -s
- git diff --check
- verified all 21 stale comment blocks were replaced

### DIFF
--- a/examples/advanced/experiment-tracking/mlflow/hello-pt-mlflow-client/client.py
+++ b/examples/advanced/experiment-tracking/mlflow/hello-pt-mlflow-client/client.py
@@ -104,8 +104,9 @@ def main():
 
         # Handle each Client API task according to its type:
         # - "train" (flare.is_train()): train and optionally evaluate the received model, then send the updated
-        #   model plus any supplied evaluation metrics. Setting "train_with_evaluation" to True in config_fed_client
-        #   makes evaluation metrics required; it does not enable their transmission.
+        #   model plus any supplied evaluation metrics. When "train_with_evaluation" is True, framework integrations
+        #   such as Lightning and Hugging Face require evaluation metrics; the raw Client API does not enforce them.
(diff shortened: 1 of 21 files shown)


YuanTingHsieh:
The replacement text fixes the real issue: metrics supplied in the FLModel are always transmitted, and the flag never gated that.

One accuracy nit on the new wording: "makes evaluation metrics required" is only enforced by the framework integrations (`nvflare/app_opt/lightning/api.py:237`, `nvflare/app_opt/hf/api.py:790`). The raw Client API path these examples use has never enforced it — the executor and backends just transport the flag. In the recipe-based examples here the flag isn't even exposed (recipes/ScriptRunner have no such param); only the two integration fixtures set it via `config_fed_client.conf`. So a raw script that sets the flag True and returns no metrics will not error, which contradicts "required".

Suggested tweak:

> Setting "train_with_evaluation" to True in config_fed_client makes evaluation metrics required (enforced by framework integrations such as Lightning and Hugging Face); it does not control whether supplied metrics are transmitted.

Not blocking — this matches the executor docstring framing from #5145. LGTM otherwise.

chesterxgchen:
@YuanTingHsieh Addressed in 4300e7bdf. All 21 comment blocks now state that Lightning and Hugging Face enforce the metric requirement, while the raw Client API does not; they separately clarify that the flag never controls whether supplied metrics are transmitted. The full style gate and diff checks pass.

YuanTingHsieh (approved):

PR #5153, opened 2026-08-18
Remove stale package version fallback

nvkevlu:
Removes stale package version fallback.

### Description

Removes stale package version fallback.

### Types of changes

- [ ] Documentation updated.


YuanTingHsieh:
Do we need to support building directly from a source tree with no usable Git/Versioneer metadata? Versioneer handles normal Git checkouts and embeds the resolved version in official sdists/wheels. The checked-in CI builds from Git checkouts, the Docker builds assert that Versioneer succeeds, and no in-repo workflow sets `NVFL_BASE_VERSION`.

Instead of maintaining another hard-coded release version, should we keep `NVFL_BASE_VERSION` as the explicit escape hatch and fail clearly when both Versioneer metadata and that override are unavailable? Is there an external/nightly packaging pipeline that relies on the current implicit fallback?

YuanTingHsieh (approved):

PR #5155, opened 2026-08-18
Fix POC tutorial CLI commands

nvkevlu:
Fix POC tutorial CLI commands.

### Description

This PR updates the POC tutorial notebooks to match the current NVFlare CLI.

- Replace unsupported `nvflare poc status` with `nvflare system status`.
- Correct examples that repeat scalar `-p` and `-ex` options.
- Align the documented default start behavior: server and clients start by default; admin consoles are explicit.
- Use participant and administrator names from the active project configuration.
- Replace obsolete piped `poc prepare` confirmations with notebook-compatible commands and place a concise safety note beside the first example.
- Add focused command-parsing coverage for both notebooks without enforcing a specific overwrite flag.

This is a documentation and regression-test change only; runtime behavior is unchanged.

### Types of changes

- [ ] Documentation updated.


examples/tutorials/setup_poc.ipynb
```diff
@@ -498,7 +500,7 @@
     "    nvflare poc stop -p [SERVICE]\n",
     "```\n",
     "\n",
-    "Some examples of service participant names can be \"server\", \"site-1\", \"admin@nvidia.com\", or \"lead@hospital.org.us\" depending on your prepared POC. "
+    "Some examples of service participant names in this custom project are \"nonprofit-server\", \"us_hospital\", \"admin@nonprofit.org\", and \"lead@hospital.org.us\". "
```
YuanTingHsieh: Three issues in this stop section:

1. `"nonprofit-server"` is not a participant in `custom_project.yml` — the server there is `general-hospital-server` (custom_project.yml:15). `nvflare poc stop -p nonprofit-server` fails with `CLIException: participant 'nonprofit-server' is not defined, expecting one of: [...]` (exit 4). The same wrong name is in the pre-existing project-description cell above ("nonprofit-server, served as Server"); worth fixing both while here.

2. `admin@nonprofit.org` is a bad `poc stop -p` example: `_stop_poc` unconditionally appends the project admin to the exclusion list, so `nvflare poc stop -p admin@nonprofit.org` prints "Starting shutdown of ['admin@nonprofit.org'] using the stop_fl.sh script" but builds zero commands — a silent no-op with exit 0. The stop parser's own help says "project admin console is not a default managed service".

3. The unchanged sentence at the top of this cell still claims `nvflare poc stop` shuts down "all participants including server, clients, and admin consoles". The default path calls the admin-API `shutdown_system` (server/clients only) and force-excludes the project admin console. Since this PR now teaches starting the console with `-p admin@nonprofit.org`, users will expect `poc stop` to close a console it doesn't manage.

examples/tutorials/setup_poc.ipynb
```diff
@@ -434,19 +434,21 @@
    "source": [
     "## Start NVFlare system in POC mode\n",
     "\n",
-    "When starting the POC deployment, it's necessary to use a separate terminal since the `nvflare poc start` command will run  in the foreground emitting output from the server and any connected clients.\n",
+    "You can run `nvflare poc start` from a terminal or notebook cell. The command starts the services, waits for them to become ready, and then returns.\n",
```
YuanTingHsieh: "from a terminal or notebook cell" contradicts the unchanged intro callout at the top of this notebook: "It is ideal to start your NVFLARE system in POC mode from a **terminal**, not from a notebook." The new guidance is correct at HEAD (start returns after readiness), so the intro callout should be updated in the same PR.

Also minor: "waits for them to become ready, and then returns" is the success path only — the wait is bounded at 30s (`POC_START_READY_TIMEOUT`) and on expiry the command exits 2 with CONNECTION_FAILED while services continue starting in the background. Worth a "waits up to 30 seconds (configurable with `--timeout`)" phrasing.

examples/tutorials/setup_poc.ipynb
```diff
@@ -457,7 +459,7 @@
     "1. We start the system with two hospital clients and server, without the project admin console:\n",
     "\n",
     "```\n",
-    "    nvflare poc start -ex admin@nonprofit.org\n",
+    "    nvflare poc start\n",
     "```\n",
     "\n",
     "2. Next we like to start US hospital's lead's FLARE console:\n",
```
YuanTingHsieh: Retained text, but worth fixing while touching this cell: `nvflare poc start -p lead@hospital.org.us` does not open an interactive console. Only the project admin goes through the foreground `sync_process`; every other admin is launched via `async_process` with stdout redirected to `poc_console.log`. The user runs this in a terminal, gets the prompt back, and no console ever appears — fl_admin sits in the background.

Also, `Note the project admin is "admin@nonprofit.org" instead of "admin@nvidia.com"` above now compares against a default this PR deleted from the notebook — the earlier text establishing admin@nvidia.com is gone, so the "instead of" references nothing. And step 1's bare `nvflare poc start` is now byte-identical to the command shown two paragraphs up; the "complex example" could shrink to just the lead-console step.

examples/tutorials/self-paced-training/part-2_federated_learning_system/chapter-3_federated_computing_platform/03.2_deployment_simulation/simulate_real_world_deployment.ipynb
```diff
@@ -319,10 +317,10 @@
     "To check the status of all components:\n",
     "\n",
     "```bash\n",
-    "nvflare poc status\n",
+    "nvflare system status\n",
```
YuanTingHsieh, line 324: These commands are only correct at repo HEAD. On the NVFlare versions this course targets — the course README says the notebooks were developed with 2.6, and sibling chapters pin `nvflare~=2.7.2rc` — there is no `system` subcommand at all (verified against the 2.7.2 tag: `nvflare system status` → argparse "invalid choice: 'system'", exit 2). Likewise 2.7.2's `_start_poc` excludes only non-project admins, so on 2.7.x `nvflare poc start` does start the project admin console, in the foreground — contradicting the new "Admin consoles are not started by default" text above.

If the self-paced notebooks are meant to work with the pinned/released versions, this needs a version callout; if they track main, an explicit "requires 2.8+" note would prevent confusion given the README's 2.6 disclaimer.

examples/tutorials/self-paced-training/part-2_federated_learning_system/chapter-3_federated_computing_platform/03.2_deployment_simulation/simulate_real_world_deployment.ipynb
```diff
@@ -270,13 +270,13 @@
...
     "\n",
-    "Once you've created your POC environment, you can start the federated learning system. In a real deployment, each site would start their own component. In POC mode, you can start all components with a single command:\n",
+    "Once you've created your POC environment, you can start the federated learning system. In a real deployment, each site would start their own component. In POC mode, you can start the server and all clients with a single command:\n",
     "\n",
     "```bash\n",
     "nvflare poc start\n",
     "```\n",
     "\n",
-    "This command starts the server and all client processes in the background. \n",
+    "This command starts the server and all client processes in the background. Admin consoles are not started by default.\n",
```
YuanTingHsieh: At this point in the notebook the active workspace is the docker one (the last executed prepare is `! echo 'y'| nvflare poc prepare -d 'nvflare/nvflare'`), where first start runs each `docker.sh` → `docker pull`. The default readiness wait is 30s (`POC_START_READY_TIMEOUT`), so a first-ever start on a machine without the image cached exits 2 with "POC system did not become ready" even though the containers come up shortly after. Worth mentioning `--timeout`/`--no-wait` here.

examples/tutorials/self-paced-training/part-2_federated_learning_system/chapter-3_federated_computing_platform/03.2_deployment_simulation/simulate_real_world_deployment.ipynb
```diff
@@ -286,23 +286,21 @@
     "# Start only the server\n",
     "nvflare poc start -p server\n",
     "\n",
-    "# Start specific clients\n",
-    "nvflare poc start -p site-1 -p site-2\n",
+    "# Start specific clients, one per command\n",
```
YuanTingHsieh, line 293: These client-only examples work only if the server is already running: `poc start -p site-1` blocks in `_wait_for_poc_system_ready` (which connects through the admin port), and with no server it waits the full 30s then exits 2 — even though the client process launched fine. (`_ensure_server_running_for_admin_console` fails fast only for admin consoles, not clients.) Worth stating the server-first ordering as a prerequisite rather than presenting `-p server` as a parallel alternative.

tests/unit_test/examples/poc_tutorial_commands_test.py
```diff
@@ -0,0 +1,92 @@
...
+
+
+def test_setup_tutorial_describes_default_poc_services():
+    text = _notebook_text(_SETUP_TUTORIAL)
+
+    assert "starts the server and all clients" in text
+    assert "Admin consoles are optional and are not started unless explicitly selected" in text
+    assert "as well as project admin console" not in text
+    assert "nvflare poc start -p admin@nonprofit.org" in text
+    assert "nvflare poc start -p admin@nvidia.com" not in text
```
YuanTingHsieh: This guards a string that never existed in any revision — the command this PR actually removed was `nvflare poc start -ex admin@nvidia.com` (`-ex`, not `-p`). As written, reintroducing the exact removed command passes all five tests: it parses, satisfies the ≤1 `-ex` count, and trips no text assert. Suggest `assert "nvflare poc start -ex admin@nvidia.com" not in text` instead (or in addition).

tests/unit_test/examples/poc_tutorial_commands_test.py
```diff
@@ -0,0 +1,92 @@
...
+
+def _notebook_text(path: Path) -> str:
+    notebook = json.loads(path.read_text(encoding="utf-8"))
+    return "\n".join("".join(cell.get("source", [])) for cell in notebook["cells"])
+
+
+def _nvflare_commands(path: Path):
+    for line in _notebook_text(path).splitlines():
+        command = line.strip().removeprefix("!").strip()
+        if command.startswith("nvflare "):
```
YuanTingHsieh: Two gaps in the extractor:

1. It only matches lines that start with `nvflare` after stripping `!`, so every actually-executed piped cell — `! echo 'y' | nvflare poc prepare ...`, 8 cells across the two notebooks — is skipped. The parse test currently validates only markdown fences, none of the commands the notebooks execute. Splitting on `|` and scanning each segment would cover them.

2. `shlex.split` runs before any filter with no ValueError guard: a future prose line starting with `nvflare ` that contains an apostrophe (e.g. "nvflare poc start doesn't require a separate terminal") raises `ValueError: No closing quotation` and errors 4 of the 5 tests with a traceback pointing nowhere near the edit. All 32 currently-matching lines are safe today; a try/except-continue (or extracting only from fenced blocks) makes this robust.

tests/unit_test/examples/poc_tutorial_commands_test.py
```diff
@@ -0,0 +1,92 @@
...
+            yield shlex.split(command)
+
+
+def _tutorial_parser():
+    parser = argparse.ArgumentParser(prog="nvflare")
+    subcommands = parser.add_subparsers(dest="subcommand")
+    def_poc_parser(subcommands)
```
YuanTingHsieh: Consider validating against the real CLI instead of a reconstructed parser — `tests/unit_test/tool/cli_schema_missing_args_test.py` already has the pattern (patch `sys.argv`, drive the real `cli.parse_args`). That removes the `{"poc", "system"}` filter (so the notebook's `nvflare config -pw ... --job_templates_dir ...` line gets validated too), covers global flags like `--format`, and can't drift from cli.py wiring. Today the hand-built wiring happens to match `cli.py` exactly, but nothing enforces that.

YuanTingHsieh (commented):
Reviewed at 05a87a837. The direction is right — I verified every core claim against the CLI at HEAD: `-p`/`-ex` are scalar (`nargs="?"`, repetition silently last-wins), `poc status` never existed, admin consoles are excluded by default, and `poc start` waits for readiness then returns. The new tests pass 5/5 and style/license are clean. But the rewrite introduces a few factual errors of the same class it fixes, and the tests have gaps that miss the exact regressions they target. Details inline; three items that can't be anchored to diff lines:

**1. `poc clean` claim (simulate_real_world_deployment.ipynb, "Cleaning Up", unchanged line in the edited cell).** The text says clean removes "all processes and temporary files". `_clean_poc` refuses to run while the system is running (`CLIException: system is still running, please stop the system first.`, exit 4) unless `--force`, and never removes processes — it only deletes the workspace. A reader who takes the sentence literally and skips `poc stop` gets an error instead of cleanup.

**2. The deeper fix for repeated `-p`/`-ex` is CLI-side.** `define_start_parser`/`define_stop_parser` use plain scalar `store` options, so `nvflare poc start -p site-1 -p site-2` silently starts only site-2 with exit 0 — worse than an error. Everything downstream already consumes lists (`get_service_list`/`get_excluded` wrap the scalar), so `action="append"` is a small change that would make the previously-documented commands just work. As written, `test_poc_tutorial_commands_do_not_repeat_scalar_service_options` enshrines the limitation and would have to be deleted if the CLI is ever fixed properly. Fine to defer, but worth a tracking issue.

**3. Scope: this fixes 2 of ~14 files teaching the stale idiom.** `docs/user_guide/data_scientist_guide/poc.rst:26` still affirmatively says bare `poc start` starts "an admin console" and prescribes `-ex admin@nvidia.com` — directly contradicting the tutorials this PR just fixed. Same stale pattern in `docs/user_guide/nvflare_cli/poc_command.rst:215`, `examples/hello-world/hw_pre_start.sh:24`, `examples/advanced/client-api-attach/README.md:144`, `examples/advanced/monitoring/jobs/README.md:136`, `examples/tutorials/job_recipe.ipynb`, and 7 self-paced-training notebooks (chapters 3.3, 3.4, 5.3, 6.3, 12.1). Suggest a follow-up sweep so the repo teaches one mental model.

YuanTingHsieh (approved):
Re-reviewed at b4597691d. All review items are addressed and verified:

- Doc errors fixed: `general-hospital-server` name corrected (both occurrences), the no-op `poc stop -p admin@nonprofit.org` and non-interactive lead-console examples removed, stop section now correctly treats the console as separate, intro callout updated, `poc clean` ordering clarified, version callout (2.8+) and `--timeout 120` docker guidance added, server-first ordering stated.
- Test rewritten against the real `cli.parse_args` with sys.argv patched; extraction now covers fenced markdown and executed `!` cells (piped commands split on shell operators), shlex errors surface with notebook/cell/line attribution, and non-empty extraction is asserted per notebook. Verified: passes alone and alongside `tests/unit_test/tool/poc/` in one process (99 passed); black/isort/flake8 clean; both notebooks nbformat-valid.
- The `--force` callout is accurate (`_prepare_poc` stops a running POC, deletes, rebuilds without prompting) — and correctly replaces `echo 'y' |`, which now exits 4 in non-tty contexts.

One non-blocking note: pruning the text asserts also dropped the guard against reintroducing repeated `-p`/`-ex` or `-ex admin@nvidia.com` — argparse accepts repeats silently (last-wins), so the parse test can't catch them. A small check over the extracted command tokens (no repeated scalar flag within a command) would restore that without the prose brittleness. Fine as a follow-up, along with the CLI-side `action="append"` fix and the sweep of the ~12 other files still teaching `-ex admin@nvidia.com` (including `docs/user_guide/data_scientist_guide/poc.rst`, which still says the admin console starts by default).

PR #5158, opened 2026-08-18
Authenticate Cell messages with AES-GCM

IsaacYangSLA:
## Summary

- replace unauthenticated AES-CBC/PKCS7 Cell payload encryption with AES-256-GCM
- sign each versioned encrypted envelope, including the nonce, wrapped key, ciphertext, and authentication tag
- verify the sender signature on every message, including cached-key messages, before decrypting
- bind cached decryption keys to the validated origin certificate
- reject malformed, tampered, and legacy unversioned ciphertexts

## Root cause

`SimpleCellCipher` previously signed only the RSA-OAEP-wrapped symmetric key. The IV and AES-CBC ciphertext were not covered by a MAC, AEAD tag, or signature, so an active attacker could modify ciphertext without detection. CBC bit flipping could produce altered same-length plaintext that passed the existing post-decryption length check.

## Impact

Cell payloads now have confidentiality, ciphertext integrity, and sender-specific message authenticity. Any change to the authenticated envelope is rejected before plaintext is returned. The explicit format version prevents an accidental fallback to the legacy unauthenticated CBC format, so mixed-version peers must be upgraded together.

Replay protection is separate from message authenticity and is not introduced by this change.

## Validation

- `./.venv/bin/python -m pytest -q tests/unit_test/fuel/f3/cellnet/identity_binding_test.py` (`40 passed`)
- `./.venv/bin/python -m pytest -q tests/unit_test/fuel/f3/cellnet` (`180 passed`)


IsaacYangSLA:
## Large-message performance

I benchmarked the existing implementation at `29506703` against this PR at `9aa7f4af`. Total time is encryption plus decryption.

| Plaintext size | Existing total | PR total | Improvement |
|---:|---:|---:|---:|
| 16 MiB | 34.76 ms | 33.36 ms | 4% faster |
| 64 MiB | 248.81 ms | 219.69 ms | 12% faster |
| 128 MiB | 504.00 ms | 441.80 ms | 12% faster |

For the 128 MiB workload, peak process RSS was approximately 699 MiB with this PR versus 953 MiB with the existing implementation—a reduction of approximately 27%.

Despite authenticating every message, the PR’s OpenSSL AES-GCM path was faster and used less peak memory for large messages on this system.

Results are averages of two runs using warm caches and per-size medians. The test used Python 3.12.1, `cryptography` 49.0.0, and OpenSSL 4.0.1.

nvidianz (approved):
Safe to go in 2.9

PR #5159, opened 2026-08-18
Align Network Attach profile with internal mTLS defaults

pcnudde:
## Summary

- align the shipped Network Attach profile with the default internal mTLS listener used by prepared Docker, Kubernetes, and Slurm runtimes
- document the existing clear-transport opt-outs and their matching `tcp`/`clear` trainer profile
- extend the existing real-Cell test to cover both matching clear and mTLS configurations

## Root cause

Prepared runtime parents now use `stcp` with mTLS by default, while the shipped Network Attach profile still selected `tcp` with clear transport. Network Attach already supports the matching mTLS path with the provisioned credentials located beside `ca_cert`, so no runtime protocol change is needed.

## Impact

The normal example now works with prepared-runtime defaults. Deployments that intentionally use clear transport can continue to set `parent.internal_connection_security: clear` for Docker/Kubernetes or `job_launcher.internal_connection_security: clear` for Slurm and keep a matching `tcp`/`clear` trainer profile.

## Validation

- `python -m pytest -q tests/integration_test/fast/client_api_attach_e2e_test.py::test_secure_network_attach_uses_site_cell_identity_over_cp` — 2 passed
- `git diff --check`

The focused real-Cell test uses locally generated credentials; no live Docker, Kubernetes, or Slurm deployment was mutated.

### DIFF
--- a/examples/advanced/client-api-attach/README.md
+++ b/examples/advanced/client-api-attach/README.md
@@ -38,18 +38,22 @@ Start from `attach_profile_network.json`:
   "execution_mode": "attach",
   "attach_id": "numpy_trainer",
   "site_name": "site-1",
-  "connect_url": "tcp://site-1.example.com:8004",
-  "connection_security": "clear",
+  "connect_url": "stcp://site-1.example.com:8004",
+  "connection_security": "mtls",
   "secure_mode": true,
   "ca_cert": "/absolute/path/to/site/startup/rootCA.pem",
   "job_wait_timeout": null
 }
 ```
 
-`connect_url` must be the provisioned CP internal listener, not a CJ URL. Keep
-`client.crt` and `client.key` beside `rootCA.pem`. `secure_mode=true` uses those
-site credentials for Cell authentication and end-to-end payload encryption even
-when the CP transport itself is `connection_security="clear"`.
+`connect_url` must be the provisioned CP internal listener, not a CJ URL. Prepared
+Docker, Kubernetes, and Slurm parents use `stcp` with mTLS by default. Keep
+`client.crt` and `client.key` beside `rootCA.pem`; they provide transport and Cell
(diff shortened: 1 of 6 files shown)


examples/advanced/client-api-attach/README.md
```diff
@@ -38,18 +38,22 @@ Start from `attach_profile_network.json`:
   "execution_mode": "attach",
   "attach_id": "numpy_trainer",
   "site_name": "site-1",
-  "connect_url": "tcp://site-1.example.com:8004",
-  "connection_security": "clear",
+  "connect_url": "stcp://site-1.example.com:8004",
```
YuanTingHsieh, line 41: **LOW — Keep the Local POC instructions aligned with this profile.** The Local POC section later tells users to reuse `attach_profile_network.json` after replacing its placeholders. The documented process-mode POC does not provision an `internal` listener configuration, so `ConnectorManager` creates a dynamic `tcp`/`clear` CP listener. Leaving this profile at `stcp`/`mtls` can cause a TLS-to-clear connection failure (or a misleading transport-policy mismatch if the actual `tcp://` URL is pasted). Please add a sentence there such as: “For the local process-mode POC above, use the CP’s `tcp://` URL and set `connection_security` to `"clear"`; keep `secure_mode` enabled for end-to-end Cell protection.” Docker POC and prepared runtimes can retain mTLS.

YuanTingHsieh (commented):
One low-severity documentation issue in the Local POC flow.

YuanTingHsieh (dismissed):

examples/advanced/client-api-attach/README.md
pcnudde: will change this

YuanTingHsieh (approved):

PR #5162, opened 2026-08-18
Enable best-model selection for Swarm recipes

chesterxgchen:
## Summary

- configure a client-side `IntimeModelSelector` by default in `SwarmLearningRecipe`
- expose `key_metric` and `key_metric_mode` while preserving positional compatibility
- support higher-is-better and lower-is-better metrics without introducing a new selector or service
- update the canonical Swarm LoRA example to evaluate the received global adapter before local training and report `val_loss`
- align recipe, Job API/JSON, CSE, user-guide, release-note, and example documentation with the client-side Swarm topology and metric-timing contract

## Behavior

When clients report the configured pre-training metric, Swarm tracks the best global model across aggregation clients, distributes it at workflow completion, and persists `best_FL_global_model.pt` through the existing client-side PyTorch persistor. If the metric is absent, the workflow continues to persist the last global model and does not create a best-model checkpoint.

This uses the existing full `IntimeModelSelector`, including dictionary-metric selection and `min`/`max` handling; it does not depend on changes to `SimpleIntimeModelSelector`.

## Validation

- `python3 -m pytest tests/unit_test/recipe/swarm_recipe_test.py tests/unit_test/recipe/component_config_verification_test.py tests/unit_test/examples/recipe_export_arguments_test.py -q` (47 passed, 5 skipped)
- full recipe unit suite (532 passed, 22 skipped)
- Python syntax compilation for the updated Swarm example
- `./runtest.sh -s`
- pre-push agent-skill lint (0 findings)
- `git diff --check`

The full HTML documentation build entered Sphinx but did not complete locally because unrelated API autodoc imports stalled. The changed RST, Markdown, recipe snippets, exported JSON expectations, and example behavior were audited together for consistency.

### DIFF
--- a/docs/release_notes/flare_290.rst
+++ b/docs/release_notes/flare_290.rst
@@ -119,4 +119,5 @@ Compatibility and Migration Notes
   and ``key_metric_mode="min"`` for lower-is-better metrics. Clients must report
   a pre-training validation metric with the configured name for selection to
   occur; jobs without that metric continue to persist the last global model but
-  do not create ``best_FL_global_model.pt``.
+  do not create ``best_FL_global_model.pt``. Set ``key_metric=None`` to opt out
+  and preserve the pre-2.9 last-model-only behavior.
--- a/docs/release_notes/flare_290.rst
+++ b/docs/release_notes/flare_290.rst
@@ -114,3 +114,9 @@ Compatibility and Migration Notes
   missing client. The accepted envelope may already have exposed references to
   downstream consumers and cannot be safely withdrawn. An explicit job abort
   that wins the terminal-state race remains ``ABORTED``.
+- ``SwarmLearningRecipe`` now configures client-side best-model selection by
+  default. Use ``key_metric`` to select a dictionary-valued validation metric
+  and ``key_metric_mode="min"`` for lower-is-better metrics. Clients must report
+  a pre-training validation metric with the configured name for selection to
+  occur; jobs without that metric continue to persist the last global model but
+  do not create ``best_FL_global_model.pt``.
--- a/docs/release_notes/flare_290.rst
+++ b/docs/release_notes/flare_290.rst
@@ -120,4 +120,11 @@ Compatibility and Migration Notes
   a pre-training validation metric with the configured name for selection to
   occur; jobs without that metric continue to persist the last global model but
   do not create ``best_FL_global_model.pt``. Set ``key_metric=None`` to opt out
-  and preserve the pre-2.9 last-model-only behavior.
+  and preserve the pre-2.9 last-model-only behavior. Selection skips round 0,
+  so a one-round job does not create a best-model checkpoint. With
+  ``key_metric_mode="min"``, Swarm best-metric logs and records expose the
+  negated comparison value (for example, a loss of 2.31 is shown as -2.31).
+  ``client_config_overrides`` can no longer replace ``model_selector``: migrate
+  the former ``{"model_selector": None}`` opt-out to ``key_metric=None``, and
+  use ``BaseSwarmLearningRecipe`` with an explicit ``SwarmClientConfig`` for a
+  custom selector.
--- a/docs/user_guide/data_scientist_guide/available_recipes.rst
+++ b/docs/user_guide/data_scientist_guide/available_recipes.rst
@@ -1040,6 +1040,8 @@ Decentralized federated learning without a central server.
         min_clients=3,
         num_rounds=5,
         train_script="client.py",
+        key_metric="accuracy",
+        key_metric_mode="max",
         initial_ckpt="path/to/pretrained.pt",  # Optional: pre-trained weights
         progress_timeout=7200,
         learn_task_timeout=None,  # No training-task time limit
@@ -1055,6 +1057,12 @@ Decentralized federated learning without a central server.
 For ``initial_ckpt``, a relative path is bundled and distributed to every client.
 An absolute path is not distributed; it must be readable at the same path on every
 client because the Swarm model persistor runs client-side.
+
+The recipe configures client-side best-model selection by default. Training clients
+must report a pre-training validation metric matching ``key_metric``; for dictionary
+metrics, the named entry is selected. Set ``key_metric_mode="min"`` for metrics such
+as loss. The best model is distributed at workflow completion and the default
+PyTorch persistor writes ``best_FL_global_model.pt`` at each result client.
 
 .. note::
    For large models (>2 GB), tune the following parameters:
--- a/docs/user_guide/data_scientist_guide/available_recipes.rst
+++ b/docs/user_guide/data_scientist_guide/available_recipes.rst
@@ -1061,8 +1061,14 @@ client because the Swarm model persistor runs client-side.
 The recipe configures client-side best-model selection by default. Training clients
 must report a pre-training validation metric matching ``key_metric``; for dictionary
 metrics, the named entry is selected. Set ``key_metric_mode="min"`` for metrics such
-as loss. The best model is distributed at workflow completion and the default
-PyTorch persistor writes ``best_FL_global_model.pt`` at each result client.
+as loss, or set ``key_metric=None`` to disable best-model selection. In ``"min"``
+mode, Swarm logs and best-metric records expose the negated comparison value. Model
+selection starts after round 0, so a one-round job does not create a best-model
+checkpoint. The best model is distributed at successful workflow completion and the
+default PyTorch persistor writes ``best_FL_global_model.pt`` at each result client.
+Rotating aggregation clients can write interim per-client best checkpoints during
+the run; those files are authoritative only after successful completion at result
+clients.
 
 .. note::
    For large models (>2 GB), tune the following parameters:
@@ -1088,12 +1094,14 @@ PyTorch persistor writes ``best_FL_global_model.pt`` at each result client.
 
 For advanced controller settings, ``server_config_overrides`` and
 ``client_config_overrides`` are shallow-merged into ``SwarmServerConfig`` and
-``SwarmClientConfig`` after the named parameters. Overlapping dictionary values
-therefore take precedence over the documented named API. Client overrides cannot
-replace the recipe-managed executor, aggregator, persistor, shareable generator, or
-``min_responses_required``; use ``BaseSwarmLearningRecipe`` for custom components or
-quorum settings. Server overrides cannot replace ``min_clients``; set it through the
-named parameter so all scheduler and workflow quorum settings remain aligned.
+``SwarmClientConfig`` after the named parameters. Overlapping dictionary values for
+non-recipe-managed fields therefore take precedence over the documented named API.
+Client overrides cannot replace the recipe-managed executor, aggregator, persistor,
+shareable generator, model selector, or ``min_responses_required``. Use
+``key_metric=None`` to disable selection; for a custom selector or other custom
+components, use ``BaseSwarmLearningRecipe`` with an explicit ``SwarmClientConfig``.
+Server overrides cannot replace ``min_clients``; set it through the named parameter
+so all scheduler and workflow quorum settings remain aligned.
 
 
 Edge Recipes
--- a/examples/advanced/swarm_learning/swarm_pt/README.md
+++ b/examples/advanced/swarm_learning/swarm_pt/README.md
@@ -84,7 +84,7 @@ input_model = flare.receive()         # 2. Receive global LoRA adapter from the 
 apply_global_adapter(model, input_model.params)  # 3. Apply global adapter weights
 
 metrics = {}
-if round_num > 0:
+if input_model.current_round != 0:
     val_loss = evaluate_loss(model, validation_dataloader, validation_steps)
     metrics = {"val_loss": val_loss}
 adapter_diff = local_train(model, train_dataloader, local_steps)  # local fine-tuning
--- a/examples/advanced/swarm_learning/swarm_pt/README.md
+++ b/examples/advanced/swarm_learning/swarm_pt/README.md
@@ -54,6 +54,11 @@ A CUDA-capable GPU is strongly recommended. The base model is loaded in `bfloat1
 
 This example uses the [wikitext-2-raw-v1](https://huggingface.co/datasets/wikitext) dataset (Apache-2.0, no license acceptance required). The training split is partitioned into disjoint shards — one per client — to simulate real-world data silos where each participant trains only on their own local data.
 
+When `--data_dir` is supplied, the directory must contain both each site's local
+training dataset at `<data_dir>/<site>/train` and a shared validation dataset at
+`<data_dir>/validation`. Custom corpora must provide the same layout; they do not
+need to be regenerated with this example's `prepare_data.py`.
+
 ## Model
 
 [Qwen2.5-0.5B](https://huggingface.co/Qwen/Qwen2.5-0.5B) (Apache-2.0, default) or [Qwen2.5-1.5B](https://huggingface.co/Qwen/Qwen2.5-1.5B) is used as the base model, selectable via `--model_size`. LoRA adapters are attached to the query and value projection layers of every attention block:
@@ -78,27 +83,31 @@ flare.init()                          # 1. Initialize NVFlare Client API
 input_model = flare.receive()         # 2. Receive global LoRA adapter from the workflow
 apply_global_adapter(model, input_model.params)  # 3. Apply global adapter weights
 
-val_loss = evaluate_loss(model, validation_dataloader, validation_steps)
-adapter_diff = local_train(model, dataloader, steps)  # local fine-tuning
+metrics = {}
+if round_num > 0:
+    val_loss = evaluate_loss(model, validation_dataloader, validation_steps)
+    metrics = {"val_loss": val_loss}
+adapter_diff = local_train(model, train_dataloader, local_steps)  # local fine-tuning
 
 flare.send(flare.FLModel(             # 4. Send LoRA adapter diff back
     params_type=ParamsType.DIFF,
     params=adapter_diff,
-    metrics={"val_loss": val_loss},
+    metrics=metrics,
 ))
 ```
 
-Each client first evaluates the received global adapter on the validation split, then fine-tunes for a fixed number of gradient steps and returns the adapter weight **diff** (after − before). The selected Swarm aggregation client averages both the pre-training `val_loss` values and the model diffs. Because the metric describes the received global model rather than a locally updated model, it can be used safely for global best-model selection.
+Starting in round 1, each client first evaluates the received global adapter on the validation split, then fine-tunes for a fixed number of gradient steps and returns the adapter weight **diff** (after − before). Round 0 skips validation because no aggregated global adapter exists yet. The selected Swarm aggregation client averages both the pre-training `val_loss` values and the model diffs. Because the metric describes the received global model rather than a locally updated model, it can be used safely for global best-model selection.
 
 ## Peer-to-Peer Workflow
 
 This example uses [Swarm Learning](https://nvflare.readthedocs.io/en/main/apidocs/nvflare.app_common.ccwf.html), a decentralized federated workflow where clients act as both trainers and aggregators in a peer-to-peer fashion — there is no central aggregation server. The workflow proceeds as follows:
 
 1. A designated starting client scatters the global LoRA adapter to all peers.
-2. Each client trains locally and submits its adapter diff to a designated aggregator.
-3. The aggregator accumulates diffs and produces a new global adapter.
-4. The updated adapter is scattered to the next round of trainers.
-5. This repeats for `num_rounds` rounds.
+2. Starting in round 1, each client validates the received global adapter before training.
+3. Each client trains locally and submits its adapter diff to a designated aggregator.
+4. The aggregator accumulates diffs and produces a new global adapter.
+5. The updated adapter is scattered to the next round of trainers.
+6. This repeats for `num_rounds` rounds.
 
 ## Job Recipe Code
 
@@ -144,6 +153,10 @@ recipe = SwarmLearningRecipe(
 - `InTimeAccumulateWeightedAggregator` for peer-to-peer adapter aggregation
 - `SimpleModelShareableGenerator` for model ↔ shareable conversion
 - `IntimeModelSelector` for client-side best-model selection using pre-training `val_loss`
+
+Because `key_metric_mode="min"` converts loss into a higher-is-better comparison
+value, Swarm best-metric logs and records display the negated loss. For example,
+an actual `val_loss` of 2.31 is displayed as -2.31.
 
 ## Step-by-Step Instructions
 
@@ -257,11 +270,11 @@ This writes a standard NVFlare job folder that can be submitted to a production 
 
 #### Each Round
 - **Scatter**: Starting client sends global LoRA adapter to all trainers.
-- **Validate**: Each client scores the received global adapter on the shared validation split before training.
+- **Validate**: Starting in round 1, each client scores the received global adapter on the shared validation split before training.
 - **Local training**: Each client fine-tunes for `local_steps` gradient steps on its data shard, reporting loss every 5 steps.
-- **Submit**: Each client sends its pre-training `val_loss` and LoRA adapter diff to the designated aggregator.
+- **Submit**: Each client sends its LoRA adapter diff and, starting in round 1, its pre-training `val_loss` to the designated aggregator.
 - **Aggregate**: Aggregator accumulates metrics and diffs, produces the updated global adapter, and records a new best model when `val_loss` improves.
 
 #### Completion
-- Final and best-scoring global LoRA adapters are distributed and persisted to the simulation workspace.
+- Final and best-scoring global LoRA adapters are distributed and persisted to the simulation workspace. Best-model selection starts after round 0, so at least two rounds are required to create a best-model checkpoint.
 - Results available at `<workspace>/ccwf_swarm_pt_lora/`.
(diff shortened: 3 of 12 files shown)


examples/advanced/swarm_learning/swarm_pt/client.py
```diff
@@ -177,6 +184,27 @@ def local_train(model, dataloader: DataLoader, steps: int) -> dict:
...
+                break
+            loss = model(
+                input_ids=batch["input_ids"].to(device),
+                attention_mask=batch["attention_mask"].to(device),
+                labels=batch["labels"].to(device),
+            ).loss
+            losses.append(loss.item())
+
+    if not losses:
+        raise RuntimeError("validation dataloader produced no complete batches")
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in d7b6eed0e. The empty validation-loader path now raises ValueError, which identifies invalid validation input consistently. Verified Python compilation, the Swarm recipe unit tests, ./runtest.sh -s, and the pre-push agent-skill lint.

YuanTingHsieh:
**Blocking ask: coerce metric values to `float` in `IntimeModelSelector._before_accept` as part of this PR (or land it as a precursor before this merges).**

The selector this PR turns on by default never coerces the reported metric, and swarm is the one workflow where that matters: `NumberMetricComparator.compare` (`nvflare/app_common/ccwf/common.py:204`) raises `ValueError` unless the value is `isinstance(a, (int, float))`. An `np.float32` — or a 0-d tensor from a forgotten `.item()`, a very common Client API habit — survives the whole pipeline intact:

- `FLModel.metrics` goes into DXO meta `INITIAL_METRICS` unmodified (`fl_model_utils.py:69`)
- the selector's negation `-1.0 * validation_metric` and weighted average keep the numpy type
- `VALIDATION_RESULT` → `executor.best_metric` (`swarm_client_ctl.py:523`) and the `Constant.METRIC` header; FOBS round-trips `np.float32` back as `np.float32` (`Float32ScalarDecomposer` reconstructs the numpy scalar), so cross-client propagation preserves the poisonous type

Failure sequence (verified empirically at this PR's head): round 1's aggregator adopts the value on the no-compare branch (`current_best_global_metric` is still `None`); at round 2, `Gatherer.aggregate()` (`swarm_client_ctl.py:226-233`) calls `compare(np.float32, ...)` → `ValueError` → caught in `_end_gather` (`swarm_client_ctl.py:640-645`) → `update_status(error=EXECUTION_EXCEPTION)` → server `system_panic` (`server_ctl.py:520-522`). **The whole job aborts mid-run**; `num_rounds >= 3` is the only precondition.

The missing coercion is pre-existing in the widget, but it was unreachable for swarm recipe users before this PR (no selector deployed → `best_metric` stays `None` → the compare never runs), and FedAvg is unaffected (the selector-internal `>` comparisons work fine on numpy scalars). Making the selector default-on is what arms it, so it should be defused with this change.

The fix is the same one #5150 applies to `SimpleIntimeModelSelector` — in `IntimeModelSelector._before_accept`, after the dict-key extraction and before negation:

```python
validation_metric = float(validation_metric)
```

plus a unit test feeding `np.float32`. One line in the widget fixes it for every consumer, not just swarm. Note #5150 does not cover this: it touches only the Simple variant, and this PR explicitly doesn't depend on it.

chesterxgchen:
Addressed the blocking metric-type issue in d036f0b6a.

- `IntimeModelSelector._before_accept` now coerces the selected metric to a native `float` before negation and aggregation.
- Added regression coverage using a dict-valued `np.float32` metric and verified both selected and raw metrics are Python floats.
- Rebased the branch onto `upstream/main` at 6cc8886e0.

Validation: selector tests (20 passed), Swarm recipe tests (35 passed), and `./runtest.sh -s` passed.

chesterxgchen:
Fixed the stale generated recipe catalog in e791e7103 by running `python -m nvflare.tool.recipe.generate_recipe_catalog`. The generated `swarm-pt` entry now includes `key_metric` and `key_metric_mode`.

Rebased again onto current `upstream/main` at 6592a45ef. Validation: recipe CLI tests (42 passed), selector tests (20 passed), Swarm recipe tests (35 passed), and `./runtest.sh -s` passed.

nvflare/app_opt/pt/recipes/swarm.py
```diff
@@ -366,6 +380,10 @@ def __init__(
                 allow_numpy_conversion=aggregation_format != ExchangeFormat.PYTORCH,
             ),
             "shareable_generator": SimpleModelShareableGenerator(),
+            "model_selector": IntimeModelSelector(
```
holgerroth: Making the selector default-on changes behavior for every existing `SwarmLearningRecipe` job on upgrade, and there's no named way to turn it off:

- Scripts that already report metrics measured *after* local training (e.g. `metrics={"accuracy": ...}` for tracking) will silently get a `best_FL_global_model.pt` selected on wrong semantics — `FLModelUtils.to_shareable` (fl_model_utils.py:69) maps any `FLModel.metrics` to `INITIAL_METRICS`, and nothing checks the pre-training timing the docs require.
- Scripts that send no metrics now log `validation metric not existing in <client>` once per trainer per round on the aggregation client.
- The only opt-out is `client_config_overrides={"model_selector": None}` — it works (`SwarmClientConfig` treats None as not-configured) but is documented nowhere. And since `model_selector` isn't in `_RECIPE_MANAGED_CLIENT_CONFIG_KEYS`, an override like `{"model_selector": IntimeModelSelector()}` silently *replaces* the selector configured by `key_metric`/`key_metric_mode` — the exact conflict the frozenset raises `ValueError` for with executor/aggregator/persistor.

Suggest: a first-class opt-out (e.g. `key_metric=None` skips the selector), add `model_selector` to the protected keys, and call out the behavior change + opt-out in the release note.

Also relevant for large models: each improving round now does a synchronous full-model `copy.deepcopy` (swarm_client_ctl.py:547, retained until job end) plus a second full checkpoint write (file_model_persistor.py:301-306) on the aggregation path, on top of the existing per-round save. Negligible for LoRA adapters, but this recipe explicitly targets 7B+ — and `enable_tensor_disk_offload` doesn't reduce it since `GLOBAL_MODEL` holds full in-memory weights. Another reason an explicit opt-out matters.

nvflare/app_opt/pt/recipes/swarm.py
```diff
@@ -366,6 +380,10 @@ def __init__(
...
+                key_metric=key_metric,
+                negate_key_metric=key_metric_mode == "min",
```
holgerroth: With `key_metric_mode="min"`, the *negated* value is what escapes into every user-facing swarm record: the selector publishes its (negated) `best_val_metric` as `VALIDATION_RESULT` → `swarm_client_ctl.py:546` `best_metric` → `Constant.METRIC` header and the `"global best metric is -X"` log → final BEST broadcast → result clients re-fire `GLOBAL_BEST_MODEL_AVAILABLE` with the negated value → `CWEResultPrinter` prints `My Best Metric: -2.31` for an actual val_loss of 2.31. Ranking stays correct (everything compares in negated space), but the shipped example uses exactly this mode and none of the new docs mention the sign flip — a user grepping for their loss value will find nothing.

The selector already tracks `raw_val_metric`; longer term, propagating the raw value with a mode-aware `metric_comparator` (the extension point already exists on `SwarmClientConfig`) would keep negation an internal detail. Short term, at least a docs note that logged/recorded best metrics are negated in min mode.

nvflare/app_opt/pt/recipes/swarm.py
```diff
@@ -42,6 +43,8 @@
 
 class _SwarmValidator(BaseModel):
     initial_ckpt: Optional[str] = None
+    key_metric: str = "accuracy"
```
holgerroth: `key_metric_mode` gets strict `Literal` validation, but `key_metric` is only type-checked: `SwarmLearningRecipe(key_metric="")` (e.g. an empty templated config value) constructs and exports cleanly, then at runtime `"" in metrics` never matches, so every contribution warns and no best model is ever produced while the job otherwise succeeds. A `min_length=1` constraint (or strip check) here would fail fast the way the mode does.

nvflare/app_common/widgets/intime_model_selector.py
```diff
@@ -163,6 +163,7 @@ def _before_accept(self, fl_ctx: FLContext):
                 )
                 return False
 
+        validation_metric = float(validation_metric)
```
holgerroth: Two suggestions on this coercion:

1. `SimpleIntimeModelSelector._before_accept` wraps the identical coercion in `try/except (TypeError, ValueError)` → one-line warning + `return False` (simple_intime_model_selector.py:137-143). Unguarded here, a non-numeric metric value (`[0.5, 0.6]`, `"N/A"`, `None`) raises out of the handler, and `fire_event_to_components` logs a full traceback per contribution per round instead of the sibling's clean skip. Matching the guarded pattern keeps the two selectors' behavior consistent.
2. `float()` passes NaN through: one client's NaN poisons the round's weighted sum (lines 182-184) and `nan > best` is always False (line 194), so the round silently can never become best — a persistently diverging client freezes `best_FL_global_model.pt` at a stale round with only an info-level log. This line is the natural place for a `math.isfinite` check (warn + skip).

tests/unit_test/app_common/widgets/in_time_model_selector_test.py
```diff
@@ -171,3 +172,28 @@ def test_model_selection_publishes_metrics_selection_info(self):
...
+        shareable.add_cookie(AppConstants.CONTRIBUTION_ROUND, 1)
+        peer_ctx.set_prop(FLContextKey.SHAREABLE, shareable, private=True)
+        fl_ctx = engine.fl_ctx_mgr.new_context()
+        fl_ctx.set_prop(AppConstants.CURRENT_ROUND, 1, private=True, sticky=False)
+        fl_ctx.set_peer_context(peer_ctx)
+
+        handler.handle_event(AppEventType.BEFORE_CONTRIBUTION_ACCEPT, fl_ctx)
+        handler.handle_event(AppEventType.BEFORE_AGGREGATION, fl_ctx)
+
+        assert isinstance(handler.val_metric, float)
```
holgerroth: These `isinstance(..., float)` assertions don't actually guard the fix on numpy < 2 (numpy is unpinned in setup.cfg): with the coercion at intime_model_selector.py:166 reverted, `-1.0 * np.float32(0.2)` — and the `* weight` accumulation on the raw path — promote to `np.float64`, which *subclasses* Python `float`, so all three isinstance checks and `pytest.approx` still pass. Verified empirically on numpy 1.26: the test passes with the fix removed. `assert type(handler.val_metric) is float` (and same for the other two) makes the regression test meaningful on both numpy majors.

holgerroth (commented):
Took a deep pass over this (multi-angle review + per-finding verification against the PR head). The core mechanism checks out end-to-end: `FLModel.metrics` → `INITIAL_METRICS` → selector on the aggregation client, export serializes both selector args, the catalog regeneration is consistent, and the `float()` coercion fixes a real pre-existing `np.float32` failure in `NumberMetricComparator.compare` during swarm aggregation. The pre-training-metric design is the right call.

Inline comments below fall into three groups: consequences of the selector being default-on (opt-out, upgrade semantics, min-mode presentation, large-model cost), a few robustness edges (NaN, round 0, `validation_steps`, empty `key_metric`), and example/docs accuracy.

examples/advanced/swarm_learning/swarm_pt/client.py
```diff
@@ -177,6 +184,27 @@ def local_train(model, dataloader: DataLoader, steps: int) -> dict:
...
+                break
+            loss = model(
+                input_ids=batch["input_ids"].to(device),
+                attention_mask=batch["attention_mask"].to(device),
+                labels=batch["labels"].to(device),
+            ).loss
+            losses.append(loss.item())
+
+    if not losses:
+        raise ValueError("validation dataloader produced no complete batches")
```
holgerroth: This raise fires in two realistic cases and kills every client at round 0:

- `--validation_steps 0` (or negative) — neither argparse here nor in job.py bounds it, the loop breaks before appending anything, and the message then wrongly blames the dataloader (it produced batches; steps consumed none).
- `--batch_size` larger than the validation set, because line 132 hard-codes `drop_last=True` for the validation split too.

Suggest `drop_last=(split == "train")` plus validating `validation_steps >= 1` at parse time (and distinguishing the two messages).

examples/advanced/swarm_learning/swarm_pt/client.py
```diff
@@ -81,34 +83,39 @@ def build_dataloader(
...
     """
+    if split not in ("train", "validation"):
+        raise ValueError(f"split must be 'train' or 'validation', but got {split!r}")
+
     if data_dir is not None:
         from datasets import load_from_disk
 
-        shard_path = os.path.join(data_dir, site_name, "train")
-        if not os.path.isdir(shard_path):
+        dataset_path = os.path.join(data_dir, site_name, "train") if split == "train" else os.path.join(data_dir, split)
```
holgerroth, line 112: This narrows the `--data_dir` contract: pre-PR only `{data_dir}/{site}/train` was read, so hand-built custom-corpus layouts without a `validation/` dir worked; they now crash at startup — and the remediation text (`Run prepare_data.py ...`) would regenerate wikitext-2 over the user's own data. Worth a README note that `validation/` is now required and an error message that doesn't assume the wikitext workflow.

examples/advanced/swarm_learning/swarm_pt/client.py
```diff
@@ -177,6 +184,27 @@ def local_train(model, dataloader: DataLoader, steps: int) -> dict:
     return diff
 
 
+def evaluate_loss(model, dataloader: DataLoader, steps: int) -> float:
```
holgerroth, line 207: Since this PR makes `val_loss` load-bearing for best-model selection: `tokenize()` sets `labels = enc["input_ids"].copy()` with `padding="max_length"`, and Qwen's pad token is eos. HF causal-LM loss only ignores `-100` labels (attention_mask does not exclude label positions), so a sizable share of each 128-token row's loss is trivial pad/eos prediction: absolute values understate the real LM loss, between-model deltas are attenuated, and early rounds partly "improve" by learning eos→eos. Ranking is mostly preserved (deterministic eval, identical pad structure for every model), but masking pad positions to `-100` in `tokenize()` would make the selection metric a clean LM loss.

examples/advanced/swarm_learning/swarm_pt/client.py
```diff
@@ -122,7 +129,7 @@ def tokenize(batch):
 
     dataset = dataset.map(tokenize, batched=True, remove_columns=["text"])
```
holgerroth, line 150: Nit: the whole validation split (~2.9k filtered rows) gets tokenized per client, but with `shuffle=False`, `drop_last=True`, and the steps cap in `evaluate_loss`, only the first `validation_steps × batch_size = 40` examples are ever read. `dataset.select(range(validation_steps * batch_size))` before `.map` on the validation path is metric-identical.

docs/release_notes/flare_290.rst
```diff
@@ -114,3 +114,9 @@ Compatibility and Migration Notes
   missing client. The accepted envelope may already have exposed references to
   downstream consumers and cannot be safely withdrawn. An explicit job abort
   that wins the terminal-state race remains ``ABORTED``.
+- ``SwarmLearningRecipe`` now configures client-side best-model selection by
+  default. Use ``key_metric`` to select a dictionary-valued validation metric
+  and ``key_metric_mode="min"`` for lower-is-better metrics. Clients must report
```
holgerroth, line 119: One more condition worth stating: the selector always skips round 0 (intime_model_selector.py:139-141) and swarm rounds start at 0, so a `num_rounds=1` job (the `SwarmServerConfig` default) never creates `best_FL_global_model.pt` even when clients report the metric every round — as written this note reads as metric ⇒ best file. Relatedly, the example pays `evaluate_loss` in round 0 for a value the selector is guaranteed to discard.

docs/user_guide/data_scientist_guide/available_recipes.rst
```diff
@@ -1056,6 +1058,12 @@ For ``initial_ckpt``, a relative path is bundled and distributed to every client
 An absolute path is not distributed; it must be readable at the same path on every
 client because the Swarm model persistor runs client-side.
 
+The recipe configures client-side best-model selection by default. Training clients
+must report a pre-training validation metric matching ``key_metric``; for dictionary
+metrics, the named entry is selected. Set ``key_metric_mode="min"`` for metrics such
+as loss. The best model is distributed at workflow completion and the default
```
holgerroth: Accurate for the default happy path, but worth a caveat: every rotating aggregation client's *first* improving round writes its own local best to `best_FL_global_model.pt` mid-run (the persistor save on `GLOBAL_BEST_MODEL_AVAILABLE` isn't gated by the cross-client comparison), and the true global best replaces those only at successful completion, only at `result_clients`. With a `result_clients` subset override or a mid-run abort, a past aggregator keeps a file named "best" that isn't the global best. One sentence like "interim per-client best checkpoints are authoritative only after successful completion at result clients" would prevent surprises.

examples/advanced/swarm_learning/swarm_pt/README.md
```diff
@@ -75,20 +75,22 @@ The `QwenLoRAModelWrapper` in `model.py` overrides `state_dict()` / `load_state_
 import nvflare.client as flare
 
 flare.init()                          # 1. Initialize NVFlare Client API
-input_model = flare.receive()         # 2. Receive global LoRA adapter from server
+input_model = flare.receive()         # 2. Receive global LoRA adapter from the workflow
 apply_global_adapter(model, input_model.params)  # 3. Apply global adapter weights
 
+val_loss = evaluate_loss(model, validation_dataloader, validation_steps)
 adapter_diff = local_train(model, dataloader, steps)  # local fine-tuning
```
holgerroth: Snippet name mix: the new line above uses `validation_dataloader`/`validation_steps`, while this line keeps `dataloader`/`steps` — and this PR renames the actual variable to `train_dataloader` (client.py:239), so `dataloader` no longer matches anything in the file. `local_train(model, train_dataloader, local_steps)` keeps the snippet consistent. The numbered Peer-to-Peer Workflow list below also still omits the new validate step.

holgerroth:
@chesterxgchen to help prioritize my inline comments: the only one I would treat as a pre-merge ask for 2.9 is the opt-out/ownership point on the default selector (a named way to disable it, e.g. `key_metric=None`, plus adding `model_selector` to `_RECIPE_MANAGED_CLIENT_CONFIG_KEYS`, with a release-note line about the behavior change). That is public API surface that would be hard to change quietly after 2.9.0 ships. Everything else — the example hardening, docs wording, and test tweaks — works fine as fast-follows if you would rather keep this moving for the release. Overall the feature looks right and verified working end-to-end; nice work.

nvflare/app_opt/pt/recipes/swarm.py
chesterxgchen: Addressed in 6d4c71437. `key_metric=None` now omits the client-side selector, `model_selector` is protected as a recipe-managed client config field, and the 2.9 release note documents the default-on behavior and opt-out. I also regenerated the recipe catalog and added construction, override-rejection, and exported-config coverage. Validation: Swarm recipe + recipe CLI suites (79 passed) and `./runtest.sh -s` passed.

nvflare/app_opt/pt/recipes/swarm.py
chesterxgchen: Addressed in ec885f670. The recipe API docstring, release notes, recipe guide, and example README now state that `key_metric_mode="min"` exposes the negated comparison value in Swarm best-metric logs and records.

nvflare/app_opt/pt/recipes/swarm.py
chesterxgchen: Addressed in ec885f670. `key_metric` now fails fast when empty or whitespace-only while retaining `None` as the explicit opt-out, with parameterized recipe coverage.

nvflare/app_common/widgets/intime_model_selector.py
chesterxgchen: Addressed in ec885f670. Metric coercion now matches the guarded sibling-selector behavior, and non-finite values are warned and skipped before aggregation. Regression coverage includes list/string/None and NaN/Inf inputs.

tests/unit_test/app_common/widgets/in_time_model_selector_test.py
chesterxgchen: Addressed in ec885f670. The NumPy coercion regression now uses exact `type(...) is float` assertions for selected, raw, and published metrics, so it remains meaningful across NumPy versions.

examples/advanced/swarm_learning/swarm_pt/client.py
chesterxgchen: Addressed in ec885f670. Both job and client parsers now reject non-positive validation steps, validation loaders retain partial batches with `drop_last=False`, and the empty-loader diagnostic no longer misstates that only complete batches count.

examples/advanced/swarm_learning/swarm_pt/client.py
chesterxgchen: Addressed in ec885f670. The error now describes the required generic train/validation layout without directing custom-corpus users to regenerate WikiText, and the README explicitly documents the `--data_dir` contract for custom corpora.

examples/advanced/swarm_learning/swarm_pt/client.py
chesterxgchen: Addressed in ec885f670. Tokenization now assigns `-100` to labels wherever `attention_mask` marks padding, so causal-LM validation loss excludes padded eos positions.

examples/advanced/swarm_learning/swarm_pt/client.py
chesterxgchen: Addressed in ec885f670. The validation dataset is capped to `validation_steps * batch_size` before tokenization, while retaining a smaller available validation split.

docs/release_notes/flare_290.rst
chesterxgchen: Addressed in ec885f670. The example now skips round-0 evaluation, and the release notes and guides state that selection starts after round 0 and a one-round job does not create a best-model checkpoint.

docs/user_guide/data_scientist_guide/available_recipes.rst
chesterxgchen: Addressed in ec885f670. The recipe guide now explains that rotating aggregation clients can write interim per-client best checkpoints and that these are authoritative only after successful completion at result clients.

examples/advanced/swarm_learning/swarm_pt/README.md
chesterxgchen: Addressed in ec885f670. The README snippet now uses `train_dataloader` and `local_steps`, mirrors the round-0 metric skip, and includes validation in the numbered peer-to-peer workflow.

YuanTingHsieh:
Re-reviewed at 594a6a8c2 — all the earlier asks check out in code (I re-verified each fix rather than the replies, re-ran the swarm recipe + selector suites at head, and confirmed the new selector tests fail with the coercion reverted). Code-wise I have nothing left. Two doc lines are still wrong, both cheap to fix in this PR:

**1. The documented export command doesn't exist** — `examples/advanced/swarm_learning/swarm_pt/README.md:245` (Step 5) and the `job.py:25` module docstring both instruct:

```
python job.py --export --export-dir /tmp/swarm_lora_job
```

but `define_parser()` defines neither `--export` nor `--export-dir`, so the command exits with `error: unrecognized arguments`. Pre-existing lines, but this PR rewrote the README end-to-end and added an export unit test, which makes the claim look freshly validated. Either add the two flags (wiring `recipe.export(...)`) or drop/replace the two lines.

**2. The key-matching contract is overstated for the Job-API/JSON surface** — `docs/programming_guide/controllers/client_controlled_workflows.rst:424` says "The metric dictionary key must match the recipe's ``key_metric``", but `IntimeModelSelector` applies `key_metric` only when the reported metric is a dict (`intime_model_selector.py` — scalar `INITIAL_METRICS` is accepted unconditionally, and the new `float()` coercion happily accepts it). That matters because this sentence sits in the section documenting the Job-API/JSON configuration, where classic Learner-based executors emit *scalar* `INITIAL_METRICS` — including an injected placeholder `0` when the learner's validate result lacks the key (`learner_executor.py:114`, `.get(MetaKey.INITIAL_METRICS, 0)`). Paired with such an executor, `key_metric` is dead weight and round 1 always records a "best" (0 beats −inf), producing a `best_FL_global_model.pt` the release note says cannot exist. The recipe/ScriptRunner path is safe (`FLModel` enforces dict metrics), so this is a docs-scoping fix, e.g.: "When metrics are reported as a dictionary, the key must match ``key_metric``; a scalar metric value is used as-is regardless of ``key_metric``. Legacy Learner-based executors report a scalar and inject 0 when the metric is missing — do not rely on omitting the metric to disable selection on that path."

The learner-side placeholder-0 behavior itself is pre-existing and out of scope here; happy to see it tracked as a follow-up instead.

YuanTingHsieh (approved):

holgerroth (approved):
All review feedback is resolved as of 594a6a8c2 — the `key_metric=None` opt-out with `model_selector` in the recipe-managed keys settles the API-surface question for 2.9, the release note now covers the behavior change, round-0 condition, and min-mode display, and the guarded/finite coercion, test hardening, and example fixes (round-0 skip via workflow round, label masking, drop_last/validation-steps validation) all check out. The remaining items are non-blocking follow-ups (mode-aware comparator for raw best-metric records; the duplicated `positive_int` helper in the example). LGTM — thanks for the thorough turnaround, @chesterxgchen!

ZiyueXu77 (approved):

PR #5163, opened 2026-08-18
Honor configured internal listener host

nvidianz:
## Summary

- honor an explicitly configured `internal.resources.host` instead of replacing it with the localhost loopback default
- preserve an explicitly configured `listen_host` when applying the default internal listener host
- add regression coverage for the configured wildcard host and fixed-port case

## Root cause

`ConnectorManager` loaded `comm_config.json` and then unconditionally applied `internal_listener_host` to both `host` and `listen_host`. When a server service target was `localhost`, this silently replaced an explicit internal host such as `0.0.0.0` with `127.0.0.1` while leaving the configured port intact.

The loopback value is now applied only as a default when the internal host was not explicitly configured.

## Impact

Explicit internal listener configuration is honored again, including deployment configurations where job containers must reach the parent cell. Localhost-based environments without an explicit host continue to bind their internal listener to loopback.

## Testing

- `python -m pytest -q tests/unit_test/fuel/f3/cellnet/core_cell_test.py tests/unit_test/fuel/f3/drivers/net_utils_test.py tests/unit_test/private/fed/server/fed_server_test.py` (72 passed)
- `./runtest.sh -s` under Python 3.12
- focused reproduction produces connect URL `tcp://0.0.0.0:19100` and wildcard listen URL `tcp://0:19100`


YuanTingHsieh (dismissed):
LGTM, please resolve the conflicts thanks

nvidianz:
Resolved the conflicts with current `upstream/main` in 0bd0759e0, preserving both the configured-host precedence regression and the new resource-immutability test from `main`. Validation: 32 focused CoreCell tests passed and full `./runtest.sh -s` passed.

YuanTingHsieh (approved):

PR #5165, opened 2026-08-18
Drop late streamed replies without routing errors

nvidianz:
## Summary

- mark streamed response transport frames as optional so a reply that outlives its requester is dropped without ERROR-level routing noise
- log asynchronous undeliverable response diagnostics at DEBUG
- preserve reliable response delivery, receiver-side stream error propagation, and oversized server-job response fail-fast behavior
- add focused regression coverage for late response handling

## Root cause

A Collab caller can time out and remove its request waiter while the remote method continues running. If the server workflow exits before that method returns, the server job cell route is also removed. The later client response is sent as a reliable stream whose outer data frames were marked required, so the root server reported `cannot forward req: no path` at ERROR even though the caller had already received the expected timeout.

Response streams are now transport-optional. This affects transport logging only: active callers still receive the response or the correlated stream error, and reliable delivery remains enabled.

## Impact

Expected late replies following a handled Collab timeout no longer create false server/client error events. The workflow continues to receive `CollabCallError` for the original timeout.

## Testing

- `python -m pytest -q tests/unit_test/collab/runtime tests/unit_test/fuel/f3/cellnet` (225 passed)
- focused stream-adapter regressions (2 passed)
- `./runtest.sh -s` under Python 3.12

Jira: [FLARE-3134](https://jirasw.nvidia.com/browse/FLARE-3134)


nvflare/fuel/f3/streaming/byte_streamer.py
```diff
@@ -274,17 +281,34 @@ def __init__(
...
 
+    @staticmethod
+    def _new_send_error(msg: str, send_error=None) -> StreamError:
+        if send_error == ReturnCode.TARGET_UNREACHABLE:
+            return StreamTargetUnreachable(msg)
+        return StreamError(msg)
+
+    def _new_pending_error(self, msg: str, seq=None) -> StreamError:
+        if seq is not None:
+            return self._new_send_error(msg, self.pending_send_errors.get(seq))
```
YuanTingHsieh: Could we add a regression test for the actual routed topology before merging? The current test makes the sender’s local `fire_and_forget()` return `TARGET_UNREACHABLE`, but in the reported scenario the responder still has a valid first hop to the root server. That send succeeds locally; the root later finds the destination job cell missing and drops the fire-and-forget frame, so no unreachable result reaches `TxTask`.

With `pending_send_errors[seq]` remaining `None`, retry exhaustion produces a generic `StreamError`, and `_handle_reply_stream_done()` still logs it at ERROR level on the responder. This appears to remove the server routing ERROR while potentially leaving a delayed client ERROR.

Please test first-hop success plus downstream route removal and assert that the expected late reply creates no ERROR-level server or client logs. If this reproduces, the routed failure needs to be propagated or otherwise correlated as a stale reply rather than relying only on the sender’s local send result.

YuanTingHsieh (commented):
One routed late-response case still needs coverage.

nvflare/fuel/f3/streaming/byte_streamer.py
nvidianz: Addressed in fd1751a6b. Forwarding cells now report downstream stream-routing failures to the original sender through the existing stream-error topic, preserving the failed receiver for correlation. `TARGET_UNREACHABLE` is surfaced as `StreamTargetUnreachable`, so a stale reply remains diagnostic-only; other forwarding failures remain regular stream errors. I added a real multiprocess CoreCell/ConnManager regression that connects the destination, removes it, verifies the sender’s first hop to `server` succeeds, and asserts no ERROR-level server or client-callback logs. Focused tests: 151 passed; full `./runtest.sh -s`: passed.

nvflare/fuel/f3/streaming/byte_streamer.py
```diff
@@ -728,8 +804,14 @@ def _purge_error_contexts(cls, now: float):
...
     @staticmethod
     def _matches_error_context(message: Message, context: _TxTaskContext) -> bool:
+        origin = message.get_header(MessageHeaderKey.ORIGIN)
+        receiver = message.get_header(StreamHeaderKey.ERROR_RECEIVER, origin)
         return (
-            message.get_header(MessageHeaderKey.ORIGIN) == context.target
+            receiver == context.target
+            and (
+                origin == context.target
+                or message.get_header(StreamHeaderKey.ERROR_TYPE) == StreamTargetUnreachable.__name__
```
YuanTingHsieh: **P2 — Generic forwarded errors cannot correlate.** `_forward_error_handler()` emits an ordinary `StreamError` for downstream failures such as `COMM_ERROR`, and sending that notification sets its origin to the forwarding cell. Here `origin != context.target`, while the forwarding exception is allowed only for `StreamTargetUnreachable`, so every generic forwarded error is rejected as uncorrelated and an active stream waits for retry exhaustion instead of failing promptly. Please either accept appropriately correlated forwarded errors generally or emit feedback only for the supported unreachable case, and add coverage for a downstream `COMM_ERROR`.

nvflare/fuel/f3/streaming/byte_streamer.py
```diff
@@ -274,17 +281,34 @@ def __init__(
...
+    @staticmethod
+    def _new_send_error(msg: str, send_error=None) -> StreamError:
+        if send_error == ReturnCode.TARGET_UNREACHABLE:
+            return StreamTargetUnreachable(msg)
+        return StreamError(msg)
+
+    def _new_pending_error(self, msg: str, seq=None) -> StreamError:
+        if seq is not None:
+            return self._new_send_error(msg, self.pending_send_errors.get(seq))
+        if self.pending_send_errors and all(
```
YuanTingHsieh: **P2 — Protect pending-error inspection with `retry_lock`.** `send_loop()` calls `_new_pending_error()` without holding the lock, while `handle_ack()` can remove entries and the retry scheduler can update them concurrently. Iterating this live dictionary can raise `dictionary changed size during iteration` or classify the timeout from inconsistent state. Please inspect the mapping under `retry_lock` (or take a locked snapshot); the lock is an `RLock`, so callers already holding it remain safe.

YuanTingHsieh (commented):
Two remaining P2 issues after the routed late-response fix.

nvflare/fuel/f3/streaming/byte_streamer.py
nvidianz: Addressed in 944f7607f. Forwarded stream errors now correlate through the explicit ERROR_RECEIVER header for all error types, while still requiring the stream ID, channel, topic, and request ID context to match. A downstream COMM_ERROR regression verifies that the active sender fails promptly with a regular StreamError rather than waiting for retry exhaustion.

nvflare/fuel/f3/streaming/byte_streamer.py
nvidianz: _new_pending_error() now inspects pending_send_errors while holding retry_lock; the existing RLock keeps calls from retry paths safe. Added regression coverage that asserts both keyed lookup and values iteration occur while the lock is owned. Addressed in 944f7607f. Validation: 153 focused tests and full ./runtest.sh -s passed.

YuanTingHsieh:
The latest change introduces a regression for non-optional reliable streams. In a three-cell route with a transient 3-second next-hop outage, the stream previously recovered and delivered in about 5.01 seconds. After this change, the first forwarded `COMM_ERROR` is correlated through `ERROR_RECEIVER`, immediately terminates the stream at t=0, and fast-fails the live waiter instead of allowing the reliable retry mechanism to recover.

Please gate the bounced error—or the sender-side stream termination and waiter overwrite—on `MessageHeaderKey.OPTIONAL`. `_forward_error_handler` already reads this header. Late optional replies should still fail promptly with `StreamTargetUnreachable`, while non-optional reliable streams must retain their retry-until-`retry_timeout` contract.

Please also add a regression test where a transient no-route window heals within `retry_timeout` and a non-optional reliable stream successfully delivers.

nvidianz:
Addressed in 76c05d3cf. Forwarded route errors are now bounced only for optional streams, so non-optional reliable streams retain their retry-until-retry_timeout contract. Added both focused gating coverage and a live multiprocess CoreCell/ConnManager regression: the first downstream route attempt fails, the receiver reconnects within retry_timeout, and the reliable stream successfully delivers on retry. Validation: 155 focused tests and full ./runtest.sh -s passed.

YuanTingHsieh (dismissed):

nvflare/fuel/f3/cellnet/cell.py
```diff
@@ -160,6 +160,14 @@ def _handle_reply_stream_done(self, reply_future):
             )
             os._exit(1)
 
+        if isinstance(error, StreamTargetUnreachable):
+            # A response can finish after the requester has timed out and removed its waiter or job cell. The
+            # requester already owns the request outcome, so an unreachable target is diagnostic only.
+            self.logger.debug(
+                f"streamed response from {self.my_info.fqcn} was not delivered: {secure_format_exception(error)}"
+            )
+            return
```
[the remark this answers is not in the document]
nvidianz: Fixed in d2cd1f54d. The response completion callback now downgrades `StreamTargetUnreachable` only when the original request was explicitly optional. Required response delivery failures remain ERROR-level diagnostics. Added focused coverage for both optional and required cases, and updated the routed optional-stream test to pass the request context through the callback. All 69 targeted tests and the full `./runtest.sh -s` style suite pass.

YuanTingHsieh:
**P1 — The default Collab timeout path still produces a responder-side ERROR.**

`_send_response()` sends the response stream with transport `optional=True`, but registers `_handle_reply_stream_done()` with `optional`, which is the original request's flag. A normal Collab call uses `CallOption.optional=False`. In the default timeout sequence, the request times out, its job-cell route is removed, and the late response finishes with `StreamTargetUnreachable`; the callback therefore receives `optional=False` and logs the terminal result at ERROR. This suppresses the relay/server routing ERROR but still violates the clean client-log requirement.

The updated routed test avoids this path by manually passing `True` to `_handle_reply_stream_done()`. Please add a regression using a default/non-optional Collab request (or an equivalent request with `MessageHeaderKey.OPTIONAL=False`) that times out before the remote result returns, then assert that the late response produces no ERROR-level server or responder logs.

The original request's `optional` flag is not a proxy for whether its waiter or job-cell route still exists. Please classify this using the response transport/lifecycle context instead.

nvidianz:
Addressed in 4c744bdb0. The responder no longer uses the original request optional flag as a proxy for lifecycle state. Request streams now carry their wait budget, and the responder establishes a local monotonic deadline when dispatch begins. A StreamTargetUnreachable is downgraded only when the response was produced after that deadline; failures while the requester should still be waiting remain ERROR-level. Progress-extended requests intentionally omit this fixed deadline. Added coverage for a default optional=False request whose result is produced after timeout, plus coverage that the wait budget is propagated. All 89 targeted tests and the full ./runtest.sh -s suite pass.

nvflare/fuel/f3/cellnet/cell.py
```diff
@@ -160,11 +166,19 @@ def _handle_reply_stream_done(self, reply_future):
             )
             os._exit(1)
 
+        if response_was_late and isinstance(error, StreamTargetUnreachable):
+            # A response produced after the request's wait budget can outlive its requester or job cell. The
+            # requester already owns the timeout outcome, so an unreachable late target is diagnostic only.
+            self.logger.debug(
+                f"streamed response from {self.my_info.fqcn} was not delivered: {secure_format_exception(error)}"
+            )
+            return
```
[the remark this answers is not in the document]
nvidianz: Addressed in a3fd5fc75 by removing elapsed-time classification entirely. The completion callback now uses the response stream transport context: streamed responses are deliberately transport-optional because they may outlive the requester, while request-level failure remains owned by the requester through its correlated error or timeout. The callback cannot reliably infer remote waiter state, so StreamTargetUnreachable on this optional transport is diagnostic; other stream failures remain ERROR-level. The default optional=False request regression exercises this actual path. All 88 focused tests and the full style suite pass.

nvidianz:
Follow-up: a3fd5fc75 simplifies the fix to use the response transport context directly, as requested. It removes the local deadline heuristic because it cannot authoritatively represent the remote waiter lifecycle. Default Collab requests still use optional=False, but their response stream is intentionally transport-optional; an unreachable response is therefore diagnostic while the requester owns the timeout/error outcome. The focused default-request regression remains, and 88 targeted tests plus the full style suite pass.

nvflare/fuel/f3/streaming/stream_const.py
```diff
@@ -54,6 +54,7 @@ class StreamHeaderKey:
     OFFSET = STREAM_PREFIX + "os"
     ERROR_MSG = STREAM_PREFIX + "em"
     ERROR_TYPE = STREAM_PREFIX + "et"
+    ERROR_RECEIVER = STREAM_PREFIX + "er"
```
YuanTingHsieh: Nit before this ships: rename `StreamHeaderKey.ERROR_RECEIVER` -> `FAILED_DESTINATION` (and optionally the wire value `stream_er` -> `stream_fd`). The header holds the `DESTINATION` of the frame that could not be forwarded, not the receiver of the error — `ERROR_RECEIVER` reads as the cell the error is delivered to, which here is the sender. Since the header is introduced by this PR, the rename (including the wire string) is free now; after release it becomes a mixed-version compatibility surface between relays and senders. Optionally the `receiver=` fields in the "Ignored uncorrelated stream error" log lines could follow to `failed_destination=` for the same clarity.

nvidianz: Addressed in c06f9f315. Renamed the new header to StreamHeaderKey.FAILED_DESTINATION and changed its wire key to stream_fd before release. Related variables and diagnostic fields now use failed_destination consistently. The focused streaming, Cell, and Collab suite passes with 110 tests.

YuanTingHsieh (approved):

Issue #142, opened 2022-01-22
shell scripts missing x permission in poc


IsaacYangSLA:
The shell script files generated from poc command do not have original permission settings, especially the execute permission, after switching to shutil.unpack_archive.

PR #141, opened 2022-01-22
Fix create_analytic_dxo method calling

YuanTingHsieh:

PR #5170, opened 2026-08-19
Harden PyTorch-family conversion skills and artifact resolution

chesterxgchen:
## Summary

- harden Hugging Face model and dataset artifact resolution for offline caches and explicitly authorized downloads
- make PyTorch, Lightning, and Hugging Face conversion routing, model construction, topology ownership, export behavior, and validation consistent
- prevent generated jobs from redefining NVFLARE-reserved export arguments or losing audited model/site configuration during recovery
- tighten Lightning validation sequencing and persistent site-local state guidance
- add behavioral regression coverage for the conversion contracts

### Artifact resolution

- Add a maintained resolver for local artifacts and Hugging Face model or dataset repositories.
- Require callers to distinguish local paths from Hub IDs and resolve relative local paths against an explicit absolute source root.
- Report expected cache misses as structured exit-zero results.
- For authorized Hub downloads, resolve and validate a full commit SHA before calling `snapshot_download()`.
- Support older compatible `huggingface_hub` releases through the public `huggingface_hub.utils.LocalEntryNotFoundError` export.
- Keep documentation and evaluations aligned: a cache-only exception mentioning `huggingface.co` does not by itself prove network access.

### Conversion behavior

- Route directly to a single PyTorch-family converter when the user names the framework or preliminary inspection identifies one owner; retain `nvflare-orient` for unresolved or conflicting ownership.
- Preserve required and overridden constructor values through explicit `class_path`/`path` plus complete `args`; allow direct instances only when unchanged zero-argument defaults reconstruct the model.
- Centralize the single simulator-topology-owner contract in `conversion-common.md` and rely on `SimEnv` for named-client/count consistency and default thread handling.
- Make explicit export opt-in. Local `python job.py` validation uses the workspace materialized by Recipe/`SimEnv`; generated parsers must not declare or interpret reserved `--export` or `--export-dir` arguments.
- Select one final validation target and preserve inspected argument types, guards, fallbacks, and correct values during retries.
- Prevent the three observed partition-probe recoveries by preserving annotated `Path` values, normalizing CSV-reloaded frame dtypes before equality checks, and retaining complete local import closures in isolated checks.
- Prevent local Lightning conversions from probing unsupported `Recipe`, `nvflare.recipe.run`, export, run, or lifecycle APIs after `recipe show`; use the concrete recipe with `recipe.execute(SimEnv(...))`.
- Require Lightning to load both validation references immediately after file generation and before any preflight, smoke test, cleanup, validation, or execution command.
- Clarify that persistent Lightning site-local state requires supported filters in both network directions; aggregation-only `exclude_vars` is insufficient.

## Validation

- `python3 -m pytest tests/unit_test/tool/agent_skill_checks/conversion_templates_test.py tests/unit_test/tool/agent_skill_checks/seed_skills_test.py -q --deselect tests/unit_test/tool/agent_skill_checks/seed_skills_test.py::test_pytorch_recipe_capability_profiles_match_tensor_disk_offload_support` — 147 passed, 1 skipped, 1 deselected; the deselected XGBoost catalog test requires unavailable local `libomp`
- focused resolver, routing, simulator-topology, validation-recovery, and Lightning bidirectional-filter regression tests passed
- `python3 dev_tools/agent/skills/checks/cli.py --skills-root skills` — 0 findings
- Bandit B615 against the resolver — passed
- `./runtest.sh -s` — Black, isort, Flake8, and agent-skill lint passed
- pre-push agent-skill lint passed

## Review follow-up

- Addresses the unpinned `snapshot_download()` finding by passing the validated immutable revision through an explicit `revision=` keyword visible to static analysis.
- Addresses caller-working-directory ambiguity through the explicit source-root contract and fresh-working-directory coverage.
- Addresses the Lightning site-local-state finding with bidirectional filtering guidance and two-round regression coverage.
- Addresses the dataset-repository, offline-eval, routing-symmetry, validation-sequencing, topology-duplication, and `SimEnv` helper findings.
- Addresses the two-machine Lightning validation recoveries without changing Lightning workflow or dataset behavior.

### DIFF
--- a/skills/nvflare-convert-huggingface/references/huggingface-validation.md
+++ b/skills/nvflare-convert-huggingface/references/huggingface-validation.md
@@ -64,15 +64,18 @@ reaches the applicable phase, and stop at the first failure.
   rerun the Hub resolver once with the complete canonical invocation:
 
   ```bash
-  python <skill-dir>/scripts/resolve_model_snapshot.py --source hub --allow-download --revision <commit-sha> <org/model>
+  python <skill-dir>/scripts/resolve_model_snapshot.py --source hub --allow-download <org/model>
   ```
 
-  The revision is the full immutable 40-character commit SHA. Use its
-  immutable `resolved_path` for the server and every client. Do not run a
-  preceding `snapshot_download(..., local_files_only=True)` probe, copy resolver
-  logic into generated `job.py`, or download without authorization. Resolve
-  datasets through their source-prescribed local path or cache with the same
-  policy.
+  In that authorized invocation, the resolver uses the public
+  `HfApi().model_info(...).sha` result, validates the full immutable
+  40-character commit SHA, and passes it to `snapshot_download`. An already
+  audited full SHA may instead be supplied with `--revision`. Use the returned
+  immutable `revision` and `resolved_path` for the server and every client. Do
+  not run a preceding `snapshot_download(..., local_files_only=True)` probe,
+  copy resolver logic into generated `job.py`, or download without
+  authorization. Resolve datasets through their source-prescribed local path or
+  cache with the same policy.
 - Never recover from an offline/cache-only miss by removing
   `HF_HUB_OFFLINE`/`TRANSFORMERS_OFFLINE`, dropping `local_files_only=True`, or
   rerunning online unless the user explicitly requested the download. Do not
--- a/skills/nvflare-convert-huggingface/references/huggingface-validation.md
+++ b/skills/nvflare-convert-huggingface/references/huggingface-validation.md
@@ -54,33 +54,44 @@ reaches the applicable phase, and stop at the first failure.
   ```bash
   python <skill-dir>/scripts/resolve_model_snapshot.py --source local --source-root <absolute-source-root> <configured-path>
   python <skill-dir>/scripts/resolve_model_snapshot.py --source hub <org/model>
+  python <skill-dir>/scripts/resolve_model_snapshot.py --source hub --repo-type dataset <org/dataset>
   ```
 
   A relative local identifier requires the absolute original source-project
   root so resolution never depends on the validation command's working
   directory. An absolute local identifier must omit `--source-root`; the option
   resolves relative identifiers and is not a sandbox boundary.
-- Only when the user authorizes downloading a public checkpoint if uncached,
-  rerun the Hub resolver once with the complete canonical invocation:
+- Only when the user authorizes downloading the specific public Hub artifact if
+  uncached, rerun its resolver once with the matching canonical invocation:
 
   ```bash
   python <skill-dir>/scripts/resolve_model_snapshot.py --source hub --allow-download <org/model>
+  python <skill-dir>/scripts/resolve_model_snapshot.py --source hub --repo-type dataset --allow-download <org/dataset>
   ```
 
-  In that authorized invocation, the resolver uses the public
-  `HfApi().model_info(...).sha` result, validates the full immutable
+  In that authorized invocation, the resolver uses the matching public
+  `HfApi().model_info(...).sha` or
+  `HfApi().dataset_info(...).sha` result, validates the full immutable
   40-character commit SHA, and passes it to `snapshot_download`. An already
-  audited full SHA may instead be supplied with `--revision`. Use the returned
-  immutable `revision` and `resolved_path` for the server and every client. Do
-  not run a preceding `snapshot_download(..., local_files_only=True)` probe,
-  copy resolver logic into generated `job.py`, or download without
-  authorization. Resolve datasets through their source-prescribed local path or
-  cache with the same policy.
+  audited full SHA may instead be supplied with `--revision`. For a model
+  repository, use the returned immutable `revision` and `resolved_path` for the
+  server and every client. For a dataset repository, pass `resolved_path` only
+  to a source-compatible local dataset loader; if the source cannot load that
+  snapshot layout, ask or fail closed rather than inventing a loader. Do not run
+  a preceding `snapshot_download(..., local_files_only=True)` probe, copy
+  resolver logic into generated `job.py`, or download without authorization.
+  For a configured Hub dataset ID, use the same resolver with
+  `--repo-type dataset`; for a source-prescribed local dataset path, use the
+  existing local path without a Hub lookup.
 - Never recover from an offline/cache-only miss by removing
   `HF_HUB_OFFLINE`/`TRANSFORMERS_OFFLINE`, dropping `local_files_only=True`, or
   rerunning online unless the user explicitly requested the download. Do not
   substitute another local checkpoint. Keep the job as a draft and report
   full-model validation blocked when the required artifact is unavailable.
+  A cache-only Hugging Face error can mention `huggingface.co` while reporting
+  a local cache miss; that URL alone is not evidence that a network request was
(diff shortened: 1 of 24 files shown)


chesterxgchen:
Reviews #141 and #142 are addressed in the latest commits. Artifact source is now explicit (`local` or `hub`), with regression coverage proving that missing relative local paths never contact the Hub and an existing `org/model` directory cannot override an explicit Hub source. Constructor guidance now requires the selected recipe’s documented `class_path` or `path` key plus complete `args`. The Cyclic regression exports a parameterized `class_path` input and verifies its server config is normalized to `path` with all arguments retained.

skills/nvflare-convert-huggingface/scripts/resolve_model_snapshot.py
```diff
@@ -0,0 +1,127 @@
...
+    cache_dir: Optional[str] = None,
+) -> Dict[str, Any]:
+    """Return structured resolution evidence; an expected cache miss is not an error."""
+    if source not in {"hub", "local"}:
+        raise ValueError("source must be 'local' or 'hub'")
+
+    if source == "local":
+        if allow_download or revision is not None or cache_dir is not None:
+            raise ValueError("Hub download, revision, and cache options cannot be used with source='local'")
+        candidate = Path(identifier).expanduser()
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in d6b74fa10. Relative local identifiers now require an explicit absolute `source_root` and resolve against that stable original-project base, while absolute local paths remain unchanged and Hub identifiers reject `source_root`. Added a CLI regression that changes to a different caller directory and verifies the expected source-relative model is selected. The full scoped conversion skill suites pass (139 passed, 1 skipped, with the unrelated XGBoost/libomp catalog test deselected), and `./runtest.sh -s` passes.

skills/nvflare-convert-huggingface/scripts/resolve_model_snapshot.py
```diff
@@ -0,0 +1,142 @@
...
+import re
+from pathlib import Path
+from typing import Any, Dict, Optional, Sequence
+
+_COMMIT_SHA_RE = re.compile(r"[0-9a-fA-F]{40}")
+
+
+def _load_huggingface_hub():
+    from huggingface_hub import snapshot_download
+    from huggingface_hub.errors import LocalEntryNotFoundError
```
holgerroth: `from huggingface_hub.errors import LocalEntryNotFoundError` fails on `huggingface_hub` < 0.25.0: `errors.py` didn't exist before 0.22.0, and `LocalEntryNotFoundError` only moved into it at 0.25.0 (verified against the huggingface_hub release tags). Nothing in the skill tree pins a minimum hub version, so on an older environment the canonical `--source hub` probe crashes with a `ModuleNotFoundError` traceback and a nonzero exit instead of the documented exit-zero `"status": "missing"` JSON — exactly the false-failure recovery evidence this resolver was written to eliminate.

Suggestion: import from `huggingface_hub.utils`, which re-exports `LocalEntryNotFoundError` in both old and current versions:

```python
from huggingface_hub.utils import LocalEntryNotFoundError
```

skills/nvflare-convert-huggingface/references/huggingface-validation.md
```diff
@@ -41,28 +41,45 @@ reaches the applicable phase, and stop at the first failure.
...
+  root so resolution never depends on the validation command's working
+  directory. An absolute local identifier does not require `--source-root`.
+- Only when the user authorizes downloading a public checkpoint if uncached,
+  rerun the Hub resolver once with the complete canonical invocation:
+
+  ```bash
+  python <skill-dir>/scripts/resolve_model_snapshot.py --source hub --allow-download --revision <commit-sha> <org/model>
+  ```
+
+  The revision is the full immutable 40-character commit SHA. Use its
```
holgerroth: The authorized-download path requires a full 40-character commit SHA (`resolve_model_snapshot.py` raises `ValueError` without one), but no skill document explains how the agent should *obtain* that SHA. A cache-only run returns `"status": "missing"` with no SHA, so on a cold cache there's a chicken-and-egg: the agent must fail closed on a fully authorized download, ask the user (against the hello-world prompt's "Proceed without asking for additional confirmation"), or improvise an undocumented online lookup (`HfApi().model_info().sha`) — the invented-API behavior the evals prohibit. The packaged `huggingface-conversion` example ships an unpinned `hf-internal-testing/tiny-random-distilbert`, so the quick-start hits this directly.

Suggestion: document one sanctioned SHA-discovery step for authorized downloads (e.g. a resolver flag that reports the current commit SHA for a repo ID as part of the authorized flow, or reading the `snapshots/<sha>` component of `resolved_path` after the pinned download), so the requirement is satisfiable without inventing an API.

holgerroth (commented):
Two findings from a review pass (both verified against the branch):

skills/nvflare-convert-huggingface/scripts/resolve_model_snapshot.py
chesterxgchen: Fixed on the current head in d32a6e686. The resolver imports LocalEntryNotFoundError from the public huggingface_hub.utils re-export, and the regression test loads the resolver against a legacy-style module layout with no huggingface_hub.errors module. The combined conversion suites pass (143 passed, 1 skipped, 1 unrelated XGBoost/libomp test deselected), and the full style/skill-lint gate passes.

skills/nvflare-convert-huggingface/references/huggingface-validation.md
chesterxgchen: Fixed in d32a6e686 without making the SHA a user-facing quick-start requirement. In the single authorized --allow-download invocation, the maintained resolver now calls the public HfApi.model_info(...).sha API, validates the returned full commit SHA, and passes that immutable revision to snapshot_download. An explicit --revision remains optional for an already-audited SHA. The canonical skill command and focused resolver tests cover the cold-cache path; users only authorize the public model download.

holgerroth (dismissed):
Verified the latest push: the resolver's self-resolving commit SHA via the public HfApi model-info lookup is the right call (keeps the download pinned while making cold-cache resolution completable), and the huggingface_hub.utils import restores compatibility with older hub versions — both confirmed against the branch with the updated tests. Remaining observations are non-blocking follow-ups. LGTM.

holgerroth (dismissed):
Re-reviewed the latest push (rebase onto main plus three new commits). All remaining follow-ups from the previous review are addressed: the resolver now supports Hub dataset repos via --repo-type with a matching dataset_info SHA lookup, the offline-error interpretation note is restored so docs and evals agree again, build_sim_env now delegates topology consistency to SimEnv's own validator, the single-topology-owner rule is consolidated into conversion-common.md with the other references pointing at it, and the pytorch/lightning skill descriptions gained the inspection-routing clause. Verified the full agent-skill test suite passes locally on this head (148 passed, 1 skipped). LGTM.

holgerroth (dismissed):
Reviewed the new commit (Prevent recipe lifecycle discovery probes): the Lightning skill prohibition on post-recipe-show lifecycle probing is consistent with the existing shared construction guidance, and the eval assertion, prohibited-behavior id, and seed-test pins move in lockstep. seed_skills_test.py passes locally on this head. Still LGTM.

holgerroth (approved):

PR #5171, opened 2026-08-19
Harden certless admin sessions and listener binding

nvidianz:
## Summary

- fail closed when an admin session token cannot be signature-verified
- reject admin login when the server signing identity is unavailable
- bind the admin listener to the explicit `admin_host` setting
- normalize localhost and IPv4/IPv6 loopback literals to IPv4 loopback
- reject unsupported non-loopback IPv6 admin bind literals before constructing a CellNet URL
- warn when an insecure admin listener is exposed on a non-loopback interface

## Root cause

`Session.decode_token()` treated the identity asserter as optional and decoded attacker-controlled session attributes without signature verification when it was absent. `LoginModule.pre_command()` could then recreate an authenticated session from that unverified token.

Separately, `BaseServer.deploy()` used the general listener host for the admin URL and ignored the explicit `admin_host` setting, causing the admin/HCI listener to bind more broadly than intended. The provisioned `admin_server` identity remains an advertised endpoint and is intentionally not used as a local bind address.

## Impact

Certless deployments can no longer authenticate or recreate sessions from unsigned or forged tokens. Admin listener exposure now follows the configured host, and insecure non-loopback exposure is made explicit at startup.

## Validation

- `574 passed`: `tests/unit_test/fuel/hci` and `tests/unit_test/private/fed/server`
- `54 passed`: focused final regression suite
- scoped Black, isort, Flake8, and `git diff --check` passed
- full `./runtest.sh -s` reached Black but was blocked by three pre-existing generated Python files under `node_modules`; no changed file failed style validation


pcnudde:
remove the jira reference please.

nvidianz:
Removed the Jira reference from the PR description.

pcnudde (dismissed):

nvflare/private/fed/server/fed_server.py
```diff
@@ -82,6 +83,35 @@
...
+    return _normalize_loopback_host(host) == "127.0.0.1"
+
+
+def _normalize_admin_host(host: str) -> str:
+    """Normalize a configured admin bind host and reject unsupported IPv6 listeners."""
+    normalized_host = _normalize_loopback_host(host)
+    try:
+        address = ipaddress.ip_address(host.strip("[]"))
+    except ValueError:
+        return normalized_host
```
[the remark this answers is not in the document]
nvidianz: Fixed in d9ef0e651. IPv6 parsing now trims surrounding whitespace before removing brackets, so padded non-loopback IPv6 literals are rejected before URL construction. I also added a focused regression test. The targeted server tests pass (42 tests), and scoped Black/isort/flake8 plus agent-skill lint pass. The full style command was also run; it is blocked only by three pre-existing generated node_modules Python files outside this change.

nvflare/private/fed/server/fed_server.py
```diff
@@ -82,6 +83,35 @@
...
+def _normalize_admin_host(host: str) -> str:
+    """Normalize a configured admin bind host and reject unsupported IPv6 listeners."""
+    normalized_host = _normalize_loopback_host(host)
+    try:
+        address = ipaddress.ip_address(host.strip().strip("[]"))
+    except ValueError:
+        return normalized_host
+    if isinstance(address, ipaddress.IPv6Address) and not address.is_loopback:
+        raise ValueError(f"IPv6 admin_host is not supported: {host}")
+    return normalized_host
```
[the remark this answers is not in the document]
nvidianz: Fixed in a901f4110. `_normalize_admin_host` now trims the configured value before normalization and URL construction, covering padded IPv4 addresses and hostnames as well as IPv6 literals. Added regression cases for both IPv4 and hostname inputs. All 44 focused tests and scoped Black/isort/flake8 plus agent-skill checks pass.

nvflare/private/fed/server/fed_server.py
```diff
@@ -172,14 +203,22 @@ def deploy(self, args, grpc_args=None, secure_train=False, enable_admin_listener
...
+            configured_admin_host = grpc_args.get("admin_host")
+            admin_host = _normalize_admin_host(configured_admin_host) if configured_admin_host else url_host
+            if not secure_train and not _is_loopback_host(admin_host):
+                self.logger.warning(
+                    f"insecure admin listener is exposed on non-loopback host '{admin_host}'; "
+                    "set secure_train=true or configure admin_host as a loopback address"
+                )
+            admin_url = f"{scheme}://{admin_host}:{admin_port}?{ADMIN_LISTENER_KEY}=true"
             if admin_port == fl_port:
                 root_url = [admin_url]
```
YuanTingHsieh, line 261: **[P1] Preserve FL reachability when the admin host shares the FL port**

With `service.target = server.example:8002`, `admin_host = 127.0.0.1`, and no explicit `admin_port`, `admin_port == fl_port` and this replacement makes the only Cell listener `tcp://127.0.0.1:8002?admin_listener=true`. Remote FL clients can no longer connect. This is also the configuration suggested by the warning above. Require a distinct `admin_port` whenever the admin bind host differs from the FL bind host, or reject this shared-port configuration before constructing `root_url`; please add a regression case for it.

nvflare/fuel/hci/server/sess.py
```diff
@@ -73,9 +73,11 @@ def make_token(self, id_asserter: IdentityAsserter):
         return f"{bds}:{signature}"
 
     @staticmethod
-    def decode_token(token: str, id_asserter: IdentityAsserter = None):
+    def decode_token(token: str, id_asserter: IdentityAsserter):
```
YuanTingHsieh, line 77: **[P1] Keep `_check_session` working after requiring an asserter**

`handle_check_session()` still calls `self.get_session(token)` without an identity asserter. That call now reaches `decode_token(token, None)`, the exception is swallowed by `get_session()`, and every valid live session is reported as `SESSION_INACTIVE`. Obtain the HCI server's asserter in the handler and pass it through, and add a regression test exercising the registered `_check_session` command with a signed live token.

YuanTingHsieh (commented):
Two blocking regressions remain on the current head. Both need focused regression coverage before merge.

nvflare/private/fed/server/fed_server.py
nvidianz: Fixed in f0b03f95b. A shared FL/admin port is now rejected when  differs from the FL bind host, preventing a loopback-only admin listener from replacing the externally reachable FL listener. Added focused regression coverage, and clarified the insecure-listener warning to require a distinct admin port.

nvflare/fuel/hci/server/sess.py
nvidianz: Fixed in f0b03f95b. handle_check_session now obtains the HCI server identity asserter from the connection and passes it to get_session. Added a regression that invokes the registered _check_session handler with a signed live token and verifies the OK response. All 54 focused tests and scoped style checks pass.

nvflare/private/fed/server/fed_server.py
```diff
@@ -82,6 +83,36 @@
...
+
+
+def _is_loopback_host(host: str) -> bool:
+    return _normalize_loopback_host(host) == "127.0.0.1"
+
+
+def _normalize_admin_host(host: str) -> str:
+    """Normalize a configured admin bind host and reject unsupported IPv6 listeners."""
+    host = host.strip()
+    normalized_host = _normalize_loopback_host(host)
```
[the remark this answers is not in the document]
nvidianz: Fixed in 025b3571a (current branch head b1ca28e39 after integrating the latest main update). Whitespace-only admin_host values now raise ValueError before URL construction, preventing an empty host from becoming a wildcard bind. Added focused regression coverage. All 46 server tests and scoped Black/isort/flake8 checks pass.

nvflare/private/fed/server/fed_server.py
```diff
@@ -172,14 +205,26 @@ def deploy(self, args, grpc_args=None, secure_train=False, enable_admin_listener
...
+        listening_host = _normalize_loopback_host(service_host) if _is_loopback_host(service_host) else None
         fl_port = int(parts[1])
 
         url_host = listening_host or "0"
         root_url = [f"{scheme}://{url_host}:{fl_port}"]
         if enable_admin_listener:
             admin_port = int(grpc_args.get("admin_port", fl_port))
-            admin_url = f"{scheme}://{url_host}:{admin_port}?{ADMIN_LISTENER_KEY}=true"
+            configured_admin_host = grpc_args.get("admin_host")
+            admin_host = _normalize_admin_host(configured_admin_host) if configured_admin_host else url_host
```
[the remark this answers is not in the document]
nvidianz: Fixed in 975f824ab. The deploy path now distinguishes an unset admin_host (None, which keeps the default bind) from an explicitly configured empty value (which is passed to validation and rejected). The regression is parameterized for both an empty string and whitespace-only input. All 47 focused server tests and scoped style checks pass.

nvflare/private/fed/server/fed_server.py
```diff
@@ -172,14 +203,22 @@ def deploy(self, args, grpc_args=None, secure_train=False, enable_admin_listener
...
+            if not secure_train and not _is_loopback_host(admin_host):
+                self.logger.warning(
+                    f"insecure admin listener is exposed on non-loopback host '{admin_host}'; "
+                    "set secure_train=true or configure admin_host as a loopback address"
+                )
+            admin_url = f"{scheme}://{admin_host}:{admin_port}?{ADMIN_LISTENER_KEY}=true"
```
YuanTingHsieh, line 259: **Blocker.** `admin_port` defaults to `fl_port` here, and provisioning defaults it the same way (`lighter/ctx.py:45`). On that branch `root_url = [admin_url]`, so a configured `admin_host` replaces the bind host of the *only* external listener — FL traffic included.

Repro: `service.target="server.example:8002"` + `"admin_host": "127.0.0.1"`, no `admin_port` → `root_url = ['tcp://127.0.0.1:8002?admin_listener=true']` — every remote FL client gets connection refused. The warning below tells operators to do exactly this.

Suggest failing fast (or ignoring `admin_host` with a warning) when `admin_port == fl_port` and `admin_host` differs from the service host. Also note: in secure mode the reverse direction (non-loopback `admin_host` silently rebinding a loopback FL target) gets no warning at all. Tests only cover `admin_host` with a distinct `admin_port`.

YuanTingHsieh (commented):
Round-2 review at a901f4110. The token-side fix is solid: I traced all wire-token paths (login, pre_command, get_session, recreate_session, logout) and they fail closed; the TokenVerifier binding and algorithm check out. The two new commits address whitespace trimming only.

Two blockers remain (inline): the shared-port `admin_host`/FL-bind coupling and the `_check_session` regression. Remaining inline comments are correctness fixes or non-blocking notes. All findings were reproduced against this head, not just read from the diff.

nvflare/fuel/hci/server/sess.py
```diff
@@ -73,9 +73,11 @@ def make_token(self, id_asserter: IdentityAsserter):
...
         if not isinstance(token, str):
             raise ValueError(f"token must be str but got {type(token)}")
+        if not id_asserter:
```
YuanTingHsieh, line 80: **Blocker.** This raise breaks `handle_check_session` (line 318): it still calls `self.get_session(token)` with the retained `id_asserter=None` default, the ValueError is swallowed by `get_session`'s `except Exception`, and the registered, filter-exempt `_check_session` command now always answers SESSION_INACTIVE.

Verified with a live session: `get_session(token, asserter)` returns it; `get_session(token)` returns None for the same valid token.

Fix: fetch the asserter from `conn.get_prop(ConnProps.HCI_SERVER)` in `handle_check_session`, and drop the `None` default on `get_session` so no future caller can hit this silently.

nvflare/private/fed/server/fed_server.py
```diff
@@ -172,14 +203,22 @@ def deploy(self, args, grpc_args=None, secure_train=False, enable_admin_listener
         if len(parts) != 2:
             raise RuntimeError(f"bad service target: {target}")
 
-        listening_host = "127.0.0.1" if parts[0].rstrip(".").lower() == "localhost" else None
+        service_host = parts[0]
+        listening_host = _normalize_loopback_host(service_host) if _is_loopback_host(service_host) else None
```
YuanTingHsieh, line 231: Behavior change without opt-in: the old line special-cased only the literal `localhost`; this maps *any* loopback literal to a `127.0.0.1` bind. A config with target `127.0.0.2:8002` previously bound wildcard and served clients dialing 127.0.0.2 — now they get connection refused, and multi-alias setups (127.0.0.2/127.0.0.3 sharing a port) collide on 127.0.0.1. Also breaks target `127.0.0.1` in containers with `docker -p` published ports (the proxy reaches the container via eth0, not loopback).

If tightening the bind for loopback targets is intended, bind the *configured* literal instead of collapsing everything to 127.0.0.1.

nvflare/private/fed/server/fed_server.py
```diff
@@ -82,6 +83,36 @@
...
+def _normalize_admin_host(host: str) -> str:
+    """Normalize a configured admin bind host and reject unsupported IPv6 listeners."""
+    host = host.strip()
+    normalized_host = _normalize_loopback_host(host)
+    try:
+        address = ipaddress.ip_address(host.strip("[]"))
+    except ValueError:
+        return normalized_host
+    if isinstance(address, ipaddress.IPv6Address) and not address.is_loopback:
+        raise ValueError(f"IPv6 admin_host is not supported: {host}")
```
YuanTingHsieh: Three issues with this rejection:

1. It only catches IP literals — a hostname takes the `except ValueError` path and bypasses the guard. Under the default `grpc` scheme a failed admin bind is logged only at DEBUG (`grpc_driver.py:191-193`), so an AAAA-only or unresolvable `admin_host` leaves the server running with **no admin listener at all**, or lets IPv6-capable gRPC bind the very family this rejects.
2. `::ffff:127.0.0.1` is an `IPv6Address` whose `is_loopback` is False before Python 3.13 — the same config crashes on 3.10–3.12 and works on 3.13+ (repo supports 3.10–3.14).
3. The ValueError escapes `server_train.py`'s `except ConfigError` and exits via mpm's generic EXCEPTION path with a raw traceback. Raising `ConfigError` classifies the misconfiguration correctly.

nvflare/private/fed/server/fed_server.py
```diff
@@ -82,6 +83,36 @@
...
+    """Return the IPv4 loopback address for localhost and loopback IP literals."""
+    if not host:
+        return host
+    if host.rstrip(".").lower() == "localhost":
+        return "127.0.0.1"
+    try:
+        address = ipaddress.ip_address(host.strip().strip("[]"))
+    except ValueError:
+        return host
+    return "127.0.0.1" if address.is_loopback else host
```
YuanTingHsieh, line 113: Silent coercion: an explicitly configured `admin_host` of `127.0.0.2` or `::1` is rewritten to `127.0.0.1` with no log line, so the listener binds a different address than configured — and for `::1` no IPv6 listener exists at all (a console dialing `[::1]:8003` is refused). Inconsistent with the loud ValueError for non-loopback IPv6 below. At minimum log the coercion; better, honor the configured loopback literal.

nvflare/private/fed/server/fed_server.py
```diff
@@ -172,14 +203,22 @@ def deploy(self, args, grpc_args=None, secure_train=False, enable_admin_listener
...
         fl_port = int(parts[1])
 
         url_host = listening_host or "0"
         root_url = [f"{scheme}://{url_host}:{fl_port}"]
         if enable_admin_listener:
             admin_port = int(grpc_args.get("admin_port", fl_port))
-            admin_url = f"{scheme}://{url_host}:{admin_port}?{ADMIN_LISTENER_KEY}=true"
+            configured_admin_host = grpc_args.get("admin_host")
+            admin_host = _normalize_admin_host(configured_admin_host) if configured_admin_host else url_host
+            if not secure_train and not _is_loopback_host(admin_host):
```
YuanTingHsieh: This warning fires on every default insecure deployment: with no `admin_host` configured it falls back to `url_host = "0"`, which `_is_loopback_host` can't classify — that's why `test_production_listener_bindings_remain_wildcard_by_default` needed `server.logger = MagicMock()`. It also false-positives on genuinely-loopback hosts (`127.1`, /etc/hosts aliases). And the remediation it suggests is a trap on shared-port kits (see comment at line 221) for a key that is undocumented and never provisioned.

Suggest gating it on an explicitly configured `admin_host`, and documenting the new key.

nvflare/fuel/hci/server/login.py
```diff
@@ -172,7 +176,9 @@ def handle_logout(self, conn: Connection, args: List[str]):
         if self.session_mgr:
             token = conn.get_prop(ConnProps.TOKEN)
             if token:
-                self.session_mgr.end_session_by_token(token)
+                hci = conn.get_prop(ConnProps.HCI_SERVER)
+                id_asserter = hci.get_id_asserter() if hci else None
+                self.session_mgr.end_session_by_token(token, id_asserter)
```
YuanTingHsieh: The handler still replies "OK" when this silently no-ops: `end_session_by_token` swallows the decode ValueError (asserter unavailable) and skips on a failed verify (e.g. server cert rotated), leaving the session alive until idle timeout.

Simpler fix that also deletes the plumbing: `pre_command` already verified this token and stored the session via `conn.set_prop(ConnProps.SESSION, sess)` (LOGOUT is not filter-exempt), so use `conn.get_prop(ConnProps.SESSION)` + `end_session_by_id(...)` — then `end_session_by_token` (this is its only caller) and the new asserter threading can be removed entirely.

nvflare/fuel/hci/server/sess.py
```diff
@@ -174,6 +175,8 @@ def create_session(self, user_name, user_org, user_role, origin_fqcn, active_stu
 
     def recreate_session(self, token: str, origin_fqcn, id_asserter: IdentityAsserter):
```
YuanTingHsieh, line 177: Non-blocking, pre-existing — but worth a follow-up ticket since this PR rewrites these exact lines: any validly-signed token is resurrectable. After logout or idle expiry, the next request falls through `pre_command` to `recreate_session`, which unconditionally re-mints the session; there is no revocation list, jti, or issued-at check, so logout cannot invalidate a captured token until the admin cert expires. Verified end-to-end: create → logout (session popped) → `recreate_session` re-inserts the same sess_id/user/role.

nvflare/private/fed/server/fed_server.py
```diff
@@ -82,6 +83,36 @@
...
+    return "127.0.0.1" if address.is_loopback else host
+
+
+def _is_loopback_host(host: str) -> bool:
+    return _normalize_loopback_host(host) == "127.0.0.1"
+
+
+def _normalize_admin_host(host: str) -> str:
+    """Normalize a configured admin bind host and reject unsupported IPv6 listeners."""
+    host = host.strip()
```
YuanTingHsieh, line 128: The strip fixes whitespace, but a non-string value still crashes opaquely: `"admin_host": 8003` (easy confusion with `admin_port`, which *is* an int in provisioned configs) → `AttributeError: 'int' object has no attribute 'strip'` via the generic exception exit. `host = str(host).strip()` — or an isinstance check raising `ConfigError` — finishes this fix.

nvflare/private/fed/server/fed_server.py
```diff
@@ -82,6 +83,36 @@
 from .server_status import ServerStatus
 
 
+def _normalize_loopback_host(host: str) -> str:
```
YuanTingHsieh, line 99: Non-blocking: this is now the fourth in-repo copy of loopback normalization, and it already disagrees with the F3-layer copy it descends from — `net_utils.get_tcp_urls` maps `::1` → wildcard `"0"` and preserves `127.0.0.2`, while this maps both → `127.0.0.1` (`external_process_backend.py:242` carries a third semantics). The whitespace fix also had to land in two places here and still misses the `localhost` string check at line 90 on the service-target path.

Suggest consolidating into a shared helper in `net_utils.py` — the driver layer already has the `LISTEN_HOST` mechanism — which would cover client/relay/adhoc listeners too.

tests/unit_test/fuel/hci/session_study_test.py
```diff
@@ -29,6 +31,24 @@ def sign(data, return_str=True):
...
+
+    @staticmethod
+    def verify(_nonce, _data, signature):
+        return signature == "signature"
+
+
+class _FakeCell:
+    pass
+
+
```
YuanTingHsieh, line 51: Non-blocking: with this autouse fake plus `_FakeIdAsserter.sign`, no unit test anywhere exercises the real `IdentityAsserter.sign` ↔ `TokenVerifier.verify` pair for session tokens — in a PR whose point is that signature verification gates decode, every test stubs it out. One round-trip test with a real self-signed cert would close that. Also note the autouse fixture silently applies to any test later added to this module.

nvflare/private/fed/server/fed_server.py
nvidianz: Fixed on the current head. Shared-port configurations now fail fast with ConfigError when admin_host differs from the FL bind host, and the regression test covers an external FL target with a loopback admin_host.

nvflare/fuel/hci/server/sess.py
nvidianz: Fixed on the current head. handle_check_session now obtains the identity asserter from ConnProps.HCI_SERVER, and get_session no longer has a None default. The registered command regression test verifies a signed live token returns OK.

nvflare/private/fed/server/fed_server.py
nvidianz: Addressed in 4ae4fa6e6. IPv4 loopback literals now retain the configured address, so 127.0.0.2 is not collapsed to 127.0.0.1. Abbreviated 127/8 forms are recognized without wildcard exposure. IPv6 loopback is still normalized to IPv4 because the F3 URL parser and TCP drivers do not support IPv6 end to end, and that coercion is now logged. Binding loopback-only for a loopback service target is the intended security posture; container deployments that require external reachability must use a non-loopback service target.

nvflare/private/fed/server/fed_server.py
nvidianz: Addressed in 4ae4fa6e6. Invalid host types and unsupported IPv6 literals now raise ConfigError, and IPv4-mapped IPv6 loopback is normalized consistently across supported Python versions. Hostnames remain supported intentionally as bind host configuration; the IPv6 rejection is limited to literals because F3 cannot parse them end to end. Silent transport-driver bind failures are a separate lower-layer issue.

nvflare/private/fed/server/fed_server.py
nvidianz: Addressed in 4ae4fa6e6. Configured IPv4 loopback literals are now honored exactly. IPv6 loopback remains normalized to IPv4 due to the documented F3 IPv6 limitation, but the server now emits an explicit warning describing the actual bind address.

nvflare/private/fed/server/fed_server.py
nvidianz: Addressed in 4ae4fa6e6. The insecure-listener warning is now gated on an explicitly configured admin_host, so default production deployments no longer warn. Loopback classification also recognizes abbreviated IPv4 loopback and IPv4-mapped IPv6. The shared-port remediation is protected by the fail-fast ConfigError already added on the current head.

nvflare/fuel/hci/server/login.py
nvidianz: Addressed in 4ae4fa6e6. Logout now consumes the Session placed on ConnProps.SESSION by pre_command and ends it by ID. The token-decoding helper and identity-asserter plumbing were removed, and a regression test confirms logout removes the preverified session.

nvflare/fuel/hci/server/sess.py
nvidianz: Confirmed. This is a pre-existing token revocation design gap and is not introduced by the signature-verification hardening in this PR. It should be tracked separately because fixing it requires a revocation or token-generation policy rather than a local correction here.

nvflare/private/fed/server/fed_server.py
nvidianz: Fixed in 4ae4fa6e6. Non-string admin_host values now fail with a classified ConfigError and a clear type-specific message; empty and whitespace-only values use ConfigError as well.

nvflare/private/fed/server/fed_server.py
nvidianz: Agreed that consolidation is worthwhile, but it is broader than this server-side security fix because the existing helpers intentionally differ around F3 IPv6 limitations and listener resource semantics. I kept this PR scoped and added regression coverage for the supported server cases; the cross-layer helper consolidation should be a follow-up.

tests/unit_test/fuel/hci/session_study_test.py
nvidianz: Addressed in 4ae4fa6e6. Added a real IdentityAsserter.sign to TokenVerifier.verify round-trip using a generated certificate and private key, while retaining the focused fake-verifier tests for invalid signatures.

nvflare/private/fed/server/fed_server.py
```diff
@@ -82,6 +85,58 @@
...
+    except ValueError:
+        try:
+            return ipaddress.IPv4Address(socket.inet_aton(value))
+        except OSError:
+            return None
+
+
+def _normalize_loopback_host(host: str) -> str:
+    """Return a loopback bind address for localhost and loopback IP literals."""
+    if not host:
```
YuanTingHsieh: **Nit: make the falsey-host contract explicit.** `host` is annotated as `str`; the admin path validates it before calling this helper, and the service-target path also supplies a string. `if not host` therefore suggests that `None` or another falsey non-string is supported, and silently propagating such a value could later be converted to the wildcard host by `listening_host or "0"`. If an empty string is intentionally supported as wildcard shorthand, please check `host == ""` and document that; otherwise reject it at the configuration boundary.

YuanTingHsieh (commented):
One non-blocking cleanup comment on the loopback helper contract.

nvflare/private/fed/server/fed_server.py
nvidianz: Addressed in 700b330e7. The helper now checks host == "" explicitly and documents that an empty service host is preserved as wildcard shorthand. Added a deployment regression test confirming :8002 produces tcp://0:8002 with no loopback listener override.

YuanTingHsieh (approved):

PR #5177, opened 2026-08-20
Fixes repeatable participant selection and exclusion handling for POC commands.

nvkevlu:
Fixes repeatable participant selection and exclusion handling for POC commands.

### Description

Allows -p/--service and -ex/--exclude to be repeated with nvflare poc start and nvflare poc stop, preserving every participant instead of silently keeping only the last value. It also fixes stop routing so exclusions are honored and updates the related documentation, examples, migration guide, and 2.9 release notes.

Follows up after https://github.com/NVIDIA/NVFlare/pull/5155.

### Types of changes

- [ ] Documentation updated.


nvflare/tool/poc/poc_commands.py
```diff
@@ -1994,10 +2005,15 @@ def _stop_poc(
...
     if services_list is None:
         services_list = []
-    if excluded is None:
-        excluded = [service_config[SC.FLARE_PROJ_ADMIN]]
-    else:
-        excluded.append(service_config[SC.FLARE_PROJ_ADMIN])
+    user_excluded = list(excluded or [])
+    excluded = list(user_excluded)
+    project_admin = service_config[SC.FLARE_PROJ_ADMIN]
+    if project_admin not in excluded:
```
YuanTingHsieh, line 2011: [P2] Reject or actually stop explicitly selected admin consoles

This still produces a false-success result on the current head. For example, `nvflare poc stop -p admin@nvidia.com -p site-1` always adds the project admin to `excluded`, so only `site-1` gets a stop command; after that client exits, the command returns `status: stopped` even though the explicitly selected admin console is still running. A selected secondary admin similarly gets only an unused `shutdown.fl` touch and is omitted from the PID wait. Since `-p` accepts participant names and this PR supports mixed repeated selections, please either reject admin stop targets explicitly or terminate and wait for them, with a regression test for the mixed admin/client case.

YuanTingHsieh (commented):
One remaining correctness issue in the repeated stop-selection path.

nvflare/tool/poc/poc_commands.py
```diff
@@ -2029,6 +2054,7 @@ def _stop_poc(
             project_config,
             excluded=excluded,
             services_list=services_list,
+            wait=wait,
```
YuanTingHsieh, line 2057: Is making local/subset stops blocking by default an intentional behavior change? Historically, a bare/default stop used coordinated shutdown and waited, while `nvflare poc stop -p site-1` used the local `_run_poc` path and was effectively fire-and-forget. Passing `wait=True` here changes that command to block for up to 30 seconds and potentially fail with a timeout; exclusion-based stops now inherit the same new behavior. If this is the desired contract, should we call out the subset-stop compatibility change explicitly in the migration/release notes? If not, we may want to preserve the previous local behavior and distinguish its result from `status: stopped`. What behavior do we intend?

YuanTingHsieh (commented):
Question about the intended compatibility contract for local POC shutdown.

nvflare/tool/poc/poc_commands.py
nvkevlu: Previously, subset stops returned immediately while incorrectly reporting status: stopped and wait: true. The new behavior makes the contract consistent.

This also matches behavior for bare nvflare poc stop with no -p or -ex

The release and migration notes have been updated.

YuanTingHsieh (approved):
Approved. The shutdown behavior is now internally consistent: normal stops wait before reporting `status: stopped`, while `--no-wait` is the explicit fire-and-forget path. The repeated participant/exclusion handling, admin-console validation, timeout handling, tests, and user-facing documentation look good.

Non-blocking documentation suggestion: the 2.9 release notes now mention that targeted and exclusion-based stops wait, but the migration-guide POC section still only describes repeatable options and honored exclusions. Consider adding the same wait/`--no-wait` compatibility note there.

PR #5179, opened 2026-08-20
Fix external trainer settlement and receiver parsing

YuanTingHsieh:
## Summary

- give `RESULT_SOURCE_SETTLED` a 30-second per-attempt control budget, matching `RESULT_READY`, while retaining three retries and warning when all attempts are unacknowledged
- make END_RUN re-sample the trainer's transfer barrier and perform a final bounded SHUTDOWN probe, so a completed download is not reported lost merely because the separate settlement acknowledgement is delayed
- validate and deduplicate trainer-side `RECEIVER_IDS` while preserving count-based tracking (`receiver_ids=None`) when no valid identity is supplied
- document all five current roles of `shutdown_timeout`, its zero-value behavior for ordinary and accepted-source teardown, and the accepted-result reaper's 5-second `stop_grace_period` clamp

## Why

A client job can legitimately remain busy for more than one second while streaming the next task. The previous settlement timeout could therefore miss all acknowledgements and leave a clean trainer exit classified as an unsettled source loss after every result download had completed.

During END_RUN, equal 30-second trainer and reaper deadlines alone are insufficient because the reaper interval can start before downstream transfer completion. The trainer now publishes transfer-barrier completion before waiting for the task-correlated settlement reply. The backend keeps probing while that barrier is live and makes a final probe at its deadline, allowing the completed source to become settled even if `RESULT_SOURCE_SETTLED` is still queued.

Separately, invalid or duplicate `RECEIVER_IDS` entries were interpreted differently by the trainer and client job. The aligned parser keeps valid ordered identities, removes duplicates, and retains the prior identity-free count fallback when none are usable, avoiding an invented receiver identity that could wait for the full streaming timeout.

## Testing

- `python -m pytest -q tests/unit_test/client/cell/api_test.py tests/unit_test/app_common/executors/external_process_backend_test.py` — 231 passed, 2 skipped on the exact signed commit rebased on current `main`
- Black, isort, and flake8 passed for every touched Python file
- `git diff --check` passed

### DIFF
--- a/docs/design/client_api_execution_modes.md
+++ b/docs/design/client_api_execution_modes.md
@@ -303,11 +303,13 @@ trainer to stop and starts the natural-exit reaper. This also covers inline resu
 reply can race END_RUN even though they create no download transaction. Teardown cannot safely
 return with a daemon reaper because ClientRunner tears down streaming and the CJ Cell immediately
 after END_RUN. A result transfer retains its own `DownloadService` idle/receiver policy, but that
-longer data-transfer budget does not govern CJ process ownership. END_RUN gives a still-live result
-source one acknowledged SHUTDOWN interval, then force-stops every remaining owned process group.
-A source already known to be settled instead receives its configured natural-exit budget and a
-small configured TERM grace, both inside a 30-second cleanup cap, so normal post-send checkpoint
-work is not cut off at the live-source acknowledgement deadline. A run-abort-marked `ABORT_TASK`
+longer data-transfer budget does not govern CJ process ownership. END_RUN uses a short acknowledged
+SHUTDOWN request to sample send-barrier state throughout a 30-second session-scale interval. The
+trainer publishes transfer-barrier completion before waiting for `RESULT_SOURCE_SETTLED`, and the
+reaper makes a final SHUTDOWN probe at the deadline. Therefore a completed transfer is treated as
+settled even if its separate settlement acknowledgement is delayed. A source already known to be
+settled instead receives its configured natural-exit budget and a small configured TERM grace inside
+the cleanup budget. A run-abort-marked `ABORT_TASK`
 latches aborted teardown even when `execute()` already returned the lazy result, sends ABORT
 immediately, and bypasses the normal result-source drain interval. Task/workflow-only cancellation
 sends ABORT without poisoning later tasks, and normal END_RUN carries no run-abort marker. Other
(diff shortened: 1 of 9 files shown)


chesterxgchen (approved):
Verdict: solid, well-tested fix. A few minor nits, nothing blocking.

This PR bundles three related but distinct fixes for the external_process Client API backend:

1. Settlement timeout too aggressive (_RESULT_SOURCE_SETTLED_TIMEOUT 1s→30s)

Correct fix for the stated bug: a busy CJ (streaming the next task) could legitimately take >1s to ack RESULT_SOURCE_SETTLED, causing a completed, accepted result to be misreported as an unsettled/lost source. Bumping to a shared SESSION_CONTROL_TIMEOUT = 30.0 constant is reasonable and matches the backend's own reaper budget.

- Minor concern: _RESULT_SOURCE_SETTLED_ATTEMPTS stayed at 3, so worst-case blocking in _notify_result_source_settled went from 3s to ~90s if the CJ is truly unreachable. In practice this is harmless because the backend's own reaper forcibly kills the trainer process at its independent 30s deadline (_RESULT_REAPER_MAX_TOTAL_TIMEOUT), so the trainer won't actually sit blocked for 90s — but the two budgets (3×30s vs 30s) aren't obviously coupled in the code, only in intent. Worth a comment or a reduced attempt count, not a blocker.

2. END_RUN transfer-barrier resampling (_request_trainer_shutdown/_wait_for_result_reapers)

This is the core race fix. The trainer now flips _result_send_active = False (the send barrier) before attempting the separately-acked RESULT_SOURCE_SETTLED notification (nvflare/client/cell/api.py:576-583), so a concurrent SHUTDOWN probe from the backend can observe RESULT_SOURCE_LIVE: False even if the settlement RPC is still delayed/lost. The backend's reaper loop now gives a live source the full 30s session-scale budget instead of just 5s, and performs one final forced probe (force_probe=True, bypassing the retry throttle) right at the deadline before declaring the source lost.

- I traced the locking (_shutdown_request_lock, early-return-without-setting-shutdown_requested when still live) and it's consistent — no deadlock or busy-loop risk since force_probe is only invoked once, at the deadline.
- Doc nit: the docstring/design-doc language ("samples send-barrier state throughout a 30-second interval") overstates what the code does — it's actually just two probes (one at t=0 in finalize(), one at the deadline in _wait_for_result_reapers()), not continuous sampling. Not a functional issue, just slightly misleading prose.

3. RECEIVER_IDS validation/dedup

_normalize_result_receiver_ids now uses FQCN.validate() (confirmed: returns ""/falsy for valid, non-empty error string for invalid) plus dict.fromkeys for order-preserving dedup, falling back to None (count-based tracking) when nothing valid survives. Verified against the actual FQCN.validate implementation — logic is correct.
- Minor behavior narrowing: it now only accepts str/list/tuple inputs (previously any iterable via try: list(x)). Since this field comes from an internal Cell header (FOBSContextKey.RECEIVER_IDS), this is unlikely to matter in practice.

chesterxgchen:
@YuanTingHsieh approved with minor comments.

PR #5181, opened 2026-08-21
Align agent skills with current specification

chesterxgchen:
## Summary

Align the bundled NVFLARE skills with the current Agent Skills frontmatter contract, improve instruction quality, document the Hugging Face helper script, and remove duplicated local-validation guidance.

### Agent Skills frontmatter

- Move `version: "0.1.0"` from the root of every bundled `SKILL.md` into the `metadata` mapping.
- Require `metadata.version` in the NVFLARE frontmatter validator.
- Remove root-level `version` from the validator allowlist so misplaced versions are reported instead of silently accepted.
- Update the skills README, valid fixture, and frontmatter/lint tests for the new structure.
- Add regression coverage confirming all nine bundled skills use `metadata.version` and reject root-level `version`.

### Instruction quality

- Rewrite the `nvflare-shared` purpose with direct imperative verbs: `Use`, `Resolve`, and `Do not`.
- Retain its existing `## Instructions` and `## Examples` sections, whose actions use imperative verbs such as `Load`, `Treat`, and `Apply`.
- Add an `## Available Scripts` table to `nvflare-convert-huggingface` for its bundled `scripts/resolve_model_snapshot.py` helper.
- Add explicit `run_script()` examples for local-path and Hugging Face Hub snapshot resolution.
- Keep the Hugging Face skill within the repository hard limit of 200 lines by reflowing its unchanged reference-loading guidance.

### Shared-reference deduplication

- Make `references/validation-evidence.md` the single owner of the complete local-validation path, including:
  - `SimEnv` construction and execution
  - local versus exported-job command selection
  - simulation environment constraints
  - destructive-cleanup safeguards
  - terminal evidence and post-simulation reporting
- Replace the duplicated validation block in `references/conversion-workflow.md` with a one-line pointer to the canonical reference.
- Add regression checks preventing export commands and topology guidance from being duplicated back into the workflow reference.

## Validation

- Hugging Face SkillEvaluator quality score: 91.5 (grade A), increased from 84.5 and above the 85.0 Tier 1 floor.
- Official Agent Skills reference validator passed for all nine bundled skills.
- `python -m pytest tests/unit_test/tool/agent_skill_checks -q` — 478 passed, 37 skipped.
- `python -m dev_tools.agent.skills.checks` — 0 errors, warnings, or informational findings.
- `./runtest.sh -s` — Black, isort, Flake8, and agent-skill lint passed.
- `git diff --check` passed.

## Validation limitations

- The complete local Tier 1 command reports the corrected 91.5 quality score, but its Semgrep check cannot complete on the macOS development host because the local Semgrep installation has an empty X.509 trust store. The corresponding Linux CI prerequisite is installed by the repository workflow.
- The local Tier 2 dedup scan requires an embedding provider, which is not configured in the development environment. Repository regression tests cover the intended single-owner structure.
- Dedicated NVSkills CI is not requested because it does not support fork-based PRs.

### DIFF
--- a/dev_tools/agent/skills/checks/frontmatter.py
+++ b/dev_tools/agent/skills/checks/frontmatter.py
@@ -32,7 +32,7 @@ YAML_ANCHOR_OR_ALIAS_RE = re.compile(r"(^|[:\s\[{,])([&*])[A-Za-z0-9_-]+(?=\s|$|
 REQUIRED_FRONTMATTER_FIELDS = ("name", "description")
 # `author` (name plus NVIDIA email inside the spec `metadata` map) is required
 # by the company skills guideline (NVCARPS).
-REQUIRED_METADATA_FIELDS = ("author", "min-flare-version", "blast-radius")
+REQUIRED_METADATA_FIELDS = ("author", "version", "min-flare-version", "blast-radius")
 # The NVIDIA skills catalog uses a team identity, never an individual: once
 # `npx skills add` copies the frontmatter out of the repo, `author` acts as the
 # support contact, and a personal inbox goes stale when people change teams.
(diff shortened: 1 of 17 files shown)


holgerroth (approved):
Approving to keep this moving — the frontmatter migration, validator change, and dedup all check out (validator type-checks `metadata.version`, all nine skills conform, deleted validation content is re-established in `validation-evidence.md`). A few non-blocking items worth follow-up:

1. **`nvflare-convert-huggingface/SKILL.md` (Available Scripts):** the examples call `run_script(...)`, a helper that doesn't exist in agent runtimes — `nvflare-autofl-report/SKILL.md` explicitly says not to invent one, and `references/huggingface-validation.md` declares the `python <skill-dir>/scripts/...` form canonical. Suggest switching the examples to the canonical form.
2. **Same file, reference-loading paragraph:** the reflow replaced the inline trigger clause for `site-data-and-paths.md` with "only for its stated triggers", which can be read as scoping all three references in that clause (including the always-required `conversion-common.md`). The pytorch/lightning converters kept the inline wording; restoring it here would match them.
3. **`dev_tools/agent/skills/checks/lints.py:2554`:** `_line_for_frontmatter_issue`'s field tuple lacks `version`, so the new top-level-`version` rejection is attributed to the `name:` line and a non-string `metadata.version` falls back to line 1 (reproduced locally). Extracting the quoted field name like the field-required branch does would fix it generally.
4. **`seed_skills_test.py:715`:** the negative export-command assertion runs on raw `workflow_text` while the positive one whitespace-normalizes, so a line-wrapped reintroduction would evade the guard — normalize both sides.
5. **Publish pipeline:** the NVIDIA skills catalog still declares `version` at the top level in its published skills, and no visible catalog schema reads either form. Worth confirming the downstream skill-validation run passes on these nine skills after merge (it must be triggered from a branch on this repo, not a fork).

PR #5183, opened 2026-08-21
Make blocking Collab group results re-iterable

nvidianz:
## Problem

Blocking Collab group calls wait for every site outcome but return the same live, destructive result iterator used by non-blocking calls. A second pass over those completed results silently yields no entries, which can cause multi-stage aggregation code to compute an empty result without any error.

## Changes

- freeze completed blocking-call results into an ordered, replayable snapshot
- preserve the existing `(site, result)` iteration contract, partial-result ordering, and `.failures`
- keep non-blocking result streams live and single-pass
- document the blocking and non-blocking contracts
- add regression coverage for repeated list/dict conversion, partial results, failures, and invalid freezing

## Validation

- `python -m pytest -q tests/unit_test/collab` — 120 passed
- focused regression tests — 14 passed
- scoped Black, isort, and Flake8 checks passed

The full style wrapper could not install into the externally managed Homebrew Python environment, so the equivalent scoped checks were run directly.

### DIFF
--- a/nvflare/collab/api/group_call_context.py
+++ b/nvflare/collab/api/group_call_context.py
@@ -47,6 +47,9 @@ class ResultQueue:
         self.num_successes = 0
         self.failures = {}
         self.update_lock = threading.Lock()
+        self._iteration_started = False
+        self._frozen_items = None
+        self._frozen_next_index = 0
 
     def append(self, item, is_whole=True):
         """Append an item to the result queue.
@@ -85,12 +88,22 @@ class ResultQueue:
             return self.num_whole_items_received == self.limit
 
     def __iter__(self):
+        with self.update_lock:
+            if self._frozen_items is not None:
+                return iter(self._frozen_items)
         return self
 
     def __next__(self):
         while True:
             # Check the queue and completion count atomically with append().
             with self.update_lock:
+                if self._frozen_items is not None:
+                    if self._frozen_next_index >= len(self._frozen_items):
+                        raise StopIteration()
+                    item = self._frozen_items[self._frozen_next_index]
+                    self._frozen_next_index += 1
+                    return item
+                self._iteration_started = True
                 try:
                     item = self.q.get_nowait()
                 except queue.Empty:
@@ -109,6 +122,25 @@ class ResultQueue:
             item = self.q.get(block=True)
             if item is not _FAILURE_MARKER:
                 return item
+
+    def freeze(self):
+        """Freeze a completed, unconsumed queue into a re-iterable snapshot."""
+        with self.update_lock:
+            if self.num_whole_items_received < self.limit:
+                raise RuntimeError("cannot freeze result queue before all outcomes are received")
+            if self._iteration_started:
+                raise RuntimeError("cannot freeze result queue after iteration has started")
+            if self._frozen_items is None:
+                items = []
+                while True:
+                    try:
+                        item = self.q.get_nowait()
+                    except queue.Empty:
+                        break
+                    if item is not _FAILURE_MARKER:
+                        items.append(item)
+                self._frozen_items = tuple(items)
+            return self
 
     def __len__(self):
         """Return the number of successful whole results received.
--- a/nvflare/collab/api/group_call_context.py
+++ b/nvflare/collab/api/group_call_context.py
@@ -33,6 +33,21 @@ def _get_site_name(target_name: str):
     return target_name.split(".", 1)[0]
 
 
+class FrozenResultQueue:
+    """Immutable, re-iterable snapshot of completed group-call results."""
+
+    def __init__(self, items, failures, num_successes):
+        self._items = tuple(items)
+        self.failures = dict(failures)
+        self._num_successes = num_successes
+
+    def __iter__(self):
+        return iter(self._items)
+
+    def __len__(self):
+        return self._num_successes
+
+
 class ResultQueue:
     def __init__(self, limit: int, retain_history=True):
         if limit <= 0:
@@ -48,7 +63,7 @@ class ResultQueue:
         self.failures = {}
         self.update_lock = threading.Lock()
         self._items = [] if retain_history else None
-        self._frozen_items = None
+        self._frozen_results = None
 
     def append(self, item, is_whole=True):
         """Append an item to the result queue.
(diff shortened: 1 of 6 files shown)


nvflare/collab/api/group_call_context.py
```diff
@@ -110,6 +123,25 @@ def __next__(self):
             if item is not _FAILURE_MARKER:
                 return item
 
+    def freeze(self):
+        """Freeze a completed, unconsumed queue into a re-iterable snapshot."""
+        with self.update_lock:
+            if self.num_whole_items_received < self.limit:
+                raise RuntimeError("cannot freeze result queue before all outcomes are received")
+            if self._iteration_started:
+                raise RuntimeError("cannot freeze result queue after iteration has started")
```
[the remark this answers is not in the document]
nvidianz: Fixed in 42c98d58e. Blocking calls now retain an append-only result history, so freeze() snapshots every successful result even when process_resp_cb consumed items from the live queue. Added a group-level regression test covering callback consumption.

nvflare/collab/api/group_call_context.py
```diff
@@ -85,12 +88,22 @@ def append_failure(self, site_name: str, error: CollabCallError):
             return self.num_whole_items_received == self.limit
 
     def __iter__(self):
+        with self.update_lock:
+            if self._frozen_items is not None:
+                return iter(self._frozen_items)
         return self
```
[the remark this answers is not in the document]
nvidianz: Fixed in 42c98d58e. Removed the competing frozen cursor: frozen results are now collection-style and create a fresh iterator through iter(), while direct next(results) raises a clear TypeError. Regression coverage verifies both behaviors.

nvflare/collab/api/group_call_context.py
```diff
@@ -85,12 +89,17 @@ def append_failure(self, site_name: str, error: CollabCallError):
...
 
     def __next__(self):
         while True:
             # Check the queue and completion count atomically with append().
             with self.update_lock:
+                if self._frozen_items is not None:
+                    raise TypeError("frozen results are re-iterable; use iter(results) before calling next()")
```
holgerroth: Two related problems with the frozen-mode `TypeError`:

**1. Race — `TypeError` where `StopIteration` is due.** `__iter__` returns `self` while the queue is live, so an iterator obtained before `freeze()` — e.g. a `process_resp_cb` (or a thread it spawns) iterating `gcc.waiter.results` mid-call, the same access `test_blocking_group_results_include_results_consumed_by_callback` exercises — hits this raise on its next step once the blocking call returns and `group.py` freezes the queue. Before this PR that step was a clean `StopIteration`. A consumer parked in `q.get(block=True)` at freeze time still receives the final item, then crashes on the following advance (reproduced with a side-thread `list(results)` racing `append_failure()` + `freeze()`).

**2. Iterator-protocol oddity.** The class still defines `__next__`, so `isinstance(results, collections.abc.Iterator)` stays `True`, yet `next(results)` raises and `next(results, default)` crashes as well (the default only suppresses `StopIteration`). Previously, `next(results)` on a blocking call returned the first result.

Both go away if `freeze()` returns a small frozen-view object (snapshot tuple + `.failures` + `__len__`, no `__next__`) instead of switching the shared queue into a third mode: stale live handles then keep draining the queue and end in `StopIteration` naturally, and `next(frozen)` fails with Python's standard "not an iterator" `TypeError`. A minimal alternative that keeps the current shape: when frozen, have `__next__` serve remaining queued items and then raise `StopIteration` rather than `TypeError`. Worth settling now while Collab is unreleased — once it ships, changing the blocking return type becomes a breaking change.

tests/unit_test/collab/group_test.py
```diff
@@ -96,3 +96,52 @@ def invoke():
...
+    assert dict(results) == expected_by_site
+
+
+def test_blocking_group_results_include_results_consumed_by_callback():
+    app = ClientApp(object())
+    consumed = []
+
+    def process_response(gcc, result):
+        if gcc.target_name == "site-2":
+            consumed.append(next(gcc.waiter.results))
```
holgerroth: This is the first in-repo endorsement of calling `next(gcc.waiter.results)` inside a `process_resp_cb`, and that pattern can deadlock a blocking group call against the real backend: `GroupCallContext.set_result` invokes the callback *before* the site's own outcome is appended (the append happens in the `finally`), so a callback's `next()` can only be satisfied by *other* sites' outcomes. The test is safe only because the mocked dispatch here is synchronous and ordered; the production `CellDispatcher` submits calls via a `ThreadPoolExecutor`, so callbacks can run before any outcome has been queued.

Concrete hangs (reproduced on a 1-site group): the callback blocks forever in `q.get(block=True)` because its own site's append is stuck behind the callback's return. The same happens when the consumed prior outcome was a failure — `append_failure` enqueues only the internal marker, which `__next__` skips before blocking.

The deadlock mechanism predates this PR, but before this test canonizes the pattern, suggest either exercising the history-retention behavior without modeling `next()`-in-callback as supported usage, or adding an explicit hazard note here and tracking the callback/append ordering issue separately.

holgerroth (commented):
Ran a deep review of the re-iterability change. The core approach checks out: `freeze()`'s error paths are unreachable in the production blocking flow (per-site timeouts arrive as counted failures, and aborts raise before the freeze is reached), and the regression coverage is thorough. Two issues are worth addressing before merge — inline below. Smaller items (draining `q` in `freeze()`, an explicit `retain_history=False` at `_invocation.py:57`, a couple of docstring gaps) can ride in a follow-up.

nvflare/collab/api/group_call_context.py
nvidianz: Addressed in a9ce9d336. `freeze()` now returns a separate `FrozenResultQueue` snapshot, leaving the live `ResultQueue` unchanged so pre-existing iterators drain normally and finish with `StopIteration`. The snapshot has no `__next__`, and regression coverage verifies both behaviors.

tests/unit_test/collab/group_test.py
nvidianz: Addressed in a9ce9d336. Removed the callback-consumption test so it no longer presents `next(gcc.waiter.results)` from `process_resp_cb` as supported usage. History retention remains covered independently, and failure markers now reliably lead live iterators to `StopIteration`.

holgerroth (approved):
Verified the fixes in a9ce9d336: `freeze()` now returns a detached `FrozenResultQueue`, the live `ResultQueue` keeps single-mode iterator semantics (pre-freeze iterators drain normally and end in `StopIteration` — regression-tested), `next(frozen)` fails with Python's standard non-iterator `TypeError`, and the callback-consumption pattern is no longer presented as supported usage. Collab unit suite passes locally (123/123) and CI is green. Approving.

Non-blocking follow-ups noted during review, fine to land separately: an explicit `retain_history=False` at `_invocation.py:57`; a docstring note that the frozen snapshot includes results consumed during the call; the empty `Returns:` section in `ProxyList.__call__`; and the pre-existing interaction where `dict(results)` can present a failed site's last streamed partial unless `.failures` is cross-checked.

PR #5185, opened 2026-08-21
Allow client-to-client Collab calls through transit cells

nvidianz:
## Summary

- authorize Collab stream frames only at their final job-cell destination
- allow intermediate CellNet nodes to forward client-to-client Collab traffic
- preserve pre-allocation participant and envelope authorization at the receiving job cell
- add focused coverage for transit streams

## Background

CoreCell invokes incoming filters before checking whether a message is addressed locally or needs forwarding. The Collab stream filter therefore rejected generation-1 client-to-client traffic at the intermediate server because the destination did not match the server cell. The sender then received an uncorrelated stream error and the workflow stalled.

Transit cells now leave the stream untouched. The final destination still runs the existing exact destination, participant, protocol, target, and method checks before ByteReceiver allocates receive state.

## Validation

- 29 focused Collab authorization and security tests passed
- 117 Collab unit tests passed
- shipped two-client swarm example completed successfully in SimEnv with client-to-client traffic routed through the server
- scoped Black, isort, Flake8, and whitespace checks passed
- repository style wrapper was run with Python 3.12.13 from ~/venv/3.12; it reached Black and reported only three generated node_modules/flatted.py files outside this change that would be reformatted

### DIFF
--- a/nvflare/collab/runtime/dispatch.py
+++ b/nvflare/collab/runtime/dispatch.py
@@ -100,17 +100,10 @@ class _CollabStreamFilter:
         ):
             return None
 
-        # CoreCell invokes incoming filters before it decides whether to
-        # deliver or forward a message. Only authorize streams addressed to
-        # this job cell; a transit stream will be authorized by its final
-        # destination before receiver state is allocated there.
-        if message.get_header(MessageHeaderKey.DESTINATION) != self.authorizer.local_fqcn:
-            return None
-
         rejection = self.authorizer.authorize(message.headers)
         if rejection:
             self.byte_receiver.reject(message, rejection)
-            # Incoming filters stop delivery when they return a Message. The
+            # Incoming request filters stop delivery when they return a Message. The
             # stream rejection itself is reported on ByteStreamer's generic
             # error topics, so this outer fire-and-forget reply is discarded.
             return Message()
@@ -144,7 +137,9 @@ def prepare_for_remote_call(cell, app, logger, executor, participants: dict[str,
     authorizer = CollabCallAuthorizer(app, cell.get_fqcn(), participants, logger)
     adapter = Adapter(_submit_app_method, cell.core_cell.my_info, cell)
     stream_filter = _CollabStreamFilter(authorizer, cell.byte_receiver)
-    cell.core_cell.add_incoming_filter(STREAM_CHANNEL, STREAM_DATA_TOPIC, stream_filter.filter)
+    # Request filters run only after CoreCell has routed the frame to this
+    # destination, but before ByteReceiver allocates stream state.
+    cell.core_cell.add_incoming_request_filter(STREAM_CHANNEL, STREAM_DATA_TOPIC, stream_filter.filter)
     cell.register_blob_cb(
         channel=MSG_CHANNEL,
         topic=MSG_TOPIC,
--- a/nvflare/collab/runtime/dispatch.py
+++ b/nvflare/collab/runtime/dispatch.py
@@ -100,6 +100,13 @@ class _CollabStreamFilter:
         ):
             return None
 
+        # CoreCell invokes incoming filters before it decides whether to
+        # deliver or forward a message. Only authorize streams addressed to
+        # this job cell; a transit stream will be authorized by its final
+        # destination before receiver state is allocated there.
+        if message.get_header(MessageHeaderKey.DESTINATION) != self.authorizer.local_fqcn:
+            return None
+
         rejection = self.authorizer.authorize(message.headers)
         if rejection:
             self.byte_receiver.reject(message, rejection)
--- a/tests/unit_test/collab/runtime/dispatch_test.py
+++ b/tests/unit_test/collab/runtime/dispatch_test.py
@@ -109,8 +109,8 @@ def test_prepare_for_remote_call_registers_blob_callback():
     assert registration["logger"] is logger
     assert registration["executor"] is executor
     assert "preflight_cb" not in registration
-    cell.core_cell.add_incoming_filter.assert_called_once()
-    filter_registration = cell.core_cell.add_incoming_filter.call_args.args
+    cell.core_cell.add_incoming_request_filter.assert_called_once()
+    filter_registration = cell.core_cell.add_incoming_request_filter.call_args.args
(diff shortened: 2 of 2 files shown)


nvflare/collab/runtime/dispatch.py
```diff
@@ -100,6 +100,13 @@ def filter(self, message: Message):
         ):
             return None
 
+        # CoreCell invokes incoming filters before it decides whether to
+        # deliver or forward a message. Only authorize streams addressed to
+        # this job cell; a transit stream will be authorized by its final
+        # destination before receiver state is allocated there.
+        if message.get_header(MessageHeaderKey.DESTINATION) != self.authorizer.local_fqcn:
```
holgerroth: Consider registering this filter through CoreCell's post-routing stage instead of re-deriving the routing decision here.

Because the filter is registered via `add_incoming_filter`, CoreCell invokes it (core_cell.py:2084) *before* its deliver-vs-forward decision (core_cell.py:2123) — which is why this hand re-checks `DESTINATION != local_fqcn`. That works, but it becomes a third copy of the "is this addressed to me" predicate alongside core_cell.py:2123 and `authorize()` (line 56), and all three have to stay in sync for the "authorize iff delivered locally" property to hold; a later routing change (FQCN normalization, aliasing, broadcast destinations) would diverge them silently.

`cell.core_cell.add_incoming_request_filter(STREAM_CHANNEL, STREAM_DATA_TOPIC, ...)` expresses the intended semantics natively: request filters run inside `_process_request` (core_cell.py:1820), reached only after the `destination == my_info.fqcn` decision, yet still before `ByteReceiver._data_handler` (the request cb registered at byte_receiver.py:730, invoked at core_cell.py:1833) allocates the RxTask. Stream frames are fire-and-forget REQ with a BYTES payload, so the extra decode/decrypt/reply-discard steps are no-ops, and it additionally covers the bulk-delivery path (core_cell.py:1722) that the general in-filter stage skips. That would let you delete this guard entirely.

If you'd rather keep the manual guard in this PR, two smaller notes:

1. The comment above (lines 103-106) slightly overstates the guarantee. For a destination that isn't a prepared Collab job cell (a non-Collab cell, or a peer that hasn't run `prepare_for_remote_call` yet), `_data_handler` allocates the RxTask (byte_receiver.py:769) *before* discovering there's no callback and tearing it down (:775-777) — so "before receiver state is allocated there" only holds for a prepared Collab cell. Something like "before the Collab call is dispatched" would be accurate.

2. This check is fail-open on an absent DESTINATION (`None != local_fqcn` → skip). It's unreachable today because CoreCell raises on a missing DESTINATION before filters run (core_cell.py:2071-2075), but `if not dest or dest != self.authorizer.local_fqcn` would make the filter's own contract fail-closed regardless of caller.

holgerroth (commented):
Reviewed the transit-authorization fix — the core change looks correct. The filter's skip predicate (`DESTINATION != authorizer.local_fqcn`) reads the same operands CoreCell uses for its own deliver-vs-forward decision (`destination != my_info.fqcn`, both the once-normalized `my_info.fqcn`), so a frame skips authorization exactly when CoreCell forwards it instead of delivering locally, and `authorize()` still overwrites the sender-supplied `AUTHENTICATED_CALLER` at the true destination. I couldn't find an authorization-bypass path.

Two suggestions inline — one production, one test. Neither blocks correctness.

One optional cleanup not worth its own thread: once the guard on line 107 is in place, `CollabCallAuthorizer.authorize`'s own `destination != self.local_fqcn` check (lines 55-57) can no longer fire from its sole caller. It's harmless defense-in-depth; I'd just leave a one-line note that the caller pre-filters transit so that branch reads as an intentional backstop rather than live logic (or let the request-filter refactor below make the relationship obvious).

tests/unit_test/collab/runtime/dispatch_test.py
```diff
@@ -137,6 +138,25 @@ def test_collab_stream_filter_rejects_before_receive_state_is_created():
...
+        }
+    )
+
+    assert stream_filter.filter(message) is None
+    authorizer.authorize.assert_not_called()
+    byte_receiver.reject.assert_not_called()
+
+
 def test_collab_stream_filter_ignores_other_streams():
     authorizer = MagicMock()
```
holgerroth: After the new destination guard, this test no longer isolates the channel/topic check it's named for. With a bare `MagicMock()` (so `local_fqcn` is an auto-created mock) and no DESTINATION header, control reaches the new guard at dispatch.py:107 where `None != <mock>.local_fqcn` is True, so `filter()` returns None whether or not the channel/topic branch runs. I confirmed by mutation — deleting the channel/topic early-return (dispatch.py:97-101) leaves this test green.

Setting `local_fqcn` on the mock and giving the message a matching DESTINATION would keep the early return attributable to the channel/topic branch only, so the test can still fail if that guard regresses.

nvflare/collab/runtime/dispatch.py
nvidianz: Addressed in b225f383a. The Collab admission filter is now registered with add_incoming_request_filter, so CoreCell performs routing first and only the final destination invokes authorization. The exact destination check remains as a fail-closed backstop, and authorization still runs before ByteReceiver allocates stream state. Focused security tests, the full Collab suite, and the real server-routed swarm PoC pass.

tests/unit_test/collab/runtime/dispatch_test.py
nvidianz: Addressed in b225f383a. The channel/topic test now gives the mock an explicit local FQCN and supplies the matching destination header, so its early return is attributable only to the non-Collab channel and topic.

holgerroth (approved):
The follow-up commit resolves both review comments the right way: registration moved to `add_incoming_request_filter` (so the filter runs only after CoreCell routes the frame here, before ByteReceiver allocates stream state, with the hand-rolled destination guard deleted), and the other-streams test now isolates the channel/topic guard — verified by mutation that removing the guard fails the test. Re-verified the refactor's semantics: rejection replies are discarded for fire-and-forget stream frames, decode is a passthrough for byte frames, and the bulk-delivery path is now covered as well. All 116 collab unit tests pass at this head, including the raw-socket e2e authorization tests. LGTM.

PR #5190, opened 2026-08-21
Avoid per-stream error context scans

nvidianz:
## Summary

- replace full-map expiration scans with an insertion-ordered expiry queue
- keep retained stream-error contexts bounded while reducing contention on the shared stream map lock
- add regression coverage for expiry cleanup without scanning all retained contexts

## Validation

- `/Users/zhihongz/venv/3.12/bin/python -m pytest tests/unit_test/fuel/f3/streaming/byte_streamer_ack_watchdog_test.py tests/unit_test/fuel/f3/cellnet/cell_blob_size_test.py -v`
- Scoped Black, isort, flake8, and `git diff --check`

### DIFF
--- a/nvflare/fuel/f3/streaming/byte_streamer.py
+++ b/nvflare/fuel/f3/streaming/byte_streamer.py
@@ -14,6 +14,7 @@
 import logging
 import threading
 import time
+from collections import OrderedDict
 from concurrent.futures import TimeoutError, as_completed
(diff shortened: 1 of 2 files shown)


chesterxgchen:
why do we need this ?

nvidianz:
> why do we need this ?

The new ticket from QA

chesterxgchen (approved):

PR #5191, opened 2026-08-21
Handle unexpected job admission failures

YuanTingHsieh:
## Summary

- cancel successful resource reservations when job admission exits unexpectedly
- record unexpected admission failures in schedule history so retry backoff and maximum-attempt handling remain effective
- continue scanning later candidates after cleanup succeeds
- stop the scheduler pass when resource results are unavailable or cancellation fails, because reservation state cannot be proven safe
- replace the scheduler's bare exception handlers with `except Exception`

## Root cause

`DefaultJobScheduler.schedule_job()` caught exceptions around the complete candidate scan. If `_try_job()` raised after resource reservation, the exception bypassed normal cancellation and scheduling bookkeeping, then aborted the scan before later candidates were considered.

## Implementation

`_try_job()` now owns the reservation lifecycle after resource-check results are returned. Reservations are retained only when the job is admitted successfully; all other exits run cancellation from a `finally` block. Empty results for non-empty client resource requests are treated as unsafe because reservation tokens may have been lost with missing replies.

`_do_schedule_job()` handles unexpected exceptions at the individual-candidate boundary, logs the traceback, records a retryable failed attempt, and proceeds to the next candidate after successful cleanup. Errors that leave reservation state uncertain still abort the pass after bookkeeping.

## Validation

- `30 passed`: `tests/unit_test/app_common/job_schedulers/job_scheduler_test.py`
- Black check across `nvflare examples tests dev_tools`
- isort check across `nvflare examples tests dev_tools`
- Flake8 across `nvflare examples tests dev_tools`: 0 errors
- agent skill checks: 0 findings


nvflare/app_common/job_schedulers/job_scheduler.py
```diff
@@ -195,69 +200,82 @@ def _try_job(self, job: Job, fl_ctx: FLContext) -> (int, Optional[Dict[str, Disp
...
+            ) from e
+
+        keep_resources = False
+        try:
+            fl_ctx.set_prop(FLContextKey.RESOURCE_CHECK_RESULT, resource_check_results, private=True, sticky=False)
+            self.fire_event(EventType.AFTER_CHECK_CLIENT_RESOURCES, fl_ctx)
+
+            if not resource_check_results:
+                self.log_debug(fl_ctx, f"Job {job.job_id} can't be scheduled: resource check results is None or empty.")
+                return SCHEDULE_RESULT_NO_RESOURCE, None, "error checking resources"
```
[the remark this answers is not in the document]
YuanTingHsieh: Addressed in `44eb8f2ea`. When `resource_reqs` is non-empty, an empty result now raises `_UnsafeAdmissionError`, so the failed attempt is recorded and the scheduler pass stops. Empty results remain a normal no-resource result only when no client resource checks were expected. I also added coverage for empty results, resource-check exceptions, and cancellation failures; all 29 focused scheduler tests pass.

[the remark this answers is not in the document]
YuanTingHsieh: Added dedicated regression coverage in `859c95e5d`: one test verifies that empty results with expected client replies stop the candidate scan, and a contrasting test verifies that empty results remain safe when no client resource checks were issued. The focused scheduler suite now passes all 30 tests.

nvflare/app_common/job_schedulers/job_scheduler.py
```diff
@@ -195,69 +200,84 @@ def _try_job(self, job: Job, fl_ctx: FLContext) -> (int, Optional[Dict[str, Disp
...
+            return SCHEDULE_RESULT_OK, sites_dispatch_info, ""
+        finally:
+            if resource_check_results and not keep_resources:
+                try:
+                    self._cancel_resources(
+                        resource_reqs=resource_reqs,
+                        resource_check_results=resource_check_results,
+                        fl_ctx=fl_ctx,
+                    )
+                except Exception as e:
```
IsaacYangSLA, line 313: **P1 — Verify cancellation acknowledgements before continuing.** This treats cleanup as successful unless `_cancel_resources()` raises, but the production `ServerEngine.cancel_client_resources()` discards `_send_admin_requests()` replies. A timeout or a client-side `EXECUTION_EXCEPTION` therefore returns normally, and the scheduler continues even though reservation state is unknown. Please propagate and validate cancellation results, raising `_UnsafeAdmissionError` for missing or non-OK acknowledgements.

nvflare/app_common/job_schedulers/job_scheduler.py
```diff
@@ -195,69 +200,84 @@ def _try_job(self, job: Job, fl_ctx: FLContext) -> (int, Optional[Dict[str, Disp
...
+            raise _UnsafeAdmissionError(
+                f"resource check for job {job.job_id} failed before reservation results were available"
+            ) from e
+
+        keep_resources = False
+        try:
+            fl_ctx.set_prop(FLContextKey.RESOURCE_CHECK_RESULT, resource_check_results, private=True, sticky=False)
+            self.fire_event(EventType.AFTER_CHECK_CLIENT_RESOURCES, fl_ctx)
+
+            if not resource_check_results:
```
IsaacYangSLA, line 231: **P1 — Treat partial resource-check results as unsafe too.** This only rejects an entirely empty result. If `resource_reqs` contains `site1` and `site2` but the result contains only a successful `site1`, a job with `min_sites=1` is admitted even though `site2` may have an untracked reservation. I reproduced that path. Please compare expected and returned site sets and stop the pass whenever an expected result is missing.

nvflare/app_common/job_schedulers/job_scheduler.py
```diff
@@ -195,69 +200,84 @@ def _try_job(self, job: Job, fl_ctx: FLContext) -> (int, Optional[Dict[str, Disp
...
+                app_name=sites_to_app[SERVER_SITE_NAME], resource_requirements={}, token=None
             )
-
-        # add server dispatch info
-        sites_dispatch_info[SERVER_SITE_NAME] = DispatchInfo(
-            app_name=sites_to_app[SERVER_SITE_NAME], resource_requirements={}, token=None
-        )
-
-        return SCHEDULE_RESULT_OK, sites_dispatch_info, ""
+            keep_resources = True
```
IsaacYangSLA, line 303: **P2 — Retain cleanup ownership through post-admission bookkeeping.** Setting `keep_resources` here ends the cleanup guard before `_do_schedule_job()` updates schedule history and returns the ready job. If that later processing raises, the outer handler returns no job while the successful reservation remains active; I reproduced it with malformed schedule-history metadata. Please keep cleanup ownership until the ready job is successfully returned, or cancel the returned dispatch reservations when post-admission processing fails.

IsaacYangSLA (commented):
I found three resource-lifecycle gaps that can leave reservation state unknown or leak a successful reservation. The submitted 30 scheduler tests and targeted Black/Flake8 checks pass, but the cases described inline are not covered.

nvflare/app_common/job_schedulers/job_scheduler.py
YuanTingHsieh: Addressed in `585c3db7d`. `_do_schedule_job()` now retains ownership of a successful dispatch through logging and schedule-history bookkeeping. If post-admission processing raises, it reconstructs cancellation inputs from the returned `DispatchInfo`, cancels those reservations, records the failed attempt, and continues only after cleanup succeeds; an uncertain cleanup stops the pass. Added a regression where the first history update raises, the first token is cancelled, and a later candidate schedules with a new token.

nvflare/app_common/job_schedulers/job_scheduler.py
YuanTingHsieh: Addressed in `585c3db7d`. The scheduler now compares the exact expected and returned site sets before evaluating admission. Missing or unexpected sites raise `_UnsafeAdmissionError`; known successful reservations are cancelled by the existing `finally` guard, and the pass stops because reservations at missing sites cannot be proven clean. Added a two-site partial-reply regression that verifies known cleanup, retry bookkeeping, and that the later candidate is not examined.

nvflare/app_common/job_schedulers/job_scheduler.py
YuanTingHsieh: Addressed in `585c3db7d`. `ServerEngine.cancel_client_resources()` now requires a positive acknowledgement for every site with a reservation token and raises on missing/time-out replies, non-OK message return codes, malformed acknowledgement bodies, and non-OK cancellation return codes (including `EXECUTION_EXCEPTION`). The scheduler already converts that cleanup failure into `_UnsafeAdmissionError`, so the candidate scan stops. Added production-level success and parameterized failure tests; the combined focused suites pass all 69 tests.

pcnudde:
The valuable core of this PR is that cleanup now always runs: the `finally`-centralized cancellation in `_try_job`, the post-admission cleanup via `_cancel_dispatch_resources`, and the bare `except` fixes. Please keep all of that. But I don't think the hard-fail half — raising on unacknowledged cancels and aborting the whole scheduling pass — is the right trade, and I'd like it changed to verify-and-log.

**Why the abort is disproportionate:**
- The failure it guards against self-heals: all built-in resource managers inherit `AutoCleanResourceManager`'s expiry (default 30s), so an unacked cancel leaks a reservation for at most ~30s. Nothing is running against that reservation; it's a hold taken while asking "can you fit this job?".
- The abort protects nothing in practice: `JobRunner` re-runs `schedule_job` ~1s later, so later candidates get checked against the (still-live, <=30s) reservation on the very next pass anyway.
- Meanwhile the triggers are routine and some are deterministic: a client disconnecting between check and cancel can *never* ack (`reservation_sites` is populated before the `if client:` guard, so no request is even built); a client that passes the 15s check window can reliably miss the 10s cancel-ack default; a lost reply after a successful client-side cancel also raises. Each occurrence stalls the entire queue for that pass and burns the affected job's `SCHEDULE_COUNT`; a client down for a few minutes drives a perfectly schedulable job to `FINISHED_CANT_SCHEDULE` with a misleading "failed to cancel resources" history.

**Requested changes:**
1. `cancel_client_resources`: keep the reply verification, but log the per-site errors (error level) instead of raising. Reuse `check_client_replies()` from `admin.py` for the verification — it already does missing/timeout/return-code checks — plus the new `get_return_code()` body check. Revert the `Raises:` addition to `ServerEngineSpec` (nothing enforces it and other implementations don't honor it).
2. Drop `_UnsafeAdmissionError` and the completeness raises in `_try_job` (missing sites / empty results are a normal client-disconnect race the old code handled; the non-dict check can stay as a logged skip). Failed admission of one job should `continue` to the next candidate, as before.
3. If we want a stronger guarantee against leaks for custom resource managers, the right place is `ResourceManagerSpec` — document/require reservation expiry there rather than compensating in the scheduler.
4. Independent fixes to keep regardless: guard the second `_update_schedule_history` call in the except handler (a deterministic history failure currently escapes and defeats backoff entirely); preserve the original exception when wrapping (the current `from` chaining hides the root-cause admission error from logs and history); drop the dead `attempt_recorded` flag and the redundant `rc == SCHEDULE_RESULT_OK` conjunct.

This keeps every leak observable (loud logs on unacked cancels) without letting one slow or departed client stop scheduling for the whole queue. Most of the new tests would shift from asserting scan-abort to asserting log-plus-continue.

PR #5195, opened 2026-08-24
Fix cyclic external result materialization

YuanTingHsieh:
## Summary

- declare that CyclicClientController consumes a materialized learn-task result
- when ClientAPIExecutor is called from a nested client-side workflow task, resolve the executor responsible for the outer active task through ClientRunner and honor its local-consumer declaration
- preserve existing task-result-filter and component materialization for direct tasks while keeping unrelated nested workflows in pass-through mode
- add production-topology regression coverage where the cyclic controller is an executor and is absent from engine components

## Testing

- targeted cyclic controller and ClientAPIExecutor tests: 121 passed
- broader CCWF and ClientAPIExecutor tests: 283 passed, 2 subtests passed
- full project style check: Black, isort, flake8, and agent-skill lint passed
- fresh GPU qualification at exact signed head b2d77cc5a51fe5c656aff79c41958734b5399474: 132 focused tests passed
- unchanged two-client external-process cyclic POC job: FINISHED:COMPLETED in 35.89 seconds; both clients completed both rounds with no LazyDownloadRef arithmetic or TypeError signatures


IsaacYangSLA (approved):
Seems later we can consolidate those calls.  LGTM.

nvidianz (approved):

PR #5198, opened 2026-08-25
Fix watchdog PID tracking

IsaacYangSLA:
## Summary

- launch the watchdog-managed FL application with `exec` so the recorded PID belongs to the Python process
- capture and persist the background PID directly, then reap the process after forced termination
- add a regression test for the generated startup script's PID tracking behavior

## Testing

- `.venv/bin/python -m pytest -q tests/unit_test/lighter/static_file_builder_test.py` (22 passed)
- `./runtest.sh -s`


YuanTingHsieh (approved):
LGTM

PR #5199, opened 2026-08-25
Advertise protocol and NVFlare versions during registration

pcnudde:
## Summary
- define federation protocol version 1
- advertise the protocol and NVFlare release versions in the client registration request
- advertise the protocol and NVFlare release versions in the successful server registration response
- leave compatibility enforcement unchanged in 2.9

## Rationale
This seeds additive registration fields so a future release can detect protocol compatibility in both directions and report the peer software release in mismatch diagnostics. Existing peers continue to ignore the extra fields, so this does not change 2.9 admission behavior.

## Testing
- ./runtest.sh -s
- python -m pytest -q tests/unit_test/private/fed/authenticator_test.py tests/unit_test/private/fed/server/fed_server_test.py (66 passed)

### DIFF
--- a/nvflare/private/fed/authenticator.py
+++ b/nvflare/private/fed/authenticator.py
@@ -32,6 +32,7 @@ from nvflare.fuel.f3.cellnet.fqcn import FQCN
 from nvflare.fuel.f3.message import Message
 from nvflare.fuel.f3.message import Message as CellMessage
 from nvflare.fuel.utils.log_utils import get_obj_logger
+from nvflare.private import defs as private_defs
 from nvflare.private.defs import CellChannel, CellChannelTopic, CellMessageHeaderKeys, ClientRegMsgKey, new_cell_message
 from nvflare.private.fed.utils.identity_utils import (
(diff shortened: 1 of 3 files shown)


nvflare/private/fed/authenticator.py
```diff
@@ -247,6 +248,7 @@ def authenticate(self, shared_fl_ctx: FLContext, abort_signal: Signal):
             CellMessageHeaderKeys.CLIENT_TYPE: self.client_type,
             CellMessageHeaderKeys.CLIENT_IP: local_ip,
             CellMessageHeaderKeys.PROJECT_NAME: self.project_name,
+            CellMessageHeaderKeys.FEDERATION_PROTOCOL_VERSION: private_defs.FEDERATION_PROTOCOL_VERSION,
```
nvidianz, line 252: We should add NVFLare version like 2.8.1 here too.

pcnudde: added

nvidianz (approved):

PR #5205, opened 2026-08-25
[2.7] Surface client job log configuration failures

chesterxgchen:
## Summary
- backport #5138 to the 2.7 release branch
- propagate client-side configure_job_log failures through a legacy HCI ERROR data item and detailed command metadata
- report non-OK, missing, and partial client replies while preserving successful no-client all behavior
- add focused release-compatible regression coverage for both response representations because the newer Session and CLI test scaffolding is not present on 2.7

## Validation
- python3 -m pytest tests/unit_test/private/fed/server/job_cmds_test.py tests/unit_test/private/fed/server/job_cmds_configure_log_test.py -q (14 passed)
- ./runtest.sh -s

### DIFF
--- a/nvflare/private/fed/server/job_cmds.py
+++ b/nvflare/private/fed/server/job_cmds.py
@@ -328,7 +328,8 @@ class JobCommandModule(CommandModule, CommandUtil, BinaryTransfer):
                     client_errors.append(f"{client_name}: {detail}")
 
             if client_errors:
-                conn.update_meta(make_meta(MetaStatusValue.ERROR, info="\n".join(client_errors)))
+                error_info = "\n".join(client_errors)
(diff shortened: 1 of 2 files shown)


YuanTingHsieh (approved):
Approving. One non-blocking P3 follow-up: `configure_job_log ... all ...` can still raise when a running job has no currently connected clients because `send_request_to_clients()` returns `None` and `process_replies_to_table()` iterates it before the replies are normalized. This is a narrow edge case and can be addressed consistently across main, 2.8, and 2.7.

chesterxgchen:
Addressed consistently across the supported branches:

- main / 2.9: #5217
- 2.8: #5219
- 2.7: #5218

The fix normalizes a missing client reply collection before `process_replies_to_table()` runs, preserving the existing no-response output when `configure_job_log ... all ...` has no connected clients. Each branch includes regression coverage for the actual `None` return.

PR #5206, opened 2026-08-25
[2.8] Surface client job log configuration failures

chesterxgchen:
## Summary
- backport #5138 to the 2.8 release branch
- propagate client-side configure_job_log failures through a legacy HCI ERROR data item and detailed command metadata
- report non-OK, missing, and partial client replies while preserving successful no-client all behavior
- include regression coverage for both response representations

## Validation
- python3 -m pytest tests/unit_test/private/fed/server/job_cmds_test.py tests/unit_test/fuel/flare_api/session_new_methods_test.py tests/unit_test/tool/job/job_log_test.py -q (180 passed)
- ./runtest.sh -s

### DIFF
--- a/nvflare/private/fed/server/job_cmds.py
+++ b/nvflare/private/fed/server/job_cmds.py
@@ -399,15 +399,27 @@ class JobCommandModule(CommandModule, CommandUtil, BinaryTransfer):
             replies = self.send_request_to_clients(conn, message)
             self.process_replies_to_table(conn, replies)
 
+            client_errors = []
             client_replies = replies or []
-            expected_tokens = set((conn.get_prop(self.TARGET_CLIENTS, {}) or {}).keys())
             received_tokens = {client_reply.client_token for client_reply in client_replies}
-            if not expected_tokens.issubset(received_tokens) or any(
-                client_reply.reply is None
-                or client_reply.reply.get_header(MsgHeader.RETURN_CODE, AdminReturnCode.OK) != AdminReturnCode.OK
-                for client_reply in client_replies
-            ):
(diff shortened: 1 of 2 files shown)


nvflare/private/fed/server/job_cmds.py
```diff
@@ -399,15 +399,27 @@ def configure_job_log(self, conn: Connection, args: List[str]):
...
+                    client_errors.append(f"{client_name}: no reply")
+                    continue
+
+                return_code = client_reply.reply.get_header(MsgHeader.RETURN_CODE, AdminReturnCode.OK)
+                if return_code != AdminReturnCode.OK:
+                    detail = client_reply.reply.body or f"return code {return_code}"
+                    client_errors.append(f"{client_name}: {detail}")
+
+            if client_errors:
+                conn.update_meta(make_meta(MetaStatusValue.ERROR, info="\n".join(client_errors)))
```
YuanTingHsieh: This update records the failure only in response metadata and removes the ERROR data item previously emitted by append_error. The legacy HCI/admin client renders response data but does not display metadata, so missing or partial client failures can appear successful there. Please preserve an error data item while keeping the detailed metadata, for example by using append_error with make_meta.

chesterxgchen: Fixed in c96400d34. The failure path now emits the detailed per-client text as an ERROR data item via append_error while passing the same make_meta payload, so legacy HCI displays the failure and Session/CLI consumers retain detailed ERROR metadata. Regression assertions cover both representations. The same fix is applied to #5205, and the main follow-up to #5138 is #5212.

YuanTingHsieh (approved):

PR #5208, opened 2026-08-26
Fix FedJob transitive package import export

YuanTingHsieh:
## Summary

- fix FedJob export recursion for qualified package imports reached from a top-level external script
- preserve the parent-directory helper lookup added by #4885
- carry the registered script's ordered roots through unqualified flat-helper recursion
- keep qualified package modules anchored to the root where they were resolved
- prevent both missing project dependencies and same-named parent files from corrupting parent-helper exports
- add regression coverage for package recursion, flat siblings, source-root precedence, parent-package anchoring, and parent-helper back-imports

## Root cause

For a depth-0 registered script, the exporter always moved source_root one directory above the project. Direct imports could still resolve from the script directory, but recursive package imports inherited the parent root. As a result, pkg.a importing pkg.b either omitted the project module or copied a same-named file from the parent directory.

A single recursive root also cannot model flat helpers correctly: subdirectory siblings need the project root as a fallback, while parent helpers need the original project-first lookup order for their own unqualified imports. This change carries that ordered flat-import root chain across unqualified modules and drops it when entering a qualified package, which remains anchored to its matched root.

## Testing

- 55 tests passed: tests/unit_test/job_config/fed_job_config_test.py
- project style checks passed for both changed files (black, isort, flake8, and agent skill lint)
- public FedJob.export_job reproductions passed for qualified packages, transitive flat siblings, and parent-helper project back-imports with and without a parent shadow
- a 12-case differential matrix against tag 2.8.1 found zero compatibility regressions

### DIFF
--- a/nvflare/job_config/fed_job_config.py
+++ b/nvflare/job_config/fed_job_config.py
@@ -492,7 +492,7 @@ class FedJobConfig:
         self._check_destination_collision(source_file, dest_file)
         return source_file, source_root, dest_file
 
-    def _get_custom_file(self, custom_dir, module, source_file, source_root=None):
+    def _get_custom_file(self, custom_dir, module, source_file, source_root=None, flat_import_roots=None):
         module_parts = self._module_parts(module)
         if source_root is None:
             source_root = self._derive_source_root(module=module, source_file=source_file)
@@ -514,7 +514,14 @@ class FedJobConfig:
 
         self.custom_modules.append(module)
         try:
-            self._copy_source_file(custom_dir, module, source_file, dest_file, source_root=source_root)
+            self._copy_source_file(
+                custom_dir,
+                module,
+                source_file,
+                dest_file,
+                source_root=source_root,
+                flat_import_roots=flat_import_roots,
+            )
         except Exception:
             self.custom_modules.remove(module)
             raise
@@ -538,7 +545,16 @@ class FedJobConfig:
         resolved_parts = package_parts[:keep_parts] + import_parts
         return ".".join(resolved_parts) if resolved_parts else None
 
-    def _copy_source_file(self, custom_dir, module, source_file, dest_file, source_root, is_external_script=False):
+    def _copy_source_file(
+        self,
+        custom_dir,
+        module,
+        source_file,
+        dest_file,
+        source_root,
+        is_external_script=False,
+        flat_import_roots=None,
+    ):
         source_file, source_root, dest_file = self._validate_copy_paths(
             custom_dir=custom_dir,
             source_file=source_file,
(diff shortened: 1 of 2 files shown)


nvkevlu:
Greptile has already found this, but when a match comes from source_dir, narrowing recursion to that directory loses the original project-root fallback. For example, src/client.py importing src/net.py, which imports project-root root_helper.py, omits root_helper.py from the export and subsequently fails at import time. Please retain the original project root as a fallback for flat-sibling matches while keeping package/parent matches anchored correctly, and add a transitive regression test.

nvflare/job_config/fed_job_config.py
```diff
@@ -556,16 +572,20 @@ def _copy_source_file(self, custom_dir, module, source_file, dest_file, source_r
...
             import_module = self._resolve_import_module(module, import_source, level, source_file)
             if not import_module:
                 continue
             import_path = os.path.join(*self._module_parts(import_module)) + ".py"
-            search_roots = [source_root]
-            # Flat registered scripts and non-package modules can resolve unqualified imports from source-dir siblings.
-            if level == 0 and search_source_dir and os.path.basename(source_file) != "__init__.py":
-                search_roots.insert(0, source_dir)
+            # Flat modules retain the registered script's ordered roots; package modules stay anchored to their root.
+            search_roots = list(flat_import_roots) if level == 0 and is_flat_module else [source_root]
```
holgerroth, line 588: Verified this fixes both failure modes of the underlying export bug: the transitive package repro (`train.py` -> `pkg.a` -> `pkg.b` through the public `FedJob.export_job`) now packages `pkg/b.py` and the exported app imports cleanly, and with a same-named decoy in the parent directory the project file wins. Thanks also for locking in the parent-helper ordering with the parametrized shadow tests.

One non-blocking follow-up idea: when an import spec matches none of the search roots, the loop below still falls through silently, so a project module the resolver cannot map (an import shape outside the supported ones) is dropped from the export and only surfaces as `ModuleNotFoundError` at the deployed sites, far from the cause. An unconditional warning would be noisy since third-party imports also land here, but a debug-level log of unresolved import specs - or a warning gated to modules whose top-level name exists as a `.py` file or package directory under one of the candidate roots - would make the next unsupported-shape gap diagnosable at export time instead of at run time. Fine as a separate PR if you'd rather keep this one scoped.

nvkevlu (approved):

nvflare/job_config/fed_job_config.py
YuanTingHsieh: Thanks for verifying both failure modes. Agreed that unresolved-import visibility is a real diagnostics gap, but it predates this change: import discovery is best-effort, and unresolved specs also include third-party dependencies, so a useful warning policy needs separate consideration to avoid noise. The FedJob guide already documents best-effort discovery and notes that imports which cannot be resolved locally may not be discovered. We will keep this code-freeze fix scoped and handle improved diagnostics or more explicit import-shape guidance as a follow-up.

PR #5213, opened 2026-08-26
[2.9] Improve Lightning and Hugging Face conversion skill reliability

chesterxgchen:
## Summary
- tighten nvflare-convert-lightning triggering so it requires both federation intent—either explicit federated/NVFLARE conversion or collaborative training across sites while data remains local—and verified PyTorch Lightning ownership, with adjacent-negative routing evals for DDP, profiling, serving, training-loop, and TensorFlow/Keras requests
- add a positive Lightning routing eval for implicit federation intent expressed as collaborative cross-hospital training without moving data, protecting recall without naming FLARE or federated learning
- move the Lightning standard-path reference order ahead of the detailed workflow and consolidate non-standard conditional loads to reduce unnecessary reference reads without changing validation gates
- document shared PyTorch-family Recipe contracts for ExchangeFormat/TransferType imports and the required set_per_site_config ordering before files, filters, or components
- make Hugging Face rank resolution product-owned: nvflare.client.hf.init resolves an initialized process-group rank or global RANK, canonicalizes validated environment ranks, rejects negative ranks, and rejects unresolved multi-process launches before creating a Client API context; generated clients remain rankless and contain no launcher-parsing logic
- centralize the foreground final-simulation and no-active-finalization invariant in the shared validation reference used by both Lightning and Hugging Face converters
- make Hugging Face source-data handling conditional: remove the generic --data_root assumption, add an opt-in resolve_source_local_path helper, preserve source argument names, and distinguish source-relative, absolute, per-site, Hub, URL, and no-path inputs
- keep Hugging Face dependency inventory separate from optional cache/filesystem discovery, defer artifact availability checks to the maintained resolver, and make separately necessary missing-path probes exit zero
- add focused eval fixtures and deterministic regression coverage for trigger selection, reference ordering, Recipe construction, product-owned distributed rank behavior, source-path resolution, dependency-probe isolation, per-site overrides, and alternate caller working directories

## Validation
- `python3 -m pytest tests/unit_test/app_opt/hf -q` — 112 passed, 8 skipped
- `python3 -m pytest tests/unit_test/tool/agent_skill_checks -q` — 491 passed, 37 skipped
- `./runtest.sh -s` — Black, isort, Flake8, and agent-skill lint passed
- pre-push agent-skill lint — 0 errors, 0 warnings, 0 informational findings

### DIFF
--- a/skills/nvflare-convert-huggingface/assets/client_with_eval.py
+++ b/skills/nvflare-convert-huggingface/assets/client_with_eval.py
@@ -13,40 +13,13 @@ best-model selection is required. Set it to ``False`` only for a valid
 train-only source path.
 
 The entry is rankless for CPU, single-GPU, and other single-process training.
-With an initialized multi-process group, it passes
-``torch.distributed.get_rank()`` to FLARE. If environment markers declare a
-multi-process launch before the process group is initialized, global ``RANK``
-must already be available; otherwise the entry fails before FLARE initialization.
-For a preserved multi-GPU launch, follow
+The Hugging Face Client API resolves an initialized process-group rank or global
+``RANK`` and rejects an unresolved multi-process launch. For a preserved
+multi-GPU launch, follow
 ``references/huggingface-state-and-distributed.md``.
 """
 
-import os
-
-import torch.distributed as dist
-
 import nvflare.client.hf as flare
-
-
-_MULTIRANK_SIZE_ENV_VARS = ("WORLD_SIZE", "LOCAL_WORLD_SIZE", "OMPI_COMM_WORLD_SIZE", "SLURM_NTASKS")
-
-
-def _environment_declares_multirank() -> bool:
-    for name in _MULTIRANK_SIZE_ENV_VARS:
-        try:
-            if int(os.environ.get(name, "1") or 1) > 1:
-                return True
-        except (TypeError, ValueError):
-            continue
-    return False
-
-
-def _has_valid_global_rank_env() -> bool:
-    try:
-        int(os.environ["RANK"])
-    except (KeyError, TypeError, ValueError):
-        return False
-    return True
 
 
 def make_hf_argument_parser(dataclass_types):
@@ -58,17 +31,7 @@ def make_hf_argument_parser(dataclass_types):
 
 def main(trainer_factory, *, evaluate_before_train=True):
     """Run one persistent patched Trainer in single-process or distributed mode."""
-    if dist.is_available() and dist.is_initialized() and dist.get_world_size() > 1:
-        flare.init(rank=dist.get_rank())
-    else:
-        if _environment_declares_multirank() and not _has_valid_global_rank_env():
-            raise RuntimeError(
-                "multi-process launch detected but global RANK is unavailable; initialize torch.distributed "
-                "or export a valid global RANK before FLARE Client API initialization"
-            )
-        # CPU and single-GPU runs stay rankless. Under torchrun, flare.init()
-        # resolves the global RANK environment value.
-        flare.init()
+    flare.init()
     trainer = trainer_factory()
     flare.patch(trainer)
 
--- a/skills/nvflare-convert-huggingface/assets/client_with_eval.py
+++ b/skills/nvflare-convert-huggingface/assets/client_with_eval.py
@@ -14,13 +14,39 @@ train-only source path.
 
 The entry is rankless for CPU, single-GPU, and other single-process training.
 With an initialized multi-process group, it passes
-``torch.distributed.get_rank()`` to FLARE. For a preserved multi-GPU launch, follow
+``torch.distributed.get_rank()`` to FLARE. If environment markers declare a
+multi-process launch before the process group is initialized, global ``RANK``
+must already be available; otherwise the entry fails before FLARE initialization.
+For a preserved multi-GPU launch, follow
 ``references/huggingface-state-and-distributed.md``.
 """
+
+import os
 
 import torch.distributed as dist
 
 import nvflare.client.hf as flare
+
+
+_MULTIRANK_SIZE_ENV_VARS = ("WORLD_SIZE", "LOCAL_WORLD_SIZE", "OMPI_COMM_WORLD_SIZE", "SLURM_NTASKS")
+
+
+def _environment_declares_multirank() -> bool:
+    for name in _MULTIRANK_SIZE_ENV_VARS:
+        try:
+            if int(os.environ.get(name, "1") or 1) > 1:
+                return True
+        except (TypeError, ValueError):
+            continue
+    return False
+
+
+def _has_valid_global_rank_env() -> bool:
+    try:
+        int(os.environ["RANK"])
+    except (KeyError, TypeError, ValueError):
+        return False
+    return True
 
 
 def make_hf_argument_parser(dataclass_types):
@@ -35,6 +61,11 @@ def main(trainer_factory, *, evaluate_before_train=True):
     if dist.is_available() and dist.is_initialized() and dist.get_world_size() > 1:
         flare.init(rank=dist.get_rank())
     else:
+        if _environment_declares_multirank() and not _has_valid_global_rank_env():
+            raise RuntimeError(
+                "multi-process launch detected but global RANK is unavailable; initialize torch.distributed "
+                "or export a valid global RANK before FLARE Client API initialization"
+            )
         # CPU and single-GPU runs stay rankless. Under torchrun, flare.init()
         # resolves the global RANK environment value.
         flare.init()
--- a/skills/nvflare-convert-huggingface/assets/client_with_eval.py
+++ b/skills/nvflare-convert-huggingface/assets/client_with_eval.py
@@ -12,10 +12,13 @@ Keep ``evaluate_before_train=True`` when source-backed per-round evaluation or
 best-model selection is required. Set it to ``False`` only for a valid
 train-only source path.
 
-This standard asset assumes one training process per site. For a preserved
-distributed multi-process launch, adapt Client API initialization using
+The entry is rankless for CPU, single-GPU, and other single-process training.
+With an initialized multi-process group, it passes
+``torch.distributed.get_rank()`` to FLARE. For a preserved multi-GPU launch, follow
 ``references/huggingface-state-and-distributed.md``.
 """
+
+import torch.distributed as dist
 
 import nvflare.client.hf as flare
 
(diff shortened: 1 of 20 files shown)


tests/unit_test/tool/agent_skill_checks/seed_skills_test.py
```diff
@@ -952,6 +952,30 @@ def test_orientation_routes_only_unresolved_explicit_conversions():
...
+        "lightning-negative-tensorflow-keras",
+    }
+
+    assert "use only when the request expresses federated or NVFLARE conversion intent" in description
+    assert "Lightning source evidence alone is not sufficient" in skill_text
+    for eval_id in negative_eval_ids:
+        negative_eval = _eval_by_id(eval_data, eval_id)
+        assert negative_eval["files"]
+        assert negative_eval["nvflare"]["expected_skill"] is None
+        assert negative_eval["nvflare"]["negative_for"] == "nvflare-convert-lightning"
```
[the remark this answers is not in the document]
chesterxgchen: Thanks. The pytest check is intentionally a trigger-contract preservation test, not a behavioral skill router: NVFlare has no in-repo routing implementation to execute because skill selection is performed by the external agent/model. The five evals.json cases are the executable Harbor routing cases for DDP-only, profiling, inference serving, training-loop changes, and TensorFlow/Keras. Adding a local pytest classifier would test a duplicate implementation rather than production behavior. Commit af78b36bc renames and documents the test to make that scope explicit. Behavioral confirmation should come from rerunning these cases through the model-backed trigger evaluation.

pcnudde (dismissed):

skills/nvflare-shared/references/site-data-and-paths.md
```diff
@@ -43,12 +43,39 @@ Apply the `conversion-common.md` "Data Location" invariants first: a configurabl
...
-reinterpret it relative to packaged `client.py`, the export directory, or the
-simulator process working directory. Validate relative-path conversions from a
-fresh caller working directory.
+Classify each detected input before adapting `job.py`, preserving the source's
+argument name and semantics:
+
+- For a source-project-relative local file or directory, resolve it in `job.py`
+  against the source-project root before Recipe construction. Hugging Face
+  conversions use the maintained `resolve_source_local_path()` helper, which
+  resolves against `SOURCE_DIR`, not the caller's working directory.
```
holgerroth: This equates `SOURCE_DIR` with the source-project root, but in the maintained asset `SOURCE_DIR = Path(__file__).resolve().parent` — the *generated* `job.py`'s directory. Those diverge in the read-only-source flow that `runtime-output-guidance.md` explicitly supports: with source at read-only `/src` (data at `/src/datasets/sst2`) and the adapted `job.py` written to `/out`, `resolve_source_local_path("datasets/sst2")` resolves to `/out/datasets/sst2`, where no data exists. The new validation rung in `huggingface-validation.md` then *certifies* that wrong value ("absolute and equal to the expected location under `SOURCE_DIR`"), so the error only surfaces later as a data-load failure in the full simulation.

The sibling `scripts/resolve_model_snapshot.py` takes an explicit `--source-root` for exactly this reason. Suggest either giving the helper (or its documented adaptation) an explicit source-root anchor, or constraining this guidance to the colocated case and directing the read-only-source flow to resolve against the inspected source root instead.

skills/nvflare-convert-huggingface/references/huggingface-validation.md
```diff
@@ -38,6 +39,15 @@ reaches the applicable phase, and stop at the first failure.
...
   structured argument surface exists, fail closed.
+- Only when source inspection found a source-project-relative local file or
+  directory argument, change to a fresh working directory outside the project,
+  construct the Recipe with that relative source value, and inspect the final
+  `train_args`. Require the source argument name to be unchanged and its
+  transmitted value to be absolute and equal to the expected location under
+  `SOURCE_DIR` before running the full simulation. Skip this path-specific check
+  when the source has no local path argument. Do not pass absolute local paths,
+  per-site paths, Hub identifiers, or URLs through the relative-path test;
+  validate their classification using `site-data-and-paths.md` instead.
```
holgerroth: Broken reference: bare `site-data-and-paths.md` resolves to this `references/` directory, where no such file exists. Elsewhere in this file, bare names denote same-directory files (`huggingface-state-and-distributed.md`) and cross-directory references use the prefix (`../../nvflare-shared/references/validation-evidence.md` on line 3). Should be `../../nvflare-shared/references/site-data-and-paths.md` so the agent can actually load the classification contract at the moment it's directed there.

skills/nvflare-convert-huggingface/assets/client_with_eval.py
```diff
@@ -23,9 +30,14 @@ def make_hf_argument_parser(dataclass_types):
...
-    """Run one persistent patched Trainer using the caller's resolved global rank."""
-    flare.init(rank=rank)
+def main(trainer_factory, *, evaluate_before_train=True):
+    """Run one persistent patched Trainer in single-process or distributed mode."""
+    if dist.is_available() and dist.is_initialized() and dist.get_world_size() > 1:
+        flare.init(rank=dist.get_rank())
+    else:
+        # CPU and single-GPU runs stay rankless. Under torchrun, flare.init()
+        # resolves the global RANK environment value.
+        flare.init()
```
holgerroth: This drops the old contract's early rejection ("Reject a multi-process launch when global rank cannot be resolved" was also deleted from `huggingface-state-and-distributed.md`). The distributed reference only requires `torch.distributed` init before `flare.patch(trainer)`, not before this entry runs — so under a launcher that sets no `RANK` env (mpirun/SLURM expose `OMPI_COMM_WORLD_RANK`/`SLURM_PROCID`) with the process group initialized inside trainer construction, every worker takes this rankless branch and the Client API defaults rank to `"0"`. `flare.patch()`'s guards in `nvflare/app_opt/hf/api.py` do raise a `RuntimeError` later, but only after each worker completed a full rank-0 Client API init — in ex-process mode that's N concurrent rank-0 cell sessions before the crash.

Suggest a fail-fast here when multirank environment markers exist without a resolvable global rank (e.g. `WORLD_SIZE`/`LOCAL_WORLD_SIZE` > 1 while `dist` is uninitialized and `RANK` is unset), restoring the removed early rejection.

skills/nvflare-shared/references/site-data-and-paths.md
chesterxgchen: Addressed in 6cf72dc07. `resolve_source_local_path()` now requires an explicit absolute `source_root`; `SOURCE_DIR` is valid only for colocated generation. The read-only-source guidance and HF validation now require the inspected original source root rather than the generated output directory. Regression coverage loads `job.py` from a separate generated directory and verifies that a relative dataset path resolves under the distinct source root.

skills/nvflare-convert-huggingface/assets/client_with_eval.py
chesterxgchen: Addressed in 6cf72dc07. The maintained HF client asset now fails before `flare.init()` when `WORLD_SIZE`, `LOCAL_WORLD_SIZE`, `OMPI_COMM_WORLD_SIZE`, or `SLURM_NTASKS` declares multiple processes but neither an initialized process group nor a valid global `RANK` is available. Tests cover every marker and retain the delayed-process-group path when global `RANK` is present.

skills/nvflare-convert-huggingface/references/huggingface-validation.md
chesterxgchen: Addressed in 6cf72dc07. The reference now uses the complete `../../nvflare-shared/references/site-data-and-paths.md` path, with a regression assertion preserving that link.

holgerroth (dismissed):
Approving — all three review comments are addressed in 6cf72dc07 and verified against the code:

- `resolve_source_local_path()` now takes an explicit absolute `source_root` (with validation), and the references/evals consistently scope `SOURCE_DIR` to the colocated case only.
- The `site-data-and-paths.md` cross-reference is fixed.
- The client asset fails fast before `flare.init()` when multi-rank environment markers (`WORLD_SIZE`, `LOCAL_WORLD_SIZE`, `OMPI_COMM_WORLD_SIZE`, `SLURM_NTASKS`) are present without a resolvable global `RANK` — covering the mpirun/SLURM corner.

CI is green across the full matrix, and I ran the two changed test files locally on this head (161 passed, 1 skipped).

Two non-blocking follow-up suggestions:

1. The "run the final simulation with `run_in_background: false`; never finalize with an active simulation" rule currently lives only in the Hugging Face SKILL.md. The Lightning converter follows the same shared validation ladder and has the same false-green risk — consider moving the rule into `nvflare-shared/references/validation-evidence.md` so both converters inherit it.
2. The new Lightning triggering negatives all state "do not federate" explicitly. There is no positive eval guarding recall for *implicit* federation intent (e.g. "train this collaboratively across 3 hospitals without moving the data"). Worth adding one before the next skill-eval run so the tightened description doesn't over-correct.

skills/nvflare-convert-huggingface/assets/client_with_eval.py
chesterxgchen: Agreed. Commit fd066496f moves this invariant into the product instead of adding more generated-code logic. The Hugging Face-specific `nvflare.client.hf.init()` wrapper now resolves an initialized process-group rank or global `RANK`, and rejects `WORLD_SIZE`, `LOCAL_WORLD_SIZE`, OpenMPI, or Slurm multi-rank launches with no resolvable global rank before the generic Client API initializer can create a context. The generated HF asset is back to an unconditional `flare.init()` call; its local environment parsing, `torch.distributed` import, and associated prescriptive prose/tests were removed. Product tests now own the single-process, initialized-group, delayed-group/global-RANK, explicit-rank, and fail-before-context cases. I kept this in the HF API rather than generic `nvflare.client.init()` so framework-specific launcher semantics do not affect unrelated clients.

holgerroth (approved):
Re-approving at the current head after reviewing the three commits pushed since my earlier approval:

- **Product-level rank enforcement** — `nvflare.client.hf.init()` now owns distributed-rank resolution (initialized process group → `dist.get_rank()`; declared multi-rank env requires a valid non-negative global `RANK`, rejected with a clear error *before* any Client API context is created), with dedicated contract tests including normalization and negative-rank edge cases. The skill asset collapses back to a rankless `flare.init()` and the corresponding guard prose is deleted — net reduction in skill mass with stronger guarantees.
- **Shared finalization rule** — the foreground-simulation requirement moved from HF SKILL.md into `nvflare-shared/references/validation-evidence.md` as a framework-neutral rule, so all converters inherit it.
- **Implicit federation intent** — the trigger description now covers "multiple sites training collaboratively while keeping each site's data local" (with a positive eval that deliberately avoids the words FLARE/NVFLARE/federated), closing the recall-side risk of the tightened triggering.

Verified: CI fully green at this head; locally 505 agent-skill checks and 404 client/HF product tests pass.

Remaining follow-ups (non-blocking, post-merge cleanup):
1. Unify the two multirank-env helpers — `nvflare/client/hf/__init__.py` checks `WORLD_SIZE`/`LOCAL_WORLD_SIZE`/`OMPI_COMM_WORLD_SIZE`/`SLURM_NTASKS` while `nvflare/app_opt/hf/api.py::_env_declares_multirank` checks only the first two; they should share one helper.
2. Consider generalizing the rank enforcement beyond the `hf` namespace so the Lightning/PyTorch converters can shed their rank prose the same way.

PR #5216, opened 2026-08-26
Fix concurrent job lifecycle accounting

### LINKED ISSUE #5215: Fix concurrent job lifecycle accounting using explicit event job IDs
## Description

`DefaultJobScheduler` currently identifies `JOB_STARTED`, `JOB_COMPLETED`, and `JOB_ABORTED` events using the sticky `CURRENT_JOB_ID` value from `FLContext`.

Job launch and completion processing can run concurrently. Because sticky context properties are shared across derived contexts, another job can replace `CURRENT_JOB_ID` before a lifecycle handler reads it. The event can then be attributed to the wrong job.

## Impact

Incorrect event attribution corrupts `scheduled_jobs` accounting. A completed job can remain counted as active, or another job can be removed instead. Once stale entries reach `max_jobs`, the scheduler can reject valid admissions even when execution capacity is available. This can collapse concurrency to serial execution or leave jobs waiting indefinitely.

This is an accounting correctness bug; reduced throughput is one observable consequence.

## Expected behavior

Every lifecycle event must update scheduler accounting for the job that actually generated that event, independent of subsequent changes to sticky context state.

## Proposed fix

- Publish the originating job ID as non-sticky lifecycle event data from `JobRunner`.
- Prefer that explicit event job ID in `DefaultJobScheduler`.
- Retain `CURRENT_JOB_ID` as a compatibility fallback for existing event publishers.
- Cover concurrent sticky-context mutation and lifecycle publication with focused regression tests.

## Acceptance criteria

- Start, completion, abort, and launch-failure events carry the originating job ID.
- Scheduler accounting uses the explicit event ID when sticky context contains a different job ID.
- Existing lifecycle publishers without explicit event data remain compatible.
- Concurrent multi-job runs do not retain stale scheduled-job entries or collapse admission because of misattributed lifecycle events.

pcnudde:
## Summary

- publish the originating job ID as non-sticky data on server job lifecycle events
- make `DefaultJobScheduler` prefer the explicit event job ID while preserving the existing sticky-ID fallback
- add regression coverage for sticky context changes and start/completion/abort publication

## Problem

Concurrent job launch and completion processing can replace the sticky `CURRENT_JOB_ID` before a lifecycle event handler reads it. The scheduler can then account a start, completion, or abort against the wrong job, leaving stale entries in `scheduled_jobs`. Once those entries reach `max_jobs`, valid admissions can be blocked even though execution capacity is available.

## Validation

- `72 passed` in the focused scheduler and JobRunner unit-test suites
- Black, isort, and Flake8 passed for all four changed files
- `git diff --check` passed
- a 100-job, one-server/two-client validation with `max_jobs=20` completed all jobs successfully, with each job scheduled exactly once and no stale-accounting serial tail

Fixes #5215

### DIFF
--- a/nvflare/apis/utils/job_utils.py
+++ b/nvflare/apis/utils/job_utils.py
@@ -18,7 +18,8 @@ import os
 from typing import Optional
 from zipfile import ZipFile
 
-from nvflare.apis.fl_constant import JobConstants
+from nvflare.apis.fl_constant import FLContextKey, JobConstants
+from nvflare.apis.fl_context import FLContext
 from nvflare.apis.job_def import ALL_SITES, JobMetaKey
 from nvflare.fuel.utils.config import ConfigFormat
 from nvflare.fuel.utils.config_factory import ConfigFactory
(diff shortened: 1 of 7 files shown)


YuanTingHsieh (approved):

YuanTingHsieh:
Follow-up audit (non-blocking for this PR): the explicit lifecycle event job ID and workspace fixes here correctly address the reported race. A separate cleanup should remove or convert the five server-parent sticky CURRENT_JOB_ID writes as one coordinated change, while retaining sticky job identity inside single-job child processes. It should also audit the default-sticky JOB_RUN_NUMBER alias and the legacy ambient-ID fallback in _save_workspace. JobMetricsCollector needs related concurrency hardening: use the explicit lifecycle job ID, key duration state by job ID, and avoid mutating a shared tags dictionary. Concurrent-context regression coverage should verify that one job operation cannot leak identity or timing state into another; the lifecycle helper fallback can remain temporarily for compatibility with external publishers.

PR #5226, opened 2026-08-28
[Research] Add AgenticFL workflow as a research example

ZiyueXu77:
### Description

This PR adds AgenticFL as a self-contained NVFlare research example. The workflow uses a Codex agent backend to negotiate the task, assess privacy-safe client metadata, generate client-local data adapters and task-specific training code, validate generated artifacts, and launch FedAvg through the NVFlare Client API.

The example includes:

- fixed `job_data.py` and `job_train.py` entry points for the showcased workflow;
- structured agent, data-contract, NVFlare runtime, prompt, logging, and utility modules;
- fail-closed client admission and generated-code validation with no deterministic backend fallback;
- a loopback local VLM for existing raw-image inspection, spatial-label visual review, rendering, and automatic orientation repair;
- no acquisition-quality review;
- reference preparation from available cohort data rather than distributed or synthetic reference images;
- a 39-site retinal cohort guide and dataset-source documentation;
- focused tests for data boundaries, job wiring, research scope, reference preparation, and visual-QC contracts.

The recorded end-to-end experiment queried 39 retinal sites, admitted 18 sites to client-local extraction, retained 10 sites and 26,549 records for training, and completed 100 FedAvg rounds with 10/10 client updates in every round. The README reports the latest client-mean metrics and training curve. It also explains expected run-to-run variation from agent-generated model and augmentation strategies while the principal FL settings remain fixed.

### Validation

- `PYTHONPATH=research/agentic-fl python -m unittest discover -s research/agentic-fl/tests -v` — 25 tests passed.
- The complete 39-site data phase and 10-client, 100-round GPU training workflow completed end to end.
- Every client produced 100 finite metric records and the final global model was persisted.

### Types of changes

- [x] Documentation updated.


research/agentic-fl/agenticfl/agents/local_adapter.py
```diff
@@ -0,0 +1,1723 @@
...
+        return filename if filename in {"adapter.py", "adapter_manifest.json"} else "[redacted-filename]"
+
+    sanitized = re.sub(
+        r"(?<![A-Za-z0-9_.-])[A-Za-z0-9_.-]+\.(?:png|jpe?g|tiff?|bmp|gif|csv|tsv|txt|json|xml|mat|npy|npz|nii(?:\.gz)?|dcm|mha|mhd|nrrd|nhdr|h5|hdf5|svs|zip|tar|gz)(?![A-Za-z0-9_.-])",
+        redact_filename,
+        sanitized,
+        flags=re.IGNORECASE,
+    )
+    sanitized = re.sub(
+        r"(?<![A-Za-z0-9_.-])[A-Za-z][A-Za-z0-9-]*(?:[_-][A-Za-z0-9-]+)*[_-]\d{3,}[A-Za-z0-9_-]*(?![A-Za-z0-9_.-])",
```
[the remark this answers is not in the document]
ZiyueXu77: Fixed in 6a22e875. The ambiguous nested identifier regex is replaced by a linear token match plus a simple numeric-segment check. A regression test exercises a 20,000-hyphen identifier and verifies fail-safe redaction.

research/agentic-fl/agenticfl/data/contracts/classification.py
```diff
@@ -0,0 +1,319 @@
...
+
+
+def label_value(value: Any) -> int | None:
+    if isinstance(value, bool):
+        return int(value)
+    if isinstance(value, int) and value >= 0:
+        return value
+    if isinstance(value, str) and re.fullmatch(r"\d+", value.strip()):
+        return int(value.strip())
+    return None
```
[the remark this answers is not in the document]
ZiyueXu77: Fixed in 6a22e875. Boolean labels now fail closed as invalid instead of being coerced to 0 or 1, with regression coverage for both boolean values and the supported integer forms.

research/agentic-fl/agenticfl/flare/agenticfl_training_client_runtime.py
```diff
@@ -0,0 +1,444 @@
...
+        return False
+    if decision.get("reviewed") is False or decision.get("consensus_reached") is False:
+        return False
+    selected_transform = decision.get("selected_transform")
+    if not isinstance(selected_transform, str):
+        return False
+    expected_transform = "as_is"
+    if isinstance(label_orientation, dict) and isinstance(label_orientation.get("selected_transform"), str):
+        expected_transform = label_orientation["selected_transform"]
+    return selected_transform == expected_transform
```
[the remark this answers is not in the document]
ZiyueXu77: Fixed in 6a22e875. The client training launcher now reuses the shared visual_qc_decision_passed predicate, eliminating the divergent implementation and enforcing the canonical transform allow-list. The focused test also covers a matching but unknown transform.

research/agentic-fl/scripts/prepare_references.py
```diff
@@ -0,0 +1,371 @@
...
+        rgb.save(destination, format="PNG")
+    return size
+
+
+def _write_mask_png(source: Path, destination: Path, *, expected_size: tuple[int, int]) -> None:
+    with Image.open(source) as image:
+        mask = image.convert("L")
+        if mask.size != expected_size:
+            raise ValueError(f"reference mask size {mask.size} does not match image size {expected_size}")
+        mask.save(destination, format="PNG")
```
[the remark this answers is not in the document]
ZiyueXu77: Fixed in 6a22e875. Reference masks are now required to be binary, include background value 0 and nonzero foreground, and are normalized to values 0 and 255. Tests cover normalization and multi-class rejection.

research/agentic-fl/README.md
```diff
@@ -0,0 +1,377 @@
...
+- general image enhancement, reconstruction, inpainting, or content repair;
+- OpenHands, hosted chat APIs, or deterministic runtime fallbacks;
+- production isolation or clinical/model-quality claims.
+
+## Repository Layout
+
+```text
+research/agentic-fl/
+|-- README.md
+|-- findings.md              # Review issues, acceptance criteria, and status
```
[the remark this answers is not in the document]
ZiyueXu77: Fixed in 6a22e875. findings.md is intentionally not part of the public research contribution, so the stale repository-tree entry has been removed from the README.

research/agentic-fl/agenticfl/data/extractor.py
```diff
@@ -0,0 +1,3389 @@
...
+            raise ValueError("generated materializer did not include source_files")
+        for entry in source_files:
+            if not isinstance(entry, dict):
+                raise ValueError("generated materializer source file entry was not an object")
+            rel_path = entry.get("path")
+            content = entry.get("content")
+            expected_sha = entry.get("sha256")
+            if not isinstance(rel_path, str) or not rel_path.strip() or not isinstance(content, str):
+                raise ValueError("generated materializer source file missing path or content")
+            if isinstance(expected_sha, str) and expected_sha and payload_digest(content) != expected_sha:
```
holgerroth: The checksum here is effectively opt-in by the sender. `path` and `content` are mandatory — the line above raises when either is missing — but `sha256` is not: when the key is absent, `isinstance(expected_sha, str)` is `False` and the comparison is skipped entirely. The content is then written at L1365 and executed at L1397 via `subprocess.run([sys.executable, entry_path, ...])` inside the client boundary, with the raw dataset in reach.

`source_digest` from `_pack_generated_data_materializer` (`agenticfl/server.py:1536`) isn't verified client-side either — it only reaches the logs — so there is no path on which the client confirms that what it executes is what the server intended to send.

Suggest making `sha256` mandatory per entry and failing closed when it is absent, matching how `path`/`content` are already handled:

```python
if not isinstance(expected_sha, str) or not expected_sha:
    raise ValueError(f"generated materializer source file missing sha256 for {rel_path}")
if payload_digest(content) != expected_sha:
    raise ValueError(f"generated materializer source checksum mismatch for {rel_path}")
```

research/fedready/fedready/agents/__init__.py
```diff
@@ -0,0 +1,3811 @@
...
+        "entry_script",
+        "source_digest",
+    ):
+        value = materializer.get(key)
+        if value is not None:
+            summary[key] = value
+    source_files = materializer.get("source_files")
+    if isinstance(source_files, list):
+        summary["source_file_count"] = len(source_files)
+        summary["source_files_redacted"] = True
```
holgerroth, line 2589: This drops `source_files` before the guardrail review, so the guardrail that gates server-to-client code execution never sees the code it is admitting.

Path: `agenticfl/client.py:410` → `authorize_extraction` → `GuardrailAgent.inspect` → `_guardrail_review_payload` → `_guardrail_redact_value`, which substitutes `_generated_materializer_agent_summary` on the `agenticfl.generated_data_materializer.v1` node. The guardrail then reviews `{entry_script, source_digest, source_file_count, source_files_redacted: true}`, returns `allowed: true`, and `_run_generated_data_materializer` writes and executes the source it never read.

Summarising large nodes to keep the review payload manageable is reasonable for most of the tree, but it is inverted for the one node whose contents are executable — that is the node the guardrail most needs to read.

Taken together with the optional `sha256` in `data/extractor.py` (separate comment), the remaining gate on executing server-supplied Python on a client is a file count.

holgerroth (commented):
Two findings from a review pass focused on the server-to-client trust boundary. Both sit on the generated-data-materializer path and compound each other, so I am flagging them together rather than separately.

These are the two highest-severity items from a longer list — happy to share the rest if useful.

research/agentic-fl/agenticfl/data/extractor.py
ZiyueXu77: Good catch — fixed in `0380fbbc`. The client now requires a non-empty `sha256` for every source file and verifies it before writing any source. It also requires and recomputes `source_file_count` and the bundle `source_digest` over `entry_script` plus the ordered path/digest manifest. Source files are only written after all validation succeeds. `test_generated_materializer_rejects_unbound_source_before_execution` covers missing and mismatched file digests, missing and mismatched bundle digests, and count mismatch, and verifies that the materializer is neither written nor executed.

research/fedready/fedready/agents/__init__.py
ZiyueXu77: Agreed — fixed in `0380fbbc`. `_guardrail_redact_value` now preserves the complete `agenticfl.generated_data_materializer.v1` node for client guardrail review, including every `source_files[].content`; redaction behavior for the rest of the payload is unchanged. The guardrail prompt now explicitly requires reviewing each source body as untrusted code and denying behavior outside the declared local interface or privacy contract. `test_client_guardrail_receives_exact_generated_materializer_source` captures the review context and verifies that the exact executable source reaches the guardrail.

holgerroth (commented):
Three findings from checking training validation and runtime argument handling at 9f3f55ae. All 30 included unit tests pass; focused local probes reproduced the issues below. The live Codex/VLM/GPU workflow was not run.

research/fedready/fedready/job_train.py
```diff
@@ -0,0 +1,2771 @@
...
+    if errors:
+        message = "; ".join(str(error.get("message")) for error in errors)
+        if any(str(error.get("kind", "")).endswith("not_found") for error in errors):
+            raise FileNotFoundError(message)
+        raise ValueError(message)
+    validated = dict(spec)
+    package_dir = _resolve_training_package_dir(str(spec["package_dir"]), code_workspace=code_workspace)
+    validated["package_dir"] = str(package_dir.resolve())
+    if require_local_simulation:
+        validated["package_integrity"] = build_training_package_integrity(package_dir)
```
holgerroth: [P2] Compare the package digest with the artifact that passed preflight

This recomputes and overwrites `package_integrity` while retaining the existing `local_simulation` approval. Consequently, when a saved training spec is reused after its source files change, export accepts code that never passed that preflight. In a local probe, replacing `train.py` with `raise RuntimeError("modified after preflight")` still passed `validate_training_code_spec`, and the recorded digest was silently replaced. Please compare the current digest against the digest recorded when preflight completed and require another preflight if it differs. Recording the newly computed digest alone does not bind the simulation evidence to the exported code.

research/fedready/fedready/job_train.py
```diff
@@ -0,0 +1,2771 @@
...
+        if path == log_txt:
+            continue
+        with path.open("r", encoding="utf-8", errors="replace") as handle:
+            for line in handle:
+                record_message(line)
+
+    model_paths = sorted(str(path) for path in server_dir.rglob("FL_global_model.pt"))
+    metric_paths = sorted(path for path in root.rglob(DEFAULT_METRICS_JSONL))
+    metric_artifacts = [str(path) for path in metric_paths]
+    nonempty_metric_artifacts = [str(path) for path in metric_paths if path.is_file() and path.stat().st_size > 0]
```
holgerroth, line 2341: [P2] Validate metric contents before treating them as success evidence

A nonempty file is sufficient for `metric_artifact_available`, which feeds both `succeeded` and the local-preflight promotion gate. With the other completion markers present, a local probe returned `succeeded=True` for a metrics file containing either `not JSON` or `{"round": 0, "loss": NaN}`. Generated training code can therefore pass preflight without producing valid, finite metric records. Please parse and validate the records before accepting the artifact; malformed or nonfinite metrics should fail the gate rather than qualify solely by file size.

research/fedready/fedready/job_train.py
```diff
@@ -0,0 +1,2771 @@
...
+
+def _format_task_args(
+    *,
+    template: str,
+    client_id: str,
+    config: FedAvgTrainingConfig,
+    dataset_root: str | None = None,
+) -> str:
+    values = {
+        "dataset_root": shlex.quote(dataset_root or config.dataset_root),
```
holgerroth: [P2] Preserve space-containing paths through TaskScriptRunner argument parsing

`shlex.quote` produces shell syntax, but this job uses the in-process `TaskScriptRunner`, whose `get_sys_argv()` calls `split_command_preserving_secret_refs(..., posix=False)` and preserves legacy whitespace splitting. For example, formatting `/tmp/my project/mock_data` here produces argv entries `"'/tmp/my"` and `"project/mock_data'"` instead of one dataset-root argument. Preflight derives an absolute mock-data path from the project workspace, so a checkout or project path containing spaces cannot start training. Please pass the path through a configuration/argument mechanism compatible with this runner, and cover the actual runner-to-launcher parsing in a regression test.

research/fedready/fedready/job_train.py
ZiyueXu77: Valid finding, fixed in `74cf5687`. `validate_training_code_spec` no longer overwrites the recorded package integrity. The required-simulation path now recomputes the deterministic package manifest and requires exact equality with the manifest recorded after preflight; any changed source requires another preflight. `test_training_package_must_match_local_preflight_digest` covers the reported reuse case.

research/fedready/fedready/job_train.py
ZiyueXu77: Valid finding, fixed in `74cf5687`. Simulator status now parses every nonblank JSONL record, requires an object with at least one numeric aggregate metric, and rejects non-finite round or metric values. Only validated artifacts feed `metric_artifact_available` and `succeeded`. `test_simulator_status_requires_valid_finite_metric_records` covers malformed JSON, `NaN`, and a valid finite record.

research/fedready/fedready/job_train.py
ZiyueXu77: Valid finding, fixed in `74cf5687`. The dataset root is now percent-encoded into a whitespace-free task-argument token and decoded by the client runtime launcher before local path resolution and forwarding to the generated trainer. The regression test runs the formatted arguments through the actual `TaskScriptRunner.get_sys_argv()` and launcher parsing with a space-containing preflight path.

research/fedready/fedready/job_train.py
```diff
@@ -0,0 +1,2828 @@
...
+        "file_count": len(files),
+        "files": files,
+    }
+
+
+def _training_package_integrity_errors(spec: dict[str, Any], *, package_dir: Path) -> list[dict[str, Any]]:
+    """Bind local preflight approval to the exact package being exported."""
+
+    recorded = spec.get("package_integrity")
+    current = build_training_package_integrity(package_dir)
```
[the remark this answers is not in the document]
ZiyueXu77: Valid finding, fixed in `7f840afe`. For Python packages, `_add_training_package` now exports `package_dir` into `custom/<package_dir.name>` instead of exporting `package_dir.parent`. This preserves the package import root while excluding mutable sibling modules, so the tree covered by `package_integrity` is exactly the tree deployed to the server and clients. `test_training_package_export_uses_integrity_root` protects that contract.

PR #5230, opened 2026-08-28
Fix StatisticsController._validate_min_clients returning True when no clients responded

cch0:
### Description

`_validate_min_clients` is supposed to check that at least `min_clients`
sites actually contributed real statistics. But its checks are written
inside `for` loops that go over the results — so if there's nothing to
loop over, the checks never run at all, and the function falls through
to `return True` by default. It should return `False` in that case, but
instead it says everything is fine.

This can happen in more than one place in the same function: when zero
clients responded at all, and separately, when clients responded but
none of them actually produced any per-dataset statistics. Both cases
mean the job effectively got no usable data, and both were silently
treated as success.

This PR adds an explicit check for each of those empty cases, returning
`False` with a warning log instead of falling through. Nothing else
changes — cases where enough real data was collected still work exactly
the same as before.

### Types of changes

  full parallel suite run hit unrelated pre-existing environment/infra noise (missing optional extras + an
 xdist worker crash), but the specific test file for this change passed cleanly

- [ ] Documentation updated.


nvflare/app_common/workflows/statistics_controller.py
```diff
@@ -367,12 +373,19 @@ def _validate_min_clients(self, min_clients: int, client_statistics: dict) -> bo
                     if ds_feature_statistics[ds_name]:
                         resulting_clients[ds_name].update([client])
 
+        if not resulting_clients:
+            self.logger.warning(
```
chesterxgchen: since we are aborting, this should error log, not warning.

nvflare/app_common/workflows/statistics_controller.py
```diff
@@ -350,6 +350,12 @@ def results_cb(self, client_task: ClientTask, fl_ctx: FLContext):
     def _validate_min_clients(self, min_clients: int, client_statistics: dict) -> bool:
         self.logger.info("Check if min_client result received for all features")
 
+        if not client_statistics:
+            self.logger.warning(
```
chesterxgchen: logger.error() not warning, since we are aborting the job

chesterxgchen (commented):
please remove or complete the Fixes # . placeholder.
add few minor comment, but look good for me

nvflare/app_common/workflows/statistics_controller.py
cch0: Updated.

nvflare/app_common/workflows/statistics_controller.py
cch0: Updated.

cch0:
> please remove or complete the Fixes # . placeholder. add few minor comment, but look good for me

Thank you for the feedback, Changes have been made. PR description has been updated, too.

chesterxgchen (approved):

PR #5236, opened 2026-08-31
Correct Slurm internal connection security in design doc

pcnudde:
### Description

Correct the Slurm job launcher design document to match the current internal connection behavior:

- worker-to-parent TCP connections use mTLS by default
- `job_launcher.internal_connection_security: clear` is an explicit insecure opt-out for trusted, isolated networks

Runtime behavior is unchanged and existing tests already verify the mTLS default.

### Types of changes

- [x] Documentation updated.

### Validation

- `VIRTUAL_ENV="$PWD/.venv" PATH="$PWD/.venv/bin:$PATH" UV_CACHE_DIR=/private/tmp/uv-cache-flare3160 ./runtest.sh -s --skip-install`
- `git diff --check`


IsaacYangSLA (approved):
LGTM

PR #5240, opened 2026-09-01
Add NVIDIA FLARE implementation of FedSCS aggregation

Rakib-Ul-Haque:
# FedSCS: Robust Federated Learning via Stable Cosine Similarity

🏆 Award: This work received the Distinguished Conference Paper Award at the IEEE ICCST 2025.

FedSCS is a robust federated aggregation method that assigns adaptive weights to client updates using **peer-update similarity** and **temporal stability**. It is intended for federated learning settings with heterogeneous or potentially unreliable client updates.

## Motivation

Standard FedAvg primarily weights clients according to local dataset size. When client data are non-IID or a client update is substantially different from the updates of other clients, dataset-size weighting alone may assign substantial influence to an update that is inconsistent with the prevailing update direction.

FedSCS addresses this by comparing each client update with the aggregate direction of its peers and tracking the client's similarity across communication rounds. Updates that are consistently aligned with their peers receive higher aggregation weights.

FedSCS does **not** require access to clients' raw training data and should not be interpreted as providing guaranteed malicious-client detection.

## FedSCS

For client \(i\) at round \(t\), FedSCS:

1. Computes the peer-consensus update from the other participating clients.
2. Computes the non-negative cosine similarity between the client update and peer consensus.
3. Maintains a rolling similarity score across rounds.
4. Penalizes rapidly changing similarity through a stability term.
5. Normalizes the resulting scores into aggregation weights.
6. Aggregates client updates using these weights.

The method requires \(O(Nd)\) operations for \(N\) participating clients and update dimension \(d\), without constructing a pairwise client-similarity matrix.

## NVIDIA FLARE Integration

This example implements FedSCS as a **custom NVIDIA FLARE model aggregator** and uses the standard `FedAvgRecipe` for the federated workflow.

```text
FedAvgRecipe
     │
     ├── Client 1 ──┐
     ├── Client 2 ──┤
     ├── Client 3 ──┤
     ├── Client 4 ──┤── DIFF updates
     └── Client 5 ──┘
                    │
             FedSCSAggregator
                    │
          Peer similarity + stability
                    │
             Adaptive weights
                    │
               Global model
```

Model updates are transferred using NVIDIA FLARE's `DIFF` transfer type.

## Example Dataset

The example uses CIFAR-10 with five simulated clients. The dataset is prepared locally and is not included in the repository.

The dataset preparation creates four regular clients and one intentionally noisy client (`site-5`). For `site-5`, 80% of the training samples are corrupted using salt-and-pepper noise, with a salt probability of 0.30 and a pepper probability of 0.30. The standard clean CIFAR-10 test set is used for evaluation.

This setup provides a controlled example of heterogeneous client updates and allows FedSCS to evaluate client updates based on their similarity to the updates of the other participating clients.

## Update Safety

FedSCS uses cosine similarity, whose trust score is invariant to positive scalar rescaling of a client update. To provide defense in depth against arbitrarily large finite DIFF values, this implementation applies a configurable L2 norm bound to received client updates before they are used by the aggregator. The default example uses a maximum update norm of 10.0.

This magnitude bound is an implementation-level safety control and is separate from the Stable Cosine Similarity scoring formulation described in the FedSCS paper.

## Project Structure

```text
research/fedscs/
├── README.md
├── requirements.txt
├── job.py
├── client.py
├── prepare_data.sh
└── src/
    ├── fedscs_aggregator.py
    └── model.py
```

## Requirements

* Python 3.10+
* NVIDIA FLARE ~= 2.9.0
* PyTorch
* torchvision
* NumPy

Install the dependencies with:

```bash
pip install -r research/fedscs/requirements.txt
```

## Prepare CIFAR-10

From the NVIDIA FLARE repository root:

```bash
./research/fedscs/prepare_data.sh
```

The preparation script downloads CIFAR-10 and validates the required dataset files before reporting successful preparation.

## Run the Example

From the NVIDIA FLARE repository root:

```bash
cd research/fedscs
python job.py
```

The example runs a simulated federated learning experiment using the standard NVIDIA FLARE Recipe workflow and the custom `FedSCSAggregator`.

During training, the FedSCS aggregator reports the per-round client scores,
Stable Cosine Similarity (SCS) values, and normalized aggregation weights.
In simulator runs, these per-client `score`, `scs`, and `weight` INFO messages
are written to `<workspace>/server/log.txt`. Inspect these values across
rounds to see how FedSCS adapts the contribution of each client.

In particular, compare the aggregation weight assigned to `site-5` with the weights of the other clients. Because `site-5` contains intentionally noisy training data, its FedSCS weight reflects its similarity and stability relative to the other participating clients.

## Limitations

FedSCS relies on peer consensus as a reference for evaluating client updates. The example uses five clients, with four regular clients and one noisy client, and provides a controlled CIFAR-10 demonstration rather than covering all forms of non-IID data, client heterogeneity, data corruption, or adversarial behavior.

## Citation

If you use FedSCS in academic work, please cite:

```text
Rakib Ul Haque and Panagiotis (Panos P.) Markopoulos,
"Robust Federated Learning via Stable Cosine Similarity,"
IEEE ICCST, 2025.
```

## License

This research example follows the licensing terms of the NVIDIA FLARE repository.


research/fedscs/data/cifar-10-batches-py/batches.meta
ZiyueXu77, line 1: Please replace data files with a "prepare_data.sh" script for this step. Example here "https://github.com/NVIDIA/NVFlare/blob/main/examples/advanced/cifar10/pt/cifar10-sim/prepare_data.sh"

Rakib-Ul-Haque: Dear @ZiyueXu77 

Thank you for your valuable comments. The file is updated accordingly.

research/fedscs/src/fedscs.py
```diff
@@ -0,0 +1,175 @@
...
+It is framework-independent and can be used by the NVFlare
+research example.
+"""
+
+from typing import Dict, List, Tuple
+
+import numpy as np
+
+
+def flatten_update(update: Dict[str, np.ndarray]) -> np.ndarray:
```
ZiyueXu77, line 14: Is this module intended to be used? A repository-wide search found no imports of `flatten_update`, `compute_fedscs_weights`, or `aggregate_models`; `FedSCSAggregator` independently reimplements these operations, and its temporal-score behavior differs from this module. Please either have the aggregator reuse this framework-independent implementation (with focused tests), or remove this file so there is one authoritative implementation of the algorithm.

research/fedscs/job.py
```diff
@@ -0,0 +1,98 @@
...
+
+    aggregator = FedSCSAggregator(
+        eps=1e-12,
+        max_update_norm=10.0,
+    )
+
+    recipe = FedSCSRecipe(
+        name="fedscs_cifar10",
+        model=model,
+        min_clients=2,
```
ZiyueXu77, line 60: Could we note the two-client limitation in the README, or use at least three clients if this example is meant to demonstrate robust discrimination? With exactly two participants, each client's peer sum is just the other update, so cosine symmetry gives both clients the same raw score and, from equal history, 0.5 weights. Two clients are sufficient as an integration smoke test, but they cannot demonstrate identification of an outlier from this signal alone.

research/fedscs/prepare_data.sh
```diff
@@ -0,0 +1,30 @@
...
+
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+DATA_DIR="${SCRIPT_DIR}/data"
+DATASET_DIR="${DATA_DIR}/cifar-10-batches-py"
+ARCHIVE="${DATA_DIR}/cifar-10-python.tar.gz"
+URL="https://www.cs.toronto.edu/~kriz/cifar-10-python.tar.gz"
+
+mkdir -p "${DATA_DIR}"
+
+if [ -d "${DATASET_DIR}" ] && [ -f "${ARCHIVE}" ]; then
```
ZiyueXu77, line 13: This existence check can accept an incomplete or corrupt preparation. I interrupted the download locally; extraction left a partial dataset directory. After resuming the archive, this branch printed `CIFAR-10 dataset already prepared`, while `train.py` failed with `Dataset not found or corrupted`. Please download to a temporary file with `curl --fail`, verify the published checksum, extract into a temporary directory, and only return success after validating the required batch files.

research/fedscs/job.py
```diff
@@ -0,0 +1,98 @@
+import sys
+from pathlib import Path
+from typing import Any, Dict
```
ZiyueXu77, line 3: The required repository checks currently fail for this example. The license checker rejects all seven new Python files; Black would reformat six files; isort rejects five; and flake8 reports unused imports here and in `fedscs.py`/`fedscs_aggregator.py`. `git diff --check` also flags the extra blank line at the end of the README. Please run the repository formatting/import cleanup and add the canonical license headers before merge.

ZiyueXu77 (commented):
thanks for the PR! Some comments

chesterxgchen (commented):
FedSCS: Robust Federated Learning via Stable Cosine Similarity
FedSCS is a robust federated aggregation method that uses client update similarity and temporal stability scores to compute adaptive aggregation weights.


Please add some background information, the title only describe the name and what it does, but did not explain why do we need this ? in what cases, we should use this algorithm ? what kind of the problem it is trying to solve.

research/fedscs/src/fedscs.py
```diff
@@ -0,0 +1,175 @@
+"""
```
holgerroth, line 1: This module appears unused: `job.py`, `fedscs_controller.py` and `fedscs_aggregator.py` never import it, and the aggregator re-implements the flatten/cosine/weight logic. The two also disagree on the algorithm: `compute_fedscs_weights` uses a running average divided by `len(previous_scores) + 1` (number of clients, not rounds), while the aggregator uses `stability = 1 - |cur - prev| / max(...)`. Since the README points readers here as the algorithm, I would either drop this file or have the aggregator import from it so there is one definition.

research/fedscs/client.py
```diff
@@ -0,0 +1,168 @@
+import logging
```
holgerroth, line 1: Missing the Apache-2.0 license header. Same for `job.py`, `train.py`, `src/fedscs.py`, `src/model.py` and `src/fedscs_controller.py`. The repo license check (`./runtest.sh -l`) covers `research/`, so CI will fail on these.

research/fedscs/src/fedscs_aggregator.py
```diff
@@ -0,0 +1,927 @@
...
+
+    w_global_next = w_global + sum_i(alpha_i * clipped_delta_i)
+
+where clipped_delta_i is the client update after applying the
+configured L2 update-norm bound.
+"""
+
+from __future__ import annotations
+
+from typing import Any, Dict, List, Optional, Tuple
```
holgerroth, line 42: `List`, `Tuple` and `Shareable` are imported but never used (flake8 F401).

holgerroth (commented):
Low-effort single-pass review; three inline notes below.

holgerroth:
Thanks for the contribution. One structural suggestion that would simplify this a lot and also fix the server-side memory profile.

**Suggestion: drop the custom `FedSCS` controller and `FedSCSRecipe`, and implement FedSCS as a custom aggregator only, using DIFF transfer.**

Right now `fedscs_controller.py` copies the whole `FedAvg.run()` loop just to add `aggregator.reset_stats()` + `aggregator.load_model(model)` so the aggregator can subtract the global model. That freezes a copy of FedAvg internals (`_aggr_helper`, `_received_count`, `_maybe_cleanup_memory`, the disk-offload plumbing, ...) which will drift from upstream, and it needs a `FedAvgRecipe` subclass in `job.py` on top.

You don't need the global model on the aggregator side at all. The stock recipe already supports sending deltas:

```python
from nvflare.client.config import TransferType

recipe = FedAvgRecipe(
    ...,
    aggregator=FedSCSAggregator(...),
    params_transfer_type=TransferType.DIFF,
)
```

With that, the Client API computes `delta_i = w_i - w_global` on the client and `accept_model()` receives the delta directly. The aggregator then clips, scores cosine similarity against the peer sum, and returns the weighted delta:

```python
return FLModel(params=aggregated_delta, params_type=ParamsType.DIFF, metrics=...)
```

`FedAvg.update_model()` -> `FLModelUtils.update_model()` adds a DIFF result onto the server's global model, which is exactly `w_global + sum_i(alpha_i * clipped_delta_i)` from your docstring. Stock `FedAvg` + stock `FedAvgRecipe`, no controller, no recipe subclass. This is also the pattern used by the other custom-aggregator examples (`examples/advanced/cifar10/pt/cifar10-sim/cifar10_custom_aggr`, `examples/advanced/medgemma`).

**Memory.** This matters beyond tidiness. The current aggregator keeps a float64 copy of the global model (`load_model`), a float64 copy of every client model (`client_models`), and a second float64 copy of every delta (`client_updates`), and rebuilds the N-1 peer sum from scratch for each client. For float32 weights that is roughly 4N+2 model-sizes at peak, and the `np.asarray` calls also materialize any tensor-disk-offload refs FedAvg passes through. FedSCS inherently needs all N deltas before it can weight anything, so O(N) is the floor, but it can be close to that floor: store each incoming delta once (flattened, original dtype, clipped on arrival), keep one running total, and compute the peer sum for client i as `total - delta_i`. That gets you to about N+1 flattened models with no extra copy of the global model.

Happy to look at a revised version. If it helps, I can sketch the accept/aggregate skeleton for the DIFF-based aggregator.

Rakib-Ul-Haque:
Thanks for the detailed review. I’ve addressed the requested changes:

* Removed the unused duplicate `src/fedscs.py` implementation.
* Removed the custom FedSCS controller/recipe and implemented FedSCS as a custom `ModelAggregator` using the standard `FedAvgRecipe`.
* Switched the example to `TransferType.DIFF`.
* Increased the simulation to five clients and documented the two-client limitation.
* Hardened CIFAR-10 preparation with temporary download/extraction, checksum verification, and dataset validation.
* Updated the README with the motivation, workflow, integration details, and limitations of FedSCS.
* Removed update clipping and corrected the temporal stability calculation to follow the published FedSCS formulation.
* Added the required license headers and cleaned up formatting and unused imports.
* Verified the example with a complete 10-round NVIDIA FLARE simulation.
* Confirmed that the targeted repository checks pass, including Black, isort, flake8, agent-skill-lint, license checks, and `git diff --check`.

The changes have been pushed to the `fedscs` branch. Thanks again for the helpful feedback!

research/fedscs/src/fedscs_aggregator.py
```diff
@@ -0,0 +1,524 @@
...
+            current_scores[client_name] = float(score)
+            current_scs[client_name] = float(scs)
+
+        total_scs = sum(current_scs.values())
+
+        if not np.isfinite(total_scs):
+            raise RuntimeError("FedSCS produced a non-finite total trust score.")
+
+        if total_scs <= self.eps:
+            raise RuntimeError("FedSCS produced zero total trust score; " "unable to compute aggregation weights.")
```
holgerroth: **Blocking: this aborts the whole job.** A `RuntimeError` here propagates out of `aggregate_model()`, FedAvg reports `FATAL_SYSTEM_ERROR`, and the run stops.

It is also not a corner case. Whenever the client updates lack a dominant common direction, the total sum `S` is small relative to each individual update, so for every client `g_i = S - d_i ≈ -d_i` and `cos(d_i, g_i) ≈ -1`. All `rho_i` are then exactly 0 at the same time. I reproduced this in a 5-client SimEnv run (synthetic data with no shared signal): every client scored `0.000000` and round 0 ended with

```
ERROR - Exception in model controller run: RuntimeError: FedSCS produced zero total trust score; unable to compute aggregation weights.
ERROR - Aborting current RUN due to FATAL_SYSTEM_ERROR received
```

Early rounds with heterogeneous data or small batches can hit this in practice. The previous revision fell back to uniform weights, which is the right behaviour; please restore that (with a `self.warning(...)` so it is visible in the log) instead of raising. Your earlier `RuntimeError` for the non-finite case can stay.

research/fedscs/src/fedscs_aggregator.py
```diff
@@ -0,0 +1,524 @@
...
+        """Return the FedSCS weighted DIFF."""
+        if not self.client_updates:
+            raise RuntimeError("No client updates received.")
+
+        self._compute_scs()
+
+        client_names = list(self.client_updates)
+
+        # Construct the output strictly from the authoritative schema.
+        aggregated_delta = {name: np.zeros(shape, dtype=np.float64) for name, shape in self.expected_schema.items()}
```
holgerroth: **Blocking: the global model silently becomes float64.** `aggregated_delta` is returned as float64 regardless of the incoming dtype. `FLModelUtils.update_model()` does `global + diff`, and NumPy/torch promote to float64, so after round 0 every server-side parameter is float64. Server model memory doubles and the persisted checkpoints are float64 too. I confirmed this in the simulator: `FL_global_model.pt` and `best_FL_global_model.pt` hold `torch.float64` weights for a float32 `SimpleCNN`.

Accumulating in float64 is fine; just cast back before returning, e.g. record the incoming dtype per parameter in `accept_model` (or take it from the server model alongside `expected_schema`) and do

```python
aggregated_delta[name] = aggregated_delta[name].astype(input_dtype[name], copy=False)
```

before building the `FLModel`.

research/fedscs/client.py
```diff
@@ -0,0 +1,423 @@
...
+        raise ValueError("Training data loader is empty. " "Check the dataset preparation.")
+
+    return total_loss / total_batches
+
+
+def main(args):
+    """Run the NVFLARE client."""
+    script_dir = os.path.dirname(os.path.abspath(__file__))
+
+    data_dir = os.path.join(
```
holgerroth: Portability note (non-blocking for a research example, but worth a line in the README or a flag): the data directory is resolved relative to `__file__`. This works in `SimEnv` only because the in-process runner executes the original script path. When I export the job, `app/custom/` contains just `client.py` and `src/`, so with `launch_external_process=True` or any POC/production deployment the script runs from `custom/` and `load_client_dataset` raises `FileNotFoundError`.

A `--data_dir` argument (defaulting to the current location) would make it work in both modes, matching how the other CIFAR-10 examples handle data paths.

research/fedscs/requirements.txt
```diff
@@ -0,0 +1,10 @@
+# NVIDIA FLARE
+nvflare==2.9.0rc2
```
holgerroth: 2.9.0 was tagged on 2026-08-31, so this pins a superseded release candidate. Other `research/` examples use `~=`; `nvflare~=2.9.0` would be consistent. `torchaudio` is unused and can be dropped.

research/fedscs/README.md
```diff
@@ -0,0 +1,124 @@
...
+├── client.py
+├── prepare_data.sh
+└── src/
+    ├── fedscs_aggregator.py
+    └── model.py
+```
+
+## Requirements
+
+* Python 3.9+
```
holgerroth: The repo requires Python 3.10+ (`setup.cfg` `python_requires`), and the FLARE version should match the requirements pin (2.9.0 rather than rc2).

research/fedscs/README.md
```diff
@@ -0,0 +1,124 @@
...
+               Global model
+```
+
+Model updates are transferred using NVIDIA FLARE's `DIFF` transfer type.
+
+## Example Dataset
+
+The example uses CIFAR-10 with five simulated clients. The dataset is prepared locally and is not included in the repository.
+
+The example can be configured to study heterogeneous client updates, including a client with corrupted/noisy training data, while the standard CIFAR-10 test set is used for evaluation.
```
holgerroth: This says the noisy client *can be configured*, but `prepare_data.sh` unconditionally makes client 5 noisy (80% of its samples, salt/pepper 0.30/0.30). Since that is the whole point of the demonstration, please state it explicitly and tell the reader what to look for: `site-5` should receive a visibly lower weight in the per-round FedSCS printout. A short sample of that output would help a lot.

Also, the Limitations section says the example uses "at least three clients" while `job.py` hardcodes five.

research/fedscs/src/fedscs_aggregator.py
```diff
@@ -0,0 +1,524 @@
...
+
+        self.current_scores = current_scores
+        self.current_scs = current_scs
+
+        self.current_weights = {client_name: scs / total_scs for client_name, scs in current_scs.items()}
+
+        if not all(np.isfinite(weight) for weight in self.current_weights.values()):
+            raise RuntimeError("FedSCS produced non-finite aggregation weights.")
+
+        print("")
```
holgerroth: Nit: `ModelAggregator` has `self.info()`, which goes to the FLARE log with timestamps; `print` is fine in the simulator but is lost in real deployments. Also `round_number` starts at 1 while FedAvg starts at 0, so the log shows `FedSCS Round 1` inside `Round 0 started.` Using `self.fl_ctx.get_prop(AppConstants.CURRENT_ROUND)` (FedAvg sets it on the aggregator's context each round) would keep them aligned.

holgerroth (changes requested):
Thanks for the restructuring. The aggregator-only DIFF design is exactly what I had in mind, and the schema/norm hardening in the latest commit looks good.

I re-reviewed head b7f0aaa by exporting the job, unit-testing the aggregator, and running 5-client SimEnv simulations against the local main checkout. Two things below are blocking (a hard job abort and a dtype regression), the rest are smaller. Details inline.

research/fedscs/src/fedscs_aggregator.py
Rakib-Ul-Haque: Thank you for catching this. I have fixed this case by falling back to uniform weights when the total SCS is zero. The aggregator now also emits a warning instead of raising an exception, so the round can continue safely.

The non-finite total SCS case is still treated as an error, since that indicates an invalid numerical state rather than a valid zero-score fallback.

Fixed in commit 97d4b8317.

research/fedscs/src/fedscs_aggregator.py
Rakib-Ul-Haque: Thank you. I have addressed this by keeping the aggregation accumulation in float64 for numerical stability, while tracking the incoming dtype of each parameter and casting the aggregated DIFF back to that dtype before returning it.

Therefore, for the default PyTorch float32 model, the aggregated update remains float32 and does not change the global model dtype.

Fixed in commit 97d4b8317.

research/fedscs/client.py
Rakib-Ul-Haque: Thank you for pointing this out. I have updated the client to accept --data_dir explicitly and pass the data directory from job.py through train_args.

This also fixes the SimEnv execution path where the client process may otherwise resolve the data directory relative to the copied/exported client script.

I agree that a deployment-independent data path would be preferable for a fully portable deployment. For this example, the explicit --data_dir argument keeps the dataset location configurable while ensuring the provided SimEnv example works correctly.

research/fedscs/requirements.txt
Rakib-Ul-Haque: Thank you. I updated requirements.txt to use nvflare~=2.9.0 and removed the unused torchaudio dependency.

The remaining dependencies are limited to NVIDIA FLARE, PyTorch, torchvision, and NumPy.

research/fedscs/README.md
Rakib-Ul-Haque: Updated the README to explicitly document Python 3.10+ as a requirement and aligned the dependency information with the updated requirements.txt.

research/fedscs/README.md
Rakib-Ul-Haque: Thank you. I expanded the README to explicitly document the site-5 setup: 80% of its training samples are intentionally corrupted using salt-and-pepper noise with salt probability 0.30 and pepper probability 0.30, while the test set remains clean.

I also added instructions for inspecting the per-round FedSCS scores/weights so that the effect of the noisy client can be observed directly.

research/fedscs/src/fedscs_aggregator.py
Rakib-Ul-Haque: Correct. I updated the README to consistently describe the example as using five simulated clients and clarified that the five-client setup provides the peer group used by the FedSCS consensus calculation.

holgerroth (commented):
Thanks for the quick turnaround. I re-verified head 6c766ce the same way as before (job export, 5-client SimEnv run on synthetic no-consensus data, black/isort/flake8 with the repo config):

- Zero-consensus round now logs `WARNING - FedSCS produced zero total trust score; falling back to uniform aggregation weights.` and the job runs to `Finished FedAvg.` instead of aborting.
- `FL_global_model.pt` / `best_FL_global_model.pt` are back to `torch.float32`; `expected_dtypes` serializes cleanly into `config_fed_server.json`.
- `--data_dir` is passed through `train_args` and resolved as an absolute path on the client.
- Round numbering now follows `AppConstants.CURRENT_ROUND`, output goes through `self.info`/`self.warning`.
- Requirements/README versions, torchaudio, noisy-client description, and the Limitations text are all updated. Style and license checks pass on all four Python files.

One small optional note: in the simulator the per-client `score=/scs=/weight=` INFO lines land in `<workspace>/server/log.txt` rather than on the console, so a one-line pointer in the README's "Run the Example" section would save readers a hunt.

From my side the review comments are resolved.

holgerroth (commented):
Re-reviewed 6c766ce. Two remaining issues are detailed inline: clipping can invalidate float32 updates on NumPy 2, and best-model selection uses local rather than global accuracy. Job export succeeded, and five focused checks covered both reproductions plus the zero-trust fallback, unclipped output dtype, and score history. Full CIFAR-10 training was not run.

research/fedscs/src/fedscs_aggregator.py
```diff
@@ -0,0 +1,651 @@
...
+        for name, value in update.items():
+            array = np.asarray(value)
+
+            try:
+                with np.errstate(
+                    over="raise",
+                    invalid="raise",
+                    under="ignore",
+                ):
+                    bounded_value = array * scale
```
holgerroth, line 251: **[P1] Preserve the authoritative dtype when clipping updates.**

`_flatten_update()` produces float64, so `np.linalg.norm()` and consequently `scale` are NumPy float64 scalars. On NumPy 2, `array * scale` promotes a float32 update to float64. The update is stored successfully, but `_compute_scs()` subsequently calls `_validate_dtypes()` and raises, aborting round aggregation whenever clipping activates.

I reproduced this on NumPy 2.5.3 with three float32 updates `[20, 20]`, an authoritative float32 schema, and the default norm bound of 10. All stored updates became float64, and `aggregate_model()` raised `ValueError: Parameter 'w' for client 'site-1' has dtype float64; expected float32.` Unclipped updates passed.

Please cast each bounded value back to its authoritative parameter dtype before storing it, retaining the finite-value check, and add a regression test that exercises clipping with float32 inputs under NumPy 2. Casting only the final aggregate is too late for this path.

research/fedscs/client.py
```diff
@@ -0,0 +1,430 @@
...
+        # ---------------------------------------------------------------
+
+        output_model = flare.FLModel(
+            params=model_diff,
+            params_type=ParamsType.DIFF,
+            metrics={
+                "accuracy": local_accuracy,
+                "loss": local_loss,
+                "global_accuracy": global_accuracy,
+                "global_loss": global_loss,
```
holgerroth, line 373: **[P2] Select global checkpoints using the received global model's accuracy.**

`job.py` leaves `FedAvgRecipe.key_metric` at its default, `"accuracy"`, but that entry here contains the locally trained model's accuracy. NVFlare serializes metrics accompanying parameters as `INITIAL_METRICS`, and `IntimeModelSelector` uses the selected metric before aggregation to decide whether to save the existing global model. The metric and saved model therefore refer to different weights.

A focused check with the actual selector reproduced selection of a worse global model: when global accuracy decreased from 0.8 to 0.6 while local accuracy increased from 0.3 to 0.9, the second global model was marked as the new best.

Please set `key_metric="global_accuracy"` in `job.py`, or report `global_accuracy` under `"accuracy"` and give the post-training metric a separate local name. The client already computes the required global metric.

holgerroth:
Thank you for addressing all of the previously raised issues, @Rakib-Ul-Haque. I re-verified the latest head (uniform-weight fallback, float32 checkpoints, `--data_dir`, logging, requirements and README) and everything checks out on my side. We will do a final pass before approving.

Rakib-Ul-Haque:
Thanks for the thorough review. I addressed the remaining issues and pushed the updates.

Preserved the authoritative dtype when clipping updates, including the NumPy 2 / float32 case.
Added a regression test for float32 norm clipping; the test suite passes (2 passed).
Set FedAvgRecipe.key_metric="global_accuracy" for best-model selection.
Added the README pointer to <workspace>/server/log.txt for the per-round FedSCS score, scs, and weight messages.
Re-ran the 5-client CIFAR-10 FedSCS simulation through all 10 rounds successfully, ending with Finished FedAvg.
Black, isort, flake8, and git diff --check all pass.

The updated commit has been pushed to the fedscs branch. Thanks again for the detailed feedback.

holgerroth:
@Rakib-Ul-Haque I ran the FedSCS example and a standard `FedAvgRecipe` baseline on H100 at commit 5544548fab04d4c4673dbfc744bb69a62cd9607a. Both completed all 10 rounds successfully, and the saved checkpoints loaded with finite float32 parameters.

The runs used the same prepared CIFAR-10 datasets, `SimpleCNN`, and unchanged client training code: 5 clients, 10 rounds, 4 local epochs, batch size 64, SGD with learning rate 0.01 and momentum 0.9, DIFF transfer, and `global_accuracy` checkpoint selection. Four clients had 10,000 clean samples each; the fifth had 10,000 samples with 80% corrupted. I reused the exact prepared data files for the baseline and independently evaluated both runs' saved checkpoints on the same full, clean 10,000-image test set.

| Checkpoint | Standard FedAvg | FedSCS |
| --- | ---: | ---: |
| Final global model | 67.57% | 69.17% |
| Selected best global model | 67.39% | 69.10% |

The final-model difference is **+1.60 percentage points** for FedSCS. This is encouraging, but these were single runs with independent random initialization and training order, so the difference does not yet establish a reproducible advantage. Best-model selection scores incoming global models; the final post-aggregation model was evaluated separately, which explains why the final checkpoint can score higher.

Could you add a reproducible FedAvg comparison to this example, ideally with repeated, matched seeds? A few implementation suggestions:

- Add a method option such as `--method fedavg|fedscs`. The baseline can reuse the existing `FedAvgRecipe` configuration and client script with `aggregator=None`; keep the other training and checkpoint-selection settings identical.
- Run, for example, **5 seeds** (at least 3 would already help), with both methods paired within each seed. Reuse identical client partitions and corruption realizations within each pair, and initialize both methods from the same saved initial checkpoint via `initial_ckpt`. Seed the actual client training processes and data-loader RNGs as well as Python/NumPy/PyTorch; setting a seed only in the job-building process will not cover simulator-side initialization or client shuffling. Record data and training seeds explicitly.
- Include the commands and a small README results table with every seed, mean ± standard deviation, and the paired FedSCS-minus-FedAvg differences. Use the final model after the fixed 10 rounds as the primary comparison; report selected-best results separately.
- One useful additional ablation would be **FedAvg with the same update-norm clipping**. The current FedSCS implementation clips at 10.0, while standard FedAvg does not, so the comparison above combines the effects of scoring and clipping. A clipping-matched baseline would help distinguish those effects.

A baseline runner plus the repeated-seed results would make the contribution substantially easier to evaluate and reproduce.

Rakib-Ul-Haque:
<p dir="auto" data-start="0" data-end="150" class="PDq2pG_selectionAnchorContainer">Thank you for running the initial comparison and for the detailed feedback. I have now implemented the requested reproducibility and ablation changes.<span aria-hidden="true" class="PDq2pG_selectionAnchor"></span></p>
<p dir="auto" data-start="152" data-end="174">Specifically, I added:</p>
<ul data-start="176" data-end="891">
<li data-section-id="1xywf39" data-start="176" data-end="298">
A <code data-start="180" data-end="219">--method fedavg|fedavg_clipped|fedscs</code> option, with standard FedAvg using the native <code data-start="266" data-end="280">FedAvgRecipe</code> aggregation path.
</li>
<li data-section-id="17oazlc" data-start="299" data-end="450">
Five matched seeds (<code data-start="321" data-end="332">1001–1005</code>) using the same prepared client data/corruption setup and the same seed-specific initial checkpoints for each method.
</li>
<li data-section-id="16d7xbg" data-start="451" data-end="538">
Explicit client-side seeding for Python, NumPy, PyTorch/CUDA, and DataLoader workers.
</li>
<li data-section-id="1xyvfqx" data-start="539" data-end="630">
A clipping-matched FedAvg baseline using the same L2 update-norm bound of 10.0 as FedSCS.
</li>
<li data-section-id="1qc4g4h" data-start="631" data-end="731">
Separate evaluation of the final model after the fixed 10 rounds and the selected-best checkpoint.
</li>
<li data-section-id="1i069ea" data-start="732" data-end="816">
An evaluation script and initial-checkpoint generation script for reproducibility.
</li>
<li data-section-id="kyf22k" data-start="817" data-end="891">
The commands, methodology, and complete five-seed results in the README.
</li>
</ul>
<p dir="auto" data-start="893" data-end="917">The updated results are:</p>
<div dir="auto" class="group TyagGW_tableContainer"><div tabindex="-1" class="TyagGW_tableWrapper flex flex-col-reverse w-fit">
Metric | FedAvg | FedAvg + L2 clipping | FedSCS + L2 clipping
-- | -- | -- | --
Final, mean ± std | 66.19 ± 0.97% | 66.23 ± 1.27% | 68.88 ± 0.72%
Best, mean ± std | 67.13 ± 0.80% | 67.01 ± 1.21% | 68.56 ± 0.62%

</div></div>
<p dir="auto" data-start="1162" data-end="1224">For the paired final-model comparison, the mean difference is:</p>
<ul data-start="1226" data-end="1343">
<li data-section-id="9wz87f" data-start="1226" data-end="1257">
FedSCS − FedAvg: <strong data-start="1245" data-end="1257">+2.69 pp</strong>
</li>
<li data-section-id="vjjove" data-start="1258" data-end="1300">
FedSCS − FedAvg + clipping: <strong data-start="1288" data-end="1300">+2.65 pp</strong>
</li>
<li data-section-id="1jxfjwc" data-start="1301" data-end="1343">
FedAvg + clipping − FedAvg: <strong data-start="1331" data-end="1343">+0.04 pp</strong>
</li>
</ul>
<p dir="auto" data-start="1345" data-end="1374">For the selected-best models:</p>
<ul data-start="1376" data-end="1493">
<li data-section-id="s4codj" data-start="1376" data-end="1407">
FedSCS − FedAvg: <strong data-start="1395" data-end="1407">+1.44 pp</strong>
</li>
<li data-section-id="11xsl9m" data-start="1408" data-end="1450">
FedSCS − FedAvg + clipping: <strong data-start="1438" data-end="1450">+1.55 pp</strong>
</li>
<li data-section-id="1lou7pd" data-start="1451" data-end="1493">
FedAvg + clipping − FedAvg: <strong data-start="1481" data-end="1493">−0.11 pp</strong>
</li>
</ul>
<p dir="auto" data-start="1495" data-end="1643">The changes and results have been pushed to the <code data-start="1543" data-end="1551">fedscs</code> branch in commit <a data-start="1569" data-end="1642" class="decorated-link cursor-pointer" rel="noopener" target="_new"><code data-start="1570" data-end="1581">523f34432</code></a>.</p>
<p dir="auto" data-start="1645" data-end="1826" data-is-last-node="" data-is-only-node="">Thank you again for the suggestion. The matched-seed comparison and clipping ablation should make the experimental comparison substantially more reproducible and easier to evaluate.</p>

Rakib-Ul-Haque:
The final changes are included in commit [68a86b3]

Issue #337, opened 2022-03-22
Logging Interface/API


Nintorac:
We will need to tie NVFlare into our logging infrastructure at some point. When that happens we will need for NVFlare to emit logs in JSON format in order for ingestion to be simple. I would propose adding a Logging interface that then lets users implement their own backends.

The interface should look something like this
```
class ILogger():

    @abstractmethod
    def _log(self, level, message, **kwargs):
        pass

    def info(self, message, **kwargs):
        self._log("info", message, **kwargs)

    def warn(self, message, **kwargs):
        self._log("warn", message, **kwargs)

    def error(self, message, **kwargs):
        self._log("error", message, **kwargs)

    def critical(self, message, **kwargs):
        self._log("critical", message, **kwargs)
```

Dagster does this fairly well, can look to their for inspiration - https://docs.dagster.io/concepts/logging/loggers

Nintorac:
One https://docs.dagster.io/concepts/logging/loggers#logging-from-an-op

yanchengnv:
Yup, this aligns well with our direction of spec-based programming model. We'll add this item for future versions.

tal-rhino:
NVFlare seems to currently use the Python stdlib logging module. That supplies a similar interface to that described here, as well as extensive and well-documented ways to customize log formatting, where to write it, filtering and more.

I can't think of a good reason that NVFlare should implement its own way of logging. So I'd suggest:
1. Make the NVFlare code stick to the convention of using logger names prefixed with their module or part of the system, always starting with `nvflare.`. This currently doesn't seem to be the case.
2. Document how NVFlare's internal logging is done and examples of how to customize it.

With this, switching NVFlare to structured JSON logging would be a relatively simple matter of configuring logging to use an appropriate formatter.

taleinat:
NVFlare seems to currently use the Python stdlib logging module. That supplies a similar interface to that described here, as well as extensive and well-documented ways to customize log formatting, routing, filtering and more.

I can't think of a good reason that NVFlare should implement its own way of logging. So I'd suggest:
1. Make the NVFlare code stick to the convention of using logger names prefixed with their module or part of the system, always starting with `nvflare.`. This currently doesn't seem to be the case.
2. Document how NVFlare's internal logging is done and examples of how to customize it.

With this, switching NVFlare and one's own model code to use structured JSON logging would be a straightforward matter of configuring logging to use an appropriate formatter, using the existing knowledge and tools for the logging module.

[...]

PR #5244, opened 2026-09-02
Make Hello PyTorch quickstart meaningful by default

chesterxgchen:
Hello PyTorch now runs a reproducible, offline learning task with `python job.py`: two clients, three CPU training rounds, and evaluation of the persisted final global model on both sites. The synthetic labels correspond to visible image features, so completion demonstrates learning as well as successful execution.

- Keep the beginner script focused on four controls: clients, rounds, dataset, and client-local data root. CIFAR-10 remains an explicit option, prepared once before clients start. Missing cache files fail before simulation with the exact preparation command; exporting a job does not require local data.
- Align the README, documentation, website snippets, notebooks, and CI with this default. The acceptance test loads the final model and requires at least 60% accuracy on both sites and a 40-point improvement over the initial model.
- Give the CLI notebook its own small exporter for the shared PyTorch application. Both export cells work with the simplified beginner API and retain client log streaming. The abort example uses 8,192 training samples per client over 20 rounds.
- Share one helper for isolating the tests' generic `model` and `prepare_data` imports. Only tests requiring torchvision skip when it is unavailable or incompatible; the synthetic tests still run.
- Restore real full cross-site evaluation coverage in premerge CI, checking that both clients submit their models and evaluate every client model and the persisted global model. The beginner CLI remains focused on final-global evaluation.

The branch includes merged #5270. Framework prerequisites are already merged; this PR changes examples, documentation, tests, website content, and CI. The advanced POC/production continuation in #5255 is stacked on this PR and must merge afterward.

Validation:

- 24 focused Hello PyTorch unit tests passed, including real notebook exports, CIFAR export without local data, and early rejection of missing data.
- Both real simulator acceptance tests passed: the learned, loadable final model and the complete client/server cross-site evaluation matrix.
- Isolated runs with torchvision missing or failing on import each produced 7 passed and 2 skipped client tests.
- Repository style, changed Markdown links, and `git diff --check` passed.
- HTML documentation built on the combined stack using `SKIP_API_DOCS=1 sphinx-build -b html`; existing repository diagnostics remain, with none in changed files.


YuanTingHsieh:
Thanks for making the default path deterministic, meaningful, and tested. I do have a scope and usability concern for a hello-world example.

The learner-facing job.py grows from roughly 96 to 217 lines and exposes about 18 arguments. A newcomer now encounters dataset selection, tracking, cross-site evaluation, external-process execution, memory GC, export behavior, POC lifecycle management, production startup kits, usernames, and failure cleanup before they have absorbed the basic FedAvg path.

I suggest keeping Hello PyTorch focused on the meaningful deterministic dataset, final-global-model evaluation, and a small zero-argument simulation, with only a few instructional knobs such as clients, rounds, and dataset. The simulation-to-POC-to-production progression, tracking, external-process controls, and memory tuning could live in a separate continuation or advanced example, with this README linking to it.

prepare_data.py itself seems reasonable because it keeps the nontrivial synthetic-data construction out of the training loop. The larger concern is the amount of orchestration and CLI surface in the beginner entry point, not the supporting test and documentation coverage. A hello example should be complete enough to work, but it does not need to serve as the comprehensive reference application.

YuanTingHsieh:
Two implementation findings from review:

1. [P1] Monitoring failures bypass the new cleanup path and can produce a successful process exit.

In job.py, the try/except around run.get_result() does not catch most monitoring failures. Run.get_result() intentionally catches exceptions raised by ExecEnv.get_job_result(), caches a None result, and returns normally. For POC, it also calls stop(clean_up=False). Consequently, a monitoring failure can print Result can be found in : None, exit with status 0, and retain the failed POC workspace instead of reaching the clean_up=True exception handler. Please check the returned result and cached status after get_result(), then explicitly clean up and raise when monitoring did not complete successfully. A regression test should exercise a real Run whose environment raises from get_job_result(), rather than a recipe.execute() stub that raises directly.

2. [P2] The optional uncached CIFAR-10 path can still race during its initial download.

Every client uses the same DATASET_PATH (/tmp/nvflare/data) and calls torchvision.datasets.CIFAR10(..., download=True). When valid CIFAR-10 data has already been pre-staged there, the example can run offline and no download is needed. When the data is absent, however, simulated clients may concurrently download and extract the same files into that shared directory. The README statement that each client receives its own local copy is therefore also misleading. Please pre-download once before starting clients, synchronize the initial download, use distinct client roots, or restore a clear warning for the uncached path.

holgerroth (commented):
I reviewed this PR in depth at `903461f` (all changed files plus the consumers of the example: the notebook, the Sphinx pages, the website snippets, the standalone-job configs, and the sibling examples) and re-checked every point against the current head `c94c3f2` before posting. Locally the unit tests pass and the simulator acceptance test passes in about 25 s, with identical metrics and bit-identical global weights across two runs.

Resolved by the newer commits: the two Sphinx pages now match the example, thanks. Still open, as inline comments below: the seed dependence of the 60% learning gate, the requirements pin, the notebook and TensorBoard regressions, the always-on final-evaluation branch, a few README accuracy points, and some test-coverage gaps. I have not re-reviewed the new POC/production environment code beyond confirming it does not change these findings; Yuan-Ting's comments cover that area.

tests/integration_test/fast/hello_pt_quickstart_test.py
```diff
@@ -0,0 +1,131 @@
...
+    first_round_metrics = {metric["name"]: metric["value"] for metric in first_round["aggregated_metrics"]}
+    initial_accuracy = first_round_metrics["accuracy"]
+
+    with open(os.path.join(server_run_dir, "cross_site_val", "cross_val_results.json")) as results_file:
+        final_results = json.load(results_file)
+    final_accuracies = [site_results["SRV_FL_global_model.pt"]["accuracy"] for site_results in final_results.values()]
+
+    assert set(final_results) == {"site-1", "site-2"}
+    assert initial_accuracy <= 20.0
+    assert min(final_accuracies) >= 60.0
```
holgerroth, line 87: **The 60% gate holds for this particular `MODEL_SEED`, not for the training setup.**

I replicated the zero-argument job outside the simulator with the same `prepare_data`/`model` modules, loaders and seeds, persistent SGD momentum across rounds, and 7-step equal-weight FedAvg. It reproduces the recorded run exactly (round 0: 1.0/1.0, round 1: 30.0/30.0, final 70.0/70.0). Sweeping only the model-init seed with everything else fixed:

- seeds 0-19: `[40, 30, 10, 61, 60, 20, 90, 50, 30, 40, 60, 50, 50, 60, 20, 70, 40, 40, 54, 20]`, so 6 of 20 pass
- seeds 0-99: 29 of 100 pass, and 12 of those land exactly on 60.0

Thread count and matmul precision do not move the result, so platform noise is not the risk. What flips it: a torch release that changes init RNG consumption (premerge installs unpinned torch; inserting extra RNG draws after `manual_seed` gives 4 of 10 passes), any edit to `MODEL_SEED`, `batch_size` 64 (10%), `train_size` 100 (10%), or `lr` 0.05 (20%). `batch_size=16` and `epochs=2` land exactly on 60.0.

Since this is a required premerge step, I would widen the margin (3 rounds gives 75/77 here) or lower the bar, and note in the test that the threshold is calibrated against the seeded init, rather than keep a cushion of 10 of 100 eval images.

examples/hello-world/hello-pt/requirements.txt
```diff
@@ -1,4 +1 @@
-nvflare[PT]~=2.7.2rc
-torch
-torchvision
-tensorboard
+nvflare[PT]~=2.10.0
```
holgerroth: **The pin is off on two counts.**

The merge base of this PR is tag `2.9.0rc3` and nothing under `nvflare/` changes here, so every API the example uses (`add_final_global_evaluation`, `add_cross_site_evaluation`, `client_memory_gc_rounds`, `initial_ckpt`, the reserved `--export` handling, `round_metrics.jsonl`) already exists at 2.9.0rc3. Per AGENTS.md the pin should name the first upcoming version that supports the feature, which is 2.9.0.

Independently, `~=2.10.0` desugars to `>=2.10.0, ==2.10.*`, and `2.10.0rc1 < 2.10.0`, so no 2.10 pre-release can ever satisfy it, not even with `--pre`. The README's first command then fails through the whole 2.9 cycle and every 2.10.0rcN, while the CI gate in this same PR shows the example works on 2.9.0rc3. Every other current-line pin under `examples/` carries the `rc` suffix for this reason.

```suggestion
nvflare[PT]~=2.9.0rc
```

examples/hello-world/hello-pt/client.py
```diff
@@ -47,62 +52,87 @@ def evaluate(net, data_loader, device):
...
+        train_size=args.train_size,
+        test_size=args.test_size,
+        batch_size=args.batch_size,
+        num_workers=args.num_workers,
+    )
     last_params = None
 
     # (optional) metrics tracking
-    summary_writer = SummaryWriter()
+    summary_writer = SummaryWriter() if args.track_metrics else None
```
holgerroth: **Gating the client writer on `--track_metrics` breaks TensorBoard for every existing caller.**

Only `job.py` emits `--track_metrics` (when `--experiment_tracking` is set). `hello-pt.ipynb` cell 21 passes `train_args=f"--batch_size {batch_size}"` and cell 24 still adds the TensorBoard receiver, so `tb_events/` is created but stays empty, and cell 29's `%tensorboard` shows "No dashboards are active" while cell 28 promises per-client loss curves. The website's `jobCode_pt` snippet has the same shape. Nothing on the server side writes to TensorBoard on its own.

The client-side `SummaryWriter` does not need the tensorboard package, so it can stay unconditional as before; the receiver is the optional part.

examples/hello-world/hello-pt/README.md
```diff
@@ -1,227 +1,302 @@
...
-* **Aggregation**: Models aggregated and persisted on the server.
-
-#### Round 1
-* **Clients Sampled**: site-1, site-2.
-* **Training**:
-  * Similar process as Round 0.
-  * **Aggregation**: Models aggregated and persisted.
-#### Completion
-* **FedAvg Process**: Successfully finished with the final model persisted.
+For an interactive CIFAR-10 and TensorBoard-oriented variant, see [`hello-pt.ipynb`](hello-pt.ipynb). The canonical
```
holgerroth, line 236: **This sentence is not true of the notebook after this PR, and the notebook itself is now broken.**

`hello-pt.ipynb` is untouched, but: (1) its `%pip install -r requirements.txt` no longer installs tensorboard, so cell 24 `add_experiment_tracking(recipe, tracking_type="tensorboard")` raises `ValueError: tensorboard is not installed` in a fresh venv (Colab hides this because tensorboard is preinstalled); (2) it passes no `--dataset`, so it now trains on the synthetic data while cells 8/11/15/18 describe CIFAR-10 and cell 19 shows the old unseeded `SimpleNetwork`; (3) with the new defaults its `stop_cond="accuracy >= 70"` trips after round 1, so the advertised 5-round run ends after 2. `examples/hello-world/README.md` (lines 103 and 108) and `hello_world.ipynb` likewise still advertise CIFAR-10 and a `tb_events` path the default run never creates.

Either update the notebook (install tensorboard, pass `--dataset cifar10` and `--track_metrics`, fix the prose) or describe it accurately here.

examples/hello-world/hello-pt/job.py
```diff
@@ -49,47 +108,109 @@ def define_parser():
...
+
+    if args.experiment_tracking != "none":
+        add_experiment_tracking(recipe, tracking_type=args.experiment_tracking)
 
     if args.cross_site_eval:
+        # Full CSE also collects and evaluates each client's latest local model.
         add_cross_site_evaluation(recipe)
+    else:
+        # Always verify the persisted final global model in the basic quickstart.
+        add_final_global_evaluation(recipe)
```
holgerroth: **There is no longer a way to get a plain FedAvg job out of this example.**

The base added CSE only when `--cross_site_eval` was passed; now the `else` branch always adds `add_final_global_evaluation`, and no option turns it off. Two consequences:

- `--train_script` pointing at a user's own script that does not handle `flare.is_evaluate()` gets a mandatory `CrossSiteModelEval(submit_model_task_name="")` in the exported `config_fed_server.json`; the validate task makes the script train and return weights, and `ValidationJsonGenerator` logs "Expected dxo of kind METRICS or COLLECTION" and writes empty results with no error.
- `tests/integration_test/data/test_configs/standalone_job/pt_job.yml` ("with synthetic data", unchanged) now silently runs the CSE workflow while its only validator is `PTModelValidator`; `GlobalModelEvalValidator` (as used in `np_app.yml`) would actually check the new output.

A `--final_eval/--no-final_eval` flag, or a three-way eval mode, would restore the removed path. Minor: the export now carries two `ValidationJsonGenerator` components (BaseFedJob's plus the helper's). They write identical content, so it is harmless, but worth knowing.

examples/hello-world/hello-pt/README.md
```diff
@@ -1,227 +1,302 @@
...
+| Dataset | Deterministic synthetic images with a class-related signal |
+| Clients | 2 |
+| Federated rounds | 2 |
+| Local epochs per round | 1 |
+| Training examples per client | 200 |
+| Evaluation examples per client | 100 |
+| Batch size | 32 |
+| Data-loader workers | 0 |
+| Experiment tracking | Off |
+| Post-training evaluation | Final global model on both clients |
```
holgerroth, line 45: **The default run evaluates two server models, and the second scores lower than the "final" one.**

`IntimeModelSelector` skips round 0 and accepts round 1 against `-inf`, so `best_FL_global_model.pt` is always written, and with 2 rounds it is always the round-0 aggregate (clients score the received model before training, so the final aggregate is never scored). `cross_val_results.json` therefore shows, per site, `SRV_FL_global_model.pt: 70.0` next to `SRV_best_FL_global_model.pt: 30.0` in my runs. This row says "Final global model" only, and the paragraph under the outputs list tells readers to use `SRV_FL_global_model.pt` without saying why a model called "best" is worse.

A sentence on the one-round lag, or restricting the final-eval locator to the final model, would save newcomers the confusion.

examples/hello-world/hello-pt/README.md
```diff
@@ -1,227 +1,302 @@
-
 # Hello PyTorch
-This example demonstrates how to use NVIDIA FLARE with PyTorch to train an image classifier using federated averaging (FedAvg). The complete example code can be found in the `hello-pt directory <examples/hello-world/hello-pt/>`. It is recommended to create a virtual environment and run everything within a virtualenv.
 
-## NVIDIA FLARE Installation
-For the complete installation instructions, see [Installation](https://nvflare.readthedocs.io/en/main/installation.html)
-```
-pip install nvflare
+This quickstart trains a small image classifier with federated averaging (FedAvg). Two simulated clients train on
+distinct local datasets for two rounds, and then evaluate the persisted final global model on separate evaluation
```
holgerroth: **"Distinct local datasets" overstates what the synthetic default demonstrates.**

`SyntheticImageDataset` uses the same patch geometry and the same U(0, 0.08) noise law for every site; only the seed differs. Labels are `arange(size) % 10` permuted, so both sites get exactly 20 samples per class (10 in eval). The two sites are i.i.d. replicas of one distribution, which is more homogeneous than a random IID split of a real dataset; in the recorded run both sites report identical received-model accuracy every round and identical final numbers. The "Why this default is better" section lists "the simulated clients also used the same CIFAR-10 data" as a fixed problem, but the default does not demonstrate a partition either, and the CIFAR-10 paragraph below discloses that only for CIFAR-10.

"Independently generated" is accurate; or add a per-site class skew if heterogeneity is meant to be part of the lesson. (`docs/hello-world/hello-pt/index.rst:11` repeats the wording.)

examples/hello-world/hello-pt/README.md
```diff
@@ -1,227 +1,302 @@
...
-python job.py --export --export-dir /tmp/nvflare/jobs/job_config
+
+Use CIFAR-10 instead of the deterministic quickstart data:
+
+```bash
+python job.py --dataset cifar10 --epochs 2 --batch_size 16 --num_workers 2
 ```
 
-The exported job is written to `/tmp/nvflare/jobs/job_config/hello-pt`. You can combine the export flags with the example-specific arguments, for example:
+CIFAR-10 is downloaded on first use. In this simulated example, each client receives its own local copy of the same
```
holgerroth: +1 to Yuan-Ting's P2 above. Two additions: torchvision's `download_url` has no lock or atomic rename (it checks integrity, then writes straight to the final path), so on a cold cache two concurrent clients really do corrupt each other; and the repo convention for this is a runnable `prepare_data.py` that downloads once (hello-lightning `README.md:46-50`, hello-jax `README.md:103`; 26 of 27 `examples/**/prepare_data.py` files have a `__main__`). hello-pt now has a `prepare_data.py` with that name that does nothing when run (`python prepare_data.py` exits 0 silently), so a user carrying the sibling habit over gets no download and no hint. The base README's pre-download note was removed in this PR; restoring it, or making `prepare_data.py` do the download, would close both points.

examples/hello-world/hello-pt/job.py
```diff
@@ -49,47 +108,109 @@ def define_parser():
...
+        "--epochs",
+        str(args.epochs),
+        "--learning_rate",
+        str(learning_rate),
+        "--num_workers",
+        str(args.num_workers),
+        "--dataset",
+        args.dataset,
+    ]
+    if args.dataset == "synthetic":
```
holgerroth: **`--train_size`/`--test_size` are accepted but silently ignored with `--dataset cifar10`.**

They are forwarded only on the synthetic branch here, and the CIFAR-10 branch of `create_data_loaders` never reads them, so `python job.py --dataset cifar10 --train_size 1000 --test_size 200` parses cleanly and trains and evaluates on the full 50k/10k splits (now evaluated twice per round). Neither flag has `help=` text in either parser, and the README shows the two commands a few lines apart. Either `parser.error()` on the combination or honor the sizes with `torch.utils.data.Subset`.

tests/integration_test/slow/hello_pt_quickstart_test.py
```diff
@@ -0,0 +1,131 @@
...
+    final_accuracies = [site_results["SRV_FL_global_model.pt"]["accuracy"] for site_results in final_results.values()]
+
+    assert set(final_results) == {"site-1", "site-2"}
+    assert initial_accuracy <= 20.0
+    assert min(final_accuracies) >= 60.0
+    assert min(final_accuracies) >= initial_accuracy + 40.0
+
+    artifact_path = os.path.join(server_run_dir, "app_server", "FL_global_model.pt")
+    artifact = torch.load(artifact_path, map_location="cpu", weights_only=True)
+    assert artifact["model"]
```
holgerroth: **Two ways this gate can go green without testing anything.**

1. `assert artifact["model"]` only checks that the dict is non-empty; a checkpoint with mismatched keys (or `{"junk": tensor}`) passes, while the test name promises a *loadable* model. `job_module.create_model().load_state_dict(artifact["model"])` inside the `with` block is the one-line check.
2. The module-level `skipif(not HAS_PT)` (line 29) means a torch-less runner reports "1 skipped" and exits 0; `from nvflare.recipe import SimEnv` does not import torch, so collection succeeds, and the premerge step has no `-rs` or strict setting. Torch is installed today via `.[dev]`, so this is a silent-regression hazard rather than a live bug.

Placement: `tests/integration_test/README.md:73-74` reserves `slow/` for the nightly entry points and `fast/` for premerge (the adjacent premerge step uses `fast/`). `ci/run_integration.sh` runs the whole `slow/` directory nightly behind the xgboost wheel install, so this file will run twice. Moving it to `fast/` is a one-line change in the workflow.

tests/unit_test/examples/hello_pt_client_test.py
```diff
@@ -15,11 +15,12 @@
 import importlib.util
 import os
 import sys
+from collections import Counter
 
 import pytest
 
-HAS_PT_DEPS = all(importlib.util.find_spec(dep) is not None for dep in ("torch", "torchvision"))
-pytestmark = pytest.mark.skipif(not HAS_PT_DEPS, reason="PyTorch example dependencies are not installed")
+HAS_PT = importlib.util.find_spec("torch") is not None
```
holgerroth, line 22: **The CIFAR-10 branch of `create_data_loaders` now has zero coverage anywhere.**

No test calls `create_data_loaders`; `pt_job.yml`, `recipe_system_test.py`, and the new acceptance tests are all synthetic, and `hello_pt_job_test.py` only asserts that the string `cifar10` is forwarded. The base at least smoke-tested torchvision through the module-level import and this guard (`HAS_PT_DEPS` included torchvision, with the `RuntimeError` skip). A typo in `client.py:65-75` would now break only `python job.py --dataset cifar10`, unnoticed by CI. A monkeypatched `create_data_loaders(dataset="cifar10", ...)` call (pattern in `cifar10_fedprox_client_test.py:74`) would cover it.

examples/hello-world/hello-pt/client.py
```diff
@@ -47,62 +52,87 @@ def evaluate(net, data_loader, device):
 
         if total == 0:
             raise ValueError("Evaluation data_loader produced no samples; check data preparation and --test_size.")
-        accuracy = 100 * correct // total
-        print(f"Accuracy of the network on {total} test images: {accuracy} %")
+        accuracy = 100.0 * correct / total
```
holgerroth, line 55: **`evaluate()` never calls `net.eval()`, and nothing calls `net.train()` before the epoch loop.**

Harmless for `SimpleNetwork` (no BatchNorm/Dropout; the outputs are identical in both modes), but this PR makes the helper the example's canonical eval pattern by calling it twice per round and publishing both numbers as metrics. A user who swaps in a BatchNorm/Dropout model gets a pre-training accuracy computed in train mode, and BN running stats mutated by the eval pass before training; that number feeds `IntimeModelSelector`. Pre-existing, but since the function is touched here, an `eval()`/`train()` pair (as in `hello-dp/client.py:86,200`) is worth adding.

examples/hello-world/hello-pt/model.py
```diff
@@ -19,16 +19,23 @@
...
-        self.conv1 = nn.Conv2d(3, 6, 5)
-        self.pool = nn.MaxPool2d(2, 2)
-        self.conv2 = nn.Conv2d(6, 16, 5)
-        self.fc1 = nn.Linear(16 * 5 * 5, 120)
-        self.fc2 = nn.Linear(120, 84)
-        self.fc3 = nn.Linear(84, 10)
+        # The recipe serializes the model as a class configuration and the
+        # server constructs a new instance. Seed inside the constructor so
+        # those server-side parameters are reproducible too.
+        with torch.random.fork_rng(devices=[]):
```
holgerroth: **Two notes on seeding inside `__init__`.**

1. `fork_rng(devices=[])` restores only the CPU generator, but `torch.manual_seed` also seeds every CUDA/MPS/XPU generator, so constructing the model resets accelerator RNG for the rest of the process (server process included). `torch.random.default_generator.manual_seed(MODEL_SEED)` gives identical weights without the side effect.
2. The premise in the comment is right (the exported config is `{"path": "model.SimpleNetwork", "args": {}}`, so the server re-instantiates from the class path), but a module constant with no opt-out pins every `SimpleNetwork()` in every importer: `recipe_system_test.py`, the notebook, and anyone doing a seed sweep has to monkeypatch `MODEL_SEED`. `__init__(self, seed: Optional[int] = None)` with `self.seed = seed` is serialized by `FedJobConfig._get_args`; I verified that an export with `create_model()` returning `SimpleNetwork(seed=MODEL_SEED)` produces `{"args": {"seed": 202610}}` and identical weights, while a bare `SimpleNetwork()` keeps normal semantics. That would also give `create_model()` a purpose; today its docstring credits it for reproducibility that lives entirely in `__init__`.

examples/hello-world/hello-pt/client.py
```diff
@@ -12,24 +12,29 @@
...
 from torch.optim import SGD
-from torchvision.transforms import Compose, Normalize, ToTensor
 
 # (1) import nvflare client API
 import nvflare.client as flare
 from nvflare.client.tracking import SummaryWriter
 
 DATASET_PATH = "/tmp/nvflare/data"
+LOCAL_MODEL_PATH = "./local_model.pt"
+DEFAULT_BATCH_SIZE = 32
```
holgerroth, line 30: **These defaults, the dataset argument group, and the learning-rate fallback are byte-identical copies of the ones in `job.py`.**

The client copy is live: `recipe_system_test.py:176` and `hello-pt.ipynb` drive `client.py` without `--learning_rate`, while `python job.py` always forwards its own values. Tuning `DEFAULT_SYNTHETIC_LEARNING_RATE` in `job.py` (for example to fix the gate margin) would leave those callers on the old value with nothing failing (`hello_pt_job_test.py` asserts the job-side `0.1` only). `prepare_data.py` is already copied into the job's `custom/` directory, so a shared defaults module there would ship correctly; or `job.py` can import the defaults from `client.py`, since the config copier follows only the train script's imports.

chesterxgchen:
Thanks—agreed on both points. I have split the beginner quickstart from the environment-continuity work.

PR #5244 now contains only the meaningful simulation path: deterministic site-distinct data, final-global-model evaluation, a small zero-argument run, and its CI learning gate. `job.py` is now 91 lines (slightly smaller than the original 96-line version) and exposes only client count, round count, and dataset choice. POC/production execution, tracking, full cross-site evaluation, external-process controls, and memory tuning are preserved for a separate continuation PR after this one lands, so that PR will contain only the incremental environment work.

The CIFAR-10 documentation now states that simulated clients share `/tmp/nvflare/data`, requires a one-process pre-download before simulation, warns that concurrent initial downloads can race, and makes clear that all clients read the same logical CIFAR-10 datasets rather than receiving separate local copies.

The monitoring-failure cleanup finding applies only to the extracted POC/production continuation. I will address it there by treating a `None` result or unsuccessful cached status as failure, cleaning up only the POC lifecycle owned by that invocation, raising a nonzero error, and testing with a real `Run` whose environment raises from `get_job_result()`.

tests/integration_test/fast/hello_pt_quickstart_test.py
chesterxgchen: Good catch; I reproduced your 0–19 sweep exactly for the two-round setup. I changed the fixed default to three rounds while retaining the 60%/40-point functional thresholds. The real simulator run now produces 75% and 77% final accuracy (initial aggregate accuracy 1%) and passes in 25.63 seconds, so the absolute gate has 15 points of margin without materially changing quickstart time. The test, README, RST guide, and PR description now state explicitly that the thresholds are calibrated to the fixed model/data seeds and three-round defaults—not arbitrary initialization or hyperparameters. Addressed in 2e754328a.

examples/hello-world/hello-pt/README.md
chesterxgchen: Made `prepare_data.py` runnable. `python prepare_data.py` downloads both CIFAR-10 splits once, while every simulated client now opens the prepared cache with `download=False`, eliminating concurrent client writes. The README, RST guide, and notebook all use that command. Addressed and unit-tested in e484017b5.

examples/hello-world/hello-pt/README.md
chesterxgchen: Documented the one-round metric lag and told readers to use `SRV_FL_global_model.pt` for the post-training result. The explanation now makes clear why the earlier model selected as `best` can score below the final aggregate in a short run. Addressed in e484017b5.

examples/hello-world/hello-pt/client.py
chesterxgchen: Restored the client `SummaryWriter` unconditionally and removed `--track_metrics`; the server-side receiver remains the optional persistence control. The TensorBoard notebook now installs TensorBoard explicitly and selects the CIFAR-10 path explicitly. Addressed in e484017b5.

examples/hello-world/hello-pt/requirements.txt
chesterxgchen: The 2.10 target is intentional for this release deliverable, so I am not moving the example back to 2.9. You are right that the previous `~=2.10.0` excluded every prerelease. I replaced it with the explicit `nvflare[PT]>=2.10.0rc0,<2.11`, which admits 2.10 RCs as well as 2.10 final/patch releases and excludes 2.11. I also verified those cases with `packaging.specifiers.SpecifierSet`. Addressed in e484017b5.

examples/hello-world/hello-pt/job.py
chesterxgchen: For the beginner quickstart, final-global-model evaluation is now an intentional part of the acceptance contract rather than an optional mode. The simplified `job.py` no longer accepts arbitrary `--train_script` or evaluation-mode controls, so it cannot silently pair an incompatible script with that contract. The standalone recipe-system test now validates the final global-evaluation output. Plain FedAvg and full cross-site-evaluation controls belong in the advanced environment-continuity follow-up, where their compatibility can be documented and tested explicitly. Addressed in e484017b5.

examples/hello-world/hello-pt/client.py
chesterxgchen: Added `net.eval()` in `evaluate()` and an explicit `model.train()` before local training, with regression coverage verifying evaluation-mode execution. Addressed in e484017b5.

examples/hello-world/hello-pt/model.py
chesterxgchen: Implemented the suggested shape: `SimpleNetwork(seed=None)` preserves normal bare-constructor semantics, while `create_model()` passes the fixed recipe seed. The seeded branch seeds only PyTorch’s CPU default generator inside `fork_rng(devices=[])`, and the export test verifies that `{"seed": 202610}` is serialized in the server model configuration. Unit coverage also verifies reproducibility and restoration of the caller’s CPU RNG state. Addressed in e484017b5.

examples/hello-world/hello-pt/job.py
chesterxgchen: The beginner `job.py` no longer exposes `--train_size` or `--test_size`; its only instructional controls are client count, round count, and dataset choice. Therefore `python job.py --dataset cifar10 --train_size ...` is rejected instead of being silently accepted. Training-size tuning will be part of the advanced continuation rather than the first-run interface. Addressed in e484017b5.

examples/hello-world/hello-pt/README.md
chesterxgchen: Updated `hello-pt.ipynb`, the umbrella `hello_world.ipynb`, the hello-world index, and the website job snippet to match the new behavior. The focused notebook explicitly installs TensorBoard, prepares CIFAR-10 once, passes `--dataset cifar10`, uses the seeded model factory, and no longer uses the early-stop settings that shortened the run unexpectedly. Addressed in e484017b5.

tests/integration_test/slow/hello_pt_quickstart_test.py
chesterxgchen: Strengthened both cases: the test now instantiates the recipe model and calls `load_state_dict()` on the persisted artifact, and importing PyTorch is mandatory instead of producing a successful skip. The test was moved from `slow/` to `fast/`, with premerge updated to the new path, so it follows the repository’s integration-test placement contract and is not duplicated by the nightly slow suite. Addressed in e484017b5.

examples/hello-world/hello-pt/README.md
chesterxgchen: Agreed. Replaced the “distinct local datasets” claim with the precise statement that site and split data are generated independently from the same simple IID distribution, and explicitly state that the quickstart does not demonstrate statistical heterogeneity. Updated the Markdown, RST, and umbrella notebook in e484017b5.

tests/unit_test/examples/hello_pt_client_test.py
chesterxgchen: Added a focused CIFAR loader test that monkeypatches `torchvision.datasets.CIFAR10`, exercises both train and test branches, and asserts that clients use the prepared cache with `download=False`. The preparation command is separately tested to download both splits. Addressed in e484017b5.

examples/hello-world/hello-pt/client.py
chesterxgchen: The simplified beginner job no longer duplicates the client’s learning-rate, batch-size, epoch, worker, or sample-count defaults; those remain client implementation details. Dataset path/choices/default are now shared from `prepare_data.py`, which is already shipped with the client script. Addressed in e484017b5.

holgerroth:
Thanks for the fast turnaround on the split and on the review items. The trimmed shape (103-line `job.py` with four controls, `prepare_data.py` as the one-time CIFAR-10 download, the gate in `fast/` with three rounds) is what we were after, and it is green for me locally: unit tests, the acceptance test at 75%/77%, and style.

One remaining scope request. Two framework changes still ride in this example PR, and I would move them into their own small PR that #5244 and #5255 can depend on:

1. `nvflare/fuel/utils/secret_utils.py` and `nvflare/app_common/executors/task_script_runner.py`: switching `get_sys_argv()` to `group_shell_quotes=True` changes tokenization for every in-process client script, not just hello-pt. A quoted span such as `--pattern "a b"` used to reach the script as two literal tokens and now arrives as one. It looks like the right fix and it is tested, but it deserves its own title, description, and release note rather than living inside a quickstart PR.
2. `nvflare/recipe/sim_env.py`: `get_job_status()` now returns a `FINISHED:*` status instead of `None`, which changes `Run.get_status()` for all `SimEnv` users. Same argument; the acceptance test's status assertion can move with it.

Keeping #5244 example-only also makes the stacked #5255 easier to review, since its own `nvflare/job_config` and `script_runner` changes then stand out as the only framework deltas there. Happy to review the split PR quickly.

chesterxgchen:
Review #335 disposition:

- The `train_args` list is intentional and supplied by prerequisite PR #5257. PR #5244 is now based on `main` and contains only example/documentation/test/CI changes, so it must merge after #5257; the PR description states that dependency explicitly.
- Fixed the standalone integration configuration in `a932ba3eb`: final-global-model evaluation now uses `GlobalModelEvalValidator` instead of `CrossValResultValidator`, so it no longer expects client-local model artifacts.

`./runtest.sh -s` and `git diff --check` pass on the corrected example-only revision.

nvkevlu (dismissed):
With all the updates now, this should be good after #5257 is merged.

chesterxgchen:
Review #337 disposition:

- Removed the duplicate `Run.get_status()` assertion in `4b38eb3a0`. Synchronous `SimEnv` terminal-status behavior and its regression coverage belong to prerequisite PR #5257, as requested during the framework/example split. The quickstart acceptance test now verifies its example-owned success signals: result artifacts, final model loadability, metrics, and learning thresholds.
- The tokenized `train_args` value remains intentional. PR #5244 contains no framework changes and must merge after #5257, which updates `FedAvgRecipe` and the in-process runner to accept `list[str]` without losing spaces, quotes, or apostrophes. Reverting this to a string would restore the custom-cache path bug. The PR description now states this merge order and validation context explicitly.

Validation with #5257 applied: 34 focused unit tests passed; the real simulator acceptance test passed in 25.82s; `./runtest.sh -s` and `git diff --check` passed. Until #5257 merges into `main`, PR #5244's isolated matrix is expected to fail at the prerequisite Pydantic boundary.

chesterxgchen:
Rebased the example-only branch onto current `upstream/main` (`8f512c1f1`); the new PR head is `1021a0619`. With declared prerequisites #5257 and #5258 applied, the combined unit set passed (`100 passed`), the real simulation→POC acceptance test passed in 72.47 seconds, and `./runtest.sh -s` passed.

chesterxgchen:
Rebased onto current `upstream/main`, which now includes prerequisites #5257 and #5258. The first post-rebase CI run exposed a new test-order collision introduced by the recently merged FedCoRe example: both examples use a top-level `prepare_data.py`, and the Hello PyTorch loader could reuse FedCoRe's cached module. Commit `dd2b0dc46` now isolates and restores both generic sibling imports (`model` and `prepare_data`).

Validation:
- direct FedCoRe-before-Hello-PyTorch reproduction: 4 passed
- focused Hello PyTorch unit and real simulation tests: 18 passed
- `./runtest.sh -s` passed

No unresolved review threads or pending draft reviews were present when refreshed.

chesterxgchen:
Merge-order update: #5270 is now a prerequisite and should merge before this PR. It replaces the shared-workspace backup/restore lifecycle introduced by #5258 with a unique Recipe-owned POC workspace per deployment. After #5270 merges, this branch will be rebased onto the updated `main`; this PR remains example/documentation/test/CI-only.

holgerroth:
One consumer of `job.py` outside this diff still uses the removed flags: `examples/tutorials/nvflare_cli.ipynb`. Both of its export cells (JSON lines 216 and 437) run

```
cd ../hello-world/hello-pt && python job.py --export --export-dir /tmp/nvflare/jobs/job_config --enable_log_streaming --synthetic_data --train_size 2048 --test_size 256 --num_rounds 2 --epochs 1 --batch_size 64 --num_workers 0
```

(the second one with `--train_size 8192 --test_size 512 --num_rounds 20`). Against this head's parser both exit 2:

```
job.py: error: unrecognized arguments: --enable_log_streaming --train_size 2048 --test_size 256 --epochs 1 --batch_size 64 --num_workers 0
```

so `/tmp/nvflare/jobs/job_config/hello-pt` is never created and every later `nvflare job submit` / `job logs` / abort cell in the tutorial fails. The PR removed the identical command from `docs/examples/hello_pt_job_api.rst`; this notebook was missed, and nothing in CI parses it.

Two things to adjust there:

1. Reduce both cells to the flags that survive, e.g. `python job.py --export --export-dir /tmp/nvflare/jobs/job_config --dataset synthetic --num_rounds 2` (and `--num_rounds 20` for the abort demo). The per-client sizes/epochs/batch size now come from `client.py` defaults, so the tutorial's `--train_size 8192` for the long-running abort case needs another way to make the job run long enough, `--num_rounds` alone may do.
2. The prose in the same section (line 206) says the export "enables job log streaming so the later `job logs` cell can retrieve server and client logs", and the `nvflare job logs {JOB_ID} --sites all --tail 80` cell (line 396) relied on `recipe.enable_log_streaming()` from the old `job.py`. The trimmed `job.py` no longer calls it, so either that sentence and the `--sites all` expectation change, or the tutorial gets its own small recipe script.

Reproduction from the example directory:

```
python -c "import job; job.define_parser().parse_args('--enable_log_streaming --synthetic_data --train_size 2048 --test_size 256 --num_rounds 2 --epochs 1 --batch_size 64 --num_workers 0'.split())"
```

IsaacYangSLA (approved):
example, document and tests only.  no nvflare code change.  LGTM.

PR #5245, opened 2026-09-02
feature(security): per-job certificates for job processes

pcnudde:
## Summary

Job processes (SJ/CJ) run on the site's long-lived certificate and private key, so any job-supplied code can read the site key and impersonate the site indefinitely. This PR gives every job process its own short-lived credential and keeps the site keys out of job processes.

- **Provisioning**: `CertBuilder` generates a job-signing intermediate CA (`job_ca.crt`/`job_ca.key`, `CA:TRUE pathlen:0`, root-signed marker extension) into the server startup kit. On by default; `enable_job_ca: false` opts out. Persisted in `cert.json`, regenerated if expired. `nvflare package` kits have no root key and therefore no job CA.
- **Issuance**: at job deploy the SP issues one leaf per participating site (`CN=<site>`, job-ID extension, `notBefore` backdated for clock skew, validity `job_cert_valid_days` in the server resources config, default 30 days, clamped to the CA). The SJ credential is written to the run dir after app deployment (`AppDeployer` recreates the run dir); each CJ credential is pushed in its site's deploy message over the existing authenticated CP-SP channel.
- **Job processes**: `ssl_cert`/`ssl_private_key` point at the job credential, so nothing in SJ/CJ refers to the site key. Job cells use it for TLS and message-level crypto (CJ pins its server-role credential too, for listener-enabled sites). The startup content-integrity check no longer requires the site key in job processes.
- **Site-scope rejection**: `IdentityVerifier.verify_common_name()` rejects any certificate carrying the job-ID extension and any chain containing the job-CA marker, so neither a leaked job leaf key nor a stolen job CA key can register a site, log in as admin, or pass as the server.
- **Job binding in cellnet**: TLS drivers expose the peer cert's job ID as `PEER_JOB_ID`; a job-bound certificate is rejected on any FQCN outside its job, at the connection handshake and on the certificate cached for message-level crypto. One-directional, so older clients that fall back to site certs keep working.
- **Launchers withhold `*.key`** (including `job_ca.key`) when a job credential exists: Docker binds the startup kit file by file; Kubernetes drops keys from the startup Secret and delivers the job credential through the per-pod credential Secret (`NVFLARE_JOB_CERT`/`NVFLARE_JOB_KEY`), installed by the job process before its bootstrap cell starts, with `job_cert/` excluded from workspace bundles and result uploads; Slurm (apptainer/pyxis) binds a keyless staged copy of the kit. Kits without a job CA keep today's behavior with a warning. The in-process launcher and Slurm `sandbox: none` share the host filesystem and cannot isolate.
- Design: `docs/design/per_job_certs_design.md`.

## Compatibility

Kits without a job CA (pre-feature, `enable_job_ca: false`, `nvflare package`), older client releases, non-secure mode, and the simulator are unchanged: job cells fall back to site certificates. Re-provisioning an existing project adds the job CA without changing any other certificate.


pcnudde:
Closing for now; more changes coming before review.

PR #5255, opened 2026-09-03
Add advanced Hello PyTorch environment continuity

chesterxgchen:
Run the shared Hello PyTorch application in simulation, a local POC, or an already-provisioned production federation. The advanced continuation reuses the beginner client, model, and data code from merged #5244 and adds environment selection, training overrides, final or cross-site evaluation, optional TensorBoard tracking, log streaming, and external client execution.

Rebased onto upstream `main` at `3ca61fd9a`, including merged #5244, #5270, and #5280. Changes are scoped to examples, documentation, tests, and CI configuration.

Review fixes:

- Follow Yuan-Ting’s automatic-export/client-validation suggestion: remove the job-level CIFAR prechecks and private `_peek_recipe_args` imports from both entry points. Recipe retains the standard `--export` / `--export-dir` behavior. No new framework APIs or export-state queries are introduced.
- Keep CIFAR validation in the shared client: missing or empty cache files produce a preparation command after services start. Exports require no local cache. Document this timing and recommend absolute client-local data paths; torchvision checks integrity when loading with downloads disabled.
- Preserve downloaded results and service logs when a POC job reports an unsuccessful or unknown status, stop services, and print diagnostic locations. Monitoring interruptions before a result is obtained clean up this run, including `BaseException` paths. Deployment failures remain handled by `PocEnv`.
- Share dataset arguments across the beginner job, advanced job, and client. Explain option defaults, reject negative GC intervals early, and use the existing default admin identity and supported success statuses.
- Document retained-workspace recovery, `NVFLARE_HOME` result locations, explicit CLI download destinations, dataset-dependent learning rates, and the actual CI guarantees. Update continuation and cross-site evaluation links.
- Test observable example behavior, including simulator/POC continuity, standard CLI exports, and retained failed-job diagnostics. Replace tests of removed job-level prechecks with client-validation coverage and a real missing-CIFAR POC failure case.

Validation at `4512d100a`:

- Entire example unit suite: 188 passed, 7 skipped.
- Five real integration tests passed in 177.58 seconds: beginner learning/artifacts, cross-site evaluation, simulator/POC equality and shutdown, and failed-POC result/log retention with shutdown for invalid batch size and missing CIFAR data.
- Actual CLI export tests cover simulation, POC, and production without a local CIFAR cache; beginner exports preserve both relative and absolute paths containing spaces and apostrophes. Production tests do not connect to a live federation.
- Repository style (`./runtest.sh -s --skip-install`), license checks, and `git diff --check` passed.


chesterxgchen:
The current head `6722cd4f4` addresses the production-status compatibility finding.

The advanced example accepts both the current `FINISHED:COMPLETED` status and the Flare API compatibility status `FINISHED_OK` after a result is downloaded. A focused regression test executes the production path with `FINISHED_OK` and verifies that it returns the result and reports success.

The PR description has also been refreshed to match the final example-only split and its merge order: #5257, #5258, #5244, then this PR.

chesterxgchen:
Rebased the example-only branch onto current `upstream/main` (`8f512c1f1`); the new PR head is `12380867d`. The `FINISHED_OK` compatibility fix described above is unchanged. With declared prerequisites #5257, #5258, and #5244 applied, the combined unit set passed (`100 passed`), the real simulation→POC acceptance test passed in 72.47 seconds with identical per-site results, and `./runtest.sh -s` passed.

chesterxgchen:
Rebased and restacked this branch on current PR #5244 after #5257 and #5258 merged. This exposed and resolved the three actual stack conflicts in the premerge workflow, guided RST, and canonical README. Both CI gates are retained; the new beginner documentation remains canonical; only the POC/production continuation links are added.

Validation on the exact pushed stack:
- affected subsystem unit set: 379 passed, 2 skipped
- beginner + advanced tests with real simulation and real local POC: 31 passed in 95.57s
- import-order regression plus both job test modules: 24 passed
- relative-link tests/checks passed
- `./runtest.sh -s` passed
- HTML documentation build passed

No unresolved review threads or pending draft reviews were present when refreshed.

chesterxgchen:
Merge-order update: #5270 must merge first, followed by #5244, then this PR. Once #5270 is on `main`, this branch will be rebased and its integrated POC failure handling will be simplified to rely on the new per-deployment Recipe workspace; the temporary `workspace_owned` compatibility check will be removed.

holgerroth (commented):
Reviewed at `76747c527` (rebased on the merged https://github.com/NVIDIA/NVFlare/pull/5244). Locally: the 37 beginner and advanced unit tests and the whole `tests/unit_test/examples` session pass, style, license, and link checks are clean, exports in default, external-process plus log-streaming plus cross-site plus TensorBoard, and `--evaluation none` modes produce the expected configs, `python job.py` reproduces 75/77, and the simulation-to-POC continuity test passes in 80 s.

One item should be fixed before merge: the cleanup path on a failed job deletes the run's workspace together with the only client logs and the downloaded result (first comment on `job.py`; reproduced with a real POC run). The rest are inline.

Since #5244 merged with a few open items from its last review, this PR is the natural place for the ones that touch the same files or concerns:

- `validate_cifar10` accepts zero-byte files (seven empty files pass `is_file()` and torchvision's misleading "use download=True" error comes back); a size or checksum check belongs with the CIFAR-10 path this example documents.
- The beginner fail-fast surfaces as a raw traceback while its README says it "prints the preparation command"; catching `FileNotFoundError` in `main()` fixes both scripts the same way.
- `--help` text for the headline flags in both parsers.
- Document the dataset-dependent learning rate (0.1 synthetic, 0.01 CIFAR-10) and the third acceptance-gate condition (`initial_accuracy <= 20`) in the README and RST this PR already edits.
- Merge the "environment-continuity follow-up" sentences into the new link sections (inline).
- `examples/hello-world/hello-numpy-cross-val/README.md:259` still calls hello-pt "the PyTorch CSE example"; pointing it at this example's `--evaluation cross-site` fixes it.

Two items are better as a small framework PR ahead of this one, because both examples need them: a public export-mode accessor (the beginner `job.py` imports the private `nvflare.recipe.spec._peek_recipe_args`, and this script needs the same check to add the CIFAR preflight) and a public success-status check. The CLI-tutorial exporter's `--job-dir` handling, the website Colab wording, and the hello-pt notebook install cell are unrelated to this PR and can stay a separate follow-up.

examples/advanced/hello-pt-environments/job.py
```diff
@@ -0,0 +1,208 @@
...
+        status = None if args.env == "sim" else run.get_status()
+        if result is None:
+            raise RuntimeError("Job monitoring did not return a result. Review the execution-environment logs.")
+        if args.env != "sim" and status not in SUCCESS_STATUSES:
+            raise RuntimeError(f"Job completed with unsuccessful status: {status}")
+    except (Exception, KeyboardInterrupt):
+        # Each PocEnv has a unique workspace, so monitoring failure cleanup only
+        # targets this run. Previously retained results belong to other paths.
+        if args.env == "poc":
+            env.stop(clean_up=True)
```
holgerroth: **Cleanup on a failed job deletes the only logs and the downloaded result.**

Any non-success terminal status takes this path, not only monitoring failures: `monitor_job` treats every `FINISHED*` status as terminal, so `get_result` returns a path, line 188 raises, and `env.stop(clean_up=True)` removes the `.recipe-<hex>` workspace, which holds the site logs and the downloaded result (`poc_env.py:602-609`). Reproduced with `python job.py --env poc --dataset cifar10 --data_root <empty dir> --num_rounds 1`: the clients raise the CIFAR-10 `FileNotFoundError`, the job ends `FINISHED:EXECUTION_EXCEPTION`, the script prints `RuntimeError: Job completed with unsuccessful status: FINISHED:EXECUTION_EXCEPTION`, and afterwards no `poc.recipe-*` directory and no client log with the error exist. The message on line 187 ("Review the execution-environment logs") points at files this branch deletes one line later, and the result path is never printed on this branch. The same happens for a successful download whose `get_job_status` lookup failed transiently (status `None`). README lines 50-52 promise cleanup only for provisioning, submission, or monitoring failures, and `test_main_rejects_unsuccessful_poc_status` locks the deletion in.

Suggest keeping the workspace whenever `result` is not `None` (print the result path and the site log locations, then raise) and reserving `clean_up=True` for the `result is None` case.

examples/advanced/hello-pt-environments/job.py
```diff
@@ -0,0 +1,208 @@
...
+    if args.env == "poc":
+        return PocEnv(num_clients=args.n_clients)
+    if args.username:
+        return ProdEnv(startup_kit_location=args.startup_kit, username=args.username)
+    return ProdEnv(startup_kit_location=args.startup_kit)
+
+
+def main(argv=None):
+    args = parse_args(argv)
+    recipe = create_recipe(args)
```
holgerroth, line 184: **No CIFAR-10 preflight for sim and POC.**

The merged beginner `job.py` validates the cache before starting anything (`hello-pt/job.py:93-96`); this script does not, so `python job.py --dataset cifar10 --data_root <empty>` starts the simulator and fails inside both clients (186-line log, six `FileNotFoundError`s), and with `--env poc` it pays provisioning and startup first and then hits the deletion above, so even the helpful message is gone. For `sim` and `poc` the clients are local and the beginner check applies verbatim; for `prod` the README's client-local caveat is right. The beginner version uses the private `nvflare.recipe.spec._peek_recipe_args` to skip validation on export; see the note on line 34 about making that public rather than copying the private import here.

examples/advanced/hello-pt-environments/job.py
```diff
@@ -0,0 +1,208 @@
...
+    try:
+        # Retain the POC workspace after success so the printed result path and
+        # service logs remain accessible.
+        result = run.get_result(clean_up=args.env != "poc")
+        status = None if args.env == "sim" else run.get_status()
+        if result is None:
+            raise RuntimeError("Job monitoring did not return a result. Review the execution-environment logs.")
+        if args.env != "sim" and status not in SUCCESS_STATUSES:
+            raise RuntimeError(f"Job completed with unsuccessful status: {status}")
+    except (Exception, KeyboardInterrupt):
```
holgerroth: **`except (Exception, KeyboardInterrupt)` misses other `BaseException`s.**

pytest-timeout's `Failed` (MRO `Failed -> OutcomeException -> BaseException`), `SystemExit`, and `GeneratorExit` skip this handler, so a monitoring hang that trips `@pytest.mark.timeout(300)` leaves the POC server and clients running on 8002/8003 and the next local run fails its port preflight (`poc_env.py:357-360`). `PocEnv._deploy` already uses `except BaseException` for the same reason (`poc_env.py:537`); `except BaseException:` with the same body here matches it.

examples/advanced/hello-pt-environments/job.py
```diff
@@ -0,0 +1,208 @@
...
+from model import create_model  # noqa: E402
+from prepare_data import DATASET_CHOICES, DATASET_PATH, DEFAULT_DATASET  # noqa: E402
+
+from nvflare.app_opt.pt.recipes.fedavg import FedAvgRecipe
+from nvflare.recipe import PocEnv, ProdEnv, SimEnv, add_experiment_tracking, add_final_global_evaluation
+from nvflare.recipe.utils import add_cross_site_evaluation
+
+DEFAULT_NUM_CLIENTS = 2
+DEFAULT_NUM_ROUNDS = 3
+SUCCESS_STATUS = "FINISHED:COMPLETED"
```
holgerroth: **Success and export decisions belong to the recipe layer.**

`SUCCESS_STATUSES` duplicates the private `_LEGACY_TERMINAL_JOB_STATUSES` (`nvflare/fuel/flare_api/flare_api.py:81`) and `_TERMINAL_JOB_STATES` (`nvflare/tool/job/job_cli.py:1927`), and adding the CIFAR preflight here would mean copying the beginner's import of the private `_peek_recipe_args`. Two hello-pt scripts now decide "did the job succeed?" and "am I exporting?" on their own, so a change to the status vocabulary or to `--export` handling silently diverges from `nvflare job` and `Recipe.execute`. A public success check (`RunStatus`-aware, or `Run.succeeded()`) and a public `export_requested()` in `nvflare.recipe` are a few framework lines; they fit a small separate PR ahead of this one, after which both examples can drop the private import and the literals.

examples/advanced/hello-pt-environments/job.py
```diff
@@ -0,0 +1,208 @@
...
+        default=None,
+        help="Path to an admin startup kit. Required with --env prod.",
+    )
+    parser.add_argument(
+        "--username",
+        default=None,
+        help="Production admin identity. Must match the startup kit; defaults to admin@nvidia.com.",
+    )
+
+    dataset_group = parser.add_mutually_exclusive_group()
```
holgerroth: **Third copy of the dataset argument block.**

Lines 68-82 are the same mutually-exclusive group, `--synthetic_data` alias, `set_defaults`, and `--data_root` help as `examples/hello-world/hello-pt/job.py:46-60` and `client.py:105-121`. Importing the beginner `job` is not viable (the advanced directory precedes `HELLO_PT_DIR` on `sys.path` under the shared test loader, so `import job` resolves to this file), but a `prepare_data.add_dataset_arguments(parser)` helper in the module all three already import from removes the copies; the recipe construction (137-147) could take the same route through a shared builder in the beginner example.

examples/advanced/hello-pt-environments/job.py
```diff
@@ -0,0 +1,208 @@
...
+    env = create_environment(args)
+
+    # PocEnv handles deployment failures and preserves the workspace if service
+    # shutdown cannot be verified. Only monitor a successfully submitted run.
+    run = recipe.execute(env)
+    try:
+        # Retain the POC workspace after success so the printed result path and
+        # service logs remain accessible.
+        result = run.get_result(clean_up=args.env != "poc")
+        status = None if args.env == "sim" else run.get_status()
```
holgerroth: **The `sim` special-casing is unnecessary, and the literals can use `RunStatus`.**

`SimEnv` records `RunStatus.FINISHED_COMPLETED.value` on success and raises on failure (`sim_env.py:151-156`), and `Run.get_status()` serves the cached status once `get_result` has stopped the environment (`run.py:68-70`), so `status = run.get_status()` and one `if result is None or status not in SUCCESS_STATUSES` cover all three environments; only the printed sentence differs. `SUCCESS_STATUS` can be `RunStatus.FINISHED_COMPLETED.value`, and `create_environment` can pass `username=args.username or DEFAULT_ADMIN_USER` (from `nvflare.recipe.prod_env`) instead of branching.

examples/advanced/hello-pt-environments/job.py
```diff
@@ -0,0 +1,208 @@
...
+"""
+
+
+def define_parser() -> argparse.ArgumentParser:
+    parser = argparse.ArgumentParser(
+        description="Continue Hello PyTorch through simulation, POC, or production.",
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+        epilog=EXPORT_HELP,
+    )
+    parser.add_argument("--n_clients", type=int, default=DEFAULT_NUM_CLIENTS)
```
holgerroth: **No help text on the headline flags.**

`--n_clients`, `--num_rounds`, `--dataset`, `--batch_size`, `--epochs`, `--learning_rate`, and `--num_workers` print as bare `--flag VALUE` entries between fully described neighbours, while README line 119 sends readers to `python job.py --help` for all options. The overrides are where the client defaults matter (learning rate 0.1 for synthetic and 0.01 for CIFAR-10, one epoch, batch 32), and none of that is discoverable from `--help`. The merged beginner `job.py` has the same gap; worth fixing both parsers together.

examples/advanced/hello-pt-environments/job.py
```diff
@@ -0,0 +1,208 @@
...
+        default="none",
+        help="Optional server-side tracking receiver. TensorBoard requires the tensorboard package.",
+    )
+    parser.add_argument("--enable_log_streaming", action=argparse.BooleanOptionalAction, default=False)
+    parser.add_argument(
+        "--launch_external_process",
+        action="store_true",
+        help="Run the shared client script in a separate subprocess instead of in-process.",
+    )
+    parser.add_argument(
```
holgerroth, line 109: **Negative `--client_memory_gc_rounds` is only rejected inside the clients.**

The recipe stores the value unchanged and `ClientAPIExecutor` raises `ValueError: memory_gc_rounds must be an integer >= 0` on every client at job start (`client_api_executor.py:282-283`), so `--env poc --client_memory_gc_rounds -1` provisions and starts a POC, fails, and (per the note on line 194) deletes the logs that show why. A non-negative `type=` converter or `parser.error` at the CLI fails fast.

examples/advanced/hello-pt-environments/job.py
```diff
@@ -0,0 +1,208 @@
...
+# limitations under the License.
+"""Run the Hello PyTorch application across simulation, POC, or production."""
+
+import argparse
+import sys
+from pathlib import Path
+
+# Reuse the beginner example's real model, client, and data code. This example
+# changes orchestration and execution environment, not the learning application.
+HELLO_PT_DIR = Path(__file__).resolve().parents[2] / "hello-world" / "hello-pt"
```
holgerroth: **The repository layout is assumed without a check.**

`parents[2] / "hello-world" / "hello-pt"` goes on `sys.path` unconditionally, so copying this directory anywhere else yields `ModuleNotFoundError: No module named 'model'`, and a shallow path like `/tmp/job.py` raises `IndexError` from `parents[2]`. A one-line `if not HELLO_PT_DIR.is_dir(): raise SystemExit(...)` plus a README sentence makes the dependency explicit.

examples/advanced/hello-pt-environments/README.md
```diff
@@ -0,0 +1,158 @@
...
+## Compare the three environments
+
+| Stage | What changes | Command |
+| --- | --- | --- |
+| Simulation | Server and clients run through the local simulator. | `python job.py` |
+| POC | `PocEnv` provisions, starts, and stops a local system for this job. | `python job.py --env poc` |
+| Production | `ProdEnv` connects to an already-running system through an authorized admin startup kit. | `python job.py --env prod --startup-kit <admin-kit> --username <admin-identity>` |
+
+The Recipe, model, client script, data code, and local training loop stay the same across all three stages. With the
+fixed defaults, both simulation and POC report 75% final accuracy on site-1 and 77% on site-2. Premerge CI enforces
```
holgerroth: **This paragraph overstates what CI enforces.**

No test asserts 75%/77%: the continuity test asserts simulation == POC and the quickstart gate asserts thresholds, and the beginner README deliberately calls its numbers "not benchmark claims". "Production construction/export" covers less than it sounds: `test_environment_selection_uses_the_same_client_count` builds a `ProdEnv` against an empty temp directory and checks two attributes, and the export test uses `SimEnv`; nothing exports or runs with `ProdEnv`. Suggest stating the equivalence check only and describing the numbers as what the fixed seeds produce today.

examples/advanced/hello-pt-environments/README.md
```diff
@@ -0,0 +1,158 @@
...
+```bash
+python job.py --env poc
+```
+
+`PocEnv` provisions a local federation, starts separate server and client processes, submits the same application,
+downloads the result, and stops the services. It needs permission to start processes and bind the standard NVFlare
+POC ports. The POC lifecycle belongs to this invocation, so provisioning and process startup occur for every job and
+make this deliberately slower than simulation.
+
+Each invocation uses a unique workspace beside the configured CLI POC workspace, with a `.recipe-<UUID>` suffix.
```
holgerroth, line 48: **`NVFLARE_HOME` moves the result out of the retained workspace.**

`PocEnv._deploy` provisions with `examples_dir=None` (`poc_env.py:495-500`); when `NVFLARE_HOME` is set, `get_examples_dir` falls back to `$NVFLARE_HOME/examples` (`poc_commands.py:195-196`) and `_link_jobs_dir_to_admin_transfer` symlinks the admin `transfer` directory there, which is also the download directory. A developer with the variable exported (the POC tutorial sets it) gets one `<job_id>` directory per run in their NVFlare checkout, and deleting `poc.recipe-*` trees does not remove them. The integration test works around exactly this with `monkeypatch.delenv("NVFLARE_HOME")`; the example and this section should say so, or the script should pass an explicit examples directory.

examples/advanced/hello-pt-environments/README.md
```diff
@@ -0,0 +1,158 @@
...
+On success, the command retains that workspace so the printed result path and service logs remain available across
+later runs. The configured CLI workspace and earlier Recipe results are preserved. Stop any running CLI POC with
```
holgerroth, line 50: **Retention needs cleanup guidance.**

Each successful `--env poc` run leaves a full provisioned tree under `/tmp/nvflare/poc.recipe-<32 hex>` (three kits, service logs, job copies, downloaded results) distinguishable only by mtime, and a run killed mid-startup leaves services on 8002/8003 that `nvflare poc stop` cannot stop because it only targets the CLI workspace (`poc_env.py:378-380`), so the "Stop any running CLI POC" advice on line 48 does not clear the next preflight failure. Documentable: `NVFLARE_POC_WORKSPACE=/tmp/nvflare/poc.recipe-<hex> nvflare poc stop`, `rm -rf /tmp/nvflare/poc.recipe-*`, and `nvflare job download JOB_ID -o <dir>` (lines 87 and 103 currently download into the current directory, which is this example directory inside the checkout when following the page).

docs/hello-world/hello-pt/index.rst
```diff
@@ -120,5 +120,13 @@ experiment tracking, full cross-site evaluation, external-process execution,
 and memory tuning belong in a separate continuation workflow rather than the
```
holgerroth: **The old "separate continuation workflow" sentence now sits unlinked next to the new section.**

Lines 117-122 here and `examples/hello-world/hello-pt/README.md:224-228` ("the environment-continuity follow-up") still describe this example as something that does not exist yet, while the new heading links it a few lines later; in this file the closing "For the API concepts..." paragraph now renders under "Continue to POC and Production". Making the earlier sentence the link and dropping or merging the new section, and moving the API-concepts paragraph above the heading, reads cleanly.

tests/unit_test/examples/hello_pt_environments_job_test.py
```diff
@@ -0,0 +1,321 @@
...
+    monkeypatch.setattr(job_module, "create_recipe", lambda args: recipe)
+    monkeypatch.setattr(job_module, "create_environment", lambda args: env)
+
+    with pytest.raises(RuntimeError, match="did not return a result"):
+        job_module.main(["--env", "poc"])
+
+    assert stop_calls == [False, True]
+
+
+@pytest.mark.parametrize("error_type", [RuntimeError, KeyboardInterrupt])
```
holgerroth: **This test mostly re-tests `PocEnv`.**

It drives the real `PocEnv._deploy` with `_preflight_ports_before_provision` and the module-level `prepare_poc_provision` monkeypatched and asserts `_clean_up_failed_deployment`'s behaviour (lines 281-284), which belongs in `tests/unit_test/recipe`. The example-specific claim (main does not call `env.stop` on deploy failure) holds structurally because `recipe.execute` sits outside the `try`, and a fake env whose `deploy` raises plus `stop=pytest.fail` proves it without torch models or private patches. A `PocEnv` refactor should not break an example test.

examples/advanced/hello-pt-environments/job.py
lucasromanomr: Corrigido em 5376b168c: `update_goal` agora falha fechado quando o Harness está ON e o `CompletionVerifier` não está configurado — retorna o mesmo erro do `spec_task.complete`, emite `harness.completion.blocked` (reason=other) e não marca o goal como completo. Teste `TestUpdateGoalCompleteRejectsMissingCompletionVerifier` cobre o cenário.

examples/advanced/hello-pt-environments/job.py
chesterxgchen: Fixed in 43015416c. A downloaded result now survives unsuccessful and unavailable statuses: services are stopped with clean_up=False, and the result path, retained workspace, and service log paths are printed before the error propagates. Only failures before a result is obtained request cleanup. Unit tests cover aborted, execution-exception, and None statuses. A real POC job with an intentional client error verifies retained downloads/logs and stopped services.

examples/advanced/hello-pt-environments/job.py
chesterxgchen: Fixed in 43015416c. Both local environments validate the CIFAR cache before recipe/environment construction; export and production skip admin-local checks using the public export_requested() query from #5281. Missing or empty files produce the correct shared preparation-script command without a traceback. Tests cover sim/POC rejection, production bypass, and actual exports for all three environments.

examples/advanced/hello-pt-environments/job.py
chesterxgchen: Fixed in 43015416c. Monitoring cleanup catches BaseException and re-raises after stopping the POC. Tests cover KeyboardInterrupt, SystemExit, GeneratorExit, and pytest.fail.Exception; deployment errors remain handled by the environment.

examples/advanced/hello-pt-environments/job.py
chesterxgchen: Implemented separately in #5281 as requested: nvflare.recipe.export_requested() and Run.succeeded(), with tests for current/legacy/unknown/failed statuses and cached results. This PR is stacked on that prerequisite and both example entry points use the public queries. The private export import and example-owned success literals are removed. Merge #5281 first; its commit can then be dropped from this PR by rebasing.

examples/advanced/hello-pt-environments/job.py
chesterxgchen: Fixed in 43015416c. prepare_data.add_dataset_arguments(parser) is now used by the beginner job, advanced job, and client, keeping dataset choices, alias, defaults, and cache help in one place. Recipe construction remains explicit in each entry point so the advanced controls stay visible. The Harness/update_goal reply above appears to refer to another change; this fix addresses the dataset argument duplication.

examples/advanced/hello-pt-environments/job.py
chesterxgchen: Fixed in 43015416c using Run.succeeded() from #5281. All three environments use the same status/success path; only the displayed success sentence differs. Production construction now passes args.username or DEFAULT_ADMIN_USER without branching. The shared success query recognizes RunStatus.FINISHED_COMPLETED.value and legacy FINISHED_OK.

examples/advanced/hello-pt-environments/job.py
chesterxgchen: Fixed in 43015416c. Both job parsers describe client count, round count, and dataset selection. Advanced override help explains batch size 32, one local epoch, zero loader workers, and learning rates 0.1 for synthetic / 0.01 for CIFAR-10. Omitted overrides continue to use the shared client defaults.

examples/advanced/hello-pt-environments/job.py
chesterxgchen: Fixed in 43015416c. parse_args rejects negative --client_memory_gc_rounds with parser.error before constructing a recipe or environment. A regression test verifies that execution is not reached and the corrective message names the option.

examples/advanced/hello-pt-environments/job.py
chesterxgchen: Fixed in 43015416c. Path traversal no longer indexes parents[2], and the script checks that the shared client/model/data files exist before modifying sys.path or importing them. A copied script exits with a full-checkout instruction and no traceback. The README now states the repository-layout requirement.

examples/advanced/hello-pt-environments/README.md
chesterxgchen: Fixed in 43015416c. The README labels 75%/77% as current fixed-seed observations and states that CI compares simulator/POC per-site results rather than asserting those exact numbers. Actual CLI export tests now cover ProdEnv as well as SimEnv/PocEnv; the description explicitly excludes validation against a live production federation.

examples/advanced/hello-pt-environments/README.md
chesterxgchen: Addressed in 43015416c. The README explains that NVFLARE_HOME redirects downloaded results to its examples directory, recommends unsetting it when downloads should stay in the run workspace, and explains that external downloads need separate removal. The command now prints the retained workspace as well as the result path, including failed-job diagnostics.

examples/advanced/hello-pt-environments/README.md
chesterxgchen: Addressed in 43015416c. The README shows how to target an interrupted Recipe workspace with NVFLARE_POC_WORKSPACE, locate its logs, and remove that specific retained directory after confirming shutdown and saving needed artifacts. CLI download examples use an explicit /tmp output directory. Cleanup guidance deliberately targets one identified run rather than a wildcard over all retained runs.

docs/hello-world/hello-pt/index.rst
chesterxgchen: Fixed in 43015416c. The beginner README links directly to its continuation section and lists the advanced controls there. The RST removes the future-work wording, folds the controls into the linked continuation section, and moves the API-concepts paragraph above that heading. Documentation links and HTML build checked.

tests/unit_test/examples/hello_pt_environments_job_test.py
chesterxgchen: Fixed in 43015416c. The example deployment-failure test now uses a minimal failing execute call and an environment whose stop fails the test if invoked. It no longer patches PocEnv private methods, provisions directories, or re-tests framework workspace cleanup. The separate real POC acceptance tests verify the observable service/artifact behavior of the example.

chesterxgchen:
@holgerroth, the review fixes are pushed in 43015416c, and each of the 14 inline threads now has a specific response with its disposition.

The additional summary items are addressed too: missing/zero-byte CIFAR files fail early, both entry points show the preparation command without a traceback, CLI help explains defaults, README/RST document the dataset-dependent learning rates and initial-accuracy <=20% gate, continuation wording is linked directly, and the NumPy cross-validation README points to the advanced PyTorch cross-site command.

The requested framework queries are isolated in #5281. This PR is stacked on it and adopts the public APIs in both examples. Please merge #5281 first; this PR can then be rebased to remove the prerequisite commit.

Validation: 194 example unit tests passed (7 skipped), prerequisite spec/Run tests 101 passed (3 skipped), and all four real integration tests passed in 152.91 seconds. The new failed-POC test verifies that the downloaded result and client error logs remain and services stop. Style, license, local documentation links, and HTML build checked.

The unrelated CLI-tutorial exporter, website Colab wording, and notebook-install follow-ups remain outside this PR, as suggested in the review.

YuanTingHsieh:
The export control flow is confusing: Recipe consumes `--export` automatically, but the example then needs `export_requested()` to inspect that decision before validating local data. This makes the caller depend on CLI state that `execute()` is intended to handle transparently.

Can we choose one clear approach here?

1. **Explicit export handling:** register `--export` in the entry-point parser, then branch between `recipe.export()` and local validation followed by execution. A shared parser helper could avoid repeating option definitions across examples. This would require changing the existing import-time flag consumption, not just adding an argument to this script.
2. **Automatic export with client-side validation:** remove the job-level CIFAR precheck and rely on the shared client's existing validation. Export remains transparent, but missing data is reported after simulation or POC starts.

If early local-data validation is required, explicit handling seems simpler. If automatic export is the priority, the second option avoids exposing another CLI-state query. Explicit flags were intentionally migrated away from, so this is a tradeoff to settle deliberately rather than adding `export_requested()` without resolving the ownership question.

This concern is specific to `export_requested()`. `Run.succeeded()` queries a run's outcome and fits naturally on `Run`.

examples/advanced/hello-pt-environments/job.py
chesterxgchen: Scope update in a6317ffad: we are not adopting the new public API proposal in this change. #5281 is closed, and its commit has been removed from this branch; there are no changes to nvflare/recipe or its tests.

The advanced example uses the existing RunStatus.FINISHED_COMPLETED.value and retains compatibility with FINISHED_OK. Both examples use the existing export-flag helper to skip local CIFAR checks when exporting. No new Run.succeeded() or export_requested() API is required.

Centralizing these decisions is a maintenance suggestion, but moving the same status mapping into a new method does not itself guarantee compatibility with future status changes. We are keeping the current framework API and focusing this PR on the demonstrated example issues. The failed-job result/log retention, cleanup, CIFAR preflight, and other review fixes remain intact. Validation after removal: 194 example unit tests passed (7 skipped), all four real integration tests passed in 155.05 seconds, and repository style passed.

chesterxgchen:
Correction to the earlier prerequisite note: #5281 is closed and its API commit has been removed from this PR. Head a6317ffad builds on upstream main and makes no framework API changes. The example correctness fixes remain, using existing status and export handling. We are not adopting the proposed public API addition as part of this work.

Revalidated: 194 example unit tests passed (7 skipped), all four real integration tests passed in 155.05 seconds, and repository style passed.

examples/advanced/hello-pt-environments/job.py
chesterxgchen: @lucasromanomr  is your comment related to this PR ?

chesterxgchen:
@YuanTingHsieh, your comment concerned the proposed new export_requested() API. That API has already been removed, and #5281 is closed. The current PR keeps the established Recipe-managed --export and --export-dir behavior; we are not introducing explicit export parsing or changing export ownership.

The existing local CIFAR preflight and its export bypass also remain. There are no framework API changes in the current diff. No further export-flow redesign is planned for this PR.

examples/advanced/hello-pt-environments/job.py
```diff
@@ -0,0 +1,225 @@
...
+sys.path.insert(0, str(HELLO_PT_DIR))
+
+from model import create_model  # noqa: E402
+from prepare_data import add_dataset_arguments, validate_cifar10  # noqa: E402
+
+from nvflare.apis.job_def import RunStatus
+from nvflare.app_opt.pt.recipes.fedavg import FedAvgRecipe
+from nvflare.recipe import PocEnv, ProdEnv, SimEnv, add_experiment_tracking, add_final_global_evaluation
+from nvflare.recipe.prod_env import DEFAULT_ADMIN_USER
+from nvflare.recipe.spec import _peek_recipe_args
```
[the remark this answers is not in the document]
chesterxgchen: We considered this suggestion and are not adopting a new public export-state API in this PR. The proposed export_requested() addition was removed, #5281 is closed, and the maintainer has confirmed that the established Recipe-managed --export and --export-dir behavior should remain unchanged.

The examples retain the existing helper for the local CIFAR preflight export bypass. We accept this existing internal dependency for this scoped example change; it is not a request to redesign the export flow or add another API. Please treat the public export-state API proposal as intentionally declined for this PR.

holgerroth (dismissed):
Approving at `89058467f`. Re-verified the fixes locally on this head:

- The failed-job path in POC now retains the run workspace: a job forced to fail after startup (`--env poc --batch_size 0`) ends with the result path, the retained workspace, and the three `poc_console.log` paths printed, the per-site error logs and the downloaded result still on disk, and no services left running. The unprepared CIFAR-10 case now stops in one line before the simulator or POC starts, for both the beginner and the advanced entry point.
- No regressions from the shared beginner-file changes: the merged quickstart acceptance tests (2) and the new environment-continuity tests (2, including the failed-POC case) pass on this head, 232 tests in `tests/unit_test/examples` plus the skill checks pass in one session, and every existing consumer of the client argv (the slow recipe test's `--synthetic_data ...` string, the CLI tutorial exporter, the notebook's `--dataset cifar10 --batch_size ...`) still parses through `add_dataset_arguments`. Style, license, and relative-link checks are clean. The simulation still reproduces 75/77.

One item stays open by agreement: both entry points keep the private `nvflare.recipe.spec._peek_recipe_args` import for the export bypass after https://github.com/NVIDIA/NVFlare/pull/5281 was closed. That is fine for this scoped change; a public export-state query (or the explicit handling Yuan-Ting described) can follow separately once the ownership question is settled.

examples/advanced/hello-pt-environments/job.py
YuanTingHsieh: still i think we should just not do this.

examples/advanced/hello-pt-environments/job.py
```diff
@@ -0,0 +1,225 @@
...
+
+def main(argv=None):
+    args = parse_args(argv)
+    export_only, _ = _peek_recipe_args()
+    if args.dataset == "cifar10" and args.env != "prod" and not export_only:
+        args.data_root = str(Path(args.data_root).expanduser().resolve())
+        try:
+            validate_cifar10(args.data_root, prepare_script="../../hello-world/hello-pt/prepare_data.py")
+        except FileNotFoundError as e:
+            raise SystemExit(str(e)) from None
```
YuanTingHsieh: maybe we just do not validate in job construction

examples/hello-world/hello-pt/job.py
```diff
@@ -93,15 +81,19 @@ def main(argv=None):
     export_only, _ = _peek_recipe_args()
     if args.dataset == "cifar10" and not export_only:
         args.data_root = str(Path(args.data_root).expanduser().resolve())
-        validate_cifar10(args.data_root)
+        try:
+            validate_cifar10(args.data_root)
```
YuanTingHsieh: maybe we just dont validate in job.py, validation in client.py makes sense

YuanTingHsieh (changes requested):

examples/hello-world/hello-pt/job.py
chesterxgchen: I kind of agree on this, pondering different options for a while

chesterxgchen:
@YuanTingHsieh @holgerroth — implemented the automatic-export/client-validation approach in `4512d100a` and rebased onto upstream main (`3ca61fd9a`).

Both entry points now omit the job-level CIFAR precheck and the private `_peek_recipe_args` import. Recipe continues to own `--export` / `--export-dir`; there is no new public query or validation hook.

This supersedes my earlier decision to retain the precheck: missing or empty data is now reported by the shared client after simulation/POC starts. Both READMEs describe this timing and recommend absolute client-local data paths. Failed POC results and logs are still preserved, and the new real missing-CIFAR integration case checks the preparation error in retained logs and confirms service shutdown.

Validation: 188 example unit tests passed (7 skipped); all five real Hello PyTorch integration cases passed; repository style and license checks passed. The PR description now reflects the final behavior.

examples/advanced/hello-pt-environments/job.py
chesterxgchen: Addressed in 4512d100a. Both Hello PyTorch entry points now omit the private _peek_recipe_args import. The existing Recipe-managed --export / --export-dir behavior remains unchanged; no public export query or hook was added.

examples/advanced/hello-pt-environments/job.py
chesterxgchen: Addressed in 4512d100a. Removed CIFAR cache validation from job.py. Recipe construction no longer checks data on the submitting machine; validation remains in the shared client when it loads data.

examples/hello-world/hello-pt/job.py
chesterxgchen: Implemented your automatic-export/client-validation suggestion in 4512d100a. Both entry points omit the job-level precheck. Missing or empty data is reported after startup, as documented in both READMEs. Client tests cover missing, partial, and empty caches; a real missing-CIFAR POC test verifies the preparation error remains in retained logs and all services stop. All five integration cases and 188 example unit tests passed.

examples/advanced/hello-pt-environments/job.py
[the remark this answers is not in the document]
chesterxgchen: Updated in 4512d100a: the private helper dependency is now removed from both entry points. Following Yuan-Ting’s suggestion, CIFAR validation stays in client.py, so job.py no longer needs to inspect export state. Standard Recipe export behavior is unchanged; no public export-state API is needed.

examples/advanced/hello-pt-environments/job.py
chesterxgchen: Updated disposition in 4512d100a: following Yuan-Ting’s automatic-export/client-validation suggestion, both entry points now omit the job-level CIFAR precheck and private export-helper import. Missing or empty data is reported after services start. Both READMEs document this tradeoff. Failed POC results/logs remain preserved; a real missing-CIFAR integration case verifies the preparation error is retained and services stop. This supersedes the earlier early-validation fix described above.

YuanTingHsieh (approved):

PR #5257, opened 2026-09-03
Improve Recipe script arguments and simulation status

chesterxgchen:
## Summary

This extracts two framework-level behaviors from the Hello PyTorch example work so the example PRs remain scoped to examples, documentation, tests, and CI.

- allow in-process Client API scripts to receive pre-tokenized `list[str]` arguments
- preserve the historical whitespace-only behavior for string arguments
- expose argv-valued script arguments and external commands consistently across FedAvg, PyTorch FedEval, and PyTorch Swarm recipes
- carry argv through `FedAvgRecipe`, `ScriptRunner`, `ClientAPIExecutor`, and the in-process backend without shell reparsing
- resolve supported secret references in argv elements and recognize argv-valued per-site `train_args`/`command` fields as supported runtime boundaries
- regenerate the Recipe catalog so CLI introspection advertises the argv-capable FedAvg signatures
- record server-aligned terminal status for synchronous `SimEnv` runs

## Why

Shell syntax cannot unambiguously reinterpret legacy argument strings. For example, `O'Brien` must remain one literal token, while quoting is often used to represent a path containing spaces. Pre-tokenized argv makes boundaries explicit and avoids changing the meaning of every existing in-process script string.

`SimEnv` runs synchronously. A successful deployment returns a `Run` whose status is `FINISHED:COMPLETED`. A deployment failure raises before a `Run` is constructed; the environment itself records `FINISHED:FAILED_TO_RUN` for setup/startup exceptions, `FINISHED:EXECUTION_EXCEPTION` for ordinary nonzero simulator exits, or `FINISHED:ABNORMAL` for a `-9` exit. Unknown job IDs still return `None`, and failed attempts cannot expose a stale workspace left by an earlier run with the same name.

This PR intentionally contains no Hello PyTorch example changes. PRs #5244 and #5255 consume these behaviors as example-only follow-ups.

## Compatibility

- Existing string-valued script arguments retain their whitespace-only tokenization.
- Existing exported `task_script_args: null` configurations remain accepted and normalize to empty arguments.
- Existing top-level `FedEvalRecipe(eval_args=None)` calls remain accepted and normalize consistently with per-site/runner empty arguments.
- Pre-tokenized argv is a backward-compatible extension.
- A shared argv normalizer validates and copies argument lists at Recipe, runner, executor, and task boundaries so later caller mutation does not change execution.
- Deferred FedEval top-level and per-site argv values are validated and copied when configured; Swarm external commands advertise and enforce the same contract.
- Invalid argv containers or non-string elements fail while the Recipe or runner is configured, before export or execution.
- Direct and per-site secret references remain placeholders in exported config and resolve only at the executing site.

## Validation

- broad Recipe, PyTorch Recipe, executor, runner, and catalog unit tests — 1,293 passed, 18 skipped, 3 subtests passed
- focused FedEval compatibility tests — 28 passed; broader FedEval/component configuration tests — 37 passed, 5 skipped
- focused `SimEnv` status/result tests — 18 passed, including partial workspaces from nonzero simulator exits and setup failures
- combined prerequisite/example stack unit tests — 135 passed
- stacked real simulation→POC acceptance test — passed in 75.61 seconds with identical final per-site accuracies
- manual `SimEnv` Recipe integration workflow — 1 passed in 18.11 seconds
- `python3 -m nvflare.tool.recipe.generate_recipe_catalog --check` — passed
- `./runtest.sh -s` — passed
- `git diff --check` — passed


nvflare/app_common/executors/client_api_executor.py
```diff
@@ -497,6 +499,18 @@ def _normalize_command(command: Optional[Union[str, list[str]]]) -> Optional[Uni
             raise ValueError("command argv must start with a non-empty executable")
         return list(command)
 
+    @staticmethod
+    def _normalize_script_args(script_args: Union[str, list[str]]) -> Union[str, list[str]]:
+        if isinstance(script_args, str):
+            return script_args
+        if not isinstance(script_args, list):
+            raise ValueError(
```
holgerroth: **`task_script_args=None` was accepted before this PR and is now rejected, including from exported job configs.**

Base stored `None` here untouched (falsy, so it skipped the wrong-mode check) and `TaskScriptRunner` mapped it to an empty argv. Head: `ClientAPIExecutor(execution_mode="in_process", task_script_path="custom/train.py", task_script_args=None)` raises `task_script_args must be a string or list of strings, but got NoneType`. The same happens through the real client path: a job exported by the current release with `ScriptRunner(script_args=None)` writes `"task_script_args": null` (`_values_differ("", None)` is True), and `instantiate_class` on that entry now fails at app deployment. `FedEvalRecipe(eval_args=None)` and per-site `{"eval_args": None}` hit it too (`fedeval.py:203` uses plain `.get`).

It also disagrees with the rest of this PR: `_normalize_command` four lines up accepts `None`, and the widened `TaskScriptRunner.__init__` keeps `script_args is None` legal and `get_sys_argv` keeps an `else: args_list = []` branch that is now dead. Suggest `if script_args is None: return ""` at the top of this helper, or drop `Optional` from the runner if the tightening is intended (then it needs a test and a note).

nvflare/recipe/fedavg.py
```diff
@@ -49,7 +49,7 @@ class _FedAvgValidator(BaseModel):
     min_clients: int
     num_rounds: int
     train_script: str
-    train_args: str
+    train_args: Union[str, list[str]]
```
holgerroth, line 53: **Lax pydantic mode accepts `set`, `tuple`, generators, and `bytes` here; a set scrambles argv.**

`_FedAvgValidator` is not strict, so `FedAvgRecipe(train_args={"--lr", "0.01", "--epochs", "5"})` (a `{}`-for-`[]` typo) is accepted and exported in hash order. Three interpreter runs on the real PT recipe gave `['5','0.01','--epochs','--lr']`, `['--epochs','0.01','5','--lr']`, and the correct order once, so the client sometimes receives `--lr 5` with no warning. A tuple is accepted here too, while `set_per_site_config`, `ClientAPIExecutor`, and this PR's own `test_in_process_rejects_invalid_task_script_args` treat a tuple as invalid.

Note `StrictStr` does not fix this (item strictness does not stop sequence coercion; I checked on pydantic 2.12). What works and still accepts `str` and `list[str]`: `Union[Annotated[str, Strict()], Annotated[list[str], Strict()]]`, or a `@field_validator("train_args", mode="before")` that requires `isinstance(v, (str, list))` and reuses `_normalize_script_args`'s message.

nvflare/recipe/fedavg.py
```diff
@@ -155,7 +156,7 @@ class FedAvgRecipe(Recipe):
             ``set_per_site_config(recipe, config)`` immediately after construction. Each config dict can
             contain optional overrides:
             - train_script (str): Training script path
-            - train_args (str): Script arguments
+            - train_args (str or list[str]): Script arguments
```
holgerroth, line 165: **Per-site list `train_args` aliases the caller's list until export.**

`set_per_site_config` snapshots with `dict(site_config)` (spec.py:353-356, whose comment promises that later mutation cannot change deferred client preparation), and `ScriptRunner.__init__` stores `script_args` by reference (script_runner.py:146); the copy this PR adds happens only in `ClientAPIExecutor` at `_ensure_client_apps_prepared`. Reproduced: `lst = ["--lr", "0.1"]; set_per_site_config(r, {"site-1": {"train_args": lst}}); lst[1] = "0.0001"; lst.append("--leaked"); r.export(d)` exports `['--lr', '0.0001', '--leaked']`. The natural trigger is reusing one list across sites in a loop (`shared[-1] = f"/data/{site}"`), which makes both sites export the last site's path silently. The top-level `train_args` path does not have this problem because pydantic rebuilds the list, so the two entry points behave differently.

The PR adds copy-on-accept and "args are copied" tests at the executor and runner layers; carrying the same copy (or `_normalize_script_args`) into `ScriptRunner.__init__` closes this and the late-validation comment below in one line.

nvflare/recipe/sim_env.py
```diff
@@ -110,34 +111,43 @@ def __init__(
...
+                gpu=self.gpu_config,
+                log_config=self.log_config,
             )
+            if run_status not in (None, 0):
+                raise RuntimeError(
+                    f"Simulation failed with return code {run_status}. "
+                    f"Logs are in per-site subdirectories under {os.path.join(self.workspace_root, job_id)}, "
+                    f"e.g. server/simulate_job/log.txt"
+                )
+        except Exception:
```
holgerroth: **Every failure status recorded here is unreachable through `Run`.**

`Recipe.run` does `job_id = env.deploy(self._job)` and only then `Run(env, job_id)` (spec.py:904-907, the only `Run(...)` construction). Because `deploy` re-raises, no `Run` exists for a failed simulation, so under `SimEnv` `run.get_status()` can only ever return `FINISHED:COMPLETED`; `FINISHED:ABNORMAL` is observable only by calling `env.get_job_status(job.name)` directly, which only the new unit tests do. After a preflight failure `env.get_job_result(name)` still returns `<root>/<name>`, a directory that was never created.

So the description's "makes `Run.get_status()` consistent with `FINISHED:COMPLETED` or `FINISHED:ABNORMAL`" holds for the first value only. Either document that `SimEnv` failures raise (so `get_status()` is COMPLETED or `None`), or gate `get_job_result` on the recorded status. Turning a deploy failure into a terminal `Run` inside `Recipe.run` would also change `PocEnv`/`ProdEnv` and is probably out of scope here.

nvflare/recipe/sim_env.py
```diff
@@ -110,34 +111,43 @@ def __init__(
...
+            self.last_run_failed = True
+            self._job_statuses[job_id] = RunStatus.FINISHED_ABNORMAL.value
```
holgerroth: **`FINISHED:ABNORMAL` for every failure does not match how the server classifies the same outcomes.**

The simulator's exit codes that reach line 139: `1` (setup failed, `simulator_runner.py:536-537`), `2` (exception or `FATAL_SYSTEM_ERROR`, `:532`/`:543`), `101-103` (`mpm.py:154-164`), `247` (in-simulator Ctrl-C). The server's `_classify_finished_job_status` (`job_runner.py:554-563`) maps all of those to `FINISHED:EXECUTION_EXCEPTION`, uses `FAILED_TO_RUN` for a job that never started (`job_runner.py:720`), and reserves `ABNORMAL` for `INFRASTRUCTURE_ERROR` (104, never produced by the simulator) and `-9`. Only the `-9` case matches. The new test pins the never-started case as `ABNORMAL` (`sim_env_test.py:214-221`) while asserting `simulator_run` was never called.

The integer is already in hand at line 139. Minimal mapping: non-zero rc to `FINISHED:EXECUTION_EXCEPTION` (except `-9` to `ABNORMAL`), and exceptions before or around `simulator_run` to `FAILED_TO_RUN`. Low impact today because of the reachability point above; it matters as soon as that is fixed.

nvflare/app_opt/pt/recipes/fedavg.py
```diff
@@ -117,7 +117,7 @@ def __init__(
         min_clients: int,
         num_rounds: int = 2,
         train_script: str,
-        train_args: str = "",
+        train_args: Union[str, list[str]] = "",
```
holgerroth, line 120: **The widening reaches 7 more recipes silently and misses 8 others, so `train_args` now has three behaviors.**

Because the shared `_FedAvgValidator` is widened, `NumpyFedAvgRecipe`, `FedProxRecipe`, `FedCERecipe`, `SklearnFedAvgRecipe`, `KMeansFedAvgRecipe`, `SVMFedAvgRecipe`, and the TF `FedAvgRecipe` accept a list and export a JSON array while their signatures, docstrings, and `recipe_catalog.json` entries still say `str` (verified with `train_args=["--a", "b c"]`). `FedOptRecipe`, `ScaffoldRecipe`, `FedAvgRecipeWithHE`, `CyclicRecipe`, the TF `FedOpt`/`Scaffold`, `FedAvgLrRecipe`, and `FedTaskRecipe` (`task_args`) keep their own `str` validators and reject with `Input should be a valid string`. `FedEvalRecipe.eval_args` has no validator at all. The catalog test only proves catalog equals annotation, so it cannot catch the first group.

Suggest widening the pass-through signatures and docstrings, regenerating the catalog, and declaring the type once: `_CommandArg = Union[str, list[str]]` already exists in `nvflare/job_config/script_runner.py:27`, which this module imports from.

nvflare/job_config/script_runner.py
```diff
@@ -98,7 +98,7 @@ def __init__(
         Args:
             script: Training script path.
             script_args: Arguments appended to the script. Pre-tokenized argv preserves
-                exact argument boundaries for external processes.
+                exact argument boundaries in both execution modes.
```
holgerroth, line 99: **`ScriptRunner` still stores `script_args` unvalidated (line 146), so per-site type errors surface late with the wrong name.**

`_apply_per_site_config` builds a `ScriptRunner` per site specifically to "validate every runner override while `set_per_site_config()` is still recoverable", and that works for other parameters (a per-site `{"framework": "not_a_framework"}` raises immediately). A per-site `{"train_args": ["--epochs", 3]}`, a tuple, or an int is accepted and only fails at `export()`/`run()` with `task_script_args argv must contain only strings`, naming neither `train_args` nor the site. A tuple with a `${secret:...}` element additionally gets an `UnsupportedSecretRefWarning` saying the reference "will remain a literal placeholder" (the exemption at `secret_utils.py:710` matches `list` only) before the build aborts on the tuple, pointing the user at the wrong problem.

One line at 146, `self._script_args = ClientAPIExecutor._normalize_script_args(script_args)` (the module already imports `ClientAPIExecutor`), fixes the timing, the message, the misleading warning, and the aliasing noted on `recipe/fedavg.py:159`.

tests/unit_test/recipe/fedavg_recipe_test.py
```diff
@@ -239,6 +239,31 @@ def test_external_command_secret_ref_is_supported(self, mock_file_system, base_r
                 },
             )
 
+    def test_pre_tokenized_train_args_with_secret_refs_are_supported(
```
holgerroth, line 252: **Nothing tests a list `task_script_args` across the export/reload boundary.**

This test and the runner/executor tests stop at in-memory objects. Every export or reload assertion in `tests/` uses the string form (`fedavg_recipe_test.py:1430`, `json_configer_test.py:238/246`, `recipe_k8s_test.py:135`, `script_runner_test.py:583`). The list path works today because `FedJobConfig._get_args` emits lists via `type(v).__name__ in dir(builtins)` and `_EnvUpdater.process_element` happens to handle list parents for `{SITE_NAME}` expansion (`wfconf.py:61-72`); a refactor of either would stringify the list or leave the placeholder unexpanded with no test failing. A list variant of `json_configer_test.py:223-254` plus a list assertion next to line 1430 would pin the headline behavior; I ran that path manually and it passes.

nvflare/fuel/utils/secret_utils.py
```diff
@@ -707,10 +707,11 @@ def _contains_secret_ref_outside_keys(
                 supported_value_depth == 1
                 and isinstance(key, str)
                 and key in supported_value_keys
-                and isinstance(item, str)
+                and (isinstance(item, str) or (isinstance(item, list) and all(isinstance(arg, str) for arg in item)))
```
holgerroth, line 710: **This exemption now applies to every supported key, so per-site `command` lists are blessed while the constructor still rejects them.**

With `_SUPPORTED_PER_SITE_SECRET_REF_KEYS = {"command", "train_args"}` (and `{"command", "eval_args"}` in `fedeval.py`), `set_per_site_config(r, {"site-1": {"command": ["python3", "-u", "${secret:TOK}"]}})` no longer warns (base did) and exports `["python3", "-u", "${secret:TOK}", "custom/cli.py"]`; that is correct, since `prepare_subprocess_command` resolves per token. But `FedAvgRecipe(..., command=["python3", "-u"])` still fails pydantic (`command: str` at recipe/fedavg.py:58 and :225), the per-site docstring at recipe/fedavg.py:161 still says `command (str)`, and the two new secret-ref tests cover `train_args` only. Either widen the top-level `command` consistently or document the per-site-only support and add a list-`command` case.

nvflare/recipe/sim_env.py
```diff
@@ -110,34 +111,43 @@ def __init__(
...
-            gpu=self.gpu_config,
-            log_config=self.log_config,
-        )
-        self.last_run_failed = run_status not in (None, 0)
-        if self.last_run_failed:
-            raise RuntimeError(
-                f"Simulation failed with return code {run_status}. "
-                f"Logs are in per-site subdirectories under {os.path.join(self.workspace_root, job_id)}, "
-                f"e.g. server/simulate_job/log.txt"
+        self._job_statuses.pop(job_id, None)
```
holgerroth, line 118: **`except Exception` misses `KeyboardInterrupt`/`SystemExit`, and this `pop` exists only for that gap.**

In `fed_job_config.simulator_run`, `generate_job_config` (:266) and the secret scan (:267) run before the `try/except KeyboardInterrupt` (:269), and a second Ctrl-C inside that handler (:295-298) also propagates. A Ctrl-C there passes line 145 unrecorded; the `pop` here has already erased the previous status, so `get_job_status` returns `None` while `last_run_failed` still holds the previous run's value (reproduced: `None`/`True` after a prior failure, `None`/`False` after `SystemExit`). The repo's precedent for record-then-reraise cleanup is `except BaseException` (`in_process_backend.py:145-149` and about twenty other handlers). Widening makes this `pop` dead and lets one `_record_status(job_id, failed)` helper replace the three lockstep writes at 113, 146, 149; `last_run_failed` has no readers outside `sim_env_test.py` and can become a derived property.

nvflare/app_common/executors/client_api_executor.py
```diff
@@ -497,6 +499,18 @@ def _normalize_command(command: Optional[Union[str, list[str]]]) -> Optional[Uni
             raise ValueError("command argv must start with a non-empty executable")
         return list(command)
 
+    @staticmethod
+    def _normalize_script_args(script_args: Union[str, list[str]]) -> Union[str, list[str]]:
```
holgerroth: **This is the fourth hand-written copy of the str-or-`list[str]` argv check, and the copies disagree.**

`_normalize_script_args` rejects `None` and tuples; `_normalize_command` above accepts `None`; `TaskScriptRunner.__init__` (task_script_runner.py:86-95) accepts `None` and rejects tuples; `_to_external_process_argv` (job_config/script_runner.py:31-39) rejects both but shlex-splits strings; `prepare_subprocess_command` (utils/process_utils.py:141-147) accepts tuples and raises `TypeError` for `None`. Three different message prefixes for "argv must contain only strings". The `None` regression above is a direct result of this divergence. A single `normalize_argv(value, arg_name, allow_none=...)` in `nvflare/fuel/utils` would collapse them (it cannot live in `script_runner`, which imports this module), and `_CommandArg` could move there too instead of `Union[str, list[str]]` being re-spelled in five modules.

holgerroth (commented):
I reviewed this at head `67aceb6` (every changed file plus the callers of `ScriptRunner`, `ClientAPIExecutor`, `TaskScriptRunner`, and `SimEnv.get_job_status` across `nvflare/`, `examples/`, `research/`, `skills/`, and `tests/`). Locally: the 417 tests in the touched files pass, black/isort/flake8 and the license check are clean, and a list `train_args` round-trips exactly through export, JSON reload, and `TaskScriptRunner` (spaces, quotes, apostrophes, backslashes, empty element, per-site override). Things I checked and found fine: per-element secret-reference resolution in both execution modes, the string path being byte-identical to the base, and the simulator exiting non-zero on client and server failures, so the new `FINISHED:COMPLETED` is not a false positive.

The inline comments carry the findings that sit on changed lines. Four more are in files outside the diff:

1. `tests/integration_test/slow/recipe_system_test.py:61` still asserts `run.get_status() is None`; with this PR `Run.get_status()` returns `FINISHED:COMPLETED` before `get_result()`, so the slow suite fails as documented (`pytest slow/recipe_system_test.py -v`). Same stale expectation in the skipped class at `tests/unit_test/recipe/run_test.py:287-288`.
2. `research/auto-fl-research/tasks/cifar10/job.py:336` and `tasks/vlm_med/job.py:258` build `train_args` with `" ".join(shlex.quote(...))` for recipes that default to in-process, where the runner splits with `posix=False` and never unquotes. Reproduced: `['--data_root', '/data/cifar cache', '--tag', 'run one']` arrives as `['--data_root', "'/data/cifar", "cache'", '--tag', "'run", "one'"]`. This PR adds the exact fix (pass the list); those call sites, plus the `" ".join` + whitespace `SystemExit` in `examples/devops/multicloud/e2e/verify.py:164-176` and `examples/advanced/recipe-k8s/job.py:177`, could switch to it.
3. The repo's own skills still tell agents to fail closed on whitespace values for in-process `train_args` (`skills/nvflare-shared/references/pytorch-family-recipe-construction.md:63-68`, `skills/nvflare-convert-huggingface/references/huggingface-validation.md:38-41`, and `assets/job.py:64` raises on any whitespace). That is the motivating use case of this PR, and it stays unreachable through the guided path; related prose is pinned by `seed_skills_test.py:378`, so it is a coordinated update.
4. `docs/` has no mention of the list form (all 16 `train_args` occurrences in `docs/*.rst` are strings; `recipe_api.rst:237-241, 367-372` documents only the tokenize-then-resolve contract), and `nvflare/recipe/secrets.py:38-42` still says references resolve "after argument tokenization", which is not how the list path works.

On stacking: #5244 and #5255 were closed at 16:43Z when the `stack/` branches were removed, and #5244's last head still carried its own copies of these `nvflare/` changes. I assume both will be reopened against `main` once this lands; when they are, please make sure they consume this PR rather than re-adding the tokenizer and `SimEnv` code, so it does not land twice.

chesterxgchen:
Addressed the latest review feedback in `67aceb6e4`:

- regenerated `recipe_catalog.json`; the unified and PyTorch FedAvg `train_args` entries now advertise `Union[str, list[str]]`
- expanded the synchronous `SimEnv.deploy()` failure boundary to include local-script validation and component-policy/workspace setup, so every deployment exception records `FINISHED:ABNORMAL` and sets `last_run_failed`
- added regression coverage for setup failures and reran the catalog freshness tests

Validation: 417 focused tests plus 3 subtests passed; the 56 recipe CLI/SimEnv tests passed; `./runtest.sh -s` and `git diff --check` passed.

nvflare/app_common/executors/client_api_executor.py
chesterxgchen: Fixed in e8fa1abfd. `task_script_args=None` is accepted again and normalized to empty arguments, including when an older exported config contains JSON `null`. Added a constructor regression test.

nvflare/recipe/sim_env.py
chesterxgchen: Clarified in e8fa1abfd and the PR description. A successful deploy returns a `Run` with `FINISHED:COMPLETED`; deployment failures re-raise before a `Run` exists, so failure status is available through `SimEnv.get_job_status()` directly. `get_job_result()` now returns `None` when preflight did not create a workspace.

nvflare/recipe/fedavg.py
chesterxgchen: Fixed in e8fa1abfd. A Pydantic before-validator now applies the same strict str-or-list contract before coercion, so tuple, set, bytes, and non-string argv elements are rejected deterministically. Added coverage for all of those cases.

nvflare/recipe/fedavg.py
chesterxgchen: Fixed in e8fa1abfd. Per-site `train_args` and `command` argv are normalized and copied while `set_per_site_config()` is still applying the configuration. A mutation regression test confirms later caller changes cannot leak into export.

nvflare/recipe/sim_env.py
chesterxgchen: Fixed in e8fa1abfd. Exceptions before or during simulator startup now record `FINISHED:FAILED_TO_RUN`; ordinary nonzero simulator exits record `FINISHED:EXECUTION_EXCEPTION`; `-9` records `FINISHED:ABNORMAL`. Parameterized tests cover the return-code mapping and setup exceptions.

nvflare/app_opt/pt/recipes/fedavg.py
chesterxgchen: Fixed in e8fa1abfd. The unified, NumPy, PyTorch, TensorFlow, scikit-learn, FedProx, FedCE, KMeans, and SVM FedAvg wrappers now all expose `Union[str, list[str]]` for both `train_args` and `command`. The generated catalog was regenerated and its freshness test passes. Recipes with independent validators remain unchanged.

tests/unit_test/recipe/fedavg_recipe_test.py
chesterxgchen: Fixed in e8fa1abfd. The Recipe export test now asserts list-valued `task_script_args` in `config_fed_client.json`, and a JSON configurator/reload test verifies that list shape and a `{SITE_NAME}` element survive reload while a spaced path stays one element.

nvflare/job_config/script_runner.py
chesterxgchen: Fixed in e8fa1abfd. `ScriptRunner` now normalizes and copies argv in its constructor, and the FedAvg per-site boundary validates `train_args`/`command` with their public field names during `set_per_site_config()`. Invalid elements now fail immediately with a `train_args argv...` error; regression coverage verifies both timing and message.

nvflare/app_common/executors/client_api_executor.py
chesterxgchen: Fixed in e8fa1abfd. Added one shared `normalize_argv()` contract and reused it in `ClientAPIExecutor`, `TaskScriptRunner`, `ScriptRunner`, external-process argv preparation, and FedAvg Recipe validation. Mode-specific code now only adds its intentional empty-command or legacy-string tokenization semantics.

nvflare/fuel/utils/secret_utils.py
chesterxgchen: Fixed in e8fa1abfd. Top-level and per-site `command` now consistently accept string or pre-tokenized argv across every affected FedAvg wrapper. Tests cover a top-level command list, tuple rejection, and a per-site command argv containing a secret reference with warnings treated as errors.

nvflare/recipe/sim_env.py
chesterxgchen: Fixed in e8fa1abfd. The deployment boundary now catches `BaseException`, records status through one `_record_status()` helper, and resets `last_run_failed` before each attempt. Tests cover both `KeyboardInterrupt` and `SystemExit` in addition to ordinary exceptions.

nvflare/app_common/executors/client_api_executor.py
holgerroth: Verified at `e8fa1ab`: `None` is accepted in all three modes and via JSON `null`, normalizes to `""`, and the key is omitted on export; `FedEvalRecipe(eval_args=None)` works too. Thanks.

nvflare/recipe/fedavg.py
holgerroth: Verified at `e8fa1ab`: the before-validator rejects tuple, set, generator, bytes, and non-string elements for both `train_args` and `command` with field-named messages; `str` and `list` are still accepted, and subclasses are normalized to plain types.

nvflare/recipe/fedavg.py
holgerroth: Verified at `e8fa1ab` for `FedAvgRecipe` and `ScriptRunner`: exports keep the original values after the caller mutates the list. One residual, noted separately on the per-site block: `FedEvalRecipe.eval_args` still aliases the caller's list and has no validator.

nvflare/recipe/sim_env.py
holgerroth: Verified at `e8fa1ab`: `get_job_result` returns `None` when no workspace exists, and the raise-before-`Run` contract is now stated in the description. Fine by me as the trade-off; the alternative would have to change how `deploy` fails.

nvflare/recipe/sim_env.py
holgerroth: Verified at `e8fa1ab`: rc 1/2/101/102/103/247 record `EXECUTION_EXCEPTION`, -9 records `ABNORMAL`, and setup/startup exceptions record `FAILED_TO_RUN`. That matches `job_runner`'s classification; see my note under the Greptile comment below.

nvflare/app_opt/pt/recipes/fedavg.py
holgerroth: Verified at `e8fa1ab`: nine FedAvg-family recipes now declare `Union[str, list[str]]` for `train_args` and `command`, and the regenerated catalog passes its freshness test. Two undeclared cases remain (`FedEvalRecipe.eval_args` and `SwarmLearningRecipe.command` accept lists but are still typed and catalogued `str`); noted separately.

nvflare/job_config/script_runner.py
holgerroth: Verified at `e8fa1ab`: bad per-site `train_args`/`command` now fail inside `set_per_site_config` with field-named messages, and `ScriptRunner` copies. Tiny residual: a per-site tuple containing a secret reference still prints the spurious `UnsupportedSecretRefWarning` before the `ValueError`, because `Recipe.set_per_site_config` scans before `_apply_per_site_config` and the exemption excludes tuples. Console noise only.

tests/unit_test/recipe/fedavg_recipe_test.py
holgerroth: Verified at `e8fa1ab`: the export test asserts the list in `config_fed_client.json` for both modes, and the `JsonConfigurator` test covers list shape, a `{SITE_NAME}` element, and a spaced path. Both pass, and the full chain into `TaskScriptRunner` holds when I run it by hand.

nvflare/fuel/utils/secret_utils.py
holgerroth: Verified at `e8fa1ab`: top-level `command` accepts a list, per-site list and string commands export identically, the per-site docstring is updated, and the secret-ref test is parametrized over both forms.

nvflare/recipe/sim_env.py
holgerroth: Verified at `e8fa1ab`: `BaseException` is re-raised unchanged (`KeyboardInterrupt`, `SystemExit`, `GeneratorExit` checked), `_record_status` keeps both fields in sync, and `last_run_failed` is reset per attempt. The `pop` at line 118 is now redundant but harmless.

nvflare/app_common/executors/client_api_executor.py
holgerroth: Verified at `e8fa1ab`: `normalize_argv` is used at every boundary I listed and `CommandArg` replaced the inline unions in the transport modules. Remaining: `process_utils.prepare_subprocess_command` keeps its own check (it accepts tuples), and `argv_utils.py` has no unit test of its own; both fine as follow-ups.

nvflare/recipe/sim_env.py
```diff
@@ -110,36 +111,50 @@ def __init__(
...
+            run_status = job.simulator_run(
+                workspace=workspace,
+                n_clients=self.num_clients if self.clients is None else None,
+                clients=self.clients,
+                threads=self.num_threads,
+                gpu=self.gpu_config,
+                log_config=self.log_config,
+            )
+        except BaseException:
+            self._record_status(job_id, RunStatus.FAILED_TO_RUN)
```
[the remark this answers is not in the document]
holgerroth: I don't think this one holds. The PR description promises `FAILED_TO_RUN` for setup/startup exceptions and `ABNORMAL` only for a -9 exit, and that mirrors the server: `job_runner.py` sets `RunStatus.FAILED_TO_RUN` when `_start_run` raises (around line 720) and reserves `FINISHED_ABNORMAL` for `INFRASTRUCTURE_ERROR`, -9, and orphaned RUNNING jobs (around lines 550-571 and 770). A Ctrl-C during a running simulation reaches this code as -9 (`fed_job_config.py` catches the `KeyboardInterrupt`) and is recorded `ABNORMAL`. So the current mapping is the server-aligned one.

nvflare/recipe/fedavg.py
```diff
@@ -427,6 +434,11 @@ def _apply_per_site_config(self, config: Dict[str, Dict]) -> None:
         # Validate every runner override while set_per_site_config() is still
         # recoverable; actual client apps are materialized later.
         for site_config in config.values():
+            # Preserve the caller-mutation guarantee for mutable argv overrides.
+            if site_config.get("train_args") is not None:
+                site_config["train_args"] = normalize_argv(site_config["train_args"], "train_args")
```
holgerroth, line 439: **`FedEvalRecipe` (and `SwarmLearningRecipe.command`) did not get this treatment.**

`FedEvalRecipe.eval_args` has no validator and `_prepare_client_apps` builds its runners lazily, so both the top-level and the per-site `eval_args` list still alias the caller: `lst = ['--data', '/orig']; FedEvalRecipe(..., eval_args=lst); lst[-1] = '/MUTATED'; export` gives `['--data', '/MUTATED']`, and `eval_args=('--a', '1')` is accepted at construction and fails only at export. Both `FedEvalRecipe.eval_args` and `SwarmLearningRecipe.command` also accept lists end to end while still annotated and catalogued `str`, so the catalog understates them. Suggest routing them through the same `normalize_argv` (validator plus per-site copy) and widening their annotations, or explicitly documenting them as string-only.

tests/unit_test/recipe/fedavg_recipe_test.py
```diff
@@ -233,12 +242,71 @@ def test_external_command_secret_ref_is_supported(self, mock_file_system, base_r
...
                     "site-2": {},
                 },
             )
 
+    def test_pre_tokenized_train_args_with_secret_refs_are_supported(
+        self, mock_file_system, base_recipe_params, simple_model
+    ):
+        params = {**base_recipe_params, "train_args": ["--api-key", "${secret:DEFAULT_API_KEY}"]}
+        with warnings.catch_warnings():
+            warnings.simplefilter("error", UnsupportedSecretRefWarning)
```
holgerroth: **These two guards cannot fail.**

`warnings.simplefilter("error", UnsupportedSecretRefWarning)` never raises here (same at line 234), because `_emit_safe_warning` in `secret_utils.py` catches the warning-as-error and re-emits it. I confirmed with a key that definitely warns: the function returns `True`, the warning prints, and nothing is raised, so both tests would still pass with the list exemption in `secret_utils.py` reverted. Using `with warnings.catch_warnings(record=True) as w: warnings.simplefilter("always")` and asserting that no `UnsupportedSecretRefWarning` is in `w` makes them real.

nvflare/app_common/executors/task_script_runner.py
```diff
@@ -140,18 +149,19 @@ def release_runtime(self):
...
 
     def get_sys_argv(self):
-        # Preserve the runner's legacy whitespace splitting for existing arguments. Only quoted
-        # spans containing a secret ref are grouped, so a composite such as
-        # "Bearer ${secret:TOKEN}" remains one argument without changing unrelated backslashes.
-        args_list = (
-            []
-            if not self.script_args
-            else split_command_preserving_secret_refs(
+        # Keep legacy strings on their historical whitespace-only path. New
```
holgerroth: **This comment understates the legacy path.**

The call below still passes `group_secret_ref_quotes="${secret:" in self.script_args`, so a quoted span containing a secret reference is grouped into one argument (`--auth "Bearer ${secret:T}"` arrives as one arg), which is not whitespace-only. The base comment explained exactly that and was removed here; worth keeping both sentences (the argv rationale and the secret-ref grouping exception) so nobody later 'fixes' the grouping away.

holgerroth (commented):
Re-verified the fixes at `e8fa1ab` locally: 587 tests across the touched files pass, black/isort/flake8 are clean, the argv round trip through export, JSON reload, and the runner holds, and #5244's commit passes its unit tests, its acceptance test, and `python job.py` on top of this head. Eleven of my fifteen items are resolved; the four others (the slow-suite assertion, the research callers, the skills guidance, and the docs) are outside this PR and can be follow-ups. Two new points from the fix code are inline below, plus one small comment fix.

One reminder: `tests/integration_test/slow/recipe_system_test.py:61` still asserts `run.get_status() is None`, which now fails whenever that suite runs; line 75 already expects `FINISHED:COMPLETED` for PocEnv.

nvflare/recipe/sim_env.py
```diff
@@ -178,4 +190,5 @@ def get_job_result(self, job_id: str, timeout: float = 0.0) -> Optional[str]:
         """Get job result workspace path."""
         if self.workspace_root is None:
             raise RuntimeError("Simulation workspace_root is None - SimEnv may not be properly initialized")
-        return os.path.join(self.workspace_root, job_id)
+        result_path = os.path.join(self.workspace_root, job_id)
+        return result_path if os.path.exists(result_path) else None
```
YuanTingHsieh, line 196: [P2] Avoid returning a stale workspace from an earlier attempt

`deploy()` replaces the in-memory status for this job name, but it does not remove an existing workspace. If a previous run created this directory and a later run with the same name fails preflight before `simulator_run`, this existence check returns the previous run’s artifacts while `get_job_status()` reports `FINISHED:FAILED_TO_RUN`. Recipe names and simulator workspace roots are commonly reused, so callers can receive stale models or logs. Please track result availability for the current attempt, or gate this lookup on the current recorded status, and add a regression that succeeds once (or pre-creates the workspace) and then fails preflight.

YuanTingHsieh:
[P2] The manual Recipe integration suite still has the pre-change `SimEnv` expectations.

`tests/integration_test/slow/recipe_system_test.py:61` asserts `run.get_status() is None`, but after a successful synchronous deploy this PR now returns `FINISHED:COMPLETED`, so that test is guaranteed to fail. The skipped `tests/unit_test/recipe/run_test.py:290-292` also assumes `get_job_result()` returns a path for a workspace that does not exist; it now returns `None`. These tests are outside the current CI path, which is why the checks remain green. Please update both expectations with the new status/result contract.

nvflare/recipe/sim_env.py
chesterxgchen: Fixed in `9816eb36a`. `SimEnv.get_job_result()` now returns a workspace only when the current attempt recorded `FINISHED:COMPLETED`; failed and unknown attempts return `None` even if a same-name directory already exists. The preflight regression now creates a stale prior-result directory before forcing validation failure.

nvflare/app_common/executors/task_script_runner.py
chesterxgchen: Fixed in `9816eb36a`. The comment now preserves both parts of the legacy contract: ordinary strings retain historical whitespace tokenization, while quoted spans containing secret references remain grouped so a resolved value stays one argument. It separately explains why new callers should use pre-tokenized argv.

nvflare/recipe/fedavg.py
chesterxgchen: Fixed in `9816eb36a`. `FedEvalRecipe.eval_args` and `command` now publicly accept `Union[str, list[str]]`, validate and copy top-level argv immediately, and validate/copy deferred per-site argv before storing it. `SwarmLearningRecipe.command` now exposes and normalizes the same contract. The generated Recipe catalog advertises all three widened fields, and regression tests cover caller mutation plus invalid tuple rejection.

tests/unit_test/recipe/fedavg_recipe_test.py
chesterxgchen: Fixed in `9816eb36a`. Both tests now record all warnings and explicitly assert that no `UnsupportedSecretRefWarning` was emitted, so reverting the supported argv boundary makes the tests fail even though `_emit_safe_warning()` neutralizes warning-as-error policies.

chesterxgchen:
Addressed the latest review feedback in `9816eb36a`:

- widened and normalized PyTorch `FedEvalRecipe.eval_args`/`command` and `SwarmLearningRecipe.command`, including top-level and deferred per-site copy/validation, then regenerated the Recipe catalog
- changed the secret-reference tests to record warnings and explicitly assert that `UnsupportedSecretRefWarning` is absent
- restored the legacy secret-reference grouping exception in the task-runner comment
- gated `SimEnv` result lookup on the current attempt reaching `FINISHED:COMPLETED`, so a failed same-name rerun cannot return an older workspace
- updated the manual Recipe integration assertion to expect `FINISHED:COMPLETED` and updated the skipped undeployed-`Run` expectation to return no result

Validation: 793 broad affected tests passed with 18 skipped and 3 subtests; 198 post-format focused tests passed with 3 skipped; the manual synchronous simulation test passed in 18.11 seconds; catalog freshness, `./runtest.sh -s`, and `git diff --check` passed.

chesterxgchen:
The current head addresses the failed-simulation result finding.

`SimEnv.get_job_result()` now returns a workspace only when the recorded status is `FINISHED:COMPLETED`; setup failures, ordinary nonzero exits, `-9` abnormal exits, and unknown jobs return `None` even if a partial or stale workspace exists.

Commit `75bd2092a` extends the regression coverage to create a partial workspace before both nonzero simulator exit cases and verifies that neither is exposed as a result. The focused `SimEnv` suite passes (`18 passed`), `./runtest.sh -s` passes, and `git diff --check` is clean.

chesterxgchen:
Rebased onto current `upstream/main` (`8f512c1f1`); the new PR head is `5519856b8`. Post-rebase checks: focused `SimEnv` suite `18 passed`, combined dependency-stack unit set `100 passed`, real simulation→POC acceptance test passed in 72.47 seconds, and `./runtest.sh -s` passed.

nvflare/app_opt/pt/recipes/fedeval.py
```diff
@@ -147,9 +149,9 @@ def __init__(
         self.eval_ckpt = eval_ckpt
         self.min_clients = min_clients
         self.eval_script = eval_script
-        self.eval_args = eval_args
+        self.eval_args = normalize_argv(eval_args, "eval_args")
```
YuanTingHsieh: [P2] Preserve eval_args=None compatibility

Before this PR, FedEvalRecipe(..., eval_args=None) was accepted and ScriptRunner normalized it to empty argv. This direct normalize_argv call uses the default allow_none=False, so the same constructor now raises ValueError. Per-site {"eval_args": None} still bypasses this normalization and works, leaving the top-level and per-site paths inconsistent. Please normalize with allow_none=True and map None to "" as ScriptRunner does, and add a regression test.

chesterxgchen: Fixed in `7666c22e2`. Top-level `eval_args=None` now uses `normalize_argv(..., allow_none=True)` and maps `None` to the same empty-string representation used by `ScriptRunner`, while invalid non-null containers remain rejected. Added a constructor regression test; the focused FedEval suite passes (`28 passed`) and the broader FedEval/component-config set passes (`37 passed, 5 skipped`).

chesterxgchen (commented):
Submitting the completed review-thread replies for the current implementation.

chesterxgchen (commented):
Submitting the completed review-thread reply for the current implementation.

YuanTingHsieh (approved):

Issue #338, opened 2022-03-22
Multi-process executor's assumption


YuanTingHsieh:
https://github.com/NVIDIA/NVFlare/blob/main/nvflare/app_common/executors/multi_process_executor.py#L151

It only works if the prefix "self.get_multi_process_command()" return a command that can accept the "-m nvflare.private.fed.app.client.sub_worker_process"

~~And would treat it as "python -m nvflare.private.fed.app.client.sub_worker_process".~~

I wonder how this apply to other frameworks than PyTorch.

YuanTingHsieh:
Thanks for the reply.

My question should be that if the user has some running commands that are not using Python.

Then "-m XXX" will not work for them.

chesterxgchen:
The underline structure might be changed in the future release. Let me close this discussion

PR #5258, opened 2026-09-03
Stabilize Recipe script packaging and POC lifecycle

chesterxgchen:
## Summary

This extracts the framework-level prerequisites from the advanced Hello PyTorch environment work so PR #5255 can remain scoped to examples, documentation, tests, and CI.

- freeze a locally available script's packaged relative destination when its `ScriptRunner` is registered
- use that same destination in executor configuration and export, even if `sys.path` later changes
- preserve every destination when one source is registered by multiple runners
- keep distinct relative directories for different same-named scripts
- preserve unavailable absolute in-process paths for scripts pre-installed on production clients
- make script/directory resource classification consistent
- provision a fresh POC workspace transactionally while retaining and restoring an existing stopped workspace on preparation, startup, or readiness failure
- require every selected server/client process or Docker container to remain alive and then pass the existing client-registration readiness check before discarding the retained backup
- route Docker start, inspect, and stop operations through the same endpoint selection, including `NVFL_DOCKER_SOCK` overriding a local `DOCKER_HOST` or `DOCKER_CONTEXT`
- fail closed before workspace replacement when Docker service state cannot be determined, while treating an explicit missing container as stopped
- normalize the POC workspace path before deriving its sibling backup and reject project configurations located inside that replaceable workspace
- expose whether the current `PocEnv` invocation began deployment and owns the replacement workspace, so callers can clean up only their own artifacts

## Why

Recipe construction and export can run under different import paths. Deriving the deployed script path twice can therefore configure the executor for one path while copying the file to another. Freezing the mapping once makes the executor and package agree without changing the public Job API.

`PocEnv` reuses the standard local POC path. A failed in-place reprovision could otherwise corrupt or delete logs and results retained from an earlier invocation. The updated lifecycle moves an existing stopped workspace aside, prepares and starts the replacement, waits for every selected service to remain alive, verifies exact client registration through the existing system-status check, and removes the backup only after job submission succeeds. If preparation, launch, readiness, or submission fails, every surviving managed service is stopped before the partial replacement is discarded and the original workspace is restored atomically. If partially started services cannot be stopped, the backup is retained for recovery rather than overwritten.

Docker-backed POC uses container state rather than host PID files. The same Docker CLI environment is applied to start, readiness inspection, and stop operations. For a local `NVFL_DOCKER_SOCK` override, `DOCKER_CONTEXT` is removed because Docker gives it precedence over `DOCKER_HOST`; remote endpoints retain their existing selection behavior.

This PR intentionally contains no Hello PyTorch example changes. PR #5255 consumes these behaviors as an example-only follow-up.

## Compatibility

- Relative script paths keep their existing packaged layout.
- Locally available absolute scripts use a stable `sys.path`-relative destination.
- Absolute in-process scripts unavailable on the authoring machine remain absolute for production clients that pre-install them.
- Multiple identical source/destination registrations are deduplicated; distinct destinations are all preserved.
- The existing FedBPT recipe is updated to register its source under the same import root used by export.
- POC preflight failures do not claim workspace ownership, and failed replacements restore retained workspaces and results even when asynchronous launch returns before the services become healthy.
- Readiness requires both stable service/container liveness and expected client registration; Docker cleanup inspects all managed containers before workspace restoration.
- Docker inspection timeouts, CLI/daemon errors, and unknown responses stop deployment before a retained workspace is moved.
- If inspection or stopping fails at any rollback stage—including the post-stop verification—rollback preserves both the partial active workspace and retained backup, marks the active workspace owned by the invocation, and reports both recovery paths instead of guessing that services are stopped.
- `project_conf_path` must remain outside `poc_workspace`; trailing workspace separators are normalized so backups cannot be nested inside the directory they replace.

## Validation

- broad job-config, Recipe, POC lifecycle, POC CLI, and FedBPT unit set — 817 passed, 20 skipped
- focused POC lifecycle and CLI tests after the final rollback hardening — 98 passed
- combined prerequisite/example stack unit tests — 138 passed
- stacked real-environment acceptance test — simulation followed by local POC passed in 75.61 seconds with identical final per-site accuracies
- `./runtest.sh -s` — passed
- `git diff --check` — passed

### DIFF
--- a/nvflare/recipe/poc_env.py
+++ b/nvflare/recipe/poc_env.py
@@ -28,9 +28,11 @@ from nvflare.job_config.api import FedJob
 from nvflare.recipe.spec import ExecEnv
 from nvflare.recipe.utils import collect_non_local_scripts
 from nvflare.tool.poc.poc_commands import (
+    POC_START_READY_TIMEOUT,
     _clean_poc,
     _start_poc,
     _stop_poc,
+    _wait_for_poc_system_ready,
     get_poc_workspace,
     get_prod_dir,
     is_poc_running,
@@ -42,7 +44,6 @@ from nvflare.tool.poc.service_constants import FlareServiceConstants as SC
 from .session_mgr import SessionManager
 
 STOP_POC_TIMEOUT = 10
-SERVICE_START_TIMEOUT = 3
 DEFAULT_ADMIN_USER = "admin@nvidia.com"
 _WORKSPACE_BACKUP_PREFIX = ".nvflare-recipe-backup-"
 
@@ -222,6 +223,16 @@ class PocEnv(ExecEnv):
                 excluded=[self.username],
                 services_list=[],
             )
+            project_config, service_config = setup_service_config(self.poc_workspace)
+            if not _wait_for_poc_system_ready(
+                self.poc_workspace,
+                project_config,
+                service_config,
+                services_list=[],
+                excluded=[self.username],
+                timeout_in_sec=POC_START_READY_TIMEOUT,
+            ):
+                raise RuntimeError("POC services were started but no server or clients were selected for readiness")
         except BaseException:
             if self._check_poc_running():
                 self.stop(clean_up=False)
@@ -255,9 +266,6 @@ class PocEnv(ExecEnv):
             if workspace_backup:
                 shutil.rmtree(workspace_backup, ignore_errors=True)
         self.logger.info("POC services started successfully")
-
-        # Give services time to start up
-        time.sleep(SERVICE_START_TIMEOUT)
 
         # Submit job using SessionManager
         return self._get_session_manager().submit_job(job)
--- a/nvflare/recipe/poc_env.py
+++ b/nvflare/recipe/poc_env.py
@@ -31,6 +31,7 @@ from nvflare.recipe.utils import collect_non_local_scripts
 from nvflare.tool.poc.poc_commands import (
     POC_START_READY_TIMEOUT,
     _clean_poc,
+    _docker_cli_env,
     _is_live_pid_file,
     _start_poc,
     _stop_poc,
@@ -186,6 +187,7 @@ class PocEnv(ExecEnv):
                 text=True,
                 timeout=5,
                 check=False,
+                env=_docker_cli_env(),
             )
         except (OSError, subprocess.TimeoutExpired):
             return False
--- a/nvflare/recipe/poc_env.py
+++ b/nvflare/recipe/poc_env.py
@@ -14,6 +14,7 @@
 
 import os
 import shutil
+import subprocess
 import threading
 import time
 import uuid
@@ -33,6 +34,7 @@ from nvflare.tool.poc.poc_commands import (
     _is_live_pid_file,
     _start_poc,
     _stop_poc,
+    _wait_for_poc_system_ready,
     get_poc_workspace,
     get_prod_dir,
     is_poc_running,
@@ -175,11 +177,29 @@ class PocEnv(ExecEnv):
         os.replace(backup, self.poc_workspace)
 
     @staticmethod
+    def _is_docker_service_running(service_name: str) -> bool:
+        """Return whether Docker reports the named POC container as running."""
+        try:
+            result = subprocess.run(
+                ["docker", "inspect", "--format", "{{.State.Running}}", service_name],
+                capture_output=True,
+                text=True,
+                timeout=5,
+                check=False,
+            )
+        except (OSError, subprocess.TimeoutExpired):
+            return False
+        return result.returncode == 0 and result.stdout.strip().lower() == "true"
+
+    @staticmethod
     def _running_services(project_config: dict, service_config: dict, poc_workspace: str) -> list[str]:
         """Return managed POC services whose local process is still alive."""
+        service_names = [service_config[SC.FLARE_SERVER], *service_config.get(SC.FLARE_CLIENTS, [])]
+        if service_config.get(SC.IS_DOCKER_RUN):
+            return [name for name in service_names if PocEnv._is_docker_service_running(name)]
+
         project_name = project_config.get("name")
         prod_dir = get_prod_dir(poc_workspace, project_name)
-        service_names = [service_config[SC.FLARE_SERVER], *service_config.get(SC.FLARE_CLIENTS, [])]
         running = []
         for service_name in service_names:
             service_dir = os.path.join(prod_dir, service_name)
@@ -268,9 +288,19 @@ class PocEnv(ExecEnv):
             )
             project_config, service_config = setup_service_config(self.poc_workspace)
             self._wait_for_services_ready(project_config, service_config)
+            if not _wait_for_poc_system_ready(
+                self.poc_workspace,
+                project_config,
+                service_config,
+                services_list=[],
+                excluded=[self.username],
+                timeout_in_sec=POC_START_READY_TIMEOUT,
+            ):
+                raise RuntimeError("POC services were started but no server or clients were selected for readiness")
             # Successful submission proves that the admin connection is ready,
-            # in addition to the local process-health check above. Keep the
-            # retained workspace backup until both checks have passed.
+            # in addition to the process/container and client-registration
+            # checks above. Keep the retained workspace backup until all checks
+            # and submission have passed.
             job_id = self._get_session_manager().submit_job(job)
         except BaseException:
             if self._check_poc_running():
@@ -345,7 +375,9 @@ class PocEnv(ExecEnv):
             # the server exited during startup, stop any surviving local client
             # processes directly so the workspace can be restored safely.
             services_list = []
-            if not is_poc_running(self.poc_workspace, service_config, project_config):
+            if service_config.get(SC.IS_DOCKER_RUN) or not is_poc_running(
+                self.poc_workspace, service_config, project_config
+            ):
                 services_list = self._running_services(project_config, service_config, self.poc_workspace)
             _stop_poc(
                 poc_workspace=self.poc_workspace,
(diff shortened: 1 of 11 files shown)


chesterxgchen:
Review #338 is fixed in `04dc9fc93`.

`PocEnv.deploy()` now waits for the newly launched server and selected clients to pass the existing POC system-readiness check before marking the replacement workspace owned and deleting the retained backup. A readiness timeout or health failure follows the same transactional failure path as provisioning/launch errors: stop the partial POC when necessary, discard the replacement, restore the prior workspace atomically, and re-raise.

Regression coverage verifies both sides of the boundary: the backup still exists while readiness is evaluated, is deleted only after health succeeds, and is restored when asynchronous launch returns but readiness subsequently fails.

Validation: the complete job-config plus POC/FedBPT regression set passed (`180 passed, 2 skipped`); `./runtest.sh -s` and `git diff --check` passed.

chesterxgchen:
The current rebased head `911b260b2` covers the readiness and Docker endpoint findings reported against intermediate commits:

- readiness checks all selected server/client services, using host PID liveness for local processes and `docker inspect` container state for Docker POC
- the existing POC system-status check is still required, so expected client registration is verified in addition to process/container stability
- failure cleanup discovers and stops every surviving managed service/container before restoring the retained workspace
- Docker start, readiness inspection, and stop share `_docker_cli_env()`
- when a local `NVFL_DOCKER_SOCK` override is applied, that helper sets the matching `DOCKER_HOST` and removes `DOCKER_CONTEXT`, whose precedence would otherwise select the wrong daemon

Regression coverage explicitly verifies that `docker inspect` receives the socket-aware environment and that both inspect and stop remove `DOCKER_CONTEXT` for the override combination. The focused POC lifecycle/CLI set passes (`88 passed`), and all ten current unit-test matrix jobs are green; coverage is still running.

chesterxgchen:
Rebased onto current `upstream/main` (`8f512c1f1`); the new PR head is `c18bfd17e`. The Docker endpoint/readiness fixes described above are unchanged. Post-rebase checks: focused POC lifecycle/CLI set `88 passed`, combined dependency-stack unit set `100 passed`, real simulation→POC acceptance test passed in 72.47 seconds, and `./runtest.sh -s` passed.

YuanTingHsieh (dismissed):

YuanTingHsieh:
not about this PR but as follow up

> I’m concerned that this fixes the immediate mismatch by freezing a `sys.path`-derived guess, while making that implicit behavior part of the Job API contract. `sys.path` is mutable and environment-dependent, so the selected destination can still vary across notebooks, launchers, editable installs, and different authoring machines.
>
> Could we separate the local source path from the portable packaged path explicitly? For example:
>
> ```python
> ScriptRunner(
>     source_path=Path(__file__).parent / "client.py",
>     package_path="trainer/client.py",
> )
> ```
>
> `source_path` is resolved on whichever machine builds the job, while `package_path` is the stable path stored in the artifact and used by the executor. A script expected to be preinstalled on production clients should be represented separately, such as `installed_path=...` or `bundle=False`.
>
> We could retain the current `sys.path` inference for backward compatibility, ideally with a warning showing the inferred mapping, but I don’t think `_ext_script_destinations` populated from private reflection should become the long-term mechanism. Local discovery can be convenient; it should not define the portable job contract.

nvflare/recipe/poc_env.py
```diff
@@ -140,6 +148,94 @@ def __init__(
...
+            shutil.rmtree(self.poc_workspace)
+        elif os.path.lexists(self.poc_workspace):
+            os.remove(self.poc_workspace)
+        os.replace(backup, self.poc_workspace)
+
+    @staticmethod
+    def _is_docker_service_running(service_name: str) -> bool:
+        """Return whether Docker reports the named POC container as running."""
+        try:
+            result = subprocess.run(
```
IsaacYangSLA, line 217: **P1**: Treating inspection errors as `False` can make `deploy()` replace a live POC workspace. `TimeoutExpired`, `OSError`, and all nonzero `docker inspect` exits are indistinguishable from a stopped or missing service here. `_check_poc_running()` can therefore return false, causing the stop check to be skipped before `_backup_existing_workspace()` moves the workspace, while the old containers remain running. Could this distinguish stopped/not-found from unknown and fail closed when Docker state cannot be determined? I reproduced the timeout path returning `False`.

nvflare/recipe/poc_env.py
```diff
@@ -140,6 +148,94 @@ def __init__(
...
+    @property
+    def workspace_owned(self) -> bool:
+        """Whether the latest deploy created or replaced the active POC workspace."""
+        return self._workspace_owned
+
+    def _backup_existing_workspace(self) -> Optional[str]:
+        """Move a retained workspace aside before provisioning mutates its path."""
+        if not os.path.exists(self.poc_workspace):
+            return None
+        backup = f"{self.poc_workspace}{_WORKSPACE_BACKUP_PREFIX}{uuid.uuid4().hex}"
```
IsaacYangSLA, line 170: **P2**: If `poc_workspace` ends with `/`, this backup path is inside the directory being moved (`/tmp/poc/` -> `/tmp/poc/.nvflare-recipe-backup-*`). `os.replace()` then fails with `EINVAL` because it is moving a directory into itself. Could the workspace be normalized before constructing this sibling backup path? Reproduced with a trailing-slash workspace.

nvflare/recipe/poc_env.py
```diff
@@ -162,32 +263,81 @@ def deploy(self, job: FedJob) -> str:
...
-            self.stop(clean_up=True)
-
+            # Keep the stopped workspace until fresh provisioning succeeds, so
+            # a failed replacement can restore its prior logs and results.
+            self.stop(clean_up=False)
+            if self._check_poc_running():
+                raise RuntimeError("Existing POC services could not be stopped")
+
+        self._deployment_started = True
+        workspace_backup = self._backup_existing_workspace()
```
IsaacYangSLA, line 322: **P2**: The workspace is moved before `prepare_poc_provision()` receives `project_conf_path` below. If callers place `project_conf_path` inside `poc_workspace`, the source config disappears before provisioning and every redeploy fails; the exception path only restores the workspace afterward. Could this configuration be rejected before the move, or could the config be copied outside the workspace first? The CLI guards this case, but `PocEnv.deploy()` does not.

nvflare/recipe/poc_env.py
chesterxgchen: Fixed in `346998c0c`. Docker inspection now returns `False` only for a valid `false` state or an explicit Docker “no such object/container” response. Timeouts, `OSError`, daemon errors, nonzero unknown failures, and malformed successful output raise `RuntimeError`, so deployment fails before moving the existing workspace. Added timeout, unavailable-CLI, missing-container, daemon-error, and deploy-preservation coverage.

nvflare/recipe/poc_env.py
chesterxgchen: Fixed in `346998c0c`. `PocEnv` normalizes the configured workspace path once during initialization, so a trailing separator cannot place the backup inside the workspace. Regression coverage verifies the normalized value and that the retained backup is created as a sibling directory.

nvflare/recipe/poc_env.py
chesterxgchen: Fixed in `346998c0c`. Deployment now rejects a `project_conf_path` lexically or physically inside the POC workspace during non-mutating preflight, before service checks or workspace replacement. The regression test verifies the configuration and retained workspace remain untouched and lifecycle ownership stays false.

chesterxgchen:
Agreed as a follow-up design direction. This PR keeps the existing `ScriptRunner(script=...)` contract compatible and uses the frozen internal mapping only to make current inference deterministic between configuration and export. A public separation such as authoring `source_path`, portable `package_path`, and an explicit preinstalled/bundle choice should be designed independently with migration and deprecation behavior, rather than added to this safety fix.

chesterxgchen:
Reviews #381 and #383 are fixed in `cad42396a`.

If service-state inspection fails after the prior workspace has already been moved aside, rollback now fails closed: it preserves the partial replacement at the active path, preserves the prior workspace at its generated backup path, sets `workspace_owned=True`, and raises an actionable error naming both locations. It does not attempt restoration or deletion while container state is unknown.

Regression coverage creates both workspaces, injects a rollback-time Docker inspection failure, and verifies ownership, contents, exception chaining, and the reported backup path. Validation: `96` focused POC lifecycle/CLI tests passed; the combined prerequisite/example stack passed `136` unit tests; `./runtest.sh -s` and `git diff --check` passed.

chesterxgchen:
Reviews #384 and #385 are fixed in `723525980`.

The same recovery-oriented handler now covers the entire rollback cleanup branch: the initial liveness check, exceptions from `stop(clean_up=False)`, and the verification check after stop. Any uncertainty preserves the partial active workspace and retained backup, sets `workspace_owned=True`, and raises an error that names both recovery locations.

New tests exercise both previously uncovered paths: Docker-state failure after the initial check returned running, and an exception from `stop()` itself. Validation: `98` focused POC lifecycle/CLI tests passed, the combined dependency/example stack passed `138` unit tests, `./runtest.sh -s` passed, and `git diff --check` is clean.

YuanTingHsieh (dismissed):

chesterxgchen:
Rebased onto current `upstream/main` after PR #5257 merged and force-pushed the updated fork branch.

The only rebase conflict was in the exported in-process recipe assertion. It now preserves both contracts: the packaged relative `task_script_path` from this PR and the pre-tokenized `task_script_args` introduced by #5257.

Post-rebase validation:
- `350 passed, 2 skipped` across the focused job-config, POC lifecycle, recipe, and export tests
- real two-client POC integration workflow passed (`test_end_to_end_poc_workflow`)
- `./runtest.sh -s` passed
- pre-push agent-skill lint passed

IsaacYangSLA (approved):
LGTM.

PR #5260, opened 2026-09-03
[2.9] Publish NVFlare skills with shared catalog bundle

holgerroth:
## Summary

- add the evaluation dataset required to validate and sign the existing `nvflare-shared` bundle
- base the publication change directly on the stable `2.9` branch
- preserve every existing NVFlare skill instruction, reference, asset, test, runtime file, and metadata value unchanged

The complete NVFlare skill set is expected to be installed together so the existing sibling references resolve.

## Validation

- bundled skill validator: passed
- `tests/unit_test/tool/agent_skill_checks/seed_skills_test.py`: 38 passed
- repository agent-skill checks: 0 errors, warnings, or info findings
- `skillevaluator quality-check skills/nvflare-shared -r cli`: passed, 98.2/100
- `skillevaluator validate skills/nvflare-shared --external --no-dedup -r cli`: passed, 10/10 Tier-1 checks
- the seven current non-Auto-FL signatures match their manifests; `nvflare-autofl` has the known stale hash for `references/continuous-campaigns.md`

`./runtest.sh -s` could not install dependencies because the local Homebrew Python 3.14 environment is externally managed. GitHub's style check is authoritative for the PR. Local Tier-3 contract validation could not start because the installed SkillEvaluator lacks its `harbor` dependency; the required NVCARPS run remains authoritative.

## Signing and catalog registration

NVSkills CI currently rejects the `2.9` base because the NVFlare entry in `NVIDIA/nvskills-ci/config/onboarded-repositories.json` allows only `main`. Add `2.9` to `allowed_base_refs`, then comment exactly `/nvskills-ci` again.

Merge only after NVCARPS attaches `skill-card.md`, `BENCHMARK.md`, and `skill.oms.sig` for `nvflare-shared` and refreshes `nvflare-autofl/skill.oms.sig` against the exact `2.9` content.

The follow-up `NVIDIA/skills` registration will use `ref: 2.9`, contain all nine flat NVFlare entries, and state that the complete NVFlare skill set must be installed together.


skills/nvflare-convert-huggingface/SKILL.md
```diff
@@ -43,6 +42,8 @@ workflows: do not merge or automatically chain them, do not route the combinatio
 workflow to run first before generating or running either job. Recommend `nvflare-fed-stats` first only when the
 user's purpose is to understand data distribution; handle conversion later as a separate request.
 
+**Shared prerequisite:** Run sibling `../nvflare-shared/SKILL.md` for `nvflare-convert-huggingface` first. If missing, tell the user to install `nvflare-shared` from `nvidia/skills` in the same agent and scope. Never vendor it or use a private NVCARPS dependency.
```
chesterxgchen: **Shared prerequisite:** Run sibling `../nvflare-shared/SKILL.md`

This instruction is trouble some, it will fail, as the shared skills can't run by itself.

holgerroth: Resolved in 761bec55a. I removed the instruction to run nvflare-shared first from all five dependent skills. They now only confirm that the sibling ../nvflare-shared/SKILL.md path exists; if it is absent, they ask the user to install nvflare-shared from nvidia/skills in the same agent and scope, then stop. Once present, the consuming skill loads the required references/assets directly and continues—nvflare-shared is never invoked as a separate workflow. I also aligned the shared SKILL.md, evals, README, architecture notes, and tests with that behavior.

skills/nvflare-autofl/skill-card.md
```diff
@@ -29,15 +29,14 @@ Mitigation: Review and scan skill before deployment. <br>
...
 
 
 ## Skill Output: <br>
-**Output Type(s):** [Shell commands, Analysis, Configuration instructions, Files] <br>
-**Output Format:** [JSON envelopes with Markdown summaries and TSV ledger] <br>
+**Output Type(s):** [Analysis, Configuration instructions, Shell commands] <br>
+**Output Format:** [JSON envelopes, YAML configuration, TSV ledgers, and Markdown reports] <br>
 **Output Parameters:** [1D] <br>
-**Other Properties Related to Output:** [Persists autofl.yaml, results.tsv, campaign_state.json, candidate manifests, run artifacts, and progress plots] <br>
+**Other Properties Related to Output:** [None] <br>
```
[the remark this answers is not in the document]
holgerroth: Thanks—this observation is correct. The authoritative SKILL.md states that Auto-FL persists autofl.yaml, results.tsv, state, candidate manifests, run artifacts, progress, and reports. However, this skill-card.md was regenerated and signed by NVCARPS, which replaced the previously correct output-properties value with [None]. We should not edit the signed generated artifact manually. We’ll report this as a card-generation issue to the SkillsEval/NVCARPS team and leave the NVFlare behavior unchanged in this PR.

pcnudde (approved):

Issue #5246, opened 2026-09-02
Ephemeral credentials and external CA architecture


pcnudde:
## Goal

Make external CA integration and ephemeral credentials the recommended production security posture for NVFlare while retaining the existing FLARE-generated startup-kit model for POCs, development, and smaller deployments.

The work is split into five independently adoptable features. All features retain the existing runtime credential contract: NVFlare reads credentials from established startup paths and validates them against `rootCA.pem`.

## Design principles

- Support incremental adoption by project, site, user, or federation.
- Keep endpoint issuance issuer-neutral. step-ca is the documented reference issuer, but NVFlare consumes standard certificate files rather than depending on the issuer at runtime.
- Keep the per-job issuer internal to the NVFlare server so jobs do not require an external CA at launch time.
- Allow baseline kit credentials, manually approved credentials, and automated enrollment to coexist under a common trust anchor.
- Separate frequent certificate renewal from less frequent, restart-bound private-key rotation.
- Fail closed when a feature-enabled deployment cannot obtain, validate, or deliver the required stronger credential; do not silently downgrade to a site credential.

## Tracked features

- [ ] #5247
- [ ] #5249
- [ ] #5248
- [ ] #5251
- [ ] #5250

## Expected outcome

A production deployment can keep its root key outside NVFlare, generate endpoint keys where they are used, renew short-lived endpoint certificates without interrupting running jobs, scope job credentials to a single run, and derive authorized study access from SSO-backed identity and signed entitlements.

PR #5263, opened 2026-09-03
per-job certificates for job processes

### LINKED ISSUE #5249: Per-job credentials
Parent overview: #5246

## Goal

Give every participating site in a job a short-lived, job-bound credential so isolated job processes never receive the server or client site's long-lived private key.

## High-level design

- After accepting a job, the server parent uses NVFlare's constrained internal job CA to issue a unique RSA keypair and certificate for the server job process and each participating client job process.
- Each leaf retains `CN=<site>` and carries https URI Subject Alternative Names under `https://nvidia.com/nvflare/v1/`: `job/<job id>` (the job it belongs to) and one `cell/<fqcn>` per cell it may claim (the job cell, with its descendants, and the workspace-transfer bootstrap cell under the owning CP or server FQCN). No private OIDs: an extension under NVIDIA's enterprise arc cannot be allocated, and `2.25` OIDs break Go parsers.
- Cellnet stays job-free. Drivers expose the peer certificate; `CellIdentityResolver` enforces one generic rule at the TLS handshake and the message-crypto boundary: a certificate carrying cell-scope URIs may only claim an FQCN equal to or under one of those cells. The FL layer decides the cells when it issues the credential.
- The server writes the server-job credential locally and sends each client-job credential through that site's authenticated deployment message.
- The trust anchor remains `startup/rootCA.pem`; job processes use `<run dir>/job_cert/job.crt` and `job.key`.

## Requirements

- Provision the constrained `pathlen:0` job CA in the server kit by default, with explicit project opt-out. It carries the marker URI SAN `https://nvidia.com/nvflare/v1/ca/job`, root-signed, so site-scope verification rejects anything it issued.
- Keep the job-CA private key only in the server kit and never expose it to participants or job processes.
- Issue credentials once, with no CSR flow and no renewal. `job_cert_valid_days` defaults to 30 days and defines the maximum supported job duration.
- Withhold site private keys from Docker, Kubernetes, and Apptainer/Pyxis Slurm jobs. In-process launch and Slurm `sandbox: none` remain explicitly non-isolating.
- Keep sub-worker and Client API/CellPipe trainer authentication paths unchanged; they do not receive job credentials.
- Reject job certificates for site registration, administration, server identity, or a different job.
- Fail deployment if a valid job credential cannot be issued or delivered, including when the job CA is near expiry or a participating client cannot consume the deploy header. Never fall back to site credentials.
- Preserve the internal job CA permanently so enabling the feature does not require an external CA at job launch.

## Compatibility boundary

Secure mode (`secure_train=true`) always requires the job CA. A server kit without one (provisioned before this feature, with `enable_job_ca: false`, or by `nvflare package`) fails every job deploy with an explicit reason in the job's deploy detail until the project is re-provisioned and the server kit redeployed; the root CA is reused and client and admin kits do not change. There is no mode in which secure jobs run on site certificates, so `enable_job_ca: false` is only meaningful for non-secure deployments. Server and clients must run the same release.

The distributed `nvflare cert` / `nvflare package` flow has no root key at packaging time and cannot produce the job CA; extending it is addressed separately. An operator with their own PKI can issue the job CA themselves (intermediate CA, `pathlen:0`, marker URI SAN `https://nvidia.com/nvflare/v1/ca/job`, chaining to `rootCA.pem`) and place `job_ca.crt` / `job_ca.key` in the server kit; the openssl recipe is in the user guide (`docs/user_guide/admin_guide/security/per_job_certificates.rst`).

pcnudde:
## Summary

Closes #5249 (parent overview: #5246).

Job processes (SJ/CJ) run on the site's long-lived certificate and private key, so any job-supplied code can read the site key and impersonate the site indefinitely. This PR gives every job process its own short-lived credential, makes it the only credential a secure job can run on, and keeps the site keys out of job processes.

- **Provisioning**: `CertBuilder` generates a job-signing intermediate CA (`job_ca.crt`/`job_ca.key`, `CA:TRUE pathlen:0`, root-signed marker URI SAN) into the server startup kit. On by default; `enable_job_ca: false` opts out. Persisted in `cert.json`, regenerated if expired. A customer-supplied root passed to `nvflare provision` is covered. The distributed `nvflare cert` / `nvflare package` flow is not (no root key at packaging time, so no job CA, so no secure jobs); it is addressed separately.
- **Issuance**: at job deploy the SP issues one leaf per participating site (`CN=<site>`, URI SANs naming the job and the cells it may claim, `notBefore` backdated for clock skew, validity `job_cert_valid_days` from the server's `fed_server.json` or `resources.json`, `--set`, or `NVFLARE_JOB_CERT_VALID_DAYS`, default 30 days, clamped to the CA). The SJ credential is written to the run dir after app deployment; each CJ credential is pushed in its site's deploy message over the existing authenticated CP-SP channel.
- **No fallback in secure mode**: a server kit without a job CA, or with a job CA about to expire, fails the job deploy with the reason in `job_deploy_detail`; the CP rejects a deploy request without a valid job credential; a job cell refuses to start without one; launchers refuse to launch a secure job that has no credential. Non-secure mode and the simulator use no certificates and are unaffected.
- **Job processes**: `ssl_cert`/`ssl_private_key` point at the job credential, so nothing in SJ/CJ refers to the site key. Job cells (and the Kubernetes bootstrap cell) use it for TLS and message-level crypto in both TLS roles, so cellnet's directory-based credential back-fill can never present a site certificate. The startup content-integrity check no longer requires the site key in job processes.
- **Site-scope rejection**: `IdentityVerifier.verify_common_name()` rejects any certificate carrying the job URI and any chain containing the job-CA marker URI, so neither a leaked job leaf key nor a stolen job CA key can register a site, log in as admin, or pass as the server.
- **Child links must be mTLS**: job-cert binding exists only on mTLS connections, so in secure mode the Docker and Slurm launchers refuse a clear-text parent connection (configure the client's internal listener with `stcp` / `mtls`; the shared-file transport is exempt). Kubernetes job pods already use the parent's external `stcp` listener with mTLS. A live test connects real job cells to a site parent's mTLS internal listener and verifies another job's certificate is rejected.
- **Credential lifetime**: the SP destroys the SJ credential before the run directory is archived to the job store, so `download_job` never contains it; the CP destroys the CJ credential as soon as the CJ process exits. The credential is dead at that point (no renewal, no reuse).
- **Certificate URIs, no private OIDs**: certificate attributes are https URI Subject Alternative Names under `https://nvidia.com/nvflare/v1/` (`job/<job id>` and `cell/<fqcn>` on job leaves, `ca/job` on the job CA; `nvflare/fuel/sec/cert_uri.py`). A private extension would need an OID under NVIDIA's enterprise arc, which cannot be allocated, and `2.25` OIDs break Go parsers; a domain-owned URI needs no registry. Readers match the root exactly, ignore other hosts, and fail closed on malformed URIs.
- **Cellnet stays job-free**: TLS drivers expose the authenticated peer certificate (`PEER_CERT`); `CellIdentityResolver` enforces one generic rule, a certificate carrying cell-scope URIs may only claim an FQCN equal to or under one of those cells, at the connection handshake and on the certificate cached for message-level crypto. Which cells a job credential lists is decided where it is issued: the job cell and the workspace-transfer bootstrap cell under the owning CP or server FQCN (which covers clients behind relays). `site-1.<job B>.ws_transfer_<job A>` is job B's cell and is rejected for job A's certificate. Live cross-process mTLS tests cover a server root, a site parent's internal listener, and the bootstrap cell.
- **Launchers never ship `*.key`** (including `job_ca.key`): Docker binds the startup kit file by file; Kubernetes drops keys from the startup Secret and delivers the job credential through the per-pod credential Secret (`NVFLARE_JOB_CERT`/`NVFLARE_JOB_KEY`), installed by the job process before its bootstrap cell starts, with `job_cert/` excluded from workspace bundles and result uploads; Slurm (apptainer/pyxis) binds a keyless staged copy of the kit. The in-process launcher and Slurm `sandbox: none` share the host filesystem and cannot isolate.
- Docs: design `docs/design/per_job_certs_design.md`; user guide `docs/user_guide/admin_guide/security/per_job_certificates.rst` (incl. openssl recipe for an externally issued job CA).

## Compatibility

Existing secure-mode projects must be re-provisioned (the root CA is reused; only `job_ca.*` is added to the server kit) before jobs will deploy. Server and clients must run the same release: an older client would run its job process on site certificates, which secure mode no longer allows. Non-secure mode and the simulator are unchanged.


nvflare/fuel/f3/drivers/aio_conn.py
```diff
@@ -134,6 +134,7 @@ def _get_aio_properties(self) -> dict:
         peer_cert = self.writer.get_extra_info("peercert")
         if peer_cert:
             cn = get_certificate_common_name(peer_cert)
+            add_peer_job_id(conn_props, self.writer.get_extra_info("ssl_object"))
```
nvidianz: This logic breaks our layering. F3 is a generic communication protocol — it shouldn't know about jobs, or even cells.

nvflare/fuel/f3/cellnet/fqcn.py
```diff
@@ -17,6 +17,26 @@
 class FQCN(FQN):
     VALID_PATTERN = "^[A-Za-z0-9_.~-]*$"
 
+    # A job's cells hang directly off the owning site: <owner>.<job_id> (the job cell) and its
+    # descendants, or an auxiliary cell <owner>.<name>_<job_id> (e.g. the workspace-transfer
+    # bootstrap cell). Nothing deeper counts: a cell under another job's cell is that job's.
+    JOB_AUX_SEPARATOR = "_"
```
nvidianz: Job concepts shouldn't leak into the FQCN. Cellnet shouldn't be involved in job cell naming.

nvflare/fuel/f3/drivers/driver_params.py
```diff
@@ -42,6 +42,7 @@ class DriverParams(str, Enum):
     LOCAL_ADDR = "local_addr"
     PEER_ADDR = "peer_addr"
     PEER_CN = "peer_cn"
+    PEER_JOB_ID = "peer_job_id"
```
nvidianz: Driver is a low-level interface — job_id doesn't belong here. It breaks the network architecture.

nvflare/fuel/f3/drivers/grpc_driver.py
```diff
@@ -136,9 +136,11 @@ def Stream(self, request_iterator, context):
             DriverParams.PEER_ADDR.value: context.peer(),
             DriverParams.LOCAL_ADDR.value: get_address(self.server.connector.params),
         }
-        cn_names = context.auth_context().get("x509_common_name")
+        auth_context = context.auth_context()
+        cn_names = auth_context.get("x509_common_name")
         if cn_names:
             conn_props[DriverParams.PEER_CN.value] = cn_names[0].decode("utf-8")
+            add_grpc_peer_job_id(conn_props, auth_context)
```
nvidianz: Job_id should be handled at higher layer.

nvflare/fuel/f3/drivers/net_utils.py
```diff
@@ -21,14 +21,19 @@
...
 from nvflare.apis.fl_constant import ConnectionSecurity
 from nvflare.fuel.f3.comm_error import CommError
 from nvflare.fuel.f3.drivers.driver_params import DriverParams
 from nvflare.fuel.utils.argument_utils import str2bool
+from nvflare.lighter.constants import CertExtensionOID
 from nvflare.security.logging import secure_format_exception
 
 log = logging.getLogger(__name__)
 
+JOB_ID_EXTENSION_OID = x509.ObjectIdentifier(CertExtensionOID.JOB_ID)
```
nvidianz: Job_ID handling should be at higher layer.

nvflare/fuel/f3/drivers/socket_conn.py
```diff
@@ -223,6 +223,7 @@ def _get_socket_properties(self) -> dict:
             cert = self.sock.getpeercert()
             if cert:
                 cn = get_certificate_common_name(cert)
+                add_peer_job_id(conn_props, self.sock)
```
nvidianz: Job_ID handling should be moved at a higher layer.

nvidianz:
This PR adds job_id handling in F3, which is the wrong layer. Here's how the network is currently structured:

SFM (including the transport drivers) is the bottom layer. It sends messages from endpoint to endpoint and has no concept of cells — it cannot forward messages between them.

Cellnet sits above SFM. It introduces FQCNs and knows how to relay messages between non-directly-connected cells using the FQCN hierarchy. It understands dot-separated FQCN components, but it has no concept of jobs. That's intentional — the client/server layer above Cellnet is responsible for incorporating job_id and constructing the correct FQCN so messages are properly relayed.

job_id belongs at that higher layer, not in Cellnet or below. Keeping this separation is important: Cellnet is a general-purpose communication layer that other applications can reuse. Leaking job concepts into it breaks that abstraction.

pcnudde:
Agreed: job_id does not belong in F3. Proposal to fix the layering while keeping the check at the handshake, where it has to happen.

**What stays in F3, job-free**

- Drivers expose the peer certificate as a connection property next to `PEER_CN` (`PEER_CERT`, DER). No extension parsing in drivers; `PEER_JOB_ID` and the `*_peer_job_id` helpers in `net_utils` go away.
- Cellnet gets one generic rule, in its own vocabulary: a peer certificate may carry an "FQCN scope" extension listing the FQCN prefixes it is allowed to claim. `CellIdentityResolver.require_match` rejects a handshake, and `CredentialManager` a cached certificate, whose claimed FQCN is not equal to or under one of those prefixes. Implemented with the existing `FQCN.is_ancestor`; `belongs_to_job`, `job_aux_name` and `resolve_owner` are removed.

**What moves up to `nvflare/private/fed`**

- The issuer sets the scope when it mints a job credential: `[<cp_fqcn>.<job_id>, <cp_fqcn>.ws_transfer_<job_id>]` for a client, `[server.<job_id>, server.ws_transfer_<job_id>]` for the server. Using the registered CP FQCN covers clients behind relays without any owner-prefix logic in cellnet.
- The job-id extension (`...300.1`) stays an FL-layer concept, read only by `identity_utils` (site-scope rejection) and for log messages.
- The bootstrap cell name becomes a shared constant in the FL layer so the issuer and the K8s launcher agree on it.

**Extension**: new non-critical OID `1.3.6.1.4.1.5703.300.3`, UTF-8, newline-separated FQCN prefixes. Cellnet documents it as its own ("certificate FQCN scope"), with no reference to jobs.

**Tests**: the cellnet unit and live tests become scope tests without job vocabulary; the job-layer tests assert the issuer emits the right scope. Runtime behavior is unchanged, so the K8s/Docker/Slurm validation results stay valid.

Net effect: F3 knows that a certificate can be restricted to FQCN subtrees, nothing about jobs. The alternative is an authorizer callback that the FL layer installs per cell, which keeps cellnet free of any extension parsing but relies on every cell creator (SP, CP, relay, bootstrap) remembering to install it; I think that is the worse trade. If this works for you I will rework the PR accordingly.

pcnudde:
@nvidianz See my proposal above. I also want to remove the custom OID that is used in the certificate. It should be fine, but I do not want to go through the registration steps if I can avoid it.

pcnudde:
Amendment to the proposal above: no new OID. In fact, no private OIDs at all.

This PR will do the same and drop all three custom extensions:

| Purpose | Today | URI SAN | Read by |
|---|---|---|---|
| Job binding on the leaf | OID `300.1` | `https://nvidia.com/nvflare/v1/job/<job-id>` | FL layer (site-scope rejection, log messages) |
| Job-CA marker on the intermediate | OID `300.2` | `https://nvidia.com/nvflare/v1/ca/job` | FL layer (reject anything the job CA signed at site scope) |
| Cellnet FQCN scope | proposed `300.3` | `https://nvidia.com/nvflare/v1/cell/<fqcn-prefix>`, one per allowed prefix | cellnet (handshake and certificate cache) |

Reader rules, same as the SSO grants: exact scheme/host/prefix match, URIs on other hosts ignored, malformed URIs under the root rejected. FQCN characters are all URL-unreserved, so prefixes need no encoding, but they are percent-encoded anyway for a uniform fail-closed parser. The `https://nvidia.com/nvflare/` root becomes one shared constant for both branches.

The rest of the proposal stands: cellnet only learns that a certificate may restrict which FQCN subtrees it can claim, the FL layer decides the prefixes when it issues the credential, and F3 carries no job vocabulary. The external-CA recipe gets simpler too: `subjectAltName = URI:https://nvidia.com/nvflare/v1/ca/job` in an openssl config instead of a custom ASN.1 line.

nvidianz:
Thanks. I agree that we should avoid private OIDs. I recommend using a standard field in the certificate subject DN instead of introducing NVFlare URI SAN semantics.

Of the standard DN fields, `OU` (`2.5.4.11`, organizationalUnitName) is the best fit because it is already supported by X.509 tooling, may occur more than once, and does not displace the existing `CN` identity. For example:

- `OU=nvflare-job:<job-id>`
- `OU=nvflare-job-ca`
- `OU=nvflare-cell-scope:<fqcn-prefix>` - repeated once for each allowed prefix

The reader should collect all `OU` attributes, recognize only these exact prefixes, reject malformed or conflicting values, and apply the existing FQCN ancestor check. The values are protected by the certificate signature; this does not rely on sender-controlled message headers. I would keep `CN` exclusively for the current site/cell identity.

For completeness, the common DN fields and their ASN.1 OIDs are:

| DN field | OID | Meaning |
|---|---|---|
| `CN` | `2.5.4.3` | Common Name |
| `C` | `2.5.4.6` | Country |
| `L` | `2.5.4.7` | Locality |
| `ST` | `2.5.4.8` | State or Province |
| `STREET` | `2.5.4.9` | Street Address |
| `O` | `2.5.4.10` | Organization |
| `OU` | `2.5.4.11` | Organizational Unit |
| `serialNumber` | `2.5.4.5` | Serial Number |
| `surname` | `2.5.4.4` | Surname |
| `givenName` | `2.5.4.42` | Given Name |
| `title` | `2.5.4.12` | Title |
| `postalCode` | `2.5.4.17` | Postal Code |
| `emailAddress` | `1.2.840.113549.1.9.1` | Email Address |
| `DC` | `0.9.2342.19200300.100.1.25` | Domain Component |
| `UID` | `0.9.2342.19200300.100.1.1` | User ID |

`serialNumber` or `UID` could represent one job identifier, but neither cleanly carries repeated FQCN scopes and the CA marker. `OU` gives us one consistent, standard, repeatable field for all three purposes without registering a private OID.

pcnudde:
Thanks, agreed on dropping the private OIDs; that is what the current head does. I looked at moving the three values into repeated `OU` attributes and found two problems, one small and one structural.

**1. Issuer-name coupling.** A marker in the job CA's subject DN becomes part of every leaf's issuer name. The issuer today rebuilds the issuer name from the CA's CN; with `OU=nvflare-job-ca` on the CA, `verify_cert_chain` rejects the chain (`candidates exhausted`). Fixable by copying `ca_cert.subject` into each leaf, but it is coupling a SAN marker does not have, because extensions are not part of the name.

**2. The 64-character bound on `OU`.** X.520 caps `organizationalUnitName` at 64 (RFC 5280 Appendix A, `ub-organizational-unit-name`). Job ids are UUIDs, so the cell scopes do not fit:

| Value | Length |
|---|---|
| `nvflare-job:<uuid>` | 48 |
| `nvflare-cell-scope:site-1.ws_transfer_<uuid>` | 74 |
| `nvflare-cell-scope:relay-1.site-1.ws_transfer_<uuid>` | 82 |

`cryptography` does not enforce the bound on creation or validation and Go does not enforce it on parse, so it would work inside NVFlare today, but it violates the spec, and OpenSSL enforces the bound when building names, which is exactly the tooling the external-CA recipe relies on. The only way to stay under 64 is an owner-relative encoding (`OU=nvflare-owner:<cp_fqcn>` plus `OU=nvflare-cell:<child>`), which puts FQCN structure back into the reader; the flat list of allowed FQCNs is what keeps cellnet's check to "equal to or under one of these".

URI SANs have neither issue: no length cap, no effect on the issuer name, and SAN is the extension X.509 defines for carrying additional identities of the subject. Both encodings are protected by the signature, and a prefix convention inside `OU` is as much an NVFlare convention as a URI under our own domain; the reader rules you describe (exact prefixes only, reject malformed or conflicting values, existing FQCN ancestor check, `CN` untouched) are exactly what `nvflare/fuel/sec/cert_uri.py` and `cell_scopes` implement today.

I propose to keep the URI SANs.

nvflare/fuel/f3/drivers/aio_conn.py
```diff
@@ -134,6 +134,7 @@ def _get_aio_properties(self) -> dict:
         peer_cert = self.writer.get_extra_info("peercert")
         if peer_cert:
             cn = get_certificate_common_name(peer_cert)
+            add_peer_cert(conn_props, self.writer.get_extra_info("ssl_object"))
```
nvidianz: Those steps are common to all drivers. Can they be done at higher level?

nvidianz (dismissed):

nvflare/fuel/f3/drivers/aio_conn.py
pcnudde: Yes. Done in 1dbde5a68: `Connection.record_peer` (F3 base class) takes the raw peer certificate as the transport provides it, DER from sockets or PEM from gRPC, and sets `PEER_CERT` and `PEER_CN` once. Each driver hands over the certificate in one line, and the per-driver CN derivation that predated this PR (three conventions, plus the `nvflare.fuel.hci.security` import into the transport layer) is gone. Driver files are now 22 lines smaller than upstream; consumers of `PEER_CN` / `PEER_CERT` are unchanged, including the `"N/A"` convention for TLS without client authentication.

nvflare/fuel/f3/drivers/aio_conn.py
pcnudde: Addressed in 15c4cdb78 and 1dbde5a68: job-id handling was removed from F3 (drivers only hand the peer certificate to `Connection.record_peer`; cellnet only checks the certificate's declared cell scopes). `fqcn.py` is identical to upstream.

nvflare/fuel/f3/cellnet/fqcn.py
pcnudde: Addressed in 15c4cdb78 and 1dbde5a68: job-id handling was removed from F3 (drivers only hand the peer certificate to `Connection.record_peer`; cellnet only checks the certificate's declared cell scopes). `fqcn.py` is identical to upstream.

nvflare/fuel/f3/drivers/driver_params.py
pcnudde: Addressed in 15c4cdb78 and 1dbde5a68: job-id handling was removed from F3 (drivers only hand the peer certificate to `Connection.record_peer`; cellnet only checks the certificate's declared cell scopes). `fqcn.py` is identical to upstream.

nvflare/fuel/f3/drivers/grpc_driver.py
pcnudde: Addressed in 15c4cdb78 and 1dbde5a68: job-id handling was removed from F3 (drivers only hand the peer certificate to `Connection.record_peer`; cellnet only checks the certificate's declared cell scopes). `fqcn.py` is identical to upstream.

nvflare/fuel/f3/drivers/net_utils.py
pcnudde: Addressed in 15c4cdb78 and 1dbde5a68: job-id handling was removed from F3 (drivers only hand the peer certificate to `Connection.record_peer`; cellnet only checks the certificate's declared cell scopes). `fqcn.py` is identical to upstream.

nvflare/fuel/f3/drivers/socket_conn.py
pcnudde: Addressed in 15c4cdb78 and 1dbde5a68: job-id handling was removed from F3 (drivers only hand the peer certificate to `Connection.record_peer`; cellnet only checks the certificate's declared cell scopes). `fqcn.py` is identical to upstream.

nvflare/fuel/f3/drivers/aio_conn.py
pcnudde: addressed

nvidianz (approved):

PR #5269, opened 2026-09-04
Fix Slurm multi-node GPU step allocation

### LINKED ISSUE #5268: Slurm multi-node srun can inherit the parent allocation's GPU count
## Summary

The Slurm job launcher can give a launcher-owned multi-node client step fewer GPUs per node than the client job requested when the NVFlare parent itself runs inside a Slurm allocation.

The client allocation is submitted correctly with `--gres=gpu:<gpus_per_node>`, but the generated multi-node `srun` fan-out does not specify GRES. Slurm propagates `SLURM_*` variables through nested `sbatch` submissions even with `--export=NIL`, and `srun` interprets `SLURM_GPUS_PER_NODE` as an input option. If the parent allocation provides `SLURM_GPUS_PER_NODE=K`, the client step can therefore use `K` rather than its requested `gpus_per_node=N`.

For example, a two-node client requesting four GPUs per node can receive only one GPU per node in its `srun` step when the parent allocation's environment contains `SLURM_GPUS_PER_NODE=1`. The job-level GPU allocation remains correct, but the incorrect step-level allocation affects every rank.

## Affected configuration

The issue requires all of the following:

- The NVFlare parent runs inside a Slurm allocation.
- The parent environment contains `SLURM_GPUS_PER_NODE` (including a site-supplied default on an allocation that did not explicitly request GPUs).
- The client uses launcher-owned multi-node execution with `nodes > 1`, `gpus_per_node > 0`, and `additional_node_command` configured or generated.
- The cluster accounts for GPUs through GRES.

The issue is not observed when the parent runs outside Slurm or when `SLURM_GPUS_PER_NODE` is absent from the parent environment.

## Expected behavior

Every task in the multi-node client step receives the client's requested `gpus_per_node` value, independent of GPU option variables inherited from the parent allocation.

## Proposed change

Pass `--gres=gpu:<gpus_per_node>` explicitly to the launcher-owned multi-node `srun` command when `gpus_per_node` is nonzero, matching the existing client `sbatch` request. Leave zero-GPU jobs unchanged.

## Validation

- Verify the rendered multi-node `srun` command includes the planned GPU count.
- Preserve the zero-GPU rendering path.
- On a GRES-scheduled Slurm cluster, run an inner client requesting `N` GPUs per node from a parent environment containing a different `SLURM_GPUS_PER_NODE=K` and verify the step receives `N` GPUs on every rank.

pcnudde:
## Summary

- pass the planned per-node GPU GRES request explicitly to launcher-owned multi-node `srun`
- prevent inherited Slurm GPU option variables from changing the client step allocation
- leave zero-GPU multi-node jobs unchanged

Fixes #5268

## Testing

- `python -m pytest tests/unit_test/app_opt/job_launcher/slurm_batch_test.py tests/unit_test/app_opt/job_launcher/slurm_config_test.py tests/unit_test/app_opt/job_launcher/slurm_launcher_test.py tests/unit_test/app_opt/job_launcher/slurm_manager_test.py tests/unit_test/app_opt/job_launcher/slurm_scheduler_client_test.py tests/unit_test/job_config/script_runner_test.py tests/unit_test/recipe/fedavg_recipe_test.py` (`339 passed`)
- `./runtest.sh -s`
- `git diff --check`

## Validation limits

- A live GRES-scheduled Slurm cluster reproduction was not run for this PR.

### DIFF
--- a/nvflare/app_opt/job_launcher/slurm/batch.py
+++ b/nvflare/app_opt/job_launcher/slurm/batch.py
@@ -176,7 +176,7 @@ def _apptainer_parts(plan: LaunchPlan, config: SlurmConfig, worker_words: list[s
 
 
 def _multinode_srun_words(plan: LaunchPlan) -> list[str]:
-    return [
+    words = [
         '"${NVFL_SRUN}"',
         shlex.quote(f"--nodes={plan.resources.nodes}"),
         shlex.quote(f"--ntasks={plan.resources.nodes}"),
@@ -188,6 +188,9 @@ def _multinode_srun_words(plan: LaunchPlan) -> list[str]:
         shlex.quote("--wait=0"),
         shlex.quote("--label"),
     ]
+    if plan.resources.gpus_per_node:
+        words.append(shlex.quote(f"--gres=gpu:{plan.resources.gpus_per_node}"))
+    return words
 
 
 def _multinode_parts(plan: LaunchPlan, job_dir: str, config: SlurmConfig) -> tuple[list[str], list[str]]:
--- a/tests/unit_test/app_opt/job_launcher/slurm_batch_test.py
+++ b/tests/unit_test/app_opt/job_launcher/slurm_batch_test.py
@@ -188,8 +188,18 @@ def test_multinode_batch_exports_node_group_contract_and_delegates_to_srun(tmp_p
     assert "--kill-on-bad-exit=1" in command_line
     assert "--wait=0" in command_line
     assert "--label" in command_line
+    assert "--gres=gpu:1" in command_line
     assert f"{job_dir}/node.sh" in command_line
     assert "worker.module" not in command_line
+
+
+def test_multinode_batch_without_gpus_does_not_request_gres(tmp_path):
+    plan = replace(_multinode_plan(tmp_path), resources=JobResources(nodes=2))
+
+    script, _ = _render_batch_script(plan, _job_dir(tmp_path), _config(tmp_path))
+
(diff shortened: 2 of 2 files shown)


YuanTingHsieh:
I found a compatibility issue with this fix on Slurm 26.05: adding `--gres=gpu:N` does not override an inherited `SLURM_GPUS_PER_NODE=K`.

In Slurm 26.05, `srun` reads that environment variable as the `--gpus-per-node` option. Its validation then explicitly rejects combining it with `--gres=gpu:...`, including when `--gpus-per-node` was set through the environment:

`--gpus-per-node and --gres=gpu are mutually exclusive.`

Sources:
- [Environment-to-option mapping](https://github.com/SchedMD/slurm/blob/slurm-26-05-0-1/src/srun/opt.c#L575)
- [GPU option validation](https://github.com/SchedMD/slurm/blob/slurm-26-05-0-1/src/common/slurm_opt.c#L5148-L5166)
- [The override exception applies to `--gpus`, not `--gres`](https://github.com/SchedMD/slurm/blob/slurm-26-05-0-1/src/srun/opt.c#L945-L954)

Consequently, the reported case—an inherited `SLURM_GPUS_PER_NODE=1` and a client requesting four GPUs per node—can fail before the client step starts. Please clear the conflicting inherited variable before the launcher-owned `srun`, and add regression coverage for this inherited-environment case. The current command-string assertions do not exercise Slurm's option validation.

Validation: confirmed by tracing the Slurm 26.05 source; no live Slurm cluster reproduction was performed. This finding does not assert that the same fatal error occurs on every older Slurm release.

PR #400, opened 2022-04-07
Add Job scheduler

YuanTingHsieh:
- Added JobSchedulerSpec, ResourceManagerSpec
- Added implementation

PR #5270, opened 2026-09-04
Isolate Recipe POC workspaces per deployment

chesterxgchen:
## Summary
- provision each `PocEnv` instance's single deployment in a unique Recipe-owned workspace beside the configured reusable CLI POC workspace
- make each `PocEnv` instance one-shot once provisioning begins, while allowing retry after a non-provisioning preflight rejection
- reject overlapping `deploy()` calls on the same instance and serialize `stop()` with deployment startup
- reject Recipe deployment before provisioning when the configured CLI POC has an active server or client
- probe intended server ports before provisioning and recheck the provisioned configuration immediately before startup
- give Docker Recipe deployments unique container and network identities while preserving logical participant aliases and using the same network for spawned job containers
- treat loopback port probes as authoritative for local or unresolved Docker endpoints; verified remote `DOCKER_HOST` and context deployments rely on daemon-side startup/readiness
- remove the host-wide filesystem lock, durable runtime record, shared-workspace backup/restore, and mutable ownership state machine
- remove Recipe-owned workspaces directly without clearing the reusable CLI POC configuration
- keep stop repeatable when service metadata cannot be inspected: attempt shutdown, clear the session manager, and preserve unverified state/workspace
- verify deployment-failure cleanup stopped all services and removed the current run workspace; otherwise preserve it and report its exact path
- update the current 2.10 Recipe documentation and tutorials without modifying the released 2.9 notes

## Why this is simpler and safer
`PocEnv` starts and stops a complete POC deployment for one integrated Recipe run. Sharing the reusable CLI workspace forced Recipe execution to move, back up, restore, and classify ownership of previously retained state. A unique workspace removes that file-state conflict by construction: Recipe provisioning and cleanup affect only the current run directory, while the configured `nvflare poc` workspace and earlier retained Recipe results are not replaced.

For example, when the configured CLI workspace is `/tmp/nvflare/poc`, Recipe execution uses `/tmp/nvflare/poc.recipe-<unique-id>`. `Run.get_result()` keeps its existing default cleanup behavior; callers can pass `clean_up=False` to preserve the run workspace and logs. A later deployment uses a new `PocEnv` instance and therefore a different unique workspace.

Configured server ports remain shared host resources. `PocEnv` checks the intended ports before consuming the one-shot environment, then rechecks the fully provisioned project just before startup. The probe uses `SO_REUSEADDR` to match the actual listeners and avoid rejecting a reusable port solely because prior connections remain in `TIME_WAIT`. For verified remote Docker daemons, local loopback is a different host, so Docker startup and readiness are authoritative. An empty endpoint-discovery result follows the startup script's local-socket default and keeps the local probe enabled.

Docker Recipe deployments no longer reuse participant names as container identities. Each workspace generates unique parent-container and network names, retains logical participant names as network aliases, and passes the selected network into the Docker job launcher. Concurrent deployments therefore cannot observe or stop each other's containers; if they compete for the same published ports, Docker atomically selects a winner and the loser cleans only its own identities and workspace.

This is a prerequisite for the Hello PyTorch example PRs #5244 and #5255.

## Failure behavior
A provisioning failure removes only the current unique workspace because services have not been started. After startup is attempted, cleanup makes one verified attempt to stop the current run and remove its workspace. If service state cannot be established, services remain running, the Docker network cannot be removed, or the directory cannot be removed, `deploy()` reports the original failure, the cleanup failure, and the exact workspace with manual stop/removal guidance.

Public `stop()` remains repeatable when provisioned service metadata is missing or unreadable: it logs the unknown state, still attempts the normal shutdown path, clears the stale session manager, and preserves the workspace and uncertainty flag for a later retry. Deployment-failure cleanup retains its strict post-check and still fails loudly when cleanup cannot be verified.

`PocEnv` is one-shot after provisioning begins. `Run.abort()` aborts the job but does not stop the environment; call `Run.get_result()` or `PocEnv.stop()`, then create a new `PocEnv` before another execution. This 2.10 behavior is documented in the current Recipe API guide rather than modifying the released 2.9 notes.

## Validation
- `python3 -m pytest tests/unit_test/recipe/poc_env_test.py -q` (47 passed)
- `python3 -m pytest tests/unit_test/recipe tests/unit_test/lighter/poc_commands_test.py tests/unit_test/tool/poc tests/unit_test/tool/deploy/deploy_commands_test.py tests/unit_test/app_opt/job_launcher/docker_launcher_test.py -q` (984 passed, 19 skipped)
- `python3 -m pytest tests/integration_test/slow/recipe_system_test.py::TestRecipeSystemIntegration::test_end_to_end_poc_workflow tests/integration_test/slow/recipe_system_test.py::TestRecipeSystemIntegration::test_recipe_poc_rejects_running_cli_poc -q` (2 passed)
- `./runtest.sh -s`
- `./build_doc.sh --html --skip-api` (succeeded; 120 unrelated repository-wide warnings remain)
- `git diff --check`

### DIFF
--- a/docs/user_guide/data_scientist_guide/job_recipe.rst
+++ b/docs/user_guide/data_scientist_guide/job_recipe.rst
@@ -327,7 +327,12 @@ Let's first set the path to the POC environment:
    run.get_status()
    run.get_result()
 
-The result is stored under the directory ``/tmp/nvflare/poc``.
+``PocEnv`` creates a unique Recipe-owned workspace beside the configured path,
+such as ``/tmp/nvflare/poc.recipe-<unique-id>``. The reusable ``nvflare poc``
+CLI workspace at ``/tmp/nvflare/poc`` is not replaced by Recipe provisioning.
+The active path is available as ``env.poc_workspace``. Pass ``clean_up=False``
+to ``run.get_result()`` when you want to retain that workspace and its logs
+after the run.
 
 To use a named study, point ``PocEnv`` to a custom project file that defines ``studies:``:
 
--- a/docs/user_guide/data_scientist_guide/job_recipe.rst
+++ b/docs/user_guide/data_scientist_guide/job_recipe.rst
@@ -337,7 +337,8 @@ names, so only one Recipe-managed POC deployment can be active on a host for a
 user. Also run ``nvflare poc stop`` before starting ``PocEnv`` when the
 configured CLI deployment is active. If the Recipe process exits unexpectedly,
 the next deployment checks the recorded per-run workspace and refuses to start
-while any of its services remain active.
+while any of its services remain active. It also fails closed with manual
+recovery guidance if that recorded workspace is missing or unreadable.
 
 To use a named study, point ``PocEnv`` to a custom project file that defines ``studies:``:
 
--- a/docs/user_guide/data_scientist_guide/job_recipe.rst
+++ b/docs/user_guide/data_scientist_guide/job_recipe.rst
@@ -335,7 +335,9 @@ to ``run.get_result()`` when you want to retain that workspace and its logs
 after the run. POC deployments still share default ports and Docker participant
 names, so only one Recipe-managed POC deployment can be active on a host for a
 user. Also run ``nvflare poc stop`` before starting ``PocEnv`` when the
-configured CLI deployment is active.
+configured CLI deployment is active. If the Recipe process exits unexpectedly,
+the next deployment checks the recorded per-run workspace and refuses to start
+while any of its services remain active.
 
 To use a named study, point ``PocEnv`` to a custom project file that defines ``studies:``:
 
--- a/docs/user_guide/data_scientist_guide/job_recipe.rst
+++ b/docs/user_guide/data_scientist_guide/job_recipe.rst
@@ -332,7 +332,9 @@ such as ``/tmp/nvflare/poc.recipe-<unique-id>``. The reusable ``nvflare poc``
 CLI workspace at ``/tmp/nvflare/poc`` is not replaced by Recipe provisioning.
 The active path is available as ``env.poc_workspace``. Pass ``clean_up=False``
 to ``run.get_result()`` when you want to retain that workspace and its logs
-after the run.
+after the run. The Recipe and CLI deployments still share default ports and
+Docker participant names, so run ``nvflare poc stop`` before starting
+``PocEnv`` when the configured CLI deployment is active.
 
 To use a named study, point ``PocEnv`` to a custom project file that defines ``studies:``:
 
--- a/docs/user_guide/data_scientist_guide/job_recipe.rst
+++ b/docs/user_guide/data_scientist_guide/job_recipe.rst
@@ -332,9 +332,10 @@ such as ``/tmp/nvflare/poc.recipe-<unique-id>``. The reusable ``nvflare poc``
 CLI workspace at ``/tmp/nvflare/poc`` is not replaced by Recipe provisioning.
 The active path is available as ``env.poc_workspace``. Pass ``clean_up=False``
 to ``run.get_result()`` when you want to retain that workspace and its logs
-after the run. The Recipe and CLI deployments still share default ports and
-Docker participant names, so run ``nvflare poc stop`` before starting
-``PocEnv`` when the configured CLI deployment is active.
+after the run. POC deployments still share default ports and Docker participant
+names, so only one Recipe-managed POC deployment can be active on a host for a
+user. Also run ``nvflare poc stop`` before starting ``PocEnv`` when the
+configured CLI deployment is active.
 
 To use a named study, point ``PocEnv`` to a custom project file that defines ``studies:``:
 
--- a/docs/user_guide/data_scientist_guide/job_recipe.rst
+++ b/docs/user_guide/data_scientist_guide/job_recipe.rst
@@ -333,14 +333,13 @@ CLI workspace at ``/tmp/nvflare/poc`` is not replaced by Recipe provisioning.
 The active path is available as ``env.poc_workspace``. Pass ``clean_up=False``
 to ``run.get_result()`` when you want to retain that workspace and its logs
 after the run. Each ``PocEnv`` instance owns one provisioning lifecycle; create
-a new instance for another deployment after provisioning has begun. POC
-deployments still share default ports and Docker participant names, so only one
-Recipe-managed POC deployment can be active on a host for a user. Also run
-``nvflare poc stop`` before starting ``PocEnv`` when the configured CLI
-deployment is active. If the Recipe process exits unexpectedly, the next
-deployment checks the recorded per-run workspace and refuses to start while any
-of its services remain active. It also fails closed with manual recovery
-guidance if that recorded workspace is missing or unreadable.
+a new instance for another deployment after provisioning has begun. Recipe POC
+deployments still compete for their configured server ports, although custom
+projects with distinct ports can run concurrently. Docker Recipe deployments
+use unique per-workspace container and network names. Also run ``nvflare poc
+stop`` before starting ``PocEnv`` when the configured CLI deployment is active.
+See :ref:`recipe_execution_environments` for lifecycle, conflict-checking, and
+failure-cleanup details.
 
 To use a named study, point ``PocEnv`` to a custom project file that defines ``studies:``:
 
--- a/docs/user_guide/data_scientist_guide/job_recipe.rst
+++ b/docs/user_guide/data_scientist_guide/job_recipe.rst
@@ -332,13 +332,15 @@ such as ``/tmp/nvflare/poc.recipe-<unique-id>``. The reusable ``nvflare poc``
 CLI workspace at ``/tmp/nvflare/poc`` is not replaced by Recipe provisioning.
 The active path is available as ``env.poc_workspace``. Pass ``clean_up=False``
 to ``run.get_result()`` when you want to retain that workspace and its logs
-after the run. POC deployments still share default ports and Docker participant
-names, so only one Recipe-managed POC deployment can be active on a host for a
-user. Also run ``nvflare poc stop`` before starting ``PocEnv`` when the
-configured CLI deployment is active. If the Recipe process exits unexpectedly,
-the next deployment checks the recorded per-run workspace and refuses to start
-while any of its services remain active. It also fails closed with manual
-recovery guidance if that recorded workspace is missing or unreadable.
+after the run. Each ``PocEnv`` instance owns one provisioning lifecycle; create
+a new instance for another deployment after provisioning has begun. POC
+deployments still share default ports and Docker participant names, so only one
+Recipe-managed POC deployment can be active on a host for a user. Also run
+``nvflare poc stop`` before starting ``PocEnv`` when the configured CLI
+deployment is active. If the Recipe process exits unexpectedly, the next
+deployment checks the recorded per-run workspace and refuses to start while any
+of its services remain active. It also fails closed with manual recovery
+guidance if that recorded workspace is missing or unreadable.
 
 To use a named study, point ``PocEnv`` to a custom project file that defines ``studies:``:
 
--- a/nvflare/recipe/poc_env.py
+++ b/nvflare/recipe/poc_env.py
@@ -50,6 +50,7 @@ STOP_POC_TIMEOUT = 10
 POC_READY_POLL_INTERVAL = 0.2
 POC_READY_STABLE_INTERVAL = 2.0
 DEFAULT_ADMIN_USER = "admin@nvidia.com"
+_RECIPE_WORKSPACE_SUFFIX = ".recipe-"
 
 
 # Internal — not part of the public API
@@ -156,10 +157,25 @@ class PocEnv(ExecEnv):
 
     def _new_poc_workspace(self) -> str:
         """Return a unique Recipe-owned workspace beside the configured POC path."""
-        return f"{self._poc_workspace_root}.recipe-{uuid.uuid4().hex}"
+        return f"{self._poc_workspace_root}{_RECIPE_WORKSPACE_SUFFIX}{uuid.uuid4().hex}"
+
+    def _is_recipe_workspace(self, workspace: str) -> bool:
+        """Return whether the path has this environment's exact sibling-and-UUID form."""
+        root = os.path.abspath(self._poc_workspace_root)
+        candidate = os.path.abspath(workspace)
+        if os.path.dirname(candidate) != os.path.dirname(root):
+            return False
+        prefix = f"{os.path.basename(root)}{_RECIPE_WORKSPACE_SUFFIX}"
+        candidate_name = os.path.basename(candidate)
+        if not candidate_name.startswith(prefix):
+            return False
+        identifier = candidate_name[len(prefix) :]
+        return len(identifier) == 32 and all(c in "0123456789abcdef" for c in identifier.lower())
 
     def _clean_up_failed_deployment(self) -> None:
         """Stop a failed deployment and verify its per-run workspace was cleaned."""
+        if not self._is_recipe_workspace(self.poc_workspace):
+            raise RuntimeError(f"refusing to clean unmanaged POC workspace {self.poc_workspace}")
         self.stop(clean_up=True)
         if self._check_poc_running():
             raise RuntimeError("POC services remain running")
@@ -306,7 +322,8 @@ class PocEnv(ExecEnv):
             except BaseException as cleanup_error:
                 raise RuntimeError(
                     f"POC deployment failed ({deployment_error}); cleanup could not be completed safely "
-                    f"for the per-run workspace {self.poc_workspace}: {cleanup_error}"
+                    f"for the per-run workspace {self.poc_workspace}: {cleanup_error}. Stop any remaining POC "
+                    "services and remove this workspace manually."
                 ) from cleanup_error
             raise
         self.logger.info("POC services started successfully")
--- a/nvflare/recipe/poc_env.py
+++ b/nvflare/recipe/poc_env.py
@@ -383,8 +383,11 @@ class PocEnv(ExecEnv):
                     if not self._running_services(project_config, service_config, self.poc_workspace):
                         poc_running = False
                         break
-                except Exception:
-                    poc_running = False
+                except Exception as state_error:
+                    self.logger.warning(f"Could not verify whether POC services stopped: {state_error}")
+                    # Preserve the workspace when service state is unknown. It
+                    # contains the configuration needed for manual cleanup.
+                    poc_running = True
                     break
                 time.sleep(1)
                 count += 1
--- a/nvflare/recipe/poc_env.py
+++ b/nvflare/recipe/poc_env.py
@@ -12,9 +12,13 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import errno
+import fcntl
 import os
 import shutil
+import stat
 import subprocess
+import tempfile
 import threading
 import time
 import uuid
@@ -51,6 +55,11 @@ POC_READY_POLL_INTERVAL = 0.2
 POC_READY_STABLE_INTERVAL = 2.0
 DEFAULT_ADMIN_USER = "admin@nvidia.com"
 _RECIPE_WORKSPACE_SUFFIX = ".recipe-"
+
+
+def _recipe_runtime_lock_path() -> str:
+    """Return the per-user host lock that serializes Recipe POC runtimes."""
+    return os.path.join(tempfile.gettempdir(), f".nvflare-recipe-poc-{os.geteuid()}.lock")
 
 
 # Internal — not part of the public API
@@ -154,6 +163,7 @@ class PocEnv(ExecEnv):
         self._session_manager = None  # Lazy initialization
         self._session_manager_lock = threading.Lock()
         self._workspace_used = False
+        self._runtime_lock_file = None
 
     def _new_poc_workspace(self) -> str:
         """Return a unique Recipe-owned workspace beside the configured POC path."""
@@ -189,6 +199,48 @@ class PocEnv(ExecEnv):
         except Exception:
             return False
         return bool(self._running_services(project_config, service_config, workspace))
+
+    def _acquire_runtime_lock(self) -> None:
+        """Claim the host's single Recipe-managed POC runtime slot."""
+        if self._runtime_lock_file is not None:
+            return
+
+        lock_path = _recipe_runtime_lock_path()
+        flags = os.O_CREAT | os.O_RDWR
+        if hasattr(os, "O_CLOEXEC"):
+            flags |= os.O_CLOEXEC
+        if hasattr(os, "O_NOFOLLOW"):
+            flags |= os.O_NOFOLLOW
+        fd = os.open(lock_path, flags, 0o600)
+        try:
+            lock_stat = os.fstat(fd)
+            if not stat.S_ISREG(lock_stat.st_mode) or lock_stat.st_uid != os.geteuid():
+                raise RuntimeError(f"Refusing to use unsafe Recipe POC runtime lock {lock_path}")
+            os.fchmod(fd, 0o600)
+            try:
+                fcntl.flock(fd, fcntl.LOCK_EX | fcntl.LOCK_NB)
+            except OSError as e:
+                if e.errno in (errno.EACCES, errno.EAGAIN):
+                    raise RuntimeError(
+                        "Another Recipe PocEnv deployment is active on this host. "
+                        "Stop it before starting this deployment."
+                    ) from e
+                raise
+            self._runtime_lock_file = os.fdopen(fd, "a+")
+        except BaseException:
+            os.close(fd)
+            raise
+
+    def _release_runtime_lock(self) -> None:
+        """Release this environment's Recipe POC runtime slot, if held."""
+        lock_file = self._runtime_lock_file
+        if lock_file is None:
+            return
+        self._runtime_lock_file = None
+        try:
+            fcntl.flock(lock_file.fileno(), fcntl.LOCK_UN)
+        finally:
+            lock_file.close()
 
     @staticmethod
     def _is_docker_service_running(service_name: str) -> bool:
@@ -284,18 +336,25 @@ class PocEnv(ExecEnv):
                 f"For PocEnv, all scripts must be present on the local machine."
             )
 
-        # Unique workspaces isolate files, but POC deployments still share
-        # default ports and Docker participant names. Never compete with the
-        # reusable CLI-managed deployment selected by the user's configuration.
-        if self._is_poc_workspace_running(self._poc_workspace_root):
-            raise RuntimeError(
-                f"The configured CLI POC deployment is running at {self._poc_workspace_root}. "
-                "Stop it with 'nvflare poc stop' before starting a Recipe PocEnv deployment."
-            )
+        if self._workspace_used and self._check_poc_running():
+            raise RuntimeError("This PocEnv already has a running deployment; stop it before deploying another job")
+
+        # Recipe POC workspaces have isolated files but share host ports and
+        # Docker participant names. The process-held lock closes the race
+        # between checking those resources and starting services. It is
+        # released only after this deployment's services have stopped.
+        self._acquire_runtime_lock()
+        try:
+            if self._is_poc_workspace_running(self._poc_workspace_root):
+                raise RuntimeError(
+                    f"The configured CLI POC deployment is running at {self._poc_workspace_root}. "
+                    "Stop it with 'nvflare poc stop' before starting a Recipe PocEnv deployment."
+                )
+        except BaseException:
+            self._release_runtime_lock()
+            raise
 
         if self._workspace_used:
-            if self._check_poc_running():
-                raise RuntimeError("This PocEnv already has a running deployment; stop it before deploying another job")
             self.poc_workspace = self._new_poc_workspace()
             self._session_manager = None
         self._workspace_used = True
@@ -369,6 +428,7 @@ class PocEnv(ExecEnv):
                 self.logger.info(f"Removing POC workspace: {self.poc_workspace}")
                 shutil.rmtree(self.poc_workspace, ignore_errors=True)
             self._session_manager = None  # Clear stale session manager
+            self._release_runtime_lock()
             return
 
         try:
@@ -423,6 +483,8 @@ class PocEnv(ExecEnv):
                         self.logger.warning(
                             f"Failed to clean POC workspace {self.poc_workspace}: {e}. Remove it manually."
                         )
+            if not poc_running:
+                self._release_runtime_lock()
         except Exception as e:
             self.logger.warning(f"Failed to stop and clean existing POC: {e}")
         finally:
--- a/nvflare/recipe/poc_env.py
+++ b/nvflare/recipe/poc_env.py
@@ -195,11 +195,17 @@ class PocEnv(ExecEnv):
         if os.path.exists(self.poc_workspace):
             raise RuntimeError("the per-run POC workspace could not be removed")
 
-    def _is_poc_workspace_running(self, workspace: str) -> bool:
+    def _is_poc_workspace_running(self, workspace: str, fail_if_unknown: bool = False) -> bool:
         """Return whether any managed service is running in a POC workspace."""
         try:
             project_config, service_config = setup_service_config(workspace)
-        except Exception:
+        except Exception as e:
+            if fail_if_unknown:
+                raise RuntimeError(
+                    f"Could not determine service state for the previously active Recipe PocEnv workspace {workspace}: "
+                    f"{e}. Stop any remaining services, then remove the stale runtime record "
+                    f"{_recipe_runtime_lock_path()} manually."
+                ) from e
             return False
         return bool(self._running_services(project_config, service_config, workspace))
 
@@ -232,7 +238,7 @@ class PocEnv(ExecEnv):
             lock_file = os.fdopen(fd, "r+")
             fd = None
             previous_workspace = lock_file.read().strip()
-            if previous_workspace and self._is_poc_workspace_running(previous_workspace):
+            if previous_workspace and self._is_poc_workspace_running(previous_workspace, fail_if_unknown=True):
                 raise RuntimeError(
                     f"A prior Recipe PocEnv deployment is still active at {previous_workspace}. "
                     "Stop its services and remove the workspace manually before starting another deployment."
--- a/nvflare/recipe/poc_env.py
+++ b/nvflare/recipe/poc_env.py
@@ -187,7 +187,9 @@ class PocEnv(ExecEnv):
         """Stop a failed deployment and verify its per-run workspace was cleaned."""
         if not self._is_recipe_workspace(self.poc_workspace):
             raise RuntimeError(f"refusing to clean unmanaged POC workspace {self.poc_workspace}")
-        self.stop(clean_up=True)
+        # deploy() already holds the instance lifecycle guard. Use the private
+        # implementation so failure cleanup cannot deadlock on that guard.
+        self._stop(clean_up=True)
         if self._check_poc_running():
             raise RuntimeError("POC services remain running")
         if os.path.exists(self.poc_workspace):
@@ -463,6 +465,13 @@ class PocEnv(ExecEnv):
         Args:
             clean_up (bool, optional): Whether to clean the POC workspace. Defaults to False.
         """
+        # Wait for an in-progress deployment to finish before inspecting
+        # services or clearing its durable runtime record.
+        with self._deployment_lock:
+            self._stop(clean_up)
+
+    def _stop(self, clean_up: bool = False) -> None:
+        """Stop POC while the caller holds the instance lifecycle guard."""
         # Check if already stopped (idempotent)
         if not self._check_poc_running():
             # POC already stopped or workspace doesn't exist
--- a/nvflare/recipe/poc_env.py
+++ b/nvflare/recipe/poc_env.py
@@ -207,7 +207,16 @@ class PocEnv(ExecEnv):
                     f"{_recipe_runtime_lock_path()} manually."
                 ) from e
             return False
-        return bool(self._running_services(project_config, service_config, workspace))
+        try:
+            return bool(self._running_services(project_config, service_config, workspace))
+        except Exception as e:
+            if fail_if_unknown:
+                raise RuntimeError(
+                    f"Could not determine service state for the previously active Recipe PocEnv workspace {workspace}: "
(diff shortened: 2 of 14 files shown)


nvflare/recipe/poc_env.py
```diff
@@ -348,70 +287,14 @@ def deploy(self, job: FedJob) -> str:
...
-                    self._workspace_owned = True
-                    raise RuntimeError(
-                        f"POC deployment failed and the retained workspace could not be restored from {workspace_backup}"
-                    ) from restore_error
-                self._workspace_owned = False
-            else:
-                self._workspace_owned = True
+            # This path is unique to the current Recipe execution, so failure
+            # cleanup cannot delete a retained CLI workspace or a prior run.
+            self.stop(clean_up=True)
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in `4aced538d`. Failure cleanup now performs a post-stop verification: it checks that no managed POC service remains and that the unique per-run workspace was removed. If stop suppresses an error, service state is unknown, services remain alive, or cleanup leaves the directory behind, `deploy()` raises an actionable error containing the run-workspace path plus both the deployment and cleanup failure details. Added regression coverage for a submission failure followed by incomplete shutdown; the focused suite is 30 passed and the broader Recipe/POC suite is 679 passed, 18 skipped.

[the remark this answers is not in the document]
chesterxgchen: Follow-up refinement in `2c8235c26`: cleanup failure handling remains explicit but intentionally stateless. A failed deployment gets one stop/remove attempt and post-check. If that cannot be verified, the raised error includes the original failure, cleanup failure, and exact unique workspace path with manual stop/removal guidance. There is no retry marker, workspace scan, backup/restore, or cross-run ownership state. Updated coverage passes (31 focused; 680 Recipe/POC tests with 18 skipped).

nvflare/recipe/poc_env.py
```diff
@@ -150,65 +152,19 @@ def __init__(
...
-            else "there was no retained workspace to back up"
-        )
-        raise RuntimeError(
-            f"POC deployment failed and {reason}; "
-            f"the partial replacement remains at {self.poc_workspace}; {backup_note}"
-        ) from error
+    def _clean_up_failed_deployment(self) -> None:
+        """Stop a failed deployment and verify its per-run workspace was cleaned."""
+        self.stop(clean_up=True)
+        if self._check_poc_running():
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in `743328020`. The shutdown poll now treats an inspection exception as unknown/running, preserves the per-run workspace (including the service configuration needed for manual cleanup), and does not call `_clean_poc`. The deployment-failure post-check therefore reports the cleanup problem and exact workspace instead of losing service identity after deletion. Added direct regression coverage; 32 focused tests and 681 broader Recipe/POC tests pass.

chesterxgchen:
Addressed the latest review findings in `9f8d9b9e3`:

- added a read-only preflight that rejects Recipe `PocEnv` deployment while any managed service is active in the configured CLI POC workspace; provisioning is not entered and the CLI workspace is not modified
- documented that unique workspaces isolate files, while default ports and Docker participant names remain shared
- changed cleanup failure handling to catch `Exception`, so cleanup-time `KeyboardInterrupt` and `SystemExit` propagate normally
- added unit coverage plus a real-process integration test that starts a CLI POC, verifies Recipe execution is rejected, and confirms the CLI workspace remains intact

Validation: 35 focused unit tests passed; 684 broader Recipe/POC tests passed with 18 skipped; both real POC integration cases passed; repository style and docs builds passed.

nvflare/recipe/poc_env.py
```diff
@@ -309,18 +284,23 @@ def deploy(self, job: FedJob) -> str:
                 f"For PocEnv, all scripts must be present on the local machine."
             )
 
-        self._validate_project_conf_location()
+        # Unique workspaces isolate files, but POC deployments still share
+        # default ports and Docker participant names. Never compete with the
+        # reusable CLI-managed deployment selected by the user's configuration.
+        if self._is_poc_workspace_running(self._poc_workspace_root):
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in `d63883e52`. Recipe-managed POC runtimes now hold one nonblocking per-user host advisory lock from preflight until their services stop. A second `PocEnv` object or process is rejected before provisioning, so distinct workspace paths cannot race on the shared ports or Docker participant names. The OS releases the lock if the owning process exits; no persistent ownership record, workspace scan, backup, or restore state was added. Added coverage that keeps one environment active, rejects the second without provisioning, then verifies the second can deploy after the first stops. Validation: 39 focused tests; 688 broader Recipe/POC tests with 18 skipped; both real POC integration cases passed; style and docs builds passed.

chesterxgchen:
Addressed the actionable review findings in `d8f90e449`:

- the host lock file now durably records the active Recipe workspace path; after an unexpected process exit, the next deployment acquires the OS lock, verifies services in that prior workspace, and refuses to start if any remain
- normal verified shutdown clears the recorded path; no workspace scan, backup/restore flow, retry marker, or ownership state machine was added
- a separate instance-level guard rejects overlapping `deploy()` calls on the same `PocEnv` object
- added regression coverage for post-crash service detection, recovery after those services stop, durable record clearing, and same-instance overlap

The `fcntl` import is intentionally unchanged: NVFlare currently declares POSIX/Linux operating-system support in `setup.cfg`; Windows is outside this supported package matrix.

Validation: 42 focused tests passed; 691 broader Recipe/POC tests passed with 18 skipped; both real POC integration cases passed; repository style and docs builds passed.

chesterxgchen:
Addressed the stop/deploy race in `aecd204f6`. Public `stop()` now takes the same per-instance lifecycle guard as `deploy()`, so it cannot inspect services, clear the durable workspace record, or release the host runtime lock while startup is in progress. Deployment-failure cleanup calls the private stop implementation while already inside that guard, avoiding reentrant deadlock. Added a threaded regression test proving stop waits for deployment to leave the guard.

Validation: 43 focused tests passed; 692 broader Recipe/POC tests passed with 18 skipped; both real POC integration cases passed; repository style checks passed.

nvflare/recipe/poc_env.py
```diff
@@ -150,65 +162,113 @@ def __init__(
...
+                if e.errno in (errno.EACCES, errno.EAGAIN):
+                    raise RuntimeError(
+                        "Another Recipe PocEnv deployment is active on this host. "
+                        "Stop it before starting this deployment."
+                    ) from e
+                raise
+            lock_file = os.fdopen(fd, "r+")
+            fd = None
+            previous_workspace = lock_file.read().strip()
+            if previous_workspace and self._is_poc_workspace_running(previous_workspace):
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in 0629415f8. The durable prior-workspace record is now checked in strict mode: if its service configuration is missing or unreadable, deployment fails closed before provisioning, keeps the record intact, releases the advisory lock, and reports the recorded workspace plus the runtime-record path for manual recovery. Normal first-run inspection of the configured CLI workspace retains its existing behavior. Added a regression test covering an unreadable recorded workspace and documented the recovery behavior. Validation: 44 PocEnv tests passed; 693 recipe/POC tests passed (18 skipped); ./runtest.sh -s passed; the HTML docs build completed successfully.

chesterxgchen:
Addressed Review #400 in `1661ef3e3`. Strict stale-runtime inspection now wraps both service-configuration loading and managed-service lookup. If either stage cannot establish state, deployment fails closed before provisioning, preserves the durable workspace record, releases the advisory lock, and reports the manual recovery paths. Normal configured-CLI-workspace inspection retains its prior error behavior. Added regression coverage for a readable configuration whose service state cannot be evaluated. Validation: 45 focused PocEnv tests passed; 694 Recipe/POC tests passed (18 skipped); `./runtest.sh -s` and `git diff --check` passed.

nvflare/recipe/poc_env.py
```diff
@@ -309,18 +390,35 @@ def deploy(self, job: FedJob) -> str:
...
+            if self._is_poc_workspace_running(self._poc_workspace_root):
+                raise RuntimeError(
+                    f"The configured CLI POC deployment is running at {self._poc_workspace_root}. "
+                    "Stop it with 'nvflare poc stop' before starting a Recipe PocEnv deployment."
+                )
+        except BaseException:
+            self._release_runtime_lock()
+            raise
+
+        if self._workspace_used:
```
YuanTingHsieh: [P1] Make `PocEnv` a one-shot environment

Recipe `PocEnv` should own exactly one deployment lifecycle, but this branch explicitly reuses a stopped instance by rotating it to a new workspace. Once this environment has begun provisioning, every later `deploy()` call should be rejected; `stop()` can remain idempotent. Please remove the workspace-rotation branch and the associated reuse tests and documentation.

nvflare/recipe/poc_env.py
```diff
@@ -464,27 +515,42 @@ def stop(self, clean_up: bool = False) -> None:
...
+                        else f"services are still running after {STOP_POC_TIMEOUT} seconds"
+                    )
                     self.logger.warning(
-                        f"POC still running after {STOP_POC_TIMEOUT} seconds, cannot clean workspace. Skipping cleanup."
+                        f"POC {reason}; preserving workspace {self.poc_workspace}. "
+                        "Stop any remaining services and remove it manually."
                     )
                 else:
                     try:
                         _clean_poc(self.poc_workspace)
```
YuanTingHsieh: [P1] Keep Recipe cleanup out of CLI configuration

`Run.get_result()` defaults to `clean_up=True`, so a normal successful Recipe run reaches this call. `_clean_poc()` invokes `_clean_poc_config()`, which unconditionally removes `poc.workspace`, `poc.startup_kit`, and `prod.startup_kit` from the global CLI configuration. Now that `self.poc_workspace` is a Recipe-owned sibling rather than the configured CLI workspace, this erases the reusable CLI POC settings even though their files remain. Use Recipe-only directory cleanup here, or make CLI configuration cleanup conditional on an exact configured-workspace match.

nvflare/recipe/poc_env.py
```diff
@@ -150,65 +162,130 @@ def __init__(
...
+        try:
+            project_config, service_config = setup_service_config(workspace)
+        except Exception as e:
+            if fail_if_unknown:
+                raise RuntimeError(
+                    f"Could not determine service state for the previously active Recipe PocEnv workspace {workspace}: "
+                    f"{e}. Stop any remaining services, then remove the stale runtime record "
+                    f"{_recipe_runtime_lock_path()} manually."
+                ) from e
+            return False
```
YuanTingHsieh, line 298: [P1] Preserve unknown current-runtime state

This non-strict fallback is unsafe once startup may have occurred. If `_start_poc()` succeeds but the later `setup_service_config()` call fails, deployment cleanup enters `_stop()`; its initial `_check_poc_running()` reaches this `return False`, deletes the workspace, and clears/releases the runtime lock without stopping the services. The subsequent verification also reports false because the configuration is gone, potentially leaving processes or containers running untracked. Missing metadata can mean stopped only before services could have started; afterward cleanup must fail closed and preserve the workspace and lock.

nvflare/recipe/poc_env.py
```diff
@@ -50,7 +54,12 @@
 POC_READY_POLL_INTERVAL = 0.2
 POC_READY_STABLE_INTERVAL = 2.0
 DEFAULT_ADMIN_USER = "admin@nvidia.com"
-_WORKSPACE_BACKUP_PREFIX = ".nvflare-recipe-backup-"
+_RECIPE_WORKSPACE_SUFFIX = ".recipe-"
+
+
+def _recipe_runtime_lock_path() -> str:
+    """Return the per-user host lock that serializes Recipe POC runtimes."""
+    return os.path.join(tempfile.gettempdir(), f".nvflare-recipe-poc-{os.geteuid()}.lock")
```
holgerroth: `tempfile.gettempdir()` is resolved per process from `TMPDIR`/`TEMP`/`TMP`, so two same-user processes with different temp dirs open different lock files and both proceed onto 8002/8003. See the summary comment for the two options (port preflight, or anchor beside the configured POC root). `test_recipe_runtime_lock_path_is_host_and_user_scoped` recomputes the same formula, so it can't catch this.

nvflare/recipe/poc_env.py
```diff
@@ -309,18 +390,35 @@ def deploy(self, job: FedJob) -> str:
                 f"For PocEnv, all scripts must be present on the local machine."
             )
 
-        self._validate_project_conf_location()
+        if self._workspace_used and self._check_poc_running():
```
holgerroth: This replaces the auto-stop-and-redeploy that shipped in 2.8.0/2.8.1 (`if self._check_poc_running(): self.stop(clean_up=True)`). Fine given the one-shot direction, but please add a release-note line, and note that `Run.abort()` does not stop the env so abort-then-execute on the same `PocEnv` now raises.

nvflare/recipe/poc_env.py
```diff
@@ -150,65 +162,130 @@ def __init__(
...
+        if os.path.exists(self.poc_workspace):
+            raise RuntimeError("the per-run POC workspace could not be removed")
+
+    def _is_poc_workspace_running(self, workspace: str, fail_if_unknown: bool = False) -> bool:
+        """Return whether any managed service is running in a POC workspace."""
+        try:
+            project_config, service_config = setup_service_config(workspace)
+        except Exception as e:
+            if fail_if_unknown:
+                raise RuntimeError(
```
holgerroth: Nit: this message is duplicated verbatim at 214-218, and `_deploy` has two identical `except BaseException: self._release_runtime_lock(); raise` wrappers at 407-409 and 416-418 with the workspace-rotation block unguarded between them. If the rotation branch goes away per YuanTing's comment, a single try around preflight and `_record_runtime_workspace()` and a small helper for this message would keep both in one place.

tests/unit_test/recipe/poc_env_test.py
```diff
@@ -102,403 +336,366 @@ def test_poc_env_none_num_clients_raises():
...
+        Path(kwargs["workspace"]).mkdir(parents=True)
 
-    monkeypatch.setattr(poc_env_module, "_start_poc", fail_start)
-    env = PocEnv()
+    def fail_submission(job):
+        raise RuntimeError("submission failed")
 
-    with pytest.raises(RuntimeError, match="start failed"):
+    _configure_successful_deploy(monkeypatch, env, prepare=prepare, submit=fail_submission)
+    monkeypatch.setattr(env, "stop", lambda clean_up: None)
```
holgerroth: This patches `env.stop`, but `_clean_up_failed_deployment` calls `self._stop`, so the patch is inert: the real `_stop` runs, `is_poc_running` fails on the fake workspace, and the broad `except Exception` at poc_env.py:554 swallows it (visible as a "Failed to stop and clean existing POC" warning with live logging). The assertion holds only because `_check_poc_running` is patched to True, so the test would still pass if the `_stop` call were removed. Patch `"_stop"` and record its `clean_up` argument, as the lock-contention test at line 260 does.

holgerroth (commented):
Thanks for pulling the backup/restore state machine from #5258 back out; unique sibling workspaces are the right fix for the file conflict and I'd like to see this land before the next release cut.

I agree with all three of @YuanTingHsieh's P1s (one-shot `PocEnv`, keep `_clean_poc_config` out of Recipe cleanup, preserve unknown state after `_start_poc` has run). A few additional points that they don't cover:

**Lock scope.** The advisory lock is keyed on `tempfile.gettempdir()`, so its identity depends on each process's `TMPDIR`. Two shells of the same user with different `TMPDIR` (macOS per-login temp vs. an IDE/CI/sandbox that sets it) get different lock files, both acquire, and both bind 8002/8003, which is exactly the collision the docs say is prevented. Since the lock only exists to guard ports and Docker names, I'd consider replacing it with a port-availability preflight using the existing `_is_local_port_available` in `poc_commands` (this also covers the CLI-coexistence case without any durable state). If the lock stays, anchor it beside the configured POC root (e.g. `<root>.recipe.lock`), which both processes already agree on. Related: because the record is written before `prepare_poc_provision` and cleared after `rmtree`/`_clean_poc`, a kill in either window leaves a recorded path with no directory, and every later `deploy()` fails closed until the user hand-deletes the lock file. For non-Docker mode a missing directory cannot host live pid-tracked processes, so treating it as not-running (or clearing the record before removing the directory) would remove that false positive.

**Tutorial notebooks.** `examples/tutorials/job_recipe.ipynb` ("The result are stored under the directory `/tmp/nvflare/poc`" and "The POC environment is always started fresh at the beginning of each `recipe.execute(env)` call") and `examples/tutorials/flare_api.ipynb` (the commented "Alternative method" block presenting `recipe.execute(PocEnv(...))` while the notebook's own CLI POC is running) both describe the old behavior; with this PR the first is a UUID sibling that is deleted by default and the second raises. Worth updating alongside the RST changes.

**Release note.** Rejecting `deploy()` on a running env replaces the auto-stop-and-redeploy that shipped in 2.8.0/2.8.1, and `Run.abort()` does not stop the env, so `execute(); abort(); execute()` on one `PocEnv` now raises. That is consistent with the one-shot direction, but it deserves a line in the release notes.

Smaller items are inline. CI, style, and license checks are green and the unit tests pass locally under `-n 8`.

nvflare/recipe/poc_env.py
chesterxgchen: Fixed in `5cdbbf199`. `PocEnv` is now one-shot once provisioning begins: the instance is marked consumed immediately before `prepare_poc_provision()`, the workspace-rotation branch was removed, and every later `deploy()` is rejected even after `stop()`. Pre-provisioning validation/runtime-preflight rejections remain retryable. The reuse documentation and tests were replaced with the one-shot contract, including coverage that a provisioning failure also consumes the instance.

nvflare/recipe/poc_env.py
chesterxgchen: Fixed in `5cdbbf199`. The environment now records that services may have started before entering `_start_poc()`, covering partial-start exceptions as well as the reported post-start metadata failure. From that point, current-workspace inspection is strict: missing/unreadable service metadata raises instead of reporting stopped, preserving the Recipe workspace, durable runtime record, and held lock for recovery. Added a regression test for metadata loss immediately after startup.

nvflare/recipe/poc_env.py
chesterxgchen: Fixed in `5cdbbf199`. Recipe cleanup no longer calls the CLI `_clean_poc()` path. A guarded Recipe-only helper validates the sibling workspace path and removes only that directory with `shutil.rmtree`, so the reusable CLI workspace/startup-kit configuration is untouched. Regression coverage verifies that Recipe shutdown removes the run workspace without invoking `_clean_poc_config()`.

nvflare/recipe/poc_env.py
chesterxgchen: Fixed in `bc56ac661`. Added a compatibility note to the 2.9 release notes documenting that `PocEnv` is now one-shot, that this replaces the 2.8 auto-stop-and-redeploy behavior, and that `Run.abort()` does not stop the environment—so aborting and then executing with the same `PocEnv` raises. The note directs users to stop/collect the run and create a new instance.

nvflare/recipe/poc_env.py
chesterxgchen: Fixed in `bc56ac661`. Strict unknown-state recovery guidance is now centralized in `_raise_unknown_service_state()`, and CLI preflight plus `_record_runtime_workspace()` share one release-on-failure wrapper after the workspace-rotation branch was removed.

tests/unit_test/recipe/poc_env_test.py
chesterxgchen: Fixed in `bc56ac661`. The test now patches the private `_stop()` method that `_clean_up_failed_deployment()` actually calls, records its `clean_up` argument, and asserts it received `True`. This removes the inert public-`stop()` patch and makes the test fail if the cleanup stop call disappears.

nvflare/recipe/poc_env.py
chesterxgchen: Fixed in `bc56ac661`. The lock path is now derived from the absolute configured POC root and placed beside it, with the effective UID in the filename. It no longer uses `tempfile.gettempdir()`, so same-user processes with different `TMPDIR`/`TEMP` settings converge on the same lock. The regression test changes `TMPDIR` between calls and verifies the path is unchanged.

nvflare/recipe/poc_env.py
```diff
@@ -50,7 +52,12 @@
 POC_READY_POLL_INTERVAL = 0.2
 POC_READY_STABLE_INTERVAL = 2.0
 DEFAULT_ADMIN_USER = "admin@nvidia.com"
-_WORKSPACE_BACKUP_PREFIX = ".nvflare-recipe-backup-"
+_RECIPE_WORKSPACE_SUFFIX = ".recipe-"
+
+
+def _recipe_runtime_lock_path(poc_workspace_root: str) -> str:
+    """Return the per-user runtime lock beside the configured POC workspace."""
+    return f"{os.path.abspath(poc_workspace_root)}.recipe-{os.geteuid()}.lock"
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in `461a18128`. The lock identity is now independent of both Recipe/CLI workspace selection and process environment: all same-user POC runtimes use `/tmp/.nvflare-recipe-poc-<euid>.lock` (with `/tmp` canonicalized via `realpath`). The regression test changes both `TMPDIR` and `NVFLARE_POC_WORKSPACE` between calls and verifies they resolve to the same lock.

nvflare/recipe/poc_env.py
chesterxgchen: Follow-up correction in `461a18128`: anchoring beside the configured workspace still allowed two different workspace roots to collide on shared ports. The final implementation uses one fixed `/tmp/.nvflare-recipe-poc-<euid>.lock` per user, independent of both workspace root and process temp-directory environment.

nvflare/recipe/poc_env.py
```diff
@@ -50,7 +52,12 @@
 POC_READY_POLL_INTERVAL = 0.2
 POC_READY_STABLE_INTERVAL = 2.0
 DEFAULT_ADMIN_USER = "admin@nvidia.com"
-_WORKSPACE_BACKUP_PREFIX = ".nvflare-recipe-backup-"
+_RECIPE_WORKSPACE_SUFFIX = ".recipe-"
+
+
+def _recipe_runtime_lock_path() -> str:
+    """Return the environment-independent, per-user host runtime lock."""
+    return os.path.join(os.path.realpath("/tmp"), f".nvflare-recipe-poc-{os.geteuid()}.lock")
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in `9b16f9f5d`. The lock now lives at `<OS-account-home>/.nvflare/recipe-poc-runtime.lock`, with the home resolved from the effective UID via `pwd` rather than `HOME`. Acquisition creates the directory as mode 0700 when absent and rejects it unless it is a directory owned by the effective user with no group/world write bits; the existing `O_NOFOLLOW`, regular-file ownership, and 0600 checks still protect the lock file. Added coverage for an unsafe writable lock directory and for path stability across `HOME`, `TMPDIR`, and workspace overrides.

nvflare/recipe/poc_env.py
chesterxgchen: Correction in `9b16f9f5d`: FLARE 2.9 is already released, so its release notes are restored unchanged. There is not yet a 2.10 release-notes draft in the repository; the compatibility warning (including `Run.abort()` behavior) now lives in the current Recipe API documentation for the 2.10 codebase.

nvflare/recipe/poc_env.py
```diff
@@ -50,7 +53,13 @@
 POC_READY_POLL_INTERVAL = 0.2
 POC_READY_STABLE_INTERVAL = 2.0
 DEFAULT_ADMIN_USER = "admin@nvidia.com"
-_WORKSPACE_BACKUP_PREFIX = ".nvflare-recipe-backup-"
+_RECIPE_WORKSPACE_SUFFIX = ".recipe-"
+
+
+def _recipe_runtime_lock_path() -> str:
+    """Return the environment-independent runtime lock for the OS user."""
+    user_home = pwd.getpwuid(os.geteuid()).pw_dir
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in `90be5f25e`. Lock selection now prefers the effective UID's passwd home but falls back to `$HOME/.nvflare` when the preferred location cannot be created or opened; a missing passwd entry selects `$HOME` directly. Every candidate still requires an effective-user-owned, non-group/world-writable directory and the existing protected regular-file checks. The selected path is retained for stale-runtime recovery guidance. Added regressions for both a missing passwd entry and an unwritable preferred home; the focused suite is 48 passed and style is clean.

nvflare/recipe/poc_env.py
```diff
@@ -50,7 +53,29 @@
...
+
+def _recipe_runtime_lock_paths() -> list[str]:
+    """Return preferred and environment-home fallback lock paths."""
+    paths = [_recipe_runtime_lock_path()]
+    env_home = os.environ.get("HOME")
+    if env_home:
+        fallback = os.path.join(env_home, ".nvflare", "recipe-poc-runtime.lock")
+        if fallback not in paths:
+            paths.append(fallback)
+    return paths
```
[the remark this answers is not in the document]
chesterxgchen: Addressed in 2d2d43719. The fallback no longer uses `$HOME`, so processes with different environment homes cannot select different locks. Candidate locations are derived only from the effective OS user: the passwd home first, then OS-managed per-user runtime directories (`/run/user/<uid>`/`/var/run/user/<uid>` on Linux or `DARWIN_USER_TEMP_DIR` on macOS). Every selected directory and lock file still receives the ownership/type/mode checks; if no stable per-user location is available, deployment fails closed. The regression test removes the passwd entry, changes `$HOME` between two resolutions, and verifies that the candidate paths remain identical. Validation: 48 focused tests passed and `./runtest.sh -s` passed.

holgerroth:
Thanks, the lifecycle changes in `5cdbbf1`/`bc56ac6` (one-shot env, Recipe-only `rmtree`, strict state after `_start_poc`) address my review and YuanTing's P1s; that part looks ready.

On the runtime lock: it has moved through five locations today and `_recipe_runtime_lock_paths` is now the largest piece of new code in the file, plus the durable record, fail-closed recovery text, and the `pwd`/`getconf` resolution. I'd suggest dropping it. The conflict it guards is two deployments binding 8002/8003 (and, in Docker mode, the default participant names). A direct preflight before `prepare_poc_provision` that checks those ports with the existing `_is_local_port_available` in `poc_commands`, and keeps the current `docker inspect` name check, detects the real conflict including the CLI POC and stale servers from crashed runs, with no persistent state, no lock-location logic, and no manual recovery step. A second run started in the same instant would still fail loudly at `_wait_for_services_ready`, which is acceptable for a local dev tool.

If the team prefers to keep a lock, then a single location via `get_hidden_nvflare_dir()` (same `$HOME` semantics as the CLI config) without the fallback chain would be my second choice.

One leftover from my review: `examples/tutorials/job_recipe.ipynb` and `flare_api.ipynb` still describe the old result path and running-CLI coexistence; fine as a follow-up if you prefer.

chesterxgchen:
@holgerroth Agreed on simplifying the runtime-conflict handling. Addressed in `6e74ffc92`:

- removed the host filesystem lock, durable workspace record, lock-location discovery, and manual stale-record recovery path
- added direct preflight of the actual server ports from the provisioned project immediately before service startup (this preserves custom `project_conf_path` ports)
- retained the configured CLI-workspace check before provisioning and added Docker inspection that rejects any existing selected participant container name
- made pre-start failure cleanup remove only the new unique Recipe workspace, so detecting another Docker deployment cannot stop its containers
- kept startup/readiness failure as the loud fallback for the narrow simultaneous-preflight race

The two tutorial notebooks were also updated in `d5263b347`: `job_recipe.ipynb` now describes the UUID sibling workspace/default result cleanup and one-shot lifecycle, and `flare_api.ipynb` labels `PocEnv` as an alternative only when no CLI POC is running.

Validation: 39 focused tests passed; 688 broader Recipe/POC tests passed with 18 skipped; both real-process POC integration tests passed; `./runtest.sh -s`, the HTML docs build, notebook JSON validation, and `git diff --check` passed.

nvflare/recipe/poc_env.py
```diff
@@ -437,21 +432,32 @@ def stop(self, clean_up: bool = False) -> None:
...
         """
+        # Wait for an in-progress deployment to finish before inspecting its
+        # services or removing its workspace.
+        with self._deployment_lock:
+            self._stop(clean_up)
+
+    def _stop(self, clean_up: bool = False) -> None:
+        """Stop POC while the caller holds the instance lifecycle guard."""
         # Check if already stopped (idempotent)
         if not self._check_poc_running():
```
holgerroth: With `_services_may_have_started` True, this call raises (via `_raise_unknown_service_state`) before the `try:` below, so `stop()` never attempts `_stop_poc`, never reaches the `finally` that clears the session manager, and never resets the flag, so every later `stop()` raises too. `Run.get_result` catches it, logs a warning, and marks the run stopped, so with the default `clean_up=True` the services and workspace leak silently. A scratch test with `setup_service_config` patched to raise shows `env.stop(clean_up=True)` raising on both calls with `_stop_poc` called zero times. The docstring above still promises idempotency.

Suggest handling unknown state here the way the poll loop at 479-485 already does: log, still attempt `_stop_poc`, preserve the workspace, keep the `finally`. `_clean_up_failed_deployment` keeps its own post-checks at 202-205, so failure cleanup still fails loudly. A unit test that drives real `_stop` with the flag set and `setup_service_config` raising, plus a double-`stop()` after a successful deploy, would pin this (the current failure-path tests stub `_check_poc_running`, and dropping the try/except around `_remove_recipe_workspace` at 449-452 passes all 39 tests).

nvflare/recipe/poc_env.py
```diff
@@ -233,13 +242,42 @@ def _is_docker_service_running(service_name: str) -> bool:
...
 
+    @staticmethod
+    def _is_docker_service_running(service_name: str) -> bool:
+        """Return whether Docker reports the named POC container as running."""
+        return PocEnv._get_docker_service_state(service_name) is True
+
+    @staticmethod
+    def _ensure_shared_resources_available(project_config: dict, service_config: dict) -> None:
+        """Reject ports or Docker names already owned by another POC deployment."""
+        port_conflicts = _build_poc_port_preflight(project_config).get("conflicts", [])
```
holgerroth: `_is_local_port_available` (poc_commands.py ~795) does a plain `bind()` with no `SO_REUSEADDR`, while the actual listeners bind with reuse (`tcp_driver.py:33`, gRPC defaults). A port whose previous server-side connections are in TIME_WAIT is therefore reported `in_use` for ~30 s (macOS) / 60 s (Linux) after `run.get_result()`, though the server would bind fine. Reproduced here: with a server-side TIME_WAIT on a port, this helper returned `(False, 'in_use')` while a bind with `SO_REUSEADDR` and a real `grpc.server().add_insecure_port` both succeeded. The CLI treats the same preflight as advisory; here it is fatal, and because it runs after `_deployment_started = True` and full provisioning (line 366-380), a false positive burns the env. The `get_result()` then `PocEnv()` pattern in `job_recipe.ipynb` is exactly the sequence that hits it.

Two small changes: set `SO_REUSEADDR` in `_is_local_port_available` before binding, and run the port probe before line 366 on the intended ports (defaults from poc_commands, or `_get_poc_server_port_specs` on the raw `project_conf_path` YAML), keeping this post-provision call as a re-check. While there, it would help to mention in the message that another Recipe `PocEnv` (`<root>.recipe-*`) may own the ports, since `nvflare poc stop` cannot stop those.

docs/user_guide/data_scientist_guide/job_recipe.rst
```diff
@@ -327,7 +327,20 @@ Let's first set the path to the POC environment:
...
+CLI workspace at ``/tmp/nvflare/poc`` is not replaced by Recipe provisioning.
+The active path is available as ``env.poc_workspace``. Pass ``clean_up=False``
+to ``run.get_result()`` when you want to retain that workspace and its logs
+after the run. Each ``PocEnv`` instance owns one provisioning lifecycle; create
+a new instance for another deployment after provisioning has begun. POC
+deployments still share default ports and Docker participant names, so only one
+Recipe-managed POC deployment can be active on a host for a user. Also run
+``nvflare poc stop`` before starting ``PocEnv`` when the configured CLI
+deployment is active. If the Recipe process exits unexpectedly, the next
+deployment checks the recorded per-run workspace and refuses to start while any
```
holgerroth: This paragraph still describes the removed lock/record design ("checks the recorded per-run workspace", "fails closed ... if that recorded workspace is missing or unreadable"); nothing is recorded any more, and line 337-338's "only one Recipe-managed POC deployment can be active on a host for a user" is no longer what the code enforces (two deployments with distinct ports via `project_conf_path` are allowed). `recipe_api.rst` 485-494 describes the current behavior correctly, so a short paraphrase or cross-reference there would resolve the contradiction.

holgerroth (commented):
Thanks for `6e74ffc`; dropping the lock for a resource preflight is the right shape and the notebooks now match. Re-reviewed the new head. Three things I'd fix before merge, inline below: the strict-mode check at the top of `_stop()` can make `stop()` raise and skip shutdown; the port probe binds without `SO_REUSEADDR` so a TIME_WAIT port from the previous run fails the preflight and consumes the one-shot env (reproduced locally on macOS with a real grpcio bind succeeding on the same port); and `job_recipe.rst` still describes the removed workspace record. Everything else from my pass is minor and can wait.

docs/user_guide/data_scientist_guide/job_recipe.rst
chesterxgchen: Addressed in `e9134fb32`. The stale lock/record and single-active-deployment text is removed. `job_recipe.rst` now explains that deployments compete only for configured server ports, that distinct-port projects may run concurrently, and that Docker Recipe deployments use unique per-workspace container/network identities; it links to the detailed execution-environment section for lifecycle and cleanup behavior.

nvflare/recipe/poc_env.py
```diff
@@ -233,19 +306,86 @@ def _is_docker_service_running(service_name: str) -> bool:
...
 
+    @staticmethod
+    def _is_docker_service_running(service_name: str) -> bool:
+        """Return whether Docker reports the named POC container as running."""
+        return PocEnv._get_docker_service_state(service_name) is True
+
+    @staticmethod
+    def _docker_daemon_is_local() -> bool:
+        """Return whether Docker targets a verified local Unix-socket daemon."""
+        return _get_docker_endpoint(_docker_cli_env()).startswith("unix://")
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in `116849f52`. An empty discovery result now follows `start_docker.sh` semantics and is treated as local, so the loopback port preflight remains enabled. The regression parametrizes both the empty endpoint and an explicit Unix endpoint, while the remote endpoint case continues to verify that local probing is skipped.

nvflare/recipe/poc_env.py
chesterxgchen: Addressed in `e9134fb32`. The initial strict state check is now inside the stop error boundary. If metadata inspection fails, `stop()` logs the unknown state, still invokes `_stop_poc`, preserves the workspace and uncertainty flag, and reaches `finally` to clear the session manager. Deployment-failure cleanup retains its strict post-check. Added a regression that calls real `stop(clean_up=True)` twice with unreadable metadata and verifies two shutdown attempts without an exception, plus a double-stop test after a successful deployment.

nvflare/recipe/poc_env.py
chesterxgchen: Addressed in `e9134fb32`. `_is_local_port_available` now sets `SO_REUSEADDR` before binding. `PocEnv` probes default ports or ports loaded from the raw `project_conf_path` before provisioning consumes the one-shot environment, then retains the post-provision recheck for races and normalized configuration. The error now identifies another Recipe `PocEnv` as a possible owner and clarifies that `nvflare poc stop` only stops the configured CLI deployment. Added coverage for socket reuse, early custom-port rejection without consuming the env, and the post-provision race recheck.

chesterxgchen:
Addressed Review #410 and the latest Holger/Greptile threads in `e9134fb32` and `116849f52`:

- Recipe Docker deployments now use unique per-workspace container and network identities; logical participant aliases are preserved, and spawned job containers receive the same selected network
- Docker stop/liveness paths resolve logical participants only through the current deployment mapping, and Recipe Docker shutdown bypasses coordinated shutdown so a loser cannot reach or stop another deployment
- the concurrent-start regression drives two deployments through startup simultaneously and proves losing cleanup neither observes nor stops the winner containers
- local loopback port probes are skipped only for a verified remote Docker endpoint; empty endpoint discovery follows the startup script local-socket default
- port probes now set `SO_REUSEADDR`, run before provisioning consumes the one-shot env, and are rechecked after provisioning
- stop remains repeatable when service metadata is unreadable: it attempts shutdown, preserves unverified state/workspace, and always clears the session manager
- stale lock/record documentation was replaced with the current port and Docker identity behavior

The branch was explicitly rebased on current `upstream/main` before both pushes (`git rebase upstream/main`: up to date).

Validation: 47 focused PocEnv tests passed; 984 broader Recipe/POC/Docker tests passed with 19 skipped; both real POC integration tests passed; `./runtest.sh -s`, the HTML docs build, and `git diff --check` passed.

holgerroth (dismissed):
Approving at `682243d`. Verified the three pre-merge items in substance: `stop()` no longer raises on unknown state and still attempts `_stop_poc`, with tests driving the real `_stop` path twice; the port probe sets `SO_REUSEADDR` and runs before the one-shot env is consumed, with the post-provision recheck retained; and `job_recipe.rst` now matches the implemented behavior. Also read the per-workspace Docker identity change and the network-teardown retry; the affected unit suites pass locally and style is clean.

Two non-blocking notes for a follow-up:
- Several `PocEnv` unit tests now reach the real `bind()` on 8002/8003 through the pre-provision probe (they fail in a sandbox or on a dev box with a POC running). Stubbing `_preflight_ports_before_provision` in those tests would keep the suite hermetic.
- `DockerJobLauncher` lets `NVFL_DOCKER_NETWORK` override an explicitly configured `network`, while `workspace` in the same constructor treats its env var as a fallback. Harmless in generated kits since both come from the same config, but the precedence is inconsistent.

holgerroth (approved):
Re-approving at `e28d083`. The only change since my previous approval is that service-exit waiting and Docker network teardown now share one `STOP_POC_TIMEOUT` budget after the shutdown command, with the workspace preserved and a retry path when the budget runs out; the new fake-clock tests cover both outcomes. Affected unit suites pass locally and style is clean. The two non-blocking notes from the earlier approval (unit tests binding real 8002/8003, and `NVFL_DOCKER_NETWORK` precedence in `DockerJobLauncher`) still apply as follow-ups.

Issue #5209, opened 2026-08-26
Feature Request: Heterogeneity-Aware Dynamic Weighting for FedAvg


sauravsingla:
NVFlare provides FedAvg and several approaches for federated learning under heterogeneous/non-IID data, including FedProx and SCAFFOLD. However, FedAvg aggregation is primarily driven by client sample size, which may not adequately represent the information value of individual clients when their local data distributions differ significantly. A large client whose distribution is highly redundant with other participants can dominate aggregation, while a smaller client containing underrepresented patterns may have limited influence on the resulting global model.

It would be useful to support an optional heterogeneity-aware dynamic weighting strategy for FedAvg. Instead of determining aggregation coefficients only from the number of local samples, the strategy could combine sub-linear sample-size weighting with a configurable heterogeneity or representation score. For example, a client-side distribution descriptor could be compared with an aggregated reference distribution using a measure such as Jensen-Shannon divergence, while explicitly rewarding representation of globally underrepresented patterns rather than simply rewarding maximum divergence.

Conceptually, the aggregation coefficient could take the form αᵢ ∝ nᵢ^γ × Hᵢ^β × Qᵢ^δ, followed by normalization and optional minimum/maximum weight bounds. Here, nᵢ represents local sample size, Hᵢ represents a bounded heterogeneity/representation-benefit score, and Qᵢ optionally represents update or validation quality. With 0 < γ < 1, very large clients would retain an appropriate volume advantage without automatically overwhelming smaller but informative clients. The heterogeneity metric should be pluggable so that implementations could later consider label, feature, embedding, prediction, or other privacy-preserving distribution descriptors.

This would complement rather than replace FedAvg, FedProx, SCAFFOLD, or custom aggregators. FedProx and SCAFFOLD primarily address optimization/client-drift effects of non-IID data, whereas this capability would address a different question: how much influence should each heterogeneous client’s update have during aggregation? Exposing this as an optional, configurable FedAvg weighting policy could provide a reusable mechanism for highly heterogeneous federations while preserving NVFlare’s existing extensibility and default FedAvg behavior.

ZiyueXu77:
[...]

PR #5273, opened 2026-09-06
Add heterogeneity-aware dynamic aggregation (#5209)

sauravsingla:
## Summary

This draft PR implements an opt-in heterogeneity-aware dynamic client-weighting policy related to #5209.

The policy is presented as a generic aggregation policy rather than a FedOpt-specific algorithm. Two NVFlare adapters expose the same weighting logic:

- `AdaptiveHeterogeneityAggregator` for Shareable/DXO `WEIGHT_DIFF` workflows;
- `AdaptiveHeterogeneityModelAggregator` for the current unified FedAvg `ModelAggregator` workflow.

The FedOpt path remains an integration/reference smoke path. Adaptive weighting is gated on persistent distribution heterogeneity and client-performance disparity; when activation conditions are not satisfied, native local-step weighting is preserved.

## Maintainer-requested engineering updates

Following the maintainer comment, the branch now:

- separates actual training-example count from `NUM_STEPS_CURRENT_ROUND` optimizer steps;
- uses optimizer steps only for native NVFlare fallback weighting;
- uses explicit sample counts for adaptive representation and metric-reliability calculations;
- keeps per-site final weights server-side rather than placing them in global-model metadata;
- exposes constructor arguments as attributes so exported `FedJob` configuration preserves non-default settings;
- handles one-client and changing large-client cohorts without infeasible-bound crashes;
- uses unconstraining default weight bounds (`0.0`, `1.0`) while retaining optional configured bounds;
- falls back conservatively when configured bounds are infeasible for the active cohort;
- returns a valid empty `WEIGHT_DIFF` DXO no-op for an empty Shareable/DXO aggregation round;
- returns an empty DIFF no-op for the analogous empty unified-FedAvg `ModelAggregator` round;
- removes `.github/workflows/adaptive-hetero-ci.yml`;
- removes stale checked-in development results that predated the current policy;
- adds focused job-export tests for Shareable/DXO and unified FedAvg serialization paths;
- makes the FedOpt SimEnv smoke verify adaptive activation end to end by reading the final persisted server checkpoint and failing unless `adaptive_blend_factor` is non-zero, with focused regression coverage for the checkpoint assertion path.

## Standard CIFAR-10 evaluation infrastructure

The branch contains a matched Dirichlet CIFAR-10 evaluation workflow under `research/adaptive-hetero-aggregation/cifar10_evaluation/` supporting FedAvg, FedOpt (optional full-participation reference), FedProx, SCAFFOLD, FedCE, and adaptive aggregation.

For a fixed `(alpha, seed)`, NVIDIA FLARE's Dirichlet splitter creates one common training assignment. Each site's assignment is deterministically split into 90% training and 10% held-out validation. Every compared method uses the same site train/validation subsets, server initialization seed, and per-site client RNG seed.

Matched FedAvg/FedOpt/FedProx/SCAFFOLD clients preserve their NVIDIA FLARE algorithm-specific training behavior while using held-out training validation instead of reading the CIFAR-10 test set during training. Adaptive and FedCE metrics use the same held-out validation source. The official CIFAR-10 test set is reserved for one common post-training evaluator.

## Provenance and activation evidence

The evaluation protocol is now `cifar10_dirichlet_trainval_test_v3`.

Every completed result row records canonical JSON configuration plus SHA-256 hashes for:

- common dataset/model/training/execution/evaluator configuration;
- method-specific configuration (`fedprox_mu`, `fedce_mode`, adaptive policy settings, etc.);
- the full nested experiment configuration;
- the experimental condition (`alpha`, participation rate, seed).

Campaign resume only skips rows whose protocol, common-config hash and method-config hash match the current campaign. The summarizer recomputes the hashes, rejects mixed configurations, and cross-checks nested method/condition/experiment payloads. The planned campaign matrix is tracked explicitly; failed conditions are retained as missing outcomes and reported transparently rather than imputed or silently dropped.

The unified adaptive aggregator also persists federation-level activation telemetry across valid rounds:

- aggregation rounds;
- active adaptive rounds;
- activation rate;
- mean active blend factor;
- maximum observed blend factor;
- cohort-change count.

These are federation-level diagnostics only; per-site final weights remain server-side. NVFlare persists the aggregate metadata in the final checkpoint, and the common evaluator extracts it. The main results table therefore shows adaptive activation rate alongside accuracy, making a fallback-heavy partial-participation run explicit rather than silently presenting it as actively adaptive.

## Repeated runs, confidence intervals, and partial participation

The default campaign covers FedAvg, FedProx, SCAFFOLD, FedCE and adaptive aggregation across:

- Dirichlet alpha `0.1` and `0.5`;
- seeds `7, 19, 31, 43, 57`;
- full participation and 75% participation;
- 8 clients, 50 rounds, 4 local epochs.

`summarize_results.py` reports two-sided 95% Student-t confidence intervals for global and worst-client accuracy and paired adaptive-minus-baseline intervals for common seeds. FedAvg, FedProx and FedCE completed the full five-seed matched matrix. SCAFFOLD coverage is incomplete because eight planned conditions encountered numerical divergence with NaN/Inf model differences; those outcomes are reported explicitly rather than imputed. Neutral and negative results are retained.

The FedCE partial-participation run is included for completeness with its documented limitation: current NVFlare FedCE assumes full participation. The adapted client handles a site's first participation after round 0 without fabricating a previous local model.

The campaign remains an explicit research workload rather than repository-wide CI, per maintainer request.

## Current evidence status

The matched Dirichlet CIFAR-10 campaign has now been executed. Of 100 planned conditions, 92 completed successfully; the eight missing conditions are SCAFFOLD runs that encountered numerical divergence (seven at `alpha=0.1` and one at `alpha=0.5`). No failed outcomes were imputed or replaced.

Under full participation, Adaptive activated in 47/50 rounds (94%) for every seed. At `alpha=0.5`, Adaptive achieved 82.47% global accuracy and 77.23% worst-client accuracy; relative to FedProx, paired improvements were +0.66 ± 0.27 percentage points globally and +1.83 ± 0.61 percentage points for worst-client accuracy. At `alpha=0.1`, Adaptive achieved 70.71% global accuracy, with a paired +1.69 ± 1.28 percentage-point difference versus FedProx.

Under randomly changing 75% participation, the cohort-stability safeguard prevented adaptive activation, so these runs are interpreted as conservative fallback under cohort churn rather than active adaptive reweighting. At `alpha=0.1`, FedCE was strongest (65.20% versus 61.71% for Adaptive). These limiting and negative results are retained unchanged.

Overall, the current evidence supports that Adaptive is competitive with FedAvg and FedCE and provides statistically resolved positive differences over FedProx in selected full-participation settings; it does not establish universal superiority.

The maintainer separately identified a public method write-up as a requirement for eventual `research/` acceptance. That publication item remains separate from this revision.

Relates to #5209.


research/adaptive-hetero-aggregation/src/adaptive_hetero/policy.py
```diff
@@ -0,0 +1,291 @@
...
+        if candidate_blend > 0.0:
+            self._activation_streak += 1
+        else:
+            self._activation_streak = 0
+
+        blend = candidate_blend
+        if self._rounds_seen <= cfg.activation_warmup_rounds or self._activation_streak < cfg.activation_patience:
+            blend = 0.0
+
+        weights = base_weights.copy() if blend == 0.0 else (1.0 - blend) * base_weights + blend * adaptive_weights
```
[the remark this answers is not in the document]
sauravsingla: Fixed. When adaptive weighting is active, the post-blend result is now projected onto the configured bounded simplex, so the final weights—not only the adaptive candidate—respect `min_weight` and `max_weight`. Exact native FedOpt weighting is still preserved when `blend_factor == 0`. I also added a regression test with a 90% dominant base client to verify the final output cannot exceed the configured maximum.

.github/workflows/adaptive-hetero-ci.yml
```diff
@@ -0,0 +1,97 @@
...
+      - name: Partial-participation safety benchmark
+        run: |
+          python research/adaptive-hetero-aggregation/digits_benchmark.py \
+            --rounds 25 \
+            --participation-rate 0.75 \
+            --seeds 7 19 31 \
+            --settings severe extreme \
+            --models linear mlp \
+            --methods fedopt adaptive \
+            --output /tmp/adaptive-digits-partial.json
```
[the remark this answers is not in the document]
sauravsingla: Fixed. The digits benchmark now supports holding a sampled participant cohort for a configurable number of consecutive rounds, records how many rounds actually used a non-zero adaptive blend, and can fail if adaptation never activates. The partial-participation CI uses 5-round held cohorts together with `--require-adaptive-activation`, so a pure fallback-only run no longer passes this check.

.github/workflows/adaptive-hetero-ci.yml
```diff
@@ -0,0 +1,97 @@
+name: Adaptive heterogeneity research CI
+
+on:
+  push:
+    branches:
+      - feature/heterogeneity-aware-aggregation-5209
+    paths:
+      - "research/adaptive-hetero-aggregation/**"
+      - ".github/workflows/adaptive-hetero-ci.yml"
+  workflow_dispatch:
```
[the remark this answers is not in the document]
sauravsingla: Updated. The dedicated workflow now runs for relevant pull requests targeting `main` and for relevant pushes to `main`, while retaining the feature-branch push trigger during this PR. Path filters keep the workflow scoped to this research implementation and its CI definition.

research/adaptive-hetero-aggregation/nvflare_smoke/client.py
```diff
@@ -0,0 +1,143 @@
...
+    correct = 0
+    total = 0
+    with torch.no_grad():
+        for features, labels in loader:
+            logits = model(features.to(DEVICE))
+            labels = labels.to(DEVICE)
+            total_loss += float(criterion(logits, labels).item())
+            correct += int((logits.argmax(dim=1) == labels).sum().item())
+            total += int(labels.numel())
+    return total_loss / total, correct / total
```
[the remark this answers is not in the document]
sauravsingla: Fixed. `_evaluate` now raises a descriptive `ValueError` when the validation loader contains no examples instead of dividing by zero. I also added a focused unit test for the empty-loader case.

research/adaptive-hetero-aggregation/README.md
```diff
@@ -0,0 +1,200 @@
...
+feature is already production-ready. Before proposing it upstream, the strongest
+next validation would be:
+
+- run the adapter inside the current NVFlare simulator with the real
+  `FedOptRecipe`;
+- compare against NVFlare's actual FedCE implementation and the existing
+  Auto-FedRL research example rather than only the controlled FedOpt baseline;
+- add a public image benchmark such as CIFAR-10 with Dirichlet label skew;
+- add partial-participation experiments;
+- test descriptor privacy and metric comparability assumptions;
```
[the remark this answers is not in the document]
sauravsingla: Updated. The README now reflects the completed FedOpt SimEnv, FedCE protocol, partial-participation, integration, and repository validation coverage; it also documents the current stateful safeguards and remaining production-scale limitations. The older development result file is explicitly marked as historical because it predates the current hardened policy.

holgerroth:
Thanks for this, and for asking the scoping questions up front. The idea of gating adaptive weights on persistent heterogeneity plus performance disparity is a sensible one, and the FedOpt/FedCE integration work shows you've dug into the aggregation stack. Here's where we stand and how we'd like to get this in.

**Fit with `research/`.** That directory is for reference implementations of published or publicly documented methods; see [research/CONTRIBUTING.md](https://github.com/NVIDIA/NVFlare/blob/main/research/CONTRIBUTING.md) and the [sample template](https://github.com/NVIDIA/NVFlare/blob/main/research/sample-research/README.md). Right now the method isn't written up anywhere, the evidence is synthetic data and sklearn digits over five seeds with deltas mostly within one standard deviation, and the checked-in results predate the current policy (the blend factors in the JSON aren't reachable with `max_blend_factor=0.20`). So we can't merge it yet, but this looks like a good candidate for a preprint or workshop paper, and once that exists we'd be glad to take it as a research example.

**What would get it there:** compare against FedProx, SCAFFOLD, and FedCE (already in the repo) on a standard non-IID benchmark such as Dirichlet CIFAR-10 or a multi-site medical dataset, report confidence intervals, and include partial participation in the main results.

**On your design questions:**

- *Metadata contract:* descriptor plus normalized metric via DXO meta is fine. Two fixes needed: `NUM_STEPS_CURRENT_ROUND` holds optimizer steps, not samples, so the default prior of 100 nearly zeroes the metrics; and the per-site weight table in the aggregated meta gets copied into the global model and broadcast to every client.
- *FedOpt as integration point:* this is really a generic `Aggregator` that defers weighting until all updates arrive, so it applies equally to FedAvg. I'd present it that way.
- *Safeguards:* a few gaps. With defaults, one accepted client or more than 50 clients raises "weight bounds are infeasible" before the gate runs, and an all-rejected round raises from `aggregate()` and panics the job. Also, the constructor args aren't stored as attributes, so `FedJob` serializes the aggregator with empty args and the server always runs defaults, which means the SimEnv smoke only exercises the fallback path.
- *`research/` vs. component:* `research/` once validated. A reusable component is a possible later step, not the starting point.

**One request regardless:** please drop `.github/workflows/adaptive-hetero-ci.yml`. Project-specific config belongs inside the project directory, and 84 real-data training runs with a seed-dependent activation check shouldn't gate merges.

I'll convert this to a draft so it stays open while you work on the evaluation. Happy to leave the code points above as inline comments if that helps.

sauravsingla:
I’ve kept the PR as a draft and worked through the implementation points in your comment.

The branch now addresses the engineering items you raised:

- `NUM_STEPS_CURRENT_ROUND` remains the native optimizer-step weighting signal, while an explicit training-example count is used by the adaptive policy and metric prior;
- per-site final aggregation weights stay server-side rather than being placed in global-model metadata;
- the method is presented as a generic aggregation policy, with both Shareable/DXO `Aggregator` and current FedAvg `ModelAggregator` integrations;
- constructor parameters are serialization-visible and job-export tests verify non-default settings survive reconstruction;
- single-client/large-cohort bound feasibility and empty/all-rejected rounds fall back safely instead of panicking;
- the project-specific GitHub Actions workflow and stale pre-policy result JSON/table have been removed;
- the FedOpt SimEnv smoke now verifies the adaptive path end to end: after a successful run it reads the final persisted server checkpoint and fails unless the persisted `adaptive_blend_factor` is non-zero. Focused tests also cover the checkpoint assertion path.

For the requested evaluation, the branch has a matched Dirichlet CIFAR-10 harness for FedAvg, FedProx, SCAFFOLD, FedCE, and adaptive aggregation (FedOpt is available as an optional full-participation reference). Methods share the same Dirichlet assignment, deterministic 90/10 train/held-out-validation split from CIFAR-10 training data, server initialization seed, and per-site RNG seed. The official CIFAR-10 test set is reserved for one common final-checkpoint evaluator.

I also tightened the evidence pipeline before running the expensive campaign. Protocol v3 records canonical common/method/condition/full experiment configuration plus SHA-256 hashes; resume only accepts exact matching configurations, and the summarizer rejects mixed or internally inconsistent provenance. Adaptive final checkpoints persist federation-level activation telemetry (active rounds, activation rate, mean/max blend, and cohort-change count), and the main results table reports activation rate so partial-participation fallback behavior cannot be hidden. The complete method/alpha/participation/seed matrix is required before reviewer-facing CI tables are produced, and neutral/negative outcomes are retained.

No new CIFAR-10 performance numbers are claimed yet. The remaining non-publication evidence step is to execute the documented campaign and report its 95% confidence-interval/full-and-partial-participation results. The public method write-up you mentioned is being kept as a separate eventual acceptance requirement and is not part of this revision.

sauravsingla:
We completed the matched Dirichlet CIFAR-10 evaluation you suggested and are sharing the results transparently, including neutral and negative outcomes.

We evaluated Adaptive, FedAvg, FedProx, FedCE and SCAFFOLD at α=0.1 and α=0.5, using five fixed seeds, 8 clients, 50 rounds, full and 75% participation, and a common held-out/test evaluation protocol. The CIFAR-10 test set was reserved for the common post-training evaluator, and we report two-sided 95% Student-t confidence intervals together with paired Adaptive-minus-baseline intervals.

Under full participation, the adaptive path activated in 47/50 rounds for every Adaptive run. At α=0.5, Adaptive achieved 82.47% global accuracy and 77.23% worst-client accuracy. Relative to FedProx, the paired improvements were +0.66 ± 0.27 pp global and +1.83 ± 0.61 pp worst-client. Relative to FedAvg and FedCE, the mean differences were positive but the confidence intervals include zero, so we interpret those results as competitive rather than statistically superior.

At α=0.1/full participation, Adaptive achieved 70.71%, compared with FedAvg 70.61%, FedCE 70.32% and FedProx 69.02%. The paired Adaptive-minus-FedProx difference was +1.69 ± 1.28 pp.

Partial participation produced an important limiting result. With a newly sampled 75% cohort each round, the cohort-stability safeguard prevented adaptive activation, giving 0% activation in all ten Adaptive partial-participation runs. We therefore interpret those runs as demonstrating conservative fallback under cohort churn rather than active adaptive reweighting. At α=0.1/75%, FedCE was also the strongest method, at 65.20% versus 61.71% for Adaptive, and we retain that result unchanged.

We attempted all 100 planned conditions. 92 completed successfully. Eight SCAFFOLD configurations encountered numerical divergence with NaN/Inf model differences; seven were at α=0.1 and one at α=0.5. We have not imputed or replaced those missing outcomes.

sauravsingla:
we have now completed the matched Dirichlet CIFAR-10 evaluation.

We evaluated Adaptive, FedAvg, FedProx, FedCE and SCAFFOLD at α=0.1 and α=0.5, using five fixed seeds, 8 clients, 50 rounds, and both full and 75% participation. We report two-sided 95% Student-t confidence intervals and paired Adaptive-minus-baseline intervals.

Under full participation, Adaptive was competitive with FedAvg and FedCE and showed statistically resolved improvements over FedProx in selected settings. At α=0.5, Adaptive achieved 82.47% global accuracy and 77.23% worst-client accuracy. Relative to FedProx, the paired improvements were +0.66 ± 0.27 pp global and +1.83 ± 0.61 pp worst-client. At α=0.1/full participation, Adaptive achieved 70.71%, compared with 70.61% FedAvg, 70.32% FedCE and 69.02% FedProx; the paired Adaptive-minus-FedProx difference was +1.69 ± 1.28 pp.

We also retained the limiting and negative results. Under randomly changing 75% participation, the cohort-stability safeguard prevented adaptive activation, with 0/50 adaptive rounds across those runs. We therefore interpret these experiments as validating conservative fallback under cohort churn rather than active adaptive weighting. At α=0.1/75%, FedCE was strongest at 65.20%, versus 61.71% for Adaptive.

We attempted all 100 planned conditions. 92 completed successfully. Eight SCAFFOLD configurations encountered numerical divergence with NaN/Inf model differences — seven at α=0.1 and one at α=0.5. We have not imputed or replaced those missing outcomes and will report the incomplete SCAFFOLD coverage explicitly.

Under full participation, the adaptive path itself was exercised consistently, activating in 47/50 rounds (94%) for every Adaptive run.

The engineering points from your earlier review have also been addressed in the branch, including separation of training-example counts from optimizer steps, server-side handling of final per-site weights, generic Aggregator/FedAvg integration, serialization safeguards, safe fallback behavior, and removal of the project-specific workflow.

sauravsingla:
### CIFAR-10 benchmark summary

| Dirichlet α | Participation | Method | Seeds completed | Global accuracy (%) | Worst-client accuracy (%) | Adaptive activation |
|---|---:|---|---:|---:|---:|---:|
| 0.1 | 75% | Adaptive | 5 | 61.71 ± 7.69 | 23.57 ± 22.58 | 0% |
| 0.1 | 75% | FedAvg | 5 | 61.26 ± 9.30 | 23.74 ± 23.85 | — |
| 0.1 | 75% | FedProx | 5 | 62.60 ± 6.39 | 24.49 ± 21.12 | — |
| 0.1 | 75% | FedCE | 5 | **65.20 ± 4.48** | **30.50 ± 21.43** | — |
| 0.1 | 75% | SCAFFOLD | 0 | — | — | — |
| 0.1 | 100% | Adaptive | 5 | **70.71 ± 1.56** | 50.59 ± 11.32 | **94%** |
| 0.1 | 100% | FedAvg | 5 | 70.61 ± 1.99 | 49.50 ± 14.66 | — |
| 0.1 | 100% | FedProx | 5 | 69.02 ± 2.38 | 43.97 ± 17.43 | — |
| 0.1 | 100% | FedCE | 5 | 70.32 ± 2.76 | **53.28 ± 15.59** | — |
| 0.1 | 100% | SCAFFOLD | 3 | 11.29 ± 5.55 | 0.05 ± 0.21 | — |
| 0.5 | 75% | Adaptive | 5 | **81.00 ± 1.45** | 73.32 ± 6.01 | 0% |
| 0.5 | 75% | FedAvg | 5 | 80.78 ± 1.79 | 72.21 ± 6.06 | — |
| 0.5 | 75% | FedProx | 5 | 80.85 ± 1.57 | **73.63 ± 4.52** | — |
| 0.5 | 75% | FedCE | 5 | 80.83 ± 1.85 | 73.41 ± 5.99 | — |
| 0.5 | 75% | SCAFFOLD | 5 | 10.00 ± 0.00 | 0.28 ± 0.49 | — |
| 0.5 | 100% | Adaptive | 5 | **82.47 ± 0.66** | **77.23 ± 0.74** | **94%** |
| 0.5 | 100% | FedAvg | 5 | 82.23 ± 0.97 | 76.60 ± 0.95 | — |
| 0.5 | 100% | FedProx | 5 | 81.81 ± 0.74 | 75.40 ± 0.99 | — |
| 0.5 | 100% | FedCE | 5 | 82.22 ± 0.67 | 76.28 ± 2.58 | — |
| 0.5 | 100% | SCAFFOLD | 4 | 10.00 ± 0.00 | 0.49 ± 0.81 | — |

Values are **mean ± 95% CI half-width**, using two-sided Student-t intervals across matched seeds.

### Paired Adaptive-minus-baseline results

| α | Participation | Comparison | Global Δ (pp) | Worst-client Δ (pp) |
|---|---:|---|---:|---:|
| 0.1 | 75% | Adaptive − FedAvg | +0.45 ± 1.99 | -0.17 ± 3.99 |
| 0.1 | 75% | Adaptive − FedProx | -0.89 ± 2.49 | -0.92 ± 4.00 |
| 0.1 | 75% | Adaptive − FedCE | -3.48 ± 5.10 | **-6.93 ± 5.17** |
| 0.1 | 100% | Adaptive − FedAvg | +0.10 ± 0.89 | +1.09 ± 5.02 |
| 0.1 | 100% | Adaptive − FedProx | **+1.69 ± 1.28** | +6.62 ± 7.04 |
| 0.1 | 100% | Adaptive − FedCE | +0.39 ± 2.24 | -2.69 ± 4.99 |
| 0.5 | 75% | Adaptive − FedAvg | +0.22 ± 0.67 | +1.12 ± 2.00 |
| 0.5 | 75% | Adaptive − FedProx | +0.15 ± 0.72 | -0.31 ± 1.94 |
| 0.5 | 75% | Adaptive − FedCE | +0.17 ± 0.76 | -0.09 ± 2.25 |
| 0.5 | 100% | Adaptive − FedAvg | +0.24 ± 0.59 | +0.63 ± 1.12 |
| 0.5 | 100% | Adaptive − FedProx | **+0.66 ± 0.27** | **+1.83 ± 0.61** |
| 0.5 | 100% | Adaptive − FedCE | +0.25 ± 0.59 | +0.95 ± 2.26 |

### Interpretation

- Under full participation, Adaptive was consistently competitive with FedAvg and FedCE.
- The clearest statistically resolved positive differences were against FedProx:
  - α=0.1 / 100% participation: **+1.69 ± 1.28 pp global**
  - α=0.5 / 100% participation: **+0.66 ± 0.27 pp global**
  - α=0.5 / 100% participation: **+1.83 ± 0.61 pp worst-client**
- At α=0.5 / full participation, Adaptive achieved the highest global and worst-client means: **82.47%** and **77.23%**.
- At α=0.1 / 75% participation, FedCE was strongest; we retain this negative result unchanged.
- With randomly changing 75% participation, the cohort-stability safeguard prevented adaptive activation. These runs therefore validate conservative fallback under cohort churn rather than active adaptive reweighting.
- Under full participation, Adaptive activated in **47/50 rounds (94%)** for every seed, confirming that the active policy path was exercised.

### Campaign completeness

We attempted all **100 planned benchmark conditions**.

- **92 completed successfully**
- **8 SCAFFOLD conditions encountered numerical divergence with NaN/Inf model differences under this benchmark configuration**
- 7 of those failures occurred at α=0.1
- 1 occurred at α=0.5
- No failed SCAFFOLD outcomes were imputed or replaced

We therefore treat the FedAvg, FedProx and FedCE comparisons as complete five-seed matched comparisons, while reporting the incomplete SCAFFOLD coverage explicitly.

Overall, we interpret the results as evidence that Adaptive is competitive with strong baselines, provides measurable improvements over FedProx in selected full-participation regimes, and falls back conservatively when the cohort is unstable. We do not claim universal superiority.

research/adaptive-hetero-aggregation/src/adaptive_hetero/nvflare_aggregator.py
```diff
@@ -0,0 +1,298 @@
...
+            quality_improvement=quality,
+        )
+        return True
+
+    def aggregate(self, fl_ctx: FLContext) -> Shareable:
+        if not self._contributions:
+            self.log_warning(fl_ctx, "no valid contributions were accepted for this aggregation round")
+            self.last_weights = {}
+            self.reset(fl_ctx)
+            return make_reply(ReturnCode.EMPTY_RESULT)
```
chesterxgchen: **Crash on empty round:** `make_reply(ReturnCode.EMPTY_RESULT)` produces a `Shareable` with no DXO `content_type` header. `ScatterAndGather.control_flow` (nvflare/app_common/workflows/scatter_and_gather.py:298-308) passes this straight to `shareable_gen.shareable_to_learnable(aggr_result, fl_ctx)` with no return-code check, which calls `DXO.from_shareable()` (nvflare/apis/dxo.py:166-168) and raises `ValueError("the shareable is not a valid DXO - expect content_type DXO but got {content_type}")`. The intended graceful 'empty round' handling crashes the round instead.

research/adaptive-hetero-aggregation/src/adaptive_hetero/model_aggregator.py
```diff
@@ -0,0 +1,299 @@
...
+    def _finite_positive(meta: dict, key: str, client_name: str) -> float:
+        try:
+            value = float(meta[key])
+        except (KeyError, TypeError, ValueError, OverflowError) as exc:
+            raise ValueError(f"client {client_name!r} is missing valid {key!r} metadata") from exc
+        if not math.isfinite(value) or value <= 0.0:
+            raise ValueError(f"client {client_name!r} must provide finite positive {key!r} metadata")
+        return value
+
+    def accept_model(self, model: FLModel):
```
chesterxgchen, line 148: **Bare `ValueError` aborts the whole round instead of rejecting one client:** `accept_model()` raises a bare `ValueError` on malformed client metadata (e.g. `CLIENT_METRIC` out of `[0,1]`, empty `DISTRIBUTION_DESCRIPTOR`). The base `ModelAggregator.accept()` (nvflare/app_common/aggregators/model_aggregator.py:55-58) calls this unguarded and otherwise returns `True`; the exception propagates up through `_accept_train_result`/`_process_train_result` to the top-level `except Exception` in `control_flow` (scatter_and_gather.py:344-346), aborting the entire round. This is inconsistent with the sibling `AdaptiveHeterogeneityAggregator.accept()` in `nvflare_aggregator.py`, which catches the same class of bad input and returns `False` so only that client is rejected.

research/adaptive-hetero-aggregation/src/adaptive_hetero/policy.py
```diff
@@ -0,0 +1,326 @@
...
+            if mean_heterogeneity <= cfg.heterogeneity_deadband
+            else _sigmoid((mean_heterogeneity - cfg.heterogeneity_threshold) / cfg.heterogeneity_temperature)
+        )
+        performance_gate = (
+            0.0
+            if metric_gap <= cfg.performance_gap_deadband
+            else _sigmoid((metric_gap - cfg.performance_gap_threshold) / cfg.performance_gap_temperature)
+        )
+
+        bounds_feasible = bool(
```
chesterxgchen, line 287: **Epsilon tolerance mismatch between `bounds_feasible` and `project_bounded_simplex`:** this check uses the configurable `cfg.epsilon`, but `project_bounded_simplex` (called immediately after) hardcodes `1e-12` for the same check (policy.py:123-124). A looser custom `epsilon` can make `bounds_feasible` return `True` while `project_bounded_simplex` still raises `ValueError("weight bounds are infeasible...")`, defeating the documented conservative fallback. Currently unreachable through the two shipped adapters since neither exposes `epsilon` as a constructor argument — but latent for any future caller that does.

research/adaptive-hetero-aggregation/src/adaptive_hetero/policy.py
```diff
@@ -0,0 +1,326 @@
...
+            cfg.min_weight * counts.size <= 1.0 + cfg.epsilon
+            and cfg.max_weight * counts.size >= 1.0 - cfg.epsilon
+        )
+        if bounds_feasible:
+            adaptive_weights = project_bounded_simplex(raw, cfg.min_weight, cfg.max_weight)
```
chesterxgchen: Minor/efficiency: `project_bounded_simplex()` always runs a 100-iteration bisection even when bounds are non-binding (e.g. the default `[0.0, 1.0]`), where the projection is a mathematical no-op. A short-circuit (`if lower <= 0.0 and upper >= 1.0: return raw`) would avoid the unnecessary work on the common path.

research/adaptive-hetero-aggregation/cifar10_evaluation/evaluate_result.py
```diff
@@ -0,0 +1,224 @@
...
+
+
+def find_server_checkpoint(workspace: str) -> Path:
+    """Find the final server checkpoint produced by an NVFlare simulation."""
+
+    workspace_path = Path(workspace).resolve()
+    server_dir = workspace_path / "server" / "simulate_job" / "app_server"
+    # Prefer final-round model for an identical endpoint across methods. Best
+    # checkpoints can depend on method-specific client metrics/model selection.
+    for name in ("FL_global_model.pt", "best_FL_global_model.pt"):
```
chesterxgchen: Hardcodes the literal checkpoint filenames `"FL_global_model.pt"` / `"best_FL_global_model.pt"` instead of reusing `nvflare.app_common.app_constant.DefaultCheckpointFileName`, which this same PR already imports and uses correctly in `nvflare_smoke/job.py:32,61`. Two sources of truth for the same constant within one PR; a rename of `DefaultCheckpointFileName.GLOBAL_MODEL` would silently break this file while `nvflare_smoke/job.py` keeps working.

research/adaptive-hetero-aggregation/cifar10_evaluation/summarize_results.py
```diff
@@ -0,0 +1,455 @@
...
+
+def render_markdown(summary: dict) -> str:
+    """Render full and partial participation together as main results."""
+
+    lines = [
+        "# CIFAR-10 Main Results",
+        "",
+        "Values are means across matched seeds with two-sided 95% Student-t confidence intervals.",
+        "Global and worst-client accuracy use the common post-training evaluator.",
+        "Adaptive activation rate is the fraction of valid aggregation rounds with a non-zero blend; a low rate makes conservative fallback behavior explicit.",
```
chesterxgchen: Line exceeds the repo's black line-length limit of 120 (this line is ~160 chars). `research/adaptive-hetero-aggregation/run.py:121` has the same issue (~121 chars). Both would fail a black/flake8 check per CLAUDE.md.

chesterxgchen (commented):
Automated review findings (Claude Code) — see inline comments.

research/adaptive-hetero-aggregation/src/adaptive_hetero/nvflare_aggregator.py
sauravsingla: Addressed. Empty Shareable/DXO rounds now return a valid empty `WEIGHT_DIFF` DXO no-op, so `shareable_to_learnable()` can consume the aggregation result without `DXO.from_shareable()` failing. Updated the regression test to parse the returned Shareable and assert the empty `WEIGHT_DIFF` contract.

research/adaptive-hetero-aggregation/src/adaptive_hetero/model_aggregator.py
sauravsingla: Addressed. The unified `AdaptiveHeterogeneityModelAggregator.accept()` now catches malformed client-result `TypeError`/`ValueError` and returns `False`, matching the sibling Shareable/DXO adapter's reject-one-client behavior. Added regression coverage for an out-of-contract client metric.

research/adaptive-hetero-aggregation/src/adaptive_hetero/policy.py
sauravsingla: Addressed. `project_bounded_simplex()` now accepts the caller tolerance and uses `cfg.epsilon` consistently for feasibility and numerical checks; both policy call sites pass `cfg.epsilon`. Added tolerance regression coverage.

research/adaptive-hetero-aggregation/src/adaptive_hetero/policy.py
sauravsingla: Addressed. Added a safe fast path for non-binding `[0, 1]`-style bounds when the input is already a valid normalized simplex vector, avoiding the 100-iteration bisection on the common path. Added an exact no-op regression test.

research/adaptive-hetero-aggregation/cifar10_evaluation/evaluate_result.py
sauravsingla: Addressed. `evaluate_result.py` now reuses `DefaultCheckpointFileName.GLOBAL_MODEL` and `.BEST_GLOBAL_MODEL` instead of duplicating the literal checkpoint filenames.

research/adaptive-hetero-aggregation/cifar10_evaluation/summarize_results.py
sauravsingla: Addressed both line-length findings: wrapped the long reviewer-facing activation text in `summarize_results.py` and the FedOpt partial-participation error in `cifar10_evaluation/run.py`.

sauravsingla:
Thanks Chester for the detailed review. I’ve addressed all six findings and added targeted regression coverage; the latest NVIDIA pre-merge, unit, style, coverage and CodeQL checks are now passing.

PR #5274, opened 2026-09-08
[Research] Add FedRevive CollabAPI research example

ZiyueXu77:
## What this PR does

Adds a self-contained FedRevive research contribution implemented with the NVIDIA FLARE Collab API for CIFAR-10.

- Provides one workflow for FedAvg, FedBuff, and FedRevive through configurable concurrency K, aggregation buffer B, and distribution threshold O.
- Implements server-side data-free knowledge distillation with a bounded teacher buffer.
- Exposes class-proportion source and synthetic-generation interval as independent FedRevive options.
- Supports the shifted delay schedule from the paper as a deterministic, pluggable configuration.
- Uses disk-backed returned models and reference-counted global snapshots plus bounded physical workers to reproduce arbitrary seeded arrival sequences without exhausting host memory.
- Documents setup, data preparation, execution, centralized evaluation, expected results, and resource controls.
- Includes FLARE-only CIFAR-10 curves for the default and shifted delay schedules.

Paper: https://arxiv.org/abs/2511.00655

## Results included in the example

The README includes two single-seed figures generated only from this CollabAPI implementation:

1. FedAvg, FedBuff, and FedRevive over 200 simulated-time units under the default delay schedule.
2. FedBuff and FedRevive over 100 simulated-time units under the shifted delay schedule.

| Schedule | Method | Final accuracy | Best accuracy |
|---|---|---:|---:|
| Default | FedAvg | 0.4022 | 0.4056 |
| Default | FedBuff | 0.6111 | 0.6659 |
| Default | FedRevive | 0.7612 | 0.7712 |
| Shifted | FedBuff | 0.3019 | 0.4886 |
| Shifted | FedRevive | 0.7917 | 0.8325 |

## FedRevive experiment controls

The two dimensions are configured directly rather than represented by named presets:

- --class-proportion-source {true-histogram,estimated}
- --generation-interval N

The README demonstrates true-histogram with interval 1 and estimated with interval 10. Other combinations remain valid because the options are independent.

## Development-time validation

The paper authors have an internal reference implementation that is not open-sourced. It was used only for development-time curve alignment and is not included, referenced, or required by the research example.

Under the shifted schedule, the internal comparison produced a smoothed RMSE of 0.0604 for FedBuff and 0.0172 for FedRevive. The example itself reports only its FLARE results.

## Validation

- Python compilation for all research/fedrevive modules
- Black check
- isort check
- flake8 check
- NVFlare license-header check
- CLI parsing for both independent experiment options
- Mapping check for true/estimated class proportions and generation intervals 1/10
- Full CIFAR-10 Collab runs for FedAvg, FedBuff, and FedRevive
- Shifted-delay Collab runs for FedBuff and FedRevive

The repository-level ./runtest.sh -s wrapper could not complete in this workspace because it attempts to install tooling into an externally managed, read-only Python environment. Its applicable scoped style and license checks were run directly and passed.


research/fedrevive/dfkd.py
```diff
@@ -0,0 +1,458 @@
...
+        self.generator = DFKDGenerator(
+            latent_size=self.config.latent_size,
+            features=self.config.generator_features,
+        ).to(device)
+        self.generator.train()
+        self.meta_optimizer = torch.optim.Adam(
+            self.generator.parameters(),
+            lr=self.config.generator_lr,
+            betas=(0.5, 0.999),
+            eps=0,
```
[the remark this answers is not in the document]
ZiyueXu77: Fixed in 61b59131. Removed eps=0 so Adam uses its default positive epsilon, avoiding division-by-zero and non-finite meta-optimizer updates.

research/fedrevive/requirements.txt
```diff
@@ -0,0 +1,5 @@
+kornia
+numpy
+tensorboard
+torch
+torchvision
```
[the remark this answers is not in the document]
ZiyueXu77: Fixed in 61b59131. The example now requires torch>=2.1.0, where torch.load mmap support is available, and the corresponding torchvision>=0.16.0 minimum. The other requirements remain because each is used directly by the example.

research/fedrevive/job.py
```diff
@@ -0,0 +1,268 @@
...
+# long-lived worker may inherit a host-wide BLAS/OpenMP pool.  Multiplying those
+# threads and their stacks across simulator processes can drain host resources
+# even though logical training is intended to be sequential per physical site.
+for _variable in (
+    "OMP_NUM_THREADS",
+    "MKL_NUM_THREADS",
+    "OPENBLAS_NUM_THREADS",
+    "NUMEXPR_NUM_THREADS",
+):
+    os.environ[_variable] = "1"
```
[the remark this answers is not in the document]
ZiyueXu77: Fixed in 61b59131. These variables now use setdefault: the resource-safe value of 1 remains the default, while explicit launcher or container settings are preserved.

research/fedrevive/server.py
```diff
@@ -0,0 +1,945 @@
...
+            if not job.logical_released:
+                self._available_logical.append(job.logical_name)
+                self._available_logical.sort(key=lambda name: int(name.split("-", 1)[1]))
+            self._num_open_slots += 1
+            if not accept_update:
+                self._release_outcome_result(outcome)
+                return
+            if outcome.error is not None:
+                self.logger.warning(f"assignment {job.assignment_id} failed: {outcome.error}")
+                return
```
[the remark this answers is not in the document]
ZiyueXu77: Thanks. This is a valid failure-path concern, but retrying or reconstructing a partial synchronous cohort is outside the scope of this research example. The paper-aligned experiments assume successful client calls, and the existing base-version guard prevents a mixed cohort from being applied. We are keeping the example focused rather than adding production recovery machinery.

research/fedrevive/dfkd.py
```diff
@@ -0,0 +1,466 @@
...
+
+        if self.config.should_generate(current_version):
+            metrics = self._synthesize(teachers, distributions, self.student_model, current_version)
+            metrics["synthesis_performed"] = True
+        else:
+            # The periodic configuration reuses the bounded synthetic pool
+            # between generator updates while still performing KD for every
+            # eligible client arrival. No model snapshot is retained for reuse.
+            metrics = {"synthesis_performed": False, "synthesis_time": 0.0}
+        if current_version > self.config.warmup_versions:
```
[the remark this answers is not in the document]
ZiyueXu77: Thanks, fixed in 57f5cd0d. `revive()` now returns no distilled update during synthesis-only warmup, and only returns a model delta after at least one KD optimizer step. Synthesis still proceeds during warmup, while `process_update()` leaves the ordinary client delta unchanged until KD has actually run.

research/fedrevive/server.py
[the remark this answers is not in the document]
ZiyueXu77: Follow-up: added a minimal fail-fast check in 789d84c9. If a FedAvg assignment fails or returns no result, the workflow now raises a clear incomplete-cohort error immediately instead of allowing buffered successes to mix with a later cohort. This intentionally adds no retry or cohort-recovery mechanism.

research/fedrevive/server.py
```diff
@@ -0,0 +1,951 @@
...
+
+        physical_name = group_call_context.target_name.split(".", 1)[0]
+        assignment_id = int(result[2]["assignment_id"])
+        self._outcomes.put(_ClientOutcome(physical_name=physical_name, assignment_id=assignment_id, result=result))
+        # Returning None avoids retaining or transforming a second model-sized
+        # callback result.  The scheduler is the sole owner after queue.put().
+        return None
+
+    def _watch_call_failures(self, results, assignment_by_physical):
+        for _ in results:
```
holgerroth: **Failure reporting is delayed until the whole group call finishes.** `for _ in results: pass` iterates the `ResultQueue`, whose `__next__` skips the failure marker and blocks on `q.get()` until `num_whole_items_received >= limit`. So when site-1 fails immediately (e.g. `FileNotFoundError` for a missing shard) but site-2 in the same group call is still training or hung, site-1's `_ClientOutcome(error=...)` is only queued after site-2 returns (up to `call_timeout`, default 3600 s). Meanwhile site-1 stays in `_physical_jobs` and is never returned to `_available_physical`, and the FedAvg "cohort incomplete" error is delayed the same way.

`results.failures` is populated incrementally (`group_call_context.py` `append_failure`), so the watcher can poll it (or check it inside the loop) and emit the error outcome as soon as an entry appears.

research/fedrevive/server.py
```diff
@@ -0,0 +1,951 @@
...
+
+    @staticmethod
+    def _release_outcome_result(outcome: _ClientOutcome):
+        """Drop one queued result and unlink any lazy tensor backing files."""
+
+        result = outcome.result
+        # Break the owning reference first so an exception cannot leave a full
+        # result reachable through the scheduler bookkeeping.
+        outcome.result = None
+        if isinstance(result, (tuple, list)) and result and isinstance(result[0], LazyTensorDict):
```
holgerroth: **This branch is never taken, so the explicit cleanup is dead code** (same at the `finally` in `_process_outcome`). `TensorDecomposer` is registered per `torch.Tensor`, and `ViaDownloaderDecomposer.recompose` returns `items.make_lazy_ref(item_id)` per tensor while the `LazyTensorDict` itself stays in `fobs_ctx`. The decoded reply is therefore `(dict[str, _LazyRef], metrics, metadata)` and `isinstance(result[0], LazyTensorDict)` is always False. `_materialize_result` below already assumes `_LazyRef` values, so the two paths are inconsistent.

Consequence: temp dirs are only removed by `_TempDirRef.__del__` once every ref is collected, so any lingering reference (exception traceback, `fobs_ctx`) keeps per-client `nvflare_tensors_*` dirs alive until the final rmtree. The README's "deletes its files immediately" / "cleaned through `_release_outcome_result()`" overstates what happens. Suggest releasing via the shared `_temp_ref` of any `_LazyRef` value, or adding a public `release()` to `_LazyRef` in `lazy_tensor_dict.py`.

research/fedrevive/server.py
```diff
@@ -0,0 +1,951 @@
...
+            self._num_open_slots += 1
+            if not accept_update:
+                self._release_outcome_result(outcome)
+                return
+            if outcome.error is not None:
+                if self.method is Method.FEDAVG:
+                    raise RuntimeError(
+                        f"FedAvg cohort incomplete: assignment {job.assignment_id} failed: {outcome.error}"
+                    )
+                self.logger.warning(f"assignment {job.assignment_id} failed: {outcome.error}")
```
holgerroth: **Failed assignments are silently dropped under FedBuff/FedRevive and never counted.** `_record_physical_outcome` already re-added the failing physical site to `_available_physical` before the error is inspected, so a site with a broken `data_root` or persistent CUDA OOM is re-dispatched indefinitely. Each failure still consumed its simulated download/train/upload events, so the arrival stream the benchmark is supposed to reproduce is thinned, yet `results.json` and `accuracy_history.json` carry no failure count and the README does not mention this tolerance. A run with half its arrivals lost is indistinguishable from a healthy one.

Minimal fix: count failed assignments (ideally per physical site), log the total and write it to `results.json`, and optionally abort past a threshold.

research/fedrevive/server.py
```diff
@@ -0,0 +1,951 @@
...
+                )
+                if self._simulated_time >= self.max_time:
+                    return None
+                continue
+            # Upload is the only phase that yields a model update.  It may have
+            # completed physically much earlier and waited on disk.
+            outcome = self._completed_outcomes.pop(assignment_id, None)
+            if outcome is not None:
+                heapq.heappop(self._event_heap)
+                self._simulated_time = event.finish_time
```
holgerroth, line 830: **The upload branch does not enforce `max_time`.** The download (L788) and train (L806) branches return `None` once `_simulated_time >= max_time`, but here the upload is popped, the clock is advanced, and the outcome is returned, so `execute()` aggregates a new global version past the budget. Example: an upload accepted at t=198 with budget 200 leaves another upload at t=205 at the heap top; the loop re-enters (198 < 200), aggregates at t=205, and `results.json` reports `simulated_time=205` with the final eval on that over-budget version. The dispatch guard at L554 uses the same check, so the asymmetry looks unintended. Add the same guard here.

research/fedrevive/job.py
```diff
@@ -0,0 +1,278 @@
...
+def validate_args(args):
+    if args.num_clients < 1 or args.num_logical_clients < 1:
+        raise ValueError("--num-clients and --num-logical-clients must be >= 1")
+    if args.max_time <= 0 or args.local_iterations < 1 or args.local_batch_size < 1:
+        raise ValueError("--max-time, --local-iterations, and --local-batch-size must be positive")
+    if args.local_lr <= 0 or args.call_timeout <= 0 or args.num_workers < 0 or args.max_parallel < 0:
+        raise ValueError(
+            "--local-lr and --call-timeout must be positive; --num-workers and --max-parallel must be nonnegative"
+        )
+    if args.max_parallel > args.num_clients:
```
holgerroth: **`--max-parallel` does not do what the help text and README (L112-113) say.** Collab's `parallel` is scoped to one group call: `Group` creates a fresh `ResultWaiter` per call and `wait_for_call_permission` counts only that waiter. So:
- `--num-clients 4 --max-parallel 2` with all pending assignments on one snapshot issues one group call to 4 sites with `parallel=2`; sites 3-4 are already marked busy in `_physical_jobs` but their RPCs wait for 1-2 to return, halving utilization.
- Pending work spanning two snapshots issues two calls with independent budgets, so 4 RPCs are in flight, exceeding the advertised cap.
- Values above `--num-clients` are inert, so this `ValueError` is unnecessary and makes `--num-clients 1` fail with the default.

The effective bound on in-flight RPCs is simply the physical-site count. Suggest dropping the option (pass `parallel=0`) and the README claim.

research/fedrevive/server.py
```diff
@@ -0,0 +1,951 @@
...
+                total += labels.size(0)
+        record = {
+            "model_version": self._model_version,
+            "simulated_time": self._simulated_time,
+            "accuracy": correct / total,
+            "test_loss": loss_sum / total,
+            "final": final,
+        }
+        self._history.append(record)
+        self._writer.add_scalar("test/accuracy", record["accuracy"], self._simulated_time)
```
holgerroth: **Float `global_step`.** `FileWriter.add_event` does `event.step = int(step)`, so every evaluation inside one simulated-time unit lands on the same integer step. With eval every 10 versions that is ~3 points per step for FedBuff, ~6 for FedRevive, and ~50 for the shifted-schedule FedRevive run, which renders as a vertical saw-tooth rather than a curve (`accuracy_history.json` is unaffected). Use `self._model_version` as the step and, if you want the time axis, pass `walltime=self._simulated_time` or log a second scalar.

research/fedrevive/server.py
```diff
@@ -0,0 +1,951 @@
...
+        """
+
+        # Keep the offload tree with the user-selected persistent workspace,
+        # not /tmp (which may be a small tmpfs).  Change tempfile configuration
+        # only while the NVFlare helper allocates its tree, and restore it even
+        # if setup fails.  cleanup_tensor_disk_offload handles every exit path.
+        previous_tempdir = tempfile.tempdir
+        tempfile.tempdir = self._run_dir
+        try:
+            context = setup_tensor_disk_offload(
```
holgerroth: **Process-global `tempfile.tempdir` swap.** This is restored in `finally`, but the call-watcher threads (L533) and the `SummaryWriter` (L326) already exist at this point, so any `tempfile.*` call on another thread during the window lands under the run dir, and the trick silently stops working if the helper stops using `mkdtemp`. The underlying need (a small tmpfs `/tmp`) is shared by `fedavg.py`, `scatter_and_gather.py` and `scaffold.py`, which call the same helper. A cleaner fix is an optional `root_dir` argument on `setup_tensor_disk_offload` in `nvflare/app_common/utils/tensor_disk_offload_context.py`, passed as `dir=` to `mkdtemp`. Happy to see that as a small core change alongside this PR.

research/fedrevive/requirements.txt
```diff
@@ -0,0 +1,5 @@
+kornia
```
holgerroth, line 2: **Missing NVFlare pin and license notes.** `research/CONTRIBUTING.md` asks for "Requirements and the NVFlare version used", "License notes", and to "call out any third-party code, models, datasets, or assets", and `AGENTS.md` says the requirements pin is sufficient and not to add install-from-source caveats. Currently there is no `nvflare` entry here, the README (L248-252) says to use a `main` checkout with `pip install -e ../..`, has no License section, and does not mention CIFAR-10 or kornia terms. Every imported module (`nvflare.collab`, `LazyTensorDict`, the disk-offload context, ...) exists on the 2.9 line, so `nvflare[PT]~=2.9.0rc` like the sibling projects would work. Please also add a FedRevive row to the `research/README.md` index table.

research/fedrevive/server.py
```diff
@@ -0,0 +1,951 @@
...
+        torch.set_num_threads(1)
+        torch.set_num_interop_threads(1)
+        random.seed(self.setup_seed)
+        np.random.seed(self.setup_seed)
+        torch.manual_seed(self.setup_seed)
+        self._init_outputs()
+        self._snapshot_store = _SnapshotStore(os.path.join(self._run_dir, "assignment_snapshots"))
+        self._init_data()
+        initial_model = get_model_params(create_model(), target_device="cpu")
+        inherited_model = collab.get_prop(ContextKey.RESULT, initial_model)
```
holgerroth: Nit: `collab.get_prop(ContextKey.RESULT, ...)` inside the single `@collab.main` can never return anything but `initial_model`. The runtime sets `ContextKey.RESULT` only after the main returns (`lifecycle.py`: `result = main_func(**kwargs); server_ctx.set_prop(ContextKey.RESULT, result)`), and `job.py` provides no other writer. If warm-start is intended, load from an explicit app prop or file; otherwise drop the line so readers do not expect it to work.

holgerroth (commented):
Thanks for the contribution. I went through the new example with the Collab runtime, `LazyTensorDict`, and the disk-offload helper side by side. Style, license headers, and compilation are all clean. Inline comments below cover the issues I would want addressed before merge: the failure-watcher blocking, the dead cleanup path, the uncounted dropped assignments, the `max_time` gap on uploads, the `--max-parallel` semantics, the TensorBoard step, the `tempfile.tempdir` swap, and the missing NVFlare pin / license notes.

Two smaller observations not worth their own thread: with `--buffer-size` above 1, synthesis runs once per eligible arrival rather than once per version (`should_generate` is a pure function of the version), and the DFKD boundaries use `>=` at version 50 and 100 for the freeze/adversarial gates but `>` at 100 for KD.

research/fedrevive/server.py
ZiyueXu77: Fixed in 5a6dbfad. The bounded watcher now polls the ResultQueue incremental failure state under its update lock and emits each site failure once, without waiting for the remaining group members. Success callbacks and simulated acceptance ordering are unchanged.

research/fedrevive/server.py
ZiyueXu77: Fixed in 5a6dbfad. Added an explicit release method to the lazy tensor reference and made the server release either a LazyTensorDict or the decoded dict of lazy references. Consumed and discarded replies now unlink their shared temporary directory immediately; a focused unit test covers the shared-reference cleanup.

research/fedrevive/server.py
ZiyueXu77: Fixed narrowly in 5a6dbfad. Any failed or missing assignment now aborts FedAvg, FedBuff, and FedRevive with the assignment ID and error. This prevents a degraded run from silently changing the benchmark arrival stream without adding retries, thresholds, or recovery policy to the research example.

research/fedrevive/server.py
ZiyueXu77: Fixed in 5a6dbfad. The upload branch now applies the same max-time guard as download and train before returning an update for aggregation. The outcome is retained for normal shutdown cleanup, and a focused scheduler smoke test covers the t=205 versus budget=200 case.

research/fedrevive/job.py
ZiyueXu77: Fixed in 5a6dbfad. Removed --max-parallel and its validation/documentation, and now pass parallel=0. The physical Collab site count is documented as the actual bound on concurrent RPCs.

research/fedrevive/server.py
ZiyueXu77: Fixed in 5a6dbfad. TensorBoard now uses model_version as the integer global step and records simulated_time as walltime, avoiding collapsed points while preserving both coordinates.

research/fedrevive/server.py
ZiyueXu77: Fixed in 5a6dbfad. setup_tensor_disk_offload now accepts an optional root_dir and passes it directly to mkdtemp. The FedRevive server supplies its run directory and no longer mutates process-global tempfile state. A focused unit test verifies the requested parent directory.

research/fedrevive/requirements.txt
ZiyueXu77: Fixed in 5a6dbfad. Added the NVFlare PT requirement, removed the editable-install command, documented dependency/data license terms, and added FedRevive to the research index. The pin is intentionally nvflare[PT]~=2.10.0rc rather than 2.9 because this contribution targets main and now uses the new root_dir argument added to the disk-offload helper in this PR.

ZiyueXu77: Fixed in 5a6dbfad. Added the NVFlare PT requirement, removed the editable-install command, documented dependency and data license terms, and added FedRevive to the research index. The pin is intentionally nvflare[PT]~=2.10.0rc rather than 2.9 because this contribution targets main and now uses the new root_dir argument added to the disk-offload helper in this PR.

research/fedrevive/server.py
ZiyueXu77: Fixed in 5a6dbfad. Removed the ineffective ContextKey.RESULT lookup and initialize the global model directly from the seeded initial model.

research/fedrevive/README.md
```diff
@@ -0,0 +1,466 @@
...
+with the NVIDIA FLARE Collab API. It targets the CIFAR-10 experiment in the
+paper and provides FedAvg, FedBuff, and FedRevive through one workflow.
+
+FedRevive retains the ordinary parameter-space update from an asynchronous
+client and supplements stale updates with server-side data-free knowledge
+distillation (DFKD). Recent returned client models act as teachers, a
+persistent meta-learned generator produces synthetic inputs, and a
+staleness-dependent coefficient blends the client and distilled updates.
+
+## Scope
```
holgerroth, line 13: **Suggestion: state explicitly that the simulated-time scheduler is a paper-reproduction harness, not the recommended way to run FedRevive in FLARE.**

`server.py` carries a discrete-event simulator (simulated clock, event heap, logical-to-physical worker mapping, reference-counted disk snapshots, a watcher pool for Collab result queues). That is what makes the seeded delay schedules from the paper replayable, and it is valuable for that purpose. But `research/` examples tend to get copied as templates, and a user who wants FedRevive on their own clients does not need any of it: in a real deployment the physical clients are the clients, staleness comes from actual arrival order, and the whole algorithm reduces to `process_update` plus the reviver.

Could you add a short paragraph here in **Scope** (and a sentence at the top of **Unified scheduling and aggregation**) along the lines of:

> The scheduler in `server.py` replays the paper's seeded delay schedules on a small pool of physical sites so that results are reproducible regardless of host speed. It is a reproduction harness for this study, not the recommended pattern for deploying FedRevive. For real asynchronous training, start from the `examples/advanced/collab/pt_async_cifar10` example and reuse `fedrevive.py` (`process_update`, `TeacherBuffer`, `blend_updates`) and `dfkd.py` directly.

That keeps the algorithm files as the reusable part and makes the boundary explicit for readers. We are considering a product-side FedRevive component on top of the existing FedBuff pieces as a follow-up, so having the README already draw this line will make that migration cleaner.

holgerroth (commented):
One more README suggestion, inline below.

holgerroth (commented):
I did a focused pass mapping the DFKD implementation against the paper (arXiv 2511.00655, Section 2.2, Appendix B, Appendix I). Most of it lines up: Eq. 3 blending and staleness definition, the c=8 teacher buffer and per-class teacher weights, the two-upload Gaussian-probe proxy with T_probe=0.8, the target and adversarial synthesis terms and their weights, the Reptile-style meta-update, best-of-K_synth pool insertion, the class-weighted multi-teacher KD loss with T=1 and K_KD=10, and the published learning rates and batch sizes.

The four inline comments below are the items I consider material to whether the README figures can be called a reproduction: target sampling, the unpublished warmup gates, defaults versus the paper's main setting, and the missing normalization layers. Smaller conventions (unsquared feature-loss norm, kornia augmentation of synthetic images, the loss-clipping heuristic, unpublished constants like the beta horizon of 75 and the 16k pool) follow common DFKD practice and I have not flagged them individually.

One principle for resolving these: the default code path should keep producing the figures shown in the README. Where the paper is followed differently, document the deviation rather than silently changing defaults, or regenerate the figures alongside the change.

research/fedrevive/dfkd.py
```diff
@@ -0,0 +1,468 @@
...
+            models.append(model)
+            distributions.append(teacher.class_proxy.to(self.device))
+        return models, distributions
+
+    def _synthesize(self, teachers, distributions, student, current_version: int):
+        started = time.time()
+        self.synthesis_count += 1
+        batch_size = self.config.synth_batch_size
+        combined_distribution = torch.stack(distributions).mean(dim=0)
+        targets = torch.multinomial(combined_distribution, batch_size, replacement=True)
```
holgerroth, line 327: **Synthesis targets are sampled from the mean teacher proxy; the paper says uniform.** Section 2.2 step (2): target labels are drawn "uniformly at random over classes". Here `targets = torch.multinomial(combined_distribution, ...)` draws them from the average of the buffered teachers' class proportions, which skews the synthetic pool toward whatever classes the current eight teachers hold. Appendix B does mention "client-conditioned synthesis targets", so the paper is not fully consistent with itself, but the main-text definition is the explicit one. Could you confirm with the authors which variant produced the paper's curves, and either switch to uniform or state the deviation in the README?

research/fedrevive/dfkd.py
```diff
@@ -0,0 +1,468 @@
...
+    max_images: int = 16000
+    synth_batch_size: int = 64
+    kd_batch_size: int = 32
+    kd_lr: float = 1e-4
+    temperature: float = 1.0
+    generator_lr: float = 3e-3
+    latent_lr: float = 1e-3
+    generator_steps: int = 2
+    generation_interval: int = 1
+    warmup_versions: int = 100
```
holgerroth, line 42: **`warmup_versions=100` and `freeze_versions=50` introduce a two-stage warmup that the paper does not describe.** With these gates DFKD does nothing before version 50, synthesizes but returns no distilled update between 50 and 100 (so FedRevive is plain asynchronous FedAvg with B=1 there), and only starts blending the KD delta after version 100. Nothing in Sections 2 to 3 or Appendix I mentions a warmup. If this was needed to match the reference curves, please say so in the README next to the version-50 and version-100 statements and give the source; otherwise readers will attribute the early part of the curve to FedRevive when it is not running.

research/fedrevive/job.py
```diff
@@ -0,0 +1,267 @@
...
+        "--min-open-slots",
+        type=int,
+        default=None,
+        help="O; defaults to the method preset",
+    )
+    parser.add_argument("--server-lr", type=float, default=None)
+    parser.add_argument(
+        "--class-proportion-source",
+        choices=[source.value for source in ClassProportionSource],
+        default=ClassProportionSource.TRUE_HISTOGRAM.value,
```
holgerroth, line 76: **Defaults differ from the paper's main setting; keep them tied to the shown figures, but say so.** The defaults (true label histograms, `--generation-interval 1`) match the headline results in Figure 1/2, which is the right consistency to preserve. But the paper's main results use the server-side proxy and T_gen = 10 ("2 generator steps every 10 server updates", Appendix I; K_KD = T_gen = 10, Section 3.1), and ground-truth proportions appear only in the Appendix B ablation. Right now the paper-faithful configuration is the secondary "estimated, interval 10" run in Figure 3 (0.7462 vs 0.7612). Please state in the README, next to the Results table, that the default configuration deviates from the paper in these two respects and that the estimated/interval-10 run is the one comparable to the paper. If the defaults are changed instead, the figures need regenerating so code and README stay in sync.

research/fedrevive/model.py
```diff
@@ -0,0 +1,200 @@
...
+from collections import OrderedDict
+
+import torch
+import torch.nn as nn
+import torch.nn.functional as F
+
+ModelState = dict[str, torch.Tensor]
+
+
+class StatTracker(nn.Module):
```
holgerroth, line 26: **This network has no normalization layers, while the paper describes a ResNet-18 whose feature loss uses "BatchNorm statistics" (Appendix I).** `StatTracker` records running mean and variance but returns `x` unchanged, and `reset_tracker_stats_in_state` zeros those buffers on every global version, so the model is a ResNet-18 with the BN layers removed. That may well be what the authors' reference implementation does (BN in FL is a known problem) and the README does say "with feature-statistic trackers", but a reader who sees "ResNet-18" will assume standard BN. Please state explicitly that no normalization is applied and, if possible, confirm with the authors that this matches their setup, since it changes both the optimization behaviour and what the feature-matching loss is matching against.

research/fedrevive/dfkd.py
ZiyueXu77: Confirmed that the curve-producing implementation samples synthesis targets from the mean buffered-teacher proxy. I retained that behavior and documented explicitly that it differs from the uniform-target statement in Section 2.2, while teacher-specific proportions remain separate inputs to multi-teacher KD sampling.

research/fedrevive/job.py
ZiyueXu77: Changed the FedRevive defaults to the paper-main estimated class-proportion proxy and generation interval 10, while FedAvg and FedBuff retain their neutral true-histogram/interval-1 values. Figures 1 and 2 and their result tables were regenerated from this default; Figure 3 now presents true histogram with every-version synthesis as the ablation.

research/fedrevive/README.md
ZiyueXu77: Added the requested boundary in Scope and at the start of Unified scheduling and aggregation. The README now identifies the seeded discrete-event scheduler as a controlled reproducibility harness, points real deployments to pt_async_cifar10, and names fedrevive.py and dfkd.py as the reusable algorithm pieces.

research/fedrevive/model.py
ZiyueXu77: Confirmed that the curve-producing network uses non-normalizing StatTracker modules. The configuration table and model description now say explicitly that activations are returned unchanged and that the recorded channel statistics are used only by the DFKD feature loss.

research/fedrevive/dfkd.py
ZiyueXu77: Confirmed that versions 50 and 100 are the two gates used by the curve-producing configuration. I retained them and now state next to the DFKD procedure that these gates are not specified in the paper, including what happens in each interval.

research/fedrevive/job.py
```diff
@@ -0,0 +1,277 @@
...
+    print(f"FEDREVIVE COLLAB FIGURE 2: {args.method}")
+    print(f"  Physical clients: {args.num_clients}")
+    print(f"  Logical clients: {args.num_logical_clients}")
+    print(f"  K/B/O: {active_jobs}/{buffer_size}/{open_slots}")
+    print(f"  In-time accumulation: {args.in_time}")
+    print(f"  Simulated-time budget: {args.max_time}")
+    print(f"  Delay schedule: {args.delay_schedule}")
+    print(f"  Class-proportion source: {args.class_proportion_source}")
+    print(f"  Class-proportion probe count: {args.class_proportion_probe_count}")
+    print(f"  Generation interval: {args.generation_interval}")
```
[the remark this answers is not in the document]
ZiyueXu77: Fixed in a6f2d7e0. The console summary now reads the resolved values from recipe.server, so a default FedRevive run reports estimated and 10 instead of the unresolved parser values None.

holgerroth (approved):
Thanks for the thorough follow-through. All items from my three reviews are addressed:

- The scheduler bugs, the offload cleanup path, `--max-parallel`, TensorBoard steps, the `tempfile` swap, and the packaging/license items were fixed in 5a6dbfa, with the two small core changes (`root_dir` on the offload helper, `_LazyRef.release()`) covered by unit tests.
- The README now labels the simulated-time scheduler as a reproduction harness and points real deployments to the async Collab example.
- FedRevive defaults now match the paper's main setting (estimated proxies, generation interval 10) and Figures 1 and 2 were regenerated, so code, defaults, and shown results are consistent. Mean-proxy target sampling, the non-normalizing trackers, and the version-50/100 gates are confirmed with the paper authors and documented as deviations.

One optional follow-up: the shifted-schedule FedRevive run stopped at the former 50,000-version cap (t=97.43). Rerunning it with the new 100,000 cap would let both rows in that table reach the 100-unit budget, but the README discloses this clearly, so it does not need to block the merge.

Approving.

PR #5280, opened 2026-09-10
Fix premature POC readiness failures and unbounded waits

chesterxgchen:
The reported Recipe regression had two symptoms: healthy POC services could fail the one-second admin connection probe, and subsequent admin-session cleanup could hang beyond the configured readiness timeout. Both `nvflare poc start` and Recipe `PocEnv.deploy()` use this readiness probe.

Restoring the default ten-second connection allowance made the reported reproducer submit successfully. That addresses startup latency, but authentication, login, streaming, and cleanup can still outlast their nominal timeouts. This PR also bounds how long the readiness caller waits.

Run the complete readiness probe in one daemon worker. The caller waits on a Future for the remaining startup duration. The worker owns its session, closes it when the pending operation returns, and checks cancellation and remaining time between operations and attempts. Readiness confirmed before the timeout is returned without waiting for cleanup.

- Restore the default ten-second transport connection allowance in the POC CLI.
- Validate readiness timing arguments before starting work; report invalid timing settings as `INVALID_ARGS`.
- Preserve the base branch's session/provider retry behavior. Session construction, connection, and status errors retry within the readiness window. The last observation is retained in the timeout message.
- Keep the general Session/admin APIs and certificate-provider implementations unchanged from the base branch.

Provider error classification is outside this PR. Generic provider exceptions and step CLI exit codes do not reliably distinguish configuration errors from transient request failures. The attempted classification rules and new provider exception types have been removed. Consequently a permanent provider/configuration error can still consume the readiness window, as on the base branch; this PR bounds that wait and does not promise immediate `INVALID_CONFIG` reporting. A provider error contract would need separate design and compatibility review.

A daemon thread is intentional: `ThreadPoolExecutor` joins workers at interpreter shutdown even after `shutdown(wait=False)`, so a blocked pool worker can keep the CLI alive. The timeout bounds the caller's wait and does not forcibly terminate an in-flight operation. A timed-out call can leave one worker until that operation returns; the worker then closes its session and starts no further attempt. The optional `second_to_wait` delay precedes the readiness timeout; POC callers use zero.

Diff at `c08c022c4` against upstream `main` at `83ec31a8e`:

| Category | Files | Added | Removed | Net |
|---|---:|---:|---:|---:|
| Production Python | 2 | 121 | 57 | +64 |
| Test Python | 2 | 515 | 15 | +500 |
| Documentation (`.rst`, `.md`) | 0 | 0 | 0 | 0 |
| **Total** | **4** | **636** | **72** | **+564** |

Tests account for 81% of added lines. Counts include comments and blank lines; this GitHub PR description is excluded.

Validation:

- 389 readiness, POC CLI, Flare API, HCI, and certificate-provider unit tests passed.
- Real Recipe POC deployment, job submission, and completion passed in 39.07 seconds.
- Tests retain real authentication, login, command-list retrieval, and Cell streaming wait loops, with delay/progress simulated at the network boundary. Worker phase entry is synchronized before measuring blocked-operation timeouts.
- Regressions cover successful readiness with blocked cleanup, late results, stalled session operations, slow successful transport, and session/provider failures that recover or exhaust the bounded readiness wait.
- Full repository style checks and `git diff --check` passed.

### DIFF
--- a/nvflare/fuel/flare_api/flare_api.py
+++ b/nvflare/fuel/flare_api/flare_api.py
@@ -205,7 +205,7 @@ class Session(SessionSpec):
         if isinstance(session_expired_reason, str) and session_expired_reason:
             raise SessionExpired(session_expired_reason)
 
-    def try_connect(self, timeout: Optional[float]) -> None:
+    def try_connect(self, timeout: Optional[float], *, connect_timeout: Optional[float] = None) -> None:
         """Establish the admin transport connection and log in.
 
         Args:
@@ -215,9 +215,14 @@ class Session(SessionSpec):
                 Login requests, command-list retrieval, and retry sleeps use
                 the budget remaining after transport authentication. Later
                 session operations have their own timeouts.
+            connect_timeout: Optional transport authentication limit in seconds,
+                capped by the total timeout. For example,
+                ``session.try_connect(30, connect_timeout=10)`` allows transport
+                authentication up to ten seconds and login the remainder of the
+                thirty-second total. None lets transport use the total budget.
 
         Raises:
-            ValueError: The timeout is not a finite positive number.
+            ValueError: A supplied timeout is not a finite positive number.
             NoConnection: Transport authentication fails, consumes the budget,
                 or login reports a connection failure.
             SessionClosed: The session is closed.
@@ -234,10 +239,14 @@ class Session(SessionSpec):
             timeout = self.api.default_login_timeout
         if not isinstance(timeout, (int, float)) or not math.isfinite(timeout) or timeout <= 0:
             raise ValueError("timeout must be a finite positive number of seconds")
+        if connect_timeout is not None and (
+            not isinstance(connect_timeout, (int, float)) or not math.isfinite(connect_timeout) or connect_timeout <= 0
+        ):
+            raise ValueError("connect_timeout must be a finite positive number of seconds or None")
         deadline = time.monotonic() + timeout
 
         try:
-            self.api.connect(timeout)
+            self.api.connect(min(timeout, connect_timeout) if connect_timeout is not None else timeout)
         except FLCommunicationError as e:
             message = str(e)
             if "cannot connect to server" in message or "cannot authenticate to server" in message:
--- a/nvflare/fuel/flare_api/flare_api.py
+++ b/nvflare/fuel/flare_api/flare_api.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import json
+import math
 import os
 import time
 from typing import List, Optional
@@ -204,44 +205,36 @@ class Session(SessionSpec):
         if isinstance(session_expired_reason, str) and session_expired_reason:
             raise SessionExpired(session_expired_reason)
 
-    def try_connect(self, timeout: Optional[float], *, deadline: Optional[float] = None) -> None:
+    def try_connect(self, timeout: Optional[float]) -> None:
         """Establish the admin transport connection and log in.
 
         Args:
-            timeout: Transport authentication timeout in seconds; must be positive,
-                or None to use the admin configuration's login timeout. This
-                argument alone does not limit the subsequent admin login retries.
-            deadline: Optional absolute time in seconds from ``time.monotonic()``.
-                Pass it by keyword, for example ``deadline=time.monotonic() + 30.0``.
-                After transport authentication, login requests (including command-list
-                retrieval) and retry sleeps share the remaining budget. None preserves
-                the configured login retry count and per-command timeouts. This is
-                neither a duration nor a wall-clock timestamp from ``time.time()``.
-                It does not interrupt transport authentication or bound later session
-                operations such as ``get_system_info()`` or ``close()``.
+            timeout: Total connection and login budget in seconds, for example
+                ``session.try_connect(30.0)``. Must be a finite positive number,
+                or None to use the admin configuration's login timeout.
+                Login requests, command-list retrieval, and retry sleeps use
+                the budget remaining after transport authentication. Later
+                session operations have their own timeouts.
 
         Raises:
-            NoConnection: Transport authentication fails, the deadline has already
-                expired after authentication, or login reports a connection failure.
+            ValueError: The timeout is not a finite positive number.
+            NoConnection: Transport authentication fails, consumes the budget,
+                or login reports a connection failure.
             SessionClosed: The session is closed.
             SessionExpired: The session has expired.
             AuthenticationError: Admin login rejects the credentials.
             AuthorizationError: Admin login denies access.
-            InternalError: Login otherwise fails, including deadline exhaustion
-                during login requests or retries.
-
-        Example:
-            Allow up to ten seconds for transport authentication, then use the
-            remaining time from a thirty-second budget for admin login::
-
-                import time
-
-                deadline = time.monotonic() + 30.0
-                session.try_connect(timeout=10.0, deadline=deadline)
+            InternalError: Login otherwise fails, including timeout during login.
         """
         self._raise_if_session_expired()
         if self.api.closed:
             raise SessionClosed("session closed")
+
+        if timeout is None:
+            timeout = self.api.default_login_timeout
+        if not isinstance(timeout, (int, float)) or not math.isfinite(timeout) or timeout <= 0:
+            raise ValueError("timeout must be a finite positive number of seconds")
+        deadline = time.monotonic() + timeout
 
         try:
             self.api.connect(timeout)
@@ -250,12 +243,10 @@ class Session(SessionSpec):
             if "cannot connect to server" in message or "cannot authenticate to server" in message:
                 raise NoConnection(message) from e
             raise
-        if deadline is not None:
-            if time.monotonic() >= deadline:
-                raise NoConnection("admin connection exhausted the readiness deadline")
-            result = self.api.login(deadline=deadline)
-        else:
-            result = self.api.login()
+        remaining = deadline - time.monotonic()
+        if remaining <= 0:
+            raise NoConnection("admin connection exhausted the timeout")
+        result = self.api.login(timeout=remaining)
         status = result.get(ResultKey.STATUS) if isinstance(result, dict) else None
         details = result.get(ResultKey.DETAILS, "") if isinstance(result, dict) else ""
         if status == APIStatus.SUCCESS:
--- a/nvflare/fuel/flare_api/flare_api.py
+++ b/nvflare/fuel/flare_api/flare_api.py
@@ -254,7 +254,7 @@ class Session(SessionSpec):
             raise
         remaining = deadline - time.monotonic()
         if remaining <= 0:
-            raise NoConnection("admin connection exhausted the timeout")
+            raise NoConnection("Timed out connecting to the admin server.")
         result = self.api.login(timeout=remaining)
         status = result.get(ResultKey.STATUS) if isinstance(result, dict) else None
         details = result.get(ResultKey.DETAILS, "") if isinstance(result, dict) else ""
--- a/nvflare/fuel/flare_api/flare_api.py
+++ b/nvflare/fuel/flare_api/flare_api.py
@@ -13,7 +13,6 @@
 # limitations under the License.
 
 import json
-import math
 import os
 import time
 from typing import List, Optional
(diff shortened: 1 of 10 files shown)


nvflare/tool/api_utils.py
```diff
@@ -127,17 +149,26 @@ def wait_for_system_start(
...
         sess = None
+        ready_sys_info = None
+        cleanup_timed_out = False
         try:
             sess = Session(username=username, startup_path=admin_user_dir, secure_mode=secure_mode)
-            remaining = max(deadline - time.time(), 0.1)
+            remaining = deadline - time.monotonic()
+            if remaining <= 0:
+                raise TimeoutError("readiness deadline reached before connecting to the admin server")
             sess.try_connect(min(conn_timeout, remaining))
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in 30c75e200: readiness passes the full remaining duration to Session.try_connect, with connect_timeout=10 separately capping transport authentication. The behavioral regression retains real login handling and succeeds after nine seconds of transport authentication plus a two-second login. Follow-up 8f870e1de improves timeout/progress messages, separates cleanup failure from readiness observations, and rejects invalid timing settings immediately. Validation: 337 POC/admin unit tests passed, the real Recipe deployment/submission/completion test passed in 36.18 seconds, and repository style checks passed.

pcnudde:
Would this problem not have been fixes by changing the timeout from 1 sec to 10 sec? Do we need 700 lines for this?

chesterxgchen:
@pcnudde, fair question. Changing the connection timeout from one second to ten did make the reported reproducer succeed, and this PR includes that change.

The original report also described startup hanging beyond its configured timeout. Increasing the connection timeout alone leaves two failure paths:

- Admin login can keep retrying after transport authentication, exceeding the overall readiness timeout.
- Closing the admin session after a failed attempt can block, preventing the readiness check from returning.

The additional implementation limits login requests/retry sleeps to the remaining startup time and bounds how long readiness waits for session cleanup. It also preserves a successful readiness result if cleanup subsequently times out. The background cleanup thread is not forcibly cancelled.

At the current head, the diff is 604 added lines: 406 in tests and 198 in implementation files, including API documentation. The tests exercise slow transport followed by login, stalled login responses, blocked cleanup, and failure reporting. That breakdown explains much of the size, but the broader change to `Session.try_connect()` semantics is still a maintenance tradeoff: its timeout now includes login.

The ten-second change is sufficient as a mitigation for the reproduced startup latency; the login and cleanup changes address the reported hang as well. I should have made those two parts of the scope clearer in the initial description.

nvflare/fuel/flare_api/flare_api.py
```diff
@@ -204,19 +205,57 @@ def _raise_if_session_expired(self):
...
+            raise ValueError("timeout must be a finite positive number of seconds")
+        if connect_timeout is not None and (
+            not isinstance(connect_timeout, (int, float)) or not math.isfinite(connect_timeout) or connect_timeout <= 0
+        ):
+            raise ValueError("connect_timeout must be a finite positive number of seconds or None")
+        deadline = time.monotonic() + timeout
+
         try:
-            self.api.connect(timeout)
+            self.api.connect(min(timeout, connect_timeout) if connect_timeout is not None else timeout)
```
IsaacYangSLA: [P2] Bound transport authentication by the remaining deadline

Passing the remaining duration to `AdminAPI.connect()` does not enforce the advertised total timeout. `Authenticator` still sends challenge/registration requests with its separate `msg_timeout` (5 seconds by default), sleeps for the full retry interval, and starts a fresh timeout for registration. The deadline check here only runs after that synchronous work returns. With the real `Session.try_connect()` / `AdminAPI.connect()` / authentication loop and a simulated challenge reply that stalls for the requested network timeout, `wait_for_system_start(..., second_to_wait=0, timeout_in_sec=0.05)` took 5.009 seconds and sent the challenge with `timeout=5.0`. Thus startup can still exceed its readiness deadline. Please propagate one deadline through authentication and cap each request and retry sleep by the time remaining; retain the real authentication loop in a regression test.

nvflare/fuel/hci/client/api.py
```diff
@@ -750,6 +792,12 @@ def _send_to_cell(self, ctx: CommandContext):
         timeout = self.cmd_timeout
         if not timeout:
             timeout = 5.0
+        if self._login_deadline is not None:
+            self._check_login_deadline()
+            timeout = min(timeout, self._login_deadline - time.monotonic())
```
IsaacYangSLA: [P2] Enforce the login deadline inside streamed request waits

The remaining duration passed here to `Cell.send_request()` is an idle timeout, not a total request timeout. HCI uses the streaming path: `Cell._send_one_request()` reuses it for sending, waiting for the first reply byte, and receiving, while `_future_wait()` renews the wait whenever transfer progress occurs. Consequently this synchronous call can outlive `_login_deadline`; checking the deadline after `_user_login()` returns cannot bound it. With the real login and Cell wait logic, simulated send progress every 0.05 seconds for 0.3 seconds followed by no reply made `api.login(timeout=0.1)` take 0.404 seconds. Continued progress can extend it further, also delaying readiness. Please enforce a shared deadline or deadline-triggered cancellation across these waits and cover the real streaming path in a regression test.

nvflare/fuel/flare_api/flare_api.py
```diff
@@ -204,19 +205,57 @@ def _raise_if_session_expired(self):
...
         except FLCommunicationError as e:
             message = str(e)
             if "cannot connect to server" in message or "cannot authenticate to server" in message:
                 raise NoConnection(message) from e
             raise
-        result = self.api.login()
+        remaining = deadline - time.monotonic()
+        if remaining <= 0:
+            raise NoConnection("Timed out connecting to the admin server.")
```
pcnudde: [P2] Do not interpret an exhausted connection budget as confirmed shutdown

This raises `NoConnection` after `self.api.connect()` has returned successfully. Existing callers interpret that exception as evidence the server is down: `_wait_for_server_down()` returns successfully, and `shutdown_system()` reports `already_stopped=True` without sending the shutdown command. A controlled virtual-clock reproduction with successful transport taking 1.1 seconds against the one-second polling budget made `_wait_for_server_down(30)` return successfully without calling login or requesting system status. Authentication can legitimately finish beyond the supplied duration because the underlying request waits are not bounded by one overall deadline. Please use a distinct timeout exception for this case, or update callers so an exhausted budget cannot be treated as proof of shutdown. Add regression coverage for a slow but successful connection.

nvflare/fuel/flare_api/flare_api.py
pcnudde: Confirmed independently at `bff9f725d2fc97b733af5cf098a6e569d134df5f`. A controlled virtual-clock reproduction retaining the real `Session.try_connect()`, `AdminAPI.connect()`, and authentication loop, with delay simulated at the network boundary, gave `wait_for_system_start(..., second_to_wait=0, timeout_in_sec=0.1)` a 0.1-second budget but advanced five seconds before returning. The challenge request still received `timeout=5.0`. This supports the finding that the advertised overall readiness deadline is not enforced inside transport authentication. This was a simulated-delay reproduction, not a live deployment test.

chesterxgchen:
@pcnudde, I simplified the implementation in a976e1e61. The ten-second connection allowance remains, but all changes to Session.try_connect() and AdminAPI.login() are removed.

Instead of propagating timeout bookkeeping through transport, login, retries, and streaming, the readiness helper runs its complete session lifecycle in one daemon worker. The caller waits on a Future with the startup timeout. This also handles the transport and streaming overruns identified by you and Isaac without changing general admin API behavior.

The tradeoff is explicit: this bounds the caller's wait, not the lifetime of an in-flight operation. A timed-out call can leave one worker until that operation returns; the worker then closes its session and starts no new attempt. Confirmed readiness returns without waiting for cleanup.

The current diff changes two implementation files instead of four. Implementation additions are down from 198 to 140 lines; net implementation growth is down from 167 to 85. The total is 558 added lines, of which 418 are tests. I consolidated redundant tests while retaining the actual authentication/streaming regressions.

Changing only one second to ten seconds remains a useful mitigation for the reported slow-host reproduction, but would leave the reported unbounded wait. This version addresses both with one timeout boundary confined to readiness handling.

Validation: 326 unit tests passed, the real Recipe POC deployment/submission/completion test passed in 39.91 seconds, and repository style checks passed.

nvflare/fuel/flare_api/flare_api.py
chesterxgchen: Addressed in a976e1e61 by moving the timeout boundary to the readiness caller and reverting the Session.try_connect() API changes. You are right that Authenticator's existing per-request timeout can exceed the duration passed to connect(). The readiness caller now waits on one Future for its remaining startup time, independently of that synchronous authentication operation.

The regression retains real Session.try_connect(), AdminAPI.connect(), Authenticator.authenticate(), and the challenge loop. A blocked network challenge still receives the existing five-second request timeout, but a 0.1-second readiness call returns in under 0.5 seconds. Releasing the simulated network block then verifies session cleanup.

This does not change Authenticator's general timeout semantics or forcibly cancel it. One worker owns the session and closes it when the pending operation returns. This narrower approach avoids adding deadline plumbing throughout the transport stack. All 326 unit tests and the real Recipe POC workflow passed.

nvflare/fuel/hci/client/api.py
chesterxgchen: Addressed in a976e1e61 with a readiness-level elapsed timeout; the new AdminAPI.login(timeout=...) API and its deadline bookkeeping have been removed. Stream idle timeouts may still renew with progress, but they no longer control how long the readiness caller waits.

The regression retains real login/command-list handling plus Cell._send_one_request(), _future_wait(), and conditional_wait(). Simulated send progress repeatedly renews the idle wait; readiness nevertheless returns within its configured wait, and the worker closes its session after the simulated stream is released. Both login and command-list phases are covered, along with the real login retry sleep.

This deliberately bounds readiness without changing general streaming semantics. It does not forcibly cancel the pending stream. All 326 unit tests and the real Recipe POC workflow passed.

nvflare/fuel/flare_api/flare_api.py
chesterxgchen: Fixed in a976e1e61 by restoring Session.try_connect() unchanged from the base branch. The newly introduced NoConnection after successful transport authentication is gone; the readiness timeout is enforced separately and cannot be interpreted by shutdown callers as evidence the server is down.

New virtual-clock regressions cover both callers: transport succeeds after its polling allowance, _wait_for_server_down() still logs in and requests status before reporting that the server did not stop, and shutdown_system() sends shutdown rather than returning already_stopped=True. All 326 unit tests and the real Recipe POC workflow passed.

nvflare/tool/api_utils.py
```diff
@@ -119,68 +124,134 @@ def wait_for_system_start(
...
+                    return
                 try:
-                    sess.close()
-                except Exception as e:
-                    last_error = str(e)
+                    sess.try_connect(min(conn_timeout, remaining))
+                except ConfigError as e:
+                    outcome.set_exception(e)
+                    return
+                remaining = remaining_time()
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in 1b5f8d06b. Certificate acquisition now raises a ConfigError subclass that readiness can distinguish from local configuration failures. This covers both renewal during connect() and initial issuance during Session construction. Both retry within the readiness timeout; existing ConfigError handlers remain compatible, and session method signatures are unchanged.

The regression uses real Session/AdminAPI construction, actual fed_admin.json loading, and real acquisition/renewal handling. Only the external provider and subsequent network operations are simulated. It covers first-failure recovery and persistent failure for both initial issuance and renewal, verifying bounded timeout and cleanup.

The same commit fixes missing admin files and malformed fields being retried as connection failures. Actual missing fed_admin.json and null download_dir fixtures now fail promptly as configuration errors, without mocking secure_load_admin_config.

Validation: 358 unit tests, the real Recipe POC deployment/submission/completion test (39.06 seconds), and repository style checks passed.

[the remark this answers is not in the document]
chesterxgchen: Follow-up in 17421052d corrects two classification gaps identified in review. Only marked provider request failures and connection/timeout exceptions are retryable acquisition errors; local provider validation, import, executable, and CA-file failures remain ConfigError. The connect handler now retries AdminCertAcquisitionError explicitly and fails promptly for other ConfigError instances, including malformed connector settings.

Behavioral regressions exercise actual malformed provider configurations, missing/non-executable step binaries, an unreadable CA path, and comm_config.json loading during real cell construction. They do not mock obtain_admin_cert_files or configuration loading. Existing subprocess tests verify the retryable request subtype, and readiness tests retain initial issuance/renewal recovery, bounded persistent failure, and blocked cleanup coverage.

Validation: 405 unit tests passed; the real Recipe POC deployment/submission/completion test passed in 37.69 seconds; full repository style checks passed.

nvflare/fuel/hci/client/api.py
```diff
@@ -408,6 +413,10 @@ def ensure_client_cert_valid(self):
                 config=self.admin_cert_provider_config,
                 root_ca_file=self.ca_cert,
             )
+        except (AdminCertProviderRequestError, ConnectionError, TimeoutError) as ex:
+            raise AdminCertAcquisitionError(
+                f"failed to obtain admin certificate: {secure_format_exception(ex)}"
+            ) from ex
         except Exception as ex:
             raise ConfigError(f"failed to obtain admin certificate: {secure_format_exception(ex)}") from ex
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in 63a800b55. The compatibility catch now sits directly around the dynamically loaded custom-provider call, after provider configuration validation, CA/cache setup, and provider loading. Legacy AdminCertProviderError and OSError acquisition failures are converted to the retryable subtype there. Custom providers do not need to adopt a new exception immediately. Built-in providers (including their module:function aliases) retain their explicit classification.

The same commit fixes the step CLI exit classification raised in the separate review: nonzero exits now stop readiness immediately. step's runner returns exit code 1 for ordinary command errors, so the code cannot reliably infer whether a rejected command is transient. Explicit subprocess timeouts remain retryable; existing CLI output remains visible. This deliberately treats ambiguous step exits as terminal, without parsing human-readable errors.

Behavioral tests retain real provider loading/acquisition and readiness handling. Controlled child processes reject an unknown provisioner or zero TTL; each readiness call fails promptly after one invocation. Dynamically loaded custom-provider tests verify compatibility for both legacy exception types. 409 unit tests and repository style checks passed.

Upstream exit handling: https://github.com/smallstep/cli/blob/master/internal/cmd/root.go

nvflare/fuel/sec/admin_cert_provider.py
```diff
@@ -91,7 +95,14 @@ def obtain_admin_cert_files(config: Mapping, root_ca_file: str) -> AdminCertFile
...
         provider_func = _load_provider(provider)
-        files = provider_func(config=provider_config, root_ca_file=root_ca_file)
+        try:
+            files = provider_func(config=provider_config, root_ca_file=root_ca_file)
+        except (AdminCertProviderError, OSError) as ex:
+            if provider in BUILTIN_ADMIN_CERT_PROVIDERS or provider in BUILTIN_ADMIN_CERT_PROVIDERS.values():
+                raise
+            # Legacy custom providers use these types for acquisition failures.
+            # Keep retries at the provider boundary, after local setup succeeds.
+            raise AdminCertProviderRequestError(str(ex)) from ex
```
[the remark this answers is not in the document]
chesterxgchen: Addressed by removing the new provider-classification policy from this PR in c08c022c4. The custom-provider catch and both new exception types are gone; admin_cert_provider.py, step_ca_admin_cert.py, and the general admin API now match the base branch exactly.

The existing exception contract cannot reliably identify the cause of an arbitrary custom-provider error. Readiness now preserves the base branch's retries and bounds the caller's wait. Permanent configuration failures can still consume that window; immediate INVALID_CONFIG classification is explicitly outside this PR, not claimed as fixed. The PR description documents this scope and limitation.

Validation: 389 unit tests, real Recipe deployment/submission/completion (39.07 seconds), and repository style checks passed.

nvflare/fuel/sec/step_ca_admin_cert.py
```diff
@@ -115,9 +115,13 @@ def _run_step(
...
     except subprocess.TimeoutExpired as ex:
-        raise AdminCertProviderError(f"step ca certificate timed out after {timeout} seconds") from ex
+        raise AdminCertProviderRequestError(f"step ca certificate timed out after {timeout} seconds") from ex
     except subprocess.CalledProcessError as ex:
-        raise AdminCertProviderError(f"step ca certificate failed with exit code {ex.returncode}") from ex
+        # step uses the same exit code for configuration and request errors.
+        # Only explicit timeouts above are safe to retry automatically.
+        raise AdminCertProviderError(
+            f"step ca certificate failed with exit code {ex.returncode}; see the step CLI output for details"
+        ) from ex
```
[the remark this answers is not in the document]
chesterxgchen: Addressed in c08c022c4 by removing the new terminal step-exit policy and readiness ConfigError short-circuit. step_ca_admin_cert.py and the general admin API now match the base branch exactly. Failed acquisition follows the existing readiness retry path, with the caller's total wait bounded by this PR.

The scope is now the ten-second connection allowance plus bounded readiness waiting. It does not infer transient/permanent causes from CLI exit codes or introduce a provider error contract. Permanent provider errors may also consume the readiness window, as on the base branch; this limitation is explicit in the updated description.

Validation: 389 unit tests, real Recipe deployment/submission/completion (39.07 seconds), and repository style checks passed.

chesterxgchen:
@greptile-ai

chesterxgchen:
@greptileai review this PR

pcnudde (approved):

PR #5286, opened 2026-09-12
Add progress logging and summarize recipe runs

chesterxgchen:
## Problem

The existing Recipe logs are written primarily for framework diagnosis, rather than for a data scientist following a training run. The console and `log_fl.txt` mix training progress with communication, executor, and debugging messages. In the unmodified two-client, three-round Hello PyTorch example, the default console produces 203 lines, making it difficult to identify the current round, compare client metrics, or see what the server aggregated and evaluated.

The terminal outcome is also unclear. A successful run finishes with a result workspace path rather than a summary of the trained models and evaluation results. A failed run may be discovered only by finding an exception among unrelated log messages; users can have difficulty determining whether the job succeeded, which sites failed, and which application error caused the failure.

Data scientists need a focused view of the information they care about:

- training rounds and elapsed time;
- client, aggregated, and evaluation metrics;
- an explicit successful, failed, or aborted outcome;
- saved model and result locations on success; and
- the relevant application error, affected sites, and log locations on failure.

The existing `concise` mode cannot be changed because released users rely on its current output. The focused view must therefore be a separate opt-in mode, while the detailed diagnostic logs remain available for framework debugging.

## Result

This PR adds an opt-in `progress` mode for standard round-based deep-learning Recipes and a clear terminal report from `Run.get_result()`:

- During training, `progress` shows the current round, client metrics, aggregated metrics, evaluation progress, and text-labeled warnings and errors.
- On success, the terminal report shows the recorded rounds, cross-site evaluation, elapsed time, and artifact locations.
- On failure or abort, it shows the most relevant available client/server error, application source location, logs, and retained result workspace.
- Simulation reads bounded local artifacts and log tails. POC and production retrieve bounded remote log tails through the existing session path.
- The existing `concise` mode remains the default and preserves its NVFlare 2.9 formatter, filtering, and file-handler behavior. Existing `msg_only`, `full`, and `verbose` modes are also unchanged.

The result is a data-scientist view for normal operation and a direct failure explanation when a run does not complete. Detailed diagnostic records remain in the existing log files. The PR does **not** change model training, aggregation, evaluation, metric values, saved model contents, or FL control flow.

Concise remains the simulator default for backward compatibility:

```bash
python job.py
```

Select the new presentation explicitly:

```bash
python job.py --log_config progress
```

Use the existing full diagnostic view when needed:

```bash
python job.py --log_config full
```

## What users see

For a two-client, three-round Hello PyTorch run, `progress` mode reduces the console from **203 lines to 77 lines**, including the final result summary. A representative excerpt is:

```text
============================= ROUND 2 / 3 ==============================

  Training

  Client              accuracy  accuracy_after_local_training
  site-1                    30                             40
  site-2                    30                             70
  ------------------------------------------------------------------
  Aggregated                30                             55

  ✓ Aggregated 2 client updates                                  3.0s

============================= RUN SUMMARY ==============================

  NVIDIA FLARE · hello-pt
  Simulation · 2 clients

  ✓ Completed                                                   30.8s

  Training · aggregated client metrics

  Round               accuracy  accuracy_after_local_training
  1                          1                             20
  2                         30                             55
  3                         70                             65

  Model evaluation · accuracy

  Client        SRV_FL_global_model.pt  SRV_best_FL_global_model.pt
  site-1                            75                           70
  site-2                            77                           70

  Models     server/simulate_job/app_server/
  Metrics    server/simulate_job/metrics/
  Evaluation server/simulate_job/cross_site_val/cross_val_results.json
  Logs       server/log.txt · site-1/log.txt · site-2/log.txt
  Results    /workspace/hello-pt
```

Tables are intentionally bounded: at most ten rows and two metric/model columns are printed, with the artifact path shown when more data is available. Summary generation never loads model weights and cannot replace the run result if an artifact is missing or malformed.

## Failure summaries

Failed Recipe runs retain the framework status and show the most useful available exception, affected sites, application code location, log paths, and result workspace. Repeated client failures are grouped, while full tracebacks remain in the diagnostic logs.

For example, an intentional training error in a real two-client Hello NumPy simulation produces:

```text
============================= RUN SUMMARY ==============================

  NVIDIA FLARE · hello-numpy
  Simulation · 2 clients

  ✗ Failed                                                      11.9s

  Status    FINISHED:EXECUTION_EXCEPTION

  Failure details

  Error     NameError: name 'undefined_training_variable' is not defined
  Where     site-1, site-2 / TaskScriptRunner
  Code      client.py:31 (train)
  Logs      site-1/error_log.txt · site-2/error_log.txt

  Full tracebacks and additional messages are in the logs.
  Results   /workspace/hello-numpy
```

Simulation reads bounded local log tails. POC and production retrieve only bounded server-side tails that were already streamed or stored; they do not contact clients or enable log streaming. If client logs are unavailable, the summary says so. Failure-summary errors cannot replace the original job status or result.

## Why this is larger than a formatter change

If the change only reformatted records already reaching the console, a small formatter patch would be sufficient. The missing terminal result does not exist in one source: live metrics arrive as workflow events, saved models and evaluations live in separate artifacts, failures live in server and client logs, and POC/production evidence must be retrieved remotely. The added code reads those sources with explicit size bounds, normalizes terminal status, and presents the same outcome without allowing reporting failures to alter the job result.

The production diff adds 1,049 lines and removes 157 lines, for **892 net lines**. The largest part is the new success/failure result readers at the Recipe boundary; the next two parts are the opt-in live presentation and bounded POC/production retrieval. The detailed breakdown appears below.

| Change | Why it is needed |
|---|---|
| New `progress` logger/filter configuration | Uses explicitly marked presentation records for supported workflows, retains textual severity outside color terminals, and keeps high-volume client execution diagnostics in the full log files while leaving `concise` unchanged. |
| `MetricsArtifactWriter` presentation | Uses existing metrics events to display client and aggregation progress. Standard Recipe/BaseFedJob training already configures this writer; directly assembled FedJob and JSON jobs keep their existing components. |
| Run and failure summaries | Reads existing metrics, evaluation JSON, statuses, and log tails to show the outcome and artifact locations after `get_result()`. Missing or malformed artifacts remain non-fatal. |
| Bounded remote log retrieval | POC and production summaries must retrieve the relevant server/client tail without transferring an unbounded log. Progress replay uses structured server logs with a bounded text fallback and reports persistent unavailability. The `max_bytes` API and server handling exist only for this bounded presentation. |
| Shared `--log_config` Recipe argument | Lets any Recipe script select `progress`, `concise`, `full`, or another logging configuration without adding a custom argument to every example. The parser was moved from `spec.py` to private `_args.py`; most of this diff is relocated existing export-argument logic. |
| Status and lifecycle normalization | Gives simulator, POC, and production the same completed, aborted, failed, and not-scheduled labels and avoids reporting expected end-of-run shutdown as a warning. |
| Encoding-safe rendering | Keeps summaries readable on UTF-8, ASCII, and CP1252 consoles; a console encoding error cannot change the job result. |

## Behavior by execution environment

| Environment | Presentation behavior |
|---|---|
| Simulation | With `progress`, prints round/evaluation progress live, then summarizes local artifacts. A failed simulation reports the relevant local error and workspace. |
| POC | With `progress`, uses the existing session monitor to show bounded server progress, downloads the result, and prints the same summary. |
| Production | With `progress`, uses the same bounded session/status path as POC. It does not reconfigure already-running services or Docker containers. |

The new `progress` configuration uses ordinary NVFlare loggers, explicitly marked progress records, and the existing filter/formatter mechanism. The existing `concise` and `msg_only` dictionaries match their NVFlare 2.9 behavior. The implementation does not add a progress thread, transport, cursor API, child-logger protocol, or example-specific reporting helper.

## Scope boundaries

- This PR targets standard deep-learning, round-based workflows that already emit client, aggregation, and evaluation metrics.
- Federated statistics needs a different summary vocabulary and is planned as a separate change.
- Cyclic and swarm learning need algorithm-specific progress presentation and are planned as a separate change.
- Traditional ML workflows such as XGBoost, linear models, and K-means are outside this PR.
- Application-defined values are formatted for display only. Original metrics and result artifacts are unchanged.
- Existing `concise`, `msg_only`, `full`, and `verbose` logging behavior remains available. Full diagnostic output remains in `log.txt` and `log.json`.

## Validation

- A real, unmodified Hello PyTorch simulation completed three rounds with two clients, final global-model evaluation, and the summary shown above.
- Real simulation and POC success/failure runs verified result retention, bounded error reporting, service cleanup, and consistent presentation across environments.
- The broader Recipe, reporting, status, logging, server-log, and environment suites passed. After the final simplification and replay-boundary fix, the affected suites passed **167 tests** with 3 optional-dependency skips.
- All **3** fast real-output integration cases passed for default concise, explicit concise, and progress behavior, and all **3** Hello PyTorch simulation/POC continuity cases passed.
- `./runtest.sh -s --skip-install` passed.
- The documentation HTML build passed with 118 pre-existing warnings.

## Diff size against `upstream/main`

Counts include blank lines, comments, and docstrings.

| Area | Files | Added | Removed | Net | Why |
|---|---:|---:|---:|---:|---|
| Production code and configuration | 22 | +1049 | -157 | +892 | Presentation plus the cross-environment support described below |
| Test Python | 16 | +2076 | -196 | +1880 | Unit and integration coverage for success, failure, compatibility, malformed artifacts, encoding, bounded output, and environment behavior |
| Documentation and examples | 11 | +98 | -18 | +80 | Progress-mode guidance, supported-mode references, remote diagnostic access, and example help text |
| **Total** | **49** | **+3223** | **-371** | **+2852** | |

### Why the production diff is 892 net lines

| Production area | Net LOC | Purpose |
|---|---:|---|
| Run and failure summaries | +431 | Bounded readers for existing metrics, evaluation, model, and log artifacts, plus terminal presentation from `Run.get_result()` |
| Progress log mode and metric presentation | +276 | Opt-in formatter/filter behavior, bounded metric tables, round/client/aggregation events, and expected-shutdown noise reduction |
| POC/production monitoring and bounded log retrieval | +136 | Replays complete records from bounded structured/text tails and retrieves bounded server/client error tails without changing transport |
| Shared status classification | +36 | Gives monitoring and final presentation one definition of completed, failed, aborted, and not-scheduled outcomes |
| Recipe argument and environment handling | +13 | Moves export parsing from `spec.py` into shared `argparse` handling (the move itself is +2 net) and connects the selected mode across execution environments |
| **Total production code and configuration** | **+892** | |

### Production file breakdown

| File | Purpose | Added | Removed | Net |
|---|---|---:|---:|---:|
| `nvflare/recipe/_failure_summary.py` | Bounded client-log collection, parsing, peer attribution, and failure formatting | +205 | 0 | +205 |
| `nvflare/recipe/_run_summary.py` | Discover saved artifacts and render the final summary | +177 | 0 | +177 |
| `nvflare/fuel/utils/log_utils.py` | Add progress mode, preserve existing modes, and provide bounded, severity-aware rendering helpers | +166 | -2 | +164 |
| `nvflare/recipe/_args.py` | Shared Recipe argument parsing with standard `argparse`, including moved export logic and deferred validation errors | +88 | 0 | +88 |
| `nvflare/recipe/session_mgr.py` | Bounded POC/production progress, complete-record text fallback, and unavailable-log guidance | +130 | -22 | +108 |
| `nvflare/app_common/widgets/metrics_artifact_writer.py` | Render bounded round, client, aggregation, and duration progress from existing events | +97 | -1 | +96 |
| `nvflare/recipe/run.py` | Print terminal outcomes, artifacts, and failure details from `get_result()` | +49 | 0 | +49 |
| `nvflare/fuel/flare_api/job_status.py` | Shared current and legacy terminal-status classification | +36 | 0 | +36 |
| `nvflare/private/fed/server/job_cmds.py` | Retrieve bounded, record-aligned remote log tails | +29 | -8 | +21 |
| `nvflare/app_common/workflows/cross_site_model_eval.py` | Announce the start of cross-site evaluation; final values come from the bounded result summary | +5 | 0 | +5 |
| `nvflare/recipe/sim_env.py` | Simulation context, status, and failure summary | +14 | -4 | +10 |
| `nvflare/recipe/spec.py` | Delegate moved argument handling, reject recorded argument errors, and run presentation | +9 | -95 | -86 |
| Other supporting files (10 files, each under 10 net LOC) | Logging configuration, bounded-log API wiring, client lifecycle alignment, CLI help, and type annotations | +44 | -25 | +19 |
| **Total production code and configuration** | | **+1049** | **-157** | **+892** |

`_args.py` is primarily a relocation from `spec.py`: the two rows are +88 and -86 net lines, respectively, for a combined **+2 net lines**. The earlier hand-written token scanner was replaced with standard `argparse` handling.


nvflare/recipe/sim_env.py
```diff
@@ -128,6 +128,8 @@ def deploy(self, job: FedJob):
                 )
 
             workspace = os.path.join(self.workspace_root, job_id)
+            num_clients = len(self.clients) if self.clients is not None else self.num_clients
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in 0644e0e42: the preparation message now uses self.num_clients, which the constructor has already resolved. Regression coverage includes positive num_clients with clients=[], count-only configuration, explicit names, and matching count/names. All 22 SimEnv unit tests and the project style check pass.

nvflare/app_common/workflows/fedavg.py
```diff
@@ -164,7 +164,7 @@ def run(self) -> None:
                     "falling back to in-memory tensor download"
                 )
 
-            self.info(center_message("Start FedAvg."))
+            self.info(center_message(f"Starting FedAvg training: {self.num_rounds} rounds."))
```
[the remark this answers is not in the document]
chesterxgchen: Keeping center_message() intentionally to preserve the centered, visually separated progress messages requested during maintainer review. The target base (upstream/main at 53ba7ee56) already uses these banners for FedAvg start, rounds, and completion; retaining them is not a formatting regression relative to main. The noise reduction in this PR comes from filtering routine executor/transfer messages and replacing monitor metadata dumps/dots, while preserving meaningful workflow output. A matched three-round NumPy run on main and the PR reduced raw console output from 182 to 134 lines, with all six per-client metric messages, aggregation progress, and matching round metrics retained. The PR description includes both complete captured outputs and explicitly documents the banner decision. The integration assertions protect client metrics and aggregation progress; they intentionally do not require single-line banner formatting.

nvflare/recipe/session_mgr.py
```diff
@@ -12,29 +12,73 @@
...
+                    continue  # An in-flight final line may be incomplete; retry on the next callback.
+                if not isinstance(record, dict):
+                    continue
+                name = record.get("fullName", "")
+                level = record.get("levelname", "INFO")
+                if not isinstance(name, str) or not isinstance(level, str):
+                    continue
+                if not name.endswith(".progress") and level not in ("WARNING", "ERROR", "CRITICAL"):
+                    continue
+                if line in state["seen"]:
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in c3f9e88c9 using a bounded tail and deduplication strategy. The monitor selects the last 200 lines, parses at most 64 KiB, and retains at most 200 SHA-256 hashes from the current tail instead of every displayed line over the life of the job. The behavioral test runs repeated 1,000-record snapshots and checks the output bound, bounded hash state, and overlapping-snapshot suppression. The existing API still transfers snapshots capped at 5 MiB; this does not introduce an incremental transport or guarantee network latency. The PR description documents that limitation and that live output is a preview, with saved logs providing the complete history. The existing concise mode now uses this presentation; the additional progress mode has been removed.

nvflare/app_common/workflows/cross_site_model_eval.py
```diff
@@ -475,6 +478,10 @@ def _save_validation_result(self, client_name: str, model_name: str, dxo, fl_ctx
         self.log_info(
             fl_ctx, f"Saved validation result from client '{client_name}' on model '{model_name}' in {file_path}"
         )
+        self.logger.getChild("progress").info(
+            f"  {json.dumps(client_name)} | {json.dumps(model_name)}\n"
+            f"    {json.dumps(dxo.data, default=to_serializable)}"
+        )
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in c3f9e88c9. Evaluation progress now uses the same bounded scalar-metric formatter as training and the end summary: at most six entries, shortened names/strings, and a saved-result reference for structured values or additional metrics. It no longer serializes or stringifies arbitrary application objects for display. A behavioral test saves an actual DXO containing 100,000 sample values, a 100,000-character string, and 100 additional metrics; it verifies the saved payload is intact while the progress message remains under 1,000 characters. All 207 unit tests and 5 real simulation/POC integration cases pass, as does the project style check.

nvflare/fuel/utils/log_utils.py
```diff
@@ -44,17 +47,33 @@ class LogMode:
...
     "fmt"
 ] = "%(asctime)s - %(identity)s - %(fullName)s - %(levelname)s - %(fl_ctx)s - %(message)s"
 verbose_log_dict["loggers"]["root"]["level"] = "DEBUG"
 
+# A focused view using ordinary log records. Detailed file handlers are unchanged.
+concise_log_dict["formatters"]["progressFormatter"] = {"()": "nvflare.fuel.utils.log_utils.ProgressFormatter"}
+concise_log_dict["filters"]["ProgressFilter"] = {"()": "nvflare.fuel.utils.log_utils.ProgressLogFilter"}
+for handler_name in ("consoleHandler", "FLFileHandler"):
+    concise_log_dict["handlers"][handler_name]["formatter"] = "progressFormatter"
+    concise_log_dict["handlers"][handler_name]["filters"] = ["ProgressFilter"]
```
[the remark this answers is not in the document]
chesterxgchen: Keeping the revised concise behavior intentionally, following explicit maintainer direction to improve the existing concise mode rather than introduce another configuration mode. The console and log_fl.txt are the focused human-readable view; detailed application INFO records are retained in log.txt/log.json, and msg_only or full remains available for console diagnostics. The PR description and logging documentation now explicitly describe this presentation change, including the default remote monitoring preview. The separate-mode suggestion would contradict that requested UX, so it is not being applied. Tests that inspect transport/executor diagnostics now explicitly select full instead of relying on the concise default. The final end-of-run summary supplements intermediate progress with actual saved training/evaluation results and artifact locations.

nvflare/recipe/sim_env.py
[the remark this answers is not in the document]
chesterxgchen: Follow-up fixed in b599f18a8: the presentation test exposed an existing deployment mismatch for clients=[]. SimEnv now normalizes an empty list to None, so num_clients=2 is propagated as count-based configuration. A new behavioral test leaves FedJob.simulator_run, client resolution, and export real and intercepts only subprocess.Popen; it confirms the actual simulator command contains -n 2, -c site-1,site-2, and -t 2. This covers the previously mocked-away failure path.

nvflare/app_common/workflows/fedavg.py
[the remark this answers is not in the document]
chesterxgchen: Final implementation in b599f18a8 restores FedAvg diagnostic messages exactly to upstream, including center_message(). The workflow file is now absent from the PR diff. This preserves existing Finished FedAvg. and zero-based Round N started. assertions; the focused concise view independently renders one-based round progress through the existing metrics writer. External-trainer tests explicitly select full logging because they verify diagnostics. Six affected real end-to-end tests passed, with the TensorFlow case skipped; concise and message-only real NumPy tests also passed.

nvflare/app_common/widgets/metrics_artifact_writer.py
```diff
@@ -95,6 +111,29 @@ def handle_event(self, event_type: str, fl_ctx: FLContext):
...
+    def _log_progress_metrics(self, label, metrics):
+        values = {m["name"]: m["value"] for m in metrics}
+        if not values:
+            self.logger.getChild("progress").info(f"  {json.dumps(label[:128])} · no metrics reported")
+            return
+        first = self._progress_columns is None
+        if first:
+            self._progress_columns = list(values)[:2]
+        self.logger.getChild("progress").info(
+            format_metric_table([(label, values)], columns=self._progress_columns, header=first)
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in d96eb1eee. The live table now selects columns from each displayed result and repeats the headings when the metric names change, including for the aggregated row. This preserves aligned output for ordinary rounds while displaying later clients’ differing metrics instead of all-missing rows. A regression test covers accuracy on site-1, loss on site-2, and weighted_loss from aggregation. The existing two-column display limit and artifact reference still apply.

nvflare/recipe/_failure_summary.py
```diff
@@ -0,0 +1,171 @@
...
+        data = content.encode("utf-8")
+        if len(data) > _MAX_LOG_BYTES:
+            data = data[-_MAX_LOG_BYTES:].partition(b"\n")[2]
+        (target / "error_log.txt").write_bytes(data)
+
+
+def _records(data, plain_text):
+    if plain_text:
+        for match in _TEXT_ERROR.finditer(data.decode("utf-8", errors="replace")):
+            yield dict(name=match[1], levelname=match[2], message=match[3])
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in 2d491710e. The text-log reader now separates the leading FL context from the message and retains it as fl_ctx, allowing the same peer/peer_rc handling as JSON logs. Regression coverage generates text with the actual BaseFormatter and verifies both grouping/removal of a linked TASK_ABORTED record and preservation of an unrelated peer error, including server-only downloaded workspace layouts. All 20 failure-summary/session tests and the full project style check pass.

chesterxgchen:
Addressed in 05c6a7352.

The two earlier regressions are fixed:
- Shared server setup supplies the existing MetricsArtifactWriter when a job has none. A real two-client, two-round plain FedJob export, loaded by the simulator CLI as JSON, now shows each round and aggregation. Explicit writers are preserved; Recipe integration tests assert that round headings appear only once.
- Failure summaries try error_log.txt when the bounded log.json tail has no usable error. Both reads consume the shared 20-file allowance. Tests cover empty, malformed, INFO-only, and oversized JSON, JSON preference when usable, and the total read limit.

Disposition of the eight additional findings:

| Finding | Resolution |
|---|---|
| 1. Terminal status classification | FINISHED:CAN_NOT_SCHEDULE now displays “Not scheduled” and does not request training error logs. ABANDONED remains an unsuccessful execution: JobRunner.update_unfinished_jobs applies it to previously running jobs, so it does not establish that training never ran. |
| 2. Filtered metrics disappear | Accepted updates whose metrics are all filtered now retain a client line saying “no displayable metrics.” The count continues to describe accepted updates. |
| 3. Private Run context setup | Retained intentionally. Recipe supplies metadata it owns internally; direct Run(env, job_id) construction has an environment-only fallback. No demonstrated broken caller, and adding constructor metadata parameters would expand the public API solely for presentation. |
| 4. Constructor log configuration | The cited SimEnv path does not use SessionManager. Its constructor value reaches FedJob.simulator_run. The existing simulator-boundary test now explicitly covers concise and full constructor values. PocEnv and ProdEnv do not expose that constructor parameter. |
| 5. Parser not tied to formatter tests | Already covered: test_groups_client_errors_and_omits_only_the_linked_abort generates text with the actual BaseFormatter, including bracket-bearing workflow names, and checks extracted exceptions and peer attribution. |
| 6. Repeated tail reads | One private bounded-tail helper now serves all three readers. Raw byte tails remain appropriate for transport; summaries explicitly request complete-record tails. Tests verify byte bounds and both contracts. |
| 7. Formatter allocation | Retained intentionally: BaseFormatter stores mutable per-record state, while console and file handlers can share ProgressFormatter under different handler locks. Constructing it locally avoids cross-handler state races without adding locks or thread-local caches. No measured performance regression was supplied. |
| 8. Test filename | Renamed to _failure_summary_test.py. |

Validation: 272 targeted tests passed, 3 skipped; all four real simulator output cases passed; full project style check passed. This follow-up adds 13 net production Python lines. The PR description retains the actual output examples and includes refreshed overall and per-file LOC tables.

chesterxgchen:
Fixed in d1ab06336. Server setup now checks the explicit metrics_artifact_writer component ID as well as existing MetricsArtifactWriter instances before adding a default. Independent and no-op FLComponent replacements therefore retain control of collection and output paths.

A behavioral test uses real BaseFedJob/FedJob export and ServerJsonConfigurator loading, without mocking component construction. The replacement case failed before the fix and passes afterward. Companion cases verify that configured writers retain their custom output directory, writers under other IDs are not duplicated, and a missing writer still gets the default.

All 40 targeted tests and the full project style check passed. The production fix adds two net lines; the PR description and LOC tables are updated.

chesterxgchen:
Pushed two focused follow-ups:

- a4ceba0d6: metrics_artifact_writer is Optional[FLComponent] in the unified, PyTorch, and TensorFlow BaseFedJob signatures. Their API docstrings describe independent replacements; runtime validation is unchanged.
- 8e7ad291e: full/verbose Recipe monitoring retains resource_spec/deploy_map metadata on status changes, table labels always have a separator before values, and round/outcome duration padding excludes newline characters and uses the same column.

Disposition of the runtime/output review:

| Finding | Resolution |
|---|---|
| 1. Metadata visibility | Restored in full/verbose monitoring; concise retains the intended compact output. Tests verify metadata is emitted only on status changes. |
| 2. Automatic writer scope/replacements | Shared reporting across simulation, POC, and production is intentional. The standard metrics_artifact_writer ID reserves the role for independent replacements; existing MetricsArtifactWriter instances under other IDs are also respected. An arbitrary component under an unrelated ID does not declare that role, and setup cannot infer semantic equivalence. The documented ID provides that explicit contract. |
| 3. Missing logging-argument value | Verified in a real subprocess: argparse prints “argument --log_config: expected one argument” and exits 2, without a traceback. This is a normal CLI usage error, not an unhandled exception traceback. Silently ignoring this shared option could start a job with unintended settings. |
| 4. Missing table separator | Fixed and covered with a full-width label/value regression case. |
| 5. Newline-aware padding | Simplified: prepend newlines after padding the actual line; round and run durations now explicitly start at the same column. |
| 6. Repeated tail polling | Retained as the bounded existing-API tradeoff: at most one normal poll per five seconds, 64 KiB/200 parsed lines, and 200 digests. An offset/resume protocol is not provided by the current endpoint; adding one would broaden this reporting PR. |
| 7. Sequential error-log requests | Retained intentionally. target="all" would reintroduce the previously fixed unbounded fan-out/transfer problem. Each named-site request has a server-enforced byte cap and the collection stops after 20 sites. |
| 8. Terminal-state helpers | Both CLI and session terminal helpers already exist on the PR base, 53ba7ee56. Recipe outcome classification reuses the session helper. New standard FINISHED:* statuses are recognized by the prefix logic without maintaining another list. Broader legacy-helper consolidation is pre-existing scope. |

Validation: 199 targeted tests passed, 3 skipped; full project style check passed. Unified/PyTorch type hints and independent replacement construction were verified; the TensorFlow annotation was verified from source without a TensorFlow runtime. Fresh real Hello PyTorch success and intentional Hello NumPy failure captures are included in the PR description. Overall and per-file LOC tables are refreshed.

chesterxgchen:
Addressed in fde38e95f:

- Console encoding: the existing console formatter and Recipe output now use an encoding-aware fallback. A real Hello NumPy run with PYTHONIOENCODING=ascii completed through result retrieval without encoding/logging errors. Tests also exercise ASCII/CP1252 success and failure summaries, Unicode job names, and closed output streams.
- Missing traceback: when JSON supplies an error message without traceback details, the summary checks the corresponding text error log within the same 20-file/1-MiB-per-file limits. Tests use the actual JSON/text formatters with logger.exception("Training failed") and verify the application location.
- Message-less exceptions: the parser recognizes bare exception types such as AssertionError and extracts the final exception before truncation. Coverage uses an actual nested assertion traceback exceeding the display limit.

The logging simplification is substantive: removed ProgressFormatter, ProgressFilter, and .progress child-loggers; ordinary component loggers now use the existing concise filter and console formatter configuration. This also makes ordinary evaluation-controller INFO messages visible. The PR includes a new complete real output capture (89 lines), preserving the original before example.

Shared argument parsing/state moved from public spec.py into private _args.py, preserving flag behavior and DEFAULT_EXPORT_DIR.

Validation: 774 recipe/reporting unit tests passed, 18 skipped (two socket tests rerun outside the sandbox); final focused runs passed 104 encoding/failure/logging tests and 76 parser/spec tests. All 51 simulator-integration and example-export cases passed. Real UTF-8 success, ASCII success, and intentional training-failure runs completed as expected. Full project style check passed. PR description and both LOC tables updated.

chesterxgchen:
Fixed both findings in 6b66d9f52.

1. Inherited evaluation and metrics reporting now uses the defining module's ordinary logger. The existing concise configuration already admits those modules, so GlobalModelEval, HECrossSiteModelEval, and user subclasses retain inherited reporting locally and in remote replay. Other subclass diagnostics retain their own logger. No extra mode, formatter, record marker, or child-logger channel was added.
2. Exception extraction now selects the header after the final traceback frame instead of the last identifier-looking line. Actual multiline ValueError messages, including a continuation that itself looks like a typed exception, preserve the original type/message. Bare assertions and chained exceptions remain covered.

125 targeted tests passed, including real subclass result saving, existing concise filtering, actual JSON formatting and remote replay, custom-writer reporting, and real exception tracebacks. The full project style check passed. These changes add 10 net production Python lines. The PR description and both LOC tables have been updated; the captured output examples are preserved.

chesterxgchen:
Fixed in 88b54ab93: addressed the frame-less traceback finding and checked the adjacent parser boundaries together.

Frame and exception searches now begin inside the final traceback section, so a prefix such as Failure cannot become the exception header and a previous chained exception cannot supply its frame. Only Python's explicit/implicit exception-chain separators advance to another section; literal traceback-marker text within an exception message does not.

The initial 24 regression combinations all failed on the previous commit and pass with this fix. The final coverage also includes implicit chains and marker-bearing multiline messages, alongside the existing framed, bare, multiline, JSON/text fallback, and bounded-read cases. Final targeted validation: 129 passed, 3 existing skips. Full project style check passed.

The production change is 10 net lines in _failure_summary.py. Updated the PR description and both LOC tables; existing output examples are preserved.

nvflare/fuel/utils/log_utils.py
```diff
@@ -42,19 +45,36 @@ class LogMode:
...
-concise_log_dict["formatters"]["consoleFormatter"]["fmt"] = "%(asctime)s - %(levelname)s - %(message)s"
-concise_log_dict["handlers"]["consoleHandler"]["filters"] = ["ConciseFilter"]
+# Configure the existing concise view using ordinary component loggers.
+concise_log_dict["formatters"]["consoleFormatter"]["fmt"] = "%(message)s"
+concise_log_dict["filters"]["ConciseFilter"] = {
+    "()": "nvflare.fuel.utils.log_utils.LoggerNameFilter",
+    "logger_names": [
+        "nvflare.app_common.widgets.metrics_artifact_writer",
+        "nvflare.app_common.workflows.cross_site_model_eval",
+    ],
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in 29fe1c3a2. Concise now uses the existing LoggerNameFilter exclusion option to suppress class-level diagnostic INFO while retaining the defining module's reporting records. Warnings/errors still pass through regardless of class, and inherited reporting remains visible for GlobalModelEval, HECrossSiteModelEval, and user subclasses. Detailed log.txt/log.json and full/verbose retain the diagnostic INFO.

This is four lines of configuration using the existing filter, with no additional logging mode or formatter.

Validation: the new base-controller regression failed before the change; all 93 targeted tests now pass, including actual saved-result logs through local filtering and JSON-based remote replay. Full project style check passed. A real, unmodified three-round Hello PyTorch run retained all round/evaluation results and the final summary; formatter/model-location/result-path diagnostics were absent from the console and present in server/log.txt. The PR description preserves the before capture and now includes the complete 77-line after capture plus refreshed LOC tables.

nvflare/recipe/_run_summary.py
```diff
@@ -0,0 +1,165 @@
...
+                        lines.append(format_metric_table(rows, label="Client"))
+                    if len(evaluations) > 6 or len(metric_names) > 2:
+                        lines.append("  Additional evaluation results are available in the saved artifact.")
+            except (OSError, ValueError, TypeError):
+                pass
+            artifacts.append(f"  Evaluation {evaluation_path.relative_to(root)}")
+        app_dir = run_dir / "app_server"
+        if app_dir.is_dir():
+            if any(p.is_file() and p.suffix in (".pt", ".pth", ".npy", ".npz") for p in app_dir.iterdir()):
+                artifacts.insert(0, f"  Models    {app_dir.relative_to(root)}/")
```
holgerroth: [P2] Include standard NumPy and TensorFlow model locations in the summary

Model discovery only checks files directly inside `app_server` with `.pt`, `.pth`, `.npy`, or `.npz` extensions. The default NumPy persistor saves to `<run_dir>/models/server.npy`, and the TensorFlow persistor defaults to `app_server/tf_model.weights.h5`, so both supported configurations silently omit the `Models` location from the new run summary.

I reproduced the NumPy case using an actual `NPModelPersistor.save_model()` call: `server/simulate_job/models/server.npy` exists, but `result_summary()` reports only training metrics and the metrics directory. A file with the default TensorFlow filename is also excluded by this check. Please include these standard paths/formats and add regression coverage; model weights do not need to be loaded for discovery.

nvflare/fuel/utils/log_utils.py
```diff
@@ -42,19 +45,40 @@ class LogMode:
...
+}
+# Presentation uses the module logger; class loggers retain diagnostic INFO.
+concise_log_dict["filters"]["ConciseFilter"]["exclude_logger_names"] = [
+    name + "." for name in concise_log_dict["filters"]["ConciseFilter"]["logger_names"]
+]
+for handler_name in ("consoleHandler", "FLFileHandler"):
+    concise_log_dict["handlers"][handler_name]["filters"] = ["ConciseFilter"]
```
holgerroth, line 67: Please preserve the existing `log_fl.txt` filtering for compatibility

The new concise console experience looks useful, but please apply its focused filter only to `consoleHandler` and leave `FLFileHandler` using its existing `FLFilter`. This loop currently changes the contents of `log_fl.txt` too, suppressing application/training INFO records that existing scripts or user workflows may parse. Keeping detailed records in `log.txt` and `log.json` does not preserve compatibility for consumers already reading `log_fl.txt`.

In the actual concise Hello PyTorch run, the original `Finished FedAvg.` record remains in `server/log.txt` but disappears from `server/log_fl.txt`. Please add a regression test showing that representative application INFO records remain in `log_fl.txt` while being omitted from the concise console, and align the documentation accordingly.

For clarity, the current Auto-FL score extractor reads `cross_val_results.json`, and its reporting skill reads `results.tsv`; I verified the extractor still works against the concise demo results. This request is to preserve the existing training-log interface, not a claim that Auto-FL scoring is currently broken.

examples/advanced/hello-pt-environments/README.md
```diff
@@ -191,6 +191,9 @@ with downloads disabled, so concurrent processes do not race while writing it. T
 nonzero size; torchvision performs its own integrity checks when loading. This option is useful for experimentation
 but is not a federated data partition.
 
-For production, `--data_root` is a path on each client—not on the admin machine running `job.py`. Every site operator
+For simulation and POC, relative `--data_root` paths are resolved from the submission working directory before
+running or exporting the recipe. This only normalizes the path; it does not validate or download the cache.
```
pcnudde: I do not understand this text. "This only normalizes the path; it does not validate or download the cache"

examples/hello-world/hello-pt/job.py
```diff
@@ -51,7 +52,8 @@ def define_parser() -> argparse.ArgumentParser:
 def create_recipe(args):
     train_args = ["--dataset", args.dataset]
     if args.dataset == "cifar10":
-        train_args.extend(("--data_root", args.data_root))
```
pcnudde, line 54: Is this change related to the log change, or does it address another issue

examples/hello-world/hello-pt/prepare_data.py
```diff
@@ -108,14 +108,22 @@ def __getitem__(self, index):
...
-    batch_dir = Path(data_root).expanduser() / "cifar-10-batches-py"
+    expanded_root = Path(data_root).expanduser()
+    batch_dir = expanded_root / "cifar-10-batches-py"
     missing = [
         name for name in _CIFAR10_FILES if not (batch_dir / name).is_file() or (batch_dir / name).stat().st_size == 0
     ]
     if missing:
         command = f"python {shlex.quote(prepare_script)} --data_root {shlex.quote(str(data_root))}"
+        hint = ""
+        if not expanded_root.is_absolute():
```
pcnudde: is this related to this PR? Do we need this complexity ion the example?

nvflare/fuel/flare_api/api_spec.py
```diff
@@ -326,6 +328,9 @@ def get_job_logs(
             tail_lines: deprecated compatibility option to return the last N lines
             grep_pattern: deprecated compatibility option to return matching lines
             log_file_name: internal log file selector. Defaults to ``log.txt``.
+            max_bytes: positive UTF-8 log-byte limit per site, enforced on the server before
```
pcnudde, line 331: Why do we need this?

pcnudde (changes requested):
Overall it feels like there are many different issues mixed in this PR. 
Do we need it all? Seems that just printing a summary at the end should be simple.

YuanTingHsieh:
After reviewing the changes, I have five design and maintainability points to address:

1. **Keep the Recipe documentation focused on current user-facing behavior.** The additions to `docs/user_guide/data_scientist_guide/recipe_api.rst` are too extensive for explaining run output. Please shorten them to what users see, how to select a logging mode, what `get_result()` returns and summarizes, and where to find logs or retain the workspace. Remove change-history language, implementation details, and repeated assurances. Keep the mode comparison in the logging guide and link to it.

2. **Make progress reporting explicit instead of depending on module versus class logger names.** In `cross_site_model_eval.py`, `_logger.info()` is used for visible progress while `self.log_info()` is treated as diagnostic output. Both are INFO records, but their visibility depends on filtering rules elsewhere. This is an implicit convention that maintainers must know. Please consider a shared reporting interface that makes the intent clear at the call site.

3. **Classify expected shutdown consistently without matching human-readable messages.** Downgrading normal shutdown to INFO in `client/in_process/api.py` makes sense, but checking the exact string `"END_RUN received"` in two places is brittle. The normal API shutdown path uses a different reason and still logs a warning. Please use an explicit stop-reason classification, or at minimum a shared constant and consistent handling of expected shutdown paths.

4. **Clarify the decision to install a metrics writer automatically for every server job.** The change in `server_json_config.py` adds a `MetricsArtifactWriter` when none is configured, including outside concise mode. This changes default job behavior and creates metric artifacts; it is more than a presentation change. Please justify whether this should be a default server capability and whether it belongs in this PR or a separate change.

5. **Give cross-package shared helpers a clear interface.** Private imports within the Recipe package are reasonable, but importing helpers such as `_read_log_tail` from `fuel.utils.log_utils` and `_job_status_outcome` from `fuel.flare_api.flare_api` makes other packages depend on implementation details. Please define an appropriate shared interface/location for functionality intended to be reused across packages.

The placement of execution reporting in `Run`, with deployment and remote communication delegated to environments and `SessionManager`, makes sense. These concerns are about the implementation and default behavior, not about those modules living inside the `nvflare.recipe` package.

nvflare/fuel/utils/log_utils.py
chesterxgchen: frankly I disagree. if user use log files to extract information, the log.txt is source of truth. we create log_fl.txt intended for better readability and only isolate the log that data scientists care. if we are now change not change this one, that means we have maintain two log formats and details for compatibility, which not only increase the maintenance effort, but also defeat the original purpose of two log files. In that case, we might just drop the log_fl.txt.

holgerroth: Agreed, Chester. That distinction makes sense, and I withdraw the request to preserve the old filtering of `log_fl.txt`.

Let's follow this log-file policy: keep `log.txt` and `log.json` detailed; let `log_fl.txt` follow the readable presentation. Prefer structured result artifacts for automated score extraction.

Our Auto-FL check supports this: scoring reads `cross_val_results.json`, and reporting reads `results.tsv`, so it does not depend on the presentation in `log_fl.txt`.

nvflare/recipe/_run_summary.py
chesterxgchen: Fixed in d547ffd74. Result summaries now discover standard model files in both app_server and models, including TensorFlow .h5/.keras files and NumPy server.npy, without loading model contents. Regression coverage exercises PyTorch, TensorFlow, and NumPy locations. The targeted suite passes with 470 tests and 3 skips.

examples/advanced/hello-pt-environments/README.md
chesterxgchen: Removed in d547ffd74. The CIFAR cache-path documentation and related path-normalization changes were unrelated to the logging and run-summary scope, so the README is restored to upstream/main.

examples/hello-world/hello-pt/job.py
chesterxgchen: Removed in d547ffd74. The data-root normalization was not required for the logging and summary feature, so the Hello PyTorch job is restored to upstream/main.

examples/hello-world/hello-pt/prepare_data.py
chesterxgchen: Removed in d547ffd74. The additional relative-cache error handling and its tests were outside this PR's logging and summary scope, so prepare_data.py is restored to upstream/main.

nvflare/fuel/flare_api/api_spec.py
chesterxgchen: Retained and clarified in the PR description. Failure summaries for POC and production need a bounded client-log tail. max_bytes lets the server enforce the 1 MiB limit before transfer, so the submitter never falls back to downloading an unbounded remote log. The shared tail and status helpers now have explicit public interfaces instead of private cross-package imports.

chesterxgchen:
Addressed in d547ffd74 after rebasing onto current `upstream/main`.

1. **Recipe documentation:** reduced the run-output addition from 176 lines to 53 lines. It now covers only what users see, `--log_config`, `get_result()`, success/failure summaries, result retention, and where to find full logs. The logging guide retains the mode comparison.
2. **Explicit progress reporting:** added the shared `log_progress()` interface and used it at every user-facing metrics-writer and cross-site-evaluation call site. Class `log_info()` calls remain diagnostics, so intent is visible where a message is emitted.
3. **Expected shutdown:** replaced repeated string comparisons with one expected-stop classification covering both `END_RUN` and explicit Client API shutdown. Both paths log at INFO; unexpected stops remain warnings.
4. **Default metrics writer:** retained intentionally. The end summary reads the existing metrics artifacts, and standard FedJob/JSON jobs otherwise may not install the writer that creates them. This is the shared data source for the presentation in every logging mode, not a concise-only side effect. Explicit writers and components registered under the standard writer ID remain authoritative and are not duplicated.
5. **Shared helper interfaces:** promoted the bounded-tail and console-text helpers to named interfaces and moved terminal-status classification into `nvflare.fuel.flare_api.job_status`, removing Recipe imports of private helpers from implementation modules.

The same commit also:

- discovers standard PyTorch, TensorFlow (`.h5`/`.keras`), and NumPy (`models/server.npy`) model locations without loading weights;
- removes the unrelated CIFAR cache-path source, documentation, and test changes; and
- preserves the focused `log_fl.txt` behavior, per the maintainer decision already recorded in its review thread.

Validation: 470 targeted tests passed with 3 skips; all 4 real-output integration cases passed; project style checks passed; and the documentation build passed with 117 pre-existing warnings.

nvflare/client/in_process/api.py
```diff
@@ -327,22 +335,22 @@ def __ask_to_abort(self, topic, msg, databus):
...
+            self.logger.log(_stop_log_level(self.stop_reason), f"request to stop the job for reason {self.stop_reason}")
             self.fl_model = None
             return False
 
         return True
 
     def shutdown(self):
         self.stop = True
+        self.stop_reason = _API_SHUTDOWN_STOP_REASON
         self.event_manager.fire_event(TOPIC_STOP)
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in 97cb2fd34. shutdown() now passes the expected API-shutdown reason as TOPIC_STOP event data, so the callback preserves the reason and both stop messages remain INFO. A behavioral test calls shutdown() through the real synchronous DataBus callback and then the continuation check; the focused shutdown and shared-argument isolation run passes all 100 tests.

chesterxgchen:
Rebased onto current `upstream/main` at 13d3c4a22 and addressed the requested scope/design cleanup.

For the concern that the PR mixed unrelated issues, the CIFAR cache-path, example job, preparation, README, and associated path-only test changes have been removed. The remaining production changes support one user-visible flow: concise live progress plus success/failure summaries from `Run.get_result()` across simulation, POC, and production.

Yuan-Ting's five design points are reflected in the current head:

1. The Recipe run-output documentation was reduced from 176 lines to 53 and now covers only visible output, `--log_config`, `get_result()`, failure/success summaries, logs, and workspace retention.
2. User-facing workflow messages use the explicit shared `log_progress()` interface; ordinary component logging remains diagnostic.
3. Expected `END_RUN` and Client API shutdown reasons share one classification and remain INFO, with the shutdown reason preserved through the event callback.
4. The default metrics writer is documented and retained as the artifact source used by end-of-run summaries in all logging modes. Explicit writers or replacements registered under its component ID remain authoritative.
5. Cross-package consumers now use public `read_log_tail()`, `console_text()`, and the shared `fuel.flare_api.job_status` classification module.

The PR description now shows both successful output and an actual failure summary, explains why artifact collection and remote status/log retrieval are needed, and lists only files with at least 10 net LOC in the detailed table.

Validation after the rebase: 153 targeted tests passed with 3 skips, and the full project style check passed. Please re-review the current head.

YuanTingHsieh:
Re-reviewed and reproduced three remaining issues on **5c7fbd357b17288ecfb194b02cf3baae7cc88b23**. I added three local regression checks; each fails on the behavior described below.

1. **[P2] Preserve progress for workflows outside the new presentation's scope.**  
   [Concise filter, `log_utils.py:50–55`](https://github.com/NVIDIA/NVFlare/blob/5c7fbd357b17288ecfb194b02cf3baae7cc88b23/nvflare/fuel/utils/log_utils.py#L50-L55)

   The new INFO allowlist applies globally, even though statistics and other workflows are declared outside this PR's scope. Those workflows lose their existing progress without receiving replacement reporting.

   Reproduction: execute the real `StatisticsController.control_flow()`, stubbing task execution, result waiting, and post-processing so the check needs no clients. Capture its actual log records and apply the previous concise filter and the new filter. The previous filter retains all three lifecycle messages; the new filter retains none:

   ```text
   Before: fed_stats control flow started.
           start post processing
           task fed_stats control flow end.
   After:  []
   ```

   Please preserve existing progress for unsupported workflows, or scope the new filtering to jobs with replacement progress reporting. Deferring a new statistics presentation should not remove its current status messages.

2. **[P2] Reset aggregation counts independently of ROUND_STARTED.**  
   [Metrics writer, `metrics_artifact_writer.py:100–106`](https://github.com/NVIDIA/NVFlare/blob/5c7fbd357b17288ecfb194b02cf3baae7cc88b23/nvflare/app_common/widgets/metrics_artifact_writer.py#L100-L106)

   `Scaffold` emits contribution and aggregation events but does not emit `ROUND_STARTED`. The new counter therefore accumulates across rounds.

   Reproduction: run the real `Scaffold._run_rounds()` for two rounds with two supplied client results per round, keeping SCAFFOLD aggregation and model updates real, stubbing transport/client selection/persistence, and routing events to the real writer. Each supplied result also goes through the writer's contribution-accept handler. Both rounds aggregate two results, but the completion messages are:

   ```text
   ✓ Aggregated 2 client updates
   ✓ Aggregated 4 client updates
   ```

   Please reset the count by contribution round or after aggregation, or ensure the supported controllers emit the required round lifecycle events. Add coverage for SCAFFOLD rather than only FedAvg.

3. **[P2] Enforce the live table's row limit across contributions.**  
   [Metrics writer, `metrics_artifact_writer.py:129–139`](https://github.com/NVIDIA/NVFlare/blob/5c7fbd357b17288ecfb194b02cf3baae7cc88b23/nvflare/app_common/widgets/metrics_artifact_writer.py#L129-L139)

   Each contribution passes a single row to `format_metric_table()`, so its ten-row limit never applies across the live round.

   Reproduction: start one round on a real `MetricsArtifactWriter` and deliver 25 accepted client contributions with `loss=0.5`. All 25 client rows are printed, with no omission notice. The expected assertion `len(displayed_client_rows) <= 10` fails with `25 <= 10`.

   Please track displayed client rows per round, retain the aggregate row, and emit an artifact reference when additional client rows are omitted. Keep all accepted contributions in the artifacts.

chesterxgchen:
Addressed all three findings in `1108b05e6`.

1. **Unsupported workflow progress:** concise filtering now recognizes explicitly marked presentation records only for the metrics writer and cross-site evaluation. Other `nvflare.app_common`/`nvflare.app_opt` workflows retain their existing lifecycle INFO, including the StatisticsController messages from the reproduction. High-volume client execution/transfer diagnostics remain excluded, and warnings/errors still pass through. Remote POC/production replay applies the same classification.
2. **SCAFFOLD aggregation counts:** the accepted-contribution counter now resets after every aggregation, independently of `ROUND_STARTED`. Regression coverage performs two real `scaffold_aggregate_fn()` aggregations with two contributions each and verifies both completion messages report two updates rather than two then four.
3. **Live row bound:** the writer tracks displayed client rows across the round, prints at most ten, emits one saved-artifact notice when more are omitted, and always retains the aggregate row. The regression supplies 25 accepted contributions and verifies ten live client rows, one omission notice, the aggregate output, and all 25 sites in `round_metrics.jsonl`.

Validation: 213 targeted statistics, SCAFFOLD/FedAvg, metrics-writer, concise-filter, cross-site-evaluation, and remote-replay tests passed with 2 skips. All 4 real-output integration cases passed, and the full project style check passed. The PR description and LOC tables are updated.

nvflare/fuel/utils/log_utils.py
```diff
@@ -42,19 +45,45 @@ class LogMode:
     default_log_dict = json.load(f)
 
 concise_log_dict = copy.deepcopy(default_log_dict)
-concise_log_dict["formatters"]["consoleFormatter"]["fmt"] = "%(asctime)s - %(levelname)s - %(message)s"
-concise_log_dict["handlers"]["consoleHandler"]["filters"] = ["ConciseFilter"]
+# Configure the existing concise view using ordinary component loggers.
+concise_log_dict["formatters"]["consoleFormatter"]["fmt"] = "%(message)s"
```
holgerroth, line 49: [P2] Preserve textual severity for warnings and errors in concise output

With this message-only format, `ColorFormatter` distinguishes severity only through color when stdout is a TTY. In redirected output, CI logs, or an agent's captured terminal output, INFO, WARNING, and ERROR records with the same message render identically. I reproduced this with the actual formatter and remote progress replay: all three levels printed only `connection unavailable`.

Could we retain textual WARNING/ERROR/CRITICAL labels for those records, while leaving ordinary progress messages uncluttered? A regression test covering non-TTY console output and remote replay would protect the debugging path. Keeping severity in the detailed files is useful, but those files may not be accessible to the person or agent reading the captured console.

nvflare/recipe/session_mgr.py
```diff
@@ -12,29 +12,98 @@
...
-    """Shared callback to print job meta during monitoring."""
-    # cb_run_counter is a dictionary that is passed to the callback and is used to keep track of the number of times the callback has been called
-    if cb_kwargs["cb_run_counter"]["count"] == 0:
-        print("Job ID: ", job_id)
-        print("Job Meta: ", job_meta)
+def _show_job_progress(session, job_id, state):
+    """Replay a bounded tail of existing server logs with bounded deduplication."""
+    try:
+        response = session.get_job_logs(job_id, target="server", log_file_name="log.json", tail_lines=200)
+        logs = response.get("logs", {})
```
holgerroth, line 44: [P2] Surface unavailable remote logs instead of silently showing status only

`get_job_logs()` can return an `unavailable` entry without raising an exception. Here that becomes an empty server log, so the warning in the exception handler never appears. Using the real server log handler with a workspace containing `log.txt` but no `log.json`, I got `{"logs": {}, "unavailable": {"server": "server log not available for this job"}}`; the monitor then printed only the job ID/status and never requested the existing text log.

This can hide diagnostics on custom or older deployments that retain text logs without JSON. Could we handle the `unavailable` response explicitly, give a once-per-job actionable notice when logs remain unavailable, and use a bounded `log.txt` fallback where appropriate? Please cover the normal unavailable response as well as exceptions in tests, with allowance for logs not existing yet during startup.

nvflare/recipe/session_mgr.py
```diff
@@ -78,31 +147,32 @@ def abort_job(self, job_id: str) -> None:
...
+        _print_output(f"Job {job_id} aborted successfully with message: {msg}")
         sess.close()
 
     def get_job_result(self, job_id: str, timeout: float = 0.0) -> Optional[str]:
         """Get the result workspace of the job."""
         sess = self._get_session()
         cb_run_counter = {"count": 0}
+        if os.environ.get(FL_LOG_LEVEL, LogMode.CONCISE) == LogMode.CONCISE:
+            cb_run_counter["progress"] = {"seen": set()}
         rc = sess.monitor_job(job_id, timeout=timeout, cb=_job_monitor_callback, cb_run_counter=cb_run_counter)
```
holgerroth, line 159: Clarify the diagnostic escape hatch for remote runs

The remote log replay is enabled only in concise mode here. With the real `SessionManager` and a mocked transport, concise mode made a log request and displayed the diagnostic, while full mode made no log requests and displayed job metadata only. Thus `--log_config full` is not a way to obtain full remote service logs; it also does not reconfigure those services. This is an access/documentation limitation, rather than a claim that a previously available full remote stream regressed.

Could the logging guidance explicitly distinguish simulator/local output from remote monitoring and give the supported retrieval path and its access requirements? An agent with access to the submitter's console is not guaranteed access to server/client `log.txt`. The default site streamer carries `error_log.txt`, and full client logs depend on configured collection, retained workspace contents, or authorized site access. The agreed policy of keeping `log.txt`/`log.json` detailed still makes sense; we should make the route to those diagnostics clear when the readable view is insufficient.

nvflare/fuel/utils/log_utils.py
chesterxgchen: Fixed in 5648237ff. The clean presentation now belongs to the new opt-in progress mode, while concise is restored to its NVFlare 2.9 formatter and filtering behavior. ProgressFormatter leaves INFO messages uncluttered and prefixes WARNING, ERROR, and CRITICAL in plain text, including remote replay. Tests cover non-TTY output and replayed warnings.

nvflare/recipe/session_mgr.py
chesterxgchen: Fixed in 5648237ff. Progress replay now detects the normal unavailable response, tries a bounded server log.txt tail when log.json is absent, and reports persistent unavailability once after a startup grace attempt. Exception handling remains non-fatal. Tests cover the text fallback, startup absence, one-time notice, and exceptions.

nvflare/recipe/session_mgr.py
chesterxgchen: Clarified in 5648237ff. The Recipe guide now distinguishes local progress output from remote monitoring: progress replay reads a bounded server log.json tail with a bounded log.txt fallback, does not grant client access or reconfigure services, points users to nvflare job logs, and explains that client logs must already be streamed or require authorized site access.

nvflare/recipe/session_mgr.py
```diff
@@ -12,29 +12,129 @@
...
+def _show_job_progress(session, job_id, state):
+    """Replay a bounded tail of existing server logs with bounded deduplication."""
+    try:
+        response = session.get_job_logs(job_id, target="server", log_file_name="log.json", tail_lines=200)
+        logs = response.get("logs", {})
+        text = logs.get("server", "")
+        structured = True
+        unavailable = response.get("unavailable", {})
+        if not text and isinstance(unavailable, dict) and "server" in unavailable:
+            fallback = session.get_job_logs(job_id, target="server", log_file_name="log.txt", tail_lines=50)
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in 739bb978e. The plain-text fallback now passes `max_bytes=64 * 1024` together with `tail_lines=50`, so the server bounds the transfer before the monitor applies its local tail and deduplication. The focused session-manager suite passes (21 tests), and the project style check passes.

nvflare/fuel/utils/log_utils.py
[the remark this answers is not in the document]
chesterxgchen: Final behavior supersedes my earlier response. In 5648237ff, `concise` was restored to the NVFlare 2.9 configuration and remains the default; the cleaner focused presentation and remote replay now use the separate opt-in `progress` mode. Tests cover the compatibility boundary and the progress behavior.

nvflare/recipe/_failure_summary.py
```diff
@@ -0,0 +1,192 @@
...
+from itertools import islice
+from pathlib import Path
+
+from nvflare.apis.fl_constant import ReturnCode
+from nvflare.fuel.utils.log_utils import _read_log_tail, wrap_log_message
+
+_MAX_LOG_BYTES = 1024 * 1024
+_MAX_LOG_FILES = 20
+_FRAME = re.compile(r'^  File "([^"\n]+)", line (\d+), in ([^\n]+)', re.MULTILINE)
+_EXCEPTION = re.compile(r"^[A-Za-z_][\w.]*(?::[^\n]*)?$", re.MULTILINE)
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in 6b930e3a8 and retained at the current head. Failure parsing now selects the exception header immediately after the final traceback frame, so later unindented continuation lines cannot replace the real exception. Behavioral tests cover multiline `ValueError` messages, chained exceptions, bare exceptions, and both JSON and text logs.

chesterxgchen:
Addressed the focused-progress filtering findings in `8054d5dfc` and `75c886ad5`.

Locally, `progress` now accepts INFO only when the record is explicitly marked by `log_progress()`. `JsonFormatter` persists that boolean, and remote replay restores it. For compatibility with older logs that do not contain the field, replay infers progress only from the two existing presentation loggers. Ordinary `nvflare.app_common`, `nvflare.app_opt`, custom training, statistics, model-content, and Newton-Raphson diagnostic INFO remains in `log.txt`/`log.json`; WARNING, ERROR, and CRITICAL records remain visible. The default `concise` behavior is unchanged.

Regression coverage verifies local filtering, serialized true/false markers, remote replay, and legacy-log fallback. The expanded reporting and real-output suite passes 178 tests, and `./runtest.sh -s --skip-install` passes. The PR LOC tables were refreshed for the current head.

chesterxgchen:
Addressed findings 1–7 in `fdc525a0a`:

1. Marked progress INFO now bypasses filtering only when `progress_logger_names` is configured. Default and explicit `concise` suppress the new banners/tables while retaining their NVFlare 2.9 output; the real example test asserts both sides.
2. Removed the catch-all `MetricsArtifactWriter` injection from `ServerJsonConfigurator`, along with the type/documentation changes and plain-FedJob integration case added to support it. Standard Recipe/BaseFedJob training keeps its existing writer; directly assembled FedJob/JSON and out-of-scope jobs keep their prior components.
3. Malformed or dangling `--log_config` now aborts import-time consumption transactionally, without raising `SystemExit`, mutating `sys.argv`, changing `FL_LOG_LEVEL`, or enabling export.
4. The primary structured progress request now applies the same 64 KiB server-side cap as the text fallback.
5. Any empty structured-log response tries the bounded text fallback, including older-server `{ "logs": {} }` responses, and then follows the existing one-time unavailable notice path.
6. Aggregation events without metrics, sites, or skipped values return before emitting the separator, aggregate row, or completion message; round display state is still reset.
7. Progress rendering ignores malformed normalized-metric entries without propagating an exception into event dispatch.

I did not change finding 8. The agreed file policy in the earlier review thread is that `log_fl.txt` follows the readable presentation in `progress` mode, while `log.txt` and `log.json` retain detailed INFO diagnostics. `log_fl.txt` is not WARNING-only: explicitly marked round and aggregation INFO records pass the progress filter, and the real integration test reads `server/log_fl.txt` and verifies both. The PR documentation states this behavior.

Validation: 211 targeted unit and real-output integration tests pass, including actual default/explicit concise and progress runs. `./runtest.sh -s --skip-install` passes. The PR description and LOC tables are updated for the reduced 49-file diff.

nvflare/app_common/widgets/metrics_artifact_writer.py
```diff
@@ -126,8 +189,28 @@ def _handle_after_aggregation(self, fl_ctx: FLContext):
         if not aggregated_metrics and not sites and not skipped:
             if custom_aggregator_metrics:
                 self._custom_aggregator_no_metric_rounds.append(current_round)
+            self._round_contribution_count = 0
+            self._progress_client_rows = 0
+            self._progress_rows_omitted = False
+            self._progress_columns = None
             return
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in bdc1d6d83. A metric-free aggregation now distinguishes an event with no round context from one that follows accepted contributions. With no contribution or aggregation details it emits nothing; with an accepted update it emits only the completion line, without a misleading separator or empty aggregate row. Regression tests cover both cases.

nvflare/recipe/_args.py
```diff
@@ -0,0 +1,128 @@
...
+            if i + 1 >= len(argv) or argv[i + 1].startswith("--") or not argv[i + 1].strip():
+                _CONSUMED = True
+                return False, DEFAULT_EXPORT_DIR
+            log_config = argv[i + 1]
+            i += 2
+        elif argv[i].startswith("--log_config="):
+            log_config = argv[i].split("=", 1)[1]
+            if not log_config.strip():
+                _CONSUMED = True
+                return False, DEFAULT_EXPORT_DIR
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in bdc1d6d83 without restoring import-time `SystemExit`. Import records the malformed shared argument transactionally and leaves `sys.argv` and `FL_LOG_LEVEL` unchanged; constructing the Recipe execution environment then raises a clear `ValueError` before deployment. This also rejects the option for scripts that do not parse their own arguments. Tests cover dangling, option-valued, and empty equals forms.

chesterxgchen:
Addressed the remaining compatibility finding and the deeper-pass findings in `d341532c1`.

- Restored the original `FLFilter` on progress-mode `log_fl.txt`. The file now retains ordinary `nvflare.app_common` / `nvflare.app_opt` INFO records as well as marked progress records, so existing readers such as the Auto-FL partial-aggregation detector continue to see `Aggregated N/M results`. The real simulator test verifies the actual `Aggregated 2/2 results` line in progress mode.
- Consolidated the saved-artifact notice to at most once per round instead of once per client/aggregate row when more than two metrics are present.
- Expanded the progress-table label column so `hospital-north` and `hospital-south` remain distinct while preserving the 80-column bound.
- Added an `ExcludeProgressFilter` to the baseline/full/verbose console configuration. The new marked records appear only in the opt-in progress presentation; existing detailed records remain unchanged.
- Removed exception and stack text only from the copied progress-console record. The original record still carries its traceback to `log.txt`, `log.json`, `error_log.txt`, and `log_fl.txt` where applicable.
- Classified current and legacy aborted statuses as `Aborted`, without a fabricated failure-details block.
- Made `--tail-bytes` reserve space for its truncation notice and discard a partial first record, so bounded JSON/text tails remain within the requested byte limit and begin on a record boundary.
- Deduplicated model artifact locations before rendering the final summary.

Validation:

- 249 targeted unit tests passed (3 skipped).
- All 3 real Recipe output integration cases passed.
- Both Auto-FL simulator watchdog regression tests passed.
- `./runtest.sh -s --skip-install` passed.

The PR description and LOC tables have been refreshed for the current 50-file diff.

nvflare/app_common/widgets/metrics_artifact_writer.py
```diff
@@ -123,9 +199,32 @@ def _handle_after_aggregation(self, fl_ctx: FLContext):
...
-        if not aggregated_metrics and not sites and not skipped:
+        has_aggregation_details = bool(aggregated_metrics or sites or skipped)
+        if not has_aggregation_details:
             if custom_aggregator_metrics:
                 self._custom_aggregator_no_metric_rounds.append(current_round)
+        else:
+            log_progress(_logger, "  " + "─" * 66)
+            self._log_progress_metrics("Aggregated", aggregated_metrics)
+            if self._progress_metrics_omitted:
+                log_progress(_logger, "  Additional metric results are available in the saved metrics artifacts.")
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in 70a86d2dd. The once-per-round saved-artifact notice now runs independently of aggregation details, so omitted client metrics remain disclosed even when the aggregator intentionally reports no metrics or sites. Added a regression with three accepted client metrics, use_contribution_sites=False, and an empty aggregation result; the notice and completion line are each emitted once.

chesterxgchen:
Fixed the blocking POC continuity-test mismatch in `59516e5ac`. The test asserts progress-only round, aggregation, evaluation, and live-status output, so it now explicitly sets `FL_LOG_LEVEL=progress`; implementation and documentation remain unchanged. The complete `hello_pt_environment_continuity_test.py` file passes (3 tests, 24 warnings), and `./runtest.sh -s --skip-install` passes.

chesterxgchen:
Addressed the latest blocking and worthwhile follow-up findings:

- `8dc6c2d36` implements Option A for aborted jobs. `■ Aborted` remains a distinct terminal label, while both local failure summarization and remote client-error collection treat aborted runs as eligible for error context. The older failure-summary test also required a label adjustment because it hard-coded `✗ Failed`; it now verifies `■ Aborted` together with the original checkpoint error.
- `660260b6c` records malformed/dangling/empty `--export-dir` as a deferred Recipe argument error. Parsing remains transactional: `sys.argv`, export state, and `FL_LOG_LEVEL` are unchanged, and execution setup raises a clear error instead of silently falling back to concise mode.
- `660260b6c` suppresses `RUN SUMMARY` and the missing-result message when `get_result(timeout=...)` returns while the job status is still nonterminal. Terminal completed, aborted, failed, and not-scheduled summaries are unchanged.

Validation:

- 149 aborted-status, failure-summary, remote-error, status, and run tests passed (3 skipped).
- 171 Recipe argument/run/session/failure/status tests passed (3 skipped) after the two follow-up fixes.
- The focused argument and run suites passed again after formatting (71 passed, 3 skipped).
- `./runtest.sh -s --skip-install` passed.

The six remaining low/latent observations were intentionally left out of this PR. The PR description and LOC tables are updated for the current head.

nvflare/recipe/run.py
```diff
@@ -110,13 +117,52 @@ def get_result(self, timeout: float = 0.0, clean_up: bool = True) -> Optional[st
                 self.logger.warning(f"Failed to get job status: {e}")
                 self._cached_status = None
 
+            report = ""
+            failure_report = ""
+            outcome = job_status_outcome(self._cached_status)
+            if outcome in ("failed", "aborted"):
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in 0f2c871a2. The shared no-detail fallback now says “Check the job’s server and client logs,” which is accurate for both failed and aborted runs. The regression test asserts the neutral guidance and rejects the old “failed job” wording. The focused failure-summary/run suites pass (120 passed, 3 skipped), and the style check passes.

chesterxgchen:
Addressed the remaining old-server bounded-log compatibility gap in `0623b141b`.

- `Session.get_job_logs()` now recognizes an old server's `unrecognized arguments: --tail-bytes` response for the default `log.txt` selector as well as non-default selectors such as `log.json` and `error_log.txt`.
- The client returns an empty log response instead of raising `InternalError`, allowing Recipe progress monitoring to use its existing fallback/unavailable-log path rather than emitting the generic retrieval failure.
- The request is not retried without `--tail-bytes`, preserving the bounded-transfer guarantee and avoiding repeated full-log downloads from an old server.
- Regression coverage verifies both default `log.txt` and non-default `error_log.txt` requests and confirms each command is attempted only once.

Validation:

- The Session log-API and Recipe session-manager suites pass: 99 tests.
- `./runtest.sh -s --skip-install` passes.

The earlier ABORTED findings were against stale head `59516e5ac`; Option A remains present at the current head through `8dc6c2d36`. The PR description and LOC tables are updated for the current diff.

chesterxgchen:
Reviewed the production LOC and simplified the argument/status implementation in `33be4a337`.

- Replaced the hand-written shared-argument token scanner with `argparse.ArgumentParser(..., exit_on_error=False)`. It preserves import-time consumption, strict option names, `--` handling, transactional failures, both `--option value` and `--option=value`, and deferred errors without import-time `SystemExit`.
- Reduced `nvflare/recipe/_args.py` from 136 to 88 lines.
- The new file is primarily relocated code: paired with the `-86` net lines in `nvflare/recipe/spec.py`, shared argument handling is now only **+2 net production lines**.
- Removed redundant `FINISHED_OK` and `FINISHED_EXCEPTION` entries from the legacy terminal-status set because the existing `FINISHED*` check already covers them.
- The total production diff drops from **+979 net lines to +929**.

I also audited the remaining high-LOC production areas. Their code corresponds to separate required boundaries rather than duplicated argument machinery: bounded success-artifact parsing, bounded failure/traceback extraction, remote progress replay, reusable metric formatting/filtering, and round-progress emission. Further line-count reductions there would combine independent responsibilities or remove reviewed compatibility and bounded-output behavior.

Validation:

- 182 argument, Recipe, simulator, run, session, and status tests passed (3 skipped).
- All 3 real Recipe-output integration tests passed.
- `./runtest.sh -s --skip-install` passed.

The PR description now presents `_args.py` together with the `spec.py` removal and includes the updated exact LOC totals.

chesterxgchen:
Audited the full production diff for scope and simplification, then pushed `e693953ac`.

Removed:

- per-model/per-client cross-site evaluation log lines, which duplicated the bounded final evaluation summary and could grow as clients × models;
- the dedicated scalar-summary formatter and its tests, since nothing else used it;
- ordinary metrics-artifact path INFO records, since the final summary already reports those paths and the records could affect default concise output;
- SHA-256 hashing for a replay window already bounded to 200 lines / 64 KiB;
- repeated progress-reset state and a separate row-omission flag.

This removes 40 net production lines and 111 net test lines from the previous head. The production diff is now +1,046 / -157, net +889. Its net LOC breaks down as:

| Area | Net LOC |
|---|---:|
| Run and failure summaries | +431 |
| Progress log mode and metric presentation | +276 |
| POC/production monitoring and bounded log retrieval | +133 |
| Shared status classification | +36 |
| Recipe argument and environment handling | +13 |

The two largest standalone files are the bounded run-summary reader (+177) and failure-summary/log reader (+205). They implement the requested terminal and failure summaries across simulator, POC, and production; they do not change training or FL control flow. The remaining support is the progress formatter/emitter and the bounded remote-log path needed to present the same information outside simulation.

Validation after simplification:

- 166 targeted tests passed, 3 skipped;
- 3 Recipe output integration tests passed;
- 3 Hello PyTorch simulation/POC continuity tests passed;
- `./runtest.sh -s --skip-install` passed.

The PR description and LOC tables have been updated with the reduced counts and this breakdown.

chesterxgchen:
@pcnudde I went back through the current PR specifically for the scope/complexity concern in your review.

The three unrelated Hello PyTorch data/cache changes you identified are removed entirely. The `max_bytes` log API remains because POC/production failure summaries need the server to bound client/server log tails before transfer; the PR description now calls out that boundary directly.

I also simplified the remaining implementation in two follow-up passes:

- replaced the 136-line hand-written Recipe argument scanner with an 88-line `argparse` implementation; paired with the removal from `spec.py`, argument handling is +2 net lines;
- removed duplicate per-result cross-site evaluation output, its dedicated formatter/tests, duplicate artifact-path INFO messages, hashing of an already bounded replay window, and repeated progress state.

The production diff is now +1,046 / -157, net +889. Of that, +431 is the requested bounded run/failure summary, +276 is the opt-in progress formatter and round metrics, +133 is bounded POC/production monitoring and log retrieval, +36 is shared status classification, and +13 is Recipe argument/environment integration.

The current head is `e693953ac`; all inline threads are resolved, the focused and integration suites pass, and the PR description includes the updated LOC and file-level breakdown. Could you please re-review the current head?

chesterxgchen:
Fixed the partial `log.txt` replay finding in `4535f5dbf`.

The local 64 KiB guard now detects truncation and drops the first partial byte record before UTF-8 decoding, line limiting, or deduplication. This applies to structured and text fallback input, matching the complete-record behavior used by the other bounded-tail readers.

The regression exercises two growing fallback snapshots whose first long line is cut at different offsets. It verifies that no partial fragment is printed, an existing complete line is not repeated, and a newly appended line is printed once. The complete `session_mgr_test.py` suite passes (23 tests), and `./runtest.sh -s --skip-install` passes.

chesterxgchen:
Reviewed the six optional follow-ups against `4535f5dbf`; no additional code change is warranted:

1. `failure-logs-*` is intentionally retained inside the downloaded result. It contains the bounded client error streams referenced by the failure summary so the user can inspect them afterward; deleting it after presentation would remove those diagnostics. The directory is removed with the result workspace when the caller chooses cleanup.
2. `_read_json()` uses a one-byte oversize probe. A named constant would be cosmetic and add code without changing the bound or behavior.
3. Progress replay intentionally retains only the prior bounded window. This caps state at 200 lines / 64 KiB and allows an identical message that genuinely recurs after leaving the window to be shown again. Retaining every historical line would make memory grow for the duration of a job.
4. `self.clients = v.clients or None` is confined to `SimEnv`. An empty list with a positive resolved `num_clients` must select the simulator's count-based `n_clients` path; the SimEnv regression asserts this. `PocEnv` has separate validation and preserves its own `clients` value.
5. The terminal-status observation is stale: `_LEGACY_TERMINAL_JOB_STATUSES` now contains only `ABORTED`, `ABANDONED`, and `FAILED`; `FINISHED_OK` and `FINISHED_EXCEPTION` were removed because `startswith("FINISHED")` already covers them.
6. Progress records intentionally remain in file handlers in every mode. `ExcludeProgressFilter` protects the existing non-progress console views, while `log.txt`, `log.json`, and the existing application-selected `log_fl.txt` retain emitted records for diagnostics and automation. The `FLFileHandler` filter/formatter behavior itself is unchanged.

These are documented design choices or already-resolved cleanup, so addressing them would either remove useful diagnostics, weaken bounded state, or add cosmetic churn.

chesterxgchen:
Rebased onto current `upstream/main` (`3d2b05288`) and force-pushed the branch with lease protection. New head: `72ee4cd88`.

The rebase completed without merge conflicts, and the PR diff/LOC totals are unchanged.

Post-rebase validation:

- 167 targeted reporting, logging, status, Recipe, and session tests passed; 3 optional-dependency cases skipped;
- all 3 real Recipe output integration cases passed;
- all 3 Hello PyTorch simulation/POC continuity cases passed;
- `./runtest.sh -s --skip-install` passed.

holgerroth (approved):
Approving reviewed head `72ee4cd88`. The phased rollout is preserved: existing `concise` remains the default, and the new `progress` presentation is opt-in. Detailed logs remain available; textual severity labels, bounded remote-log fallback and unavailable-log notices, and remote-access documentation address my earlier findings.

Validation of the latest substantive changes: 188 targeted tests passed (3 skipped), plus all 3 real simulator logging cases. I verified that the subsequent rebase changed none of the 49 PR file contents. All current CI checks, including Blossom, now pass. No blocking regressions found in the reviewed changes.

pcnudde (approved):

PR #5290, opened 2026-09-14
Block client props in Dashboard creation and PATCH requests

IsaacYangSLA:
### Description

Dashboard client properties can supply provisioning paths that cause server-local files to be included in downloadable kits. Reject client creation and PATCH requests containing `props` with HTTP 400 before changing any stored data, for both client creators and project admins, regardless of approval state or the supplied value.

Client-kit generation also ignores previously stored `Client.props`, including malformed JSON, so legacy records cannot continue injecting provisioning properties after the API restriction takes effect. Client capacity and operator-managed properties from `properties.yml` or `properties.json` remain supported. Existing rows do not require migration.

API integrations that submit `props` must omit it. The current Dashboard UI's client name and capacity requests continue to work.

### Types of changes

- [x] Breaking change: client API requests containing `props` are now rejected.
- [x] New tests cover creation and PATCH, creator and admin access, pending and approved clients, empty/null/invalid values, and rejection without partial updates.
- [x] Kit-generation regression tests seed legacy properties directly in the database and download the actual ZIP as the client creator. They verify that the target file is excluded, stored connection security is ignored, and client capacity and operator-managed transport properties still reach the kit.

### Validation

- `./.venv/bin/python -m pytest tests/unit_test/dashboard -q --disable-warnings`: 73 passed; existing SQLAlchemy and JWT warnings remain.
- Before the provisioning-boundary fix, the new legacy-path regression failed because the downloaded kit contained `customRootCA.pem`; both legacy-path and malformed-JSON cases pass after the fix.
- Scoped equivalent of the project style check passed: Black on each changed file, isort, flake8, agent skill checks, and `git diff --check`. Black's combined file check stalled, so formatting was checked per file.


pcnudde:
Blocking security concern: this change only prevents future API writes. Client.props values stored before the upgrade remain active. _prepare_client in nvflare/dashboard/application/blob.py still deserializes those values and passes them to provisioning, and StaticFileBuilder copies a custom_ca_cert path from those properties into the downloadable client kit. An authenticated client creator who stored such a path before rollout can therefore request the blob after rollout, once the client is approved, and disclose any file readable by the Dashboard process.

Please neutralize legacy values at the provisioning boundary, preferably by not merging Client.props into client-kit properties, or migrate/sanitize existing rows or reject blob generation when legacy props are present. A regression test should seed a client with legacy props and prove that blob generation ignores or safely rejects them.

IsaacYangSLA:
@pcnudde Addressed in 298d6e85e54eb74a100fd94091860b8264a37df8. `_prepare_client` now ignores stored `Client.props` entirely; only client capacity and operator-managed configuration are supplied to provisioning. This covers existing rows without requiring a migration.

Added regression tests that seed an approved client's legacy properties directly in the database and download the actual kit as its creator. Before this fix, the legacy-path test failed because `customRootCA.pem` appeared in the ZIP. After the fix, both the legacy-path and malformed-JSON cases pass, with no target file disclosed; GPU capacity and operator-managed transport configuration remain effective. All 73 Dashboard tests and scoped style checks passed.

pcnudde (approved):
Re-reviewed at 298d6e85e54eb74a100fd94091860b8264a37df8. The legacy Client.props concern is addressed at the provisioning boundary, the new regression exercises the actual downloadable kit, all 73 Dashboard tests pass locally, and the full CI suite is green.

PR #5295, opened 2026-09-14
Prevent failed FedAvg callbacks from publishing partial aggregates

### LINKED ISSUE #5293: FedAvg saves partial weights from a rejected contribution
With the supported, opt-in `enable_tensor_disk_offload=True` FedAvg workflow, a later tensor-file read can fail after earlier tensors from that client have entered the accumulator. External scratch cleanup or permission changes are plausible triggers; no spontaneous production incident was demonstrated.

**Actual:** FedAvg rejects the contribution but saves its partial weights: a healthy contribution `(a=2,b=4)` followed by a rejected contribution `(a=100,b=failing)` saves `(51,4)` instead of `(2,4)`. A subsequent DIFF round retains the error.

**Expected:** Failed aggregation must not alter the saved global model.

**Reproduction:** Drive a FedAvg round with a healthy result followed by a two-key result whose second real lazy tensor file is inaccessible. Controlled file deletion and permission denial reproduced corruption using real tensor components and PyTorch checkpoint persistence; transport was substituted. Tested SHA: `83ec31a8e14bc41f788ec5a739327298173632b5`.

The [public Specula report and reproduction](https://github.com/specula-org/specula-case-studies/blob/main/systems/nvflare/modules/fedavg-task-lifecycle/runs/nvflare-fedavg-20260913/confirmed-bugs.md) document the same failure under a controlled lazy-file fault at an earlier revision. This is a conditional reliability defect, not a claim of ordinary network truncation or measured production frequency.

**Proposed fix:** Invalidate the round on aggregation exceptions and synchronize callback completion with finalization, preventing update/persistence while preserving the previous checkpoint and existing memory bounds. Discard successful contributions from that failed round. This fix is proposed, not yet implemented.

pcnudde:
A built-in FedAvg callback can accumulate early tensors and then fail loading a later tensor. The contribution is rejected, but its partial values can still reach the saved model.

Fail the affected round before model update or persistence, preserving the previous checkpoint. A per-round lock synchronizes callback completion with finalization, and closed callbacks cannot modify a later round. Successful contributions from the failed round are discarded. Tensor-at-a-time loading and memory bounds are unchanged; custom aggregators are outside this change.

Fixes #5293

Validation: 158 passed, 2 skipped across the FedAvg failure, FedAvg workflow, and weighted aggregation helper tests. Before the fix, the new suite had 6 failures and 2 passing controls. `./runtest.sh -s` passed; notebook formatting was skipped because optional Jupyter formatting dependencies were absent. Final staged whitespace checks passed.

Evidence uses controlled file/metric failures and deterministic callback timing, including real checkpoint persistence. No spontaneous production incident or deployed transport test is claimed. The full unit suite was not run.


nvflare/app_common/workflows/fedavg.py
```diff
@@ -211,13 +212,26 @@ def run(self) -> None:
                 self._params_type = None
                 self._site_metric_weights = {}
 
+                # Keep a completed/failed round from publishing partial callback mutations.
+                round_state = {"lock": Lock(), "failed": False, "closed": False}
+
+                def aggregate_one_result(result, state=round_state):
+                    with state["lock"]:
+                        if state["closed"] or state["failed"]:
```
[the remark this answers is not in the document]
pcnudde: Addressed in bda8f5523. The abort branch now acquires the round lock and closes the round before returning, so an active callback finishes first and subsequent callbacks are rejected. Extended the deterministic callback/finalization test to cover abort; it fails on the previous commit and passes with this change. Focused tests: 159 passed, 2 skipped; project style checks passed. No model copies or custom-aggregator behavior changes.

nvflare/app_common/workflows/fedavg.py
```diff
@@ -211,22 +212,43 @@ def run(self) -> None:
...
                 self.send_model(
                     task_name=self.task_name,
                     targets=clients,
                     data=model,
-                    callback=self._aggregate_one_result,
+                    callback=aggregate_one_result if not self.aggregator else self._aggregate_one_result,
                 )
 
                 # Wait for all results to be processed
                 while self.get_num_standing_tasks():
```
YuanTingHsieh, line 238: **[P1] Surface callback failure before waiting for task retirement**

`aggregate_one_result()` records a failure in `round_state["failed"]`, but this loop does not inspect that state. `BaseModelController._process_result()` catches the raised callback exception, so the task itself is not forced to retire. Since FedAvg defaults to waiting for all targets with `timeout=0`, one failed aggregation callback plus one connected client that remains slow or silent leaves the workflow waiting indefinitely; the `RuntimeError` below is never reached. I reproduced this against this head: after one callback failed while the standing-task count remained positive, the run thread stayed alive until an explicit abort. Please make callback failure wake the controller and retire/cancel the current task, and add a regression for this combination. When implementing that path, avoid holding the round lock while performing task-lifecycle operations; the callback currently holds it across lazy tensor materialization, so abort/finalization can already wait behind slow or stuck I/O. This is an indefinite-wait risk rather than a demonstrated cyclic deadlock.

nvflare/app_common/workflows/fedavg.py
```diff
@@ -211,22 +212,43 @@ def run(self) -> None:
...
+                            raise
+
                 # Non-blocking send with callback for streaming aggregation
                 set_fedprox_metadata(model, self.fedprox_mu)
                 self.send_model(
                     task_name=self.task_name,
                     targets=clients,
                     data=model,
-                    callback=self._aggregate_one_result,
+                    callback=aggregate_one_result if not self.aggregator else self._aggregate_one_result,
```
YuanTingHsieh, line 234: **[P1] Preserve failed-round protection for custom aggregators**

This conditional bypasses `aggregate_one_result` whenever a custom aggregator is configured. If `accept_model()` partially mutates its internal state and then raises, `BaseModelController._process_result()` catches the exception, `round_state["failed"]` remains false, and `aggregate_model()` can subsequently publish and save that partial state—the same corruption pattern this PR fixes for the built-in aggregator.

The current `ModelAggregator.accept_model()` contract does not require atomic mutation or rollback. Please apply failed-round tracking to custom aggregators as well, or formally add and enforce an atomicity requirement in the interface. A regression should use a custom aggregator that mutates and then raises, and verify that model update and persistence are skipped. This failure tracking is required even if custom aggregators remain responsible for their own internal locking.

nvflare/app_common/workflows/fedavg.py
pcnudde: Fixed in f217bb8d0. The wait loop now checks the recorded callback failure, marks outstanding tasks ERROR outside the round lock, and breaks into the existing synchronized close-and-raise path before any model update or persistence. Added regressions with a permanently positive standing-task count, both with and without an earlier healthy contribution; they fail on the prior head and now pass, checking cancellation and unchanged checkpoint bytes. Focused tests: 161 passed, 2 skipped; project style checks passed. Cancellation marks server tasks completed; it does not itself synchronously stop remote training.

nvflare/app_common/workflows/fedavg.py
pcnudde: The partial-mutation risk is valid, and the current interface does not require exception-atomic accept_model(). We are keeping this PR limited to built-in FedAvg, as stated in its description. Applying the wrapper to custom aggregators would impose a new fail-the-whole-round policy on them; that extension is deferred rather than claiming their state is already protected. Resolving this thread as an explicit scope decision, not as a fix for custom aggregators.

YuanTingHsieh (approved):

PR #5297, opened 2026-09-14
Fix multi-target fire-and-forget payload lifetime

### LINKED ISSUE #5291: [Bug] Multi-target fire-and-forget retires downloadable payload after the first receiver
A custom P2P workflow can send a NumPy snapshot to two peers using `send_aux_request(..., timeout=0)`. If one peer completes while the other still needs source data, `Cell._fire_and_forget` permits premature source retirement: encoding has omitted the expected receiver count, which defaults to one.

**Actual:** the slower peer gets `invalid_request`; eager decoding fails before its application handler runs. The payload receipt reports completion for only one receiver. **Expected:** retain the source until all intended receivers settle, and certify success only when all succeed.

Reproduced on `83ec31a8e14bc41f788ec5a739327298173632b5` with production NumPy serialization, an 8 MiB array, separate Cell processes and loopback TCP. Receiver scheduling was controlled so A completed before B downloaded. Single-target and explicit two-receiver declaration controls succeeded. This is not evidence of a spontaneous production incident or an affected default workflow: the shipped P2P greeting example uses one-way sending but no downloadable array, while distributed optimization uses a positive timeout.

**Proposed fix:** declare distinct target count and identities before encoding; preserve count-only accounting for explicit pass-through and existing timeout handling for missing recipients. The fix is currently proposed, not applied or tested. Suggested priority: Medium.

Public mechanism and executable baseline: [Specula CR-2 investigation](https://github.com/specula-org/specula-case-studies/blob/main/systems/nvflare/modules/payload-transfer/runs/nvflare-transfer-20260913/confirmation/CR-2/investigation.md) and [reproduction](https://github.com/specula-org/specula-case-studies/blob/main/systems/nvflare/modules/payload-transfer/runs/nvflare-transfer-20260913/repro/test_bugCR-2_fire_forget_fanout.py). The public baseline substitutes in-process transport; the separate-process TCP corroboration described above uses controlled timing too.

pcnudde:
Fixes #5291.

A multi-target streamed fire-and-forget send could retire its downloadable payload after the first receiver finished, leaving later recipients unable to download it. Declare the distinct receiver count before encoding and avoid creating a source for an empty target list.

Keep completion count-based: receivers can enable pass-through locally, so routing targets are not necessarily the final consumers. Missing recipients retain the source until existing timeout handling settles them; this change does not add retries or exact recipient-identity verification.

Validation:
- 119 focused Cell/downloader tests passed, including delayed/missing receivers and actual Adapter channel/topic forwarding.
- Separate-process TCP with an 8 MiB NumPy payload and controlled receiver scheduling passed; this is not a claim of a spontaneous production incident.
- `./runtest.sh -s` passed for Python/style targets; the tool skipped notebook formatting because Jupyter extras were unavailable. No notebooks changed.
- Full repository tests and TLS deployment validation were not run.


YuanTingHsieh:
The `(["a", "a", "b"], ...)` case exposes a target-normalization gap shared by both streaming broadcast paths. Receiver completion is tracked by distinct identity, but dispatch still follows every entry in the original list. In this test, three messages are sent while only the two distinct consumers are decoded; if the second delivery to `a` arrives after `a` and `b` complete, the source has already been released and that delivery fails.

The blocking `_broadcast_request` path updated in #5075 has the same underlying assumption: it also iterates the original target list. Could we normalize once, preserving order, and use the normalized list for both receiver accounting and dispatch in `_broadcast_request` and `_fire_and_forget`, for example `targets = list(dict.fromkeys(targets))`? Alternatively, if duplicate targets are invalid input, we should validate that explicitly and remove this test case. This does not affect the original unique-target `["a", "b"]` fix, but the current duplicate-target test overstates what is exercised.

I am fine leaving it to the author whether to address the shared normalization in this PR or track it in a follow-up PR; it should not block the primary unique-target lifetime fix here.

pcnudde:
Agreed—the duplicate-target case checks distinct-consumer accounting and preservation of the existing dispatch behavior; it does not demonstrate safe decoding of every duplicate delivery. The test decodes each distinct consumer once and separately asserts that dispatch still follows the original list.

I’m keeping shared dispatch normalization out of this PR, as you suggested. Normalizing both `_broadcast_request` and `_fire_and_forget` consistently is suitable follow-up work, with coverage that exercises every dispatched delivery. This PR remains focused on premature retirement for unique targets and preserves receiver-configured forwarding through count-only accounting.

YuanTingHsieh (approved):

PR #5302, opened 2026-09-15
Add certificate-derived study memberships

### LINKED ISSUE #5250: Study entitlements in administrator certificates
Parent overview: #5246
Implementation: #5302

## Goal

Allow SSO-issued administrator certificates to carry signed study entitlements, reducing per-study user-list maintenance while preserving registry control and existing static administrators.

## Agreed design

- The issuer derives explicit study names from administrator-controlled, signed and validated IdP claims, never from user-supplied CLI template data or CSR fields.
- Carry one standard URI Subject Alternative Name (SAN) per study:
  `https://nvidia.com/nvflare/v1/project/<percent-encoded-label>/study/<study>`.
- The project label is an informational, UTF-8 percent-encoded path segment, not an authorization boundary. It is not compared with the running project name; matching studies can be authorized in any federation trusting the issuing CA.
- After normal certificate-chain, validity, identity and administrator-purpose checks, NVFlare parses and validates the study claims.
- Authorization is additive: a named study must exist in the registry, and either registry membership or a certificate claim must authorize the administrator.
- The certificate's existing identity, organization and certificate-wide role remain unchanged. Study membership does not override the role.
- Retain certificate memberships in the server-signed session token so discovery survives session reconstruction.

## Requirements

- Accept at most 64 unique study names. Names use lowercase alphanumeric characters, hyphens or underscores, are 1–63 characters long, and start and end with an alphanumeric character.
- Reject duplicate names, reserved `default`, wildcards, and malformed or unsupported NVFlare study claims. Invalid claims reject the entire certificate login, including default-study login; registry membership must not hide an invalid claim.
- Ignore unrelated URI namespaces. A certificate without study claims grants no additional membership and preserves existing registry/default behavior.
- Keep the registry authoritative for study existence, site enrollment, scheduling and static administrator grants. Claims cannot create studies, enroll sites, alter scheduling or remove registry grants.
- Preserve job/CA/cell certificate restrictions when study SANs coexist with other NVFlare URI SANs.
- Keep entitlement authorization independent from certificate acquisition and renewal.
- Removing IdP entitlements or registry membership does not revoke existing certificate-derived access; it lasts until certificate expiry. Study names must not be reused for different studies.
- Provide vendor-neutral published documentation and one downloadable step-ca template using validated OIDC ID-token claims.

## Explicit scope decisions

- Use the versioned URI namespace, not a custom enterprise OID or a separate JSON extension.
- Bound the study count and identifier length as above; no separate aggregate entitlement byte-size cap is required.
- Do not add authorization-source audit events, an issuer-side study allowlist, a local denylist or study-name-reuse enforcement.
- Membership is visible in the certificate and signed session token; this is a shared-trust deployment model.

These decisions supersede the initial custom-OID, authorization-source-audit and aggregate-payload-size requirements.

## Compatibility boundary

Study SANs are signed authorization evidence, not a trust anchor. Registry-only/static administrators and SSO administrators with or without entitlements remain supported on updated servers. Older servers with strict NVFlare URI parsing may reject study-bearing certificates, so servers must be upgraded before these SANs are issued. Existing certificates without study claims retain their behavior.

pcnudde:
Closes #5250.

## Summary

Allow administrators to access named studies through either registry membership or membership carried in a validated certificate. The registry remains authoritative for study existence and site enrollment, and the certificate-wide role controls command permissions.

- Encode one study membership per URI SAN: `https://nvidia.com/nvflare/v1/project/<label>/study/<study>`.
- Accept up to 64 unique study names, each 1–63 characters long.
- Apply certificate membership consistently to login, study discovery, and authorized study-management commands.
- Preserve memberships in the server-signed session token so they survive session reconstruction.
- Share study-URI syntax validation across admin, site, job, and CA certificate readers. Malformed claims reject authentication, including default-study login.
- Provide published RST guidance and a downloadable step-ca template that maps issuer-validated OIDC claims to certificate identity, role, and study memberships.

## Compatibility and lifecycle

Existing registry-based certificates and default-study access remain supported. Issued memberships remain effective until certificate expiry. Deployments should upgrade servers before issuing certificates with study URI SANs.

## Validation

- 411 tests passed, covering real-certificate login, session reconstruction, study commands, role and organization restrictions, malformed URI rejection, and three CPU-only local live-mTLS tests.
- Full `./runtest.sh -s` and `git diff --check` passed.
- Verified published documentation links and the downloadable template.

End-to-end validation for this revision covers local mTLS; live IdP/step-ca/Kubernetes integration requires a separate run.


YuanTingHsieh (approved):
Do we intend for study_registry.json and certificate entitlements to remain two long-term sources of admin study membership, or is registry-managed admin membership primarily a backward-compatibility/migration mechanism? Since effective membership is their union, operators need to consider both sources when granting or revoking access.

pcnudde:
Yes, for now we need to support both in parallel. The SSO feature allows mixing of some people using OIDC while others use the traditional method. This means we need also the study membership to allow both paths.

PR #5313, opened 2026-09-17
Fail with the real cause when a tensor arrives as LazyDownloadRef

rfilgueiras:
chunk_tensors_from_params handles torch.Tensor, np.ndarray and dict. Anything else is skipped silently, so an unresolved LazyDownloadRef yields an empty blob: the client logs nothing and the server aborts five seconds later with No tensors received from peer, a timeout that names neither the cause nor the place.

This PR rejects that one type with an explicit error. It does not change the underlying behaviour.

## Why

On the ClientAPILauncherExecutor path, client/ex_process/api.py sets pass_through_on_send unconditionally for a CellPipe, so the client job holds references rather than materialised tensors. TensorClientStreamer then finds nothing to stream.

No configuration avoids this: ViaDownloaderDecomposer.decompose() checks for LazyDownloadRef before reading the download chunk size, so setting tensor_download_chunk_size and np_download_chunk_size to 0 on both server and client still leaves the refs in place.

The result is a silent failure on the client and a fatal, misleading one on the server. Diagnosing it took instrumenting the function to print value types. That is the cost this PR removes.

## What to look at

The error is raised inside a generator consumed by TensorProducer, so it surfaces as `Producer TensorProducer failed to produce next object: TypeError: ...` and stops the client runner with the real reason attached.

## Left alone on purpose

Only LazyDownloadRef is rejected. Other unhandled types are still skipped, because params legitimately carry non-tensor entries and a blanket raise would break callers that pass mixed data.

The incompatibility itself is untouched. Gating pass_through_on_send on whether the downloader is enabled, or resolving refs before AFTER_TASK_RESULT_FILTER, is a larger change and a separate discussion.

## Testing

Verified two ways. A direct unit-level call with a constructed LazyDownloadRef raises the new TypeError and confirms the import introduces no cycle. Then a real 4-round job on a POC deployment with an external training subprocess and a 256 MB payload: previously the server aborted on a 5 s timeout with no client-side error, and now the client fails immediately at the producer with the message above, and the server timeout no longer appears.


rfilgueiras:
Tensor Stream will be retired.

PR #5318, opened 2026-09-18
Prevent netrc from overriding example API tokens

### LINKED ISSUE #5317: examples get: .netrc credentials override GITHUB_TOKEN
Follow-up to merged PR #5316.

`nvflare examples get` supplies its selected `GITHUB_TOKEN` / `GH_TOKEN` through an `Authorization` header. With the default Requests session settings, an `api.github.com` entry (or a matching `default` entry) in `.netrc` overrides that header with HTTP Basic authentication. This silently ignores the explicitly configured GitHub token.

The affected call is in [examples_cli.py at the merge commit](https://github.com/NVIDIA/NVFlare/blob/4de11978dfd608b128122559c3dd6e26f76a0fef/nvflare/tool/examples/examples_cli.py#L214-L215). Requests [documents this precedence](https://requests.readthedocs.io/en/latest/user/authentication/#netrc-authentication).

### Impact

Users with stale or different matching `.netrc` credentials can receive authentication failures or use another account's rate-limit allowance despite configuring a valid GitHub token. This is a normal-priority follow-up affecting that credential configuration.

### Offline reproduction

Run from an NVFlare checkout containing #5316. This uses synthetic credentials, makes no network requests, and leaves existing credentials unchanged.

```python
import os
import tempfile
from pathlib import Path
from unittest.mock import patch

import requests

from nvflare.tool.examples.examples_cli import _github_api_headers

with tempfile.TemporaryDirectory() as directory:
    netrc = Path(directory) / "netrc"
    netrc.write_text("machine api.github.com login old-user password old-token\n")
    netrc.chmod(0o600)
    with patch.dict(os.environ, {
        "NETRC": str(netrc),
        "GITHUB_TOKEN": "ghp_FAKE_TOKEN_FOR_REPRODUCTION_ONLY",
    }):
        request = requests.Request(
            "GET",
            "https://api.github.com/repos/NVIDIA/NVFlare/git/trees/HEAD",
            headers=_github_api_headers(),
        )
        with requests.Session() as session:
            prepared = session.prepare_request(request)
        print(prepared.headers["Authorization"].split()[0])
```

Expected: `Bearer`. Actual: `Basic`.

This was also reproduced through `_download_example()` while intercepting `Session.send`, so the actual downloader's prepared API request exhibits the same behavior.

### Proposed fix and validation

Apply the selected bearer token with a request-scoped authentication handler that takes precedence over `.netrc`. Preserve environment proxy and CA configuration, and keep the token off raw public-file requests.

Add a regression test that inspects the prepared API request with a conflicting temporary `.netrc` entry. The existing authentication test mocks `Session.get`, bypassing the request preparation where the header is replaced.

chesterxgchen:
An `api.github.com` entry in `.netrc` could replace the bearer header selected from `GITHUB_TOKEN` or `GH_TOKEN` with HTTP Basic authentication during Requests preparation. Users could therefore receive authentication failures or consume a different account's rate-limit allowance despite explicitly configuring a token.

The original code supplied the token as an ordinary header:

```python
session.get(tree_url, headers={"Authorization": f"Bearer {token}"})
```

Because the request had no `auth=` value, Requests loaded matching `.netrc` credentials during request preparation and `prepare_auth()` replaced that header with Basic authentication.

The fix has two parts:

1. `_github_api_auth()` returns a request-scoped `AuthBase` implementation, and the tree request passes it through `auth=`. Requests therefore sees explicit authentication and skips the `.netrc` lookup; the handler writes the Bearer header during request preparation. This preserves `HTTP_PROXY`, `HTTPS_PROXY`, and `REQUESTS_CA_BUNDLE`, unlike setting `session.trust_env = False`.
2. The fixed, revision-pinned tree request sets `allow_redirects=False`. This prevents Requests from calling `rebuild_auth()` on a redirect and reapplying `.netrc` credentials after the initial request. A 3xx response returns `EXAMPLE_NETWORK_ERROR` before creating the destination.

Raw public-file downloads do not receive the auth handler or configured token.

Regression coverage prepares a request with a conflicting temporary `.netrc` entry and verifies that the bearer token wins. It also verifies that tree redirects are disabled and rejected without a second request.

Validation:

- 234 focused examples, CLI output, and CLI schema tests passed.
- Full project style checks passed: Black, isort, flake8, and agent-skill lint.
- `git diff --check` passed.

Closes #5317.


chesterxgchen:
Addressed the redirect-time `.netrc` finding in f0d7e1159. The fixed, revision-pinned GitHub tree request now sets `allow_redirects=False`; any 3xx response returns `EXAMPLE_NETWORK_ERROR` before destination creation, so Requests cannot run `rebuild_auth` and replace the bearer token during redirect handling. Regression coverage verifies redirects are disabled, only one request is attempted, and no destination is created. Validation: 234 focused tests and the full project style check pass.

holgerroth (approved):
Reviewed f0d7e115946fade2f52d7b0f369c0209d6ab6736; no actionable findings.

Validation: 233 focused tests passed, with one catalog coverage test skipped because the isolated source archive lacked Git metadata. Additional offline checks using Requests' real preparation path verified bearer-token precedence over conflicting .netrc credentials, isolation from raw-file downloads, preservation of environment proxy and CA settings, and rejection of 301/302/303/307/308 metadata redirects without a second request. All reported CI checks passed. Live GitHub downloads were not exercised.

PR #547, opened 2022-05-16
Enable Overseer shutdown system command.

yhwen:

PR #5319, opened 2026-09-18
Extend progress reporting across workflows

chesterxgchen:
## Problem

The opt-in `progress` log mode gives FedAvg jobs a focused view of rounds, metrics, evaluation, and the final outcome. Other common data-science workflows still show little or no useful progress. Federated statistics, PSI, Kaplan–Meier survival analysis, Swarm learning, cyclic learning, and XGBoost can appear idle until they exit, even though they may run for a long time. Traditional-ML and evaluation recipes use the shared model workflow but several of their controllers did not publish the complete standard event sequence, so their output could miss round headings or completion context.

This makes the same CLI mode behave differently depending on the recipe. A data scientist can see that a deep-learning FedAvg job is advancing, while another supported recipe may provide no indication of its current round or protocol phase.

## Change

- Reuse the standard `ROUND_STARTED`, `AFTER_CONTRIBUTION_ACCEPT`, `AFTER_AGGREGATION`, and `ROUND_DONE` path for Cyclic, SCAFFOLD, Newton-Raphson logistic regression, FedEval, and Swarm. Successful rounds emit a balanced lifecycle; timed-out cyclic rounds do not publish `ROUND_DONE`. The component that owns the work emits the event, and `MetricsArtifactWriter` owns the bounded presentation. Existing Cyclic jobs with custom event subscribers now receive this standard lifecycle.
- Recheck cancellation at blocking workflow boundaries. FedEval does not log partial results or publish `ROUND_DONE` after its wait is aborted. DH-PSI checks cancellation immediately after every blocking broadcast and multicast, before consuming results or starting the next phase; aborted backward results are not merged into workflow state.
- Install `MetricsArtifactWriter` in the base Cyclic recipe. PyTorch and TensorFlow Cyclic recipes inherit the same behavior.
- Run Swarm presentation on the client selected to aggregate that round, where the aggregation and metrics actually exist. Because aggregation ownership can rotate, the client writers are presentation-only and do not publish competing partial metric summaries.
- Persist site-only round records for non-aggregation workflows such as Cyclic and FedEval, including metrics omitted from the bounded console table, and render those records under the workflow's title in `RUN SUMMARY`.
- Emit concise workflow phases for federated statistics and DH-PSI. FedStats checks cancellation after both statistics passes and before post-processing, so an aborted derived-statistics phase cannot report success. DH-PSI progress reports every logarithmic encrypted-reduction pass, distribution, and agreement verification without printing site identities, set contents, or set sizes. Existing diagnostic-file logging remains unchanged.
- Emit cleartext and homomorphic-encryption phases for the Kaplan–Meier survival workflow. Each blocking collection is followed by an abort check before aggregation or the next phase begins. Completion is reported only after every explicitly sampled target acknowledges the final survival curve and the workflow remains un-aborted.
- Carry FedEval's single evaluation-round identity on the transmitted model so result callbacks with fresh contexts still persist round `0` rather than `null`.
- Replace collective XGBoost's stdout evaluation monitor with one callback that preserves five-decimal per-iteration diagnostic records on every rank while only rank zero emits the bounded progress table. Tree-based bagging retains its incoming global-model evaluation in `INITIAL_METRICS` for model-selection consumers and carries its post-training result in progress-only metadata. Metric artifacts keep those phases separate as per-site `metrics` and `progress_metrics`. The console likewise groups incoming-model client values with their aggregate, then renders local post-training values in a separately labeled table, so values from different phases cannot appear to share one calculation. Metric metadata uses XGBoost's returned name, including normalized aliases such as configured `error@0.5` being returned as `error`. Cyclic and scratch-training paths publish only their post-training `PROGRESS_METRICS`; the built-in cyclic event flow has no model-selection consumer for an incoming metric, so it retains a single post-update evaluation.
- Recognize federated-statistics directories and scikit-learn/XGBoost model formats in `RUN SUMMARY` without deserializing artifacts.
- Have the affected runnable examples request their result so success/failure summaries are presented.

`concise` remains the simulator default. The new records use the existing progress marker, so full, verbose, concise, and diagnostic-file behavior retain their existing filters.

Accurate tree-based bagging progress requires a post-training validation evaluation per local round in addition to the incoming-model evaluation used by model-selection consumers. A fresh bagging executor receiving a restored model performs the incoming evaluation once before local training. Cyclic performs only its existing post-update evaluation. The bagging progress evaluation is not gated on console log mode because the same current-model value is written to logging-mode-independent metric artifacts; gating it would make those artifacts change meaning between `concise` and `progress` runs.

## Recipe coverage audit

The 24 packaged recipes and the example-owned survival workflow were checked by workflow ownership rather than framework name:

| Progress contract | Recipes |
| --- | --- |
| Standard round events and shared metric writer | FedAvg (core, NumPy, JAX through NumPy exchange, PyTorch, TensorFlow, and scikit-learn), FedAvg HE, FedCE, FedProx, FedOpt, SCAFFOLD, Cyclic, KMeans, SVM, and Newton-Raphson logistic regression |
| Workflow-owned round reporting | Swarm and all three XGBoost recipe families |
| Workflow phase reporting | FedStats and DH-PSI |
| Example-owned phase reporting | Kaplan–Meier cleartext and homomorphic-encryption survival analysis |
| Evaluation reporting | FedEval and cross-site evaluation |
| No inferred algorithm progress | `FedTaskRecipe`, because it is a generic user-defined one-shot task; Flower, because its external Flower runtime owns its iteration semantics |
 Edge FedBuff already installs the shared writer and emits standard round events. `CollabRecipe` remains user-defined orchestration and cannot infer a universal round or metric contract.

## User-facing output

Values and elapsed times vary by job. These examples show the records selected by `--log_config progress`.

### Traditional ML

Scikit-learn FedAvg, KMeans, SVM, and Newton-Raphson logistic regression use the same presentation as deep-learning FedAvg. Joblib and Pickle models also appear in the final artifact list.

```text
============================= ROUND 1 / 5 ==============================

  Training

  Client                               AUC
  site-1                            0.8124
  site-2                            0.8398
  ------------------------------------------------------------------
  Aggregated                        0.8261

  ✓ Aggregated 2 client updates                              0.6s

============================= RUN SUMMARY ==============================

  ✓ Completed

  Models    server/simulate_job/app_server/
```

### Federated statistics

```text
  Federated statistics · 2 clients

  Preparing client datasets…
  Computing first-pass statistics…
  Computing derived statistics…

  ✓ Federated statistics completed

============================= RUN SUMMARY ==============================

  ✓ Completed

  Statistics server/simulate_job/statistics/
```

### Swarm learning

Swarm progress is produced by the client selected to aggregate that round.

```text
============================= ROUND 1 / 3 ==============================

  Swarm learning

  Client                          accuracy          loss
  site-1                               0.81          0.42
  site-2                               0.84          0.38

  ✓ Aggregated 2 client updates                              1.2s

============================= ROUND 2 / 3 ==============================

  Swarm learning

  Client                          accuracy          loss
  site-1                               0.86          0.34
  site-2                               0.87          0.31

  ✓ Aggregated 2 client updates                              1.1s

============================= ROUND 3 / 3 ==============================

  Swarm learning

  Client                          accuracy          loss
  site-1                               0.88          0.29
  site-2                               0.89          0.27

  ✓ Aggregated 2 client updates                              1.0s
```

### Private set intersection

The number of reduction passes depends only on the participant count, which is already displayed. No private item values or intersection sizes are printed by the new progress records. Existing diagnostic-file logging remains available for troubleshooting.

```text
  Private set intersection · 3 clients

  Preparing private inputs…
  ✓ Private inputs prepared
  Computing encrypted intersection…
  Reducing encrypted inputs · pass 1/2
  Reducing encrypted inputs · pass 2/2
  Distributing encrypted intersection…
  Verifying intersection agreement…

  ✓ Private set intersection completed
```

### Survival analysis

The Kaplan–Meier examples report the protocol phases they own. The encrypted variant adds the histogram-bounds and encrypted-histogram phases. A success record appears only after the final survival curve reaches every sampled client.

```text
  Federated Kaplan-Meier survival analysis · homomorphic encryption

  Collecting histogram bounds…
  Collecting encrypted survival histograms…
  Aggregating encrypted histograms…
  Distributing the encrypted global survival curve…

  ✓ Encrypted survival analysis completed
```

### Cyclic, SCAFFOLD, and evaluation

Cyclic reports processed updates because it relays the model rather than aggregating parallel updates. SCAFFOLD uses the normal aggregation presentation. FedEval presents one bounded evaluation round.

```text
============================= ROUND 1 / 3 ==============================

  Training

  Client                          accuracy          loss
  site-1                               0.81          0.42
  site-2                               0.84          0.38

  ✓ Processed 2 client updates                               0.8s
```

```text
============================= ROUND 1 / 1 ==============================

  Model evaluation

  Client                          accuracy          loss
  site-1                               0.88          0.29
  site-2                               0.89          0.27

  ✓ Processed 2 client updates                               0.5s
```

### XGBoost

Collective XGBoost reports the latest evaluation and training metrics from rank zero. Tree-based bagging and cyclic XGBoost reuse the shared server-side metric table.

```text
============================= ROUND 1 / 10 =============================

  XGBoost training

  Dataset                              auc       logloss
  eval                              0.8124        0.4362
  train                             0.8398        0.4017
```

## Validation

- Abort-boundary regression: `19 passed` across the complete PSI suite, including cancellation at each of the four forward and four backward communication phases.
- Latest XGBoost, workflow, and metric-writer suite: `260 passed, 2 skipped`; focused final regressions: `92 passed`.
- The broader Recipe, workflow, PSI, CCWF, and example suites passed (socket-dependent cases rerun with host networking).
- PyTorch recipe suite: `43 passed`.
- XGBoost suite: `36 passed`, including first-round `auc` and normalized `error@0.5` metric metadata.
- Metric-writer, run-summary, and XGBoost review suite: `138 passed, 3 skipped`, including distinct aggregation/progress phases and a writer-to-summary site-only evaluation regression.
- Final phase-label regression suite: `109 passed, 3 skipped`, covering metric artifacts, run summaries, XGBoost metadata, and distinct incoming-model/aggregate/post-training console sections.
- Cyclic event and metric-writer review suite: `62 passed`; unconvertible payloads do not publish accepted-contribution progress events.
- Kaplan–Meier cleartext/HE progress and completion suite: `9 passed`.
- Documentation HTML build passed with the repository's existing warnings.
- A real two-client `hello-tabular-stats` simulation under `--log_config progress` showed the phase messages, completed summary, statistics artifact location, logs, and result path.

The broader suites were run in separate processes because a combined Torch/TensorFlow/XGBoost process completed 95% without assertion failures and then segfaulted during native-library teardown; each package suite passes independently.


nvflare/app_common/workflows/cyclic_ctl.py
```diff
@@ -226,6 +226,9 @@ def _process_result(self, client_task: ClientTask, fl_ctx: FLContext):
                         fl_ctx,
                         f"Ignored {rc} from client {client_task.client.name} because early termination is not allowed",
                     )
+            fl_ctx.set_prop(AppConstants.TRAINING_RESULT, result, private=True, sticky=False)
+            fl_ctx.set_prop(AppConstants.AGGREGATION_ACCEPTED, True, private=True, sticky=False)
+            self.fire_event(AppEventType.AFTER_CONTRIBUTION_ACCEPT, fl_ctx)
```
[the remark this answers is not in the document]
chesterxgchen: Addressed in 4cbca6245. In this workflow, disabling early termination ignores only the request to stop: when the payload converts successfully, its learnable is deliberately forwarded to the next client, so it remains a processed update. The acceptance flag now follows conversion success. If an EARLY_TERMINATION payload cannot be converted, AFTER_CONTRIBUTION_ACCEPT is emitted with AGGREGATION_ACCEPTED=False, so MetricsArtifactWriter does not count or display it. Tests cover both the successfully forwarded and rejected cases.

nvflare/app_common/workflows/cyclic_ctl.py
```diff
@@ -281,6 +285,9 @@ def control_flow(self, abort_signal: Signal, fl_ctx: FLContext):
                     abort_signal=abort_signal,
                 )
 
+                if not self._is_done and not abort_signal.triggered:
+                    log_progress(self.logger, "  ✓ Cyclic round completed")
```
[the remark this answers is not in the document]
chesterxgchen: Addressed in 55f073a16. The direct “✓ Cyclic round completed” message was removed. Cyclic now emits the shared ROUND_DONE event, and MetricsArtifactWriter reports only the accepted client updates actually processed; a timeout or dead-client exit with no accepted contribution produces no completion line. Partial work is described as processed updates rather than a completed round, while the terminal job status remains authoritative.

nvflare/app_opt/xgboost/histogram_based_v2/controller.py
```diff
@@ -594,13 +602,17 @@ def control_flow(self, abort_signal: Signal, fl_ctx: FLContext):
...
+        log_progress(self.logger, "  Training collectively…")
         self.log_info(fl_ctx, "Waiting for clients to finish ...")
         while not self._is_stopped():
             done = self._check_job_status(fl_ctx)
             if done:
                 break
             time.sleep(self.job_status_check_interval)
 
+        if self.client_statuses and all(status.xgb_done for status in self.client_statuses.values()):
+            log_progress(self.logger, "\n  ✓ XGBoost training completed")
```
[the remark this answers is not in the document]
chesterxgchen: Addressed in 55f073a16. The server-side completion message based on all(status.xgb_done) was removed. Collective XGBoost progress and completion are now owned by the rank-zero training process: per-round metrics come from an XGBoost TrainingCallback, and “✓ XGBoost training completed” is emitted only after xgb.train() returns normally. Nonzero or aborted client termination can no longer trigger that success message.

examples/advanced/kaplan-meier-he/server.py
```diff
@@ -27,9 +28,13 @@ def __init__(self, min_clients: int):
         self.num_rounds = 2
 
     def run(self):
+        log_progress(self.logger, "\n  Federated Kaplan-Meier survival analysis\n\n  Collecting local histograms…")
         hist_local = self.start_fl_collect_hist()
+        log_progress(self.logger, "  Aggregating survival histograms…")
         hist_obs_global, hist_cen_global = self.aggr_hist(hist_local)
+        log_progress(self.logger, "  Distributing the global survival curve…")
         _ = self.distribute_global_hist(hist_obs_global, hist_cen_global)
+        log_progress(self.logger, "\n  ✓ Survival analysis completed")
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in 1742325a4. Both cleartext and HE workflows now retain the final distribution results and emit the completion record only when every targeted client acknowledged the result and the workflow was not aborted. Partial delivery and abort therefore leave the phase output without a false success line. A focused regression covers partial delivery, complete delivery, and abort; all three cases pass.

chesterxgchen:
Addressed the remaining progress-review findings in `fb93d17ca`:

- `ROUND_DONE` now persists normalized site-only records for Cyclic/FedEval and enables the normal end-of-run metrics summary, so bounded console omissions point to real artifacts.
- Collective XGBoost no longer installs an unconditional `EvaluationMonitor`; `verbose_eval=False` now produces only the formatted callback output, while callers can still explicitly enable native verbose output.
- Swarm client writers are presentation-only. The selected aggregation client still renders each round, but rotating owners no longer publish several incomplete metric histories as competing summaries. Omission messages no longer claim an artifact exists in this mode.
- FedStats displays the connected participant count recorded at controller start instead of labeling `min_clients` as the federation size.
- Newton-Raphson FedAvg and SCAFFOLD now pair successful `ROUND_STARTED` events with `ROUND_DONE`.
- Cyclic emits `ROUND_DONE` only for an `OK` relay task. Full-model weight-diff conversion stages every update before committing it, so a late malformed entry cannot leave a partially modified model to forward.

The earlier XGBoost server-completion, cyclic direct-completion, accepted-update, and Kaplan-Meier final-delivery findings remain resolved by the preceding commits.

Validation: 142 focused regressions passed; the broader Recipe/workflow/PSI/CCWF/example suite completed with 1,371 passed, 20 skipped, and 2 subtests passed; XGBoost has 31 passed; the full Black, isort, flake8, and agent-skill style check passed. The six socket-dependent cases denied by the sandbox all passed when rerun with host networking. The pushed commit has a verified signature.

nvflare/app_common/workflows/cyclic_ctl.py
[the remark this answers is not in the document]
chesterxgchen: Follow-up in fb93d17ca closes the remaining task-status edge directly. Cyclic now emits ROUND_DONE only when relay_and_wait leaves task.completion_status == TaskCompletionStatus.OK and the workflow is neither stopped nor aborted. TIMEOUT, CLIENT_DEAD, ERROR, and other unsuccessful relay exits therefore produce no round-completion event. A focused timeout regression verifies ROUND_STARTED is emitted while ROUND_DONE is not.

nvflare/app_opt/xgboost/histogram_based_v2/runners/xgb_client_runner.py
```diff
@@ -111,7 +149,7 @@ def _xgb_train(self, num_rounds, xgb_params: dict, xgb_options: dict, train_data
         # Specify validations set to watch performance
         watchlist = [(val_data, "eval"), (train_data, "train")]
 
-        callbacks = [callback.EvaluationMonitor(rank=self._rank)]
+        callbacks = [_ProgressCallback(self.logger, self._rank, num_rounds)]
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in 0637f2bc4 without restoring the duplicate stdout monitor. The rank-zero callback now writes every current eval/train metric through the runner normal INFO logger before emitting the bounded progress table. Full/verbose output and log files therefore retain detailed per-iteration records, while progress mode filters ordinary INFO and displays only the formatted progress record. Nonzero ranks emit neither record. The XGBoost suite passes (31 tests), including assertions for the diagnostic line, formatted rows, and absence of an explicit EvaluationMonitor when verbose_eval=False.

chesterxgchen:
Addressed both findings in `eb9f2b8c4`:

- FedStats now checks the abort signal immediately after the second statistics task and again after waiting for results. An abort cannot reach post-processing or emit `✓ Federated statistics completed`, even when first-pass results already satisfy `min_clients`.
- FedEval now sets `start_round=0`, `current_round=0`, and `total_rounds=1` on the outgoing `FLModel`. Those fields travel with the model/result path, so fresh callback contexts still produce round `0` in `round_metrics.jsonl` and `metrics_summary.json`.

Focused regression validation: `84 passed`. Broader workflow and PyTorch recipe validation: `219 passed, 2 skipped`. The full Black, isort, flake8, and agent-skill style check passed. The commit is signed and the branch remains `0` commits behind `upstream/main`.

chesterxgchen:
Addressed the two gated findings and the related compatibility note in signed commit `715d5cf2f`:

- Tree-based bagging now evaluates the locally updated booster after the local training step and publishes that value through `INITIAL_METRICS`. The existing TensorBoard record remains the incoming global-model metric and is labeled accordingly. A regression test distinguishes the incoming `0.80` value from the post-training `0.85` value.
- The Swarm test now invokes `SwarmClientController.do_learn_task()` and asserts that the controller publishes `PROGRESS_OWNER`, current round, total rounds, and title on `ROUND_STARTED`. Removing the producer assignment makes this test fail.
- `StatisticsController.post_fn()` retains its established return contract. Completion is derived from controller state and the abort signal, so existing overrides that return `None` continue to receive an accurate completion line.
- The PR description now limits the PSI privacy statement to the new progress records. Existing diagnostic-file logging remains unchanged so `log.txt` retains troubleshooting details.

Validation: the broader XGBoost, Swarm-controller, and statistics suite passes (`43 passed`), and the full project style check passes (Black, isort, flake8, and agent-skill lint).

chesterxgchen:
Scope correction: Kaplan–Meier progress is restored directly in #5319 by signed commit `301378caf`, because this PR covers progress logging across all supported workflow categories. Both cleartext and HE phases are included, and completion requires responses from every explicitly sampled final target with no abort. The logging guide, Recipe API guide, PR description, and six focused tests are restored here. #5321 is closed as superseded.

Independent example defects remain separate: #5323 sanitizes the pre-existing DH-PSI diagnostic leak while retaining useful checkpoints and timings, and #5322 fixes dataframe scalar compatibility.

chesterxgchen:
Addressed the latest correctness and scope findings in signed commit `7749fb52d`:

- Tree-based XGBoost again keeps the incoming global-model score in `INITIAL_METRICS`, preserving `IntimeModelSelector` semantics. The post-training local score uses a new progress-only metadata field that `MetricsArtifactWriter` prefers for display and metric artifacts.
- Collective XGBoost retains five-decimal per-iteration diagnostic records on every rank. Only rank zero emits the bounded progress table, so progress stays concise without reducing `log.txt` diagnostics.
- Statistics post-processing now has one validation on the base path. An explicit `False` suppresses completion, while legacy overrides returning `None` remain compatible through one fallback validation. The preparation phase is printed only when pre-run preparation is enabled.
- Newton-Raphson FedAvg and SCAFFOLD recheck the abort signal immediately after `send_model_and_wait()` and return before aggregation, persistence, or `ROUND_DONE`.
- The two documentation references to Kaplan-Meier/survival progress were removed; that example-local work is now isolated in #5321.
- Renamed the writer's side-effect-only round helper to `_update_total_rounds()` and cleared per-round XGBoost metrics at the start of training to prevent stale reuse.

Validation: `260 passed, 2 skipped` across XGBoost, workflow, and metric-writer suites; `92 passed` in the focused final regressions; the two socket-dependent Recipe cases pass with host networking; full Black, isort, flake8, and agent-skill style checks pass.

chesterxgchen:
The pre-existing DH-PSI diagnostic-log privacy issue is now isolated in #5323. That follow-up retains the preparation, ordering, forward/backward, intersection, validation, and timing checkpoints, while replacing raw site-to-size mappings with participant/response counts and protocol status. No PSI algorithm or message behavior changes in this PR.

nvflare/app_opt/xgboost/histogram_based_v2/runners/xgb_client_runner.py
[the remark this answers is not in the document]
chesterxgchen: Follow-up correction for the current head (`7749fb52d`): the earlier reply's statement that nonzero ranks emit neither record is no longer accurate. The callback now writes the five-decimal per-iteration `eval`/`train` diagnostic record through the normal INFO logger on **every rank**, preserving the detail that existing jobs had in `log.txt`. Only the bounded progress table and final progress completion line are restricted to rank zero, avoiding duplicate terminal progress without reducing diagnostic logs.

The regression test now verifies both sides of that contract: a nonzero rank emits the normal diagnostic record and does not call `log_progress_round`. This follows the same rule used for the PSI follow-up: keep useful diagnostic information in retained logs while making the focused progress view concise.

docs/user_guide/admin_guide/configurations/logging_configuration.rst
```diff
@@ -331,14 +331,15 @@ This argument can be any of the following:
 
     - ``concise`` (default for simulator mode): all non-NVFlare logs plus selected NVFlare application logs, with
       simplified log attributes. This preserves the existing NVFlare 2.9 behavior.
-    - ``progress``: round progress, client and aggregated metrics, evaluation progress,
-      warnings, and errors. INFO progress is formatted as messages only; WARNING,
+    - ``progress``: training rounds and metrics for deep-learning, traditional-ML, Swarm, cyclic,
+      and XGBoost jobs, evaluation progress, and workflow phases for federated statistics, PSI, and
+      survival analysis, plus warnings and errors. INFO progress is formatted as messages only; WARNING,
```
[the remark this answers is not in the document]
chesterxgchen: Corrected the scope in signed commit `301378caf`: Kaplan–Meier progress belongs in this PR because #5319 is the cross-workflow progress-logging change. Both cleartext and homomorphic-encryption phase reporting are restored here, together with truthful completion gating: the success record is emitted only after every explicitly sampled target acknowledges the final survival curve and the workflow is not aborted. Six focused success, partial-response, and abort tests pass. The logging guide, Recipe API guide, and PR description again include survival analysis.

The separate #5321 split is superseded and has been closed. Independent example defects such as the PSI diagnostic privacy issue (#5323) and dataframe scalar compatibility (#5322) remain separate because they exist regardless of the progress logging mode.

chesterxgchen:
Addressed Review #535 in signed commit `c99cc883f`. On a fresh bagging executor that receives a restored/pretrained model, the client now loads and evaluates that supplied booster before local training. The incoming score is returned as `INITIAL_METRICS` for model-selection consumers, while the final `evals_result` score is returned as `PROGRESS_METRICS`. The behavioral regression exercises the complete first-call `train()` path and verifies `0.80` incoming versus `0.85` post-training metadata. The complete XGBoost unit suite passes (`33 passed`).

I retained the post-training validation evaluation noted in the re-review. It supplies the current-model value to `MetricsArtifactWriter` and its persisted metrics artifacts as well as the focused console view. Console log mode is not a sound gate: the writer is installed independently of `progress`, and gating would make artifact semantics differ between `concise` and `progress` runs. The PR description now states this one-evaluation-per-local-round cost explicitly.

chesterxgchen:
Addressed the two current review findings in signed commit `5d40ff067`:

- Both Kaplan–Meier controllers now check the abort signal after every blocking collection and again before starting a later protocol phase. An abort during local-histogram collection, HE bounds collection, or encrypted-histogram collection returns immediately; no partial aggregation, later broadcast, later phase message, or success message occurs. The regressions assert those later operations are not called.
- Tree-based XGBoost now uses one metadata contract in every path. `INITIAL_METRICS` is present only when the incoming model was actually evaluated. Post-training bagging, cyclic, restored-model, and scratch-training values use `PROGRESS_METRICS`; scratch training omits `INITIAL_METRICS` because no incoming model exists.

The additional post-training evaluation remains intentional and is documented in the PR description. The client has no supported signal for the console's selected log mode, and the value also feeds logging-mode-independent metric artifacts. Gating it on console mode would make the artifact contract differ between `concise` and `progress` runs.

Validation: the full XGBoost unit directory plus the Kaplan–Meier suite passes (`44 passed`), the focused files pass (`14 passed`), `git diff --check` is clean, and the full Black/isort/flake8/agent-skill style check passes.

chesterxgchen:
Applied the two useful documentation follow-ups in signed commit `39dfe744f`:

- The logging configuration and Recipe API guides now document `AppConstants.PROGRESS_METRICS` as the post-training display/artifact channel while preserving `INITIAL_METRICS` for incoming-model and model-selection semantics.
- The centralized tree-based XGBoost parser now records the exact `eval_set()` string shape it expects.

The other two observations were already explicit in the PR description: current tree-based progress requires an additional post-training validation evaluation, and collective XGBoost preserves five-decimal per-iteration diagnostics on every rank while limiting the bounded progress presentation to rank zero.

Validation: the XGBoost suite passes (`35 passed`), the full documentation build succeeds with existing warnings, and the full Black/isort/flake8/agent-skill style check passes.

chesterxgchen:
Addressed all three findings in signed commit `8655acdcd`:

- Both Kaplan–Meier `run()` methods now return immediately when the abort signal is already triggered, before the first phase record or client collection is dispatched. The pre-abort regression now asserts that `start_fl_collect_hist()` / `start_fl_collect_max_idx()` is never called.
- Cyclic tree-based XGBoost again performs one evaluation, after the local update, and publishes only `PROGRESS_METRICS`. Its built-in event flow has no `IntimeModelSelector` consumer for an incoming metric, so the added pre-update evaluation was unnecessary. Bagging retains the incoming `INITIAL_METRICS` evaluation used by model selection and the post-training progress evaluation. A fresh cyclic executor receiving a model is also covered and does not construct a separate `Booster` for an incoming evaluation.
- Both guides now specify that `AppConstants.PROGRESS_METRICS` must contain a metric mapping and include `{"auc": 0.85}` as an example.

The PR description now distinguishes the one-evaluation cyclic path from the dual-evaluation bagging path and notes that the temporary incoming-model booster is used only on a fresh bagging executor receiving a restored model.

Validation: the full XGBoost directory plus Kaplan–Meier suite passes (`44 passed`), the focused files pass (`14 passed`), the documentation build succeeds with existing warnings, and the full Black/isort/flake8/agent-skill style check passes.

nvflare/app_opt/xgboost/tree_based/executor.py
```diff
@@ -248,22 +258,35 @@ def train(
...
                     num_boost_round=self.num_local_round,
                     evals=[(self.val_data, "validate"), (self.train_data, "train")],
+                    evals_result=evals_result,
                 )
             else:
                 loadable_model = bytearray(model_update["model_data"])
+                if self.training_mode == "bagging":
+                    incoming_bst = xgb.Booster(params=params)
+                    incoming_bst.load_model(loadable_model)
+                    incoming_metric = self._evaluate_model(incoming_bst, fl_ctx)
```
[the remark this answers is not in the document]
chesterxgchen: This does not apply to the built-in cyclic XGBoost event path. `CyclicController` publishes accepted updates only through `AFTER_CONTRIBUTION_ACCEPT`; it never fires `BEFORE_CONTRIBUTION_ACCEPT`, which is the event consumed by `IntimeModelSelector`. The cyclic recipe installs `MetricsArtifactWriter`, which reads the post-training `PROGRESS_METRICS` on the after-accept event. Therefore an `INITIAL_METRICS` value on a cyclic contribution has no built-in model-selection consumer and cannot change whether that contribution is accepted or selected.

Restoring it would require a full validation evaluation before every cyclic local update, in addition to the post-update evaluation used for progress and artifacts, for the default 100-round path. That performance regression was removed deliberately in signed commit `8655acdcd` after focused review. Bagging retains `INITIAL_METRICS` because its scatter-and-gather path does fire `BEFORE_CONTRIBUTION_ACCEPT` and model selection consumes the incoming-model metric there. Cyclic intentionally emits only the post-training `PROGRESS_METRICS`, and the regression tests pin both contracts.

nvflare/app_common/shareablegenerators/full_model_shareable_generator.py
```diff
@@ -63,8 +63,10 @@ def shareable_to_learnable(self, shareable: Shareable, fl_ctx: FLContext) -> Mod
             weights = base_model[ModelLearnableKey.WEIGHTS]
             if dxo.data is not None:
                 model_diff = dxo.data
+                updated_weights = {}
                 for v_name, v_value in model_diff.items():
-                    weights[v_name] = weights[v_name] + v_value
+                    updated_weights[v_name] = weights[v_name] + v_value
```
nvkevlu: [P1] Avoid retaining another full model during diff application

`updated_weights` retains every newly allocated parameter while `weights` still retains the entire old model; the incoming `model_diff` is also resident. For the existing 4-GB/large-model `WEIGHT_DIFF` jobs that use this generator, this increases peak memory from roughly two model copies plus one parameter to roughly three full copies and can turn a working run into an OOM. Please preserve the no-partial-mutation behavior without materializing all updated tensors at once—for example, prevalidate all keys/shapes and then update one parameter at a time—and add a large-model memory regression test.

nvflare/app_opt/pt/recipes/fedeval.py
```diff
@@ -46,11 +48,19 @@ def __init__(self, persistor_id: str, timeout: int):
...
+        self.fl_ctx.set_prop(AppConstants.CURRENT_ROUND, 0, private=True, sticky=False)
+        self.fl_ctx.set_prop(AppConstants.NUM_ROUNDS, 1, private=True, sticky=False)
+        self.fl_ctx.set_prop(AppConstants.PROGRESS_TITLE, "Model evaluation", private=True, sticky=False)
+        self.event(AppEventType.ROUND_STARTED)
         self.info("Sending model for evaluation")
         results = self.send_model_and_wait(targets=None, data=model, task_name="validate", timeout=self.timeout)
         self.info(f"Got {len(results)} results")
         for r in results:
             self.info(f"Metrics: {r.metrics}")
+        self.event(AppEventType.ROUND_DONE)
```
nvkevlu, line 65: [P2] Do not mark an aborted evaluation round complete

`send_model_and_wait()` waits with `self.abort_signal` and returns when that signal is triggered. This unconditional event therefore marks an aborted run as a finished one-round evaluation; if any results arrived before cancellation, the default `MetricsArtifactWriter` can display and persist those partial results as the completed round. Please return when `self.abort_signal.triggered` before logging the results and firing `ROUND_DONE`, and add an abort-path test as the new FedAvgLR and Scaffold paths do.

nvflare/app_common/psi/dh_psi/dh_psi_workflow.py
```diff
@@ -74,6 +79,7 @@ def run(self, abort_signal: Signal):
 
         self.check_processed_sites(intersect_site, self.forward_processed)
 
+        log_progress(self.logger, "  Distributing encrypted intersection…")
```
nvkevlu, line 84: [P2] Stop phase reporting when a protocol pass is aborted

Please check the abort signal after each protocol pass before announcing or starting the next phase. If cancellation arrives during or just after `forward_pass`, execution can still announce “Distributing…”; `backward_pass` then returns an empty result immediately, but the workflow proceeds to announce “Verifying…” and calls `check_final_intersection_sizes`, which accepts an empty mapping or may inspect stale entries if the component is reused. The same misleading transition occurs when cancellation arrives during `backward_pass`. Could we return after each pass when `self.abort_signal.triggered`?

nvflare/app_common/shareablegenerators/full_model_shareable_generator.py
chesterxgchen: Fixed in signed commit `6da352355` without retaining a third model. The generator first evaluates every requested addition while leaving the base model untouched, discarding each validation result immediately. Only after all additions succeed does it apply them one parameter at a time. A missing key or late incompatible value therefore leaves every original weight unchanged, while peak retained data is the base model, the incoming diff, and at most one temporary parameter.

The regression uses tracked parameter objects to measure live allocations: for N base parameters plus N diff parameters, the conversion may retain at most `2N + 1` objects. The prior `updated_weights` implementation reaches `3N` and fails the test. A separate late-value failure verifies that validation of a later parameter cannot leave an earlier parameter mutated.

nvflare/app_opt/pt/recipes/fedeval.py
chesterxgchen: Fixed in signed commit `6da352355`. `EvalController.run()` now checks `self.abort_signal.triggered` immediately after `send_model_and_wait()` and returns before logging returned metrics or firing `ROUND_DONE`. The regression triggers cancellation while returning a partial metric result and verifies that only `ROUND_STARTED` is emitted and neither the result count nor metrics are reported.

nvflare/app_common/psi/dh_psi/dh_psi_workflow.py
chesterxgchen: Fixed in signed commit `6da352355`. DH-PSI now checks the abort signal immediately after both `forward_pass()` and `backward_pass()`. An abort during forward reduction returns before validation or the distributing phase; an abort during backward distribution returns before the verification phase and final-size validation. Focused regressions cover both boundaries and verify that later protocol calls and progress records are absent.

holgerroth (dismissed):
Reviewed commit 6da352355. No actionable findings; the earlier review fixes appear addressed. Local validation: 270 tests passed, 4 skipped, plus a real XGBoost check of incoming-model and post-training metric metadata. Distributed/GPU end-to-end execution was not validated.

chesterxgchen:
Addressed Review #547 in signed commit `366be6594`.

DH-PSI now checks the abort signal immediately after every blocking broadcast and multicast before consuming the returned value or starting the next protocol phase. The recursive forward pass and backward pass also stop on entry when cancellation is already active. `run()` now checks cancellation before merging backward results, so an aborted partial result cannot update workflow state.

The regressions execute the real forward and backward phase chains and interrupt each of their four communication boundaries. They verify that no later operator is created, no `None` result is indexed, and aborted backward data is not retained.

Validation: `19 passed` across the complete PSI suite; the full Black/isort/flake8/agent-skill style check passed.

nvflare/app_opt/xgboost/tree_based/executor.py
```diff
@@ -248,22 +258,35 @@ def train(
...
                 bst = xgb.train(
                     params,
                     self.train_data,
                     num_boost_round=self.num_local_round,
                     xgb_model=loadable_model,
                     evals=[(self.val_data, "validate"), (self.train_data, "train")],
+                    evals_result=evals_result,
                 )
+            validation_metrics = evals_result.get("validate", {})
+            metric_values = validation_metrics.get(self.eval_metric, [])
```
holgerroth: [P2] Handle XGBoost’s normalized metric names

XGBoost reports `eval_metric="error@0.5"` under the key `"error"`, so this exact-name lookup silently drops the post-training metric on the first invocation. Reproduced with the default one-round bagging configuration: training succeeds, but the contribution has no progress metadata and `MetricsArtifactWriter` writes neither `round_metrics.jsonl` nor `metrics_summary.json`. The equivalent `eval_metric="error"` configuration produces both artifacts for the same data.

Please resolve the metric name returned by XGBoost and add a regression covering this valid alias.

holgerroth (commented):
One progress-reporting issue reproduced with real XGBoost; see the inline finding. Targeted validation: 282 tests passed, 4 skipped.

nvflare/app_common/widgets/metrics_artifact_writer.py
```diff
@@ -281,14 +330,17 @@ def _handle_after_contribution_accept(self, fl_ctx: FLContext):
         except Exception:
             return
         self._round_contribution_count += 1
-        if not model.metrics:
+        meta = model.meta or {}
+        progress_metrics = meta.get(AppConstants.PROGRESS_METRICS)
+        metrics_source = progress_metrics if isinstance(progress_metrics, dict) else model.metrics
```
YuanTingHsieh: [P2] Preserve the evaluation phase when overriding client metrics

For an executor publishing both incoming-model metrics and PROGRESS_METRICS with a standard FedAvg aggregator, this replaces the displayed and persisted client values with post-training values, while aggregation still consumes the original FLModel.metrics. Using the real aggregation helper and writer, two incoming accuracies of 0.2/0.4 with progress accuracies of 0.8/0.9 produce one round artifact containing site values 0.8/0.9 and aggregated accuracy 0.3. The record labels the aggregation as weighted_average and provides no indication that the site and aggregate values describe different model states. Please preserve distinct metric phases/sources in the presentation and artifact contract, or keep the aggregation-related client values unchanged and record progress values separately. Add coverage through aggregation and persistence together.

nvflare/app_common/widgets/metrics_artifact_writer.py
```diff
@@ -243,6 +260,38 @@ def _handle_after_aggregation(self, fl_ctx: FLContext):
...
+        log_progress(_logger, "\n" + f"  {completion}".ljust(64) + duration)
+        self._reset_progress()
+
+        if sites or skipped:
+            self._append_round_record(
+                fl_ctx,
+                {
+                    "round": current_round,
+                    "aggregated_metrics": [],
+                    "sites": sites,
```
YuanTingHsieh, line 309: [P2] Update the run summary to consume site-only metric records

The new Cyclic/FedEval records put all metrics in sites and leave aggregated_metrics empty, but result_summary() in nvflare/recipe/_run_summary.py still reads only aggregated_metrics and labels every such table "Training · aggregated client metrics". I reproduced a one-round evaluation with accuracy 0.88 using ROUND_STARTED, AFTER_CONTRIBUTION_ACCEPT, ROUND_DONE, and END_RUN: the metric is saved correctly, but result_summary() emits that training heading and a round number with no metric columns or values. Please update the summary consumer alongside this new record format so site-only/evaluation metrics are presented with the correct meaning, and add a writer-to-summary regression.

YuanTingHsieh (commented):
Reviewed commit 366be65945d12b13464a6ae2243ef1b43f8ada95. The two inline P2 findings are reproducible inconsistencies between metric production, artifact storage, and the final summary.

Please also separate these execution changes from the progress-reporting work:

- Tree-based XGBoost bagging now performs an additional post-training evaluation over both the training and validation data each local round, regardless of console log mode. Generating new measurements and defining their relationship to incoming-model/model-selection metrics changes the evaluation policy and runtime cost. Please move that evaluation/metric-contract change into a focused PR with an explicit policy and performance validation. Existing measurements can still be reported with their actual evaluation phase identified; evaluation policy should be independent of console verbosity.
- Please revert the new preflight loop in FullModelShareableGenerator from this PR. It computes every parameter addition, discards the result without checking a model-update invariant, and then repeats the addition. Every successful WEIGHT_DIFF conversion pays for two additions and two result allocations per parameter. Surfacing arithmetic exceptions before mutation does not establish update validity or guarantee failure atomicity during the second pass. Any intended model-validation or partial-update contract belongs in a separate change with explicit checks and tests for that contract.

The standard lifecycle events are a useful foundation. For this PR, please keep the scope around those events, progress presentation, and consistent artifact/summary handling. A small shared reporting contract should distinguish metric phases and let console presentation and artifact persistence consume the same meaning; the new PROGRESS_METRICS override currently allows those meanings to diverge.

Validation: 139 targeted tests passed, 3 skipped; both inline findings were additionally reproduced with the real MetricsArtifactWriter and result_summary reader. Distributed/GPU execution was not tested.

nvflare/app_opt/xgboost/tree_based/executor.py
chesterxgchen: Fixed in 3b568ab8a. Tree-based XGBoost now reads the metric name from XGBoost's actual evaluation output and uses that returned name consistently for incoming-model and post-training metadata. The first-round evals_result lookup keeps exact configured names when present and falls back to the normalized base name, so eval_metric="error@0.5" resolves the returned "error" series instead of dropping it. The first-bagging-invocation regression now covers both auc and error@0.5 and verifies INITIAL_METRICS plus PROGRESS_METRICS. Validation: all 36 XGBoost tests and the full project style check passed.

nvflare/app_common/shareablegenerators/full_model_shareable_generator.py
```diff
@@ -63,6 +63,11 @@ def shareable_to_learnable(self, shareable: Shareable, fl_ctx: FLContext) -> Mod
             weights = base_model[ModelLearnableKey.WEIGHTS]
             if dxo.data is not None:
                 model_diff = dxo.data
+                # Validate every update before mutating the model, but discard each
+                # validation result immediately to keep peak memory bounded for large models.
+                for v_name, v_value in model_diff.items():
+                    validated_value = weights[v_name] + v_value
+                    del validated_value
                 for v_name, v_value in model_diff.items():
                     weights[v_name] = weights[v_name] + v_value
```
YuanTingHsieh: we are duplicating these codes here

nvflare/app_common/widgets/metrics_artifact_writer.py
chesterxgchen: Fixed in signed commit `790b6e79c`. `sites[].metrics` now retains the contribution's aggregation/evaluation-phase values, while `AppConstants.PROGRESS_METRICS` is normalized and stored separately as `sites[].progress_metrics`. The bounded console still prefers the post-training progress values, but the artifact no longer labels those values as the inputs associated with the aggregate. The regression records incoming AUC `0.8`, post-training AUC `0.85`, and aggregate AUC `0.3`, and verifies all three remain distinct. The public artifact schema and progress-metadata documentation now describe this separation.

nvflare/app_common/widgets/metrics_artifact_writer.py
chesterxgchen: Fixed in signed commit `790b6e79c`. Each round record now carries its sanitized `progress_title`, and `result_summary()` renders records with no aggregate as `<workflow title> · client metrics`, using per-site progress metrics when present and evaluation metrics otherwise. The writer-to-summary regression publishes a `Model evaluation` round through `MetricsArtifactWriter`, finishes the run, and verifies the final summary shows the client AUC without an empty `Training · aggregated client metrics` table. Validation: 138 focused writer/run-summary/XGBoost tests passed with 3 skips; full style and documentation builds passed.

nvflare/app_common/shareablegenerators/full_model_shareable_generator.py
chesterxgchen: The repeated addition is intentional and is documented precisely in signed commit `3acf09e18`. The first pass evaluates every update while the base model remains untouched, catching missing keys and deterministic value incompatibilities before mutation. Each temporary result is immediately discarded, keeping peak memory to the base model, incoming diff, and one parameter. The second pass repeats those preflighted operations to update the weights. Retaining the first-pass results would remove the repeated computation but recreate the three-full-model peak that the earlier P1 review identified; mutating once and retaining rollback values has the same memory cost. This is a preflight guarantee rather than unconditional transactionality: failures arising only during the second pass, such as OOM or a stateful custom `__add__`, can still leave partial updates. The tests cover predictable late incompatibility and the `2N + 1` live-parameter bound.

nvflare/recipe/_run_summary.py
```diff
@@ -109,20 +115,45 @@ def result_summary(result):
...
+                            metrics = site.get("progress_metrics") or site.get("metrics")
+                            if not isinstance(metrics, list):
+                                continue
+                            site_values = {
+                                metric["name"]: metric["value"]
+                                for metric in metrics
+                                if isinstance(metric, dict) and "name" in metric and "value" in metric
+                            }
+                            if site_values:
+                                site_rows.append((_text(site.get("name", "unknown")), site_values))
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in signed commit `d98fe2f21`. Site-only rows now retain the parsed round label. A single-round evaluation keeps the concise `Client` table, while a multi-round Cyclic record uses `Round / client` labels such as `1 / site-1` and `2 / site-1`, so repeated clients are unambiguous. A two-round regression verifies the workflow title, table heading, both round/client labels, and their values. Validation: 106 focused tests passed with 3 skips, and the full style check passed.

nvflare/app_common/shareablegenerators/full_model_shareable_generator.py
YuanTingHsieh: Thanks, the narrowed preflight guarantee is clear. My concern remains the unconditional extra arithmetic and allocation pass. For the supported parameter types, why are explicit compatibility checks insufficient? Please separate this from progress reporting and justify the guarantee and cost with representative model benchmarks.

chesterxgchen: Agreed that this generic converter tradeoff should not be decided inside the progress-reporting PR. I removed the preflight and its converter-specific tests from #5319 in d76f6d0a8. Cyclic now stops on any conversion failure before publishing `AFTER_CONTRIBUTION_ACCEPT` or forwarding a model (3ada0d936); it does not change the shared converter's mutation behavior.

I opened draft #5327 to evaluate the converter change independently. Its description now covers the supported-value contract and the explicit-check alternative: `ModelLearnable` does not enforce a parameter value type, NumPy is the default exchange format, and raw/native values remain possible, so framework-specific key/shape/device checks would either narrow the API or reproduce only part of the actual `+` semantics. The draft keeps the limited real-addition preflight and explicitly excludes second-pass-only failures from its guarantee.

The draft also includes representative 25M/100M float32 NumPy and PyTorch CPU measurements. The preflight costs about 1.88–1.97x the addition time while retaining the same one-parameter peak RSS delta. We can review that tradeoff in #5327 and either adopt or close it without holding up #5319.

nvflare/app_common/workflows/cyclic_ctl.py
```diff
@@ -226,6 +228,14 @@ def _process_result(self, client_task: ClientTask, fl_ctx: FLContext):
...
                         f"Ignored {rc} from client {client_task.client.name} because early termination is not allowed",
                     )
+            # When early termination is disabled, only the request to stop is ignored. A
+            # successfully converted payload is still forwarded to the next client and is
+            # therefore a processed update. A conversion failure is not published as an
+            # accepted contribution; converter recovery semantics are outside this event.
+            if converted:
+                fl_ctx.set_prop(AppConstants.TRAINING_RESULT, result, private=True, sticky=False)
+                fl_ctx.set_prop(AppConstants.AGGREGATION_ACCEPTED, True, private=True, sticky=False)
+                self.fire_event(AppEventType.AFTER_CONTRIBUTION_ACCEPT, fl_ctx)
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in signed commit 3ada0d936. Any `shareable_to_learnable()` exception now stops and cancels the cyclic task immediately, including a disallowed `EARLY_TERMINATION` result. The controller returns before serializing `_last_learnable`, firing `AFTER_CONTRIBUTION_ACCEPT`, or preparing data for the next client, so a potentially partial model is never forwarded.

The focused regression verifies cancellation, absence of the accepted-contribution event, and that `learnable_to_shareable()` is not called. The cyclic/event-writer suite passes (`62 passed`), and the full project style check passes. Generic converter preflight remains isolated in draft #5327 for its separate cost/guarantee decision.

nvflare/app_common/shareablegenerators/full_model_shareable_generator.py
chesterxgchen: Update after the scope review: the converter preflight described in my earlier reply has been removed from #5319 in d76f6d0a8 and moved to draft #5327 for an independent cost/guarantee decision. The earlier reply should therefore not be read as the final #5319 behavior.

For this progress PR, signed commit 3ada0d936 now stops the cyclic task on any conversion exception and returns before serializing or forwarding the model. This prevents a partially converted model from reaching the next client without adding another full model or an unconditional validation pass to the shared converter. The regression verifies task cancellation and that no outgoing shareable is prepared.

nvflare/app_common/widgets/metrics_artifact_writer.py
```diff
@@ -301,26 +357,29 @@ def _handle_after_contribution_accept(self, fl_ctx: FLContext):
...
-        meta = model.meta or {}
+        if normalized_progress_metrics:
+            site["progress_metrics"] = normalized_progress_metrics
         weight = self._safe_weight(meta.get(FLMetaKey.NUM_STEPS_CURRENT_ROUND))
         if weight is not None:
             site["weight"] = weight
             site["weight_key"] = FLMetaKey.NUM_STEPS_CURRENT_ROUND
         sites.append(site)
-        self._log_contribution_progress(site["name"], metrics)
+        self._log_contribution_progress(site["name"], normalized_progress_metrics or metrics)
```
YuanTingHsieh: [P2] Label the evaluation phases in the console table

The artifact separation fixes the storage issue, but this line still displays the post-training values while _handle_after_aggregation() prints the aggregate of the incoming-model values directly underneath them.

For example, with equal client weights:

| Client | Incoming-model accuracy (FLModel.metrics) | Post-training accuracy (PROGRESS_METRICS) |
| --- | --- | --- |
| site-1 | 0.2 | 0.8 |
| site-2 | 0.4 | 0.9 |

The aggregate is correctly 0.3, because it averages 0.2 and 0.4. However, the actual console output at this commit is:

```text
Client       accuracy
site-1            0.8
site-2            0.9
---------------------
Aggregated        0.3
```

That layout makes 0.3 look like the aggregate of 0.8 and 0.9. This applies when an executor supplies both metric phases to a workflow that also reports aggregated metrics; it is not a claim that XGBoost bagging itself averages those accuracy values.

Please distinguish the phases in the console, for example by showing incoming-model client values with their aggregate and putting post-training values in a separately labeled section. Keep the existing aggregate calculation. A console-output regression should check that the two phases remain distinguishable.

YuanTingHsieh (commented):
Thanks for separating the shared weight-diff preflight into #5327 and fixing the site-only evaluation summary. The artifact now preserves incoming-model metrics and post-training progress metrics separately.

I'm okay keeping the additional tree-based XGBoost evaluation in this PR with its unconditional runtime cost documented; I am no longer requesting that split as a condition of this review.

The remaining concern is the console presentation described inline. The aggregate calculation and artifact separation are correct, but the console still places values from different evaluation phases in one unlabeled table.

nvflare/app_common/widgets/metrics_artifact_writer.py
chesterxgchen: Fixed in signed commit 6a4ad9c9f. When a contribution supplies both `FLModel.metrics` and `PROGRESS_METRICS`, the console now renders the incoming-model values first, keeps the aggregate directly with that phase, and renders the post-training values in a second labeled section:

```text
  Incoming-model client metrics
  Client                          accuracy
  site-1                               0.2
  site-2                               0.4
  ------------------------------------------------------------------
  Aggregated                           0.3

  Post-training client metrics
  Client                          accuracy
  site-1                               0.8
  site-2                               0.9
```

Aggregation inputs and math are unchanged, and the artifact still stores the two phases separately as `metrics` and `progress_metrics`. The console regression uses the values from the report and asserts that `0.2`, `0.4`, and aggregate `0.3` appear only in the incoming section, while `0.8` and `0.9` appear only in the post-training section. Validation: `109 passed, 3 skipped`; full project style passed.

holgerroth (approved):
Approved at 6a4ad9c9f after re-reviewing the follow-up fixes. The XGBoost normalized-metric issue is fixed, metric phases are separated in artifacts and clearly labeled in console output, all review threads are resolved, targeted validation passed (109 passed, 3 skipped), and the full CI matrix is green.

PR #5321, opened 2026-09-18
Add Kaplan-Meier progress reporting

chesterxgchen:
## Problem

The Kaplan-Meier example runs several potentially long collection, aggregation, encryption, and distribution phases. Under NVFlare's `progress` log mode, users receive no indication of which phase is active and cannot distinguish useful work from a stalled job.

The example must also avoid reporting success when the final global survival curve was not acknowledged by every targeted client or when the run was aborted.

## Change

- Add concise progress messages for the cleartext and homomorphic-encryption Kaplan-Meier workflows.
- Sample the final distribution targets explicitly and pass that same list to `send_model_and_wait()`.
- Print completion only when every final target responds and the abort signal is clear.
- Keep the cleartext and HE modules self-contained; the HE job does not import implementation details from `server.py`.
- Use only public controller operations and defensive abort-signal access rather than `_current_num_targets` or another private controller field.

Representative cleartext output:

```text
  Federated Kaplan-Meier survival analysis

  Collecting local histograms…
  Aggregating survival histograms…
  Distributing the global survival curve…

  ✓ Survival analysis completed
```

The HE workflow reports histogram-bound collection, encrypted histogram collection and aggregation, global-curve distribution, and completion.

## Validation

- `6 passed` for cleartext/HE completion, partial-response, and abort paths.

Targets NVFlare 2.10.


chesterxgchen:
Superseded by #5319. Kaplan–Meier is part of the intended cross-workflow progress-logging scope, so signed commit `301378caf` restores the cleartext/HE phases, completion guards, documentation, and six focused tests directly to that PR. Separate follow-ups remain limited to example defects that exist independently of progress logging.

PR #5322, opened 2026-09-18
Handle Python scalars in dataframe statistics

chesterxgchen:
## Problem

Dataframe statistics unconditionally call `.item()` on the results of pandas `sum()`, `std()`, and variance calculations. Depending on the pandas version and dataframe dtype, these operations can return an ordinary Python `float`, which has no `.item()` method. The federated-statistics client then fails before it can return its statistics.

This surfaced while running the `hello-tabular-stats` example with current pandas.

## Change

Convert NumPy scalar results with `.item()` while leaving ordinary Python scalar values unchanged. This preserves the existing wire values and supports both pandas return forms.

## Validation

- Regression coverage exercises sum, standard deviation, and variance when the dataframe operations return Python floats.
- Statistics tests: `1 passed, 12 skipped` (optional `fastdigest` cases skipped on this platform).

Targets NVFlare 2.10.


nvkevlu (approved):
LGTM

PR #5323, opened 2026-09-18
Avoid logging PSI site sizes and identities

chesterxgchen:
## Problem

The DH-PSI workflow writes raw preparation responses and several derived structures to diagnostic logs. Those values associate each site identity with its private input-set size and later intersection sizes. Because diagnostic logs are retained in the result bundle, the association remains available after the run.

The logs are useful for troubleshooting, so removing the diagnostic checkpoints entirely would make PSI failures harder to investigate.

## Change

Keep the existing preparation, ordering, forward-pass, backward-pass, intersection, validation, and timing diagnostics, but replace raw site/size structures with:

- participant and response counts;
- the protocol phase that completed;
- whether the forward or final-size validation failed; and
- the existing forward/backward timing measurements.

Every DH-PSI request also uses the shared broadcast operator's client-name-redaction option. This removes participant identities at the real result-callback boundary while preserving the operator's existing identity-bearing diagnostics for non-PSI workflows. When redaction is enabled, the callback clones its `FLContext` for logging and removes the peer context from that clone, preventing `generate_log_message()` from restoring the participant identity in the log prefix. The live callback context is not mutated and remains available for result processing.

`PSIController` also supplies a neutral participant label to the shared error handler. Within the DH-PSI workflow, broadcast callback, and PSI-controller diagnostics changed here, success records, logged errors, panic messages, and raised exceptions therefore use a neutral participant label. Forward-pass diagnostics describe the number of retained intermediate-result holders rather than mislabeling that value as a participant count. The callback's pre-existing direct `print()` is removed so all remaining output follows the configured logging path.

The operator records its client-name logging policy on each task. If a result callback raises, `WFCommServer` applies that task policy to its own exception log and removes the peer context from a logging-only clone. This closes the wrapper-level failure path without disabling participant diagnostics for any task that uses the default policy.

Validation errors retain the failed invariant and response count without embedding site identities or set sizes. PSI computation, ordering, messages, and returned results are unchanged.

This is not a global participant-identity anonymity mode. Framework task-receipt, completion, and audit records outside the callback retain peer identities, and protocol task ordering can reveal relative size ordering. Those records predate this change and remain available for operational audit. The privacy boundary here is the removal of raw set/intersection sizes and identity-to-size structures from DH-PSI workflow and callback diagnostics. The separate client-side duplicate-input exception that can include raw input values is tracked in #5326.

All participant-name- and private-size-keyed response maps are validated before access. Missing responses produce phase-only incomplete-response errors, and malformed DXO payloads produce phase-only malformed-response errors. Initial preparation verifies every connected participant response before ordering sizes; valid zero-sized responses retain the prior behavior of excluding empty participants from that worklist, while an all-empty federation stops with a privacy-safe error instead of reaching an indexing failure. The same validation covers the forward pairwise stages, backward setup, request processing, the nested backward response map, and intersection calculation.

## Validation

- `25 passed` across the complete PSI unit suite; `28 passed` when combined with the focused workflow-communicator tests.
- Regression coverage verifies that preparation, phase, and validation diagnostics contain useful counts but no sentinel site names or sizes.
- Real successful, aborting, and non-aborting `ClientTask` callbacks include a peer identity and inspect actual emitted log records. PSI suppresses the identity from message text and the generated context prefix, while generic defaults still include it.
- The callback tests also verify that creating the logging-only context does not alter the live callback peer context.
- Partial and malformed response-map regressions verify that exceptions contain neither sentinel participant identities nor private sizes.
- An empty-participant regression verifies that a valid zero-sized response is filtered from PSI processing without exposing its identity or size.
- An all-empty regression verifies an explicit abort and privacy-safe error, and a nested-response regression prevents partial backward results from reaching intersection processing.
- Real `WFCommServer` submission tests cover both broadcast and multicast tasks. Callback exception logs follow the PSI redaction policy, while default tasks continue to include the client identity.

Targets NVFlare 2.10.


chesterxgchen:
Addressed Review #536 in signed commit `51d9b4684`:

- `BroadcastAndWait` now has an opt-in client-name logging control whose default preserves existing diagnostics for every non-PSI caller.
- Every DH-PSI broadcast and multicast creates the operator with client-name logging disabled, so participant identities are removed at the actual `ClientTask` result callback rather than only from workflow-level messages.
- The regression invokes the real callback with a sentinel client identity and private size, verifies neither reaches its log messages, and also verifies the generic default still logs the identity.
- The forward record now reports the number of retained intermediate-result holders instead of incorrectly calling that dictionary size a processed-participant count.

Validation: `11 passed` across the PSI unit suite; the full Black, isort, flake8, and agent-skill style check passes.

chesterxgchen:
Addressed Review #540 in signed commit `fd612566c` while keeping the privacy policy PSI-specific:

- The shared `ErrorHandlingController` gains only a backward-compatible optional display label; all existing callers still default to the real client name.
- `PSIController` supplies `a PSI participant`, so non-OK result logs, `system_panic` messages, and raised exceptions omit the participant identity.
- The common `BroadcastAndWait` operator continues to log client identities by default. DH-PSI alone opts out for every success callback.
- Real callback tests cover both an aborting `EXECUTION_EXCEPTION` and a non-aborting `EXECUTION_RESULT_ERROR`, along with the successful-result and generic-default cases.

Validation: `14 passed` across the PSI and common error-handler tests; the full Black, isort, flake8, and agent-skill style check passes.

chesterxgchen:
Removed the pre-existing raw `print("task_name", task_name)` from `BroadcastAndWait.results_cb()` in signed commit `b18ae23d7`. Task processing continues through the existing configured logger, so PSI redaction and normal logging filters are no longer bypassed by direct stdout output.

Validation: the PSI unit suite passes (`13 passed`), `git diff --check` is clean, and the full Black/isort/flake8/agent-skill style check passes.

nvflare/app_common/workflows/broadcast_operator.py
```diff
@@ -93,12 +94,13 @@ def update_result(self, client_name: str, dxo: DXO):
...
-        print("task_name", task_name)
-        self.log_info(fl_ctx, f"Processing {task_name}, {self.task} result from client {client_name}")
+        processing_client_detail = f" from client {client_name}" if self.log_client_names else ""
+        self.log_info(fl_ctx, f"Processing {task_name}, {self.task} result{processing_client_detail}")
         result = client_task.result
         rc = result.get_return_code()
         if rc == ReturnCode.OK:
-            self.log_info(fl_ctx, f"Received result from client:{client_name} for task {task_name} ")
+            received_client_detail = f" from client:{client_name}" if self.log_client_names else ""
+            self.log_info(fl_ctx, f"Received result{received_client_detail} for task {task_name} ")
```
IsaacYangSLA: **[P2] Redact the peer context as well as the PSI message text**

`log_client_names=False` removes the participant name from the message body, but these calls still pass the original callback `FLContext` to `log_info()`. Real submissions attach the participant's peer context in `server_commands.py:238`, preserve it through `server_runner.py:559` and `wf_comm_server.py:507`, and `FLComponent` calls `generate_log_message()`, which prepends the peer identity again.

I reproduced this at `b18ae23d7ad2d3868a3aeba42cb356760e987196` by invoking the actual PSI broadcast callback with a peer context named `private-site-alpha` and capturing the actual logger:

```text
[identity=server, run=?, peer=private-site-alpha, peer_run=?]: Received result for task PSI
[identity=server, run=?, peer=private-site-alpha, peer_run=?]: Execution failed for a PSI participant
```

The identity remains for successful and failed callbacks (`OK`, `EXECUTION_EXCEPTION`, and `EXECUTION_RESULT_ERROR`). The new tests mock `log_info()`/`log_error()` and omit peer context, so they miss the emitted prefix.

Please sanitize a logging-specific context or add context-aware redaction for these PSI records and the shared error path, without mutating the live callback context, and test actual emitted records. This is an incomplete implementation of the new redaction option; the underlying context-based identity leak predates this PR.

chesterxgchen: Fixed in signed commit `cf72227ce`. When client-name logging is disabled, `BroadcastAndWait` now clones the callback `FLContext` for logging and removes `ReservedKey.PEER_CTX` from that clone. Both success records and the shared error path use the logging-only context, so `generate_log_message()` cannot reintroduce the participant identity in its `peer=...` prefix. The live callback context is not mutated and still retains its original peer context for result processing.

The regressions now attach a real peer context named `private-site-alpha` and inspect actual logger output rather than mocked `log_info()` / `log_error()` calls. They cover `OK`, aborting `EXECUTION_EXCEPTION`, and non-aborting `EXECUTION_RESULT_ERROR`, verify the identity is absent from emitted records and exceptions, verify the neutral PSI label remains, and verify the live callback peer context is unchanged. The generic `log_client_names=True` case still includes the peer identity.

Validation: the complete PSI unit suite passes (`13 passed`), `git diff --check` is clean, and the full Black/isort/flake8/agent-skill style check passes.

holgerroth: Rechecked this finding against the current head, `091323145d84e4623c59b285f2615f787bc36179`. The logging-context change addresses the peer-identity leak I reproduced on `b18ae23`: the success records and shared error handler now use a cloned context with the peer context removed, while the original callback context remains intact.

I reran the PSI suite and common error-handler test: **14 passed**. The updated callback regressions capture actual emitted records with a populated peer identity, cover `OK`, `EXECUTION_EXCEPTION`, and `EXECUTION_RESULT_ERROR`, and verify that generic identity logging and the original peer context are preserved.

This finding is resolved in the current revision.

nvflare/app_common/psi/dh_psi/dh_psi_workflow.py
```diff
@@ -179,7 +172,7 @@ def pairwise_requests(self, ordered_sites: List[SiteSize], setup_msgs: Dict[str,
             inputs[PSIConst.SETUP_MSG] = setup_msgs[s.name]
```
nvkevlu, line 176: [P1] Validate sensitive-keyed response maps before indexing them

`setup_msgs` contains only successfully collected responses. If a participant's response is absent, this lookup raises `KeyError('<site name>')`; normal workflow exception logging includes that exception text when secure logging is not enabled, exposing the identity this PR is intended to redact. The name-keyed lookups at lines 190, 208, 298, and 363 have the same issue, while line 264 can raise a `KeyError` containing a private item/intersection size. Please validate response completeness and raise a value-free error before these lookups, with coverage for partial and malformed response maps.

chesterxgchen: Fixed in signed commit `fcbbdf0b5`. DH-PSI now validates every expected response key and required DXO payload field through centralized helpers before performing participant-name- or private-size-keyed access. Missing response maps raise phase-only `incomplete` errors; invalid map/DXO/payload shapes raise phase-only `malformed` errors. No missing key or private value is interpolated into either message.

The validation covers initial preparation, all four pairwise forward stages, backward setup keyed by private sizes, request creation/processing, setup preparation, and intersection calculation. Initial preparation also verifies that every connected participant responded before ordering the sizes. Regressions cover a missing participant, missing pairwise input, missing private-size key, partial single-target result, and malformed DXO payload, and assert that neither sentinel identities nor sizes appear. The complete PSI suite passes (`18 passed`) and the full project style check passes.

nvflare/app_common/psi/dh_psi/dh_psi_workflow.py
```diff
@@ -347,24 +352,54 @@ def prepare_sites(self, abort_signal):
...
+        response_count = len(results) if isinstance(results, dict) else 0
+        self.log_info(self.fl_ctx, f"{PSIConst.TASK_PREPARE} received {response_count} participant responses")
         if not results:
             abort_signal.trigger("no items to perform PSI")
             raise RuntimeError("There is no item to perform PSI calculation")
         else:
-            self.ordered_sites = self.get_ordered_sites(results)
+            expected_clients = [client if isinstance(client, str) else client.name for client in clients]
+            sizes = self._response_values(results, expected_clients, PSIConst.ITEMS_SIZE, "PSI preparation")
+            if any(not isinstance(size, Integral) or size <= 0 for size in sizes.values()):
```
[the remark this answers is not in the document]
chesterxgchen: Fixed in 8063488a2. Zero is again treated as a valid preparation response and the participant is omitted from the ordered PSI worklist, preserving the existing empty-site behavior. Negative and non-integral sizes remain malformed, and complete response-map validation still runs before ordering. The regression verifies that one empty participant is filtered while the aggregate response-count log contains neither participant identities nor private sizes. Validation: 19 PSI tests and the full project style check passed.

chesterxgchen:
Addressed the latest review findings in signed commit `9c8a7d737`:

- `BroadcastAndWait` records its client-name logging policy on both broadcast and multicast tasks. When a result callback raises, `WFCommServer` now applies that policy to its wrapper-level exception log and removes the peer context from a logging-only clone. Real communicator submission tests verify that PSI omits the identity while the default policy retains it.
- The backward-pass response extracted from `RESPONSE_MSG` is validated against every request key before intersection tasks can be created. Partial nested maps now return a phase-only incomplete-response error with no participant identities or private values.
- Valid zero-sized participant responses continue to be filtered as before. If every participant is empty, preparation triggers the abort signal and raises the existing privacy-safe no-items error instead of reaching `ordered_sites[0]`.

Validation: `25 passed` across the PSI suite, `28 passed` with the focused communicator tests, and the full Black/isort/flake8/agent-skill style check passed.

IsaacYangSLA:
**PSIRT follow-up: duplicate PSI inputs expose raw identifiers in client error logs (pre-existing; provisional medium severity)**

**Not introduced by PR #5323.** This issue already exists in the PR’s base revision, and the affected code is unchanged by this PR. It is a separate security follow-up, not a regression or a merge-blocking finding against this PR.

Reviewed at `9c8a7d737a6ae74f1daf87bcac457ee7e24cddf0`. This is outside the changed files and is not a newly introduced regression, but it remains relevant to the privacy of retained PSI logs.

[`check_items_uniqueness()`](https://github.com/NVIDIA/NVFlare/blob/9c8a7d737a6ae74f1daf87bcac457ee7e24cddf0/nvflare/app_opt/psi/dh_psi/dh_psi_task_handler.py#L30-L33) includes duplicate item strings and occurrence counts in a `ValueError`. The actual PSI executor's [exception handler](https://github.com/NVIDIA/NVFlare/blob/9c8a7d737a6ae74f1daf87bcac457ee7e24cddf0/nvflare/app_common/executors/error_handling_executor.py#L89-L91) records that traceback. With `NVFLARE_SECURE_LOGGING` unset, synthetic duplicate input produced:

```text
ValueError: the items must be unique, the following items with duplicates {'PRIVATE_PATIENT_SENTINEL_7419': 2}
```

The same check with `NVFLARE_SECURE_LOGGING=true` suppressed the identifier. The reproduction used the exact standalone checker extracted from its source AST through the real `PSIExecutor` error path, with a substituted task-handler adapter. The optional native PSI dependency was unavailable in the Python 3.14 `.venv`; this failure occurs before native cryptographic operations.

This can cross the site boundary: the provisioning template configures `SiteLogStreamer` for `error_log.txt`, and `allow_log_streaming` defaults to true. Stored logs can be retrieved using job-authorized APIs. Thus an ordinary duplicate-input error can disclose raw identifiers to the server and authorized diagnostic readers who may not be entitled to the underlying PSI inputs. Streaming/access controls were traced in source, not exercised in a deployed federation; no unauthenticated access is claimed.

Please track a separate fix that replaces the value-bearing exception with a constant message or aggregate duplicate count, and adds an actual executor-log regression. This is a [CWE-532](https://cwe.mitre.org/data/definitions/532.html) follow-up; the current server-log changes do not address it. Secure logging mitigates this specific exception path, while previously retained logs remain unaffected.

IsaacYangSLA:
**PSIRT scope clarification: outer server logs still expose PSI identities and relative size ordering (pre-existing; low, requirement-dependent impact)**

**Not introduced by PR #5323.** This issue already exists in the PR’s base revision, and the affected code is unchanged by this PR. It is a separate security follow-up, not a regression or a merge-blocking finding against this PR.

Reviewed at `9c8a7d737a6ae74f1daf87bcac457ee7e24cddf0`. The earlier callback-prefix finding is fixed. However, the redaction guarantee does not extend to all server diagnostics.

I exercised the real `ServerRunner -> WFCommServer -> BroadcastAndWait` submission path with the task's client-name logging policy set to false. The callback records correctly omit the participant identity, but [`ServerRunner.process_submission()`](https://github.com/NVIDIA/NVFlare/blob/9c8a7d737a6ae74f1daf87bcac457ee7e24cddf0/nvflare/private/fed/server/server_runner.py#L474) still logs `got result from client private-site-alpha` before the callback, and [the completion record](https://github.com/NVIDIA/NVFlare/blob/9c8a7d737a6ae74f1daf87bcac457ee7e24cddf0/nvflare/private/fed/server/server_runner.py#L562) retains `peer=private-site-alpha`. The sentinel numeric size `91001` did not appear. Only the engine, scheduler, and task manager were substituted; the logging path was real.

There is also residual ordering information: the workflow sorts participants by input size and sends its [first pairwise setup](https://github.com/NVIDIA/NVFlare/blob/9c8a7d737a6ae74f1daf87bcac457ee7e24cddf0/nvflare/app_common/psi/dh_psi/dh_psi_workflow.py#L145-L157) to the smaller half. A two-site probe confirmed that the outer submission logs identify that first setup participant. A reader who correlates identities with protocol phases can infer relative ordering, although equality cannot be excluded from order alone and exact cardinalities are not disclosed.

These outer logging paths predate the PR; this is a scope limitation, not a recurrence of the fixed callback bug or a demonstrated cryptographic break. Please explicitly limit the privacy claim to the changed workflow/callback diagnostics and removal of raw size structures. If the intended requirement is identity/ordering privacy throughout retained logs, track the remaining task-assignment, submission, and late/duplicate-result logs and add an entire-run log-capture test, while preserving required audit behavior.

Overall review validation: **59 tests passed** using the project `.venv`, including 12 complete orchestration cases with simulated PSI set operations. Native cryptographic PSI and deployed log retrieval were not tested. No newly introduced blocking security defect was established in this revision; these two PSIRT comments document residual pre-existing exposure and follow-up scope.

chesterxgchen:
Addressed both PSIRT follow-ups.

- The client-side duplicate-input exposure is independent of this server workflow/callback change. I created #5326 for the constant/aggregate-only validation error and real executor-log regression, and assigned it to the 2.10 milestone.
- The PR description now states the privacy boundary explicitly. #5323 removes raw set/intersection sizes and identity-to-size structures from the DH-PSI workflow, callback, and PSI-controller diagnostics changed here. It does not claim global participant anonymity: framework task lifecycle/audit records may retain peer identities, and protocol task ordering can reveal relative ordering. Those pre-existing audit paths remain unchanged.

No additional #5323 product-code change was made for either follow-up because both comments identify behavior outside this PR's changed privacy boundary.

IsaacYangSLA (approved):
All discovered issues were addressed.

PR #5330, opened 2026-09-21
Do not include duplicate item values in DH-PSI validation error

### LINKED ISSUE #5326: DH-PSI duplicate-input errors expose input identifiers in diagnostic logs
## Problem

DH-PSI rejects duplicate input items in `nvflare/app_opt/psi/dh_psi/dh_psi_task_handler.py`, but the current `ValueError` includes the raw duplicate values and their occurrence counts. `ErrorHandlingExecutor` records the resulting traceback in client diagnostic logs. With secure logging disabled, an ordinary input-validation failure can therefore expose PSI input identifiers in retained or streamed logs.

Example shape of the current diagnostic:

```text
ValueError: the items must be unique, the following items with duplicates {'PRIVATE_PATIENT_SENTINEL': 2}
```

This behavior predates #5323. That PR removes identity-to-size structures from server-side DH-PSI workflow and callback diagnostics; it does not change this client-side validation path. Secure logging mitigates the traceback content when enabled, but the default error should not contain raw PSI items.

## Expected behavior

- Reject duplicate items without embedding their values in the exception.
- Retain actionable diagnostics, such as a constant message or aggregate duplicate count.
- Add a regression through the real PSI executor error/logging path that uses a sentinel duplicate identifier and verifies the identifier is absent from emitted records.
- Preserve the existing duplicate-input rejection behavior and exit status.

Denimworld12:
Fixes #5326.

## Problem

`check_items_uniqueness` in `nvflare/app_opt/psi/dh_psi/dh_psi_task_handler.py` rejected duplicate PSI inputs by building a mapping of the duplicate values to their occurrence counts and formatting that mapping into the `ValueError`:

```
ValueError: the items must be unique, the following items with duplicates {'PRIVATE_PATIENT_SENTINEL': 2}
```

`PSIExecutor` extends `ErrorHandlingExecutor`, which records the resulting traceback through `log_exception`. `log_exception` formats the traceback with `secure_format_traceback`, and that returns the full `traceback.format_exc()` unless `NVFLARE_SECURE_LOGGING` is set. Secure logging is opt-in and unset by default, so in a default deployment an ordinary input-validation failure writes raw PSI input identifiers into retained client diagnostic logs.

This is the client-side path left open by #5323, which changed server-side workflow and callback diagnostics only and does not touch this module.

## Change

The duplicate check now counts how many distinct items occur more than once and reports that count instead of the values:

```
ValueError: the items must be unique, found 2 items with duplicates
```

The mapping keyed by input values is no longer constructed at all, so no structure holding the private items remains for later code to log. The rejection itself, the `ValueError` type, and the resulting `ReturnCode.EXECUTION_RESULT_ERROR` are unchanged. This also brings the module in line with its siblings in the same package, `dh_psi_client.py` and `dh_psi_server.py`, which already raise constant messages with no input values embedded.

## Validation

`tests/unit_test/app_opt/psi/dh_psi/dh_psi_task_handler_test.py` adds five regressions using sentinel identifiers: unique items are still accepted, duplicates are still rejected, the exception message contains neither sentinel, the aggregate count is still reported without a mapping, and the formatted traceback contains no sentinel when `NVFLARE_SECURE_LOGGING` is explicitly unset. That last case covers the configuration in which the leak is actually reachable.

Reverting only the source change while keeping the tests fails three of the five, with the sentinel visible in the traceback assertion, so the tests guard the reported behaviour rather than restating the new implementation.

`45 passed` across `tests/unit_test/app_opt/psi` and `tests/unit_test/app_common/psi`. `./runtest.sh -l` and `./runtest.sh -s` both pass, covering license headers, Black, isort, flake8 and agent-skill lint.

## One open question on test placement

The issue asks for the regression to run through the real PSI executor error and logging path. `dh_psi_task_handler` imports `PSIClient` and `PSIServer`, which import the optional native `private_set_intersection` package, so the module cannot be imported where that package is unavailable. The existing `tests/unit_test/app_opt/psi/dh_psi/dh_psi_test.py` has its PSI cases commented out for that reason, noting the optional-requirement problem and a missing `GLIBC_2.29` for `_openmined_psi.so`.

I stubbed `private_set_intersection` in `sys.modules` before importing the handler, because it is the only option that exercises the real function on all ten CI configurations. `pytest.importorskip` would read more idiomatically but would skip in CI, leaving the regression unguarded where it matters.

If you would rather the validation moved into a module that imports without the optional dependency, I am happy to restructure it that way instead. I raised this on #5326 and did not want to block the fix on the answer.

Targets NVFlare 2.10.


Denimworld12:
Addressed both points from the review.

The stub is now scoped to the import and sys.modules gets restored afterwards, so it doesn't leak into other tests. I added a small test that checks this.

The count assertion compares the full message now instead of a substring.

I re-checked the regression coverage against the unfixed code and 3 of the 6 tests still fail there, so it's intact.

tests/unit_test/app_opt/psi/dh_psi/dh_psi_task_handler_test.py
```diff
@@ -0,0 +1,96 @@
...
+    def test_traceback_excludes_item_values_when_secure_logging_is_off(self, monkeypatch):
+        # Secure logging is opt-in, so an unset NVFLARE_SECURE_LOGGING is the
+        # configuration in which log_exception records the full traceback.
+        monkeypatch.delenv("NVFLARE_SECURE_LOGGING", raising=False)
+
+        traceback_text = None
+        try:
+            check_items_uniqueness([DUPLICATE_SENTINEL, DUPLICATE_SENTINEL])
+        except ValueError:
+            traceback_text = secure_format_traceback()
```
chesterxgchen: The issue acceptance criteria call for a regression through the real PSI executor error/logging path and preservation of the exit status. This test calls the validator and `secure_format_traceback()` directly, so it would still pass if `PSIExecutor.execute()` logged the sentinel through another message or returned a different code. Please drive duplicate input through `PSIExecutor` with the real `DhPSITaskHandler`, capture the emitted log record, assert that the sentinel is absent, and assert `ReturnCode.EXECUTION_RESULT_ERROR`.

tests/unit_test/app_opt/psi/dh_psi/dh_psi_task_handler_test.py
```diff
@@ -0,0 +1,96 @@
...
+    saved = {name: sys.modules.get(name) for name in _STUBBED_MODULES}
+    for name in _STUBBED_MODULES:
+        sys.modules[name] = MagicMock()
+
+    before = set(sys.modules)
+    try:
+        module = importlib.import_module("nvflare.app_opt.psi.dh_psi.dh_psi_task_handler")
+        return module.check_items_uniqueness
+    finally:
+        for name in set(sys.modules) - before:
```
chesterxgchen: This cleanup removes newly imported entries from `sys.modules`, but it does not restore submodule attributes attached to an already-loaded parent package. If `nvflare.app_opt.psi.dh_psi` was imported before this helper, its `dh_psi_client` and `dh_psi_server` attributes can remain bound to modules whose `psi` globals are `MagicMock` objects after cleanup. A later `from nvflare.app_opt.psi.dh_psi import dh_psi_client` can therefore receive the stubbed module, making the suite order-dependent. The current cleanup assertion checks only the two `private_set_intersection` keys in `sys.modules`, so it misses this state. Please explicitly restore or remove the affected parent-package attributes and limit cleanup to the targeted modules, or move the validator into a dependency-free module that can be imported without these stubs.

tests/unit_test/app_opt/psi/dh_psi/dh_psi_task_handler_test.py
Denimworld12: Done. `TestPSIExecutorDuplicateInput` now drives duplicate input through `PSIExecutor` with a real `DhPSITaskHandler`, captures the emitted records, and asserts both that the sentinel is absent and that the reply carries `ReturnCode.EXECUTION_RESULT_ERROR`. I added a companion case asserting that unique input still returns `ReturnCode.OK`, so the exit status is pinned in both directions rather than only on the failure path.

Checked that it guards the reported behaviour: reverting only the source change and keeping the tests fails this case, with the sentinel visible in the captured log.

tests/unit_test/app_opt/psi/dh_psi/dh_psi_task_handler_test.py
Denimworld12: You are right, and it was worse than the parent-attribute case alone: the cleanup assertion I added only inspected the two `private_set_intersection` keys, so it would not have detected that state either. It gave false confidence.

Rather than patch the cleanup, I removed the stubbing. `openmined.psi==2.0.6` is a declared dependency in `setup.cfg` under the `PSI` extra, reached through `app_opt` -> `all` -> `test` -> `dev`, so the handler imports without any stub. The file now uses `pytest.importorskip("private_set_intersection")`, which skips only on Python 3.14, where the environment marker excludes the dependency. There is no longer any `sys.modules` manipulation in the test.

For context on why the stub was there at all: the disabled cases in `dh_psi_test.py` refer to the optional requirement being unavailable, but that note describes a `GLIBC_2.29` failure on the old Ubuntu 18.04 Jenkins image rather than the current runners. I took it at face value instead of checking whether the dependency installs today.

tests/unit_test/app_opt/psi/dh_psi/dh_psi_task_handler_test.py
```diff
@@ -0,0 +1,107 @@
...
+        executor = _executor_with_items([DUPLICATE_SENTINEL, DUPLICATE_SENTINEL, OTHER_SENTINEL])
+
+        # Secure logging is opt-in, so an unset NVFLARE_SECURE_LOGGING is the
+        # configuration in which the full traceback reaches the log.
+        with caplog.at_level(logging.DEBUG):
+            reply = executor.execute(PSIConst.TASK, _prepare_shareable(), FLContext(), Signal())
+
+        assert reply.get_return_code() == ReturnCode.EXECUTION_RESULT_ERROR
+        assert DUPLICATE_SENTINEL not in caplog.text
+        assert OTHER_SENTINEL not in caplog.text
```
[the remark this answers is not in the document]
Denimworld12: Good catch, and it was a real hole rather than a theoretical one.

I checked it against the unfixed source: with `NVFLARE_SECURE_LOGGING=true` the traceback is sanitized, the sentinel never reaches the log, and this test passed while the leak was fully present. Only the direct message assertion failed. So in an environment with secure logging enabled the regression test was giving false confidence.

The cause was mine: the earlier revision unset the variable, and when I reworked the file to drive the real executor path I kept the comment explaining why secure logging has to be off but dropped the line that did it.

Both executor tests now call `monkeypatch.delenv("NVFLARE_SECURE_LOGGING", raising=False)`. Verified in both directions: against the unfixed source the duplicate-input test fails with secure logging on and off, and with the fix all five pass in both modes.

chesterxgchen (approved):
LGTM

